PR #13, opened 2021-07-29
Added ONNXRuntime as dependency

saganatt:
Requires alidist from https://github.com/alisw/alidist/pull/3197

PR #20, opened 2021-08-18
Migrating Analysis to O2Physics

jgrosseo:


ktf:
@TimoWilken this seems to be stuck.

TimoWilken:
It took 20 minutes between the PR being marked as non-draft and being picked up because the builder was rebuilding #13 in the meantime.

jgrosseo:
Thanks for the explanation, @TimoWilken 

If I look at the build log, it seems to me that the CI builds and tests O2 (which it shouldn't) and I don't see anything of building O2Physics.

https://ali-ci.cern.ch/alice-build-logs/AliceO2Group/O2Physics/20/5997e0c8f481cbd1f8acfa617fe8a8f9de2fbff2/build_O2Physics_o2/fullLog.txt

PR #26, opened 2021-08-20
Event Filtering: Add first version of HF triggers

fgrosa:
Equivalent of https://github.com/AliceO2Group/AliceO2/pull/6619


mpuccio (dismissed):
As the code is the same that was in O2 and was already in shape to be merged and tested

mpuccio (approved):

PR #28, opened 2021-08-20
PWGHF: Fix doMC option

lucamicheletti93:
Fix doMC option to "true" in taskX and taskJpsi


vkucera:
@lucamicheletti93 Why is this needed?
See https://github.com/AliceO2Group/O2Physics/pull/27

PR #39, opened 2021-08-25
Speedup hf 3 prong combinatorics

ktf:


ktf:
@ginnocen at the moment what really seem to happen in PbPb is that we have either an infinite loop (for tiny tracks tables) or we are simply slow. This **should** protect the infinite loop and speed up things a bit. The next performance issue in this bit of code is the lookup by string in the Label2D.

fgrosa:
hi @ktf thanks a lot! You could implement the same improvement for the speedup in the method for 2-prong preselections
https://github.com/AliceO2Group/O2Physics/blob/29538ee5dcd878b904b03f5a650ee09a4a8664de/Tasks/PWGHF/HFTrackIndexSkimsCreator.cxx#L547 
or if you prefer I can do it.

PR #42, opened 2021-08-26
PWGHF: Add PID flags for the J/ψ task

zhangbiao-phy:
PWGHF: add the pid flag for jpsi task

### DIFF
--- a/Common/Core/TrackSelectorPID.h
+++ b/Common/Core/TrackSelectorPID.h
@@ -434,7 +434,7 @@ class TrackSelectorPID
   {
     bool isSelTOF = false;
     bool isSelRICH = false;
-    bool hasRICH = track.richId() > -1;
+    bool hasRICH = isValidTrackPIDRICH(track);
     bool hasTOF = isValidTrackPIDTOF(track);
     auto nSigmaTOFEl = track.tofNSigmaEl();
     auto nSigmaTOFPi = track.tofNSigmaPi();
@@ -471,7 +471,7 @@ class TrackSelectorPID
         isSelRICH = false; // is selected as pion by RICH
       }
     } else {
-      isSelRICH = false;
+      isSelRICH = true;
     }
 
     return isSelRICH || isSelTOF;
--- a/Common/Core/TrackSelectorPID.h
+++ b/Common/Core/TrackSelectorPID.h
@@ -434,7 +434,7 @@ class TrackSelectorPID
   {
     bool isSelTOF = false;
     bool isSelRICH = false;
-    bool hasRICH = isValidTrackPIDRICH(track);
+    bool hasRICH = track.richId() > -1;
     bool hasTOF = isValidTrackPIDTOF(track);
     auto nSigmaTOFEl = track.tofNSigmaEl();
     auto nSigmaTOFPi = track.tofNSigmaPi();
@@ -471,7 +471,7 @@ class TrackSelectorPID
         isSelRICH = false; // is selected as pion by RICH
       }
     } else {
-      isSelRICH = true;
+      isSelRICH = false;
     }
 
     return isSelRICH || isSelTOF;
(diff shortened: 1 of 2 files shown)


Core/include/AnalysisCore/TrackSelectorPID.h
```diff
@@ -471,7 +471,7 @@ class TrackSelectorPID
         isSelRICH = false; // is selected as pion by RICH
       }
     } else {
-      isSelRICH = false;
+      isSelRICH = true;
     }
```
vkucera, line 475: This does not seem right.

Tasks/PWGHF/taskJpsi.cxx
```diff
@@ -47,6 +47,10 @@ struct TaskJpsi {
   Configurable<bool> d_modeJpsiToMuMu{"d_modeJpsiToMuMu", false, "Perform Jpsi to mu+mu- analysis"};
   Configurable<double> cutYCandMax{"cutYCandMax", -1., "max. cand. rapidity"};
   Configurable<std::vector<double>> bins{"pTBins", std::vector<double>{hf_cuts_jpsi_toee::pTBins_v}, "pT bin limits"};
+  Configurable<int> noPid{"noPid", 0, "Pid Flag for Jpsi"};
+  Configurable<int> tofPid{"tofPid", 0, "Pid Flag for Jpsi"};
+  Configurable<int> richPid{"richPid", 0, "Pid Flag for Jpsi"};
+  Configurable<int> combPid{"combPid", 0, "Pid Flag for Jpsi"};
```
vkucera, line 53: You use the same and wrong description for all parameters.
These are not PID flags. These are selection switches. Why are they `int`?
`noPid` does not switch off PID, it selects with topological cuts. The name is misleading.
The other names are also quite cryptic.

vkucera (changes requested):

Tasks/PWGHF/taskJpsi.cxx
zhangbiao-phy: Hi Vit, thanks for your comments. you're right, I have committed a new version. please have a look.

Core/include/AnalysisCore/TrackSelectorPID.h
zhangbiao-phy: > This does not seem right.
If the tracks p > 0.6(not use tod) and no rich response, It will be rejected if isSelRICH = false; I think this is not what we expected.

vkucera: The function is supposed to tell
> whether a track is identified as electron and rejected as pion by TOF or RICH.

That means that a particle has to pass the electron selection and fail the pion selection in a given detector.
This cannot be satisfied if the detector response is not available or disabled. In that case the track is not selected as electron and therefore the results should be `false`.

What is the behaviour that you would expect?

Tasks/PWGHF/taskJpsi.cxx
```diff
@@ -47,6 +47,10 @@ struct TaskJpsi {
   Configurable<bool> d_modeJpsiToMuMu{"d_modeJpsiToMuMu", false, "Perform Jpsi to mu+mu- analysis"};
   Configurable<double> cutYCandMax{"cutYCandMax", -1., "max. cand. rapidity"};
   Configurable<std::vector<double>> bins{"pTBins", std::vector<double>{hf_cuts_jpsi_toee::pTBins_v}, "pT bin limits"};
+  Configurable<bool> selectedTofPid{"selectedTofPid", false, "select tof PID for Jpsi"};
+  Configurable<bool> selectedRichPid{"selectedRichPid", false, "select rich PID for Jpsi"};
+  Configurable<bool> selectedCombPid{"selectedCombPid", false, "select tof and rich PID for Jpsi"};
+  Configurable<bool> selectedMid{"selectedMid", false, "select MID PID for Jpsi to mu+mu-"};
```
vkucera, line 53: Please unify the configurable names and fix the capitalisation of detector names in the helper strings.

vkucera:
@zhangbiao-phy Any updates?

vkucera (dismissed):

vkucera:
@zhangbiao-phy Please rebase and force push to restart the CI tests.

vkucera (approved):

PR #44, opened 2021-08-27
PWGHF: RICH PID implementation of D0 meson decay daughter and fixed clang-format

skundu692:
RICH PID implementation for D0 decay daughter with correct clang-format

### DIFF
--- a/Core/include/AnalysisCore/TrackSelectorPID.h
+++ b/Core/include/AnalysisCore/TrackSelectorPID.h
@@ -18,7 +18,6 @@
 #define O2_ANALYSIS_TRACKSELECTORPID_H_
 
 #include <TPDGCode.h>
-
 #include "Framework/Logger.h"
 
 /// Class for track selection using PID detectors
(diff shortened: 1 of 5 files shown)


Core/include/AnalysisCore/TrackSelectorPID.h
```diff
@@ -477,6 +476,190 @@ class TrackSelectorPID
...
+    bool ispionselected = false;
+
+    if (!hasTOF) {
+      return ispionselected;
+    }
+
+    if (usedetector == 0) {
+      if (std::abs(nSigmaTOFPi) < mNSigmaTOFMax) {
+        ispionselected = true;
+      } else {
```
vkucera, line 509: Use `switch` for `usedetector`.
Why do you need `ispionselected = true` instead of `return true`?

Core/include/AnalysisCore/TrackSelectorPID.h
```diff
@@ -477,6 +476,190 @@ class TrackSelectorPID
     return isSelRICH || isSelTOF;
   }
 
+  ////////////////////////////////new sourav: is pion/////////////////////////////////
+  template <typename T>
+  bool isPion(const T& track, int usedetector = 0)
```
vkucera, line 481: Add documentation.
Use `enum` for `usedetector` values.

vkucera (changes requested):
General comments:
* I don't think this big code duplication is needed or useful. Why don't you extend the existing tasks as was done in the J/psi code?
* Fix capitalisation and spacing.
* Simplify the code.
* Add comments.
* Fix existing comments (e.g. do not use `////`).

vkucera:
Hi @skundu692 , is this PR still needed?

PR #45, opened 2021-08-27
PWGHF: Adding the information about the prong ordering (Lc -> p K pi or Lc -> pi K p)

DelloStritto:
The prong ordering information is needed to consider only the right signal candidate in the ML training when the PID is used.
@ginnocen @vkucera

### DIFF
--- a/Tasks/PWGHF/HFCandidateCreator3Prong.cxx
+++ b/Tasks/PWGHF/HFCandidateCreator3Prong.cxx
@@ -160,6 +160,7 @@ struct HFCandidateCreator3ProngMC {
     int8_t sign = 0;
     int8_t flag = 0;
     int8_t origin = 0;
+    int8_t swapping = 0;
     int8_t channel = 0;
     std::vector<int> arrDaughIndex;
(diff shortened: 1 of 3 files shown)


Tasks/PWGHF/HFCandidateCreator3Prong.cxx
```diff
@@ -191,17 +193,22 @@ struct HFCandidateCreator3ProngMC {
           flag = sign * (1 << DecayType::LcToPKPi);
 
           //Printf("Flagging the different Λc± → p± K∓ π± decay channels");
+          if (std::abs(arrayDaughters[0].mcParticle().pdgCode()) == kProton) {
+            swapping = 0;
+          } else {
+            swapping = 1;
+          }
```
vkucera, line 200: How about this?
```c++
swapping = int(std::abs(arrayDaughters[0].mcParticle().pdgCode()) != kProton);
```

vkucera (approved):

vkucera:
@jgrosseo @iarsene Can you please give your blessing?

jgrosseo (approved):

PR #47, opened 2021-08-29
PWGHF: Cache lookups in Array2D

ktf:
String lookups are done per candidate, making it extremely slow for
no reason. Caching the position of the indices should give a rather
significant boost to loosely cut setups.


ktf:
@ginnocen this should speed up quite significantly the 3 prong business...

ktf:
(not yet tested)

ktf:
Do not merge. This crashes. I think i missed something trivial. Looking into it.

ktf:
Verified with https://alimonitor.cern.ch/train-workdir/tests/0001/00010515/stdout.log by Sourav. 3x times in that particular case.

ginnocen (approved):

PR #52, opened 2021-08-30
PWGHF: avoid square root in comparison

ktf:


ktf:
Maybe I am doing some trivial algebra / error propagation mistake, but can't we simply avoid the square root (which is expensive)?

ktf:
Also notice that a few lines above there is a .px(), .py() which if I understand correctly uses a dynamic column. It would be probably better to use an expression column if the cut is loose, otherwise you keep recomputing the various sincos...

vkucera (approved):

PR #66, opened 2021-09-06
Fix bugs Selection Daughters, PID Daughters and histos

sartozza:
- Fix bug in Daughters selection V0 (delete const static in isSelecte…dMinimal of Track
- Add possibility to init the histograms for Daughters from Track
- Add PIDSpecies function to put PID cuts on Daughters V0

### DIFF
--- a/PWGCF/FemtoDream/FemtoDreamTrackSelection.h
+++ b/PWGCF/FemtoDream/FemtoDreamTrackSelection.h
@@ -50,12 +50,35 @@ enum TrackSel { kSign,        ///< Sign of the track
 class FemtoDreamTrackSelection : public FemtoDreamObjectSelection<float, femtoDreamTrackSelection::TrackSel>
 {
  public:
+  FemtoDreamTrackSelection() : nPtMinSel(0),
+                               nPtMaxSel(0),
+                               nEtaSel(0),
+                               nTPCnMinSel(0),
+                               nTPCfMinSel(0),
+                               nTPCcMinSel(0),
+                               nTPCsMaxSel(0),
+                               nDCAxyMaxSel(0),
+                               nDCAzMaxSel(0),
+                               nDCAMinSel(0),
+                               nPIDnSigmaSel(0),
+                               pTMin(9999999.),
+                               pTMax(-9999999.),
+                               etaMax(-9999999.),
+                               nClsMin(9999999.),
+                               fClsMin(9999999.),
+                               cTPCMin(9999999.),
+                               sTPCMax(-9999999.),
+                               dcaXYMax(-9999999.),
+                               dcaZMax(-9999999.),
+                               dcaMin(9999999.),
+                               nSigmaPIDMax(9999999.){};
+
   /// Initializes histograms for the task
   /// \tparam part Type of the particle for proper naming of the folders for QA
   /// \tparam cutContainerType Data type of the bit-wise container for the selections
   /// \param registry HistogramRegistry for QA output
   template <o2::aod::femtodreamparticle::ParticleType part, typename cutContainerType>
-  void init(HistogramRegistry* registry);
+  void init(HistogramRegistry* registry, const std::string WhichDaugh = "");
 
   /// Passes the species to the task for which PID needs to be stored
   /// \tparam T Data type of the configurable passed to the functions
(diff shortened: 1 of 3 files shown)


PWGCF/FemtoDream/FemtoDreamTrackSelection.h
```diff
@@ -268,29 +274,29 @@ bool FemtoDreamTrackSelection::isSelectedMinimal(T const& track)
...
-  const static float nClsMin = getMinimalSelection(femtoDreamTrackSelection::kTPCnClsMin, femtoDreamSelection::kLowerLimit);
-  const static float fClsMin = getMinimalSelection(femtoDreamTrackSelection::kTPCfClsMin, femtoDreamSelection::kLowerLimit);
-  const static float cTPCMin = getMinimalSelection(femtoDreamTrackSelection::kTPCcRowsMin, femtoDreamSelection::kLowerLimit);
-  const static float sTPCMax = getMinimalSelection(femtoDreamTrackSelection::kTPCsClsMax, femtoDreamSelection::kUpperLimit);
-  const static float dcaXYMax = getMinimalSelection(femtoDreamTrackSelection::kDCAxyMax, femtoDreamSelection::kAbsUpperLimit);
-  const static float dcaZMax = getMinimalSelection(femtoDreamTrackSelection::kDCAzMax, femtoDreamSelection::kAbsUpperLimit);
-  const static float dcaMin = getMinimalSelection(femtoDreamTrackSelection::kDCAMin, femtoDreamSelection::kAbsLowerLimit);
-  const static float nSigmaPIDMax = getMinimalSelection(femtoDreamTrackSelection::kPIDnSigmaMax, femtoDreamSelection::kAbsUpperLimit);
+  int nPtMinSel = getNSelections(femtoDreamTrackSelection::kpTMin);
```
jgrosseo, line 277: You removed the static. This means this will be executed once per track and therefore will be much slower. Why the static is gone?

sartozza: The function isSelectedMinimal for Track objects is also called in the V0selection class for the daughters.
The number of selections (e.g nPtMinSel) and the minimal values (e.g pTMin) for tracks and V0 daughters are different. 
And with the const static types they were kept set to the Track values, making impossible to obtain candidates V0s.
So I removed in that function all the const static. I first removed const, keeping static but the problem was still there.

jgrosseo: I understand but this will be extremely slow. 

You have to find a different solution then. E.g. a class which is then instantiated twice. Once for tracks and once for V0s. It sets its values once at the beginning, but not for each track/V0

jgrosseo (changes requested):
See inline comments.

PWGCF/FemtoDream/FemtoDreamTrackSelection.h
sartozza: @jgrosseo ok working on it.
Thanks

PWGCF/FemtoDream/FemtoDreamTrackSelection.h
```diff
@@ -268,29 +319,29 @@ bool FemtoDreamTrackSelection::isSelectedMinimal(T const& track)
...
-  const static float nClsMin = getMinimalSelection(femtoDreamTrackSelection::kTPCnClsMin, femtoDreamSelection::kLowerLimit);
-  const static float fClsMin = getMinimalSelection(femtoDreamTrackSelection::kTPCfClsMin, femtoDreamSelection::kLowerLimit);
-  const static float cTPCMin = getMinimalSelection(femtoDreamTrackSelection::kTPCcRowsMin, femtoDreamSelection::kLowerLimit);
-  const static float sTPCMax = getMinimalSelection(femtoDreamTrackSelection::kTPCsClsMax, femtoDreamSelection::kUpperLimit);
-  const static float dcaXYMax = getMinimalSelection(femtoDreamTrackSelection::kDCAxyMax, femtoDreamSelection::kAbsUpperLimit);
-  const static float dcaZMax = getMinimalSelection(femtoDreamTrackSelection::kDCAzMax, femtoDreamSelection::kAbsUpperLimit);
-  const static float dcaMin = getMinimalSelection(femtoDreamTrackSelection::kDCAMin, femtoDreamSelection::kAbsLowerLimit);
-  const static float nSigmaPIDMax = getMinimalSelection(femtoDreamTrackSelection::kPIDnSigmaMax, femtoDreamSelection::kAbsUpperLimit);
+  nPtMinSel = getNSelections(femtoDreamTrackSelection::kpTMin);
```
jgrosseo, line 322: I am confused. Why here the cut value is here still set on **each** track (this will still be slow). You should do that only once somewhere, e.g. in init.

jgrosseo (approved):

PR #70, opened 2021-09-08
Addition of a mixing handler for event mixing

vfeuilla:
Creation of MixingHandler.h and MixingHandler.cxx that:
- Allow to add several mixing variables (centrality, VtxZ, etc...) and the binning used
- Compute a mixing category for a given event
and other functionnalities

Creation of MixingLibrary.h that has a list a predifined mixing configurations to be used in analysis

Implementation of the MixingHandler tool in dileptonMuMu.cxx and tableReader.cxx

### DIFF
--- a/PWGDQ/Core/MixingHandler.cxx
+++ b/PWGDQ/Core/MixingHandler.cxx
@@ -59,7 +59,9 @@ void MixingHandler::AddMixingVariable(int var, int nBins, float* binLims)
   // add a mixing variable
   //
   fVariables.push_back(var);
-  fVariableLimits[fVariables.size()].Set(nBins, binLims);
+  TArrayF varBins;
+  varBins.Set(nBins, binLims);
+  fVariableLimits.push_back(varBins);
   VarManager::SetUseVariable(var);
 }
 
@@ -128,9 +130,10 @@ int MixingHandler::FindEventCategory(float* values)
     Init();
   }
 
-  int bin[fVariables.size()];
+  std::vector<int> bin;
   for (int i = 0; i < fVariables.size(); ++i) {
-    bin[i] = TMath::BinarySearch(fVariableLimits[i].GetSize(), fVariableLimits[i].GetArray(), values[fVariables[i]]);
+    int binValue = TMath::BinarySearch(fVariableLimits[i].GetSize(), fVariableLimits[i].GetArray(), values[fVariables[i]]);
+    bin.push_back(binValue);
     if (bin[i] == -1 || bin[i] == fVariableLimits[i].GetSize() - 1) {
       return -1; // all variables must be inside limits
     }
--- a/PWGDQ/Core/MixingHandler.cxx
+++ b/PWGDQ/Core/MixingHandler.cxx
@@ -26,37 +26,22 @@ ClassImp(MixingHandler);
 MixingHandler::MixingHandler() : TNamed(),
                                  fIsInitialized(kFALSE),
                                  fVariableLimits(),
-                                 fVariables(),
-                                 fVariableNames(),
-                                 fNMixingVariables(0)
+                                 fVariables()
 {
   //
   // default constructor
   //
-  float dummyRange[2] = {-99999., +99999.};
-  for (int iVar = 0; iVar < kNMaxVariables; ++iVar) {
-    fVariableLimits[iVar].Set(2, dummyRange);
-    fVariables[iVar] = VarManager::kNothing;
-    fVariableNames[iVar] = "";
-  }
 }
 
 //_________________________________________________________________________
 MixingHandler::MixingHandler(const char* name, const char* title) : TNamed(name, title),
                                                                     fIsInitialized(kFALSE),
                                                                     fVariableLimits(),
-                                                                    fVariables(),
-                                                                    fNMixingVariables(0)
+                                                                    fVariables()
 {
   //
   // Named constructor
   //
-  float dummyRange[2] = {-99999., +99999.};
-  for (int iVar = 0; iVar < kNMaxVariables; ++iVar) {
-    fVariableLimits[iVar].Set(2, dummyRange);
-    fVariables[iVar] = VarManager::kNothing;
-    fVariableNames[iVar] = "";
-  }
 }
 
 //_________________________________________________________________________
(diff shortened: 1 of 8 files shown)


PWGDQ/Core/MixingHandler.cxx
```diff
@@ -0,0 +1,194 @@
...
+        tempCategory *= (fVariableLimits[iVar2].GetSize() - 1);
+      }
+    }
+    category += tempCategory;
+  }
+  return category;
+}
+
+//_________________________________________________________________________
+int MixingHandler::GetBinFromCategory(int iVar, int category) const
```
iarsene, line 179: This is a public interface function, and the user might not know the internal representation of the variables. Here iVar would have to be the exact index of a given variable as used in the calculation of the event category. I think it would be more useful to have a function which would have the form: GetBinFromCategory(VarManager::Variables var, int category). Also, for the user it may be more useful to get information on the range of the variable corresponding to this category.
In the AliPhysics, this function was mainly used as an internal utility for the mixing handler, with not much use externally. But with the interface i suggested, it could be used in a loop over mixing categories and retrieve info on variable ranges, etc.

PWGDQ/Core/MixingHandler.h
```diff
@@ -0,0 +1,70 @@
...
+ private:
+  MixingHandler(const MixingHandler& handler);
+  MixingHandler& operator=(const MixingHandler& handler);
+
+  // User options
+  bool fIsInitialized; // check if the mixing handler is initialized
+
+  TArrayF fVariableLimits[kNMaxVariables];
+  int fVariables[kNMaxVariables];
+  TString fVariableNames[kNMaxVariables]; //! variable names
```
iarsene, line 64: I am not sure the variable names are really needed, but as long as the strings will not be used for comparisons in the CPU critical parts of the code, i am fine with them. Also because the names we have in the VarManager are designed more for histogram axis titles...

PWGDQ/Tasks/dileptonMuMu.cxx
```diff
@@ -286,13 +304,15 @@ struct DQEventMixing {
             continue;
           }
           VarManager::FillPair(muon1, muon2, fValues, VarManager::kJpsiToMuMu);
+          int centBin = fMixHandler->GetBinFromCategory(fMixHandler->GetMixingVariable("Cent"), event1.mixingHash());
```
iarsene, line 307: This line of code is just for your own debugging right? Otherwise i would avoid using this function in the mixing loop.

PWGDQ/Tasks/dileptonMuMu.cxx
```diff
@@ -286,13 +304,15 @@ struct DQEventMixing {
...
+          //            std::cout<<"  CentralityBin "<< centBin<<std::endl;
           if (muon1.sign() * muon2.sign() < 0) {
-            fHistMan->FillHistClass("PairsMuonMEPM", fValues);
+            fHistMan->FillHistClass(Form("PairsMuonMEPM_%s", fCentBinNames[centBin].Data()), fValues);
```
iarsene, line 310: I am not sure i understand the logic here. Why do you want to fill histogram directories for separate centrality ranges? You can have one single histogram directory with histograms which have centrality as one of their dimensions.

PWGDQ/Tasks/dileptonMuMu.cxx
```diff
@@ -286,13 +304,15 @@ struct DQEventMixing {
...
           } else {
             if (muon1.sign() > 0) {
-              fHistMan->FillHistClass("PairsMuonMEPP", fValues);
+              fHistMan->FillHistClass(Form("PairsMuonMEPP_%s", fCentBinNames[centBin].Data()), fValues);
```
iarsene, line 313: similar comment as above

PWGDQ/Tasks/dileptonMuMu.cxx
```diff
@@ -319,19 +339,30 @@ struct DQDileptonMuMu {
...
     fHistMan->SetDefaultVarNames(VarManager::fgVariableNames, VarManager::fgVariableUnits);
 
     TString histNames = "";
-    histNames += "PairsMuonSEPM;PairsMuonSEPP;PairsMuonSEMM;";
+    std::vector<float> fCentLimsHashing = fMixHandler->GetMixingVariableLimits("Cent");
+    for (int icent = 0; icent < fCentLimsHashing.size() - 1; ++icent) {
+      fCentBinNames.push_back(Form("%.1f%.1f", fCentLimsHashing[icent], fCentLimsHashing[icent + 1]));
+      histNames += Form("PairsMuonSEPM_%s;PairsMuonSEPP_%s;PairsMuonSEMM_%s;", fCentBinNames[icent].Data(), fCentBinNames[icent].Data(), fCentBinNames[icent].Data());
+      //        std::cout<<"  CentralityBin "<< icent<<" : "<< fCentLimsHashing[icent]<<"-"<<fCentLimsHashing[icent + 1]<<std::endl;
+    }
```
iarsene, line 358: As i commented above, you do not need more than one mixed event histogram directory.

PWGDQ/Core/MixingHandler.h
```diff
@@ -0,0 +1,70 @@
...
+
+#include "PWGDQ/Core/HistogramManager.h"
+
+class MixingHandler : public TNamed
+{
+
+ public:
+  enum Constants { // event mixing for correlations
+    kNMaxVariables = 10
+  };
```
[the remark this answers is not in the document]
vfeuilla: Yes, done

PWGDQ/Core/MixingHandler.h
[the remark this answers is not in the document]
vfeuilla: Maybe it's because I am not very good at coding but I believe that in the case of std::array, I need to specify the number of elements in the array (std::array<int, 3>, for instance). But the number of bins for the mixing variable is not fixed, so I am not sure how to address that. At the moment, I used std::vector<TArrayF> for the variable limits.

PWGDQ/Core/MixingHandler.h
```diff
@@ -0,0 +1,70 @@
...
+  MixingHandler(const MixingHandler& handler);
+  MixingHandler& operator=(const MixingHandler& handler);
+
+  // User options
+  bool fIsInitialized; // check if the mixing handler is initialized
+
+  TArrayF fVariableLimits[kNMaxVariables];
+  int fVariables[kNMaxVariables];
+  TString fVariableNames[kNMaxVariables]; //! variable names
+  int fNMixingVariables;
```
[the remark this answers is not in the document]
vfeuilla: Indeed. Done.

PWGDQ/Core/MixingHandler.h
```diff
@@ -0,0 +1,70 @@
...
+
+  ClassDef(MixingHandler, 4);
```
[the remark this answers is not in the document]
vfeuilla: Nice catch! I forgot to change it ;-)

PWGDQ/Core/MixingHandler.cxx
```diff
@@ -0,0 +1,194 @@
...
+
+//_________________________________________________________________________
+void MixingHandler::Init()
+{
+  //
+  // Initialization of pools
+  // NOTE: The master array holding tracks is a 1D array but it will be represented as an n-dim array
+  //       The size of the array will be N_1 x N_2 x ... x N_n
+  //       The correct event category will be retrieved using the function FindEventCategory()
+  //
```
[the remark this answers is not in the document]
vfeuilla: It has been removed

PWGDQ/Core/MixingHandler.cxx
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  fIsInitialized = kTRUE;
+}
+
+//_________________________________________________________________________
+int MixingHandler::FindEventCategory(float* values)
+{
+  //
+  // Find the event category corresponding to the centrality, vtxz and ep values
```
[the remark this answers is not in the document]
vfeuilla: Done.

PWGDQ/Core/MixingHandler.cxx
vfeuilla: This is why I added "names" for the variables. Because I wanted to access the internal representation of the variables, which I do not know prior. Using a VarManger::variable is a much more elegant way to do it!

PWGDQ/Core/MixingHandler.h
vfeuilla: See answer above. I removed the names

PWGDQ/Tasks/dileptonMuMu.cxx
vfeuilla: It wasn't for debugging, I wanted to know in which centrality class I was to fill the adequate histogram directory, but it has been removed, due to your next comment

PWGDQ/Tasks/dileptonMuMu.cxx
vfeuilla: I simply didn't think of using a 2D histogram. In Run2, I used to have a lot of 1D histograms instead of 2D or 3D histograms, that were more memory consuming, and I tend to still think like that. A histogram with centrality as a dimension is a more efficient way though, I agree.

PWGDQ/Tasks/dileptonMuMu.cxx
vfeuilla: See above

PWGDQ/Tasks/dileptonMuMu.cxx
vfeuilla: See above

iarsene (approved):

PR #76, opened 2021-09-13
Update QA tasks

njacazio:
- Update A3 mult task
  - Add MinContrib cut

- Update efficiency task
  - Add multiplicity in efficiency task
  - Add track length to efficiency task

### DIFF
--- a/ALICE3/Tasks/alice3-qa-multiplicity.cxx
+++ b/ALICE3/Tasks/alice3-qa-multiplicity.cxx
@@ -24,18 +24,19 @@ struct ALICE3MultTask {
   Configurable<float> MinEta{"MinEta", -0.8f, "Minimum eta in range"};
   Configurable<float> MaxEta{"MaxEta", 0.8f, "Maximum eta in range"};
   Configurable<float> MaxMult{"MaxMult", 1000.f, "Maximum multiplicity in range"};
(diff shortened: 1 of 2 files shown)


jgrosseo (approved):

iarsene (approved):

njacazio:
Do we really need more than two approvals?

ktf:
I think you need one for PWGPP and one for ALICE3.

vkucera (approved):

ktf:
I checked. PWGPP does not have an entry in CODEOWNERS, so that's why Core is involved.

njacazio:
Thanks Giulio! I will add PWGPP CODEOWNERS then

jgrosseo:
Why wasn't my approval for Core enough?

jgrosseo:
I think the one from ALICE3 was missing (mine was OK for core)

jgrosseo:
The point is that one cannot approve oneself. This is rather annoying (not only for us), and does not seem to have a solution:
https://github.community/t/do-not-require-owner-approval-if-the-pull-request-is-from-an-owner/369/66

We may need an external tool then like we have for AliPhysics...

njacazio:
Ok indeed you are right, this would not help

PR #80, opened 2021-09-14
ALICE3: Add table for ECAL

njacazio:


vkucera (approved):

vkucera:
@njacazio If you want the ALICE3 string to be displayed in the commit history, you need to add it in the commit message as well. The PR title is ignored in case of single-commit PRs.

njacazio:
Hi @vkucera thanks for the suggestion, good to know. It should be ok having it only in the PR title

PR #74, opened 2021-09-11
Add tracktime task

njacazio:

PR #93, opened 2021-09-18
tofSignal -> tracktime transition

jgrosseo:
Squashed PR https://github.com/AliceO2Group/O2Physics/pull/74 from @njacazio

### DIFF
--- a/ALICE3/TableProducer/alice3-pidTOF.cxx
+++ b/ALICE3/TableProducer/alice3-pidTOF.cxx
@@ -89,8 +89,7 @@ struct ALICE3pidTOFTask {
   template <o2::track::PID::ID id>
   float nsigma(Trks::iterator track)
   {
-    // return (track.tofSignal() - tof::ExpTimes< Trks::iterator, id>::GetExpectedSignal(track.collision(), track)) / sigma<id>(track);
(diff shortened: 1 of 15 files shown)


vkucera (approved):
It is quite unfortunate that the [corresponding AliPhysics PR](https://github.com/alisw/AliPhysics/pull/18517) was already merged 2 days ago which now breaks running on converted data.

jgrosseo:
Hi Vit.
We are reconverting the data over the weekend for which we obviously need the changes in AliPhysics. How would you do it?

vkucera:
> Hi Vit.
> We are reconverting the data over the weekend for which we obviously need the changes in AliPhysics. How would you do it?

OK, if the plan is to deploy reprocessed data on AliHyperloop when the corresponding O2Physics tag is released, I agree there is not much that could be done better.

jgrosseo:
> 
> 
> > Hi Vit.
> > We are reconverting the data over the weekend for which we obviously need the changes in AliPhysics. How would you do it?
> 
> OK, if the plan is to deploy reprocessed data on AliHyperloop when the corresponding O2Physics tag is released, I agree there is not much that could be done better.

Yes, this is the plan. As announced here: https://mattermost.web.cern.ch/alice/pl/i59pk56sx3bkzefz4dn971gxgy

PR #100, opened 2021-09-20
make track selection work with run3 simulations

mario-krueger:


Common/Core/TrackSelection.h
```diff
@@ -77,37 +78,54 @@ class TrackSelection
   template <typename T>
   bool IsSelected(T const& track, const TrackCuts& cut)
   {
+    bool isRun2 = track.trackType() == o2::aod::track::Run2Track || track.trackType() == o2::aod::track::Run2Tracklet;
```
jgrosseo, line 81: Why do you need to evaluate this on each track?

jgrosseo: You want to autodetect this setting?

mario-krueger: how can this can be implemented?

jgrosseo:
Until we have the meta data reading, I propose you keep the run check as you did it now.
This PR https://github.com/AliceO2Group/AliceO2/pull/7075 fixes the TrackType, so maybe we can get rid of the work around, right away. What do you think?

jgrosseo:
Merging despite further changes are coming.

PR #104, opened 2021-09-22
Add default case to fix compiler warning

njacazio:


njacazio:
@mario-krueger here it is!

PR #115, opened 2021-09-24
Temporarilly adding a copy of the filter-PP task to study the impact of the low resolution PID tables on the DQ pp triggers

iarsene:
additional small fixed in the CMakeLists to avoid double "dq" in workflow names


jgrosseo:
Did you consider adding a second process function and using the PROCESS_SWITCH instead of copying the task?

iarsene:
The idea was to have a workflow that we can run in hyperloop. With a switch, i would have to run both tasks simultaneously right ? Or i can switch between them in hyperloop?

iarsene:
The intention is to keep just one of these anyway. We will remove the other one as soon as we conclude on whether we can use or not the compressed PID information.

jgrosseo:
Yes, you can switch on hyperloop

jgrosseo:
image

jgrosseo:
I think one task is better (these things always stay ;-))

jgrosseo:
You can also set the default value for the switch in the code

iarsene:
This duplicated task would not stay, but if the switch works on hyperloop (now i see it works on a workflow in which i use the switch) i prefer this way. I will update the commit.

iarsene:
Hey @jgrosseo,
So i tried to make an implementation where i use two process switches, one that uses the compressed TPC and TOF pid and one that uses the full ones. The task compiles however it crashes at runtime. I don't get anything else except this:
[50960:tpc-pid-full]: [21:51:38][INFO] Received device shutdown request (signal 15).
At this point i don't know how to debug this, but now i would prefer to merge this PR as it is and be done with our tests, which are unrelated to this specific implementation. Although on the long term most likely i am not interested in having a switch between the full and compressed PID (we will use just one of them), i can ask Nicolo J to help me have a look into this later.

jgrosseo (approved):

PR #119, opened 2021-09-26
Selection cuts into the reco <-> gen mapping check task

victor-gonzalez:
Also incorporated the active detectors output both before and after
selection cuts

Strange, with the same selection cuts we get more tracks (one thousand (5%)
with the latest nightly MBTB production) if we go through the tracks table
based on a concrete collision than if we go through the whole tracks table
checking the selection of each track associated collision

### DIFF
--- a/PWGCF/Tasks/dptdptcorrelations-simchl.cxx
+++ b/PWGCF/Tasks/dptdptcorrelations-simchl.cxx
@@ -294,6 +294,7 @@ template <typename CollisionObject>
 bool IsEvtSelected(CollisionObject const& collision, float& centormult)
 {
   using namespace filteranalysistask;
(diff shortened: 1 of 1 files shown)


jgrosseo (approved):

jgrosseo:
> Strange, with the same selection cuts we get more tracks (one thousand (5%)
> with the latest nightly MBTB production) if we go through the tracks table
> based on a concrete collision than if we go through the whole tracks table
> checking the selection of each track associated collision

This is indeed strange. I assume this also happens locally? Could you find out which tracks are those?

victor-gonzalez:
Definitely!
I have to find out the tracks/collisions which are discarded/incorporated

PR #121, opened 2021-09-27
Extend caching and improve performance of timestamp task

njacazio:


jgrosseo (approved):

jgrosseo:
@ktf should long be avoided everywhere or just when we write to disk? Here a 64 bit value is needed...

PR #122, opened 2021-09-27
First step towards the new centrality task

victor-gonzalez:
So far a separated new task only tested by the
dpt-dpt correlations task


Common/TableProducer/centralityTable_new.cxx
```diff
@@ -0,0 +1,149 @@
...
+    mhVtxAmpCorrV0A = nullptr;
+    mhVtxAmpCorrV0C = nullptr;
+    mhMultSelCalibV0M = nullptr;
+  }
+
+  void process(soa::Join<aod::Collisions, aod::Mults>::iterator const& collision, aod::BCsWithTimestamps const&, aod::Tracks const& tracks)
+  {
+    /* check the previous run number */
+    auto bc = collision.bc_as<aod::BCsWithTimestamps>();
+    if (bc.runNumber() != mRunNumber) {
```
jgrosseo, line 86: This is for caching, right? In the future the CCDBConfigurable will do this automatically.

Common/TableProducer/centralityTable_new.cxx
```diff
@@ -0,0 +1,149 @@
...
+      LOGF(debug, "timestamp=%llu", bc.timestamp());
+      TList* callst = ccdb->getForTimeStamp<TList>("Users/v/victor/Centrality", bc.timestamp());
+
+      if (callst != nullptr) {
+        auto zvtxcalib = [callst](const char* tpname, const char* hname) {
+          TProfile* tp = (TProfile*)callst->FindObject(tpname);
```
jgrosseo, line 92: Why is this here and not stored already "ready to use" in the CCDB?

Common/TableProducer/centralityTable_new.cxx
```diff
@@ -0,0 +1,149 @@
...
+        auto multcalib = [callst](const char* ccdbhname, const char* hname) {
+          TH1* h = (TH1*)callst->FindObject(ccdbhname)->Clone(hname);
+          return h;
+        };
+        if (estV0M == 1) {
+          mhVtxAmpCorrV0A = zvtxcalib("hVtx_fAmplitude_V0A", "zvtxCalibV0A");
+          mhVtxAmpCorrV0C = zvtxcalib("hVtx_fAmplitude_V0C", "zvtxCalibV0C");
+          mhMultSelCalibV0M = multcalib("hMultSelCalib_V0M", "MultSelCalibV0M");
+          if ((mhVtxAmpCorrV0A != nullptr) and (mhVtxAmpCorrV0C != nullptr) and (mhMultSelCalibV0M != nullptr)) {
+            mV0MCallibrationStored = true;
```
jgrosseo, line 110: There is a typo in Calibration

Common/TableProducer/centralityTable_new.cxx
```diff
@@ -0,0 +1,149 @@
...
+          }
+        }
+        if (estSPD == 1) {
+          LOGF(fatal, "Calibraltion information estimated from SPD tracklets still not available");
```
jgrosseo, line 114: Also typo here :)

Common/TableProducer/centralityTable_new.cxx
```diff
@@ -0,0 +1,149 @@
...
+        if (estCL1 == 1) {
+          LOGF(fatal, "Calibraltion information estimated from CL1 still not available");
+        }
+        if (mV0MCallibrationStored) {
+          mRunNumber = bc.runNumber();
+        }
+      } else {
+        /* we dont change the run number to keep trying */
+        mV0MCallibrationStored = false;
+        LOGF(error, "Centrality calibration is not available in CCDB for run=%d at timestamp=%llu", bc.runNumber(), bc.timestamp());
```
jgrosseo, line 128: So if it is not found, we make one request per collision? This is not good I think because it makes a high load on the CCDB. I would not retry for this run then.

jgrosseo: Shouldn't the error be fatal?

Common/DataModel/Centrality_new.h
```diff
@@ -0,0 +1,35 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
```
jgrosseo, line 1: When we merge this, we will replace `Centrality.h` right away (and all uses of it of course)...

Common/TableProducer/centralityTable_new.cxx
victor-gonzalez: Good!
For the time being if the run number does not change the CCDB is only queried once

victor-gonzalez:
Sorry!
Wrong action!
Not ready for reviewing yet!
Still in draft!

Common/TableProducer/centralityTable_new.cxx
victor-gonzalez: Of course, if we so decide, we can store the zvtx calibration already ready to be used in the CCDB

Common/TableProducer/centralityTable_new.cxx
victor-gonzalez: Thanks!

Common/TableProducer/centralityTable_new.cxx
victor-gonzalez: :)

Common/TableProducer/centralityTable_new.cxx
victor-gonzalez: My logic was
a) if there is a glitch in the CCDB we don't discard the whole run, let's keep trying
b) if a run does not have calibration information perhaps others do, don't kill the workflow

But you are right a) will originate high load towards CCDB, and b) will make no sense if the CCDB is down for any reason.

I can set counters for both cases and act accordingly
- one counter for the number of tries on a concrete run number before we give up trying for that run and set all centralities out of range for that run
- one counter for the number of runs without CCDB access before we kill the workflow

or just act without any counters at all

Common/DataModel/Centrality_new.h
victor-gonzalez: Correct!
I did not do it because I had to change all usages of the `Cents` table
Perhaps we can do that as a previous step?

victor-gonzalez:
@ekryshen I don't know how to formally add you to the reviewers of this draft PR
I hope it will be enough with this 
Your inputs will be really appreciated
Thanks in advance!

Common/TableProducer/centralityTable_new.cxx
jgrosseo: I think this would be better. @ekryshen what do you think?

Common/TableProducer/centralityTable_new.cxx
jgrosseo: I don't think we need failsafe at many levels (CCDB will already try to get the file from all the replicas). We also resubmit the job on failure.
If we ask centrality for a run which is not available, it is better to bail out. So I would go for 
- no counter
- Fatal, right away

Common/TableProducer/centralityTable_new.cxx
victor-gonzalez: The calibration information for LHC18q and LHC18r are stored at CCDB fully processed there is nothing extra to do when loading them. The task has been modified accordingly

Common/TableProducer/centralityTable_new.cxx
victor-gonzalez: Fatal right away if the centrality calibration is not available or if it is corrupted (not the whole set of histograms are present)

Common/DataModel/Centrality_new.h
jgrosseo: We can do this in bulk once this is ready

ddobrigk:
Hi @victor-gonzalez, thank you very very much, this looks very good! Note that the files I sent you contain also other estimators, such as CL0 and CL1, in them already and that this could be done in a second step. 

For posterity, we may also want to have a broader discussion (with @jgrosseo and others) on standardizing estimator naming and especially how to proceed with the actual Run 3 data, where we may want to use FIT signals etc.

Common/DataModel/Centrality_new.h
victor-gonzalez: OK!

victor-gonzalez:
Thanks @ddobrigk 
Yes, the information in CCDB for CL0, CL1 and the other estimators, is already normalized and ready to be used by the centrality task. As the only information which is currently available/required is based on the V0M estimator, I believe it is better to go to production with only that estimator available (although based in V0A and V0C) and then progressively modify the centrality task to incorporate the other ones

jgrosseo:
Yes, before merging we should discuss the column names. Some of them should include Run2 I guess because we will not have an SPD in Run 3 for instance :)

ekryshen:
> Very nice! We should add your converted centrality files then to the production CCDB - I have asked access for you. What would be a good prefix? `Analysis/Centrality` ?

We already have /Centrality folder in the production CCDB:
http://alice-ccdb.cern.ch/browse
Similar to /EventSelection
Do you think it would be better to move everything to /Analysis? Then we should ask Costin to clean up.

victor-gonzalez:
Very good!
The upload takes few seconds only
Should we leave the current calibration for LHC15o?
I can remove it before uploading the new one

victor-gonzalez:
> > > Very nice! We should add your converted centrality files then to the production CCDB - I have asked access for you. What would be a good prefix? `Analysis/Centrality` ?
> > 
> > 
> > We already have /Centrality folder in the production CCDB: http://alice-ccdb.cern.ch/browse Similar to /EventSelection Do you think it would be better to move everything to /Analysis? Then we should ask Costin to clean up.
> 
> Sorry for this. Yes, we should keep the old path. However, then we cannot upload yet the objects, because the old centrality task will stop working...

The old path is `Centrality/CumMultV0M` which is not any longer representative because the new calibration information incorporates estimation from other detectors.
I guess we should use directly the `Centrality` prefix
But having in mind that the next step we are addressing is the pp multiplicity percentiles, does it make sense to have a common `CentralityMultiplicity` prefix?

jgrosseo:
Thanks. So we will not have an issue with name clashes. 
I see your point but `CentralityMultiplicity` is not nice. So due to lack of other suggestions I'd keep `Centrality`
You can then already upload your objects into `Centrality/Estimators` ?

victor-gonzalez:
Also the conflicts were resolved

victor-gonzalez:
`LHC15o`, `LHC18q` and `LHC18r` centrality calibration and estimation information is already uploaded to the `CCDB` production endpoint under `Centrality/Estimators` 
Nice output from the `centrality-qa` from three data files from two different runs from `LHC15o`
[https://cernbox.cern.ch/index.php/s/PBnkasvlHFxkJgO](https://cernbox.cern.ch/index.php/s/PBnkasvlHFxkJgO)

PR #124, opened 2021-09-28
Added a producer task for PID ML data. To be used on Hyperloop.

saganatt:
Hi @jgrosseo ,

In general, I didn't consider merging unstable machine-learning PID codes for now, but apparently only master versions can be ran on Hyperfloop? In that case, could you accept this small task? It would make easier to pre-process much larger pieces of data for ML input.
Let me know if there is any issue with the task code or whether it should be put in another place.


jgrosseo (changes requested):
There is no problem, but the task has to go elsewhere.

Table producing tasks are in the TableProducer folder and we have to put this into a PWG folder. Which one?

jgrosseo:
Maybe PWGPP for now?

saganatt:
OK, I put it in PWGPP for now.

jgrosseo (approved):

PR #126, opened 2021-09-28
ThreeBody Trigger

lauraser:
Dear experts,
this is the first version of 3-body trigger. As you will see, it is a bit slow, because there are sliceBy() and Partitions used for every collision. The current idea is to run femto skimming task and pipe the output to trigger. The CFFilter task currently takes as input aod::FullCollision const& col as well as the full FemtoDreamCollisions and FemtoDreamParticles tables. Then inside the code, I check, if the current collision is present in FemtoDreamCollision table -> if it is not, I set the trigger to false. Otherwise, the trigger is applied by forming triplets out of the particle candidates but for this, I have to use only the particles which are coming from specific collision thus inside the process task I apply the partitions. Please, let me know if there is a more efficient way to do this.
P.S. I as well had an idea to take as input only the FemtoDreamCollisions and FemtoDreamParticles. The FemtoDreamCollisions table has a column that stores the globalIndex from FullCollision table. If the skimming task is fed the collisions in order and there will be no parallelization on such level, one could then check the globalIndex and if for example the first collision in FemtoDreamCollisions has collisionID 5, one knows that the first 4 collisions from FullCollision table were rejected and one can update then the trigger table 4 times using trigger value false and then apply the trigger on current collision with collisionID 5. However, I am new to O2Physics and I don't know if this would be possible. Any advices will be greatly appreciated.
Cheers
Laura

### DIFF
--- a/EventFiltering/CMakeLists.txt
+++ b/EventFiltering/CMakeLists.txt
@@ -37,3 +37,8 @@ o2physics_add_dpl_workflow(hf-filter
                            SOURCES PWGHF/HFFilter.cxx
                            PUBLIC_LINK_LIBRARIES O2::Framework O2::DetectorsBase O2Physics::AnalysisCore O2::DetectorsVertexing
                            COMPONENT_NAME Analysis)
+
+o2physics_add_dpl_workflow(cf-filter
+                            SOURCES PWGCF/CFFilter.cxx
+                            PUBLIC_LINK_LIBRARIES O2::Framework O2::DetectorsBase O2Physics::AnalysisCore 
+                            COMPONENT_NAME Analysis)
--- a/EventFiltering/CMakeLists.txt
+++ b/EventFiltering/CMakeLists.txt
@@ -38,7 +38,7 @@ o2physics_add_dpl_workflow(hf-filter
                            PUBLIC_LINK_LIBRARIES O2::Framework O2::DetectorsBase O2Physics::AnalysisCore O2::DetectorsVertexing
                            COMPONENT_NAME Analysis)
 
-o2physics_add_dpl_workflow(cf-filter
+o2physics_add_dpl_workflow(cf-threebodyfemto-filter
                             SOURCES PWGCF/CFFilter.cxx
-                            PUBLIC_LINK_LIBRARIES O2::Framework O2::DetectorsBase O2Physics::AnalysisCore 
+                            PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::AnalysisCore 
                             COMPONENT_NAME Analysis)
--- a/EventFiltering/PWGCF/CFFilter.cxx
+++ b/EventFiltering/PWGCF/CFFilter.cxx
@@ -29,7 +29,6 @@
 #include "../../PWGCF/FemtoDream/FemtoDreamMath.h"
 #include "../../PWGCF/FemtoDream/FemtoDreamPairCleaner.h"
 
-
 #include "Common/DataModel/EventSelection.h"
 #include "Common/DataModel/Multiplicity.h"
 
@@ -39,9 +38,9 @@
 namespace
 {
 
-static constexpr int nTriplets{4}; 
+static constexpr int nTriplets{4};
 
-enum CFTriggers { 
+enum CFTriggers {
   kPPP = 0,
   kPPL,
   kPLL,
@@ -55,8 +54,8 @@ static const std::vector<std::string> CfTriggerNames{"ppp", "ppL", "pLL", "LLL"}
 namespace o2::aod
 {
 using FullCollision = soa::Join<aod::Collisions,
-                                                      aod::EvSels,
-                                                      aod::Mults>::iterator;
+                                aod::EvSels,
+                                aod::Mults>::iterator;
 } // namespace o2::aod
 
 using namespace o2;
@@ -78,7 +77,7 @@ struct CFFilter {
     registry.add("fProcessedEvents", "CF - event filtered;;events", HistType::kTH1F, {{6, -0.5, 4.5}});
     std::array<std::string, 6> eventTitles = {"all", "rejected", "p-p-p", "p-p-L", "p-L-L", "L-L-L"};
     for (size_t iBin = 0; iBin < eventTitles.size(); iBin++) {
-     registry.get<TH1>(HIST("fProcessedEvents"))->GetXaxis()->SetBinLabel(iBin+1, eventTitles[iBin].data());
+      registry.get<TH1>(HIST("fProcessedEvents"))->GetXaxis()->SetBinLabel(iBin + 1, eventTitles[iBin].data());
     }
     registry.add("fSameEvent", "CF - same event distribution;;events", HistType::kTH1F, {{8000, 0, 8}});
   }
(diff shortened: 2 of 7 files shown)


jgrosseo:
Thanks! We will have a detailed look. I convert it to draft for now for the discussion.

EventFiltering/CMakeLists.txt
```diff
@@ -37,3 +37,8 @@ o2physics_add_dpl_workflow(hf-filter
                            SOURCES PWGHF/HFFilter.cxx
                            PUBLIC_LINK_LIBRARIES O2::Framework O2::DetectorsBase O2Physics::AnalysisCore O2::DetectorsVertexing
                            COMPONENT_NAME Analysis)
+
+o2physics_add_dpl_workflow(cf-filter
```
jgrosseo, line 41: I propose a more meaningful name to the actual trigger done.

EventFiltering/CMakeLists.txt
```diff
@@ -37,3 +37,8 @@ o2physics_add_dpl_workflow(hf-filter
...
+                            SOURCES PWGCF/CFFilter.cxx
+                            PUBLIC_LINK_LIBRARIES O2::Framework O2::DetectorsBase O2Physics::AnalysisCore
```
jgrosseo, line 43: Do you really depend on O2::DetectorsBase?

EventFiltering/PWGCF/CFFilter.cxx
```diff
@@ -0,0 +1,178 @@
...
+#include "Framework/HistogramRegistry.h"
+
+#include "../filterTables.h"
+
+#include "../../PWGCF/DataModel/FemtoDerived.h"
+#include "../../PWGCF/FemtoDream/FemtoDreamParticleHisto.h"
+#include "../../PWGCF/FemtoDream/FemtoDreamPairCleaner.h"
+#include "../../PWGCF/FemtoDream/FemtoDreamContainer.h"
+#include "../../PWGCF/FemtoDream/FemtoDreamMath.h"
+#include "../../PWGCF/FemtoDream/FemtoDreamPairCleaner.h"
```
jgrosseo, line 30: I believe you can remove the `../../` everywhere

EventFiltering/PWGCF/CFFilter.cxx
```diff
@@ -0,0 +1,178 @@
...
+  void process(aod::FullCollision const& col, o2::aod::FemtoDreamCollisions const& colsFemto,
+               o2::aod::FemtoDreamParticles& partsFemto)
+  {
+    registry.get<TH1>(HIST("fProcessedEvents"))->Fill(0);
+    bool keepEvent[nTriplets]{false};
+
+    // check if current collision was selected by femto skimmer
+    auto collsInFemto = colsFemto.sliceBy(o2::aod::femtodreamcollision::collisionId, col.globalIndex());
+    if (collsInFemto.size() == 0) {
+      keepEvent[kPPL] = false;
```
jgrosseo, line 97: Here you would need all of them to 0, not only kPPL, right?

EventFiltering/PWGCF/CFFilter.cxx
```diff
@@ -0,0 +1,178 @@
...
+  float mMassLambda = TDatabasePDG::Instance()->GetParticle(3122)->Mass();
+
+  void process(aod::FullCollision const& col, o2::aod::FemtoDreamCollisions const& colsFemto,
+               o2::aod::FemtoDreamParticles& partsFemto)
+  {
+    registry.get<TH1>(HIST("fProcessedEvents"))->Fill(0);
+    bool keepEvent[nTriplets]{false};
+
+    // check if current collision was selected by femto skimmer
+    auto collsInFemto = colsFemto.sliceBy(o2::aod::femtodreamcollision::collisionId, col.globalIndex());
```
jgrosseo, line 95: You are right that this is inefficient. We use a slice to find a single value.
First of all there is no parallel processing on this level. A process function will always process a full data frame.

I would add a setting `fillAll` to the femto framework that changes how `FemtoDreamCollisions` is produced. In this case `FemtoDreamCollisions` is filled for *all* collisions and it can be therefore joined with `FullCollision`. If the collision is not interesting for femto then `FemtoDreamParticles` is not filled for that collision.

Then you subscribe here to `process(join<FemtoDreamCollisions, FullCollisions>::iterator, FemtoDreamParticles`.

You profit from the full grouping before process is called. If FemtoDreamParticles is empty, you make the event as not triggered, otherwise you do the processing as you do it now.

Let me know what you think

PWGCF/DataModel/FemtoDerived.h
```diff
@@ -24,16 +25,21 @@
...
+                                                       aod::Mults>>;
+using FilteredFullCollision = FilteredFullCollisions::iterator;
 /// FemtoDreamCollision
 namespace femtodreamcollision
 {
-DECLARE_SOA_COLUMN(MultV0M, multV0M, float);       //! V0M multiplicity
-DECLARE_SOA_COLUMN(Sphericity, sphericity, float); //! Sphericity of the event
+DECLARE_SOA_COLUMN(MultV0M, multV0M, float);                                            //! V0M multiplicity
+DECLARE_SOA_COLUMN(Sphericity, sphericity, float);                                      //! Sphericity of the event
+DECLARE_SOA_INDEX_COLUMN_FULL(Collision, collision, int, FilteredFullCollisions, "_0"); //! GlobalIndex
```
jgrosseo, line 37: With my suggestion above, this column is not needed anymore

jgrosseo (changes requested):
Thanks for your work. See the inline comments for improvement.

PWGCF/FemtoDream/femtoDreamProducerTask.cxx
```diff
@@ -32,23 +32,21 @@
...
 using namespace o2;
 using namespace o2::analysis::femtoDream;
 using namespace o2::framework;
 using namespace o2::framework::expressions;
 
 namespace o2::aod
 {
-using FilteredFullCollision = soa::Filtered<soa::Join<aod::Collisions,
-                                                      aod::EvSels,
-                                                      aod::Mults>>::iterator;
```
jgrosseo, line 44: Why this change?

PWGCF/FemtoDream/femtoDreamProducerTask.cxx
```diff
@@ -77,14 +75,18 @@ struct femtoDreamProducerTask {
...
+
   /// Event cuts
   FemtoDreamCollisionSelection colCuts;
   Configurable<float> ConfEvtZvtx{"ConfEvtZvtx", 10.f, "Evt sel: Max. z-Vertex (cm)"};
   Configurable<bool> ConfEvtTriggerCheck{"ConfEvtTriggerCheck", false, "Evt sel: check for trigger"};
   Configurable<int> ConfEvtTriggerSel{"ConfEvtTriggerSel", kINT7, "Evt sel: trigger"};
   Configurable<bool> ConfEvtOfflineCheck{"ConfEvtOfflineCheck", false, "Evt sel: check for offline selection"};
 
-  Filter colFilter = nabs(aod::collision::posZ) < ConfEvtZvtx;
+  //Filter colFilter = nabs(aod::collision::posZ) < ConfEvtZvtxSerksnyte;
```
jgrosseo, line 89: What about keeping the filters but adding `!ConfIsTrigger &&` in front?

PWGCF/FemtoDream/femtoDreamProducerTask.cxx
```diff
@@ -120,9 +122,9 @@ struct femtoDreamProducerTask {
   Configurable<std::vector<float>> ConfV0DaughPIDnSigmaMax{"ConfV0DaughPIDnSigmaMax", std::vector<float>{5.f, 4.f}, "V0 Daugh sel: Max. PID nSigma TPC"};
 
   /// \todo should we add filter on min value pT/eta of V0 and daughters?
-  Filter v0Filter = (nabs(aod::v0data::x) < V0DecVtxMax.value) &&
+  /*Filter v0Filter = (nabs(aod::v0data::x) < V0DecVtxMax.value) &&
```
jgrosseo, line 125: see above

PWGCF/FemtoDream/femtoDreamProducerTask.cxx
jgrosseo: most likely it has to be `ConfIsTrigger == false`...

PWGCF/FemtoDream/femtoDreamProducerTask.cxx
```diff
@@ -166,20 +168,24 @@ struct femtoDreamProducerTask {
...
+      }
       return;
     }
     const auto vtxZ = col.posZ();
     const auto mult = col.multV0M();
     const auto spher = colCuts.computeSphericity(col, tracks);
     colCuts.fillQA(col);
     // now the table is filled
-    outputCollision(vtxZ, mult, spher);
+    outputCollision( vtxZ, mult, spher); // globalIndexAO2D is required for the 3-body trigger
```
jgrosseo, line 188: can the comment be removed?

PWGCF/FemtoDream/femtoDreamProducerTask.cxx
lauraser: The filtering was done on the Zvtx of the collision and then I could not store all events in the femto collision table as the filtered events are not passed to process(). So I deleted the filtering from the task. The filtering on the Zvtx was actually done additionally in the isSelected() task as well. Now it is only in the isSelected task.

PWGCF/FemtoDream/femtoDreamProducerTask.cxx
lauraser: I commented it because Valentina is not currently using it, it is not related to the trigger.

PWGCF/FemtoDream/femtoDreamProducerTask.cxx
lauraser: Yes, this is much better. I will implement it.

jgrosseo: Great. In this case the filters can be kept for derived data production with FemtoDream. I have just checked the latest commits and I think it is not yet done, right? If I overlooked it, let me know and I check again.

PWGCF/FemtoDream/femtoDreamProducerTask.cxx
jgrosseo: Generally, filters are more efficient than to do it in the IsSelected function. So if the trick with checking `ConfIsTrigger` works I would also keep the filter here

PWGCF/FemtoDream/femtoDreamProducerTask.cxx
```diff
@@ -166,20 +167,24 @@ struct femtoDreamProducerTask {
...
-  void process(aod::FilteredFullCollision const& col,
+  void process(soa::Join<aod::Collisions, aod::EvSels, aod::Mults>::iterator const& col,
                aod::FilteredFullTracks const& tracks,
                o2::aod::V0Datas const& fullV0s) /// \todo with FilteredFullV0s
   {
     /// First thing to do is to check whether the basic event selection criteria are fulfilled
     if (!colCuts.isSelected(col)) {
+      // if it is a trigger run, store unselected collision as well
+      if (ConfIsTrigger) {
+        outputCollision(col.posZ(), col.multV0M(), colCuts.computeSphericity(col, tracks)); // globalIndexAO2D is required for the 3-body trigger
```
jgrosseo, line 178: Also here the comment can go because you don't have that column anymore

lauraser: Here the trigger is for the rows, not columns. To explain, that if it is a trigger run, even the events which are not selected are saved in the table. I will update the comment to make it more clear and explain in both situations

PWGCF/FemtoDream/femtoDreamProducerTask.cxx
lauraser: Please, let me know if the implementation I made at the end is fine for the filter.

EventFiltering/PWGCF/CFFilter.cxx
```diff
@@ -0,0 +1,178 @@
...
+  Produces<aod::CFFilters> tags;
+
+  HistogramRegistry registry{"registry", {}, OutputObjHandlingPolicy::AnalysisObject, true, true};
+
+  //FemtoDreamPairCleaner<aod::femtodreamparticle::ParticleType::kTrack, aod::femtodreamparticle::ParticleType::kTrack> pairCleanerTT; Currently not used, will be needed later
+  FemtoDreamPairCleaner<aod::femtodreamparticle::ParticleType::kTrack, aod::femtodreamparticle::ParticleType::kV0> pairCleanerTV;
+
+  void init(o2::framework::InitContext&)
+  {
+    registry.add("fProcessedEvents", "CF - event filtered;;events", HistType::kTH1F, {{6, -0.5, 4.5}});
```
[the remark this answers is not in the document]
lauraser: Yes, thank you. I updated the code.

PWGCF/FemtoDream/femtoDreamProducerTask.cxx
```diff
@@ -85,7 +89,7 @@ struct femtoDreamProducerTask {
   Configurable<int> ConfEvtTriggerSel{"ConfEvtTriggerSel", kINT7, "Evt sel: trigger"};
   Configurable<bool> ConfEvtOfflineCheck{"ConfEvtOfflineCheck", false, "Evt sel: check for offline selection"};
 
-  //Filter colFilter = nabs(aod::collision::posZ) < ConfEvtZvtxSerksnyte;
+  Filter colFilter = (ConfIsTrigger == (uint8_t) false && nabs(aod::collision::posZ) < ConfEvtZvtx) || ConfIsTrigger == (uint8_t) true;
```
jgrosseo, line 92: `(ConfIsTrigger == (uint8_t) true) || (nabs(aod::collision::posZ) < ConfEvtZvtx)` would be enough

PWGCF/FemtoDream/femtoDreamProducerTask.cxx
jgrosseo: I don't mind a comment in general, but bugs me is `globalIndexAO2D`. This what the old column name which you have removed. So the comment should be "store all collisions for 3-body trigger" or so... Maybe I am missing something

lauraser: I misunderstood which for which line the comment was meant. I will fix this. Sorry

jgrosseo (approved):

jgrosseo:
Needs also @mpuccio to approve...

mpuccio (approved):

PR #128, opened 2021-09-29
Add DCA cut

ddobrigk:


ALICE3/Tasks/alice3-qa-multiplicity.cxx
```diff
@@ -13,6 +13,9 @@
 // O2 includes
 #include "Framework/AnalysisTask.h"
 #include "Framework/runDataProcessing.h"
+#include "Common/Core/TrackSelection.h"
+#include "Common/DataModel/TrackSelectionTables.h"
+#include "Common/Core/trackUtilities.h"
```
vkucera, line 18: Are these needed? I don't see them used.

ddobrigk: indeed, trackUtilities are unnecessary (but the others are necessary for the trackextension)! Removed

vkucera: OK, thanks

vkucera (approved):

PR #129, opened 2021-09-29
PWGHF: Separate prompt/nonprompt D+ and add different efficiency steps

fgrosa:
@fcatalan92

### DIFF
--- a/PWGHF/DataModel/HFCandidateSelectionTables.h
+++ b/PWGHF/DataModel/HFCandidateSelectionTables.h
@@ -16,10 +16,10 @@ namespace o2::aod
 {
 // selection steps
 enum SelectionStep {
-    RecoSkims = 0,
-    RecoTopol,
-    RecoPID,
-    NSelectionSteps
+  RecoSkims = 0,
+  RecoTopol,
+  RecoPID,
+  NSelectionSteps
 };
 
 namespace hf_selcandidate_d0
--- a/PWGHF/DataModel/HFCandidateSelectionTables.h
+++ b/PWGHF/DataModel/HFCandidateSelectionTables.h
@@ -14,6 +14,14 @@
 
 namespace o2::aod
 {
+// selection steps
+enum SelectionStep {
+    RecoSkims = 0,
+    RecoTopol,
+    RecoPID,
+    NSelectionSteps
+};
+
 namespace hf_selcandidate_d0
 {
 DECLARE_SOA_COLUMN(IsSelD0, isSelD0, int);           //!
--- a/PWGHF/DataModel/HFCandidateSelectionTables.h
+++ b/PWGHF/DataModel/HFCandidateSelectionTables.h
@@ -64,15 +64,9 @@ DECLARE_SOA_TABLE(HFSelD0CandidateALICE3Forward, "AOD", "HFSELD0CANDA3F", //!
 namespace hf_selcandidate_dplus
 {
 DECLARE_SOA_COLUMN(IsSelDplusToPiKPi, isSelDplusToPiKPi, int); //!
-DECLARE_SOA_COLUMN(IsRecoHFFlag, isRecoHFFlag, int);           //!
-DECLARE_SOA_COLUMN(IsRecoTopol, isRecoTopol, int);             //!
-DECLARE_SOA_COLUMN(IsRecoPID, isRecoPID, int);
 } // namespace hf_selcandidate_dplus
 DECLARE_SOA_TABLE(HFSelDplusToPiKPiCandidate, "AOD", "HFSELDPLUSCAND", //!
-                  hf_selcandidate_dplus::IsSelDplusToPiKPi,
-                  hf_selcandidate_dplus::IsRecoHFFlag,
-                  hf_selcandidate_dplus::IsRecoTopol,
-                  hf_selcandidate_dplus::IsRecoPID);
+                  hf_selcandidate_dplus::IsSelDplusToPiKPi);
 
 namespace hf_selcandidate_lc
 {
--- a/PWGHF/DataModel/HFCandidateSelectionTables.h
+++ b/PWGHF/DataModel/HFCandidateSelectionTables.h
@@ -64,9 +64,15 @@ DECLARE_SOA_TABLE(HFSelD0CandidateALICE3Forward, "AOD", "HFSELD0CANDA3F", //!
 namespace hf_selcandidate_dplus
 {
 DECLARE_SOA_COLUMN(IsSelDplusToPiKPi, isSelDplusToPiKPi, int); //!
+DECLARE_SOA_COLUMN(IsRecoHFFlag, isRecoHFFlag, int); //!
+DECLARE_SOA_COLUMN(IsRecoTopol, isRecoTopol, int);   //!
+DECLARE_SOA_COLUMN(IsRecoPID, isRecoPID, int);
 } // namespace hf_selcandidate_dplus
 DECLARE_SOA_TABLE(HFSelDplusToPiKPiCandidate, "AOD", "HFSELDPLUSCAND", //!
-                  hf_selcandidate_dplus::IsSelDplusToPiKPi);
+                  hf_selcandidate_dplus::IsSelDplusToPiKPi,
+                  hf_selcandidate_dplus::IsRecoHFFlag,
+                  hf_selcandidate_dplus::IsRecoTopol,
+                  hf_selcandidate_dplus::IsRecoPID);
 
 namespace hf_selcandidate_lc
 {
--- a/PWGHF/DataModel/HFCandidateSelectionTables.h
+++ b/PWGHF/DataModel/HFCandidateSelectionTables.h
@@ -64,8 +64,8 @@ DECLARE_SOA_TABLE(HFSelD0CandidateALICE3Forward, "AOD", "HFSELD0CANDA3F", //!
 namespace hf_selcandidate_dplus
 {
 DECLARE_SOA_COLUMN(IsSelDplusToPiKPi, isSelDplusToPiKPi, int); //!
-DECLARE_SOA_COLUMN(IsRecoHFFlag, isRecoHFFlag, int); //!
-DECLARE_SOA_COLUMN(IsRecoTopol, isRecoTopol, int);   //!
+DECLARE_SOA_COLUMN(IsRecoHFFlag, isRecoHFFlag, int);           //!
+DECLARE_SOA_COLUMN(IsRecoTopol, isRecoTopol, int);             //!
 DECLARE_SOA_COLUMN(IsRecoPID, isRecoPID, int);
 } // namespace hf_selcandidate_dplus
 DECLARE_SOA_TABLE(HFSelDplusToPiKPiCandidate, "AOD", "HFSELDPLUSCAND", //!
--- a/PWGHF/TableProducer/HFDplusToPiKPiCandidateSelector.cxx
+++ b/PWGHF/TableProducer/HFDplusToPiKPiCandidateSelector.cxx
@@ -120,17 +120,13 @@ struct HFDplusToPiKPiCandidateSelector {
     for (auto& candidate : candidates) {
 
       // final selection flag:
-      // 0 - rejected
-      // BIT(0) - bit for D+ candidate
-      // BIT(1) - selected by topological cuts
-      // BIT(2) - selected by PID cuts
       auto statusDplusToPiKPi = 0;
 
       if (!(candidate.hfflag() & 1 << DecayType::DPlusToPiKPi)) {
         hfSelDplusToPiKPiCandidate(statusDplusToPiKPi);
         continue;
       }
-      SETBIT(statusDplusToPiKPi, 0);
+      SETBIT(statusDplusToPiKPi, aod::SelectionStep::RecoSkims);
 
       auto trackPos1 = candidate.index0_as<aod::BigTracksPID>(); // positive daughter (negative for the antiparticles)
       auto trackNeg = candidate.index1_as<aod::BigTracksPID>();  // negative daughter (positive for the antiparticles)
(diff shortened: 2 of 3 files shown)


PWGHF/Tasks/taskDPlus.cxx
```diff
@@ -113,19 +114,38 @@ struct TaskDPlusMC {
...
      {"hCPARecSig", "3-prong candidates (matched);cosine of pointing angle;entries", {HistType::kTH1F, {{110, -1.1, 1.1}}}},
      {"hCPARecBg", "3-prong candidates (unmatched);cosine of pointing angle;entries", {HistType::kTH1F, {{110, -1.1, 1.1}}}},
      {"hEtaRecSig", "3-prong candidates (matched);#it{#eta};entries", {HistType::kTH1F, {{100, -2., 2.}}}},
      {"hEtaRecBg", "3-prong candidates (unmatched);#it{#eta};entries", {HistType::kTH1F, {{100, -2., 2.}}}},
-     {"hEtaGen", "MC particles (matched);#it{#eta};entries", {HistType::kTH1F, {{100, -2., 2.}}}}}};
+     {"hEtaGen", "MC particles (matched);#it{#eta};entries", {HistType::kTH1F, {{100, -2., 2.}}}},
+     {"hPtvsYRecSig_RecoPID", "2-prong candidates (RecoPID - matched);#it{p}_{T}^{rec.}; #it{y}", {HistType::kTH2F, {{100, 0., 10.}, {100, -5., 5.}}}},
+     {"hPtvsYRecSigPrompt_RecoPID", "2-prong candidates (RecoPID - matched, prompt);#it{p}_{T}^{rec.}; #it{y}", {HistType::kTH2F, {{100, 0., 10.}, {100, -5., 5.}}}},
+     {"hPtvsYRecSigNonPrompt_RecoPID", "2-prong candidates (RecoPID - matched, non-prompt);#it{p}_{T}^{rec.}; #it{y}", {HistType::kTH2F, {{100, 0., 10.}, {100, -5., 5.}}}},
```
fcatalan92, line 131: This could be `hPtvsYRecSig_Reco` etc. storing the pT vs y information for the selected candidates. The content will be the same, it's just a question of naming (see comment above)

fcatalan92 (commented):
Hi, I have a few comments.
In addition, the histograms you added span only the 0-10 GeV/c interval in pT. You should move to something like `Configurable<std::vector<double>> bins{"pTBins", std::vector<double>{hf_cuts_dplus_topikpi::pTBins_v}, "pT bin limits"};`, or a finer pT binning for the pT weights application

PWGHF/DataModel/HFCandidateSelectionTables.h
```diff
@@ -34,9 +34,15 @@ DECLARE_SOA_TABLE(HFSelD0Candidate, "AOD", "HFSELD0CAND", //!
 namespace hf_selcandidate_dplus
 {
 DECLARE_SOA_COLUMN(IsSelDplusToPiKPi, isSelDplusToPiKPi, int); //!
+DECLARE_SOA_COLUMN(IsRecoHFFlag, isRecoHFFlag, int);           //!
```
fcatalan92, line 37: Maybe this flag is redundant, i.e. we want to process only the candidates that are flagged as Dplus from the candidate creator. Moreover, the if  at https://github.com/AliceO2Group/O2Physics/blob/2a17f74e16478e84ba55fb0e5bf402bfc9757541/PWGHF/Tasks/taskDPlus.cxx#L77-78 should be removed now that we can use binary operations in filters (it should be possible with the new arrow).

fgrosa:
Ciao @fcatalan92, thanks for your comments! Indeed I did not think about it before, but a bitmap is sufficient and there is no need to add other columns, so we can spare some space in memory.

fcatalan92 (approved):
Fine for me, but you have to change also the default value of `d_selectionFlagDPlus`, with your changes the selections are turned off by default now.
Also, the filter at line 147 should be revised (data and MC should be decoupled maybe).

fgrosa:
> Fine for me, but you have to change also the default value of `d_selectionFlagDPlus`, with your changes the selections are turned off by default now. Also, the filter at line 147 should be revised (data and MC should be decoupled maybe).

Indeed I overlooked that. I removed the configurable for the MC since it now produces the outputs for all the steps in any case, while I changed the default for the main process function in order to apply topological+PID selections as default. Thanks!

vkucera:
> > Fine for me, but you have to change also the default value of `d_selectionFlagDPlus`, with your changes the selections are turned off by default now. Also, the filter at line 147 should be revised (data and MC should be decoupled maybe).
> 
> Indeed I overlooked that. I removed the configurable for the MC since it now produces the outputs for all the steps in any case, while I changed the default for the main process function in order to apply topological+PID selections as default. Thanks!

I don't understand this. Why should the selection in real and MC data differ?

vkucera:
Why is there no histogram for `didate.isSelDplusToPiKPi(), 1`?

vkucera:
Some of the modified histograms are used for validation where the pT binning in the AliPhysics task is not synced with the JSON, I think.

fgrosa:
> > > Fine for me, but you have to change also the default value of `d_selectionFlagDPlus`, with your changes the selections are turned off by default now. Also, the filter at line 147 should be revised (data and MC should be decoupled maybe).
> > 
> > 
> > Indeed I overlooked that. I removed the configurable for the MC since it now produces the outputs for all the steps in any case, while I changed the default for the main process function in order to apply topological+PID selections as default. Thanks!
> 
> I don't understand this. Why should the selection in real and MC data differ?

They should be different because in the MC we are computing the efficiency for each step, while for the data we just need the final selection (this is done similarly in the D0 task: https://github.com/AliceO2Group/O2Physics/blob/d00ad5cbf27219e08675df678cb1499599d4b186/PWGHF/Tasks/taskD0.cxx#L70, https://github.com/AliceO2Group/O2Physics/blob/d00ad5cbf27219e08675df678cb1499599d4b186/PWGHF/Tasks/taskD0.cxx#L185), just in a less efficient way because different columns are used instead of a single one

fgrosa:
> Why is there no histogram for `TESTBIT(candidate.isSelDplusToPiKPi(), 0)`?

Because we filter candidates that have `BIT(0)` essentially and therefore there is a histogram that corresponds to `TESTBIT(candidate.isSelDplusToPiKPi(), 0)` (i.e. `hPtRecSig` and the corresponding ones for prompt and nonprompt)

fgrosa:
> Some of the modified histograms are used for validation where the pT binning in the AliPhysics task is not synced with the JSON, I think.

I will open a PR in the next days

fgrosa:
> > Some of the modified histograms are used for validation where the pT binning in the AliPhysics task is not synced with the JSON, I think.
> 
> I will open a PR in the next days

I modified the JSON in Run3Analysisvalidation AliceO2Group/Run3Analysisvalidation#305, while if I did not overlook it, it seems to me that there is no need to modify anything in the AliPhysics task

vkucera:
> > Why is there no histogram for `TESTBIT(candidate.isSelDplusToPiKPi(), 0)`?
> 
> Because we filter candidates that have `BIT(0)` essentially and therefore there is a histogram that corresponds to `TESTBIT(candidate.isSelDplusToPiKPi(), 0)` (i.e. `hPtRecSig` and the corresponding ones for prompt and nonprompt)

The worry I have is that the `hPtRecSig` et al histograms are meant for the "complete-efficiency" calculation which should include all selection steps but your changes seem to change their purpose to include just the skimming selection step.

vkucera:
> > > Why is there no histogram for `TESTBIT(candidate.isSelDplusToPiKPi(), 0)`?
> > 
> > 
> > Because we filter candidates that have `BIT(0)` essentially and therefore there is a histogram that corresponds to `TESTBIT(candidate.isSelDplusToPiKPi(), 0)` (i.e. `hPtRecSig` and the corresponding ones for prompt and nonprompt)
> 
> The worry I have is that the `hPtRecSig` et al histograms are meant for the "complete-efficiency" calculation which should include all selection steps but your changes seem to change their purpose to include just the skimming selection step.

And the "complete-efficiency" should be calculated with the same selection as applied in the real-data task. That's why I think that `d_selectionFlagDPlus` should be used in the MC task as well.

fgrosa:
> > > Why is there no histogram for `TESTBIT(candidate.isSelDplusToPiKPi(), 0)`?
> > 
> > 
> > Because we filter candidates that have `BIT(0)` essentially and therefore there is a histogram that corresponds to `TESTBIT(candidate.isSelDplusToPiKPi(), 0)` (i.e. `hPtRecSig` and the corresponding ones for prompt and nonprompt)
> 
> The worry I have is that the `hPtRecSig` et al histograms are meant for the "complete-efficiency" calculation which should include all selection steps but your changes seem to change their purpose to include just the skimming selection step.

This is fine for me but it's not what is in the D0 task (https://github.com/AliceO2Group/O2Physics/blob/a2e8df7e3c3fde78c68dfb59d281534d18e17909/PWGHF/Tasks/taskD0.cxx#L207 this histogram is filled with the loosest selection, i.e. the one in the filter only). Shouldn't we be consistent in the two tasks?

vkucera:
> > > > Why is there no histogram for `TESTBIT(candidate.isSelDplusToPiKPi(), 0)`?
> > > 
> > > 
> > > Because we filter candidates that have `BIT(0)` essentially and therefore there is a histogram that corresponds to `TESTBIT(candidate.isSelDplusToPiKPi(), 0)` (i.e. `hPtRecSig` and the corresponding ones for prompt and nonprompt)
> > 
> > 
> > The worry I have is that the `hPtRecSig` et al histograms are meant for the "complete-efficiency" calculation which should include all selection steps but your changes seem to change their purpose to include just the skimming selection step.
> 
> This is fine for me but it's not what is in the D0 task
> 
> https://github.com/AliceO2Group/O2Physics/blob/a2e8df7e3c3fde78c68dfb59d281534d18e17909/PWGHF/Tasks/taskD0.cxx#L207
> this histogram is filled with the loosest selection, i.e. the one in the filter only
> Shouldn't we be consistent in the two tasks?

You're right. Looking at the D0 task, I think it is not correct. `hPtRecSig` should have the complete selection applied. Can you adjust your code accordingly? I will fix the D0 task once your PR is final.

fgrosa:
> > > > > Why is there no histogram for `TESTBIT(candidate.isSelDplusToPiKPi(), 0)`?
> > > > 
> > > > 
> > > > Because we filter candidates that have `BIT(0)` essentially and therefore there is a histogram that corresponds to `TESTBIT(candidate.isSelDplusToPiKPi(), 0)` (i.e. `hPtRecSig` and the corresponding ones for prompt and nonprompt)
> > > 
> > > 
> > > The worry I have is that the `hPtRecSig` et al histograms are meant for the "complete-efficiency" calculation which should include all selection steps but your changes seem to change their purpose to include just the skimming selection step.
> > 
> > 
> > This is fine for me but it's not what is in the D0 task
> > https://github.com/AliceO2Group/O2Physics/blob/a2e8df7e3c3fde78c68dfb59d281534d18e17909/PWGHF/Tasks/taskD0.cxx#L207
> > 
> > this histogram is filled with the loosest selection, i.e. the one in the filter only
> > Shouldn't we be consistent in the two tasks?
> 
> You're right. Looking at the D0 task, I think it is not correct. `hPtRecSig` should have the complete selection applied. Can you adjust your code accordingly? I will fix the D0 task once your PR is final.

👍 I have added the histograms at `Skim` level and filled the `hPtRecSig` histograms with the step defined in the configurable

PWGHF/Tasks/taskDPlus.cxx
```diff
@@ -112,20 +113,41 @@ struct TaskDPlus {
...
+    registry.add("hPtvsYRecSig_RecoSkim", "3-prong candidates (RecoSkim - matched);#it{p}_{T}^{rec.}; #it{y}", {HistType::kTH2F, {{vbins, "#it{p}_{T} (GeV/#it{c})"}, {100, -5., 5.}}});
+    registry.add("hPtvsYRecSigPrompt_RecoSkim", "3-prong candidates (RecoSkim - matched, prompt);#it{p}_{T}^{rec.}; #it{y}", {HistType::kTH2F, {{vbins, "#it{p}_{T} (GeV/#it{c})"}, {100, -5., 5.}}});
+    registry.add("hPtvsYRecSigNonPrompt_RecoSkim", "3-prong candidates (RecoSkim - matched, non-prompt);#it{p}_{T}^{rec.}; #it{y}", {HistType::kTH2F, {{vbins, "#it{p}_{T} (GeV/#it{c})"}, {100, -5., 5.}}});
+    registry.add("hPtvsYGenPrompt", "3-prong candidates (matched, prompt);#it{p}_{T}^{gen.}; #it{y}", {HistType::kTH2F, {{vbins, "#it{p}_{T} (GeV/#it{c})"}, {100, -5., 5.}}});
+    registry.add("hPtvsYGenNonPrompt", "3-prong candidates (matched, non-prompt);#it{p}_{T}^{gen.}; #it{y}", {HistType::kTH2F, {{vbins, "#it{p}_{T} (GeV/#it{c})"}, {100, -5., 5.}}});
+  }
+
+  Filter filterSelectCandidates = (aod::hf_selcandidate_dplus::isSelDplusToPiKPi > 0);
```
vkucera, line 150: It seems to me that this line prevents calculation of efficiency without applying selector decision as it used to be before when `>= d_selectionFlagDPlus` was used.

PWGHF/TableProducer/HFDplusToPiKPiCandidateSelector.cxx
```diff
@@ -119,13 +119,18 @@ struct HFDplusToPiKPiCandidateSelector {
     // looping over 3-prong candidates
     for (auto& candidate : candidates) {
 
-      // final selection flag: 0 - rejected, 1 - accepted
+      // final selection flag:
+      // 0 - rejected
+      // BIT(0) - bit for D+ candidate
+      // BIT(1) - selected by topological cuts
+      // BIT(2) - selected by PID cuts
```
vkucera, line 126: Since these will be shared among taks, I would be in favour of defining them in a dedicated `enum` in `HFCandidateSelectionTables.h`.
Something like `SelectionSteps`.

PWGHF/Tasks/taskDPlus.cxx
fgrosa: Indeed it does not allow for the selection of all the 3-prong candidates, however what is the interest in selecting all the candidates that are not reconstructed as D+ in the HFTrackIndexSkimsCreator? Filtering the candidates that have at least `BIT(0)` instead should improve a bit the execution time

fgrosa:
Hi @vkucera @ginnocen is there still something you want me to change in this PR or can it be merged? Thanks!

PR #130, opened 2021-09-29
Filling fRICH MC nsigmas

njacazio:
- Add nsigma vs p

### DIFF
--- a/ALICE3/Tasks/pidRICHqa.cxx
+++ b/ALICE3/Tasks/pidRICHqa.cxx
@@ -9,6 +9,12 @@
 // granted to it by virtue of its status as an Intergovernmental Organization
 // or submit itself to any jurisdiction.
 
+///
+/// \file   qaTOFMC.cxx
+/// \author Nicolò Jacazio
+/// \brief  Task to produce QA output of the PID with ALICE3 RICH running on the MC.
+///
+
 // O2 includes
 #include "Framework/AnalysisTask.h"
 #include "ALICE3/DataModel/RICH.h"
@@ -75,7 +81,7 @@ struct richPidQaMc {
   Configurable<int> nBinsDelta{"nBinsDelta", 600, "Number of delta bins"};
   Configurable<float> minDelta{"minDelta", -0.4f, "Minimum delta plotted (rad)"};
   Configurable<float> maxDelta{"maxDelta", 0.4f, "Maximum delta plotted (rad)"};
-  Configurable<int> logAxis{"logAxis", 0, "Flag to use a log momentum axis"};
+  Configurable<int> logAxis{"logAxis", 1, "Flag to use a log momentum axis"};
   Configurable<float> nSigmaNorm{"nSigmaNorm", 0.7071067811865475f, "Normalization for the combined Nsigma"};
 
   static constexpr int Np = 5;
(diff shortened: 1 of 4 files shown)


ALICE3/Tasks/pidRICHqa.cxx
```diff
@@ -354,6 +382,11 @@ struct richPidQaMc {
         histos.fill(HIST("fRICH/pSec"), track.p());
         histos.fill(HIST("fRICH/ptSec"), track.pt());
       }
+      fillNsigmafRICH<0>(track, mcParticle, mcParticles, nsigma);
+      fillNsigmafRICH<1>(track, mcParticle, mcParticles, nsigma);
+      fillNsigmafRICH<2>(track, mcParticle, mcParticles, nsigma);
+      fillNsigmafRICH<3>(track, mcParticle, mcParticles, nsigma);
+      fillNsigmafRICH<4>(track, mcParticle, mcParticles, nsigma);
```
vkucera, line 389: Why do you call the function for all indices and check `pdgCode` each time inside?

njacazio: Because it's different at every call

njacazio: I mean, the first time I check the electron PDG, the second the muon and so on

vkucera: My point is that you call the function for every possible PDG and only inside the function checks whether it matches any value in the `PDGs` array and if it doesn't the function does nothing.
Why don't you first check the particle PDG and then call the function once with the correct index if found?

njacazio: Ah yes, I got your point, but I also need the index to identify the PDG and the name of the histogram that I want to fill.
To get what you suggest done in few lines I would probably need a loop and this cannot be done with the current implementation

njacazio: Or maybe a switch, but it would be more code anyways

vkucera: Something like this?
```c++
for (const uint8_t pidIndex : PDGs) {
  if (std::abs(particle.pdgCode()) == PDGs[pidIndex]) {
    fillNsigmafRICH<pidIndex>(track, mcParticle, mcParticles, nsigma);
    break;
  }
}
```

njacazio: eh but can it be passed to the template?

njacazio: it yes, this is indeed the best solution

ALICE3/Tasks/pidRICHqa.cxx
```diff
@@ -205,6 +216,23 @@ struct richPidQaMc {
     }
   }
 
+  template <uint8_t pidIndex, typename T, typename TTT, typename TT>
+  void fillNsigmafRICH(const T& track, const TTT& particle, const TT& mcParticles, const float& nsigma)
```
vkucera, line 220: I don't see `mcParticles` used anywhere.

njacazio: Indeed it was used in the isphysical primary

ALICE3/Tasks/pidRICHqa.cxx
vkucera: Not sure, but would be nice if it was.

mario-krueger: you can use the static for loop I added a while ago:
https://github.com/AliceO2Group/O2Physics/blob/master/Tutorials/src/histogramRegistry.cxx#L227-L236

njacazio: Hi, so going back on this, the solution proposed by Vit does not compile and the solution proposed by Mario is still a bit obscure for me. Maybe I would keep it like this, which seems the most simple to me

ALICE3/Tasks/pidRICHqa.cxx
njacazio: This I fixed

ALICE3/Tasks/pidRICHqa.cxx
mario-krueger: just for completeness, this is what it would look like with the static for loop:
````
#include "Framework/StaticFor.h"
static_for<0, 4>([&](auto i) {  
  constexpr int pidIndex = i.value;
  fillNsigmafRICH<pidIndex>(track, mcParticle, mcParticles, nsigma);
});
````
this basicall expands the loop at compile time, avoiding having to do this manually

njacazio:
@vkucera I would merge this, could you please take care of it?

vkucera (approved):

PR #132, opened 2021-09-30
Try using team as code owner

TimoWilken:
Hopefully this means that only one person from the team has to approve PRs to `PWGHF`.


ktf:
Notice this will however still prevent someone to self approve PRs.

PR #134, opened 2021-09-30
Centrality task for ALICE 3

ddobrigk:


njacazio:
Hi David, I think somehow there are unwanted changes in the cmakefile

njacazio (approved):
Looking good! In the future we might make sure that the selection is the same in both calibration and user task

ddobrigk:
Thanks a lot, @njacazio ! As for getting the calib from the same task, I think this is the natural purpose of keeping the output histogram there: we can easily use that :-)

njacazio:
I meant more, getting the limits from the ccdb object together with the calibration

PR #136, opened 2021-10-01
coll, cent tables -> same size

ddobrigk:


njacazio:
Hi @ddobrigk great! But wouldn't it be better to run the check before and return in case it is not passed? Even before fetching the ccdb object?

ddobrigk:
Okay! Let me do it that way (coding wise it's simpler, even). Will close and open another one!

PR #139, opened 2021-10-01
PWGHF: Add (ALICE3) centrality selection in the Λc task

Katka19:
Added (ALICE3) centrality as another dimension of the inv. mass histogram to be able to select offline for the desired centrality range.


vkucera:
As far as I can see, this will make the Lc analysis impossible to run in Run 3 input.

vkucera (approved):

vkucera:
To be completed in the next PR.

PR #143, opened 2021-10-04
quick fix to make track selection run with run3 sim

mario-krueger:


iarsene (dismissed):

Common/TableProducer/trackselection.cxx
```diff
@@ -40,6 +43,11 @@ struct TrackSelectionTask {
   {
     globalTracks = getGlobalTrackSelection();
     globalTracksSDD = getGlobalTrackSelectionSDD();
+
+    if (isRun3) {
+      globalTracks.SetTrackType(o2::aod::track::TrackTypeEnum::Track);
+      globalTracksSDD.SetTrackType(o2::aod::track::TrackTypeEnum::Track);
```
jgrosseo, line 49: We should only set it for globalTracks, right?

jgrosseo (approved):

PR #146, opened 2021-10-04
[O2-2514]Tutorials: avoid grouping in mc-histograms example

aalkin:
Avoids grouping between mc collisions and collisions so that mc-histograms example does not crash with Run 3 simulations that can occasionally have mc collision labels in decreasing order.

### DIFF
--- a/Tutorials/src/mcHistograms.cxx
+++ b/Tutorials/src/mcHistograms.cxx
@@ -92,19 +92,25 @@ struct AccessMcTruth {
 // For each of them get the corresponding tracks
 struct LoopOverMcMatched {
   OutputObj<TH1F> etaDiff{TH1F("etaDiff", ";eta_{MC} - eta_{Rec}", 100, -2, 2)};
-  void process(aod::McCollision const& mcCollision, soa::Join<aod::McCollisionLabels, aod::Collisions> const& collisions,
+  void process(aod::McCollisions const& mcCollisions, soa::Join<aod::McCollisionLabels, aod::Collisions> const& collisions,
                soa::Join<aod::Tracks, aod::McTrackLabels> const& tracks, aod::McParticles const& mcParticles)
   {
-    // access MC truth information with mcCollision() and mcParticle() methods
-    LOGF(info, "MC collision at vtx-z = %f with %d mc particles and %d reconstructed collisions", mcCollision.posZ(), mcParticles.size(), collisions.size());
-    for (auto& collision : collisions) {
-      LOGF(info, "  Reconstructed collision at vtx-z = %f", collision.posZ());
+    auto collision = collisions.begin();
+    for (auto& mcCollision : mcCollisions) {
+      // access MC truth information with mcCollision() and mcParticle() methods
+      LOGF(info, "MC collision at vtx-z = %f with %d mc particles and %d reconstructed collisions", mcCollision.posZ(), mcParticles.size(), collisions.size());
+      while (collision.mcCollisionId() < mcCollision.globalIndex()) {
+        ++collision;
+      }
+      if (collision.mcCollisionId() == mcCollision.globalIndex()) {
+        LOGF(info, "  Reconstructed collision at vtx-z = %f", collision.posZ());
 
-      // NOTE this will be replaced by a improved grouping in the future
-      auto groupedTracks = tracks.sliceBy(aod::track::collisionId, collision.globalIndex());
-      LOGF(info, "  which has %d tracks", groupedTracks.size());
-      for (auto& track : groupedTracks) {
-        etaDiff->Fill(track.mcParticle().eta() - track.eta());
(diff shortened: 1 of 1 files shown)


jgrosseo (changes requested):
I don't agree. We should first see if we cannot fix the data...

aalkin:
@jgrosseo that is true, the offending case is a reco event, that has 22 bcs difference with generated event, so it is something to be investigated. Regardless, using grouping for tables that are mostly 1-to-1 is a performance overhead, for the specific case of reversing index col -> mccol to mccol -> col it would make sense to use index table or something similar.

jgrosseo:
Well, here the idea is to show the grouping of mcParticles -> mcCollision which otherwise has to be done by hand and each time the groups are built (are we caching this now, btw?)

aalkin:
The algorithm with an "almost synchronous" iterator that I've put in now is simpler - collisions table is scrolled until the match is found, naturally works for multiple reco collisions for one generated. Since the fraction of not reconstructed events is not large, this is also efficient, as collisions table will be only scrolled by a few rows at a time. 

Group caching is not yet implemented.

shahor02:
Hi, such broken MC order is perfectly normal from a reconstruction point of view. The reconstructed vertices are sorted in time and due to the finite time resolution, it is possible that the times assigned to vertices from 2 close MC collisions are swapped in time.
Even worse, if e.g. group of ITS tracks from collision A were matched to TPC tracks from collision B, they time of tracks labeled as B might correspond to time A, leading to inversion by ~10\mus.

Tutorials/src/mcHistograms.cxx
```diff
@@ -92,19 +92,25 @@ struct AccessMcTruth {
...
                soa::Join<aod::Tracks, aod::McTrackLabels> const& tracks, aod::McParticles const& mcParticles)
   {
-    // access MC truth information with mcCollision() and mcParticle() methods
-    LOGF(info, "MC collision at vtx-z = %f with %d mc particles and %d reconstructed collisions", mcCollision.posZ(), mcParticles.size(), collisions.size());
-    for (auto& collision : collisions) {
-      LOGF(info, "  Reconstructed collision at vtx-z = %f", collision.posZ());
+    auto collision = collisions.begin();
+    for (auto& mcCollision : mcCollisions) {
+      // access MC truth information with mcCollision() and mcParticle() methods
+      LOGF(info, "MC collision at vtx-z = %f with %d mc particles and %d reconstructed collisions", mcCollision.posZ(), mcParticles.size(), collisions.size());
```
jgrosseo, line 101: I don't see this for mcParticles. Your PR changes the printout of the number of mcParticles *per* collision to a printout of the total length of the mcParticles table. Am I missing something?

jgrosseo:
> 
> 
> Hi, such broken MC order is perfectly normal from a reconstruction point of view. The reconstructed vertices are sorted in time and due to the finite time resolution, it is possible that the times assigned to vertices from 2 close MC collisions are swapped in time. Even worse, if e.g. group of ITS tracks from collision A were matched to TPC tracks from collision B, they time of tracks labeled as B might correspond to time A, leading to inversion by ~10\mus.

Yes, we may need to swap them in the AOD writer. I still want to check the file myself tomorrow...

aalkin:
```
colID  71  col bcs  6716000  associated to MC  84  bcs  6716000  bcdiff  0
colID  72  col bcs  6732871  associated to MC  87  bcs  6732870  bcdiff  -1
colID  73  col bcs  6732874  associated to MC  86  bcs  6732852  bcdiff  -22
colID  74  col bcs  6739596  associated to MC  88  bcs  6739594  bcdiff  -2
colID  75  col bcs  6740336  associated to MC  89  bcs  6740336  bcdiff  0
```

@shahor02 this is the example of such issue, pilot beam settings. 22 bcs seems a bit excessive. Is this something we should expect?
@jgrosseo swapping MC collisions will break their BC ordering, note how MC col 86 has smaller BC than MC col 87.

shahor02:
> 22 bcs seems a bit excessive. Is this something we should expect?

This corresponds to 0.5 \mus, one can easily have a stray vertex made of 2 low pt tracks, in this case the time resolution may reach a few \mus.

jgrosseo:
@aalkin Why the index **to something** needs to be sorted?

jgrosseo:
I am sure this is not the case for MC particles and their mothers/daughters (it is very random)...

aalkin:
> @aalkin Why the index **to something** needs to be sorted?

In this case we have three related tables, collisions, mc collisions and BCs, with an additional index between collisions and mc collisions. Since this cannot be made consistent due to time resolution, we need to decide which ordering is primary. I assume building index tables with mccols and BCs as components will be in much less demand than going from cols to mccols and back, so BC ordering is less important for mccols.

jgrosseo:
We are discussing this further in the development channel.

PR #149, opened 2021-10-05
Add first EventTime maker task

njacazio:


njacazio:
@noferini @ercolessi and here is the workflow `o2-analysis-tof-collisiontime --aod-file AO2D.root -b --aod-memory-rate-limit 1000000000 --add-qa 1`

njacazio:
This will be part of the PID task but for the development stage we would prefer to keep it separate.
Furthermore, there will anyway be the need for a Run2 task and a Run3 one, this will be needed mostly for the Run3 one as for the converted data the approach is a bit different.
In short: we plan to merge it with the PID task as soon as soon as the workflow is stable enough.
Would this be good?

Common/TableProducer/tofcollisiontimemaker.cxx
```diff
@@ -0,0 +1,126 @@
...
+
+namespace o2::aod
+{
+namespace tofeventtime
+{
+DECLARE_SOA_COLUMN(TOFEvTime, tofEvTime, float);       //! TOF event time
+DECLARE_SOA_COLUMN(TOFEvTimeErr, tofEvTimeErr, float); //! TOF event time error
+} // namespace tofeventtime
+
+DECLARE_SOA_TABLE(TOFEventTime, "AOD", "TOFSignal", //! Table of the TOF event time
```
jgrosseo, line 47: Can you unify the names?

njacazio: This I missed, thanks for spotting !

jgrosseo:
If this is a development task only, shall we put it in PWGLF?

Common/TableProducer/tofcollisiontimemaker.cxx
```diff
@@ -0,0 +1,126 @@
...
+    auto evTime = o2::tof::evTimeMaker(tracks);
+    for (auto& t : tracks) {
+      // Expected times
+      const float expPi = responsePi.GetExpectedSignal(t);
+      const float expKa = responseKa.GetExpectedSignal(t);
+      const float expPr = responsePr.GetExpectedSignal(t);
+      // Expected resolutions
+      const float expResoPi = responsePi.GetExpectedSigma(response, t);
+      const float expResoKa = responseKa.GetExpectedSigma(response, t);
+      const float expResoPr = responsePr.GetExpectedSigma(response, t);
```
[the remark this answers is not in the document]
njacazio: These are not used yet, we will use them for the event time determination

njacazio:
Indeed, alternatively we can also work on a private branch and not have the code merged yet, maybe this is the quickest solution

njacazio:
Thanks for the review Jan, I will now close this PR and we proceed offline until we have something more concrete

PR #152, opened 2021-10-05
Add helper class for bitwise cut selection

lbariogl:
Implements track  selections on `femtodreamparticle::cut`.

### DIFF
--- a/PWGCF/FemtoDream/FemtoDreamTrackBitmask.h
+++ b/PWGCF/FemtoDream/FemtoDreamTrackBitmask.h
@@ -1,0 +1,108 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+/// \file FemtoDreamTrackBitmpask.h
+/// \brief Definition of the FemtoDreamTrackBitmpask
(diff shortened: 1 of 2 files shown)


jgrosseo (commented):
My understanding was that the meaning of the bits is freely configurable when femtodream produces derived data? 
If so, what would happen to this header file then?

iarsene:
In DQ we have a similar approach, of storing a bit map for different selections, but at the moment we don't foresee a fixed mapping of these cuts because i can't estimate now how diverse this will be or how it will change in the beginning phase. Alternatively, in dq we have a cut library, and in skimming we should use a list of cuts strictly from there. Each cut will be provided by name in a Configurable, so we can just create a histogram with bins labeled with cut names used, which will correspond to each filtering bit. So then this should ensure documentation on which cuts were used at skimming time. Not sure if this helps your case though.

jgrosseo:
There is also the cutculator in o2physics which Andi has created for this.

jgrosseo:
Imagine today you produce some data with a certain bit configuration. In two weeks with another configuration. And you want to still use both configurations with O2physics, but you only have one header, then it doesn't work.

I think Andi's idea was to use this for determining which bit mask to put in the filter:
https://github.com/AliceO2Group/O2Physics/blob/master/PWGCF/FemtoDream/FemtoDreamCutculator.h

@sartozza What do you think?

jgrosseo:
Yes, I think you can do it with the cutculator and a correct usage of the bit mask. In your example what you would need an expression for which it is enough that any of the three bits match
`(bitmask & 111) != 0`
This gets a little tricky of you match different areas of the bit mask at the same time, but should be possible with correct masks. We discussed this with Andi when he was implementing it

lbariogl:
Thank you very much for your suggestion! I was exactly thinking of checking different area of the bitmask at the same time. This could be a problem because, as the structure is defined by now, the order of the selections can change and this means that the bitmasks must change accordingly. For this reason, I think that there are two options:
1. the cutculator returns exactly the bitmask corresponding to the selections we are interesed in
2. the bitmask must be built reading the `dpl-config.json` attached to the AO2D.root, putting the selections in the right order.
Am I missing something?

PR #154, opened 2021-10-06
aod-merger: Generalize negative index offsetting

sawenzel:
When AODs are merged, it could happen that the input AODs
already had shifted 'unassigned' indices. This commit, makes sure they are
taken into account properly. In result, we should now be able
to perform multi-stage merges correctly.


jgrosseo (approved):

jgrosseo:
Thanks, @sawenzel

PR #165, opened 2021-10-08
PWGHF: DDbar framework switched from cent to mult selection (ALICE3)

fcolamar:


PWGHF/Tasks/HFCorrelatorD0D0bar.cxx
```diff
@@ -101,14 +102,25 @@ struct HfCorrelatorD0D0bar {
 
   Filter filterSelectCandidates = (aod::hf_selcandidate_d0::isSelD0 >= selectionFlagD0 || aod::hf_selcandidate_d0::isSelD0bar >= selectionFlagD0bar);
 
-  void process(soa::Join<aod::Collisions, aod::Cents>::iterator const& collision, soa::Filtered<soa::Join<aod::HfCandProng2, aod::HFSelD0Candidate>> const& candidates)
+  void process(aod::Collision const& collision, soa::Join<aod::Tracks, aod::TracksExtended>& tracks, soa::Filtered<soa::Join<aod::HfCandProng2, aod::HFSelD0Candidate>> const& candidates)
   {
-    float centrality = collision.centV0M();
-    registry.fill(HIST("hCentralityPreSelection"), centrality);
-    if (centrality < centMin || centrality > centMax) {
+    int nTracks = 0;
```
aalkin, line 107: All the checked columns for tracks seem to be non-dynamic, one can a `Partition` or `.select()` method for the tracks table to get the set directly and then request its `.size()` instead of doing a loop.

vkucera: I suppose that the issues with the centrality task will be resolved soon and all the track counting code will be removed then.

fcolamar: Yes this is kind of temporary, the idea is to revert back to the centrality selection as it was implemented in the previous code version

ginnocen (approved):

PR #180, opened 2021-10-12
Fix cascade issues in analysis 

ddobrigk:
see https://alice.its.cern.ch/jira/browse/O2-2635


iarsene:
The test fails due to an error, not a warning, and its at the line where you two were discussing. I think the kTRUE might be the problem. I also think one should use "true".

jgrosseo (approved):

PR #190, opened 2021-10-14
Event selection and multiplicity tasks Run2/Run3 from hyperloop

victor-gonzalez:
For Run3 now the timestamp task shall not be incorporated to
the workflow, it gives

Exception caught: Unable to find OutputSpec with label Timestamps

### DIFF
--- a/Common/TableProducer/multiplicityTable.cxx
+++ b/Common/TableProducer/multiplicityTable.cxx
@@ -28,9 +28,11 @@ void customize(std::vector<ConfigParamSpec>& workflowOptions)
 
 struct MultiplicityTableTaskIndexed {
   Produces<aod::Mults> mult;
-  Partition<aod::Tracks> tracklets = (aod::track::trackType == static_cast<uint8_t>(o2::aod::track::TrackTypeEnum::Run2Tracklet));
+  Partition<aod::Tracks> run2tracklets = (aod::track::trackType == static_cast<uint8_t>(o2::aod::track::TrackTypeEnum::Run2Tracklet));
+  /* the Run3 tracklets type needs to be properly incorporated when defined */
+  Partition<aod::Tracks> run3tracklets = (aod::track::trackType == static_cast<uint8_t>(o2::aod::track::TrackTypeEnum::Run2Tracklet));
 
-  void process(aod::Run2MatchedSparse::iterator const& collision, aod::Tracks const& tracks, aod::BCs const&, aod::Zdcs const&, aod::FV0As const& fv0as, aod::FV0Cs const& fv0cs, aod::FT0s const& ft0s)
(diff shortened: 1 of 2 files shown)


victor-gonzalez:
If approved we have to communicate that `--selection-run 2` is not valid any longer, neither `--selection-run 3` 
By default both tasks run under Run2 conditions
Only if from `hyperloop` or from the `json` file the `processRun3`s are activated, the tasks will run under Run3 conditions

Common/TableProducer/multiplicityTable.cxx
```diff
@@ -28,17 +28,19 @@ void customize(std::vector<ConfigParamSpec>& workflowOptions)
 
 struct MultiplicityTableTaskIndexed {
   Produces<aod::Mults> mult;
-  Partition<aod::Tracks> tracklets = (aod::track::trackType == static_cast<uint8_t>(o2::aod::track::TrackTypeEnum::Run2Tracklet));
+  Partition<aod::Tracks> run2tracklets = (aod::track::trackType == static_cast<uint8_t>(o2::aod::track::TrackTypeEnum::Run2Tracklet));
+  /* the Run3 tracklets type needs to be properly incorporated when defined */
+  Partition<aod::Tracks> run3tracklets = (aod::track::trackType == static_cast<uint8_t>(o2::aod::track::TrackTypeEnum::Run2Tracklet));
```
jgrosseo, line 33: There is no Run3 tracklet. This can go

alibuild (approved):

jgrosseo:
> 
> 
> For Run3 now the timestamp task shall not be incorporated to the workflow, it gives
> 
> Exception caught: Unable to find OutputSpec with label Timestamps

@victor-gonzalez To solve this, I propose to change to
using BCsWithRun3Matchings = soa::Join<aod::BCs, aod::Run3MatchedToBCSparse, aod::Timestamps>;
(even if it is not used). I worry it creates confusion if we have to remove the timestamp task in one case (and it will be needed for CCDB anyway). Could you try this locally?

victor-gonzalez:
> > For Run3 now the timestamp task shall not be incorporated to the workflow, it gives
> > Exception caught: Unable to find OutputSpec with label Timestamps
> 
> @victor-gonzalez To solve this, I propose to change to using BCsWithRun3Matchings = soa::Join<aod::BCs, aod::Run3MatchedToBCSparse, aod::Timestamps>; (even if it is not used). I worry it creates confusion if we have to remove the timestamp task in one case (and it will be needed for CCDB anyway). Could you try this locally?

@jgrosseo The order matters!!
```
using BCsWithRun3Matchings = soa::Join<aod::BCs, aod::Timestamps, aod::Run3MatchedToBCSparse>; 
```
(I didn't know it)
Being produced!
I will test it and push the PR

PR #191, opened 2021-10-15
Use PR author to determine whether to auto-approve

TimoWilken:
Previously, the user who enabled auto-merge was used instead.


TimoWilken:
See https://github.com/AliceO2Group/O2Physics/pull/192 instead.

PR #202, opened 2021-10-19
TableMaker multi-process functions to allow producing barrel-only, muon-only or combined skims, and allow running w/ or w/o the centrality, filterPP or v0-selector tasks

iarsene:


alibuild (approved):

PR #204, opened 2021-10-20
PWGHF: Minor improvements for execution time

fgrosa:
Two improvements:
- return pointer to histograms when added to the registry to avoid to search for the histogram in the registry every time is filled
- cache indices of cut variables in each selection method


PWGHF/TableProducer/HFTrackIndexSkimsCreator.cxx
```diff
@@ -117,9 +118,10 @@ struct HfTagSelCollisions {
     labels[2 + EventRejection::NContrib] = "rej. # of contributors";
     labels[2 + EventRejection::Chi2] = "rej. #it{#chi}^{2}";
     AxisSpec axisEvents = {nBinsEvents, 0.5, nBinsEvents + 0.5, ""};
-    registry.add("hEvents", "Events;;entries", HistType::kTH1F, {axisEvents});
+    histos[0] = std::get<std::shared_ptr<TH1>>(registry.add("hEvents", "Events;;entries", HistType::kTH1F, {axisEvents}));
```
mario-krueger, line 121: you can just do `registry.add<TH1>(...)`

fgrosa: Thanks for the suggestion!

fgrosa: It's likely that the overhead is small as you say, however I am trying to improve this task as much as possible since it often gives problems of CPU time. Let's see also what @vkucera and @ginnocen thinks about this.

ginnocen (dismissed):

fgrosa:
> hi @fgrosa, could you please fix the conflict?

Hi @ginnocen I fixed the comments, but I also reverted it to draft since there are a couple of things that I still want to test and might enter in this PR

ginnocen:
hi @fgrosa. OK thanks.

fgrosa:
Hi @ginnocen in the end I managed to obtain an improvement in the execution time by adding the `px`, `py`, and `pz` static columns in the track table. In particular, using the same Pb-Pb converted files with the old version of the code this was the time:
image
while with the new version:
image
I also tried to use expression columns extending the table instead of static columns (following this example [extendedColumns.cxx](https://github.com/AliceO2Group/O2Physics/blob/b20eae8268edfeb0ba78ae9e7492a3d9f77c715d/Tutorials/src/extendedColumns.cxx)), however in this case the execution time increases by about a factor 3 (and I am not sure why, maybe @aalkin do you have some idea?)
image
So overall we gain ~3/2 of the execution time, even if we have 3 more floats in our hfsel track table (which means a bit more memory used, but in Pb-Pb this is negligible compared to the size of the candidate table).

From my side this PR is now ready for review (I might try some other improvement in the future, but in case I will open a new PR). Thanks!

PWGHF/DataModel/HFSecondaryVertex.h
```diff
@@ -42,12 +42,18 @@ namespace hf_seltrack
 DECLARE_SOA_COLUMN(IsSelProng, isSelProng, int); //!
 DECLARE_SOA_COLUMN(DCAPrim0, dcaPrim0, float);   //!
 DECLARE_SOA_COLUMN(DCAPrim1, dcaPrim1, float);   //!
+DECLARE_SOA_COLUMN(PxProng, pxProng, float);     //!
+DECLARE_SOA_COLUMN(PyProng, pyProng, float);     //!
+DECLARE_SOA_COLUMN(PzProng, pzProng, float);     //!
```
jgrosseo, line 47: Is this just to speed up? Is it filled with the propagated track information?

fgrosa: Hi @jgrosseo, 
yes indeed this is for speed up (see comment above) and it is before the propagation to the secondary vertex (actually before the secondary vertex reconstruction) essentially to avoid to recompute `px`, `py`, and `pz` in the preselection methods https://github.com/AliceO2Group/O2Physics/blob/b189d8a1de937d29583c7f14339964a03a255f2d/PWGHF/TableProducer/HFTrackIndexSkimsCreator.cxx#L730-L733 (which are called once for candidate, that in Pb-Pb is much larger than the number of tracks).
I also tried with expression columns, but in that case the execution time increases significantly (see above).

jgrosseo: OK, we follow this up offline. You can go ahead with this PR in the meanwhile.

aalkin:
> (following this example extendedColumns.cxx), however in this case the execution time increases by about a factor 3

That example is actually wrong - the extension is done inside collision loop - I'll fix it.  Probably that increase in execution time is also due to extension being done in a loop - note that this temporary extension is not cached in any way and is lost as soon as the variable goes out of scope.

fgrosa:
> > (following this example extendedColumns.cxx), however in this case the execution time increases by about a factor 3
> 
> That example is actually wrong - the extension is done inside collision loop - I'll fix it. Probably that increase in execution time is also due to extension being done in a loop - note that this temporary extension is not cached in any way and is lost as soon as the variable goes out of scope.

Thanks! I saw your PR https://github.com/AliceO2Group/O2Physics/pull/272, I will try to see if I can get similar performance implementing it similarly. I have then another question then, why in case of extended tables you use `Spawns`,
https://github.com/AliceO2Group/O2Physics/blob/a4c398da4548655a21fc0883ac705e9f6461c09c/Tutorials/src/extendedTables.cxx#L109
and in the extended columns not?

aalkin:
> Thanks! I saw your PR #272, I will try to see if I can get similar performance implementing it similarly. I have then another question then, why in case of extended tables you use `Spawns`,

These are different mechanisms: `Spawns<>` creates a table that can be subscribed by other tasks, `Extend<>` only creates a temporary extension, accessible only within the scope where it was created.

fgrosa:
> > Thanks! I saw your PR #272, I will try to see if I can get similar performance implementing it similarly. I have then another question then, why in case of extended tables you use `Spawns`,
> 
> These are different mechanisms: `Spawns<>` creates a table that can be subscribed by other tasks, `Extend<>` only creates a temporary extension, accessible only within the scope where it was created.

Thanks!

vkucera (approved):

PR #208, opened 2021-10-20
Revert "remove MC flag"

victor-gonzalez:
Reverts AliceO2Group/O2Physics#201


victor-gonzalez:
I'm not able to check it because I'm in the middle of a full update
Not ready yet

jgrosseo:
I think we should fix the CCDB instead. What do you think?

PR #238, opened 2021-10-31
PWGHF: Use process switches instead of workflow options for DPlus task

fgrosa:


ginnocen (approved):

fgrosa:
@ginnocen @vkucera can this be merged in your opinion? Thanks!

ginnocen (approved):

PR #187, opened 2021-10-13
Run 3 DCA extraction

victor-gonzalez:
I created it in a specific branch to have my master available

I will transfer the comments from the previous PR draft
After that I will update it with the comments implemented

PR #248, opened 2021-11-04
Add DCAz histograms

mfaggin:
- `doMC` flag set to `false` by default, to be able to run on data (no `PROCESS_SWITCH` implemented yet)
- DCAz distributions added
- for MC version of the task, propagation toward the MC vertex added (dirty solution)


mfaggin:
@jgrosseo I inserted the possibility to evaluate also the DCAz component. In addition, I added some other feature for the MC, even if still as "dirty solutions". Nevertheless, we would ask to approve this code to be ready to process the async reconstructed data for the pilot run. The plan is to clean the code and merge it with `PWGPP/Tasks/qaEventTrack.cxx.cxx` in collaboration with @mario-krueger .

victor-gonzalez:
Be aware your code will only run locally and having the proper files in the cwd
I would suggest to implement what is in the latest version of https://github.com/AliceO2Group/O2Physics/pull/187
Otherwise I guess you will not be able to test it in hyperloop

mfaggin:
> Be aware your code will only run locally and having the proper files in the cwd I would suggest to implement what is in the latest version of #187 Otherwise I guess you will not be able to test it in hyperloop
> 
> Ok! I saw you only run on data So you will not have problems for running it on hyperloop Sorry about the noise

Indeed by default I only run on data. No problem at all, thanks for your comment!

mfaggin:
@jgrosseo I have just implemented the possibility to configure the binning and range of the impact parameter axis at runtime, just for flexibility.

mfaggin:
Thanks a lot @mario-krueger , I followed your suggestion. It is implemented in the last commit.

mfaggin:
@jgrosseo do you think it is ok to go on?

jgrosseo (approved):

PR #249, opened 2021-11-04
make sure to access all three track tables

jgrosseo:
to keep benchmarks consistent...


alibuild (approved):

PR #251, opened 2021-11-05
Use getNominalField function

jgrosseo:


alibuild (approved):

jgrosseo:
Requires new O2 tag. Leaving it pending because it will be automatically rechecked.

victor-gonzalez (approved):

PR #259, opened 2021-11-08
Add histogram in init for SameEvent causing error with recent histogr…

sartozza:
Connected to merged PR: https://github.com/AliceO2Group/O2Physics/pull/250
Missing histogram in the init of FemtoDreamContainer was not causing error during running on Friday 05.11.2021.
Error after running tonight tag.
Added histogram to init.


mario-krueger:
interesting...
trying to fill a histogram that does not exist should have always given a fatal error
if not, we should investigate! What exactly is the error message that triggered this PR?

victor-gonzalez:
Yeah
Without this PR you should have gotten a fatal error
Are you sure your tests on 05.11.21 incorporated #250?

victor-gonzalez (approved):

PR #261, opened 2021-11-08
Compiler warning fix and use getNominalL3Field function

victor-gonzalez:


Common/TableProducer/trackextension.cxx
```diff
@@ -58,7 +58,7 @@ struct TrackExtension {
   Service<o2::ccdb::BasicCCDBManager> ccdb;
 
   int mRunNumber;
-  float mMagField;
+  int mMagField;
```
ktf, line 61: If you are going to convert the mag field to float in any case, I am not sure this is a good idea.

victor-gonzalez: You are right!
We will get it as an `int` but if we store it as such we will need the conversion on a per particle basis!
I will store it as a float then
Thanks!!

jgrosseo:
I propose to change mMagField to int to make it consistent

jgrosseo (approved):

jgrosseo:
@victor-gonzalez I propose to move the initializing of the members out of init and directly where you declare the variables.

PR #264, opened 2021-11-10
Updates for the skimming tasks (both data and MC); removed redundant tableMaker_PbPb.cxx task; other smaller changes

iarsene:
TableMakerMC: 
1) multiple process functions enabled;
2) muon tables skimming added;
3) track covariance tables subscription removed from default behaviour -> user must explicitly enable them;
4) skimming statistics histgrams added

TableMaker:
1) Modified slightly the available process functions;
2) Removed track covariance tables subscription from default behaviour -> user must explicitly enable them;

tableMaker_PbPb.cxx file removed since its case is covered in one of the process functions from TableMaker

ReducedInfoTables.h
MC data model changed such that the MC reduced event table becomes an independent table.
An additional event labels table was created, joinable with the ReducedEvents table.
A label table was added also for the reduced muons.

dqEfficiency.cxx
Small changes to accomodate the updates above.


alibuild (approved):

PR #265, opened 2021-11-10
Filter task separated in its own task dptdpt-filter

victor-gonzalez:
It feeds
- match-reco-gen independent separated from the filter-qa task
- dptdptcorrelations independent
- dptdpt-filter-qa independent separated from the match-reco-gen task

### DIFF
--- a/PWGCF/DataModel/DptDptFiltered.h
+++ b/PWGCF/DataModel/DptDptFiltered.h
@@ -21,42 +21,42 @@ namespace aod
 /* we have to change from int to bool when bool columns work properly */
 namespace dptdptfilter
 {
-DECLARE_SOA_COLUMN(EventAccepted, eventaccepted, uint8_t); //! If the collision/event has been accepted or not
-DECLARE_SOA_COLUMN(EventCentMult, centmult, float);        //! The centrality/multiplicity pecentile
+DECLARE_SOA_COLUMN(DptDptCFCollisionAccepted, collisionaccepted, bool); //! If the collision/event has been accepted or not
+DECLARE_SOA_COLUMN(DptDptCFCollisionCentMult, centmult, float);         //! The centrality/multiplicity pecentile
 } // namespace dptdptfilter
-DECLARE_SOA_TABLE(AcceptedEvents, "AOD", "ACCEPTEDEVENTS", //! Accepted reconstructed collisions/events filtered table
+DECLARE_SOA_TABLE(DptDptCFAcceptedCollisions, "AOD", "DPTDPTCFACCCOLL", //! Accepted reconstructed collisions/events filtered table
                   o2::soa::Index<>,
                   collision::BCId,
                   collision::PosZ,
-                  dptdptfilter::EventAccepted,
-                  dptdptfilter::EventCentMult);
-using AcceptedEvent = AcceptedEvents::iterator;
-DECLARE_SOA_TABLE(AcceptedTrueEvents, "AOD", "ACCTRUEEVENTS", //! Accepted generated collisions/events filtered table
+                  dptdptfilter::DptDptCFCollisionAccepted,
+                  dptdptfilter::DptDptCFCollisionCentMult);
+using DptDptCFAcceptedCollision = DptDptCFAcceptedCollisions::iterator;
+DECLARE_SOA_TABLE(DptDptCFAcceptedTrueCollisions, "AOD", "DPTCFACCGENCOLL", //! Accepted generated collisions/events filtered table
                   o2::soa::Index<>,
                   collision::BCId,
                   mccollision::PosZ,
-                  dptdptfilter::EventAccepted,
-                  dptdptfilter::EventCentMult);
-using AcceptedTrueEvent = AcceptedTrueEvents::iterator;
+                  dptdptfilter::DptDptCFCollisionAccepted,
+                  dptdptfilter::DptDptCFCollisionCentMult);
+using DptDptCFAcceptedTrueCollision = DptDptCFAcceptedTrueCollisions::iterator;
 namespace dptdptfilter
 {
-DECLARE_SOA_INDEX_COLUMN(AcceptedEvent, event);                      //! Reconstructed collision/event
-DECLARE_SOA_INDEX_COLUMN(AcceptedTrueEvent, mcevent);                //! Generated collision/event
-DECLARE_SOA_COLUMN(TrackacceptedAsOne, trackacceptedasone, uint8_t); //! Track accepted as type one
-DECLARE_SOA_COLUMN(TrackacceptedAsTwo, trackacceptedastwo, uint8_t); //! Track accepted as type two
-DECLARE_SOA_COLUMN(Pt, pt, float);                                   //! The track transverse momentum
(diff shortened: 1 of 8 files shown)


victor-gonzalez:
The filter task produced a filtered table for use of dptdpt correlations, the matching reco-gen and the filter-qa.
It is not a skimming task. The idea is that the filter task will filter the skimmed data for the above tasks as well

PWGCF/DataModel/DptDptFiltered.h
```diff
@@ -0,0 +1,68 @@
...
+#include "Framework/AnalysisDataModel.h"
+
+namespace o2
+{
+namespace aod
+{
+/* we have to change from int to bool when bool columns work properly */
+namespace dptdptfilter
+{
+DECLARE_SOA_COLUMN(EventAccepted, eventaccepted, uint8_t); //! If the collision/event has been accepted or not
```
jgrosseo, line 24: To my understanding `bool` is OK now (see TrackSelection.h`)

PWGCF/DataModel/DptDptFiltered.h
```diff
@@ -0,0 +1,68 @@
...
+DECLARE_SOA_COLUMN(EventCentMult, centmult, float);        //! The centrality/multiplicity pecentile
+} // namespace dptdptfilter
+DECLARE_SOA_TABLE(AcceptedEvents, "AOD", "ACCEPTEDEVENTS", //! Accepted reconstructed collisions/events filtered table
+                  o2::soa::Index<>,
+                  collision::BCId,
```
jgrosseo, line 29: Why do you keep the BCid?
Do you need the timestamp to later retrieve the magnetic field (or directly the magnetic field)?

PWGCF/DataModel/DptDptFiltered.h
```diff
@@ -0,0 +1,68 @@
...
+{
+namespace aod
+{
+/* we have to change from int to bool when bool columns work properly */
+namespace dptdptfilter
+{
+DECLARE_SOA_COLUMN(EventAccepted, eventaccepted, uint8_t); //! If the collision/event has been accepted or not
+DECLARE_SOA_COLUMN(EventCentMult, centmult, float);        //! The centrality/multiplicity pecentile
+} // namespace dptdptfilter
+DECLARE_SOA_TABLE(AcceptedEvents, "AOD", "ACCEPTEDEVENTS", //! Accepted reconstructed collisions/events filtered table
```
jgrosseo, line 27: You may want to call it Collision instead of Event to follow the usual nomenclature.
Also this name has to be unique ACCEPTEDEVENTS in O2, so let's call it CFDPTCOLLISONS or similar

PWGCF/DataModel/DptDptFiltered.h
```diff
@@ -0,0 +1,68 @@
...
+                  o2::soa::Index<>,
+                  collision::BCId,
+                  collision::PosZ,
+                  dptdptfilter::EventAccepted,
+                  dptdptfilter::EventCentMult);
```
jgrosseo, line 32: Do you miss an index to acceptedTrueEvent here?

PWGCF/TableProducer/dptdptfilter.cxx
```diff
@@ -0,0 +1,935 @@
...
+
+const char* speciesName[kDptDptNoOfSpecies] = {"h", "e", "mu", "pi", "ka", "p"};
+
+const char* speciesTitle[kDptDptNoOfSpecies] = {"", "e", "#mu", "#pi", "K", "p"};
+
+//============================================================================================
+// The DptDptFilter output objects
+//============================================================================================
+std::string fTaskConfigurationString = "PendingToConfigure";
+TH1F* fhCentMultB = nullptr;
```
jgrosseo, line 77: No histogram registry?

PWGCF/TableProducer/dptdptfilter.cxx
```diff
@@ -0,0 +1,935 @@
...
+  template <typename ParticleObject>
+  inline MatchRecoGenSpecies IdentifyParticle(ParticleObject const& particle)
+  {
+    using namespace dptdptfilter;
+
+    constexpr int pdgcodeEl = 11L;
+    constexpr int pdgcodeMu = 13L;
+    constexpr int pdgcodePi = 211L;
+    constexpr int pdgcodeKa = 321L;
+    constexpr int pdgcodePr = 2212L;
```
jgrosseo, line 320: Why L if it is an int then?

jgrosseo (commented):
See some comments inline

PWGCF/DataModel/DptDptFiltered.h
victor-gonzalez: Yeah, this has been in this shape from the beginning 
If now it is supported I will modify it then

PWGCF/DataModel/DptDptFiltered.h
victor-gonzalez: Not really. I introduced it when we were tracking the reco-gen issues
I will keep it for a while and probably remove it at the end unless I find a way to remove close pairs the way you do without affecting the correlators (which is not the case today). But perhaps in that case directly the magnetic field will be better

PWGCF/DataModel/DptDptFiltered.h
victor-gonzalez: Good point!
I will rename them

PWGCF/DataModel/DptDptFiltered.h
victor-gonzalez: So far, once filtered, the detector level and the generator level are completely independent and no need to link them

PWGCF/TableProducer/dptdptfilter.cxx
victor-gonzalez: You are right! It should not be!
I will change them

PWGCF/TableProducer/dptdptfilter.cxx
victor-gonzalez: Not yet for the filter task
It has been introduced in the match reco-gen and in the filter-qa

PWGCF/DataModel/DptDptFiltered.h
victor-gonzalez: `TrackSelection` table is still with `uint8_t`
Anyway, I will move it to `bool` and check if it works

jgrosseo: Yes, you are right. Anyhow it should work now. We should also then change the TrackSelection ones. Let me know if it works for you.

victor-gonzalez: It worked nice!

victor-gonzalez:
Strange. In both `build/O2Physics/o2` and `build/O2Physics/o2/macOS` there are only warnings, no errors
Has the warnings as errors been activated?

PWGCF/DataModel/DptDptFiltered.h
jgrosseo: Yes, another way to remove close pairs is better...

jgrosseo (approved):

PR #270, opened 2021-11-11
Fix compiler warnings in `TriggerAliases.h` and `EventSelection.h`

victor-gonzalez:
The `eventSelectionLabels` are not used so they were removed
If they were needed something similar to the `aliasLabels` would be needed


jgrosseo (changes requested):
Thanks @victor-gonzalez for proposing this. We are currently discussing how this could be solved for good for variables which may be needed in the header. I put this on hold for now.

victor-gonzalez:
OK
But have in mind that in the header they can be declared but not defined as in the proposed PR are now `aliasLabels`

PR #274, opened 2021-11-12
LF (strangeness) triggers implementation

ercolessi:
This code is co-authored by @ChiaraDeMartin95

Dear experts,
this is the first version of the LF strangeness trigger task. 
As you will see we save some QA histos which are crucial at this stage to make sure our selections are compatible with Run3 data. In the future if necessary we can decide to remove them.

Please let us know if you have comments or suggestions.


mpuccio (approved):
Hi @ercolessi, thanks a lot for the PR! I now approve for merging but in the next iteration please consider to replace `TMath::Power` with `std::pow`. In particular, when you do sum in quadratures (like in line 208) you can also use `std::hypot(px,py,pz)` that computes the total momentum.

PR #275, opened 2021-11-12
Trackextension: use http://alice-ccdb.cern.ch

sawenzel:
Trackextension is the only part of the code that uses https://alice-ccdb.cern.ch. A simple
move to `http:` should solve the problem with certificates discussed on mattermost. 

This would also follow
the approach in many other places of the code (see `git grep alice-ccdb`).

Maybe the CCDB location should be a global property for all tasks anyway.

### DIFF
--- a/Common/TableProducer/trackextension.cxx
+++ b/Common/TableProducer/trackextension.cxx
@@ -48,7 +48,7 @@ constexpr long run3grp_timestamp = (1619781650000 + 1619781529000) / 2;
 const char* ccdbpath_lut = "GLO/Param/MatLUT";
 const char* ccdbpath_geo = "GLO/Config/Geometry";
 const char* ccdbpath_grp = "GLO/GRP/GRP";
-const char* ccdburl = "https://alice-ccdb.cern.ch"; /* test  "http://alice-ccdb.cern.ch:8080"; */
+const char* ccdburl = "http://alice-ccdb.cern.ch"; /* test  "http://alice-ccdb.cern.ch:8080"; */
 } // namespace trackextension
 } // namespace analysis
 } // namespace o2


jgrosseo (approved):

Common/TableProducer/trackextension.cxx
```diff
@@ -48,7 +48,7 @@ constexpr long run3grp_timestamp = (1619781650000 + 1619781529000) / 2;
 const char* ccdbpath_lut = "GLO/Param/MatLUT";
 const char* ccdbpath_geo = "GLO/Config/Geometry";
 const char* ccdbpath_grp = "GLO/GRP/GRP";
-const char* ccdburl = "https://alice-ccdb.cern.ch"; /* test  "http://alice-ccdb.cern.ch:8080"; */
+const char* ccdburl = "http://alice-ccdb.cern.ch"; /* test  "http://alice-ccdb.cern.ch:8080"; */
```
sawenzel, line 51: maybe 8080 still needs to be added

jgrosseo:
> Trackextension is the only part of the code that uses https://alice-ccdb.cern.ch. A simple move to `http:` should solve the problem with certificates discussed on mattermost.
> 
> This would also follow the approach in many other places of the code (see `git grep alice-ccdb`).
> 
> Maybe the CCDB location should be a global property for all tasks anyway.

Thanks Sandro. Yes, the plan is to have this global with the CCDBConfigurable. Ongoing.

Common/TableProducer/trackextension.cxx
jgrosseo: I think 8080 points to the test instance...

victor-gonzalez: If you don't put `https` then you have to put `8080` which is the test endpoint

jgrosseo:
@sawenzel did you try this?

Common/TableProducer/trackextension.cxx
jgrosseo: I don't think so
http://alice-ccdb.cern.ch/browse/GLO/GRP/GRP --> production
http://alice-ccdb.cern.ch:8080/browse/GLO/GRP/GRP --> test

victor-gonzalez: Yeap!
You're right!
Sorry for the noise!

PR #278, opened 2021-11-12
[EMCAL-733] Port cell monitor task from EMCAL analysis framework

mfasDa:
Cell monitor task used to monitor cell-related observables (amplitude,
time, occupancy) at AOD level. Ported from AliEmcalCellMonitorTask
in AliPhysics.

Pending features to be added in the next steps:
- Trigger selection
- Retrieval of bad channel map from CCDB
- Separation of cells in cluster / isolated

### DIFF
--- a/PWGJE/Tasks/emccellmonitor.cxx
+++ b/PWGJE/Tasks/emccellmonitor.cxx
@@ -1,0 +1,244 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+#include <climits>
+#include <cstdlib>
+#include <map>
+#include <memory>
+#include <vector>
+
+#include "Framework/runDataProcessing.h"
+#include "Framework/AnalysisTask.h"
+#include "Framework/AnalysisDataModel.h"
+#include "Framework/ASoA.h"
+#include "Framework/HistogramRegistry.h"
+
+#include "EMCALBase/Geometry.h"
+#include "EMCALCalib/BadChannelMap.h"
+
+/// \struct CellMonitor
+/// \brief Simple monitoring task for cell related quantities
+/// \author Markus Fasel <markus.fasel@cern.ch>, Oak Ridge National Laoratory
+/// \since Nov 12, 2021
+///
+/// This task monitors simple cell quantities like
+/// - Amplitude distribution
+/// - Time distribution
+/// - Count rate in col-row space
+/// - Integrated amplitude in col-row space
+/// In addition, in case the task has access to reconstructed clusters,
+/// it plots
+/// - Frequency of a cell contributing to clusters
+/// - Summed cell amplitude fraction in a cluster
+///
+/// The task is a direct port of PWG::EMCAL::AliEmcalCellMonitorTask from
+/// AliPhysics
+struct CellMonitor {
+
+  o2::framework::Configurable<double> mMinCellAmplitude{"minCellAmplitude", 0., "Minimum cell amplitude for histograms."};
+  o2::framework::Configurable<double> mMinCellTimeMain{"minCellTimeMain", -50, "Min. cell time of main bunch selection"};
+  o2::framework::Configurable<double> mMaxCellTimeMain{"maxCellTimeMain", 100, "Max. cell time of main bunch selection"};
+
+  o2::framework::HistogramRegistry mHistManager{"CellMonitorHistograms"};
+
+  o2::emcal::Geometry* mGeometry = nullptr;
+  std::shared_ptr<o2::emcal::BadChannelMap> mBadChannels;
+
+  /// \brief Create output histograms and initialize geometry
+  void init(o2::framework::InitContext const&)
+  {
+    using o2HistType = o2::framework::HistType;
+    using o2Axis = o2::framework::AxisSpec;
+    LOG(INFO) << "Initializing EMCAL cell monitor task ...";
+    mGeometry = o2::emcal::Geometry::GetInstanceFromRunNumber(300000);
+    auto nCells = mGeometry->GetNCells();
+    int ntimeMain = int(mMaxCellTimeMain - mMaxCellTimeMain);
+
+    const o2Axis cellAxis{nCells, -0.5, nCells - 0.5, "cellID", "Cell abs. ID"},
+      amplitudeAxis{makeAmplitudeBinning(), "amplitude", "Amplitude (GeV)"},
+      amplitudeAxisLarge{1000, 0., 100., "amplitudeLarge", "Amplutude (GeV)"},
+      timeAxisLarge{1500, -600, 900, "celltime", "cell time (ns)"},
+      timeAxisMainBunch{ntimeMain, mMinCellTimeMain, mMaxCellTimeMain, "celltimeMain", "time (ns)"},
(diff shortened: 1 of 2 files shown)


mfasDa:
PWGJE for the moment placeholder library for components ported from the EMCAL analysis framework from AliPhysics until final location is decided (@jgrosseo).

iarsene:
@mfasDa, @jgrosseo   The task looks fine from my side. But as for the location, wouldn't this task be better suited in O2, not O2Physics ?

mfasDa:
@iarsene Thanks a lot for your review! This is the start of an EMCAL-related QA task. I would prefer to have it in O2
- The corresponding run2 task was in AliPhysics, and we might need more frequent development cycles then this would probably be for O2 in the future
- In regular running it will require a dependence on other tasks (event selection, emcal correction task) which are in O2Physics
-  Clusters will be added in the future which have have a dependency on the dynamic o2::aod::emcalcluster table which is part of O2Physics and not the general datamodel.
- I would avoid having analysis-related tasks in the workflow library of O2 if possible. Usually the task are rather heavy to compile and have no relevance for the online systems (FLP/EPN) - there we have dedicated QC tasks for. 

For the moment I would prefer to have it merged as we would like use it for tests of the AODs we produce with the corresponding tasks in development in O2 (https://github.com/AliceO2Group/AliceO2/pull/7521, stand-alone AOD producer in preparation) from the pilot beam - there EMCAL was in a stand-alone partition so the task must have a minimum of dependencies on AOD tables, as it was the case for the AliPhysics task that also was capable of analysing stand-alone runs.

iarsene:
Thanks for the explanations. I agree with your reasons, and for now i think its fine to merge it in PWGJE. But since i think there will be more groups which use EMCAL (DQ at least is one), it might be good to have these in a common directory at some point.

iarsene (dismissed):

PWGJE/Tasks/emccellmonitor.cxx
```diff
@@ -0,0 +1,215 @@
...
+    }
+    return result;
+  }
+};
+} // namespace o2::emcal
+
+o2::framework::WorkflowSpec defineDataProcessing(o2::framework::ConfigContext const& cfgc)
+{
+  return o2::framework::WorkflowSpec{
+    o2::framework::adaptAnalysisTask<o2::emcal::CellMonitor>(cfgc, o2::framework::TaskName{"emcal-cellmonitor-task"})};
```
jgrosseo, line 214: can you drop TaskName

jgrosseo:
> Thanks for the explanations. I agree with your reasons, and for now i think its fine to merge it in PWGJE. But since i think there will be more groups which use EMCAL (DQ at least is one), it might be good to have these in a common directory at some point.

Also the jet finder and HF finder are sitting in the corresponding PWG directories. I believe it is not a problem because with generating executables we don't run into dependency issues...

jgrosseo:
@mfasDa This is fine with me, see my minor comments in the PR

PWGJE/Tasks/emccellmonitor.cxx
mfasDa: Seems this is compatible only with tasks in the global namespace.

jgrosseo (commented):
my pending comment was not posted, sorry...

PWGJE/Tasks/emccellmonitor.cxx
jgrosseo: I see. Any need for the name space? The struct only exists in this executable and is never included anymore else.

PWGJE/Tasks/emccellmonitor.cxx
```diff
@@ -0,0 +1,215 @@
...
+/// The task is a direct port of PWG::EMCAL::AliEmcalCellMonitorTask from
+/// AliPhysics
+struct CellMonitor {
+  struct SupermoduleHists {
+    TH2* mCellAmp;
+    TH2* mCellCount;
+    TH2* mCellAmpTime;
+  };
+
+  o2::framework::Configurable<double> mMinCellAmplitude{"minCellAmplitude", 0., "Minimum cell amplitude for histograms."};
```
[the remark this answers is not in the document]
mfasDa: Adding using directives defeats the purpose of namespaces and should therefore be only used sparsely. If at all I would rather use using with aliasing, that should be quite safe. Not sure there is a probability for naming clashes with the framework namespace but better not risk it in the first place. 

For what concerns the HistogramRegistry I am definitely going to use that - this is my first task in O2Physics and I still need to get use to the new component. In fact the corresponding task in AliPhysics was making use of a histogram registry (of which there must have been hundreds of implementations in AliPhysics ...).

PWGJE/Tasks/emccellmonitor.cxx
jgrosseo: Yes, indeed. This is the case for the O2Physics tasks. Maybe you can remove the emcal namespace here, such that we can use the canonical task name definition. The other items I leave to you to decide.

mfasDa: Will do, future components of an analysis framework library should however go to the namespace o2::emcal (AOD content to o2::aod) for safety reasons.

PWGJE/Tasks/emccellmonitor.cxx
```diff
@@ -0,0 +1,201 @@
...
+    static constexpr std::string_view cellAmpTimeHist[20] = {"cellAmplitudeTime/cellAmpTimeCorrSM0", "cellAmplitudeTime/cellAmpTimeCorrSM1", "cellAmplitudeTime/cellAmpTimeCorrSM2", "cellAmplitudeTime/cellAmpTimeCorrSM3", "cellAmplitudeTime/cellAmpTimeCorrSM4", "cellAmplitudeTime/cellAmpTimeCorrSM5", "cellAmplitudeTime/cellAmpTimeCorrSM6", "cellAmplitudeTime/cellAmpTimeCorrSM7", "cellAmplitudeTime/cellAmpTimeCorrSM8", "cellAmplitudeTime/cellAmpTimeCorrSM9", "cellAmplitudeTime/cellAmpTimeCorrSM10", "cellAmplitudeTime/cellAmpTimeCorrSM11", "cellAmplitudeTime/cellAmpTimeCorrSM12", "cellAmplitudeTime/cellAmpTimeCorrSM13", "cellAmplitudeTime/cellAmpTimeCorrSM14", "cellAmplitudeTime/cellAmpTimeCorrSM15", "cellAmplitudeTime/cellAmpTimeCorrSM16", "cellAmplitudeTime/cellAmpTimeCorrSM17", "cellAmplitudeTime/cellAmpTimeCorrSM18", "cellAmplitudeTime/cellAmpTimeCorrSM19"};
+    mHistManager.fill(HIST(cellAmpHistSM[supermoduleID]), col, row);
+    mHistManager.fill(HIST(cellCountHistSM[supermoduleID]), col, row, amplitude);
+    mHistManager.fill(HIST(cellAmpTimeHist[supermoduleID]), celltime, amplitude); 
+  }
```
mfasDa, line 157: I have changed to the histogram registry but I am not fully happy with this part. The goal is that we have 3 histograms we want to monitor per supermodule ID. In this case the supermodule ID is determined dynamically from the tower ID and during the fill process becomes a dynamic part of the name. Based on my understanding of the histogram registry names must be constant. The only way around  I found so far is to have a templated helper function where the supermodule is a template parameter (and therefore constant) and internally 3 arrays with constant histogram names / supermodule. The mapping supermodule ID -> template parameter must therefore happen in a switch statement. This construction is a bit clumsy and inflexible, will still look for something better, maybe for the beginning it could go.

jgrosseo (approved):

PWGJE/Tasks/emccellmonitor.cxx
jgrosseo: Can the supermodule ID be used as an axis?

mfasDa: These are already 2D histograms, going to higher dimensions would be possible but not really preferred. I can also imagine at some point other detectors might have similar interests (i.e. monitor stuff per tracking sector).

PR #283, opened 2021-11-15
UPC forward task

aga01935:


PWGUD/Tasks/upcForward.cxx
```diff
@@ -27,35 +29,43 @@ using namespace std;
...
+    registry.add("hPz", "Pz;#it{P_{z}}, GeV/c;", kTH1D, {{500, -5., 5.}});
+    registry.add("hRap", "Rapidity of Mother;#it{y};", kTH1D, {{500, -10., 10.}});
+    registry.add("hEta", "Eta;#it{#eta};", kTH1D, {{500, -10., 10.}});
+    registry.add("hCharge", "Charge;#it{charge};", kTH1D, {{500, -10., 10.}});
+    registry.add("hPhi", "Phi;#it{#Phi};", kTH1D, {{500, -6., 6.}});
+
+    Configurable<float> etalow{"etalow", -4.0f, ""};   //
+    Configurable<float> etahigh{"etahigh", -2.5f, ""}; //
+    Filter etaFilter_a = aod::fwdtrack::eta > etalow;
+    Filter etaFilter_b = aod::fwdtrack::eta < etahigh;
```
iarsene, line 63: You should combine in a single Filter expression these two lines; it should be more efficient

PWGUD/Tasks/upcForward.cxx
```diff
@@ -113,18 +124,23 @@ struct UPCForward {
...
     registry.fill(HIST("hSelectionCounter"), 5);
 
-    if (-4 < p1.Eta() < -2.5 || -4 < p2.Eta() < -2.5) {
-
+    if ((p1.Eta() < -4) || (p2.Eta() < -4)) {
+      return;
+    }
+    if ((p1.Eta() > -2.5) || (p2.Eta() > -2.5)) {
       return;
     }
```
iarsene, line 139: Here you apply hardcoded selections on the muons, but you already have implemented Filters which cut on this based on Configurables. So this is not needed.

PWGUD/Tasks/upcForward.cxx
jgrosseo: Hi Ionut. I think they are combined automatically and it should have the same efficiency.
@aalkin can you comment?

iarsene: If there is such an optimization then its fine by me

aga01935: @iarsene @jgrosseo I already merged them into one.

PWGUD/Tasks/upcForward.cxx
aga01935: @iarsene I removed the selections.

PWGUD/Tasks/upcForward.cxx
aalkin: Yes, indeed, compatible filters will be joined with `and` automatically, but merging them also reduces the number of data members of the task so you can add something else if needs be.

iarsene (approved):

PR #298, opened 2021-11-17
aod-merger: add option to skip non-existing files

mario-krueger:
avoids crashes if the inputs.txt list for some reason contains files that dont exist or are not reachable

### DIFF
--- a/Common/Core/aodMerger.cxx
+++ b/Common/Core/aodMerger.cxx
@@ -28,13 +28,15 @@ int main(int argc, char* argv[])
   std::string inputCollection("input.txt");
   std::string outputFileName("AO2D.root");
   long maxDirSize = 100000000;
+  bool skipNonExistingFiles = false;
 
   int option_index = 0;
   static struct option long_options[] = {
     {"input", required_argument, nullptr, 0},
     {"output", required_argument, nullptr, 1},
     {"max-size", required_argument, nullptr, 2},
-    {"help", no_argument, nullptr, 3},
+    {"skip-non-existing-files", required_argument, nullptr, 3},
+    {"help", no_argument, nullptr, 4},
     {nullptr, 0, nullptr, 0}};
 
   while (true) {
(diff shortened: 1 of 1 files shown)


jgrosseo (changes requested):
Hi,
I am against this change. We can issue a fatal error in that case but we should not skip the file.
If the list cannot be processed an error has to be produced.

mario-krueger:
Hi,
currently it will break with a seg fault and TFile::Open will issue an error message which probably is enough to see what is the problem. I can close the PR again if skipping the file is not what we want. 
However, I was wondering if in production we would want one single corrupt file to break the whole merging chain. Maybe one could log the file names that did not work and print / save this somehow as a feedback to the operators instead?

jgrosseo:
On the Grid, we do staged merging for analysis, so indeed you can have a failure on a certain level but this only involves a few files. The risk with skipping it that you never find those errors.

However, I see the use case in the merger for local use. So we could add that in case one passes a parameter `--skip-not-existing` that it goes to the next file as you suggested. What do you think?

mario-krueger:
Hi,
sounds like a good compromise. 
For the case where we want to abort the merging, the program now ends gracefully without a seg fault and propagates that there was an error to the caller, so merging scripts can react accordingly.

Common/Core/aodMerger.cxx
```diff
@@ -48,10 +50,13 @@ int main(int argc, char* argv[])
...
+      skipNonExistingFiles = (strcmp(optarg, "true")) ? false : true;
+    } else if (c == 4) {
       printf("AOD merging tool. Options: \n");
-      printf("  --input <inputfile.txt>      Contains path to files to be merged. Default: %s\n", inputCollection.c_str());
-      printf("  --output <outputfile.root>   Target output ROOT file. Default: %s\n", outputFileName.c_str());
-      printf("  --max-size <size in Bytes>   Target directory size: %ld \n", maxDirSize);
+      printf("  --input <inputfile.txt>                  Contains path to files to be merged. Default: %s\n", inputCollection.c_str());
+      printf("  --output <outputfile.root>               Target output ROOT file. Default: %s\n", outputFileName.c_str());
+      printf("  --max-size <size in Bytes>               Target directory size. Default: %ld\n", maxDirSize);
+      printf("  --skip-non-existing-files <true|false>   Non-existing files are skipped. Default: %s\n", "false");
```
jgrosseo, line 59: I would not put an argument to this. The default is false and then one just provides `--skip-non-existing-files` if one wants this option. What do you think?

Common/Core/aodMerger.cxx
```diff
@@ -62,6 +67,7 @@ int main(int argc, char* argv[])
   printf("  Input file: %s\n", inputCollection.c_str());
   printf("  Ouput file name: %s\n", outputFileName.c_str());
   printf("  Maximal folder size (uncompressed): %ld\n", maxDirSize);
+  printf("  Skip non-existing files: %s\n", (skipNonExistingFiles) ? "true" : "false");
```
jgrosseo, line 70: Only print when true and put a WARNING in front :)

Common/Core/aodMerger.cxx
```diff
@@ -93,6 +99,16 @@ int main(int argc, char* argv[])
     printf("Processing input file: %s\n", line.Data());
 
     auto inputFile = TFile::Open(line);
+    if (!inputFile) {
+      printf("Could not open input file %s.\n", line.Data());
+      if (skipNonExistingFiles) {
+        continue;
+      } else {
+        printf("Aborting merge!\n");
```
jgrosseo, line 107: I think we should also close the output file and unlink it because it is only partial.

jgrosseo (commented):
Thanks! Here a few comments

Common/Core/aodMerger.cxx
mario-krueger: by breaking out of the while loop all buffered trees are now saved and the file is closed properly

jgrosseo (approved):

PR #301, opened 2021-11-18
Fix wrong type definition

victor-gonzalez:


alibuild (approved):

PR #307, opened 2021-11-19
small fix to extract the MC truth information before filling histograms

iarsene:


alibuild (approved):

PR #308, opened 2021-11-19
Fix for  wrong check for availability of PID info

victor-gonzalez:


alibuild (approved):

PR #318, opened 2021-11-22
Extend Efficiency task

njacazio:


mario-krueger (approved):
I suppose with this we need then also a change in the dependencies (and a config file?) here:
https://github.com/AliceO2Group/O2DPG/blob/master/MC/bin/o2dpg_sim_workflow.py#L845

njacazio:
Hi @mario-krueger this is correct, I will update it

njacazio:
For reference here it is https://github.com/AliceO2Group/O2DPG/pull/127

PR #321, opened 2021-11-22
Fix compiler warnings

iarsene:


alibuild (approved):

PR #331, opened 2021-11-23
Proper handling of GRP objects for Run3 runs

victor-gonzalez:


victor-gonzalez:
There is one thing which prevents the approval
We have to be sure that the nightly productions produce or refers to an appropriate GRP object
If not the analysis of the nightly productions will not be feasible if the `trackextended` task is required

victor-gonzalez:
Perhaps, for the time being, we can have an additional `processRun3Sim` keeping the current default `GRP` object for the nightly productions
@jgrosseo what do you think?

jgrosseo:
What is missing for the nightlies?
We have RCT and StartOrbit filled. So the only missing part would be the LUT and magnetic field? The existing object would need to be uploaded also for the validity range 1546730800-1547160770  and then it should work. @victor-gonzalez can you try this?

victor-gonzalez:
@jgrosseo I'm confused now
For the nightly productions, both alibi and grid, we use the GRP object which Ruben suggested a while ago
Its time stamp is hard coded in
```
constexpr long run3grp_timestamp = (1619781650000 + 1619781529000) / 2;
```
For the nightly productions I guess we get now a timestamp according with its run number (still being fixed in alibi productions)
So what you are suggesting, I guess, is to enlarge the timestamp validity of the default GRP object to cover the nightly productions

Is the range `1546730800-1547160770` which you mentioned the one being used for the nightly productions runs?
If that is the case, modifying the GRP object validity range with such value should be enough. No need of additional process nor hard coded timestamps

I will try that!

victor-gonzalez:
@jgrosseo OK, after my fault in uploading the GRP object with the new timestamps and the fix of Costin it seems everything works without the need of a new process function nor any hard coded timestamps

I tested with 
- the pilot beam data with one of the new runs incorporated by Costing and the its GRP object uploaded by Ruben 
- with the nightly production on run `301043` which I guess is from yesterday

The alibi nightly still needs a hack for the run number so putting anyone of the above will evidently work

I will get rid of the timestamp hard coding and we can move on and merge it

Common/TableProducer/trackextension.cxx
```diff
@@ -118,7 +113,7 @@ struct TrackExtension {
   }
   PROCESS_SWITCH(TrackExtension, processRun2, "Process Run2 track extension task", true);
 
-  void processRun3(aod::FullTracks const& tracks, aod::Collisions const&)
+  void processRun3(aod::FullTracks const& tracks, aod::Collisions const&, aod::BCsWithTimestamps const&)
```
jgrosseo, line 116: Is it enough to subscribe to Tracks instead of FullTracks?

Common/TableProducer/trackextension.cxx
```diff
@@ -131,10 +126,26 @@ struct TrackExtension {
     /* when a new run (Run3) is processed                           */
     o2::base::Propagator::MatCorrType matCorr = o2::base::Propagator::MatCorrType::USEMatCorrLUT;
 
+    int lastCollId = -1;
     for (auto& track : tracks) {
       std::array<float, 2> dca{1e10f, 1e10f};
       if (track.has_collision()) {
         if (track.trackType() == o2::aod::track::TrackTypeEnum::Track) {
+          if (lastCollId != track.collisionId()) {
```
jgrosseo, line 134: I don't think this optimization is needed. The one with the run number should be enough...

jgrosseo (commented):
Thanks, Victor. Two comments below.

I also wonder about your TODO items. I don't think we would compute these quantities here, but that we can discuss elsewhere.

Common/TableProducer/trackextension.cxx
victor-gonzalez: For Run3 yes! Tracks should be enough!

Common/TableProducer/trackextension.cxx
victor-gonzalez: OK!
I will rid of them in both Run2 and Run3

victor-gonzalez:
Regarding the TODOs, they are not mine they are legacy
But if they don't apply here I'll get rid of them

jgrosseo (approved):

PR #342, opened 2021-11-25
Change the name of the MCSignal::Print() function to avoid hiding the…

iarsene:
… virtual Print() function


alibuild (approved):

PR #348, opened 2021-11-25
new hm trigger based on transverse region

aortizve:


aortizve:
hello @aalkin, could you please help me to review the new analysis task which I implemented for the underlying event analysis ?

EventFiltering/PWGMM/multFilter.cxx
```diff
@@ -83,6 +87,26 @@ struct multFilter {
   Filter trackFilter = (nabs(aod::track::eta) < cfgTrackEtaCut) && (aod::track::isGlobalTrack == (uint8_t) true) && (aod::track::pt > cfgTrackLowPtCut);
 
   using TrackCandidates = soa::Filtered<soa::Join<aod::Tracks, aod::TrackSelection>>;
+  float computeDeltaPhi(float phia, float phib,
+                        float rangeMin = -M_PI / 2.0, float rangeMax = 3.0 * M_PI / 2.0)
+  {
+    float dphi = -999;
+    if (phia < 0)
```
aalkin, line 94: Please note that ALICE coding convention explicitly requires braces in `if () {} else {}` etc. Even if there would be a single line inside the braces.

PWGMM/Tasks/uecharged.cxx
```diff
@@ -0,0 +1,336 @@
...
+  {
+    // primaries w/o golden cut
+    globalTracks = getGlobalTrackSelection();
+    globalTracks.SetRequireGoldenChi2(false);
+    // all w/o golden cuts
+    globalTrackswoPrim = getGlobalTrackSelection();
+    globalTrackswoPrim.SetMaxDcaXYPtDep([](float pt) { return 3.f + pt; });
+    globalTrackswoPrim.SetRequireGoldenChi2(false);
+
+    std::vector<double> ptBinningL = {
```
aalkin, line 67: Note that binning can be also supplied with `ConfigurableAxis` (there are examples in the tutorials and some tasks already use it).

PWGMM/Tasks/uecharged.cxx
```diff
@@ -0,0 +1,336 @@
...
+    ue.add("hStat", "TotalEvents", HistType::kTH1F, {{1, 0.5, 1.5, " "}});
+    ue.add("hdNdeta", "dNdeta", HistType::kTH1F, {{50, -2.5, 2.5, " "}});
+    ue.add("vtxZEta", ";#eta;vtxZ", HistType::kTH2F, {{50, -2.5, 2.5, " "}, {60, -30, 30, " "}});
+    ue.add("phiEta", ";#eta;#varphi", HistType::kTH2F, {{50, -2.5, 2.5}, {200, 0., 2 * M_PI, " "}});
+    ue.add("hvtxZ", "vtxZ", HistType::kTH1F, {{40, -20.0, 20.0, " "}});
+
+    ue.add("hCounter", "Counter; sel; Nev", HistType::kTH1D, {{3, 0, 3, " "}});
+    ue.add("hPtLeadingRecPS", "rec pTleading after physics selection", HistType::kTH1D, {ptAxisL});
+
+    for (int i = 0; i < 3; ++i)
```
aalkin, line 94: The same coding convention as with `if`s applies here, braces are mandatory.

PWGMM/Tasks/uecharged.cxx
```diff
@@ -0,0 +1,336 @@
...
+    double sumptm_top[3];
+    for (int i = 0; i < 3; ++i) {
+      nchm_top[i] = 0;
+      sumptm_top[i] = 0;
+    }
+    //LOGF (INFO, "--------------------   FLAG 0!!!!!!!!!!!");
+    vector<Float_t> ptArray;
+    vector<Float_t> phiArray;
+    vector<int> indexArray;
+    ptArray.clear();
```
aalkin, line 196: No need to clear a vector that is just declared.

PWGMM/Tasks/uecharged.cxx
```diff
@@ -0,0 +1,336 @@
...
+
+    for (int i_reg = 0; i_reg < 3; ++i_reg)
+      ue_rec.push_back(1.0 * nchm_top[i_reg]);
+    for (int i_reg = 0; i_reg < 3; ++i_reg)
+      ue_rec.push_back(sumptm_top[i_reg]);
+
+    // add flags for Vtx, PS, ev sel
+
+    // for some reason loops in side donot work
+    ue.fill(HIST(pNumDenMeasuredPS[0]), flPt, ue_rec[0]);
```
aalkin, line 251: `HIST()` macro requires literal constant as an argument. To use loop here you may want to retrieve pointers to the histograms from the registry into a vector.

PWGMM/Tasks/uecharged.cxx
```diff
@@ -0,0 +1,336 @@
...
+#include <TRandom.h>
+#include <cmath>
+
+// TODO: implement flags for all, phys. sel., good vertex, ...
+
+using namespace o2;
+using namespace o2::framework;
+using namespace o2::framework::expressions;
+
+struct uecharged {
```
aalkin, line 36: To make sure that task would get a name `ue-charged` in the processing, you can use capitalization in the struct name: `UeCharged`

aalkin (commented):
Everything looks good, apart from the coding convention. Although we do not use the same checks as with O2, explicit braces make the code easier to read.

PWGMM/Tasks/uecharged.cxx
mario-krueger: Hi, 
sorry for jumping in here. Just as a small comment to this:
An alternative to working with the pointers as Anton suggested would be to use the `static_for` loop that is specifically dedicated to this purpose 
(see e.g. here: https://github.com/AliceO2Group/O2Physics/blob/master/Tutorials/src/histogramRegistry.cxx#L227-L236 
or 
https://github.com/AliceO2Group/AliceO2/blob/dev/Framework/Core/test/test_StaticFor.cxx
).
Then it would be something like
```
#include "Framework/StaticFor.h"

static_for<0, 2>([&](auto i) {
    constexpr int index = i.value;
   ue.fill(HIST(pNumDenMeasuredPS[index]), flPt, ue_rec[index]);  
   ue.fill(HIST(pNumDenData[index]), flPt, ue_rec[index]);
   ue.fill(HIST(pSumPtMeasuredPS[index]), flPt, ue_rec[index+3]);
   ue.fill(HIST(pSumPtData[index]), flPt, ue_rec[index+3]);
});
```
Cheers,
  Mario

aalkin: This is even better, as it avoids dealing with pointers. Thank you, Mario!

EventFiltering/PWGMM/multFilter.cxx
aortizve: Hello @aalkin, thanks for the suggestions. Please check the updated version.

PWGMM/Tasks/uecharged.cxx
aortizve: Thanks @mario-krueger, it works very well.

PWGMM/Tasks/uecharged.cxx
```diff
@@ -64,21 +65,11 @@ struct uecharged {
     globalTrackswoPrim.SetMaxDcaXYPtDep([](float pt) { return 3.f + pt; });
     globalTrackswoPrim.SetRequireGoldenChi2(false);
 
-    std::vector<double> ptBinningL = {
-      0.15, 0.50, 1.00, 1.50, 2.00, 2.50, 3.00, 3.50, 4.00, 4.50,
-      5.00, 6.00, 7.00, 8.00, 9.00, 10.0, 12.0, 14.0, 16.0, 18.0,
-      20.0, 25.0, 30.0, 40.0, 50.0};
-    AxisSpec ptAxisL = {ptBinningL, "#it{p}_{T}^{trig} (GeV/#it{c})"};
+    ConfigurableAxis ptBinningt{"ptBinningt", {0, 0.15, 0.50, 1.00, 1.50, 2.00, 2.50, 3.00, 3.50, 4.00, 4.50, 5.00, 6.00, 7.00, 8.00, 9.00, 10.0, 12.0, 14.0, 16.0, 18.0, 20.0, 25.0, 30.0, 40.0, 50.0}, "pTtrig bin limits"};
```
aalkin, line 68: Good, but it also requires the first entry in the list to be `VARIABLE_WIDTH` (which is 0), so that it can be correctly parsed by framework and hyperloop.

PWGMM/Tasks/uecharged.cxx
```diff
@@ -332,5 +311,5 @@ struct uecharged {
 WorkflowSpec defineDataProcessing(ConfigContext const& cfgc)
 {
   return WorkflowSpec{
-    adaptAnalysisTask<uecharged>(cfgc, TaskName{"ue-charged"})};
+    adaptAnalysisTask<uecharged>(cfgc, TaskName{"UeCharged"})};
```
aalkin, line 314: Rather than use `TaskName` argument you can rename the struct itself instead, then you do not need `TaskName` at all.

PWGMM/Tasks/uecharged.cxx
aortizve: Ok, but in this case the lowest edge of the first bin in 0.15. So, it should be ok. For ptAxis, I modified according to your suggestion. Thanks!

PWGMM/Tasks/uecharged.cxx
aortizve: Now I followed what was done in some tasks from HF

PWGMM/Tasks/uecharged.cxx
aalkin: Ok! It is just more readable with a semantic `VARIABLE_WIDTH`.

aalkin (approved):

mpuccio (approved):

PR #349, opened 2021-11-25
add TPC multiplicity to multiplicity table

jezwilkinson:
Add TPC multiplicity (estimated by # of tracks with NClsFindable >0) to mult table for run2 processing
Processing with the run3 option appears to not work (same size() method on the partition gives a single unphysical value to all rows in the dataframe), to be revisited (set to -1 in case of run3 processing for now)

### DIFF
--- a/Common/DataModel/Multiplicity.h
+++ b/Common/DataModel/Multiplicity.h
@@ -28,6 +28,7 @@ DECLARE_SOA_DYNAMIC_COLUMN(MultV0M, multV0M, //!
 DECLARE_SOA_DYNAMIC_COLUMN(MultT0M, multT0M, //!
                            [](float multT0A, float multT0C) -> float { return multT0A + multT0C; });
 DECLARE_SOA_COLUMN(MultTracklets, multTracklets, int);
+DECLARE_SOA_COLUMN(MultTPC, multTPC, int);
 
 } // namespace mult
 DECLARE_SOA_TABLE(Mults, "AOD", "MULT", //!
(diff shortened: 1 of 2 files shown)


jezwilkinson:
the run3 processor issue mentioned above appears to be that the partition as defined for run2 pulls in all of the tracks from the entire data frame and applies that number to every collision -- not only taking the tracks relating to each collision. Probably due to the fact that run2 processes with an iterator on collisions and run3 with a for loop.
I have managed to locally solve this issue by adding an internal partition within the collision loop:

```cxx
Partition<soa::Join<aod::Tracks, aod::TracksExtra>> tracksWithTPCrun3 = (aod::track::tpcNClsFindable > (uint8_t)0) && (aod::track::collisionId == collision.globalIndex());
      tracksWithTPCrun3.bindTable(tracksExtra);
```
but this winds up massively increasing the running time and virtual memory usage - should be resolved separately before that part is committed.

victor-gonzalez:
Thanks for doing this!
You are right! The filter on the Run 3 process creates a mess due to the process being subscribed to the whole collision table and the whole track table.
You can use the partition also inside the process, even in the collisions loop in the case of the process for Run 3.
You can have a look at the dataSelection tutorial
Also for the process in Run 2 you can remove the Tracks subscription because you are already subscribed via the join operation. You can filter on the join for the tracklets as well

jgrosseo (approved):

jezwilkinson:
Hi @victor-gonzalez,
Thanks for the feedback! I have made the change you suggested to the run2 processor so that it doesn't subscribe to Tracks twice.
For the run 3:  indeed the two lines I mentioned above were following the data selection tutorial for "PartitionInside" - the full version of this is at https://github.com/jezwilkinson/O2Physics/commit/f64aa66a0a90e47f80175f2bff7b5fbac35a3149 (which I do not add to this pull request).
The problem is that it causes the runtime to really explode, adding several minutes to just the multiplicity task on a single, small (27MB) pilot run AO2D:
```
without internal partition for run3: 
real    0m42.198s
user    0m9.836s
sys     0m4.796s
with internal partition for run3:
real    9m53.472s
user    9m26.048s
sys     0m7.972s
```
evidently we need to find a different way of defining this, it's far too much overhead for something so simple.

PR #350, opened 2021-11-26
Custom deltaBC, foundFV0 and bug fixes

ekryshen:
* custom deltaBC for collision-FT0 matching window, set to 300 bcs by default. Used as a workaround to increase FT0 matching efficiency for pilot beams
* foundFV0 index added in bcsel and evsel tables
* bux fixes in collision-FT0 matching algorithm

### DIFF
--- a/Common/DataModel/EventSelection.h
+++ b/Common/DataModel/EventSelection.h
@@ -72,36 +72,6 @@ enum EventSelectionFlags {
   kNsel
 };
 
-// In a struct to avoid warnings if that is not used in all executables which include this file
-struct EventSelectionLabels {
-  static constexpr const char* labels[kNsel] = {
-    "kIsBBV0A",
-    "kIsBBV0C",
-    "kIsBBFDA",
-    "kIsBBFDC",
-    "kNoBGV0A",
-    "kNoBGV0C",
-    "kNoBGFDA",
-    "kNoBGFDC",
-    "kIsBBT0A",
-    "kIsBBT0C",
-    "kIsBBZNA",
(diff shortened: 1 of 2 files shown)


Common/DataModel/EventSelection.h
```diff
@@ -123,20 +93,21 @@ DECLARE_SOA_COLUMN(NTracklets, nTracklets, int);        //! Tracklet multiplicit
 DECLARE_SOA_COLUMN(Sel7, sel7, bool);                   //! Event selection decision based on V0A & V0C
 DECLARE_SOA_COLUMN(Sel8, sel8, bool);                   //! Event selection decision based on TVX
 DECLARE_SOA_COLUMN(FoundFT0, foundFT0, int64_t);        //! FT0 entry index in FT0s table (-1 if doesn't exist)
+DECLARE_SOA_COLUMN(FoundFV0, foundFV0, int64_t);        //! FV0 entry index in FV0A table (-1 if doesn't exist)
```
jgrosseo, line 96: Is this a index to the other table as it says?
It should then be int only.
And why not declare it as index column?

jgrosseo:
Great, only one comment!

jgrosseo (approved):

Common/DataModel/EventSelection.h
ekryshen: Indeed,I changed both foundFT0 and foundFV0 types to int32_t.
I am not sure whether is should be defined as index column. I will discuss with Anton. I suggest to merge it as it is now so that people could start using evsel with pilot data.

PR #354, opened 2021-11-26
Add TPC mult for run3 processing

jezwilkinson:
put as draft for @victor-gonzalez to comment

With this patch the processRun3 function is changed to take an iterator via `soa::Join<Collisions, EvSels>::iterator`, instead of the `for (auto& collision : collisions)`. This way the same TPC partition works in both the run2 and run3 modes, without the extra overhead that would be introduced by an internal partition (as mentioned at https://github.com/AliceO2Group/O2Physics/pull/349#issuecomment-979825010)

Does anything speak against handling the iterator in this way for run3?


victor-gonzalez (approved):

victor-gonzalez:
To me it looks perfect!

jgrosseo (approved):

PR #362, opened 2021-11-29
Properly handle charge swap

ktf:


ktf:
@jgrosseo I think this was a genuine issue with the code, and it is not merely some "unused variable" nitpick.

jgrosseo:
@iarsene Can you check?

PR #363, opened 2021-11-29
[PWGMM] New MM task computing MFT tracks histograms

sarahherrmann:


PWGMM/Tasks/analyse-mft-tracks.cxx
```diff
@@ -0,0 +1,114 @@
...
+    }
+    icoll++;
+  }
+  //end of process
+
+  //aod::McCollisions
+};
+//end of MyTask
+
+struct analyseGenTracks {
```
aalkin, line 62: Any specific reason why this is a separate task and not a an extra process function in a single task?

PWGMM/Tasks/analyse-mft-tracks.cxx
```diff
@@ -0,0 +1,114 @@
...
+    }                                                                                                             //
+  };
+
+  void process(aod::McCollisions::iterator const& mcCollision, Particles const& particles)
+  {
+    int nChargedPrimaryParticles = 0;
+    auto z = mcCollision.posZ();
+
+    for (auto& particle : particles) {
+      auto p = TDatabasePDG::Instance()->GetParticle(particle.pdgCode());
```
aalkin, line 80: Please refer to how it is implemented in dNdeta task, you can declare `Service<TDatabasePDG> pdg;` inside the task. In fact, you do not need the code below, you can safely skip the particles that have no database entry (`p == nullptr`).

PWGMM/Tasks/analyse-mft-tracks.cxx
```diff
@@ -0,0 +1,114 @@
...
+          //          auto x = (std::trunc(particle.pdgCode() / 10000) - 100000);
+          //          charge = x - std::trunc(x / 1000) * 1000;
+          LOGF(debug, "[{}] Nucleus with PDG code {}", particle.globalIndex(), particle.pdgCode() /*, charge*/); // (charge %d)
+        } else {
+          LOGF(debug, "[{}] Unknown particle with PDG code {}", particle.globalIndex(), particle.pdgCode());
+        }
+      } else {
+        charge = p->Charge();
+      }
+      if (charge != 0 && particle.isPhysicalPrimary()) {
```
aalkin, line 94: You probably want to select particles in MFT acceptance?

aalkin (changes requested):

PWGMM/Tasks/analyse-mft-tracks.cxx
sarahherrmann: Not really, the only one is that when I tried to put both of the process functions inside a single task, it wouldn't compile, with the following error : 
[WARN] Task analyse-m-f-t-tracks has no inputs
[ERROR] error parsing options of o2-analysis-mm-analyse-mft-tracks: 
unrecognised option '--aod-file'

But it can also be handy when I am processing real data, I can just comment the line adaptAnalysisTask<analyseGenTracks>(cfgc)

PWGMM/Tasks/analyse-mft-tracks.cxx
sarahherrmann: If I do like you, I have the seg fault I told you about

PWGMM/Tasks/analyse-mft-tracks.cxx
aalkin: Each process function, that is not named `process()` needs to have an accompanying `PROCESS_SWITCH()` declaration. See, for example, `multiProcess.cxx` in tutorials.

PWGMM/Tasks/analyse-mft-tracks.cxx
aalkin: My code does not produce a segfault, so probably you are not using it correctly. Did you add the `Service<>` declaration in your task?

sarahherrmann: Well if I run your code with my latest AO2D I have a segfault. But my code runs fine if I don't put this line of code, so I would really prefer to keep it as is. And I added the Service<> declaration yes :)

PWGMM/Tasks/analyse-mft-tracks.cxx
sarahherrmann: Ok thanks, do you want me to write it all in one task then ?

aalkin: Having it in one task is indeed the preferred way. It also means you would not need to recompile the task to run on MC or data :)

PWGMM/Tasks/analyse-mft-tracks.cxx
aalkin: Still, please use the same code, then we can investigate the source of segfault. It could be caused by input data, or maybe you were testing during the short period where O2 dev was broken (last two days).

sarahherrmann: Ok I will do that then

PWGMM/Tasks/analyse-mft-tracks.cxx
sarahherrmann: Ok, will change that today when I have time

PWGMM/Tasks/CMakeLists.txt
```diff
@@ -1,30 +1,35 @@
-# Copyright 2019-2020 CERN and copyright holders of ALICE O2.
-# See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
-# All rights not expressly granted are reserved.
+#Copyright 2019 - 2020 CERN and copyright holders of ALICE O2.
```
aalkin, line 1: What are all these extra white spaces? CMakeLists.txt is not checked by clang-format, so you'll need to fix this yourself. Probably your text editor uses tabs, which is not a good idea for our codebase, that only uses normal spaces.

PWGMM/Tasks/analyse-mft-tracks.cxx
sarahherrmann: For the moment no, since the eta region of the MFT is changing with the z_vtx

PWGMM/Tasks/CMakeLists.txt
sarahherrmann: I don't know, I have these when I do "clang-format -style=file -i <file>", which I am obligated to do to have the right Clang format to make this pull request :( 
I don't have them in my editor (atom), they only appear when I run this command

aalkin: I see, please fix the whitespace and do not apply clang-format to CMakeListst.txt, otherwise this seems complete.

aalkin: There is still some extra whitespace in the license header - it is required to be exact. The rest is less important, but please make sure that for `CMakeLists.txt` only your additional workflow entries appear in diff.

sarahherrmann: Ok sorry I didn't notice

aalkin (approved):

jgrosseo:
@sarahherrmann Could you fix the warning?

aalkin (approved):

PR #373, opened 2021-12-01
Adjust analysis tasks for GroupSlicer separated from AnalysisTask.h

saganatt:
Follow-up to AliceO2Group/AliceO2#7781

### DIFF
--- a/PWGCF/FemtoDream/femtoDreamPairTaskTrackTrack.cxx
+++ b/PWGCF/FemtoDream/femtoDreamPairTaskTrackTrack.cxx
@@ -243,7 +243,7 @@ struct femtoDreamPairTaskTrackTrack {
   {
     cols.bindExternalIndices(&parts);
     auto particlesTuple = std::make_tuple(parts);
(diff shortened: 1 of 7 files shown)


victor-gonzalez (dismissed):

jgrosseo:
@saganatt This depends on which O2 PRs?

jgrosseo:
@saganatt We check O2Physics in the O2 CI. This means that both PR CIs will fail, right? And we possibly need to sync and force the merge.

saganatt:
> @saganatt We check O2Physics in the O2 CI. This means that both PR CIs will fail, right? And we possibly need to sync and force the merge.

Ah, right, both repos need to be updated at once... How to do it?

jgrosseo:
We first see if the O2 CI passes (except O2Physics), then I can force it (unless you see a trick to decouple it...)

saganatt:
ok, thanks!

iarsene:
@saganatt  I just merged a PR with changes in tableReader.cxx, which are now in conflict with this PR. It can be fixed easily though

saganatt:
@iarsene thanks for the info, I rebased

PR #379, opened 2021-12-02
WIP: LF triggers, make event selection configurable + fix in input tables

ercolessi:
@ChiaraDeMartin95 please have a look


ercolessi:
Hi @mpuccio, in this PR we remove temporarily the access to the v0 daughter information of the cascade since it breaks the workflow and we did not find a solution yet, but we would like to start running on trains.
We also add a multiprocess switch in order to run the task on both Run2 converted and Run3 sims smoothly.

mpuccio (approved):

mpuccio:
Thanks a lot @ercolessi, it looks good to me! We actually have the new cycle of the MC to be tested, Raluca already updated the runlist so as soon as this change is available on hyperloop you can have a look! Thanks again.

PR #382, opened 2021-12-02
Prevent undefined behaviour warning when converting from int to bool

ktf:


ktf:
this was apparently already fixed. However I think the  other fix is wrong, as it uses > 0 rather than != 0.

iarsene:
> this was apparently already fixed. However I think the other fix is wrong, as it uses > 0 rather than != 0.

I fixed a bit earlier. Somehow my previous PR was merged even if the macOS failed ...
Why is the > 0 wrong? The expression in which i do > 0 is always either 0 or positive.
Ionut

ktf:
Unless you are masking the most significant bit of a signed quantity, which a priori you do not know if it's the case or not.

PR #385, opened 2021-12-03
foundFT0, foundFV0 and foundBC as index columns

ekryshen:
* Added foundBC index to EvSel table
* Declared foundFT0, foundFV0 and foundBC as index columns
* Changes in user tasks:
    - Check for foundFT0==-1 via collision.has_foundFT0()
    - Access data directly via collision.foundFT0() instead of iterator_at(foundFT0).

### DIFF
--- a/Common/DataModel/EventSelection.h
+++ b/Common/DataModel/EventSelection.h
@@ -78,37 +78,37 @@ namespace evsel
 // TODO bool arrays are not supported? Storing in int32 for the moment
 DECLARE_SOA_COLUMN(Alias, alias, int32_t[kNaliases]);
 DECLARE_SOA_COLUMN(Selection, selection, int32_t[kNsel]);
-DECLARE_SOA_COLUMN(BBV0A, bbV0A, bool);                 //! Beam-beam time in V0A
-DECLARE_SOA_COLUMN(BBV0C, bbV0C, bool);                 //! Beam-beam time in V0C
-DECLARE_SOA_COLUMN(BGV0A, bgV0A, bool);                 //! Beam-gas time in V0A
-DECLARE_SOA_COLUMN(BGV0C, bgV0C, bool);                 //! Beam-gas time in V0C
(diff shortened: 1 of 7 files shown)


Common/DataModel/EventSelection.h
```diff
@@ -78,36 +78,37 @@ namespace evsel
...
+DECLARE_SOA_COLUMN(BGV0A, bgV0A, bool);                                     //! Beam-gas time in V0A
+DECLARE_SOA_COLUMN(BGV0C, bgV0C, bool);                                     //! Beam-gas time in V0C
+DECLARE_SOA_COLUMN(BBFDA, bbFDA, bool);                                     //! Beam-beam time in FDA
+DECLARE_SOA_COLUMN(BBFDC, bbFDC, bool);                                     //! Beam-beam time in FDC
+DECLARE_SOA_COLUMN(BGFDA, bgFDA, bool);                                     //! Beam-gas time in FDA
+DECLARE_SOA_COLUMN(BGFDC, bgFDC, bool);                                     //! Beam-gas time in FDC
+DECLARE_SOA_COLUMN(MultRingV0A, multRingV0A, float[5]);                     //! V0A multiplicity per ring (4 rings in run2, 5 rings in run3)
+DECLARE_SOA_COLUMN(MultRingV0C, multRingV0C, float[4]);                     //! V0C multiplicity per ring (4 rings in run2)
+DECLARE_SOA_COLUMN(SpdClusters, spdClusters, uint32_t);                     //! Number of SPD clusters in two layers
+DECLARE_SOA_COLUMN(NTracklets, nTracklets, int);                            //! Tracklet multiplicity
```
jgrosseo, line 92: This is not important for this PR, but aren't 16 bit enough here?

jgrosseo (approved):

Common/DataModel/EventSelection.h
ekryshen: Yes, good point. we have up to 6000 tracklets in Run2 Pb-Pb so 16 bit should be enough. I will adjust it in the next iteration

aalkin (approved):

PR #387, opened 2021-12-03
Fix PR auto-approval script

TimoWilken:
1. Parse the CODEOWNERS file properly -- `git check-ignore` uses slightly different syntax and precedence rules, so we can't use that.
2. if another codeowner has already approved the PR, then consider that OK for the relevant CODEOWNER lines.

   This means that if e.g. a PR authored by A changes files Fa and Fb, and A owns Fa, and B owns Fb, then auto-approval is still possible as long as B approves the PR.

Not properly tested yet, though I should be able to do that on Monday.

### DIFF
--- a/.github/workflows/codeowner-self-approval.yml
+++ b/.github/workflows/codeowner-self-approval.yml
@@ -1,99 +1,116 @@
 name: PR self-approval
 
 on:
-  pull_request_target:
+  pull_request:
     types:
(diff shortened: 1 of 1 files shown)


TimoWilken:
I've tested these changes and they seem to be working.

jgrosseo (approved):

PR #389, opened 2021-12-05
PWGHF: Small changes in PID condition to avoid PID rejection at highpT and histogram for  Fiducial acceptance correction

skundu692:
@vkucera Small changes in PID condition to avoid PID rejection at highpT and Add histogram to calculate a correction factor for Fiducial acceptance

### DIFF
--- a/PWGHF/Tasks/taskD0parametrizedPID.cxx
+++ b/PWGHF/Tasks/taskD0parametrizedPID.cxx
@@ -62,7 +62,7 @@ struct TaskD0parametrizedPIDMC {
   //Configurable<double> centralitySelectionMax{"centralitySelectionMax", 30000.0, "Higher boundary of centrality selection"};
 
   Filter filterSelectCandidates = (aod::hf_selcandidate_d0_parametrizedPID::isSelD0NoPID >= 1 || aod::hf_selcandidate_d0_parametrizedPID::isSelD0barNoPID >= 1);
-
+  using McParticlesHf = soa::Join<aod::McParticles, aod::HfCandProng2MCGen>;
   void process(soa::Filtered<soa::Join<aod::HfCandProng2, aod::HFSelD0CandidateparametrizedPID, aod::HfCandProng2MCRec>> const& candidates, soa::Join<aod::McParticles, aod::HfCandProng2MCGen> const& particlesMC, aod::BigTracksMC const& tracks)
   // void process(const o2::aod::Collision& collision, soa::Filtered<soa::Join<aod::HfCandProng2, aod::HFSelD0CandidateparametrizedPID, aod::HfCandProng2MCRec>> const& candidates, soa::Join<aod::McParticles, aod::HfCandProng2MCGen> const& particlesMC, aod::BigTracksMC const& tracks)
   {
@@ -118,8 +118,8 @@ struct TaskD0parametrizedPIDMC {
       //if (ncontributor<=centralitySelectionMin && ncontributor>centralitySelectionMax) {
       //  continue;
       //}
-      Float_t maxFiducialY = 0.8;
-      Float_t minFiducialY = -0.8;
+      float maxFiducialY = 0.8;
+      float minFiducialY = -0.8;
       if (std::abs(particle.flagMCMatchGen()) == 1 << DecayType::D0ToPiK) {
         if (std::abs(RecoDecay::Y(array{particle.px(), particle.py(), particle.pz()}, RecoDecay::getMassPDG(particle.pdgCode()))) > 4.0) {
           continue;
@@ -131,11 +131,13 @@ struct TaskD0parametrizedPIDMC {
           maxFiducialY = -0.2 / 15 * ptGen * ptGen + 1.9 / 15 * ptGen + 0.5;
           minFiducialY = 0.2 / 15 * ptGen * ptGen - 1.9 / 15 * ptGen - 0.5;
         }
-        float etaProngs[2];
+        /*float etaProngs[2];
         auto daught0 = particlesMC.iteratorAt(particle.daughter0Id());
         auto daught1 = particlesMC.iteratorAt(particle.daughter1Id());
         etaProngs[0] = std::abs(daught0.eta());
-        etaProngs[1] = std::abs(daught1.eta());
+        etaProngs[1] = std::abs(daught1.eta());*/
+        auto etaProng0 = std::abs(particle.daughter0_as<McParticlesHf>().eta());
+        auto etaProng1 = std::abs(particle.daughter1_as<McParticlesHf>().eta());
         if (std::abs(yGen) < 2.5) {
           registry.fill(HIST("hGenPtCMSYCutNoDaughterEtaCut"), ptGen);
         }
@@ -144,13 +146,15 @@ struct TaskD0parametrizedPIDMC {
           registry.fill(HIST("hGenPtFiducialYCutNoDaughterEtaCut"), ptGen);
         }
 
-        if (etaProngs[0] < 2.5 && etaProngs[1] < 2.5) {
+        // if (etaProngs[0] < 2.5 && etaProngs[1] < 2.5) {
+        if (etaProng0 < 2.5 && etaProng1 < 2.5) {
           if (std::abs(yGen) < 2.5) {
             registry.fill(HIST("hGenPtCMSYCutCMSDaughterEtaCut"), ptGen);
           }
           if (yGen > minFiducialY && yGen < maxFiducialY) {
             registry.fill(HIST("hGenPtFiducialYCutCMSDaughterEtaCut"), ptGen);
-            if (etaProngs[0] < 0.8 && etaProngs[1] < 0.8) {
+            // if (etaProngs[0] < 0.8 && etaProngs[1] < 0.8) {
+            if (etaProng0 < 0.8 && etaProng1 < 0.8) {
               registry.fill(HIST("hGenPtFiducialYCutALICE2DaughterEtaCut"), ptGen);
             }
           }
--- a/PWGHF/Tasks/taskD0parametrizedPID.cxx
+++ b/PWGHF/Tasks/taskD0parametrizedPID.cxx
@@ -41,7 +41,7 @@ void customize(std::vector<o2::framework::ConfigParamSpec>& workflowOptions)
 struct TaskD0parametrizedPIDMC {
   HistogramRegistry registry{
     "registry",
-    {{"hMassGen", "2-prong candidates (generated); #it{p}_{T}; #it{y}", {HistType::kTH2F, {{150, 0., 30.}, {40, 0, 4.0}}}},
+    {{"hMassGen", "2-prong candidates (generated); #it{p}_{T}; #it{y}", {HistType::kTH2F, {{150, 0., 30.}, {8, 0, 4.0}}}},
      {"hGenFixCMSYNoEta", "2-prong candidates (generated); #it{p}_{T}", {HistType::kTH1F, {{150, 0., 30.}}}},
      {"hGenPtDepYNoEta", "2-prong candidates (generated); #it{p}_{T}", {HistType::kTH1F, {{150, 0., 30.}}}},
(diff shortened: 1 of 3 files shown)


PWGHF/Tasks/taskD0parametrizedPID.cxx
```diff
@@ -113,13 +118,51 @@ struct TaskD0parametrizedPIDMC {
...
+            registry.fill(HIST("hGenFixCMSYCMSEta"), ptGen);
+          }
+          if (yGen > minFiducialY && yGen < maxFiducialY) {
+            registry.fill(HIST("hGenPtDepYCMSEta"), ptGen);
+          }
+          if (etaProngs[0] < 0.8 && etaProngs[1] < 0.8) {
+            if (yGen > minFiducialY && yGen < maxFiducialY) {
+              registry.fill(HIST("hGenPtDepYALICE2Eta"), ptGen);
+            }
+          }
```
fcolamar, line 164: L160-164: it's not fully clear to me why you nest this loop inside the one at L15. Of course it works equally fine, but would be more clear having it outside (otherwise, if you want to keep it nested, at this point you could just move it after L158, inside the already existing fidutial acceptance 'if', removing the repeated fiductial acceptance check).
Also, I see that don't you have the corresponding hGenFixYALICE2Eta histogram, it's by mistake or maybe just because you don't need it (i.e. in ALICE we always use the yFiducial cut)?

fcolamar (changes requested):
Hello @skundu692, I added a couple of comments related to the task modification, for the PID ones I'd let Vit comment

PWGHF/Tasks/taskD0parametrizedPID.cxx
```diff
@@ -41,7 +41,12 @@ void customize(std::vector<o2::framework::ConfigParamSpec>& workflowOptions)
 struct TaskD0parametrizedPIDMC {
   HistogramRegistry registry{
     "registry",
-    {{"hMassGen", "2-prong candidates (generated); #it{p}_{T}; #it{y}", {HistType::kTH2F, {{150, 0., 30.}, {8, 0, 4.0}}}},
+    {{"hMassGen", "2-prong candidates (generated); #it{p}_{T}; #it{y}", {HistType::kTH2F, {{150, 0., 30.}, {40, 0, 4.0}}}},
```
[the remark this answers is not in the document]
skundu692: This is just to check the rapidity distribution of MC generated D meson. Actually, it is not needed to calculate the efficiency and I can change it to make it consistent with MC reconstructed.

PWGHF/Tasks/taskD0parametrizedPID.cxx
```diff
@@ -113,13 +118,51 @@ struct TaskD0parametrizedPIDMC {
       //if (ncontributor<=centralitySelectionMin && ncontributor>centralitySelectionMax) {
       //  continue;
       //}
+      Double_t maxFiducialY = 100.0;
+      Double_t minFiducialY = -100.0;
```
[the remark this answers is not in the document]
skundu692: I used it only for the initialization. I agree with your suggestion that we can avoid if condition in lines 134-137 by just initializing the  Double_t maxFiducialY = 0.8;      Double_t minFiducialY = -0.8;

PWGHF/Tasks/taskD0parametrizedPID.cxx
```diff
@@ -113,13 +118,51 @@ struct TaskD0parametrizedPIDMC {
...
         auto yGen = RecoDecay::Y(array{particle.px(), particle.py(), particle.pz()}, RecoDecay::getMassPDG(particle.pdgCode()));
         registry.fill(HIST("hMassGen"), ptGen, std::abs(yGen));
+        if (ptGen < 5.0) {
+          maxFiducialY = -0.2 / 15 * ptGen * ptGen + 1.9 / 15 * ptGen + 0.5;
+          minFiducialY = 0.2 / 15 * ptGen * ptGen - 1.9 / 15 * ptGen - 0.5;
+        }
+        if (ptGen > 5.0) {
+          maxFiducialY = 0.8;
+          minFiducialY = -0.8;
+        }
```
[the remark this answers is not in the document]
skundu692: I agree

PWGHF/Tasks/taskD0parametrizedPID.cxx
skundu692: I put the L160-164 inside the if (etaProngs[0] < 2.5 && etaProngs[1] < 2.5) {} to avoid the checking for tracks with |eta| > 2.5. Yes, I can move the condition inside the if condition in line 157. I do not need the hGenFixYALICE2Eta histogram for the calculation. I think for the fiducial correction factor in ALICE we generally use 2*yFiducial cut with the assumption that the rapidity distribution is flat. Correct me if I am wrong. But I added these histograms are to calculate the correction factor for going from ALICE 1 to CMS-like acceptance eta < 2.5, as suggested by Marco:
yield_(y_D < x, eta_daughter<2.5)/yield_(y_D < x, eta_daughter < 0.8) * yield_(Y_D < 2.5, eta daughter < 2.5)/yield(Y_D < 2.5, eta daughter anywhere)

@fcolamar are we missing any information for the correction?

fcolamar: Hi @skundu692, I didn't have the time to process in detail Marco's correction, but if that's the one to be used, it would seem that you added everything. So for what concerns the taskD0 updates, what you have now is fine to me.

PWGHF/Tasks/taskD0parametrizedPID.cxx
```diff
@@ -42,6 +42,11 @@ struct TaskD0parametrizedPIDMC {
   HistogramRegistry registry{
     "registry",
     {{"hMassGen", "2-prong candidates (generated); #it{p}_{T}; #it{y}", {HistType::kTH2F, {{150, 0., 30.}, {8, 0, 4.0}}}},
+     {"hGenFixCMSYNoEta", "2-prong candidates (generated); #it{p}_{T}", {HistType::kTH1F, {{150, 0., 30.}}}},
+     {"hGenPtDepYNoEta", "2-prong candidates (generated); #it{p}_{T}", {HistType::kTH1F, {{150, 0., 30.}}}},
+     {"hGenFixCMSYCMSEta", "2-prong candidates (generated); #it{p}_{T}", {HistType::kTH1F, {{150, 0., 30.}}}},
+     {"hGenPtDepYALICE2Eta", "2-prong candidates (generated); #it{p}_{T}", {HistType::kTH1F, {{150, 0., 30.}}}},
+     {"hGenPtDepYCMSEta", "2-prong candidates (generated); #it{p}_{T}", {HistType::kTH1F, {{150, 0., 30.}}}},
```
vkucera, line 49: The names of histograms are not clear enough to indicate what the are for. Some are even missing Pt.

PWGHF/Tasks/taskD0parametrizedPID.cxx
```diff
@@ -113,13 +118,44 @@ struct TaskD0parametrizedPIDMC {
...
         registry.fill(HIST("hMassGen"), ptGen, std::abs(yGen));
+        if (ptGen < 5.0) {
+          maxFiducialY = -0.2 / 15 * ptGen * ptGen + 1.9 / 15 * ptGen + 0.5;
+          minFiducialY = 0.2 / 15 * ptGen * ptGen - 1.9 / 15 * ptGen - 0.5;
+        }
+        float etaProngs[2];
+        for (int iD = particle.daughter0Id(), counter = 0; iD <= particle.daughter1Id(); ++iD, counter++) {
+          auto daught = particlesMC.iteratorAt(iD);
+          etaProngs[counter] = std::abs(daught.eta());
+        }
```
vkucera, line 138: Why do you need an array and a for loop for getting 2 eta values?

vkucera (changes requested):

PWGHF/Tasks/taskD0parametrizedPID.cxx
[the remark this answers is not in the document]
vkucera: Why is a pT vs y histogram called `hMassGen`?

PWGHF/Tasks/taskD0parametrizedPID.cxx
```diff
@@ -113,13 +118,44 @@ struct TaskD0parametrizedPIDMC {
       //if (ncontributor<=centralitySelectionMin && ncontributor>centralitySelectionMax) {
       //  continue;
       //}
+      Double_t maxFiducialY = 0.8;
+      Double_t minFiducialY = -0.8;
```
vkucera, line 122: Why do you not use `float`?

PWGHF/Tasks/taskD0parametrizedPID.cxx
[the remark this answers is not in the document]
skundu692: I have now modified the name

PWGHF/Tasks/taskD0parametrizedPID.cxx
skundu692: I have fixed it and now avoid for loop

PWGHF/Tasks/taskD0parametrizedPID.cxx
skundu692: Histogram names are fixed now

PWGHF/Tasks/taskD0parametrizedPID.cxx
vkucera: Do you have a reason for using `Float_t` (which is a ROOT type) instead of `float` (which is a native C++ type)?

PWGHF/Tasks/taskD0parametrizedPID.cxx
```diff
@@ -113,13 +118,43 @@ struct TaskD0parametrizedPIDMC {
...
+        registry.fill(HIST("hGenPtVsY"), ptGen, std::abs(yGen));
+        if (ptGen < 5.0) {
+          maxFiducialY = -0.2 / 15 * ptGen * ptGen + 1.9 / 15 * ptGen + 0.5;
+          minFiducialY = 0.2 / 15 * ptGen * ptGen - 1.9 / 15 * ptGen - 0.5;
+        }
+        float etaProngs[2];
+        auto daught0 = particlesMC.iteratorAt(particle.daughter0Id());
+        auto daught1 = particlesMC.iteratorAt(particle.daughter1Id());
+        etaProngs[0] = std::abs(daught0.eta());
+        etaProngs[1] = std::abs(daught1.eta());
```
vkucera, line 138: Why don't you just do this?
```
        auto etaProng0 = std::abs(particle.daughter0().eta()));
```
I am not sure if you would need `daughter_as` here though.

PWGHF/Tasks/taskD0parametrizedPID.cxx
```diff
@@ -113,13 +118,47 @@ struct TaskD0parametrizedPIDMC {
...
+        registry.fill(HIST("hGenPtVsY"), ptGen, std::abs(yGen));
+        if (ptGen < 5.0) {
+          maxFiducialY = -0.2 / 15 * ptGen * ptGen + 1.9 / 15 * ptGen + 0.5;
+          minFiducialY = 0.2 / 15 * ptGen * ptGen - 1.9 / 15 * ptGen - 0.5;
+        }
+        /*float etaProngs[2];
+        auto daught0 = particlesMC.iteratorAt(particle.daughter0Id());
+        auto daught1 = particlesMC.iteratorAt(particle.daughter1Id());
+        etaProngs[0] = std::abs(daught0.eta());
+        etaProngs[1] = std::abs(daught1.eta());*/
```
vkucera, line 138: Are these lines (and related ones below) still needed?

skundu692: Commented lines are removved

vkucera: Thanks

vkucera (approved):

vkucera (approved):

fcolamar (approved):

PR #393, opened 2021-12-06
PWGHF: Implementing sel8 trigger selection and Run3-devoted track quality selection

fcolamar:


PWGHF/TableProducer/HFTrackIndexSkimsCreator.cxx
```diff
@@ -421,15 +423,25 @@ struct HfTagSelTracks {
 
       iDebugCut = 4;
       if (doCutQuality.value && (debug || statusProng > 0)) { // FIXME to make a more complete selection e.g track.flags() & o2::aod::track::TPCrefit && track.flags() & o2::aod::track::GoldenChi2 &&
-        UChar_t clustermap = track.itsClusterMap();
-        if (!(track.tpcNClsFound() >= tpcNClsFound.value && // is this the number of TPC clusters? It should not be used
-              track.flags() & o2::aod::track::ITSrefit &&
-              (TESTBIT(clustermap, 0) || TESTBIT(clustermap, 1)))) {
+        bool hasGoodQuality = true;
+        if (useIsGlobalTrack) {
+          if (track.isGlobalTrack() != (uint8_t) true) {
```
fgrosa, line 428: Why in case of global tracks you don't require a hit in one of the two innermost layers of the ITS? Shouldn't we maybe disentangle the two selections to be able to apply either one of the two or both? Thanks!

fcolamar: Ciao Fabrizio,
according to what presented by Mario in his AW talk, the isGlobalTrack requirement should already contain a hit, actually I also checked and there's indeed a call selectedTracks.SetRequireHitsInITSLayers(1, {0, 1});
Anyway, in general at some point we could find out that the isGlobalTrack selection could be not optimal for HF: in that case I think the best solution would be to have a dedicated status flag, as isGlobalTrack but checking a generic-HF selection, I discussed this with Mattia and Mario, and it would be possible to have it.

fgrosa: Ciao @fcolamar, I missed that the requirement of a hit in the first two layers was already included in `isGlobalTrack`, thanks for the explanation!

PWGHF/TableProducer/HFTrackIndexSkimsCreator.cxx
```diff
@@ -421,15 +423,25 @@ struct HfTagSelTracks {
...
+            hasGoodQuality = false;
+          }
+        } else {
+          UChar_t clustermap = track.itsClusterMap();
+          if (!(track.tpcNClsFound() >= tpcNClsFound.value && // is this the number of TPC clusters? It should not be used
+                track.flags() & o2::aod::track::ITSrefit &&
+                (TESTBIT(clustermap, 0) || TESTBIT(clustermap, 1)))) {
+            hasGoodQuality = 0;
```
vkucera, line 436:
```suggestion
            hasGoodQuality = false;
```

PWGHF/TableProducer/HFTrackIndexSkimsCreator.cxx
vkucera, line 428: > `!= (uint8_t) true`

Why is this weird type conversion needed?
Does `isGlobalTrack` not return `bool`?

PWGHF/TableProducer/HFTrackIndexSkimsCreator.cxx
```diff
@@ -336,14 +338,14 @@ struct HfTagSelTracks {
...
-      //  }
-      //}
+      // bool cutStatus[CandidateType::NCandidateTypes][nCuts];
+      // if (debug) {
+      //   for (int iCandType = 0; iCandType < CandidateType::NCandidateTypes; iCandType++) {
+      //     for (int iCut = 0; iCut < nCuts; iCut++) {
+      //       cutStatus[iCandType][iCut] = true;
+      //     }
+      //   }
+      // }
```
vkucera, line 348: Where are all these formatting changes coming from? I don't think that clang has a problem with the current version.

vkucera (changes requested):

PWGHF/TableProducer/HFTrackIndexSkimsCreator.cxx
fcolamar: The column type in Common/DataModel/TrackSelectionTables.h is actually uint8_t, I think that removing the explicit conversion would still work fine, but I left that conversion for full consistency (it's also how the check is done in the qa task). nyway, if you prefer to remove the explicit casting, I can do it

PWGHF/TableProducer/HFTrackIndexSkimsCreator.cxx
fcolamar: Actually I don't know: I also thought that the previous version was already fine, but at the end those changes are automatically suggested and implemented when running clang as suggested in the coding conventions, so I left them

aalkin:
I have a very general comment: wouldn't it be better to disentangle the switches? One can have flags to enable/disable trigger and track selection and an independent switch to chose between runs 2 and 3. And based on the latter the code would use sel7 or sel8 and different track selections.

fcolamar:
Hi @aalkin, so would you propose to add processRun2 and processRun3 configurables to HfTagSelCollisions and HfTagSelTracks (but then what about Run5)?
And in that case would you still have the new useIsGlobalTrack flag, or keep just the old doCutQuality flag, which based on the processRun2/processRun3 values checks IsGlobalTrack (for run3) or the standard quality selection (for run2)? 
While for the trigger I'd anyway have new configurables useSel8Trigger and useSel7Trigger, this still to be added w.r.t. current PR version.

fgrosa:
@fcolamar @aalkin @vkucera is there any progress on this PR? It would be very important for the studies of the pp triggers in order to have the proper track selections for Run3 (which is included in this PR). Thanks a lot!

fcolamar:
Ciao @fgrosa, indeed we were discussing whether to have a someway wider restructure of the track selection and collision selection tasks, based on the general comments by Anton. Though I agree with you that it would be useful to have quickly a working code for Run3 studies committed, as the current one, that can than be improved along the lines of what written above with more calm.

ginnocen:
@fgrosa, thanks for the ping. I also pinged Anton privately.

ginnocen:
@fgrosa if it doesn't take too much time, could you split the PR where you have the clang fixes from the one in which there is the actual code change? thanks!

fcolamar:
Hi @ginnocen, I can do it, but how to practically act? Do I leave here only the real mode modifications and create a new one with only the clang fixes (supposing that the automatic clang checks don't fail on this one?)

vkucera:
> Hi @ginnocen, I can do it, but how to practically act? Do I leave here only the effective modifications and create a new one with only the clang fixes (supposing that the automatic clang checks don't fail on this one?)

Hi @fcolamar , I agree with @ginnocen . As a start, can you remove the formatting changes and see if the PR passes the tests without them? (The log from the first commit does not complain about them.)

fcolamar:
Hi @ginnocen, @vkucera, just done, it seems that clang tests are passed with this version, that contains only the 'real' code changes

ginnocen:
@fcolamar @vkucera IMO this is good to go. Can you confirm?

ginnocen (approved):

vkucera (approved):

PR #399, opened 2021-12-08
Make ONNXRuntime optional

ktf:


ktf:
@saganatt why does it need to be required?

saganatt:
We can have this optional for ML-related activites, if you prefer so.
I am not very fluent in CMake and I simply copied the command from other packages that were at the time.

jgrosseo (approved):

PR #402, opened 2021-12-08
Workaround crash in track-extension

ktf:
Without this we get a segfault on exit (probably due to a  double delete) on the grid. Moreover, when compiling with `-fsanitize=address` it complains about:

```
[47474:track-extension]: ==47474==ERROR: AddressSanitizer: SEGV on unknown address 0x7f893300006d (pc 0x00000041ef1c bp 0x60b000236000 sp 0x7fffaf0c9f48 T0)
[47474:track-extension]: ==47474==The signal is caused by a READ memory access.
[47474:track-extension]:     #0 0x41ef1c in std::_Sp_counted_ptr<TGeoManager*, (__gnu_cxx::_Lock_policy)2>::_M_dispose() /home/alitrain/eulisse/sw2/ubuntu1804_x86-64/GCC-Toolchain/v10.2.0-alice2-1/include/c++/10.2.0/bits/shared_ptr_base.h:380
```

With this change in, address sanitizer does not complain.

### DIFF
--- a/Common/TableProducer/trackextension.cxx
+++ b/Common/TableProducer/trackextension.cxx
@@ -71,7 +71,7 @@ struct TrackExtension {
     lut = o2::base::MatLayerCylSet::rectifyPtrFromFile(ccdb->get<o2::base::MatLayerCylSet>(ccdbpath_lut));
 
     if (!o2::base::GeometryManager::isGeometryLoaded()) {
-      ccdb->get<TGeoManager>(ccdbpath_geo);
+    //  ccdb->get<TGeoManager>(ccdbpath_geo);
     }
     mRunNumber = 0;
     mMagField = 0.0;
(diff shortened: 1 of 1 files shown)


ktf:
@shahor02 @sawenzel any idea what could be the issue?

ktf:
cc @jgrosseo

shahor02:
It does not complain because the geometry is not loaded. Will check what is the problem with loading, the TGeoManager is quite a beast...

victor-gonzalez:
Could it be caused by the `BasicCCDBManager` being created within the task scope?
When the task gets out of scope the CCDB manager has to be destroyed and the double delete could happen
Perhaps having it at the global scope will fix it?

shahor02:
Hi @victor-gonzalez 
no, the problem is in the caching of gGeoManager by the BasicCCDBManager, apparently on exit the root garbage collector deletes the gGeoManager before the destructor of the BasicCCDBManager (and its cache) is called. Will see what can be done.

jgrosseo (approved):

victor-gonzalez:
@shahor02 I see!
Thanks!
Will the propagation work fine without loading the geometry?

shahor02:
> Will the propagation work fine without loading the geometry?

Depends: with yet unmerged https://github.com/AliceO2Group/AliceO2/pull/7810 one can instantiate the Propagator w/o creating geometry, but the propagation will crash if called with USEMatCorrTGeo

shahor02:
In any case, I would not merge this PR, I am trying to find a solution. As a workaround, for the moment you can set `ccdb.setCach(false)` when setting up the manager.

PR #403, opened 2021-12-08
Event filtering: Add more track selections

fgrosa:


fgrosa:
@TimoWilken sorry to bother you again, but unfortunately the self approval still does not work. Could you please have a look when you have time? Thanks a lot!

PR #417, opened 2021-12-13
Templated getTrackPar, getTrackParCov

ddobrigk:


ddobrigk:
In addition to the reviewers already tagged, this is very relevant for @vkucera : Vit, please also take a look! Thanks!

vkucera:
> In addition to the reviewers already tagged, this is very relevant for @vkucera : Vit, please also take a look! Thanks!

Hi @ddobrigk , if the new return type can be digested in `DCAFitterN::process` in the same way as `TrackParCov`, I have no problem with this.

ddobrigk:
> > In addition to the reviewers already tagged, this is very relevant for @vkucera : Vit, please also take a look! Thanks!
> 
> Hi @ddobrigk , if the new return type can be digested in `DCAFitterN::process` in the same way as `TrackParCov`, I have no problem with this.

Hi @vkucera, yes, this is fully backwards compatible as the default behaviour is to use float as the usual TrackPrecision. I tested it and it seemed to be fine... and won't break anything (fingers crossed!) :-D

iarsene (approved):

PR #420, opened 2021-12-14
Event mixing: Examples for new interface

saganatt:
2 simple examples with declaration in struct and inside `process()`.

Note: This does not work yet with Filtered and Partitioned tables, work ongoing.

### DIFF
--- a/Tutorials/src/eventMixing.cxx
+++ b/Tutorials/src/eventMixing.cxx
@@ -68,49 +68,6 @@ struct HashTask {
   }
 };
 
-// Version 1: Using categorised combinations
-struct MixedEventsTracks {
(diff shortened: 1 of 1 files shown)


iarsene (approved):

iarsene:
Hi Maja, 
The new implementation looks nice and easier for the user. I approved, but i see that there are some messages from aliBuild, although no check failed. So maybe @jgrosseo or @ktf can have a look into this.

saganatt:
Thanks! I wonder if aliBuild build on older O2 (before merge of my mixing codes in framework).

jgrosseo (approved):

jgrosseo:
Great work! Removing the hash task is next then?

saganatt:
First I will check today what is wrong with Filtered and Partitioned, maybe I will find quickly what is the problem.
But in parallel, yes, I already started removing the hash task.

PR #425, opened 2021-12-15
Skip tracks w/o MC particle

benedikt-voelkel:
* explcit check introduced in O2 PR #7866 makes certain analyses fail in
  case an invalid index is found for an MCParticle associated to a track

* before, memory was randomly accessed in that case

* to be applied in affected analyses

### DIFF
--- a/Tutorials/src/mcHistograms.cxx
+++ b/Tutorials/src/mcHistograms.cxx
@@ -72,6 +72,10 @@ struct AccessMcTruth {
       //  continue;
       //if (track.labelMask() != 0)
       //  continue;
+      if (!track.has_mcParticle()) {
+        LOGF(warning, "No MC particle for track, skip...");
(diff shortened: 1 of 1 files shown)


benedikt-voelkel:
@aalkin For me locally that works now flawlessly, including a warning each time no associated MC particle was found for a track. I think, at least in a tutorial we can be a bit more verbose.

Other analyses have to be identified. I will notify the O2Analysis channel

jgrosseo (approved):
Happy to merge, but how a track cannot have a mc particle?

benedikt-voelkel:
I talked to @aalkin that should be fake tracks. But we can x-check with @shahor02 ?!

shahor02:
sometimes ITS manages to reconstruct a track made of noise hits (injected at digitization stage)

PR #435, opened 2022-01-10
Replace deprecated isPhysicalPrimary

njacazio:


njacazio:
@jgrosseo I would wait for the green light from green light from @akalweit and @ddobrigk before this can be merged but there should be no other residuals

jgrosseo:
OK, the only question would be if the ALICE3 MCs already contain this flag...

jgrosseo (approved):

njacazio:
> OK, the only question would be if the ALICE3 MCs already contain this flag...
probably not, I will have to include it

ginnocen (approved):

jgrosseo:
@ddobrigk any comments?

alibuild (approved):

ddobrigk:
Hi everyone, sorry for the late reply as I was traveling in the past 24 hours - this all looks fine to me, thanks for taking care!

PR #437, opened 2022-01-11
Check that track has collision before fetching ev. time

njacazio:


njacazio:
@iarsene here it should be fixed

iarsene:
> @iarsene here it should be fixed

Thanks Nicolo. Did you tested yourself or you want to wait with merging until Giacomo runs a test?

njacazio:
It should be fine, but the best should be if Giacomo tested it as well

jgrosseo (approved):

PR #443, opened 2022-01-13
fix merger for VLAs (workaround ROOT problem)

jgrosseo:


jgrosseo:
macCI unrelated. Merging.

PR #448, opened 2022-01-14
DPG: add PV refit also to processMC

mfaggin:


mfaggin:
Hi @mario-krueger , same thing ad yesterday, but redone also for the processMC function.

mario-krueger (approved):

PR #452, opened 2022-01-17
PID ML: Add an interface and a simple analysis task that uses ONNX model for inference

saganatt:
Hi everyone,

this is a sample, dummy analysis that uses an ONNX model (local or on CCDB, depending on the configurable paths) to get PID predictions.
The code still will be improved, as you can read in TODO comments. However, now we need to make a quick Hyperloop test for our student @Lukaszz99 thesis, so I would like to merge it ASAP.

@jgrosseo are the paths given in simpleApplyPidOnnxModel.cxx enough to provide access to files on Hyperloop? If you have also any other comments on code, I will be happy to receive them.
@fcatalan92 FYI - for now pidONNXModel is specific to our project, but most of it could be moved to a universal parent class. This is something we can consider later in the future.


Tools/PIDML/simpleApplyPidOnnxModel.cxx
```diff
@@ -0,0 +1,81 @@
...
+namespace mlpidresult
+{
+DECLARE_SOA_INDEX_COLUMN(Track, track); //! Track index
+DECLARE_SOA_COLUMN(MlPid, mlPid, int);  //! PID predicted by the model
+} // namespace mlpidresult
+DECLARE_SOA_TABLE(MlPidResults, "AOD", "MLPIDRESULTS", o2::soa::Index<>, mlpidresult::TrackId, mlpidresult::MlPid);
+} // namespace o2::aod
+
+struct SimpleApplyOnnxModelTask {
+  PidONNXModel pidModel; // One instance per model, e.g., one per each pid to predict
```
saganatt, line 40: Otherwise we would need to load each model inside process(), and this will be slower.

alibuild (approved):

jgrosseo:
@saganatt I think you need to download the file using the CCDB framework and not a direct http open. See efficiencyGlobal.cxx in tutorials for an example...

PR #464, opened 2022-01-20
Add KFParticle to dependencies

jezwilkinson:
This commit adds the necessary CMake find_package commands to find the KFParticle package in O2Physics so that it can be used in analysis tasks.

Currently KFParticle is defined in alidist as a dependency of O2, so is already built as part of the software chain, but is not explicitly included for O2Physics -- not sure if this means we would also need a corresponding change in alidist (see previous discussion in https://github.com/AliceO2Group/AliceO2/pull/6992 where it was suggested to move the dependency from O2 to O2Physics instead)

### DIFF
--- a/dependencies/FindKFParticle.cmake
+++ b/dependencies/FindKFParticle.cmake
@@ -1,0 +1,41 @@
+# Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+# See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+# All rights not expressly granted are reserved.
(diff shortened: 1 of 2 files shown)


jgrosseo:
This is fine with me. @ktf can you comment on the question above about alidist?

ktf:
Dependencies in aliBuild are transitive, so if O2 depends on KFParticle, O2Physics will as well. However I am not sure that's the case for our CMake, so we might have to adapt the cmake command and make sure the correct -D is provided.

PR #472, opened 2022-01-24
[Event filtering] Modify order of features and replace mass histos with mass vs. pt histos

fgrosa:


alibuild (approved):

PR #475, opened 2022-01-25
PID ML Task

Lukaszz99:


saganatt (dismissed):
Approving for a quick Hyperloop test.

In the future, it could be merged with qaPid or common parts could be factored out to some common class.

saganatt:
@Lukaszz99 you need to correct Clang, it fails.

Tools/PIDML/qaPidML.cxx
```diff
@@ -321,7 +321,7 @@ struct pidml {
   template <std::size_t i, typename T>
   void pidML(const T& track, const int pdgCodeMC)
   {
-    int pidLogits[3] = {model211.applyModel(track), model2212.applyModel(track), model321.applyModel(track)};
+    float pidLogits[3] = {int(model211.applyModel(track)), model2212.applyModel(track), model321.applyModel(track)};
```
saganatt, line 324: You require an int array in pidParticle(), this causes problems.
Could you simply change the pidParticle signature to `const float pidLogits`?
And don't make ints out of elements of a float array, then you get conversion errors.

Tools/PIDML/qaPidML.cxx
```diff
@@ -0,0 +1,367 @@
...
+      histReg.fill(HIST("TPCSignalPidFalse"), track.p(), track.tpcSignal());
+      histReg.fill(HIST("TOFSignalPidFalse"), track.p(), track.beta());
+      histReg.fill(HIST(TOFPidFalseRegistryNames[i]), track.p(), track.beta());
+
+      double pt = track.pt();
+      fillContaminationRegistry(i, pdgCode, pt);
+    }
+  }
+
+  int pidParticle(const int pidLogits[])
```
saganatt, line 298: By the way, it would be less confusing if you renamed this to, e.g., `getParticlePdg()`.
`pidParticle` sounds like a class name.

saganatt (approved):

PR #482, opened 2022-01-27
Add debug task for femto

lbariogl:


PWGCF/DataModel/FemtoDerived.h
```diff
@@ -87,10 +87,10 @@ DECLARE_SOA_DYNAMIC_COLUMN(P, p, //! Compute the overall momentum in GeV/c
                            });
 // debug variables
 DECLARE_SOA_COLUMN(Sign, sign, int8_t);                                                  //! Sign of the track charge
-DECLARE_SOA_COLUMN(TPCNClsFound, tpcNClsFound, uint8_t);                                 //! Number of TPC clusters
-DECLARE_SOA_COLUMN(TPCNClsCrossedRows, tpcNClsCrossedRows, uint8_t);                     //! Number of TPC crossed rows
+DECLARE_SOA_COLUMN(TPCNClsFound, tpcNClsFound, int16_t);                                 //! Number of TPC clusters
+DECLARE_SOA_COLUMN(TPCNClsCrossedRows, tpcNClsCrossedRows, int16_t);                     //! Number of TPC crossed rows
```
jgrosseo, line 91: Why isn't uint8_t enough?

jgrosseo: There are at maximum 159...

lbariogl: I did it for consistency with the definition in O2:

https://github.com/AliceO2Group/AliceO2/blob/dev/Framework/Core/include/Framework/AnalysisDataModel.h#L225

Otherwise a cast is needed
Otherwi

lbariogl: I changed it to `uint8_t` and applied a cast

jgrosseo: Yes, this is better. The other is int16 because it formally could be larger than 8 bit, but in practice it isn't...

jgrosseo (approved):

PR #483, opened 2022-01-28
Adding dimuon vertexing process

mcoquet642:
The "processJpsiToMuMuVertexingSkimmed" function uses covariance tables which enables the use of the pair vertexing method.


iarsene (approved):

iarsene:
The macOS crash is unrelated, merging

PR #487, opened 2022-01-30
Rolling back bool columns to uint8_t

victor-gonzalez:
Also
- fixing double counting when no PID
- not selected by PID should not pass the filtering
- removed not needed using
- fixing not used warnings


alibuild (approved):

victor-gonzalez:
@jgrosseo Anything I can do for the macOS error?

PR #488, opened 2022-01-30
Moving to McParticle_001

jgrosseo:


victor-gonzalez (approved):
They look fine from my side

iarsene:
I checked the DQ files and it seems fine.
Cheers, Ionut

njacazio:
ALICE 3 and LF files should also be ok, thanks!

PR #491, opened 2022-02-01
Fixes to support version 1

jgrosseo:


sawenzel:
Thanks. Then it would be good to go ahead. I'll take away the draft status to see if it compiles.

benedikt-voelkel:
Mh, that is unexpected. I will check

jgrosseo:
> Mh, that is unexpected. I will check

It is not. We need to wait for the daily O2 tag which @ktf triggered. That's why I left this as draft actually...

benedikt-voelkel:
ah, right, sorry for my ignorance and thanks for the clarification

jgrosseo:
The tag is there. I made a dummy commit, to run the CI again

PR #492, opened 2022-02-01
PWGMM: dndeta task: use proper ITS track selection

aalkin:


alibuild (approved):

PR #495, opened 2022-02-02
PWGHF: Remove HFSelTrack from BigTracks

vkucera:
This allows to decouple execution of skimming and candidate
creators which is needed to run the HF analyses from skims
stored in the input AOD file.

PR #497, opened 2022-02-02
[Event filtering] Add back HFTrackSel in join of track tables

fgrosa:
`aod::HFSelTrack` was removed from `aod::BigTracks` here https://github.com/AliceO2Group/O2Physics/pull/495


alibuild (approved):

vkucera:
Hi @fgrosa , do you actually use that table anywhere? Sorry if I missed it.

fgrosa:
Hi @vkucera, right! I got confused because in the past I needed it for the impact parameters of the daughters, but now we take them from the track extension task indeed. I reverted this commit here https://github.com/AliceO2Group/O2Physics/pull/498

PR #501, opened 2022-02-03
Rename HF tables, adjust index to V0

jgrosseo:


jgrosseo:
This is precondition to having the Run 2 -> 3 conversion produce these tables.

@vkucera @fgrosa @ginnocen 

Please check and let me know...

jgrosseo:
For some reason clang started inserting spaces after comments, this makes the diff a bit cluttered.

Have a look how I reworked the index to V0, this is more than renaming.

PWGHF/DataModel/HFSecondaryVertex.h
```diff
@@ -64,12 +64,11 @@ using BigTracksPIDExtended = soa::Join<BigTracksPID, aod::TracksExtended>;
...
-DECLARE_SOA_INDEX_COLUMN_FULL(Index0, index0, int, Tracks, "_0");          //!
-DECLARE_SOA_INDEX_COLUMN_FULL(Index1, index1, int, Tracks, "_1");          //!
-DECLARE_SOA_INDEX_COLUMN_FULL(Index2, index2, int, Tracks, "_2");          //!
-DECLARE_SOA_INDEX_COLUMN_FULL(Index3, index3, int, Tracks, "_3");          //!
-DECLARE_SOA_INDEX_COLUMN_FULL(IndexV0, indexV0, int, aod::V0Datas, "_V0"); //!
-DECLARE_SOA_COLUMN(HFflag, hfflag, uint8_t);                               //!
+DECLARE_SOA_INDEX_COLUMN_FULL(Index0, index0, int, Tracks, "_0"); //! Index to first prong
+DECLARE_SOA_INDEX_COLUMN_FULL(Index1, index1, int, Tracks, "_1"); //! Index to second prong
+DECLARE_SOA_INDEX_COLUMN_FULL(Index2, index2, int, Tracks, "_2"); //! Index to third prong
+DECLARE_SOA_INDEX_COLUMN(V0, v0);                                 //! Index to V0 prong
```
vkucera, line 70: This makes the names of index columns inconsistent as only the track column names start with `Index`.
Actually, I would suggest to rename `Index` to `Prong` because when the column getter is called, one gets the prong table row itself, not its index (which is returned by `Id`). So my proposal is: `Prong0`,..., `ProngV0`.
@jgrosseo @ginnocen What do you think?

jgrosseo: I am fine with your proposal, as long as the branch names in the tree stay as they are (which they do with your proposal).

As you see I didn't touch the names of the track columns, they were like this already before. 

Is it OK if we merge this and you make the modifications of the names in a subsequent PR?

jgrosseo: Btw. the AliPhysics PR has been merged, so if you update the validation fw will produce the HF tables already. You have to add the wagon HFCandidateCreator before the converter, see https://alimonitor.cern.ch/trains/train.jsp?train_id=132

vkucera: > I am fine with your proposal, as long as the branch names in the tree stay as they are (which they do with your proposal).
> 
> As you see I didn't touch the names of the track columns, they were like this already before.
> 
> Is it OK if we merge this and you make the modifications of the names in a subsequent PR?

Yes, it is fine.

vkucera (approved):

PWGHF/DataModel/HFSecondaryVertex.h
jgrosseo: OK. Merged.
`DECLARE_SOA_INDEX_COLUMN(V0, v0);` will then become `DECLARE_SOA_INDEX_COLUMN_FULL(ProngV0, prongV0, int, Tracks, "");`
(note the empty argument)
You may consider also making `prongD0` for the D*

PR #503, opened 2022-02-03
Change in run2dndeta task to reweight tracks histograms for sim reco

sarahherrmann:


PWGMM/Tasks/run2dndeta.cxx
```diff
@@ -205,7 +251,7 @@ struct PseudorapidityDensity {
     }
   }
 
-  PROCESS_SWITCH(PseudorapidityDensity, processGen, "Process generator-level info", false);
```
aalkin, line 208: Due to limitations of process switches it is better to have them `false` by default, since the configuration file with have all of them set anyway.

PWGMM/Tasks/run2dndeta.cxx
```diff
@@ -100,23 +114,55 @@ struct PseudorapidityDensity {
...
+    if (doprocessGen) //we are sudying the MC, that needs correction
+    {
+      if (!useEvSel || (useEvSel && collision.sel7())) {
+
+        auto z = collision.posZ();
+        auto perCollisionSample = sample->sliceByCached(aod::track::collisionId, collision.globalIndex());
+        auto v = fRatio->Eval(z);
+
+        double r = rand->Rndm(); //random number between 0 and 1
+        if ((1 - v) <= r)        //then we accept the event
```
aalkin, line 126: This way you duplicate the processing code, if you use `if (r < v) { return; }` you can just skip the collision and have the processing fragment only once after the condition.

aalkin (approved):

PR #469, opened 2022-01-21
Introduce utilities to get nsigma with index

njacazio:

PR #504, opened 2022-02-04
Extend PID utils

njacazio:


njacazio:
@jgrosseo this extends the getter menu, also including compile time definition so as to handle @mpuccio suggestion on https://github.com/AliceO2Group/O2Physics/pull/469 
There might be better solutions but it is probably still useful to have this implemented in analyses already and then we can update it centrally

jgrosseo (approved):

PR #506, opened 2022-02-04
Including V0M definition for comparison to Run 2

sartozza:


victor-gonzalez (approved):
I approve your changes but don't understand why you get only T0A and not T0M and also why you take half the V0M multiplicity for converted Run2

PR #510, opened 2022-02-07
PWGHF: Make transition from McParticles_000 to McParticles_001

fgrosa:
@jgrosseo @vkucera @fcolamar please double check that I did not mess up anything. Thanks a lot!

### DIFF
--- a/Common/Core/RecoDecay.h
+++ b/Common/Core/RecoDecay.h
@@ -569,7 +569,8 @@ class RecoDecay
   /// \param depthMax  maximum decay tree level to check; Mothers up to this level will be considered. If -1, all levels are considered.
   /// \return index of the mother particle if found, -1 otherwise
   template <typename T>
-  static int getMother(const T& particle,
+  static int getMother(const T& particlesMC,
+                       const typename T::iterator& particle,
                        int PDGMother,
                        bool acceptAntiParticles = false,
                        int8_t* sign = nullptr,
@@ -586,8 +587,11 @@ class RecoDecay
       if (depthMax > -1 && -stage >= depthMax) { // Maximum depth has been reached.
         return -1;
       }
-      particleMother = particleMother.template mothers_first_as<typename std::decay_t<T>::parent_t>(); // get mother 0
-      auto indexMotherTmp = particleMother.globalIndex();
+      auto indexMotherTmp = particleMother.mothersIds().front();
+      if (indexMotherTmp <= -1) {
+        return -1;
+      }
+      particleMother = particlesMC.rawIteratorAt(indexMotherTmp);
       // Check mother's PDG code.
       auto PDGParticleIMother = particleMother.pdgCode(); // PDG code of the mother
       //printf("getMother: ");
@@ -620,7 +624,8 @@ class RecoDecay
   /// \note Final state is defined as particles from arrPDGFinal plus final daughters of any other decay branch.
   /// \note Antiparticles of particles in arrPDGFinal are accepted as well.
   template <std::size_t N, typename T>
-  static void getDaughters(const T& particle,
+  static void getDaughters(const T& particlesMC,
+                           const typename T::iterator& particle,
                            std::vector<int>* list,
                            const array<int, N>& arrPDGFinal,
                            int8_t depthMax = -1,
@@ -671,8 +676,8 @@ class RecoDecay
     //printf("Stage %d: %d (PDG %d) -> %d-%d\n", stage, index, PDGParticle, indexDaughterFirst, indexDaughterLast);
     // Call itself to get daughters of daughters recursively.
     stage++;
-    for (auto& dau : particle.template daughters_as<typename std::decay_t<T>::parent_t>()) {
-      getDaughters(dau, list, arrPDGFinal, depthMax, stage);
+    for (auto& idxDau : particle.daughtersIds()) {
+      getDaughters(particlesMC, particlesMC.rawIteratorAt(idxDau), list, arrPDGFinal, depthMax, stage);
     }
   }
 
@@ -713,7 +718,7 @@ class RecoDecay
       if (iProng == 0) {
         // Get the mother index and its sign.
         // PDG code of the first daughter's mother determines whether the expected mother is a particle or antiparticle.
-        indexMother = getMother(particleI, PDGMother, acceptAntiParticles, &sgn, depthMax);
+        indexMother = getMother(particlesMC, particleI, PDGMother, acceptAntiParticles, &sgn, depthMax);
         // Check whether mother was found.
         if (indexMother <= -1) {
           //Printf("MC Rec: Rejected: bad mother index or PDG");
@@ -727,13 +732,12 @@ class RecoDecay
           return -1;
         }
         // Check that the number of direct daughters is not larger than the number of expected final daughters.
-        auto dauTable = particleMother.template daughters_as<T>();
-        if (dauTable.size() > int(N)) {
+        if (particleMother.daughtersIds().back() - particleMother.daughtersIds().front() + 1 > int(N)) {
           //Printf("MC Rec: Rejected: too many direct daughters: %d (expected %ld final)", particleMother.daughtersIds().size(), N);
           return -1;
         }
         // Get the list of actual final daughters.
-        getDaughters(particleMother, &arrAllDaughtersIndex, arrPDGDaughters, depthMax);
+        getDaughters(particlesMC, particleMother, &arrAllDaughtersIndex, arrPDGDaughters, depthMax);
         //printf("MC Rec: Mother %d has %d final daughters:", indexMother, arrAllDaughtersIndex.size());
         //for (auto i : arrAllDaughtersIndex) {
         //  printf(" %d", i);
@@ -846,13 +850,13 @@ class RecoDecay
         return false;
       }
       // Check that the number of direct daughters is not larger than the number of expected final daughters.
-      auto dauTable = candidate.template daughters_as<T>();
-      if (dauTable.size() > int(N)) {
-        //Printf("MC Gen: Rejected: too many direct daughters: %d (expected %ld final)", candidate.daughtersIds().size(), N);
+
+      if (candidate.daughtersIds().back() - candidate.daughtersIds().front() + 1 > int(N)) {
+        //Printf("MC Gen: Rejected: too many direct daughters: %d (expected %ld final)", candidate.daughtersIds().back()-candidate.daughtersIds().front()+1, N);
         return false;
       }
       // Get the list of actual final daughters.
-      getDaughters(candidate, &arrAllDaughtersIndex, arrPDGDaughters, depthMax);
+      getDaughters(particlesMC, candidate, &arrAllDaughtersIndex, arrPDGDaughters, depthMax);
       //printf("MC Gen: Mother %ld has %ld final daughters:", candidate.globalIndex(), arrAllDaughtersIndex.size());
       //for (auto i : arrAllDaughtersIndex) {
       //  printf(" %d", i);
--- a/Common/Core/RecoDecay.h
+++ b/Common/Core/RecoDecay.h
@@ -587,7 +587,7 @@ class RecoDecay
         return -1;
       }
       particleMother = particleMother.template mothers_first_as<typename std::decay_t<T>::parent_t>(); // get mother 0
-      auto particleMotherIdx = particleMother.globalIndex();
+      auto indexMotherTmp = particleMother.globalIndex();
       // Check mother's PDG code.
       auto PDGParticleIMother = particleMother.pdgCode(); // PDG code of the mother
       //printf("getMother: ");
@@ -596,11 +596,11 @@ class RecoDecay
       //printf("Stage %d: Mother PDG: %d\n", stage, PDGParticleIMother);
       if (PDGParticleIMother == PDGMother) { // exact PDG match
         sgn = 1;
-        indexMother = particleMotherIdx;
+        indexMother = indexMotherTmp;
         break;
       } else if (acceptAntiParticles && PDGParticleIMother == -PDGMother) { // antiparticle PDG match
         sgn = -1;
-        indexMother = particleMotherIdx;
+        indexMother = indexMotherTmp;
         break;
       }
       stage--;
@@ -612,7 +612,6 @@ class RecoDecay
   }
 
   /// Gets the complete list of indices of final-state daughters of an MC particle.
-  /// \param particlesMC  table with MC particles
   /// \param particle  MC particle
   /// \param list  vector where the indices of final-state daughters will be added
   /// \param arrPDGFinal  array of PDG codes of particles to be considered final if found
@@ -621,8 +620,7 @@ class RecoDecay
   /// \note Final state is defined as particles from arrPDGFinal plus final daughters of any other decay branch.
   /// \note Antiparticles of particles in arrPDGFinal are accepted as well.
   template <std::size_t N, typename T>
-  static void getDaughters(const T& particlesMC,
-                           const typename T::iterator& particle,
+  static void getDaughters(const T& particle,
                            std::vector<int>* list,
                            const array<int, N>& arrPDGFinal,
                            int8_t depthMax = -1,
@@ -667,10 +665,13 @@ class RecoDecay
       return;
     }
     // If we are here, we have to follow the daughter tree.
-    // Call itself to get daughters of daughters recursively.
+    //printf("getDaughters: ");
+    //for (int i = 0; i < stage; i++) // Indent to make the tree look nice.
+    //  printf(" ");
+    //printf("Stage %d: %d (PDG %d) -> %d-%d\n", stage, index, PDGParticle, indexDaughterFirst, indexDaughterLast);    // Call itself to get daughters of daughters recursively.
     stage++;
-    for (auto& dauIdx : particle.daughtersIds()) {
-      getDaughters(particlesMC, particlesMC.rawIteratorAt(dauIdx), list, arrPDGFinal, depthMax, stage);
+    for (auto& dau : particle.template daughters_as<typename std::decay_t<T>::parent_t>()) {
+      getDaughters(dau, list, arrPDGFinal, depthMax, stage);
     }
   }
 
@@ -728,8 +729,12 @@ class RecoDecay
           return -1;
         }
         // Get the list of actual final daughters.
-        getDaughters(particlesMC, particleMother, &arrAllDaughtersIndex, arrPDGDaughters, depthMax);
-        // Check whether the number of actual final daughters is equal to the number of expected final daughters (i.e. the number of provided prongs).
+        getDaughters(particleMother, &arrAllDaughtersIndex, arrPDGDaughters, depthMax);
+        //printf("MC Rec: Mother %d has %d final daughters:", indexMother, arrAllDaughtersIndex.size());
+        //for (auto i : arrAllDaughtersIndex) {
+        //  printf(" %d", i);
+        //}
+        //printf("\n");        // Check whether the number of actual final daughters is equal to the number of expected final daughters (i.e. the number of provided prongs).
         if (arrAllDaughtersIndex.size() != N) {
           //Printf("MC Rec: Rejected: incorrect number of final daughters: %ld (expected %ld)", arrAllDaughtersIndex.size(), N);
           return -1;
@@ -839,7 +844,7 @@ class RecoDecay
         return false;
       }
       // Get the list of actual final daughters.
-      getDaughters(particlesMC, candidate, &arrAllDaughtersIndex, arrPDGDaughters, depthMax);
+      getDaughters(candidate, &arrAllDaughtersIndex, arrPDGDaughters, depthMax);
       //printf("MC Gen: Mother %ld has %ld final daughters:", candidate.globalIndex(), arrAllDaughtersIndex.size());
       //for (auto i : arrAllDaughtersIndex) {
       //  printf(" %d", i);
--- a/Common/Core/RecoDecay.h
+++ b/Common/Core/RecoDecay.h
@@ -588,7 +588,8 @@ class RecoDecay
       if (depthMax > -1 && -stage >= depthMax) { // Maximum depth has been reached.
         return -1;
       }
-      particleMother = particleMother.mothers().front(); // get mother 0 
+      auto particleMotherIdx = particleMother.mothersIds().front(); // get mother 0
+      particleMother = particlesMC.iteratorAt(particleMotherIdx);
       // Check mother's PDG code.
       auto PDGParticleIMother = particleMother.pdgCode(); // PDG code of the mother
       //printf("getMother: ");
@@ -638,8 +639,8 @@ class RecoDecay
       isFinal = true;
     }
     // Get the range of daughter indices.
-    int indexDaughterFirst = particle.daughter0Id();
-    int indexDaughterLast = particle.daughter1Id();
+    int indexDaughterFirst = particle.daughtersIds().front();
+    int indexDaughterLast = particle.daughtersIds().back();
     // Check whether there are any daughters.
     if (!isFinal && indexDaughterFirst <= -1 && indexDaughterLast <= -1) {
       // If the original particle has no daughters, we do nothing and exit.
@@ -677,22 +678,8 @@ class RecoDecay
     //printf("Stage %d: %d (PDG %d) -> %d-%d\n", stage, index, PDGParticle, indexDaughterFirst, indexDaughterLast);
     // Call itself to get daughters of daughters recursively.
     stage++;
-    // Get daughters of the first daughter.
-    if (indexDaughterFirst > -1) {
-      getDaughters(particlesMC, particlesMC.iteratorAt(indexDaughterFirst), list, arrPDGFinal, depthMax, stage);
-    }
-    // Get daughters of the daughters in between if any.
-    // Daughter indices are supposed to be consecutive and in increasing order.
-    // Reverse order means two daughters.
-    if (indexDaughterFirst > -1 && indexDaughterLast > -1) {
-      for (auto iD = indexDaughterFirst + 1; iD < indexDaughterLast; ++iD) {
-        getDaughters(particlesMC, particlesMC.iteratorAt(iD), list, arrPDGFinal, depthMax, stage);
-      }
-    }
-    // Get daughters of the last daughter if different from the first one.
-    // Same indices indicate a single daughter.
-    if (indexDaughterLast > -1 && indexDaughterLast != indexDaughterFirst) {
-      getDaughters(particlesMC, particlesMC.iteratorAt(indexDaughterLast), list, arrPDGFinal, depthMax, stage);
+    for (auto& dauIdx : particle.daughtersIds()) {
+      getDaughters(particlesMC, particlesMC.iteratorAt(dauIdx), list, arrPDGFinal, depthMax, stage);
     }
   }
 
@@ -727,7 +714,7 @@ class RecoDecay
       if (!arrDaughters[iProng].has_mcParticle()) {
         return -1;
       }
-      auto particleI = arrDaughters[iProng].template mcParticle_as<aod::McParticles_000>(); // ith daughter particle
+      auto particleI = arrDaughters[iProng].template mcParticle_as<aod::McParticles>(); // ith daughter particle
       arrDaughtersIndex[iProng] = particleI.globalIndex();
       // Get the list of daughter indices from the mother of the first prong.
       if (iProng == 0) {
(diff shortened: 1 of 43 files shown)


Common/Core/RecoDecay.h
```diff
@@ -728,7 +714,7 @@ class RecoDecay
       if (!arrDaughters[iProng].has_mcParticle()) {
         return -1;
       }
-      auto particleI = arrDaughters[iProng].template mcParticle_as<aod::McParticles_000>(); // ith daughter particle
+      auto particleI = arrDaughters[iProng].template mcParticle_as<aod::McParticles>(); // ith daughter particle
```
jgrosseo, line 717: This could also become `auto particleI = arrDaughters[iProng].mcParticle();`

PWGHF/TableProducer/HFCandidateCreator3Prong.cxx
```diff
@@ -193,7 +193,7 @@ struct HFCandidateCreator3ProngExpressions {
 
           //Printf("Flagging the different Λc± → p± K∓ π± decay channels");
           if (arrayDaughters[0].has_mcParticle()) {
-            swapping = int8_t(std::abs(arrayDaughters[0].mcParticle_as<aod::McParticles_000>().pdgCode()) == kPiPlus);
+            swapping = int8_t(std::abs(arrayDaughters[0].mcParticle_as<aod::McParticles>().pdgCode()) == kPiPlus);
```
jgrosseo, line 196: `mcParticle()`

Common/Core/RecoDecay.h
```diff
@@ -598,11 +598,11 @@ class RecoDecay
       //printf("Stage %d: Mother PDG: %d\n", stage, PDGParticleIMother);
       if (PDGParticleIMother == PDGMother) { // exact PDG match
         sgn = 1;
-        indexMother = indexMotherTmp;
+        indexMother = particleMother.globalIndex();
```
vkucera, line 601: How is `particleMother.globalIndex()` different from `particleMotherIdx`?

PWGHF/TableProducer/HFCandidateCreatorChic.cxx
```diff
@@ -233,12 +233,12 @@ struct HFCandidateCreatorChicMC {
...
           auto particleMother = particlesMC.iteratorAt(indexMother);
           hEphotonMatched->Fill(candidate.index1().e());
           hMassEMatched->Fill(sqrt(candidate.index1().px() * candidate.index1().px() + candidate.index1().py() * candidate.index1().py() + candidate.index1().pz() * candidate.index1().pz()));
-          int indexDaughterFirst = particleMother.daughter0Id(); // index of the first direct daughter
-          int indexDaughterLast = particleMother.daughter1Id();  // index of the last direct daughter
+          int indexDaughterFirst = particleMother.daughtersIds().front(); // index of the first direct daughter
+          int indexDaughterLast = particleMother.daughtersIds().back();   // index of the last direct daughter
           if ((indexDaughterFirst > -1 && indexDaughterLast > -1)) {
             std::vector<int> arrAllDaughtersIndex;
             RecoDecay::getDaughters(particlesMC, particleMother, &arrAllDaughtersIndex, array{(int)(kGamma), (int)(pdg::Code::kJpsi)}, 1);
```
jgrosseo, line 244: @vkucera Shall we also change the syntax of this function? Converting internally now in a pair of numbers where `daughterIds()` is already a pair of numbers seems not needed...

Common/Core/RecoDecay.h
fgrosa: Done!

PWGHF/TableProducer/HFCandidateCreator3Prong.cxx
fgrosa: Done! (everywhere)

Common/Core/RecoDecay.h
fgrosa: Indeed, fix it!

PWGHF/TableProducer/HFCandidateCreatorChic.cxx
jgrosseo: Replace by just `has_daughters()`

PWGHF/TableProducer/HFCandidateCreator3Prong.cxx
jgrosseo: `arrayDaughters[0].mcParticle().pdgCode()` did not work here?

Common/Core/RecoDecay.h
```diff
@@ -742,8 +728,8 @@ class RecoDecay
         }
         //Printf("MC Rec: Good mother: %d", indexMother);
         auto particleMother = particlesMC.iteratorAt(indexMother);
-        int indexDaughterFirst = particleMother.daughter0Id(); // index of the first direct daughter
-        int indexDaughterLast = particleMother.daughter1Id();  // index of the last direct daughter
+        int indexDaughterFirst = particleMother.daughtersIds().front(); // index of the first direct daughter
+        int indexDaughterLast = particleMother.daughtersIds().back();   // index of the last direct daughter
         // Check the daughter indices.
         if (indexDaughterFirst <= -1 && indexDaughterLast <= -1) {
```
jgrosseo, line 734: Replace by `has_daughters()`

jgrosseo: Further below line 753 use `daughters().size()`

PWGHF/Tasks/HFMCValidation.cxx
```diff
@@ -173,15 +173,15 @@ struct ValidationRecLevel {
...
+          indexParticle = RecoDecay::getMother(particlesMC, candidate.index0_as<aod::BigTracksMC>().mcParticle(), pdg::Code::kD0, true);
         }
         auto mother = particlesMC.iteratorAt(indexParticle);
         registry.fill(HIST("histPt"), candidate.pt() - mother.pt());
         registry.fill(HIST("histPx"), candidate.px() - mother.px());
         registry.fill(HIST("histPy"), candidate.py() - mother.py());
         registry.fill(HIST("histPz"), candidate.pz() - mother.pz());
         //Compare Secondary vertex and decay length with MC
-        auto daughter0 = mother.daughter0_as<aod::McParticles_000>();
+        auto daughter0 = particlesMC.iteratorAt(mother.daughtersIds().front());
```
jgrosseo, line 184: Does `mother.daughters().front()` work here?

PWGHF/Tasks/taskD0parametrizedPID.cxx
```diff
@@ -62,9 +62,9 @@ struct TaskD0parametrizedPIDMC {
...
 
   Filter filterSelectCandidates = (aod::hf_selcandidate_d0_parametrizedPID::isSelD0NoPID >= 1 || aod::hf_selcandidate_d0_parametrizedPID::isSelD0barNoPID >= 1);
-  using McParticlesHf = soa::Join<aod::McParticles_000, aod::HfCandProng2MCGen>;
-  void process(soa::Filtered<soa::Join<aod::HfCandProng2, aod::HFSelD0CandidateparametrizedPID, aod::HfCandProng2MCRec>> const& candidates, soa::Join<aod::McParticles_000, aod::HfCandProng2MCGen> const& particlesMC, aod::BigTracksMC const& tracks)
-  // void process(const o2::aod::Collision& collision, soa::Filtered<soa::Join<aod::HfCandProng2, aod::HFSelD0CandidateparametrizedPID, aod::HfCandProng2MCRec>> const& candidates, soa::Join<aod::McParticles_000, aod::HfCandProng2MCGen> const& particlesMC, aod::BigTracksMC const& tracks)
+  using McParticlesHf = soa::Join<aod::McParticles, aod::HfCandProng2MCGen>;
+  void process(soa::Filtered<soa::Join<aod::HfCandProng2, aod::HFSelD0CandidateparametrizedPID, aod::HfCandProng2MCRec>> const& candidates, soa::Join<aod::McParticles, aod::HfCandProng2MCGen> const& particlesMC, aod::BigTracksMC const& tracks)
```
jgrosseo, line 66: Why not use the `using` in the argument list?

PWGHF/Tasks/taskD0parametrizedPID.cxx
```diff
@@ -131,8 +131,8 @@ struct TaskD0parametrizedPIDMC {
           maxFiducialY = -0.2 / 15 * ptGen * ptGen + 1.9 / 15 * ptGen + 0.5;
           minFiducialY = 0.2 / 15 * ptGen * ptGen - 1.9 / 15 * ptGen - 0.5;
         }
-        auto etaProng0 = std::abs(particle.daughter0_as<McParticlesHf>().eta());
```
jgrosseo, line 134: This didn't work?

Common/Core/RecoDecay.h
```diff
@@ -862,15 +848,15 @@ class RecoDecay
...
+      int indexDaughterFirst = candidate.daughtersIds().front(); // index of the first direct daughter
+      int indexDaughterLast = candidate.daughtersIds().back();   // index of the last direct daughter
       // Check the daughter indices.
       if (indexDaughterFirst <= -1 && indexDaughterLast <= -1) {
         //Printf("MC Gen: Rejected: bad daughter index range: %d-%d", indexDaughterFirst, indexDaughterLast);
         return false;
       }
       // Check that the number of direct daughters is not larger than the number of expected final daughters.
-      if (indexDaughterFirst > -1 && indexDaughterLast > -1 && indexDaughterLast - indexDaughterFirst + 1 > int(N)) {
+      if (indexDaughterFirst > -1 && indexDaughterLast > -1 && candidate.daughtersIds().size() > int(N)) {
```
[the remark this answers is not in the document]
vkucera: `candidate.daughters().size()` would work the same, right?

vkucera:
Actually, there are a few places in `RecoDecay` that would be benefit from replacement with `has_daughters()` and `.daughters().size()`.

PWGHF/Tasks/taskD0parametrizedPID.cxx
jgrosseo: I mean `particle.daughter_as<McParticlesHf>().front().eta()`

PWGHF/Tasks/HFMCValidation.cxx
fgrosa: This does not work, but I will update it with the method to retrieve the first daughter implemented in https://github.com/AliceO2Group/AliceO2/pull/8113

PWGHF/Tasks/taskD0parametrizedPID.cxx
fgrosa: This does not work, but I will update it with the method to retrieve the first/last daughters implemented in https://github.com/AliceO2Group/AliceO2/pull/8113

fgrosa:
> Actually, there are a few more places in `RecoDecay` that would benefit from replacement with `.has_daughters()` and `.daughters().size()`.

Should be done everywhere now!

PWGHF/Tasks/taskD0parametrizedPID.cxx
fgrosa: The previous comment was not correct: as discussed in person I needed to use `begin()` instead of `front()` so it works now

PWGHF/Tasks/HFMCValidation.cxx
fgrosa: The previous comment was not correct: as discussed in person I needed to use `begin()` instead of `front()` so it works now

fgrosa:
@jgrosseo @vkucera the PR should be now updated using the methods implemented in https://github.com/AliceO2Group/AliceO2/pull/8113 and `rawIteratorAt()` instead of `iteratorAt()`

jgrosseo (approved):

Common/Core/RecoDecay.h
```diff
@@ -598,11 +596,11 @@ class RecoDecay
       //printf("Stage %d: Mother PDG: %d\n", stage, PDGParticleIMother);
       if (PDGParticleIMother == PDGMother) { // exact PDG match
         sgn = 1;
-        indexMother = indexMotherTmp;
+        indexMother = particleMotherIdx;
```
vkucera, line 599: Revert. Variables starting with `particle` are meant to be particles everywhere else in the code.

Common/Core/RecoDecay.h
```diff
@@ -584,12 +582,12 @@ class RecoDecay
...
       *sign = sgn;
     }
-    while (particleMother.has_mother0()) {
+    while (particleMother.has_mothers()) {
       if (depthMax > -1 && -stage >= depthMax) { // Maximum depth has been reached.
         return -1;
       }
-      auto indexMotherTmp = particleMother.mother0Id();
-      particleMother = particlesMC.iteratorAt(indexMotherTmp);
+      particleMother = particleMother.template mothers_first_as<typename std::decay_t<T>::parent_t>(); // get mother 0
```
vkucera, line 589: Is this really the simplest way to get the first mother?

Common/Core/RecoDecay.h
```diff
@@ -672,28 +667,10 @@ class RecoDecay
       return;
     }
     // If we are here, we have to follow the daughter tree.
-    //printf("getDaughters: ");
-    //for (int i = 0; i < stage; i++) // Indent to make the tree look nice.
-    //  printf(" ");
-    //printf("Stage %d: %d (PDG %d) -> %d-%d\n", stage, index, PDGParticle, indexDaughterFirst, indexDaughterLast);
```
vkucera, line 678: Do not remove the debugging prints. They are useful whenever a new decay is being added.

Common/Core/RecoDecay.h
```diff
@@ -672,28 +667,10 @@ class RecoDecay
...
-      for (auto iD = indexDaughterFirst + 1; iD < indexDaughterLast; ++iD) {
-        getDaughters(particlesMC, particlesMC.iteratorAt(iD), list, arrPDGFinal, depthMax, stage);
-      }
-    }
-    // Get daughters of the last daughter if different from the first one.
-    // Same indices indicate a single daughter.
-    if (indexDaughterLast > -1 && indexDaughterLast != indexDaughterFirst) {
-      getDaughters(particlesMC, particlesMC.iteratorAt(indexDaughterLast), list, arrPDGFinal, depthMax, stage);
+    for (auto& dauIdx : particle.daughtersIds()) {
+      getDaughters(particlesMC, particlesMC.rawIteratorAt(dauIdx), list, arrPDGFinal, depthMax, stage);
```
vkucera, line 673: This is the only place where the `particlesMC` table is used. Could it be removed by looping over `particle.daughters()` instead?

vkucera (changes requested):

Common/Core/RecoDecay.h
fgrosa: This was suggested by @aalkin, I don't think there are easier ways to do it

Common/Core/RecoDecay.h
fgrosa: Reverted

Common/Core/RecoDecay.h
fgrosa: Right! I removed the argument and I replaced the loop over indices with a loop over particles directly

Common/Core/RecoDecay.h
aalkin: All these is required because we need to call a dependent template inside a templated function, in `process()` one would not need the `template` keyword and type calculation, of course.

Common/Core/RecoDecay.h
```diff
@@ -675,25 +668,10 @@ class RecoDecay
     //printf("getDaughters: ");
     //for (int i = 0; i < stage; i++) // Indent to make the tree look nice.
     //  printf(" ");
-    //printf("Stage %d: %d (PDG %d) -> %d-%d\n", stage, index, PDGParticle, indexDaughterFirst, indexDaughterLast);
-    // Call itself to get daughters of daughters recursively.
+    //printf("Stage %d: %d (PDG %d) -> %d-%d\n", stage, index, PDGParticle, indexDaughterFirst, indexDaughterLast);    // Call itself to get daughters of daughters recursively.
```
vkucera, line 671: Looks like a line break was removed by accident.

Common/Core/RecoDecay.h
```diff
@@ -728,40 +706,35 @@ class RecoDecay
...
         // Check the daughter indices.
-        if (indexDaughterFirst <= -1 && indexDaughterLast <= -1) {
-          //Printf("MC Rec: Rejected: bad daughter index range: %d-%d", indexDaughterFirst, indexDaughterLast);
+        if (!particleMother.has_daughters()) {
           return -1;
         }
         // Check that the number of direct daughters is not larger than the number of expected final daughters.
-        if (indexDaughterFirst > -1 && indexDaughterLast > -1 && indexDaughterLast - indexDaughterFirst + 1 > int(N)) {
-          //Printf("MC Rec: Rejected: too many direct daughters: %d (expected %ld final)", indexDaughterLast - indexDaughterFirst + 1, N);
+        if (particleMother.daughtersIds().size() > int(N)) {
```
vkucera, line 728: Keep the debugging `Printf`.

Common/Core/RecoDecay.h
```diff
@@ -728,40 +706,35 @@ class RecoDecay
...
         // Get the list of actual final daughters.
-        getDaughters(particlesMC, particleMother, &arrAllDaughtersIndex, arrPDGDaughters, depthMax);
+        getDaughters(particleMother, &arrAllDaughtersIndex, arrPDGDaughters, depthMax);
         //printf("MC Rec: Mother %d has %d final daughters:", indexMother, arrAllDaughtersIndex.size());
         //for (auto i : arrAllDaughtersIndex) {
         //  printf(" %d", i);
         //}
-        //printf("\n");
-        // Check whether the number of actual final daughters is equal to the number of expected final daughters (i.e. the number of provided prongs).
+        //printf("\n");        // Check whether the number of actual final daughters is equal to the number of expected final daughters (i.e. the number of provided prongs).
```
vkucera, line 737: Looks like a line break was removed by accident.

Common/Core/RecoDecay.h
```diff
@@ -862,20 +835,16 @@ class RecoDecay
     if (N > 0) {
       //Printf("MC Gen: Checking %d daughters", N);
       std::vector<int> arrAllDaughtersIndex;            // vector of indices of all daughters
-      int indexDaughterFirst = candidate.daughter0Id(); // index of the first direct daughter
-      int indexDaughterLast = candidate.daughter1Id();  // index of the last direct daughter
       // Check the daughter indices.
-      if (indexDaughterFirst <= -1 && indexDaughterLast <= -1) {
-        //Printf("MC Gen: Rejected: bad daughter index range: %d-%d", indexDaughterFirst, indexDaughterLast);
+      if (!candidate.has_daughters()) {
```
vkucera, line 839: Keep the debugging `Printf`.
```C++
//Printf("MC Gen: Rejected: no daughters");
```

Common/Core/RecoDecay.h
```diff
@@ -862,20 +835,16 @@ class RecoDecay
...
         return false;
       }
       // Check that the number of direct daughters is not larger than the number of expected final daughters.
-      if (indexDaughterFirst > -1 && indexDaughterLast > -1 && indexDaughterLast - indexDaughterFirst + 1 > int(N)) {
-        //Printf("MC Gen: Rejected: too many direct daughters: %d (expected %ld final)", indexDaughterLast - indexDaughterFirst + 1, N);
+      if (candidate.daughtersIds().size() > int(N)) {
```
vkucera, line 843: Keep the debugging `Printf`.

vkucera (approved):
The code looks fine. Just please fix the removed comments.

vkucera (changes requested):
The code looks fine. Just please fix the removed comments.

Common/Core/RecoDecay.h
fgrosa: Done

Common/Core/RecoDecay.h
fgrosa: Done

Common/Core/RecoDecay.h
fgrosa: Fixed

Common/Core/RecoDecay.h
fgrosa: Done

Common/Core/RecoDecay.h
fgrosa: Fixed

Common/Core/RecoDecay.h
vkucera: This has not been fixed yet. Instead the original lines were pasted below.

fgrosa: RIght, I overlooked that, it should be now fixed!

vkucera (commented):
Thanks a lot Fabrizio!
Please verify that the MC plots from the validation framework match the reference.

jgrosseo (dismissed):

Common/Core/RecoDecay.h
```diff
@@ -678,22 +671,8 @@ class RecoDecay
...
-        getDaughters(particlesMC, particlesMC.iteratorAt(iD), list, arrPDGFinal, depthMax, stage);
-      }
-    }
-    // Get daughters of the last daughter if different from the first one.
-    // Same indices indicate a single daughter.
-    if (indexDaughterLast > -1 && indexDaughterLast != indexDaughterFirst) {
-      getDaughters(particlesMC, particlesMC.iteratorAt(indexDaughterLast), list, arrPDGFinal, depthMax, stage);
+    for (auto& dau : particle.template daughters_as<typename std::decay_t<T>::parent_t>()) {
+      getDaughters(dau, list, arrPDGFinal, depthMax, stage);
     }
```
fgrosa, line 676: @vkucera when this piece of code is executed I get the following error:
`[ERROR] Exception caught: Unable to find column with label fIndexMcCollisions`
even if I checked and the column `fIndexMcCollisions` is present in the `O2mccollisionlabel` table in the `AO2D` file that I am analysing. 
I wrote this piece of code following what was suggested by @aalkin for 
`particleMother = particleMother.template mothers_first_as<typename std::decay_t<T>::parent_t>();` 
however I am not sure that it works also in this case. I also tried by replacing `typename std::decay_t<T>::parent_t` with simply `aod::McParticles` but I get the same error. @aalkin @jgrosseo do you have any suggestion? Thanks!

aalkin: @fgrosa this particular exception is raised when constructing a table. When the slice index is unset it is supposed to return empty table. It may happen that this actually creates an empty table with no schema and thus the construction fails. For now you can protect this piece of code with check for `has_daughters()`.

fgrosa: Hi @aalkin thanks! Unfortunately even if I put the protection:
```cpp
    if (particle.has_daughters()) {
      for (auto& dau : particle.template daughters_as<typename std::decay_t<T>::parent_t>()) {
        getDaughters(dau, list, arrPDGFinal, depthMax, stage);
      }
    }
```
it keeps crashing with the same error. Do you have any other idea by chance?

aalkin: I will have to debug this, I assume the line `auto dauTable = particle.template daughters_as<typename std::decay_t<T>::parent_t>();` would crash as well, could you check?

vkucera: Checking for `has_daughters()` has no effect here because this line is reached only if `has_daughters()` returned `true`.

fgrosa: @aalkin indeed it crashes as well. @vkucera right because if `isFinal` is `true` and `has_daughters()` is false it goes through the first `if` statement, but the it returns anyway before this part of the code. 
I think that one possible solution would be to loop over indices in this case instead of getting the sliced table of daughters, but it would require to pass again the MC particle table as argument. However I have no better ideas at the moment. What do you think?

vkucera: @aalkin Would this work?
```c++
  template <std::size_t N, typename T>
  static void getDaughters(const typename T::iterator& particle,
    ...
    for (auto& dau : particle.daughters_as<T>()) {
```

aalkin: @vkucera I doubt this will be able to deduce type automatically from a function call, but it is better to check explicitly.

fgrosa: @vkucera @aalkin I already tried this, but the compilation fails with the following error: `couldn't deduce template parameter 'T'`

fgrosa:
Hi @jgrosseo @vkucera @aalkin with my last commit I reverted the code to the usage of `daughtersIds()` and `mothersIds()` in `RecoDecay.h` instead of `mothers_first_as<>` and `daughters_as<>` since I did not find a way to make it work otherwise. With this implementation I get the same result as compared to the reference efficiency in https://github.com/AliceO2Group/Run3Analysisvalidation/blob/master/codeHF/MC_d0_eff_ref.pdf so if you agree I would proceed with this now in the seek of moving forward (but indeed it would be good to understand why I did not manage to make it work otherwise). What do you think? Thanks a lot!

jgrosseo:
I agree to move ahead and merge this. Could you subsequently, prepare a PR where you put `mothers_first_as` and `daughters_as` in a single location, such that @aalkin could follow up on it?

jgrosseo (dismissed):

jgrosseo:
Waiting for @vkucera approval of course...

vkucera (changes requested):

Common/Core/RecoDecay.h
```diff
@@ -562,7 +562,6 @@ class RecoDecay
   }
 
   /// Finds the mother of an MC particle by looking for the expected PDG code in the mother chain.
-  /// \param particlesMC  table with MC particles
```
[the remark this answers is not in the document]
fgrosa: Fixed!

Common/Core/RecoDecay.h
```diff
@@ -584,12 +583,15 @@ class RecoDecay
...
+    while (particleMother.has_mothers()) {
       if (depthMax > -1 && -stage >= depthMax) { // Maximum depth has been reached.
         return -1;
       }
-      auto indexMotherTmp = particleMother.mother0Id();
-      particleMother = particlesMC.iteratorAt(indexMotherTmp);
+      auto indexMotherTmp = particleMother.mothersIds().front();
+      if (indexMotherTmp <= -1) {
+        return -1;
+      }
```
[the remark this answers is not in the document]
fgrosa: I added it for my debugging, but indeed it seems not necessary, removed

Common/Core/RecoDecay.h
```diff
@@ -614,7 +616,6 @@ class RecoDecay
   }
 
   /// Gets the complete list of indices of final-state daughters of an MC particle.
-  /// \param particlesMC  table with MC particles
```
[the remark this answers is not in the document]
fgrosa: Fixed!

Common/Core/RecoDecay.h
```diff
@@ -741,17 +725,15 @@ class RecoDecay
...
-        if (indexDaughterFirst <= -1 && indexDaughterLast <= -1) {
-          //Printf("MC Rec: Rejected: bad daughter index range: %d-%d", indexDaughterFirst, indexDaughterLast);
+        if (!particleMother.has_daughters()) {
+          //Printf("MC Rec: Rejected: bad daughter index range: %d-%d", particleMother.daughtersIds().front(), particleMother.daughtersIds().back());
           return -1;
         }
         // Check that the number of direct daughters is not larger than the number of expected final daughters.
-        if (indexDaughterFirst > -1 && indexDaughterLast > -1 && indexDaughterLast - indexDaughterFirst + 1 > int(N)) {
-          //Printf("MC Rec: Rejected: too many direct daughters: %d (expected %ld final)", indexDaughterLast - indexDaughterFirst + 1, N);
+        if (particleMother.daughtersIds().back() - particleMother.daughtersIds().front() + 1 > int(N)) {
```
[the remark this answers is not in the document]
fgrosa: `particleMother.daughtersIds().size()` is always 2 by definition because it's only first one and last one (suggested by @aalkin)

Common/Core/RecoDecay.h
```diff
@@ -862,16 +844,15 @@ class RecoDecay
...
+      if (!candidate.has_daughters()) {
+        //Printf("MC Gen: Rejected: bad daughter index range: %d-%d", candidate.daughtersIds().front(), candidate.daughtersIds().back());
         return false;
       }
       // Check that the number of direct daughters is not larger than the number of expected final daughters.
-      if (indexDaughterFirst > -1 && indexDaughterLast > -1 && indexDaughterLast - indexDaughterFirst + 1 > int(N)) {
-        //Printf("MC Gen: Rejected: too many direct daughters: %d (expected %ld final)", indexDaughterLast - indexDaughterFirst + 1, N);
+
+      if (candidate.daughtersIds().back() - candidate.daughtersIds().front() + 1 > int(N)) {
+        //Printf("MC Gen: Rejected: too many direct daughters: %d (expected %ld final)", candidate.daughtersIds().back()-candidate.daughtersIds().front()+1, N);
```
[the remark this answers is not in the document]
fgrosa: See reply above

fgrosa:
> I agree to move ahead and merge this. Could you subsequently, prepare a PR where you put `mothers_first_as` and `daughters_as` in a single location, such that @aalkin could follow up on it?

Indeed this is a good idea @jgrosseo to keep working on it. I will open a new PR as soon as we merge this one

Common/Core/RecoDecay.h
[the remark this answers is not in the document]
vkucera: OK, then the line below (735) should contain the same expression.

vkucera (approved):

PWGHF/Tasks/HFCorrelatorD0D0bar.cxx
```diff
@@ -436,14 +436,22 @@ struct HfCorrelatorD0D0bar {
...
+                  break;
+                }
+              }
+              for (auto& dau : particle2.daughters_as<MCParticlesPlus>()) {
+                if (std::abs(dau.eta()) > etaCut) {
+                  candidate2DauInAcc = false;
+                  break;
+                }
+              }
+              if (candidate1DauInAcc && candidate2DauInAcc && particle1.pt() > ptCut && particle2.pt() > ptCut) {
```
vkucera, line 454: @fcolamar Can you please check these changes in your tasks and confirm that you are fine with them?

jgrosseo (approved):

PR #511, opened 2022-02-08
Extend efficiency task

njacazio:
- Add trk info + phi eta maps
- Move to a sigle struct per charge


njacazio:
@mario-krueger I moved to a single struct that fills info about all particles, I added also alpha efficiency and eta-phi map that will be useful when checking the acceptance map of the TOF

mario-krueger (approved):

PR #513, opened 2022-02-08
Event selection params moved to CCDB

ekryshen:
* Event selection params moved to CCDB
* Event selection parameters tuned to reproduce Run2 results
* Fix extraction of tracklets from partition in evsel and multiplicity tasks
* PWGMM and PWGUD tasks adjusted accordingly

PR #514, opened 2022-02-08
Fix partition selection for TPC / Tracklets mult in multiplicity table

jezwilkinson:
fixes issue in multiplicity table producer to correctly use cached partitions to count TPC multiplicity & tracklet multiplicity [previously broken by change to partitions not re-calculating in each process() loop]


jezwilkinson:
closing, already covered by https://github.com/AliceO2Group/O2Physics/pull/513

PR #516, opened 2022-02-09
Add TOF event time computation also for tiny tables

njacazio:
@lbariogl here's what you are looking for


njacazio:
@jgrosseo with the PR it will be possible to run the TOF event time also when producing the compressed tables

jgrosseo:
What happens when compressed and normal tables are requested on the same train?

njacazio:
The event time is computed twice

jgrosseo (approved):

PR #524, opened 2022-02-11
[EMCAL-670] Store cluster position in phi from 0 to 2 Pi

mfasDa:
Convention for phi to be compatible
with the EMCAL geometry
implementation which uses 0 - 2 pi
in azimuth.


mfasDa:
Apparently ROOT::Math::PositionVector3D uses the convention -pi to pi for phi (see ROOT::Math::CylindricalEta3D::Restrict), the EMCAL geometry uses the convention 0 to 2 Pi. In order to align them we will store phi in the cluster with the convention 0 - 2 pi.

mfasDa:
Also applied to cells in the cell monitoring histogram now ...

jgrosseo (approved):

PR #532, opened 2022-02-15
Avoid unnecessary checks for track selection

lbariogl:


mario-krueger:
hi,
I would be very surprised if this makes any difference (except for the `isRun2` stuff maybe...).
As far as I understand, an `&&`ed if statement will not be processed further as soon as one condition evaluates to `false`. 
Cheers,
  Mario

lbariogl:
Hi,
in this case it does not male any difference at all.
I thought it evaluated all the conditions for each track.
We can forget about thir PR in this case.

PR #535, opened 2022-02-16
[EMCAL-670] Initialize GeoManager from CCDB

mfasDa:
The GeoManager is needed in order
to calculate the cluster kinematics.
Loading via the GeoManager fails on
HyperLoop due to a missing Geometry
file, instead it needs to be obtained
from the CCDB.


iarsene (approved):

mfasDa:
Running fine, but an include statement was missing - fixed. WIP label removed - ready for merge!

iarsene (approved):

PR #542, opened 2022-02-19
Add momentum selection

lbariogl:


victor-gonzalez (approved):
I guess it is what you want: to keep tracks with Pt minor than Pt_max AND P minor than P_max

PR #549, opened 2022-02-22
Avoid issues with fmt 8.1.1

ktf:


jgrosseo:
What is the issue with %f?

ktf:
The problem is the boolean, I think. On a separate note I would suggest we tell people to use LOGP, rather than LOGF, so that in general they do not have to think about the type.

jgrosseo (approved):

PR #555, opened 2022-02-23
DPG: dont use QAResults.root output file

mario-krueger:


alibuild (approved):

PR #556, opened 2022-02-23
PID as temporary workaround for primaries

pestratm:
To temporarily workaround the problem with isPhysicalPrimary, select the particles directly using PID @mario-krueger @yelmard

### DIFF
--- a/PWGLF/Tasks/spectraCharged.cxx
+++ b/PWGLF/Tasks/spectraCharged.cxx
@@ -221,11 +221,11 @@ bool chargedSpectra::initParticle(const P& particle)
 {
   vars.isChargedPrimary = false;
   auto pdgParticle = pdg->GetParticle(particle.pdgCode());
-  // if (!pdgParticle || pdgParticle->Charge() == 0.) {
-  if (!pdgParticle || TMath::Abs(pdgParticle->Charge()) != 3.) { // This is only a temporary workaround for isPhysicalPrimary and should be replaced by commented line in future
+  if (!pdgParticle || pdgParticle->Charge() == 0.) {
     return false;
   }
-  vars.isChargedPrimary = particle.isPhysicalPrimary();
+  std::list<int> primparticles{11, 13, 211, 321, 2212, 3112, 3222, 3312, 3334};                                                                                           //(Non-exhaustive) list of charged particles that can constitute primaries
+  vars.isChargedPrimary = particle.isPhysicalPrimary() && (std::find(primparticles.begin(), primparticles.end(), TMath::Abs(particle.pdgCode())) != primparticles.end()); // use PID as temporary workaround
   if (std::abs(particle.eta()) >= etaCut) {
     return false;
   }


njacazio (commented):
Hi @pestratm wouldn't it be better to iteratively check the mother? Something like https://github.com/AliceO2Group/DelphesO2/blob/master/examples/aod/createO2tables.h#L643

PWGLF/Tasks/spectraCharged.cxx
```diff
@@ -221,11 +221,11 @@ bool chargedSpectra::initParticle(const P& particle)
 {
   vars.isChargedPrimary = false;
   auto pdgParticle = pdg->GetParticle(particle.pdgCode());
-  // if (!pdgParticle || pdgParticle->Charge() == 0.) {
-  if (!pdgParticle || TMath::Abs(pdgParticle->Charge()) != 3.) { // This is only a temporary workaround for isPhysicalPrimary and should be replaced by commented line in future
+  if (!pdgParticle || pdgParticle->Charge() == 0.) {
```
njacazio, line 224: Here you are skipping neutral and those for which the PDG is not taken and this is fine

PWGLF/Tasks/spectraCharged.cxx
```diff
@@ -221,11 +221,11 @@ bool chargedSpectra::initParticle(const P& particle)
...
     return false;
   }
-  vars.isChargedPrimary = particle.isPhysicalPrimary();
+  std::list<int> primparticles{11, 13, 211, 321, 2212, 3112, 3222, 3312, 3334};                                                                                           //(Non-exhaustive) list of charged particles that can constitute primaries
+  vars.isChargedPrimary = particle.isPhysicalPrimary() && (std::find(primparticles.begin(), primparticles.end(), TMath::Abs(particle.pdgCode())) != primparticles.end()); // use PID as temporary workaround
```
njacazio, line 228: But here you're just checking that the particle PDG is in the list of the particles defined above, right?

njacazio (changes requested):

PWGLF/Tasks/spectraCharged.cxx
pestratm: Exactly. We're interested in the charged primaries and thus have the (pdgParticle->Charge() == 0.) above and the particle.isPhysicalPrimary() here. But since the isPhysicalPrimary is currently not working correctly, I added this workaround so that we exclude particles that cannot be primaries. This second part is supposed to be removed once isPhysicalPrimary is working properly, while the rest will remain here (that's why I changed it back  (pdgParticle->Charge() == 0.). It won't make any difference right now but will once all is working properly).

njacazio: Ok

njacazio (approved):

PWGLF/Tasks/spectraCharged.cxx
pestratm: > Hi @pestratm wouldn't it be better to iteratively check the mother? Something like https://github.com/AliceO2Group/DelphesO2/blob/master/examples/aod/createO2tables.h#L643 I'm not sure that you are selecting primaries like this, unless I misunderstood what you're doing

You're correct, it is not actually selecting primaries. It is more of a quick and "dirty" temporary workaround to only select particles that at least could be primary particles (and thus exclude other particles that are too short-lived to be primaries). I will change it again once isPrimaryParticle is properly working. This is why I only checked if it is one of those particles and not if it actually really is a primary

PR #558, opened 2022-02-23
Extend handling of PID response param

njacazio:
- Add common utility for CCDB handling
- Add possibility to create timestamp based objects based on run number

### DIFF
--- a/Common/Tools/handleParamBase.h
+++ b/Common/Tools/handleParamBase.h
@@ -1,0 +1,118 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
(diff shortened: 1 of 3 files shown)


jgrosseo:
Why start from the run number and not the timestamp?

njacazio:
@jgrosseo 
> Why start from the run number and not the timestamp?
Indeed the run number is just for having an easy handle but the timestamps can still be defined.
There are 2 ways of using it:
a) More general: one can send the object specifying the timestamp of the object
b) One uses the run number to get the timestamp limits and the object is pushed as such. This is when the parametrization is on the run number level (as it is for now)

jgrosseo:
I understand but it may be an additional complication. The idea was that if you have an object defined for an entire run, the validity is defined in CCDB as such and automatically the object is only retrieved once. So I don't think that the analysis code needs to know about the entity of run at all at this point. What do you think?

njacazio:
Ah sure, I fully agree! 
But this is not analysis code, this is just the way that we push PID response objects on the ccdb.
Unless I misunderstood what you are saying the way we fetch CCDB objects in analysis is still completely handled via timestamps and not run numbers. 
This change is transparent for the analysis task

jgrosseo (approved):

jgrosseo:
Thanks. I understand now

PR #559, opened 2022-02-24
PWGHF: Avoid usage of indices in getting MC particle mothers and daughters

fgrosa:
@aalkin @jgrosseo as discussed in https://github.com/AliceO2Group/O2Physics/pull/510 I reopen a PR to investigate the issue with `daughters_as<>` and `mothers_as<>` in `RecoDecay.h`

### DIFF
--- a/Common/Core/RecoDecay.h
+++ b/Common/Core/RecoDecay.h
@@ -562,7 +562,6 @@ class RecoDecay
   }
 
   /// Finds the mother of an MC particle by looking for the expected PDG code in the mother chain.
-  /// \param particlesMC  table with MC particles
   /// \param particle  MC particle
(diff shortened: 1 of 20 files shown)


jgrosseo:
Thanks a lot!
@aalkin does this help to debug or do you also need to know about the specific input files which cause the failures?

aalkin:
This should be enough, so far it seems that this may trigger two separate issues: one is with recursive dereference of the self-index, and the other is with some input files that may have set daughter slice set incorrectly (first index < second index).

benedikt-voelkel:
This should sort daughters in the `McParticle` table during AOD production
The issue is also followed up in O2-2810

PR #563, opened 2022-02-26
Improved handling of vertex and track reconstruction efficiencies

victor-gonzalez:
Also
- possibility of event selection without trigger class
- track selection FB32 and FB64 adapted to Run3


alibuild (approved):

PR #567, opened 2022-02-28
RecoDecay, PWGHF: Add protection for index double-counting

fgrosa:
@jgrosseo @vkucera I noticed that sometimes the indices are overflow (@aalkin told me something similar in a private chat when helping me with the debugging). This causes a segmentation violation (e.g. I debugged downloading locally the same inputs used for this hyperloop train https://alimonitor.cern.ch/train-workdir/tests/0002/00020645/stdout.log). I remember that this could be likely avoided asking for primary particles from a chat with @jgrosseo, but I also think that the method in `RecoDecay.h` should be generic for any usecase (also non primary particles). Let me know if you have different opinions or suggestions for better implementations!

### DIFF
--- a/Common/Core/RecoDecay.h
+++ b/Common/Core/RecoDecay.h
@@ -681,8 +681,7 @@ class RecoDecay
           getDaughters(particlesMC, particlesMC.rawIteratorAt(idxDau), list, arrPDGFinal, depthMax, stage);
         }
       }
-    }
-    else if (particle.daughtersIds().front() == particle.daughtersIds().back() && particle.daughtersIds().front() <= particlesMC.size()) {
+    } else if (particle.daughtersIds().front() == particle.daughtersIds().back() && particle.daughtersIds().front() <= particlesMC.size()) {
       getDaughters(particlesMC, particlesMC.rawIteratorAt(particle.daughtersIds().front()), list, arrPDGFinal, depthMax, stage);
     }
   }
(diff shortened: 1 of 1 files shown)


vkucera:
@jgrosseo Isn't this something that should be fixed directly in the code of `daughtersIds` to return a single daughter when `front == back`?

jgrosseo:
yes, I agree, sure. @aalkin (why) isn't it so?

aalkin:
It should return a slice of size 1, creating a single iteration when using `for(auto& d : daughters)` loop. If that is not the case, there is probably a bug in either slicing or the getter function implementation.

fgrosa:
Hi @vkucera @aalkin @jgrosseo I added a printout of the daughter indices in case of `front == back` in the for loop and I always get two of them as you can see it in the attached log. Isn't this behaviour expected?
[log.txt](https://github.com/AliceO2Group/O2Physics/files/8184343/log.txt)

vkucera (approved):

vkucera:
@jgrosseo Would you agree to merge this now and follow up on the slicing issue separately?

jgrosseo:
Fine with me.

jgrosseo:
Maybe make a JIRA ticket on the issue...

vkucera:
> Fine with me.

@jgrosseo OK, then can you approve it so that it can be merged?

jgrosseo (approved):

fgrosa:
> > Fine with me.
> 
> @jgrosseo OK, then can you approve it so that it can be merged?

Thanks a lot! I opened a JIRA, see here: https://alice.its.cern.ch/jira/browse/O2-2828

PR #571, opened 2022-02-28
PWGEM/Tasks/gammaConversionsmc.cxx Add histos and enhance

sstiefel19:
Add histos, tune binnings, introduce dynamic
retrieval of pointers of  histograms to be filled.
That is, get rid of registry.fill(HIST("compile
timeConstant" calls.

### DIFF
--- a/PWGEM/Tasks/gammaConversionsmc.cxx
+++ b/PWGEM/Tasks/gammaConversionsmc.cxx
@@ -21,6 +21,7 @@
 #include "Common/Core/PID/PIDResponse.h"
 #include "Common/Core/PID/PIDTPC.h"
 
+#include <TH1.h>
 #include <TH1F.h>
 #include <TVector3.h>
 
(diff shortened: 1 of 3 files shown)


PWGEM/Tasks/gammaConversionsmc.cxx
```diff
@@ -74,188 +76,452 @@ struct GammaConversionsmc {
...
   {
-    TAxis* lXaxis = registry.get<TH1>(HIST("IsPhotonSelected"))->GetXaxis();
-    for (size_t i = 0; i < fPhotCutsLabels.size(); ++i) {
-      lXaxis->SetBinLabel(i + 1, fPhotCutsLabels[i]);
+    for (auto bac : std::vector<std::string>{"beforeCuts/", "afterCuts/"}) {
+
+      // reconstructed data histograms
+      {
+        // track histograms
+        {
```
iarsene, line 198: Why are these two extra brackets needed?

sstiefel19: they are not strictly needed, the outer one could directly be removed, the inner when using different names for lPath. But it is good practice to put code that does not have to speak with each other but is still in the enclosing scope in their own scopes for various reasons.

PWGEM/Tasks/gammaConversionsmc.cxx
```diff
@@ -74,188 +76,452 @@ struct GammaConversionsmc {
...
+      }
+      if (lMothers[0] == lMothers[1]) {
+        LOGF(info, "size4 01");
+        lSame = 8;
+      }
+      if (lMothers[2] == lMothers[3]) {
+        LOGF(info, "size4 23");
+        lSame = 9;
+      }
+    }
```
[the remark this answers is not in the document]
sstiefel19: yes this is of course not meant to stay, Ill remove the whole block with my next commit. Its a debugging remainder which is not needed at all. (see comment above  this block)

iarsene (approved):

PR #574, opened 2022-03-01
Add task to check consistency of MC data model in AO2D files

fgrosa:
@jgrosseo I tested it offline with the file that gave me the issue with the HF code (see https://github.com/AliceO2Group/O2Physics/pull/567) and indeed it returns fatal error as expected. The same for other files of the same production (almost all those tested).

### DIFF
--- a/Common/Tasks/checkDataModelMC.cxx
+++ b/Common/Tasks/checkDataModelMC.cxx
@@ -1,0 +1,68 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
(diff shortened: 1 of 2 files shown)


Common/Tasks/checkDataModelMC.cxx
```diff
@@ -0,0 +1,68 @@
...
+using namespace o2::framework::expressions;
+
+/// Check the complete list of indices of final-state daughters of an MC particle.
+/// \param particlesMC  table with MC particles
+/// \param particle  MC particle
+template <typename T>
+void checkDaughters(const T& particlesMC,
+                    const typename T::iterator& particle)
+{
+  if (particle.has_daughters()) {
```
jgrosseo, line 30: In principle this line can be skipped because in that case the list in the next line is simply empty, but this can be changed later.

jgrosseo (approved):

PR #578, opened 2022-03-02
MCHistogram: Add simple analysis on physical primaries

sawenzel:
Motivated by https://alice.its.cern.ch/jira/browse/O2-2808

### DIFF
--- a/Tutorials/src/mcHistograms.cxx
+++ b/Tutorials/src/mcHistograms.cxx
@@ -16,6 +16,7 @@
 #include "Framework/runDataProcessing.h"
 #include "Framework/AnalysisTask.h"
 #include "CommonConstants/MathConstants.h"
+#include "TDatabasePDG.h"
 
 using namespace o2;
 using namespace o2::framework;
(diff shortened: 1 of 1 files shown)


Tutorials/src/mcHistograms.cxx
```diff
@@ -33,6 +35,26 @@ struct VertexDistribution {
...
+  {
+    for (auto& particle : mcParticles) {
+      if (!particle.isPhysicalPrimary()) {
+        continue;
+      }
+      auto pdgParticle = pdgDB.GetParticle(particle.pdgCode());
+      if (!pdgParticle) {
+        continue;
+      }
+      charge->Fill(pdgParticle->Charge());
```
jgrosseo, line 53: @aalkin @ktf @njacazio I remember some more efficient way to get the charge. Am I mistaken?

aalkin: @sawenzel there is a slightly better way to use the PDG database, see here:
https://github.com/AliceO2Group/O2Physics/blob/master/Tutorials/src/usingPDGService.cxx#L20 
Most importantly, note that the charge in PDG is in units of `|e|/3`
```cpp
 Double_t         fCharge;           // charge in units of |e|/3
 ```

@jgrosseo unfortunately, no, we still need to query PDG.

jgrosseo (approved):

PR #579, opened 2022-03-02
Add centrality x-check histo

ddobrigk:


njacazio (approved):

ddobrigk:
The MacOS problem (a timeout in o2sim_G4) does not seem related... so I believe we can force merge (or re-start the test)? Thanks!

PR #583, opened 2022-03-03
OpenData: Added first version of flow task.

mgheata:
This adds a first version of a flow analysis task, used to validate a similar open data demonstrator task.

### DIFF
--- a/Tutorials/OpenData/flowAnalysis.cxx
+++ b/Tutorials/OpenData/flowAnalysis.cxx
@@ -1,0 +1,271 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+///
+/// \brief Flow analysis.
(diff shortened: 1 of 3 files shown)


jgrosseo:
The sub folder needs an own CMakeList.txt file (you can see how it is done for the Skimming folder and do it in the same way)

mgheata:
> The sub folder needs an own CMakeList.txt file (you can see how it is done for the Skimming folder and do it in the same way)

OK, done

Tutorials/OpenData/flowAnalysis.cxx
```diff
@@ -0,0 +1,375 @@
...
+          }
+          case 8: {
+            fillCPt<8>(track, vnC);
+            break;
+          }
+          case 9: {
+            fillCPt<9>(track, vnC);
+            break;
+          }
+        }
```
iarsene, line 316: It is up to you, but here maybe you could consider having the centrality as one additional dimension to the vn histograms, so you could avoid this whole switch statement.

Tutorials/OpenData/flowAnalysis.cxx
```diff
@@ -0,0 +1,375 @@
...
+          }
+          case 8: {
+            fillAPt<8>(track, vnA);
+            break;
+          }
+          case 9: {
+            fillAPt<9>(track, vnA);
+            break;
+          }
+        }
```
iarsene, line 364: same comment as above

Tutorials/OpenData/flowAnalysis.cxx
mgheata: Thanks for the hint @iarsene, it is done like this now.

jgrosseo (approved):

iarsene (approved):

PR #587, opened 2022-03-06
Handling of MC centrality/multiplicity scaling

victor-gonzalez:


victor-gonzalez:
@jgrosseo @ddobrigk  this PR handles the needed scaling for V0M multiplicities for MC productions
The needed information for the `LHC20f6` MC production anchored to `LHC15o` has been updated to my user area in `CCDB`
```
        "ccdburl": "http://ccdb-test.cern.ch:8080",
        "ccdbpath": "Users/v/victor/Centrality/Estimators",
        "genname": "HIJING"
```
and, as said just above, it can be tested by setting the name of the generator to `HIJING`

The implementation in `CCDB` is an additional `TFormula`, with the appropriate parameters, which is incorporated to the `TList` associated to the corresponding run/timestamp in `CCDB`. The `TFormula` is given the name of the generator which is also used to find it on the `TList`

Once we agree on the PR I will upload the updated `LHC15o` `CCDB` info to the production endpoint
After that we will proceed with the rest of the MC productions

Common/TableProducer/centralityTable.cxx
```diff
@@ -183,8 +197,14 @@ struct CentralityTable {
     if (estV0M == 1) {
       float cV0M = 105.0f;
       if (V0MInfo.mCalibrationStored) {
-        float v0m = collision.multV0A() * V0MInfo.mhVtxAmpCorrV0A->GetBinContent(V0MInfo.mhVtxAmpCorrV0A->FindFixBin(collision.posZ())) +
-                    collision.multV0C() * V0MInfo.mhVtxAmpCorrV0C->GetBinContent(V0MInfo.mhVtxAmpCorrV0C->FindFixBin(collision.posZ()));
+        float v0m;
+        if (genName->length() != 0) {
+          v0m = V0MInfo.mMCScale->Eval(collision.multV0M());
```
jgrosseo, line 202: I wonder how efficient this is compared to a C++ formula with some fixed parameters.

jgrosseo:
Thanks a lot! Apart from the question on the speed this looks very good.

Common/TableProducer/centralityTable.cxx
```diff
@@ -183,8 +197,14 @@ struct CentralityTable {
     if (estV0M == 1) {
       float cV0M = 105.0f;
       if (V0MInfo.mCalibrationStored) {
-        float v0m = collision.multV0A() * V0MInfo.mhVtxAmpCorrV0A->GetBinContent(V0MInfo.mhVtxAmpCorrV0A->FindFixBin(collision.posZ())) +
-                    collision.multV0C() * V0MInfo.mhVtxAmpCorrV0C->GetBinContent(V0MInfo.mhVtxAmpCorrV0C->FindFixBin(collision.posZ()));
+        float v0m;
+        if (genName->length() != 0) {
```
jgrosseo, line 201: and probably add a bool for this as well, as length will inspect the string on each collision I believe...

ddobrigk:
Just for completeness: I also struggled with speed and optimisation when writing AliMultSelection within AliPhysics back in the day. In the end, I completely removed all per-event TString operations (they're REALLY very slow) and went with pre-optimized formulae; however, I believe that for the MC mapping the power law-based scaling is as good as it can be (it actually has to be an invertible formula for this to work), so I would have no objection to use a C++ hardcoded formula instead of invoking a TF1::Eval. We are unlikely to need more flexibility, and a linear scaling - necessary for midrapidity estimators - is only a special case of the power-law-based scaling.

Common/TableProducer/centralityTable.cxx
victor-gonzalez: `length` and `size` on std::string are of constant complexity, i.e. they don't go through the string
But as we are not looping over collisions but getting called on each of them it does make sense to have a single memory access instead of two (on looping over collisions compiler optimization will take care of it)
I will use the formula (or what the final implementation will be) pointer as 'the' bool

ddobrigk:
> Wow! C++ formula is twice quicker than `TFormula`! My idea then is to keep `TFormula` on CCDB for its naming and serialization properties but then use its parameters in a conventional C++ formula @jgrosseo @ddobrigk Would you agree?

Fully agreed, and I kind of expected this somehow :-D Please carry on and thanks a lot!!

PR #592, opened 2022-03-08
Update Run3 track selection

njacazio:
@mario-krueger I think this should be needed to ask for real global tracks, what do you think?

### DIFF
--- a/Common/Core/TrackSelection.h
+++ b/Common/Core/TrackSelection.h
@@ -62,8 +62,8 @@ class TrackSelection
         (track.itsNCls() >= mMinNClustersITS) &&
         (track.itsChi2NCl() <= mMaxChi2PerClusterITS) &&
         (track.tpcChi2NCl() <= mMaxChi2PerClusterTPC) &&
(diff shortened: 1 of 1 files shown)


mario-krueger:
good point, thanks!

mario-krueger (approved):

njacazio:
Hi @mario-krueger thanks, I mark it as ready for review

mario-krueger:
On second thought I would suspect however, that it does not change anything for the globalTrack selection, since there are other conditions which require hits in the ITS and clusters in the TPC.

njacazio:
I see. We need to investigate this issue as it seems that we get tracks that pass the global requirement that are ITS only so this triggered the change. I'm on it but probably there is no harm in merging this already

jgrosseo:
Isn't it misleading that this bypasses then the setting `mRequireTPC(ITS)Refit`?

njacazio:
@jgrosseo you are indeed right, I was hasty

jgrosseo (approved):

PR #595, opened 2022-03-08
Propagate momentum to PV

lbariogl:
Add momentum at the primary vertex to `TrackPropagation.h`

It is extremely important for femtoscopy analyses.

I added momentum as `DECLARE_SOA_DYNAMIC_COLUMN`. I wanted to add it as `DECLARE_SOA_EXPRESSION_COLUMN` (because it is used very often and to be consistent with `aod::Tracks`), but I could not because hyperbolic functions are not defined in `Core/Framework/BasicOps.h`.

### DIFF
--- a/Common/DataModel/TrackPropagation.h
+++ b/Common/DataModel/TrackPropagation.h
@@ -42,6 +42,11 @@ DECLARE_SOA_DYNAMIC_COLUMN(Pz, pz, //! z-component of the track momentum GeV/c
                              return pt * std::sinh(eta);
                            });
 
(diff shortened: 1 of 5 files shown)


victor-gonzalez (approved):

lbariogl:
Thanks @jgrosseo. For the time being it is the only way to propagate values to the primary vertex, isn'it? Then I will change it when it will be included in the usual tables.

lbariogl:
@ktf @iarsene are these changes ok?

iarsene (approved):

PR #602, opened 2022-03-09
Add debug functionalities to task for MC data model check task

fgrosa:


fgrosa:
@jgrosseo I am not sure why the tests keep failing: there is no `struct` defined at line 115 (and anyway it's not the one defined at line 90) and if I compile `O2Physics` locally I don't get any error. Any suggestion how to fix it? Should I try to close and reopen the PR?

PR #608, opened 2022-03-11
Speedup compilation

ktf:
It looks like a significant amount of compilation time can be saved by using precompiled headers. 

When compiling the Tutorials we have that before this:

```Bash
Analyzing build trace from 'report.txt'...
**** Time summary:
Compilation (110 times):
Parsing (frontend):         1059.3 s
Codegen & opts (backend):   1214.2 s

```

after:

```cpp
Analyzing build trace from 'report.txt'...
**** Time summary:
Compilation (113 times):
  Parsing (frontend):          660.6 s
  Codegen & opts (backend):   1188.6 s
```

the extra files are of course the precompiled bits and of course the impact is mostly on the frontend, while codegen and optimization stay the same.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ CMakeFiles/
 build
 compile_commands.json
 .DS_Store
+.cache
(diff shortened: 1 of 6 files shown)


jgrosseo:
Very nice!

PR #609, opened 2022-03-11
Fix producer

lbariogl:
Change needed to run with Hyperloop


victor-gonzalez (approved):
You are reverting this now then?

PR #505, opened 2022-02-04
TPC PID Response

AnnalenaKa:
These changes will move the TPC PID Response into one header file TPCPIDResponse.h and remove any headers and dependencies that are not needed anymore.

PR #610, opened 2022-03-14
New TPC PID Response after BugFix

AnnalenaKa:
**New TPC PID Response** 

This replaces the existing TPC PID Response. 

The PR gets back the changes from "TPC PID Response" PR#505 and "Update TPC QA" PR#594 that were reverted in PR#607 due to a problem when loading the Response from the CCDB. The problem was caused by using a pointer (o2::pid::tpc::Response* response = nullptr) in the table producer tasks. Apparently this is only a problem when loading the response from the CCDB instead of using a response from a file.  
In the new Pull request the bug is fixed.

Furthermore, in addition to the changes implemented in "TPC PID Response" PR#505:
-  the run dependent timestamp was implemented for the table producer tasks and in the handleParamTPCResponse.cxx
- the TFormula that is used for the resolution parameterisation in case the default is not used (mUseDefaultResolutionParam=false) is replaced by a c++ function.


AnnalenaKa:
@njacazio 

I have implemented the bug fix to load the response from the CCDB properly and we don't see this strange error message anymore. Furthermore your changes to the tpc+tof qa are included and the run dependent timestamp. The TFormula is also replaced by a c++ Formula. 
Jeremy and me have run all tests on run 2 and run 3 data loading the response from the CCDB with option 0 and -1 and loading it from a file. Everything works as expected and the QA looks good. 

Could you have a look at our modifications and tell us if this is fine or of we should change something? 
Once you are happy with the modifications Jeremy will push the new object to the CCDB and run more tests before merging our changes. Then everything should be fine and no analysis should experience any problems.

Common/TableProducer/PID/pidTPC.cxx
```diff
@@ -110,102 +112,66 @@ struct tpcPid {
...
-    } else { // Loading it from CCDB
+    const TString fname = paramfile.value;
+    if (fname != "") { // Loading the parametrization from file
+      LOGP(info, "Loading TPC response from file {}", fname);
+      try {
+        std::unique_ptr<TFile> f(TFile::Open(fname, "READ"));
+        f->GetObject("Response", responseptr);
+        response.SetParameters(responseptr);
+      } catch (...) {
+        LOGP(info, "Loading the TPC PID Response from file {} failed!", fname);
```
njacazio, line 123: Maybe this should be a fatal

Common/TableProducer/PID/pidTPC.cxx
```diff
@@ -110,102 +112,66 @@ struct tpcPid {
...
-      if (flag.value != 1) {
-        return;
+    // Check and fill enabled tables
+    auto makeTable = [&tracks, &collisions, this](const Configurable<int>& flag, auto& table, const o2::track::PID::ID pid) {
+      if (flag.value == 1) {
+        // Prepare memory for enabled tables
+        table.reserve(tracks.size());
+        int lastCollisionId = -1;        // Last collision ID analysed
+        for (auto const& trk : tracks) { // Loop on Tracks
+          auto collision = collisions.iteratorAt(trk.collisionId());
```
njacazio, line 150: This should be called after we are sure that the track has a collision

Common/TableProducer/PID/pidTPC.cxx
```diff
@@ -110,102 +112,66 @@ struct tpcPid {
...
-    constexpr auto responseTr = ResponseImplementation<PID::Triton>();
-    constexpr auto responseHe = ResponseImplementation<PID::Helium3>();
-    constexpr auto responseAl = ResponseImplementation<PID::Alpha>();
-
-    auto reserveTable = [&tracks](const Configurable<int>& flag, auto& table) {
-      if (flag.value != 1) {
-        return;
+    // Check and fill enabled tables
+    auto makeTable = [&tracks, &collisions, this](const Configurable<int>& flag, auto& table, const o2::track::PID::ID pid) {
+      if (flag.value == 1) {
```
njacazio, line 145: you could consider the opposite e.g.:
```
if (flag.value != 1)
return;
```

Common/TableProducer/PID/pidTPC.cxx
```diff
@@ -110,102 +112,66 @@ struct tpcPid {
...
+        // Prepare memory for enabled tables
+        table.reserve(tracks.size());
+        int lastCollisionId = -1;        // Last collision ID analysed
+        for (auto const& trk : tracks) { // Loop on Tracks
+          auto collision = collisions.iteratorAt(trk.collisionId());
+          if (useCCDBParam && ccdbTimestamp.value == 0 && trk.has_collision() && trk.collisionId() != lastCollisionId) { // Updating parametrization only if the initial timestamp is 0
+            lastCollisionId = trk.collisionId();
+            const auto& bc = collision.bc_as<aod::BCsWithTimestamps>();
+            response.SetParameters(ccdb->getForTimeStamp<o2::pid::tpc::Response>(ccdbPath.value, bc.timestamp()));
```
njacazio, line 154: Unless I misunderstood doing so checks for the ccdb up to 9 times for all tracks, this might be a bit expensive and is probably not needed as in the previous implementation

Common/Tools/CMakeLists.txt
```diff
@@ -1,3 +1,4 @@
+
```
njacazio, line 1: This change is not needed

Common/Tools/handleParamTPCResponse.cxx
```diff
@@ -0,0 +1,300 @@
...
+  std::map<std::string, std::string> metadata, headers;
+  o2::ccdb::CcdbApi api;
+  if (optMode.compare("push") == 0 || optMode.compare("pull") == 0) { // Initialise CCDB if in push/pull mode
+    api.init(urlCCDB);
+    if (!api.isHostReachable()) {
+      LOG(warning) << "CCDB mode (push/pull) enabled but host " << urlCCDB << " is unreachable.";
+      return 1;
+    }
+    // Init timestamps
+    if (runnumber != 0) {
```
njacazio, line 148: This looks like inspired from https://github.com/AliceO2Group/O2Physics/blob/master/Common/Tools/handleParamBase.h#L82 is there something wrong/missing with it that it needed to be reimplemented?

Common/Tools/handleParamTPCResponse.cxx
```diff
@@ -0,0 +1,300 @@
...
+
+    if (optMode.compare("push") == 0) {
+      LOG(info) << "Attempting to push object to CCDB";
+
+      if (optDelete) {
+        api.truncate(pathCCDB);
+      }
+
+      if (runnumber != 0) {
+        metadata["runnumber"] = Form("%i", runnumber);
```
njacazio, line 255: Same comment here, this looks like taken from https://github.com/AliceO2Group/O2Physics/blob/master/Common/Tools/handleParamBase.h#L64 is there something wrong/missing that it had to be reimplemented?

njacazio (commented):
Hi @AnnalenaKa great, I see the things we discussed implemented.
Concerning the ccdb access it might be better to access it only once.
Also please make an effort to either recycle the same piece of code in headers or extend the features so that they are available to everyone as it makes it easier for future development

Common/Tools/handleParamTPCResponse.cxx
```diff
@@ -0,0 +1,300 @@
...
+/// \brief exec for writing and reading Response object
+
+#include "CCDB/CcdbApi.h"
+#include <boost/program_options.hpp>
+#include <FairLogger.h>
+#include <array>
+#include <fstream>
+#include <sstream>
+#include "TFile.h"
+#include "Common/Core/PID/PIDResponse.h"
```
njacazio, line 24: Is this header useful?

Common/TableProducer/PID/pidTPC.cxx
AnnalenaKa: This is turned around now.

Common/TableProducer/PID/pidTPC.cxx
AnnalenaKa: I changed this, so that this part is done before the make Table where it would be called 9 times.

Common/Tools/CMakeLists.txt
AnnalenaKa: This is not meant to be there and removed now.

Common/Tools/handleParamTPCResponse.cxx
AnnalenaKa: The header is not used and removed.

AnnalenaKa:
Hi @njacazio,
thank you very much for looking into the changes and your suggestions. I implemented all your suggestions. Indeed the code from the handleParamBase was repeated. This is removed now. 
The ccdb should be accessed only once now with the new implementation. I hope this change is what you meant and that this makes the task faster.

njacazio:
@AnnalenaKa thank you for the fast action, I think we can put this into the ready stage!

njacazio:
@jgrosseo with this PR @AnnalenaKa included the changes from the last one (https://github.com/AliceO2Group/O2Physics/pull/505) + the run dependent calibration. This is now also tested with the calibration object taken from the CCDB.
If you agree we can merge this to have it in production

jgrosseo (approved):

jgrosseo:
Thanks @AnnalenaKa 
@njacazio Can you sync the PID service wagons once the tag of tonight appears, please?

njacazio:
@jgrosseo sure thing, I will take care of it

PR #617, opened 2022-03-15
Initialise the memory used for VLA branches

mpuccio:
@jgrosseo 🤞 (this time after 5 times I always get consistent results :D )


jgrosseo:
@ktf @TimoWilken is the CI stuck?

mpuccio:
It should, but for the mother is not relevant because only those used are read, so the unused index does not affect them

ktf:
@TimoWilken could you check what is going on with the CI?

PR #623, opened 2022-03-16
fddConverter from table 000 to 001

mbroz84:


jgrosseo:
The process of the eventselectionQA needs to subscribe to FDDs_001 (because we need to keep FDDs_000 as default in the first step otherwise the CI of O2 does not go through).
The rest looks good. Thank you!

jgrosseo (dismissed):

jgrosseo (approved):

jgrosseo:
This will fail until https://github.com/AliceO2Group/AliceO2/pull/8349 is in a tag and then auto-merge...

jgrosseo:
Disabling auto-merge for now because merging this will require adding the fdd converter in many workflows and needs explanation on the mattermost channel. So better we do this after the Easter break...

PR #624, opened 2022-03-17
PID ML: update producer task

saganatt:
- update the producer task with multiplicity and centrality info
- fix to work with older MC (MCParticles_000)


alibuild (approved):

PR #631, opened 2022-03-19
fix daughters indexes saving into reducedMCTrack

galocco:
mctrack.daughtersIds() contains only the first and the last index of the decay daughters. Therefore the size of the vector daughters was always <= 2. At the moment of filling daughterRange, all but the first two decay daughters were lost.


iarsene (approved):

iarsene:
@galocco  Thanks for spotting this!

PR #632, opened 2022-03-19
Calo Tender implemented

peressounko:
First implementation of Run2  PHOS Tender and EMCAL AnalysisFramework.
Table CaloCluster to be discussed/agreed with EMCAL and probably replace table PWGJE/DataModel/EMCALClusters.h

### DIFF
--- a/Common/DataModel/CaloClusters.h
+++ b/Common/DataModel/CaloClusters.h
@@ -1,0 +1,59 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+#ifndef O2_ANALYSIS_CALOCLUSTERS_H_
+#define O2_ANALYSIS_CALOCLUSTERS_H_
+
+#include "Framework/AnalysisDataModel.h"
+
+namespace o2::aod
+{
+namespace calocluster
+{
+// Columns to store momenta of "photons"
+DECLARE_SOA_INDEX_COLUMN(Collision, collision); //!
+DECLARE_SOA_COLUMN(Type, caloType, uint8_t);    //! calo type: 0: PHOS, 1: EMCAL
+DECLARE_SOA_COLUMN(Px, px, float);              //! momenta components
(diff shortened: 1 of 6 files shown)


peressounko:
Dear @mfasDa, @raymondEhlers 
please have a look at class CaloClusters.h and caloClusterProducer.cxx. I think we should develop common table for PHOS and EMCAL. Also it would be nice to have one producer class.
      Best regards, Dmitri

iarsene:
@mfasDa , @raymondEhlers 
I will wait for your go ahead on this.

mfasDa:
Adding @fjonasALICE who takes over responsibility over the emcal-correction-task from Raymond. For dynamic SOA we can probably agree on a common table for EMCAL and PHOS, this has worked for run2, we need to agree here that the relevant information is available for both detectors. For the producer I am not so convinced, the correction framework was done on purpose, there is a chain of corrections applied for EMCAL (clusterization, non-lin- correction, exotics rejection, hadronic correction), which to some extend requires different customisations for different objects (i.e. stronger E-cuts for photons and electrons vs. jets but no hadronic correction), so I think we might want to keep the correction framework.

peressounko:
Dear @mfasDa, all. I propose to implement minimal bricks: clusterization (can be common for PHOS and EMCAL)  in caloClusterProducer, track matching (not implemented yet, but also can be common) and configurable corrections suitable for analysis (has to be separate). All elementary bricks can be combined by some envelope class to minimize customization for final user.  Best regards, Dmitri

mfasDa:
This structure we had in the early version of the correction chain, the chain was very complicated to configure, macros with a lot of free parameters (more than 20), in order to allow for studies to tweak parameters. This setup turned out to be very complicated and error-prone for the users. Therefore decided for the new implementation of the correction task with one task to rule all components, and configuration based on a modern markup language that allows us to provide a default configuration and the user just to change the params that actually need to be changed. This improved the stability a lot. I am not in favour of going back to the old approach.

peressounko:
Dear @mfasDa, I do not see any contradiction between small classes dedicated to specific tasks which finally will be hidden in one user-available task with default full framework and list of default parameters - as you described. 
Let us then concentrate on common CaloCluster table: do we have all necessary fields implemented? Can we use current design as starting point, see how suitable it will be in real analysis and add missing fields later if necessary?

peressounko:
Dear  @fjonasALICE, @mfasDa, do you have any comment on proposed table? If not, can approve design as is and commit changes?

peressounko:
Dear @iarsene, @mfasDa, @raymondEhlers, I think we can commit draft as is and return to discussion if EMCAL will want to change something.

mfasDa:
As this is not a persistent structure it should be find to merge, however for EMCAL will probably add a couple of fields.

peressounko:
@mfasDa , adding fields is completely fine. Removing/renaming will cause modification of dependencies. Hope, give naming is acceptable for you.

iarsene (approved):

PR #633, opened 2022-03-21
Fix selection of tracks with TOF

njacazio:


PWGCF/FemtoDream/FemtoDreamTrackSelection.h
```diff
@@ -344,63 +344,18 @@ void FemtoDreamTrackSelection::init(HistogramRegistry* registry, const std::stri
...
-      return track.tpcNSigmaPr();
-      break;
-    case o2::track::PID::Deuteron:
-      return track.tpcNSigmaDe();
-      break;
-    default:
-      return 999.f;
-      break;
-  }
+  return o2::aod::pidutils::tpcNSigma(pid, track);
```
njacazio, line 347: This is not fully equivalent as in this case a Nsigma is returned even for IDs greater than deuterons

victor-gonzalez (approved):

PR #635, opened 2022-03-21
Remove TOF Signal from common user space

njacazio:
The TOF signal itself (especially in Run3) cannot be used alone to achieve PID, I propose therefore to keep it as a utility table and have it handled only by PID specific tasks. Of course it can always be computed also in specific tasks but I don't know if this will be a common use case.
Please have a look at the propose code modifications, there should not be significant changes apart in the ML section.
Thanks a lot!

### DIFF
--- a/ALICE3/TableProducer/alice3-pidTOF.cxx
+++ b/ALICE3/TableProducer/alice3-pidTOF.cxx
@@ -42,36 +42,6 @@ void customize(std::vector<o2::framework::ConfigParamSpec>& workflowOptions)
 
 #include "Framework/runDataProcessing.h"
 
-struct ALICE3tofSignal { /// Task that produces the TOF signal from the trackTime
-  Produces<o2::aod::TOFSignal> table;
-  bool enableTable = false;
-
-  void init(o2::framework::InitContext& initContext)
-  {
-    // Checking the tables are requested in the workflow and enabling them
-    auto& workflows = initContext.services().get<RunningWorkflowInfo const>();
-    for (DeviceSpec device : workflows.devices) {
-      for (auto input : device.inputs) {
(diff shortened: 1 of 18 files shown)


njacazio:
@akalweit  @saganatt @victor-gonzalez I think this is especially relevant

saganatt:
Hm, for PID ML TOF signal is one of the basic variables, next to the TPC signal. Cannot we still keep it there?

njacazio:
@saganatt yes, indeed it is used but without the event time I don't know how much it can help in the PID as it would be unbound unlike the TPC signal that can be used directly for PID

njacazio:
Before this can be merged we need confirmation from @saganatt if needed I can put this into draft PR

njacazio:
@iarsene thanks a lot for having a look, I think that the only non transparent changes are in the development of @saganatt Of course @akalweit and @victor-gonzalez please complain if this is not the case.

iarsene:
I think its still fine to leave it like this, unless @saganatt  Maja thinks she need more time.

iarsene:
@njacazio , @saganatt 
Hi, is there a solution for how we can proceed with this PR ? If not yet, lets then make it a draft please.

njacazio:
@iarsene apologies for the delay. 
Here we are with a working version which is completely transparent for the user (so far). 
**However** I realized that with this implementation the usage of dynamic columns that will try to access the expected TOF (e.g. `tofExpSignalEl`) value will fail because it will not find the tofSignal to subtract to the delta between the tofSignal and the expected signal.

Essentially if we merge this PR the expected times cannot be taken directly from the table but have to be recomputed again.
I don't know if this is then something we want.

PR #639, opened 2022-03-23
PID: Add table for TOF PID flags

njacazio:
- Now the user can access the information of the T0 during the analysis
- Update spectra tasks to subscribe to TOF flags

### DIFF
--- a/Common/Core/PID/PIDResponse.h
+++ b/Common/Core/PID/PIDResponse.h
@@ -662,6 +662,22 @@ const auto tpcExpSignalDiff(const o2::track::PID::ID index, const TrackType& tra
 
 } // namespace pidutils
 
+namespace pidflags
+{
+
+namespace enums
+{
+enum PIDFlags : uint8_t {
+  T0TOF = 0x1
+};
+}
+DECLARE_SOA_COLUMN(TOFFlags, tofFlags, uint8_t);         //! Flag for the complementary PID information
+DECLARE_SOA_COLUMN(TOFFlagsFull, tofFlagsFull, uint8_t); //! Flag for the complementary PID information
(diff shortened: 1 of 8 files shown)


iarsene (dismissed):

iarsene:
@ktf @jgrosseo 
This PR seems to still require an approval. Could one of you approve?

njacazio:
@iarsene thanks a lot for taking care of this, indeed it would be good if @jgrosseo shares the approach.
This is in view of the inclusion of the Ev. time from the FT0 that will need more flags to be introduced

iarsene:
> @iarsene thanks a lot for taking care of this, indeed it would be good if @jgrosseo shares the approach. This is in view of the inclusion of the Ev. time from the FT0 that will need more flags to be introduced

Hi @njacazio, i agree that we need this information, which existed also in Run1&2 if i remember correctly, but will be much more important in Run3&4.

Common/Core/PID/PIDResponse.h
```diff
@@ -662,6 +662,22 @@ const auto tpcExpSignalDiff(const o2::track::PID::ID index, const TrackType& tra
...
+{
+
+namespace enums
+{
+enum PIDFlags : uint8_t {
+  T0TOF = 0x1
+};
+}
+DECLARE_SOA_COLUMN(TOFFlags, tofFlags, uint8_t);         //! Flag for the complementary PID information
+DECLARE_SOA_COLUMN(TOFFlagsFull, tofFlagsFull, uint8_t); //! Flag for the complementary PID information
```
jgrosseo, line 675: The flags are different for the two cases (tiny and full)?

njacazio:
@jgrosseo this now creates only one table with the flag, at the time that the event time is computed. This is equivalent to the previous code but with only one table

jgrosseo (approved):

PR #640, opened 2022-03-23
PWGDQ: Adding a DileptonLepton process function for Bc analysis

lucamicheletti93:
The runDileptonLepton process function is added to the dqEfficiency task for the Bc analysis on MC

### DIFF
--- a/PWGDQ/Core/VarManager.h
+++ b/PWGDQ/Core/VarManager.h
@@ -379,7 +379,7 @@ class VarManager : public TObject
   template <int pairType, uint32_t collFillMap, uint32_t fillMap, typename C, typename T>
   static void FillPairVertexing(C const& collision, T const& t1, T const& t2, float* values = nullptr);
   template <uint32_t collFillMap, uint32_t fillMap, typename C, typename T1, typename T2, typename T3>
-  static void FillJpsiLepton(C const& collision, T1 const& lepton1, T2 const& lepton2, T3 const& lepton3, float* values = nullptr);
+  static void FillDleptonLepton(C const& collision, T1 const& lepton1, T2 const& lepton2, T3 const& lepton3, float* values = nullptr);
   template <typename T1, typename T2>
   static void FillDileptonHadron(T1 const& dilepton, T2 const& hadron, float* values = nullptr, float hadronMass = 0.0f);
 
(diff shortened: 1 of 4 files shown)


PWGDQ/Core/VarManager.h
```diff
@@ -1027,6 +1051,106 @@ void VarManager::FillPairVertexing(C const& collision, T const& t1, T const& t2,
   }
 }
 
+template <uint32_t collFillMap, uint32_t fillMap, typename C, typename T1, typename T2, typename T3>
+void VarManager::FillDileptonLepton(C const& collision, T1 const& lepton1, T2 const& lepton2, T3 const& lepton3, float* values)
```
iarsene, line 1055: I think we have to generalize a bit this function to be used by both muon, barrel or muon-barrel analyses, similar to the FillPairVertexing() function.
I am not sure why you need to have all the T1, T2 and T3 typenames. At the moment you use just a single type (all 3 particles are muons). I think that to make this general, you could us just two types, T1 and T2, where T1 is for the tracks that make up your dilepton (either track or fwdtrack), and T2 can be the third track, which can be independently from T1, either a fwd of or a barrel track.
I suggest to also change the name of the function, to something more general since the same functionality is needed also for J/psi + K. So i suggest to change to FillDileptonTrackVertexing maybe?

PWGDQ/Core/VarManager.h
```diff
@@ -1027,6 +1051,106 @@ void VarManager::FillPairVertexing(C const& collision, T const& t1, T const& t2,
...
+  constexpr bool eventHasVtxCov = ((collFillMap & Collision) > 0 || (collFillMap & ReducedEventVtxCov) > 0);
+  constexpr bool trackHasCov = ((fillMap & TrackCov) > 0 || (fillMap & ReducedTrackBarrelCov) > 0);
+  if (!values) {
+    values = fgValues;
+  }
+
+  float m1 = fgkMuonMass;
+  float m2 = fgkMuonMass;
+  float m3 = fgkMuonMass;
+  float mlepton = fgkMuonMass;
```
iarsene, line 1067: this part needs to be generalized, on the lines of the suggestion i made for the general scope of this function

PWGDQ/Tasks/dqEfficiency.cxx
```diff
@@ -726,6 +735,114 @@ struct AnalysisSameEventPairing {
     } // end loop over barrel track pairs
   }   // end runPairing
 
+  // Template function to run pair - lepton combinations
+  template <uint32_t TEventFillMap, uint32_t TTrackFillMap, typename TEvent, typename TTracks, typename TEventsMC, typename TTracksMC>
+  void runDileptonLepton(TEvent const& event, TTracks const& tracks, TEventsMC const& eventsMC, TTracksMC const& tracksMC)
```
iarsene, line 740: Having this function in the AnalysisSameEventPairing is breaking a bit the analysis model we used in the data analysis task (tableReader). It is also duplicating code since you are repeating the loop to build dilepton candidates which are produced in runPairing. So since we recently had an update in runPairing to produce a table which contains the indices of the lepton legs, its natural to make a separate task that subscribes to this table. You would then just need to loop over the dileptons in that table and over the selected tracks for your third leg. I think this is logically more simple, and becomes identical to the DileptonHadron task we have in tableReader.
We need also to generalize this part, since there are already ongoing barrel analyses that use this analysis topology.

PR #642, opened 2022-03-23
[EventFiltering, PWGHF] Add possibility to apply ML models to all the charm hadron species and add cut for D* soft pion

fgrosa:


fgrosa:
@vkucera the modification in the HF code is minimal (only a namespace for a `LabeledArray` for BDT selections), could you have a look and see whether it is fine for you? Thanks!

vkucera (approved):

PR #646, opened 2022-03-24
DPG: add protection for tracks without particle

mario-krueger:


alibuild (dismissed):

mario-krueger:
hi @TimoWilken,
is there a way to trigger self approval again after updating a PR?
Seems like this does not restart automatically.

alibuild (approved):

mario-krueger:
hi,
ah nice thanks. This was the trick I was looking for :)

PR #647, opened 2022-03-25
[DPG] updates in ITS bitmap tests for qaImpPar.cxx and qaEventTracks.cxx

mfaggin:


DPG/Tasks/qaEventTrack.cxx
```diff
@@ -342,11 +342,16 @@ void qaEventTrack::processReco(const C& collision, const T& tracks)
...
       }
     }
+    bool trkHasITS = false;
     for (unsigned int i = 0; i < 7; i++) {
       if (track.itsClusterMap() & (1 << i)) {
+        trkHasITS = true;
         histos.fill(HIST("Tracks/ITS/itsHits"), i, itsNhits);
       }
     }
+    if (!trkHasITS) {
```
mario-krueger, line 352: would not if(!itsNhits) do the same without the need for the additional variable 'trkHasITS'?

mfaggin: Maybe yes, indeed. But maybe is it worthwhile to have this new bool as well to check some possible pathologic cases? Not sure if it can happen that nITShits is non null but the hitmap is not filled (it shall never happen, but maybe worthwhile to have this crosscheck as well?)

mario-krueger (approved):

PR #650, opened 2022-03-29
First version of Nucleus Tree Creator

fbellini:


PWGLF/TableProducer/LFTreeCreatorNuclei.cxx
```diff
@@ -0,0 +1,177 @@
...
+    // Filling event properties
+    rowCandidateFullEvents.reserve(collision.size());
+    rowCandidateFullEvents(
+      collision.bcId(),
+      collision.numContrib(),
+      collision.posX(),
+      collision.posY(),
+      collision.posZ(),
+      0,  // iseventselected //to be set
+      1); // runnumber //to be sets
```
njacazio, line 142: this you can probably set it via: `collision.bc().runNumber()`

PWGLF/TableProducer/LFTreeCreatorNuclei.cxx
```diff
@@ -0,0 +1,177 @@
...
+  using TrackCandidates = soa::Filtered<soa::Join<aod::Tracks, aod::TracksExtra, aod::TracksExtended, aod::TrackSelection,
+                                                  aod::pidTPCFullPi, aod::pidTOFFullPi,
+                                                  aod::pidTPCFullKa, aod::pidTOFFullKa,
+                                                  aod::pidTPCFullPr, aod::pidTOFFullPr,
+                                                  aod::pidTPCFullDe, aod::pidTOFFullDe,
+                                                  aod::pidTPCFullHe, aod::pidTOFFullHe>>;
+
+  void process(soa::Filtered<soa::Join<aod::Collisions,
+                                       aod::EvSels>>::iterator const& collision,
+               TrackCandidates const& tracks)
```
njacazio, line 131: Here you could subscribe to BCs  `aod::BC const&` to have the info about the run number later

PWGLF/TableProducer/LFTreeCreatorNuclei.cxx
```diff
@@ -0,0 +1,177 @@
...
+  {
+    // Filling event properties
+    rowCandidateFullEvents.reserve(collision.size());
+    rowCandidateFullEvents(
+      collision.bcId(),
+      collision.numContrib(),
+      collision.posX(),
+      collision.posY(),
+      collision.posZ(),
+      0,  // iseventselected //to be set
```
njacazio, line 141: Since this is supposed to be working on Run3 data we can start adding the ev. sel. statement e.g. `collision.sel8()`

njacazio (commented):
@fbellini great, thanks a lot! I'm still wondering if TableProducer is the best place for this task as it just provides filtering for the data and it is not used by other tasks.
What do you think, shall we move it to tasks?

PWGLF/TableProducer/LFTreeCreatorNuclei.cxx
fbellini: Hi @njacazio, I have set runnumber as suggested. 
On the folder to store the tree creator, I have chosen TableProducer as it is done similarly in HF.
We can move it to tasks, if you believe it is a better place.

PWGLF/TableProducer/LFTreeCreatorNuclei.cxx
fbellini: See above.

PWGLF/TableProducer/LFTreeCreatorNuclei.cxx
fbellini: Done.

njacazio (dismissed):

njacazio:
We need still to fix the formatting before merging

PWGLF/TableProducer/LFTreeCreatorNuclei.cxx
iarsene: Hi @njacazio and @fbellini,
I would say that this task can stay in TableProducer since it indeed produces skims. However, for these skims to be used properly in a consumer task, one has to move the data model definition to a separate header, which can be visible to more than just this task. I guess this can be the topic of a future PR. 
I also see that this PR is modifying the aodMerger.cxx. Is this really intended ? Then maybe @jgrosseo should have a look.

jgrosseo: Thanks for flagging. This must be the result of an invalid merge. @fbellini please check before making the PR if you touch unexpected files. Github nicely shows the changes and one can scroll through...

jgrosseo: I think it sneaked in later when you pushed 
image

fbellini: @jgrosseo, @iarsene hi! Thank you for spotting this mistake and sorry for messing up with my previous push. It was due to a misconfiguration of my fork. Hopefully my settings are ok now and it should not happen again. 
To answer @iarsene, yes, I can provide the separate header in a future PR. Thanks for the suggestion.

njacazio (approved):
Perfect, this PR is now, clean! Thanks a lot!

njacazio (approved):

PR #653, opened 2022-03-30
fix FT0 MC QA

AllaMaevskaya:


AllaMaevskaya:
What is wrong here? Can it be merged?

ktf:
Merging this, it needed a review by someone at top-level because apparently Common/Tasks is apparently not in the CODEOWNERS file. @pzhristov  @iarsene @jgrosseo is that intended?

jgrosseo:
Yes, Common should be approved by @iarsene or me

jgrosseo:
Sorry that we both overlooked it

iarsene:
Sorry for the delay. Please ping me over mattermost if i don't react to the PR.

PR #654, opened 2022-03-31
PWGHF: Lc tree creator process switch for data and adjust range histos Lc task

AnnalenaKa:
This modification adds a process switch in the Lc tree creator task. By default it is set to true to process MC. When the switch is set to false the tables are filled for data processing. 
In the Lc task the histogram ranges were adjusted to match the distributions.


AnnalenaKa:
Hi @strogolo,
here are the proposed changes for the Lc tree creator. I added a process switch that allows to select whether data or MC is processed. Like this it is possible fill the tables with the candidate properties and write them to a tree. If more or other variables should be included in the tree we can still add them.
In the Lc task I adjusted the histogram ranges to match the distributions a bit better.

PWGHF/TableProducer/HFTreeCreatorLcToPKPi.cxx
```diff
@@ -307,6 +307,116 @@ struct CandidateTreeWriter {
...
+            candidate.impactParameter2(),
+            candidate.errorImpactParameter0(),
+            candidate.errorImpactParameter1(),
+            candidate.errorImpactParameter2(),
+            candidate.index0_as<aod::BigTracksPID>().tpcNSigmaPi(),
+            candidate.index0_as<aod::BigTracksPID>().tpcNSigmaKa(),
+            candidate.index0_as<aod::BigTracksPID>().tpcNSigmaPr(),
+            candidate.index0_as<aod::BigTracksPID>().tofNSigmaPi(),
+            candidate.index0_as<aod::BigTracksPID>().tofNSigmaKa(),
+            candidate.index0_as<aod::BigTracksPID>().tofNSigmaPr(),
```
vkucera, line 386: Avoid multiple calls of `index0_as`. It is expensive.

vkucera (changes requested):

PWGHF/TableProducer/HFTreeCreatorLcToPKPi.cxx
AnnalenaKa: Dear Vit,
thanks for pointing this out. I replaced this by calling it once before the table is filled with the properties.

PWGHF/TableProducer/HFTreeCreatorLcToPKPi.cxx
```diff
@@ -307,6 +310,119 @@ struct CandidateTreeWriter {
...
+      auto fillTable = [&](int CandFlag,
+                           int FunctionSelection,
+                           float FunctionInvMass,
+                           float FunctionCt,
+                           float FunctionY,
+                           float FunctionE) {
+        if (FunctionSelection >= 1) {
+          rowCandidateFull(
+            candidate.index0_as<aod::BigTracksPID>().collision().bcId(),
+            candidate.index0_as<aod::BigTracksPID>().collision().numContrib(),
```
vkucera, line 348: Same here

vkucera (changes requested):

vkucera (approved):

PR #660, opened 2022-04-05
Fix for string view in cutculator

saganatt:
@lbariogl

### DIFF
--- a/PWGCF/FemtoDream/FemtoDreamTrackSelection.h
+++ b/PWGCF/FemtoDream/FemtoDreamTrackSelection.h
@@ -160,7 +160,8 @@ class FemtoDreamTrackSelection : public FemtoDreamObjectSelection<float, femtoDr
   static int findSelectionIndex(const std::string_view& obs, std::string_view prefix = "")
   {
     for (int index = 0; index < kNtrackSelection; index++) {
-      std::string_view cmp{static_cast<std::string>(prefix) + static_cast<std::string>(mSelectionNames[index])};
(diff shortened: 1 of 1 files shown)


PWGCF/FemtoDream/FemtoDreamTrackSelection.h
```diff
@@ -160,7 +160,8 @@ class FemtoDreamTrackSelection : public FemtoDreamObjectSelection<float, femtoDr
   static int findSelectionIndex(const std::string_view& obs, std::string_view prefix = "")
   {
     for (int index = 0; index < kNtrackSelection; index++) {
-      std::string_view cmp{static_cast<std::string>(prefix) + static_cast<std::string>(mSelectionNames[index])};
+      std::string comp = static_cast<std::string>(prefix) + static_cast<std::string>(mSelectionNames[index]);
+      std::string_view cmp{comp};
```
victor-gonzalez, line 164: Why not in a single line?

victor-gonzalez (approved):

PWGCF/FemtoDream/FemtoDreamTrackSelection.h
saganatt: I think there is some memory management issue. For some reason, an one-liner results in cmp having spurious content. We need this explicit std::string comp.

victor-gonzalez: I see
Interesting but also frightening!

PR #661, opened 2022-04-05
[EMCAL-670] Map calos and collision to BCId, use BC ID as index column for the clusters

mfasDa:
Calos and Collision need to be mapped to BCId to use it as index column. The process function should also accept BCs without collision, taking (0,0,0) as vertex location for the cluster kinematics, to support stand-alone partitions. The index column of the cluster needs to be BCId instead of collision.

In addition the message about the number of cells found by the correction task need to be changed to LOG(debug), too
verbose on hyperloop.

### DIFF
--- a/PWGJE/DataModel/EMCALClusters.h
+++ b/PWGJE/DataModel/EMCALClusters.h
@@ -22,7 +22,7 @@ namespace o2::aod
 {
 namespace emcalcluster
 {
-DECLARE_SOA_INDEX_COLUMN(Collision, collision);                        //!
+DECLARE_SOA_INDEX_COLUMN(BC, bc);                                      //!
 DECLARE_SOA_COLUMN(ID, id, int);                                       //!
 DECLARE_SOA_COLUMN(Energy, energy, float);                             //!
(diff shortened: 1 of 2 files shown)


jgrosseo (approved):

jgrosseo:
This is fine with me. Is there any chance to do it overall without doing the intermediate list `mEmcalCells` ? Ideally all this would be zero-copy, except filling of the new table of course.

mfasDa:
@jgrosseo Thanks for merging!

Yes, we have sketched a way to bypass the copy. The issue is that the clusterizer uses a gsl::span of o2::emcal::cells as input as it is typically done for processors, while o2::aod::calos gives us an iterator. Furthermore the interface is different. We were thinking to define interators on iterators used as wrappers implementing certain common functions for both types to generalize the interface and then pass iterators instead of gsl::span to the clusterizer and cluster factory (for kinematics and shower shape). We did a similar trick to generalize the fitter interface in O2 to stl arrays and boost histograms. Will however need a bit of refactoring as these classes are already in use.

PR #662, opened 2022-04-06
Switched from startOrbit to orbitReset

ekryshen:


jgrosseo:
Thanks. This is fine with me. Is this filled already for all productions we have on hyperloop?

ekryshen:
> Thanks. This is fine with me. Is this filled already for all productions we have on hyperloop?

Yes, it should be already filled for all productions.

jgrosseo (approved):

PR #667, opened 2022-04-10
Modifications for adding FDD as mult selection signal

ddobrigk:
@ekryshen this is a modification that also touches the event selection classes - I hope that's alright. The point is to add FDD as an extra signal for Run 3 multiplicity selection (Note that one could save AD signals in the same slots in Run 2 converted data). Please feel free to complain, suggest changes etc. Thanks!!

@ktf @pzhristov @jgrosseo this is a small first step that's also related to [our discussion tomorrow](https://indico.cern.ch/event/1149591/).

### DIFF
--- a/Common/DataModel/EventSelection.h
+++ b/Common/DataModel/EventSelection.h
@@ -66,20 +66,21 @@ DECLARE_SOA_COLUMN(Sel8, sel8, bool);                                       //! 
 DECLARE_SOA_INDEX_COLUMN_FULL(FoundBC, foundBC, int, BCs, "_foundBC");      //! BC entry index in BCs table (-1 if doesn't exist)
 DECLARE_SOA_INDEX_COLUMN_FULL(FoundFT0, foundFT0, int, FT0s, "_foundFT0");  //! FT0 entry index in FT0s table (-1 if doesn't exist)
 DECLARE_SOA_INDEX_COLUMN_FULL(FoundFV0, foundFV0, int, FV0As, "_foundFV0"); //! FV0 entry index in FV0As table (-1 if doesn't exist)
+DECLARE_SOA_INDEX_COLUMN_FULL(FoundFDD, foundFDD, int, FDDs, "_foundFDD");  //! FDD entry index in FDDs table (-1 if doesn't exist)
 } // namespace evsel
 DECLARE_SOA_TABLE(EvSels, "AOD", "EVSEL", //!
                   evsel::Alias, evsel::Selection,
                   evsel::BBV0A, evsel::BBV0C, evsel::BGV0A, evsel::BGV0C,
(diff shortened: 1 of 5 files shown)


ekryshen:
Hi @ddobrigk,
The changes are fine with me. However note that there is another PR by Michal where he introduced changes in FDD data model. There are many places where FDD_001 is used:
https://github.com/AliceO2Group/O2Physics/pull/623
@jgrosseo can provide more details.

Also I am not sure if FDD behaviour in the pilot run and MCs is well understood. At least timing distributions look strange so be careful.

ddobrigk:
Hi @ekryshen, about the FDD signal, that's fine - I think for now we will be mostly doing early studies and code preparation. About the FDD_001: since I am not specific in the code that I wrote, I guess we always take the latest version and therefore everything should be fine, right? Maybe @jgrosseo can confirm? Thanks!

jgrosseo:
Can we hold this for a moment while we do the FDD change?

ddobrigk:
Hi @jgrosseo sure, sure. I think I can run this even privately with a local AO2D file to get output that I will then use as an example for the calibrator class in the meantime. I will keep an eye on both PRs then :-)

jgrosseo:
Thanks. As soon as https://github.com/AliceO2Group/O2Physics/pull/623/ is in, you can adjust your code to use FDDs_001 (or you wait 2 more days until FDDs_001 is the default. This may be the easiest). I change this to draft so that we do not merge it by mistake

ddobrigk:
@ekryshen @jgrosseo - Happy Easter! 
This isn't urgent, so please don't let me bug you, but whenever you have a moment, I think this can now move on. This week I will also move ahead with the centrality in Run 3 so more action might come shortly :-D Thanks a lot!

jgrosseo:
@ddobrigk The chance hasn't been done. One needs to merge https://github.com/AliceO2Group/O2Physics/pull/623 and then change the default of FDD to point to FDD_001
You can check if your new code works when you use FDD_001 instead of FDD.
I am away until Friday and if no one else takes the lead, I am reluctant to merge https://github.com/AliceO2Group/O2Physics/pull/623 now (because who then replies to issues etc)...

ddobrigk:
@jgrosseo sorry - I checked earlier today and somehow mistakenly thought #623 was merged. But never mind: I will keep on working with a private copy and do some tests and later on we merge!

ddobrigk:
I would like to close this request and then replace it with a much more complete one for review soon, as the code has already changed and it is better to do it in one single go. I will likely report on this next Monday - we can discuss then!

jgrosseo:
@ddobrigk  The changes to FDD are now merged and FDD_001 is also default in O2 dev. So if you update everything, you can "simply" build your code against the FDD table (no version tag needed)

ddobrigk:
Perfect! Thanks, @jgrosseo : I will update to dev then.

PR #668, opened 2022-04-11
[RecoDecay, Event Filtering, PWGHF] Revert to daughter indices using explicitly offset

fgrosa:
@jgrosseo @vkucera @aalkin I am reverting this again to the indices (using the table offset explicitly for grouped tables) because of the error that I still get in analysing the MC data with @aalkin offline. In case it's helpful for debug, I can open a second PR as last time with the implementation that uses slices, let me know.
I also ping @AnnalenaKa and @strogolo who asked me about this issue last week, this PR should fix it.


vkucera (approved):
Thanks, @fgrosa

jgrosseo (approved):

PR #670, opened 2022-04-12
[O2-2896] When a tree was already merge, don't merge it again.

chiarazampolli:
This assumes that the first tree that is retrieved is always the one with the
highest cycle (--> the latest). E.g. in case of:

   KEY: TTree	O2mcparticle_001;2	O2mcparticle_001 [current cycle]
   KEY: TTree	O2mcparticle_001;1	O2mcparticle_001 [backup cycle]

this tree will be merged only 1x (otherwise it is merged on itself).

### DIFF
--- a/Common/Core/aodMerger.cxx
+++ b/Common/Core/aodMerger.cxx
@@ -209,7 +209,7 @@ int main(int argc, char* argv[])
             } else {
               // key is a duplicate, let's remove it
               treeList->Remove(kj);
-	      j--;
+              j--;
             }
           } else {
             // we changed key, since they are sorted, we won't have the same anymore
--- a/Common/Core/aodMerger.cxx
+++ b/Common/Core/aodMerger.cxx
@@ -18,6 +18,7 @@
 #include "TFile.h"
 #include "TTree.h"
 #include "TList.h"
+#include "THashList.h"
 #include "TDirectory.h"
 #include "TObjString.h"
 #include <TGrid.h>
@@ -197,10 +198,16 @@ int main(int argc, char* argv[])
 
       for (auto key2 : *treeList) {
         auto treeName = ((TObjString*)key2)->GetString().Data();
+        printf("    Processing tree %s\n", treeName);
+        bool found = (std::find(foundTrees.begin(), foundTrees.end(), treeName) != foundTrees.end());
+        if (found == true) {
+          printf("    Tree %s was already merged, skipping\n", treeName);
+          continue;
+        }
         foundTrees.push_back(treeName);
 
         auto inputTree = (TTree*)inputFile->Get(Form("%s/%s", dfName, treeName));
-        printf("    Processing tree %s with %lld entries\n", treeName, inputTree->GetEntries());
+        printf("    Tree %s has %lld entries\n", treeName, inputTree->GetEntries());
 
         if (trees.count(treeName) == 0) {
           if (mergedDFs > 1) {
--- a/Common/Core/aodMerger.cxx
+++ b/Common/Core/aodMerger.cxx
@@ -18,7 +18,7 @@
 #include "TFile.h"
 #include "TTree.h"
 #include "TList.h"
-#include "THashList.h"
+#include "TKey.h"
 #include "TDirectory.h"
 #include "TObjString.h"
 #include <TGrid.h>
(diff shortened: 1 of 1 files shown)


ktf:
Maybe you could sort the list of trees to merge so that the later cycles come first and then filter duplicates out via std::unique (which removes all but the first occurrence in a set of consecutive duplicates).

ktf:
Maybe you could sort the list of trees to merge so that the later cycles come first and then filter duplicates out via std::unique (which removes all but the first occurrence in a set of consecutive duplicates).

chiarazampolli:
I don't think it is needed, my understanding is that the latest cycle is the one taken as default, which is also why it was merged 2x: the same tree was found, but the same one was taken twice.

chiarazampolli:
@ktf , I just committed the best purging that I can think we can do - the std::unique cannot be used because we work with a TList. 

Let me know. 

As discussed at the DPG meeting, actually this merging is used at job level to merge TF's output in MC, so the problem is a blocker (even if the files I tested were all healthy).

chiarazampolli:
Hello @ktf , @pzhristov 

Do you have further comments? I know that @jgrosseo  is away. 

Chiara

Common/Core/aodMerger.cxx
```diff
@@ -193,14 +194,42 @@ int main(int argc, char* argv[])
...
+      }
+
       std::list<std::string> foundTrees;
 
       for (auto key2 : *treeList) {
         auto treeName = ((TObjString*)key2)->GetString().Data();
+        printf("    Processing tree %s\n", treeName);
+        bool found = (std::find(foundTrees.begin(), foundTrees.end(), treeName) != foundTrees.end());
+        if (found == true) {
+          printf("    ***WARNING*** Tree %s was already merged (even if we purged duplicated trees before, so this should not happen), skipping\n", treeName);
```
jgrosseo, line 226: I would propose to make a FATAL error here if this shouldn't happen

Common/Core/aodMerger.cxx
```diff
@@ -193,14 +194,42 @@ int main(int argc, char* argv[])
...
+      for (auto i = 0; i < treeList->GetEntries(); ++i) {
+        TKey* ki = (TKey*)treeList->At(i);
+        for (int j = i + 1; j < treeList->GetEntries(); ++j) {
+          TKey* kj = (TKey*)treeList->At(j);
+          if (std::strcmp(ki->GetName(), kj->GetName()) == 0 && std::strcmp(ki->GetTitle(), kj->GetTitle()) == 0) {
+            if (ki->GetCycle() < kj->GetCycle()) {
+              printf("    ***WARNING*** we had ordered the keys, first cycle should be higher, let's continue anyway");
+            } else {
+              // key is a duplicate, let's remove it
+              treeList->Remove(kj);
```
jgrosseo, line 210: Given that you manipulate the list in-place while you iterate it, don't you need to add `j--` here to make sure the subsequent entry (which becomes the current on removal) is checked as well?

jgrosseo:
Thanks for fixing this @chiarazampolli 
@ktf why do we have several cycles on a tree in O2? We write in bulk in the aod writer, so cannot this be avoided?

Common/Core/aodMerger.cxx
chiarazampolli: Yes, of course, thanks for spotting.

Common/Core/aodMerger.cxx
chiarazampolli: Ok, done.

chiarazampolli:
I am recompiling O2, cannot fix right now the complaint by clang-format. 
But I actually do not understand it, I tried emacs, vi, pico, all show the lines correctly aligned, see screenshot:
image
@TimoWilken , @ktf , any idea?

TimoWilken:
Hi @chiarazampolli, the issue isn't necessarily the alignment, it's that some of your code is indented with tab characters instead of spaces. Those are the lines clang-format is complaining about.

jgrosseo:
Tried to fix it on github.com without success. Sorry...

chiarazampolli:
Thanks @TimoWilken , @jgrosseo . 
In O2, we have indications of the commands to run when the tabs/spaces are not ok. O2Physics doesn't?

Chiara

jgrosseo (approved):

TimoWilken:
@chiarazampolli I've added that check in #676, so hopefully it'll be available soon.

chiarazampolli:
Thanks @TimoWilken !

PR #671, opened 2022-04-13
PWGDQ: Adding dilepton-track process function for dqEfficiency

lucamicheletti93:
Adding dilepton-track process function for dqEfficiency

### DIFF
--- a/PWGDQ/Core/VarManager.h
+++ b/PWGDQ/Core/VarManager.h
@@ -248,6 +248,7 @@ class VarManager : public TObject
     kPairType,
     kVertexingLxy,
     kVertexingLxyErr,
+    kVertexingPseudoCTau,
     kVertexingLxyz,
     kVertexingLxyzErr,
     kVertexingLz,
@@ -262,10 +263,17 @@ class VarManager : public TObject
     kQuadDCAabsXY,
     kQuadDCAsigXY,
     kNPairVariables,
+    kCosPointingAngle,
+    kImpParXYJpsi,
+    kImpParXYK,
+    kDCATrackProd,
+    kDCATrackVtxProd,
 
     // Candidate-track correlation variables
     kPairMass,
+    kPairMassDau,
     kPairPt,
+    kPairPtDau,
     kPairEta,
     kPairPhi,
     kDeltaEta,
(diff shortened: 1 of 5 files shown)


PWGDQ/Core/VarManager.h
```diff
@@ -1029,6 +1041,144 @@ void VarManager::FillPairVertexing(C const& collision, T const& t1, T const& t2,
...
+    o2::track::TrackParCov pars3{track.x(), track.alpha(), lepton3pars, lepton3covs};
+    procCode = VarManager::fgFitterThreeProngBarrel.process(pars1, pars2, pars3);
+    procCodeJpsi = VarManager::fgFitterTwoProngBarrel.process(pars1, pars2);
+  } else {
+    return;
+  }
+
+  ROOT::Math::PtEtaPhiMVector v1(lepton1.pt(), lepton1.eta(), lepton1.phi(), mtrack);
+  ROOT::Math::PtEtaPhiMVector v2(lepton2.pt(), lepton2.eta(), lepton2.phi(), mtrack);
+  ROOT::Math::PtEtaPhiMVector v3(track.pt(), track.eta(), track.phi(), mtrack);
```
iarsene, line 1129: This vector assumes that the track is an electron or a muon, but if we want to use this function for channels like B+ -> J/psi + K+ then we need to have a way of transmitting this to the function. This can be transmitted via the template argument you call now "pairType". This could be generalized to sometihng like "candidateType" and you can have some specific entries added to the "PairCandidateType" enumeration. For example, you could have there, in addition to what is there now, kBcToThreeMuons (your case for Bc to three muons), kBtoJpsiEEK (B+ -> jpsi + K+), etc. So based on these we can properly deduce the mass assumption for all three prongs using constexpr.

PWGDQ/Core/VarManager.h
```diff
@@ -1029,6 +1041,144 @@ void VarManager::FillPairVertexing(C const& collision, T const& t1, T const& t2,
...
+    values[VarManager::kVertexingLzErr] = -999.;
+
+    values[VarManager::kVertexingTauxy] = -999.;
+    values[VarManager::kVertexingTauz] = -999.;
+    values[VarManager::kVertexingTauxyErr] = -999.;
+    values[VarManager::kVertexingTauzErr] = -999.;
+    return;
+  }
+
+  Vec3D secondaryVertex;
```
iarsene, line 1159: The secondary vertex does not seem to be computed anywhere, this is just zero i guess. So the rest of the quantities are not valid at the moment, right? You would probably need to also get the secondary vertex from the two leptons, because we could compute a few quantities from that and used them for selections.
You need also to add the uncertainties on the L values, similar to what is done in the two prong approach.

PWGDQ/Tasks/dqEfficiency.cxx
```diff
@@ -835,13 +836,210 @@ struct AnalysisSameEventPairing {
...
+
+      int indexLepton1 = dilepton.index0Id();
+      int indexLepton2 = dilepton.index1Id();
+
+      if (indexOffset == -999) {
+        indexOffset = indexLepton1;
+      }
+
+      auto lepton1 = tracks.iteratorAt(indexLepton1 - indexOffset);
+      auto lepton2 = tracks.iteratorAt(indexLepton2 - indexOffset);
```
iarsene, line 952: I am not sure yet what the index offset is needed for, but you should add a note here to be removed later on after issues are solved.

PWGDQ/Tasks/dqEfficiency.cxx
```diff
@@ -835,13 +836,210 @@ struct AnalysisSameEventPairing {
...
+      std::unique_ptr<TObjArray> objArray(configCutNamesStr.Tokenize(","));
+      fNHadronCutBit = objArray->GetEntries();
+    } else {
+      fNHadronCutBit = 0;
+    }
+  }
+
+  // Template function to run pair - track combinations
+  template <int TPairType, uint32_t TEventFillMap, uint32_t TEventMCFillMap, uint32_t TTrackFillMap, typename TEvent, typename TTracks, typename TEventsMC, typename TTracksMC>
+  void runDileptonLepton(TEvent const& event, TTracks const& tracks, soa::Join<aod::Dileptons, aod::DileptonsExtra> const& dileptons, TEventsMC const& eventsMC, TTracksMC const& tracksMC)
```
iarsene, line 933: Since for now there is nothing specific here, i would call this function "runDileptonTrack" since ideally we will use it for most 3 prong cases.

iarsene:
Hi Luca,
The modifications you add look on the lines we discussed, but i still have some comments. Also, it seems to me that the secondary vertex position for the triplet or dilepton candidates are not computed at all in the VarManager. I am adding Giacomo, @galocco , so he can also have a look and maybe contribute to this since he is developing the jpsi + K channel. His changes on the VarManager need to go in the same place and i think he has already in his code compute the secondary vertices.

PWGDQ/Core/VarManager.h
```diff
@@ -1029,6 +1041,144 @@ void VarManager::FillPairVertexing(C const& collision, T const& t1, T const& t2,
...
+  //auto lepton2 = track.iteratorAt(dilepton.index1());
+
+  //auto lepton1 = track.rawIteratorAt(dilepton.index0Id());
+  //auto lepton2 = track.rawIteratorAt(dilepton.index1Id());
+
+  //auto lepton1 = dilepton.index0Id();
+  //auto lepton2 = dilepton.index1Id();
+
+  //auto lepton1 = track.rawIteratorAt(dilepton.index0Id());
+  //auto lepton2 = track.rawIteratorAt(dilepton.index1Id());
```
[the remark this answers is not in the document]
lucamicheletti93: Removed!

PWGDQ/Core/VarManager.h
lucamicheletti93: Fixed with the new commit

PWGDQ/Tasks/dqEfficiency.cxx
lucamicheletti93: The index is needed becouse otherwise the interatorAt method does not recover the correct leptons from the track table. I would say that is just a temporary fix to solve the issue. I have added a further comment in the code

PWGDQ/Tasks/dqEfficiency.cxx
lucamicheletti93: Done!

PWGDQ/Core/VarManager.h
lucamicheletti93: Implemented with the new commit!

PWGDQ/Tasks/dqEfficiency.cxx
```diff
@@ -835,13 +836,210 @@ struct AnalysisSameEventPairing {
...
+        }
+      }
+    }
+  }
+
+  void processDimuonMuonSkimmed(soa::Filtered<MyEventsVtxCovSelected>::iterator const& event, MyMuonTracksSelectedWithCov const& tracks, soa::Join<aod::Dileptons, aod::DileptonsExtra> const& dileptons, ReducedMCEvents const& eventsMC, ReducedMCTracks const& tracksMC)
+  {
+    VarManager::ResetValues(0, VarManager::kNVars);
+    VarManager::FillEvent<gkEventFillMap>(event);
+    VarManager::FillEvent<gkMCEventFillMap>(event.reducedMCevent());
```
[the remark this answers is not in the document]
lucamicheletti93: Updated!

iarsene (approved):

PR #679, opened 2022-04-14
Add MegaLinter

vkucera:
@TimoWilken @ktf @jgrosseo 

This adds [MegaLinter](https://megalinter.github.io/latest/) as a CI action to lint and format code other than C++ and CMake.

Fixes and formatting have been applied to the existing files to pass the tests, with the exception of `Scripts/update_ccdb.py` which requires some more than trivial fixes which I leave to the next editor.

The action is set to check only modified files with `VALIDATE_ALL_CODEBASE: false`.

If a PR passes the tests without errors but formatting is still required, a formatting commit is committed directly to the PR branch. This is set with `APPLY_FIXES_MODE: commit`. (The option to create a PR instead has some problems with GitHub tokens AFAIK.)


davidrohr:
I don't really like committing to peoples branches directly. I'd investigate the option to create a PR instead, and otherwise just drop it.

vkucera:
> I don't really like committing to peoples branches directly. I'd investigate the option to create a PR instead, and otherwise just drop it.

Hi @davidrohr , the advantage is that the PR author does not need to care about reading instructions on how to run formatting tools or to approve a PR in their fork repo. What disadvantage do you see there?

davidrohr:
I don't like to concept of pushing in other people's repository. Then I'll just mark my PRs as not editable by maintainers...

davidrohr:
Also, this adds a commit on top, which should normally be squashed, but the automatic commit cannot squash properly if the PR consists of multiple commits, and also this would require a force push to the user's repository, which I also think we should not do.

vkucera:
> Also, this adds a commit on top, which should normally be squashed, but the automatic commit cannot squash properly if the PR consists of multiple commits, and also this would require a force push to the user's repository, which I also think we should not do.

Squashing the automatic formatting commit with other commits in the PR works fine.

jgrosseo:
The clang formatter makes a PR which the user can import or not. I think this would be a good solution also here. What's speak against this solution?

vkucera:
> The clang formatter makes a PR which the user can import or not. I think this would be a good solution also here. What's speak against this solution?

Does it? I cannot find a single PR created by the formatting action.

TimoWilken:
> Does it? I cannot find a single PR created by the formatting action.

These PRs are created in the *PR author's* fork. For example: https://github.com/vkucera/AliceO2/pull/10. This is done by the following code:

https://github.com/AliceO2Group/AliceO2/blob/9ed3947a4493517a97275fbe1346a6651915371a/.github/workflows/code-formatting.yml#L87-L97

vkucera:
> > Does it? I cannot find a single PR created by the formatting action.
> 
> These PRs are created in the _PR author's_ fork. For example: [vkucera/AliceO2#10](https://github.com/vkucera/AliceO2/pull/10). That is done by [this code](https://github.com/AliceO2Group/AliceO2/blob/9ed3947a4493517a97275fbe1346a6651915371a/.github/workflows/code-formatting.yml#L87-L97).

Thanks @TimoWilken , I am aware of this feature. My point is that I cannot find an example of it working in O2Physics.
If I look through [the runs of the action](https://github.com/AliceO2Group/O2Physics/actions/workflows/code-formatting.yml) where it failed, I don't see any that would create a PR in the author's branch.
For example: https://github.com/AliceO2Group/O2Physics/pull/650 clearly needed a formatting commit which was made and pushed by the author. The corresponding [run of clang-format](https://github.com/AliceO2Group/O2Physics/runs/5757092140?check_suite_focus=true#step:5:177) does not seem to have made any PR in [the concerned branch](https://github.com/fbellini/O2Physics/pulls) though.
I only see errors of this kind:
```
 ! [remote rejected] HEAD -> alibot-cleanup-670 (refusing to allow a Personal Access Token to create or update workflow `.github/workflows/code-formatting.yml` without `workflow` scope)
error: failed to push some refs to 'https://github.com/alibuild/O2Physics.git'
```

TimoWilken:
Ah, right. The action had the wrong permissions -- that should be fixed now.

vkucera:
> Ah, right. The action had the wrong permissions -- that should be fixed now.

Thanks! I will try to put that in this one.

.github/workflows/mega-linter.yml
```diff
@@ -0,0 +1,81 @@
...
+          fi
+          git push -f "$REMOTE_URL" "$CLEANUP_COMMIT:refs/heads/alibot-cleanup-ml-$PR_NUMBER"
+
+      - name: Create pull request with applied fixes
+        uses: alisw/pull-request@v2
+        if: steps.ml.outputs.has_updated_sources == 1 && ${{ github.event.repository.owner.login == 'AliceO2Group' }}
+        with:
+          source_branch: 'alibuild:alibot-cleanup-ml-${{ env.PR_NUMBER }}'
+          destination_branch: '${{ github.event.pull_request.head.label }}'
+          github_token: ${{ secrets.alibuild_github_token }}
```
TimoWilken, line 64: Use `secrets.ALIBUILD_GITHUB_TOKEN`; see above.

.github/workflows/mega-linter.yml
```diff
@@ -0,0 +1,81 @@
...
+          pr_title: >-
+            [MegaLinter] Apply linters automatic fixes to
+            #${{ env.PR_NUMBER }}
+          pr_body: >-
+            This PR cannot be merged as is. You should either run MegaLinter
+            yourself and update the pull request, or merge this PR in yours.
+
+            You can find how to run MegaLinter locally at
+            <https://megalinter.github.io/latest/mega-linter-runner/>.
```
TimoWilken, line 73: This phrasing isn't ideal in the other workflows either, but I find this a little confusingly worded. Maybe:

```
Your PR ${{ github.event.repository.full_name }}#${{ env.PR_NUMBER }}
cannot be merged as is.
...
```

.github/workflows/mega-linter.yml
```diff
@@ -0,0 +1,81 @@
...
+        continue-on-error: true  # We do not create PRs if the branch is not there.
+
+      - name: Exit with error if the PR is not clean
+        run: |
+          case "${{ steps.ml.outputs.has_updated_sources }}" in
+            0) echo 'PR clean' ; exit 0 ;;
+            1) echo 'PR not clean' ; exit 1 ;;
+          esac
```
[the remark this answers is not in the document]
vkucera: That is correct. The formatting PR should be created only if MegaLinter did not find any errors, which I think is the intended behaviour, right?

vkucera:
> Looks good to me. If the current codebase (with changes in this PR) passes MegaLinter, then let's merge and try it out!

Thanks @TimoWilken

jgrosseo:
> Looks good to me. If the current codebase (with changes in this PR) passes MegaLinter, then let's merge and try it out!

How do we check this? Let me know, I can then merge it.

vkucera:
@TimoWilken Is the action supposed to run already on this PR itself?
The original version which was set to run on `push` and `pull_request` did run but now it is set to run on `pull_request_target` only.

vkucera:
> @TimoWilken Is the action supposed to run already on this PR itself? The original version which was set to run on `push` and `pull_request` did run but now it is set to run on `pull_request_target` only.

Otherwise, the current code passes all the tests except for a few bandit issues which should be addressed at the next modification of `Scripts/update_ccdb.py `.

TimoWilken:
Seeing as the new linter checks passed before, let's merge it and see what happens.

TimoWilken:
FYI: I had to add one more thing (92b5340d379d8b8dd1524eb654fff9a9e30a3f36) to make the automatic cleanup PRs work. Apart from that, looks like it runs OK.

vkucera:
> FYI: I had to add one more thing ([92b5340](https://github.com/AliceO2Group/O2Physics/commit/92b5340d379d8b8dd1524eb654fff9a9e30a3f36)) to make the automatic cleanup PRs work. Apart from that, looks like it runs OK.

Thanks @TimoWilken . I'm glad it seems to work now.

PR #691, opened 2022-04-20
DPG: extend track selection

njacazio:
- Extend track selection table
- Uniform track selection class
- Produce output correlation tree

### DIFF
--- a/Common/Core/TrackSelection.h
+++ b/Common/Core/TrackSelection.h
@@ -52,26 +52,83 @@ class TrackSelection
   template <typename T>
   bool IsSelected(T const& track)
   {
-    const bool isRun2 = track.trackType() == o2::aod::track::Run2Track || track.trackType() == o2::aod::track::Run2Tracklet;
-    if (track.trackType() == mTrackType &&
-        track.pt() >= mMinPt && track.pt() <= mMaxPt &&
-        track.eta() >= mMinEta && track.eta() <= mMaxEta &&
-        track.tpcNClsFound() >= mMinNClustersTPC &&
-        track.tpcNClsCrossedRows() >= mMinNCrossedRowsTPC &&
-        track.tpcCrossedRowsOverFindableCls() >= mMinNCrossedRowsOverFindableClustersTPC &&
-        (track.itsNCls() >= mMinNClustersITS) &&
-        (track.itsChi2NCl() <= mMaxChi2PerClusterITS) &&
-        (track.tpcChi2NCl() <= mMaxChi2PerClusterTPC) &&
-        (mRequireITSRefit ? (isRun2 ? (track.flags() & o2::aod::track::ITSrefit) : track.hasITS()) : true) &&
-        (mRequireTPCRefit ? (isRun2 ? (track.flags() & o2::aod::track::TPCrefit) : track.hasTPC()) : true) &&
-        ((isRun2 && mRequireGoldenChi2) ? (track.flags() & o2::aod::track::GoldenChi2) : true) &&
-        FulfillsITSHitRequirements(track.itsClusterMap()) &&
-        abs(track.dcaXY()) <= ((mMaxDcaXYPtDep) ? mMaxDcaXYPtDep(track.pt()) : mMaxDcaXY) &&
-        abs(track.dcaZ()) <= mMaxDcaZ) {
-      return true;
-    } else {
+    if (!IsSelected(track, TrackCuts::kTrackType)) {
       return false;
     }
+    if (!IsSelected(track, TrackCuts::kPtRange)) {
+      return false;
+    }
+    if (!IsSelected(track, TrackCuts::kEtaRange)) {
+      return false;
+    }
+    if (!IsSelected(track, TrackCuts::kTPCNCls)) {
+      return false;
+    }
+    if (!IsSelected(track, TrackCuts::kTPCCrossedRows)) {
+      return false;
+    }
+    if (!IsSelected(track, TrackCuts::kTPCCrossedRowsOverNCls)) {
+      return false;
+    }
+    if (!IsSelected(track, TrackCuts::kTPCChi2NDF)) {
+      return false;
+    }
+    if (!IsSelected(track, TrackCuts::kTPCRefit)) {
+      return false;
+    }
+    if (!IsSelected(track, TrackCuts::kITSNCls)) {
+      return false;
+    }
+    if (!IsSelected(track, TrackCuts::kITSChi2NDF)) {
+      return false;
+    }
+    if (!IsSelected(track, TrackCuts::kITSRefit)) {
+      return false;
+    }
+    if (!IsSelected(track, TrackCuts::kITSHits)) {
+      return false;
+    }
+    if (!IsSelected(track, TrackCuts::kGoldenChi2)) {
+      return false;
+    }
+    if (!IsSelected(track, TrackCuts::kDCAxy)) {
+      return false;
+    }
+    if (!IsSelected(track, TrackCuts::kDCAz)) {
+      return false;
+    }
+    return true;
+  }
+
+  // Temporary function to check if track passes and return a flag. To be replaced by framework filters.
+  template <typename T>
+  uint16_t IsSelectedFlag(T const& track)
+  {
+    uint16_t flag = 0;
+
+    auto setFlag = [&](const TrackCuts& cut) {
+      if (IsSelected(track, cut)) {
+        flag |= 1UL << static_cast<int>(cut);
+      }
+    };
+
+    setFlag(TrackCuts::kTrackType);
+    setFlag(TrackCuts::kPtRange);
+    setFlag(TrackCuts::kEtaRange);
+    setFlag(TrackCuts::kTPCNCls);
+    setFlag(TrackCuts::kTPCCrossedRows);
+    setFlag(TrackCuts::kTPCCrossedRowsOverNCls);
+    setFlag(TrackCuts::kTPCChi2NDF);
+    setFlag(TrackCuts::kTPCRefit);
+    setFlag(TrackCuts::kITSNCls);
+    setFlag(TrackCuts::kITSChi2NDF);
+    setFlag(TrackCuts::kITSRefit);
+    setFlag(TrackCuts::kITSHits);
+    setFlag(TrackCuts::kGoldenChi2);
+    setFlag(TrackCuts::kDCAxy);
+    setFlag(TrackCuts::kDCAz);
+
+    return flag;
   }
 
   // Temporary function to check if track passes a given selection criteria. To be replaced by framework filters.
@@ -112,7 +169,7 @@ class TrackSelection
         return track.itsChi2NCl() <= mMaxChi2PerClusterITS;
 
       case TrackCuts::kITSRefit:
-        return (isRun2 && mRequireITSRefit) ? (track.flags() & o2::aod::track::ITSrefit) : true;
+        return (mRequireITSRefit ? (isRun2 ? (track.flags() & o2::aod::track::ITSrefit) : track.hasITS()) : true);
 
       case TrackCuts::kITSHits:
         return FulfillsITSHitRequirements(track.itsClusterMap());
--- a/Common/Core/TrackSelection.h
+++ b/Common/Core/TrackSelection.h
@@ -102,7 +102,7 @@ class TrackSelection
 
   // Temporary function to check if track passes and return a flag. To be replaced by framework filters.
   template <typename T>
-  uint16_t IsSelectedFlag(T const& track)
+  uint16_t IsSelectedMask(T const& track)
   {
     uint16_t flag = 0;
 
--- a/Common/Core/TrackSelection.h
+++ b/Common/Core/TrackSelection.h
@@ -160,7 +160,7 @@ class TrackSelection
         return track.tpcChi2NCl() <= mMaxChi2PerClusterTPC;
 
       case TrackCuts::kTPCRefit:
-        return (isRun2 && mRequireTPCRefit) ? (track.flags() & o2::aod::track::TPCrefit) : true;
+        return (mRequireTPCRefit ? (isRun2 ? (track.flags() & o2::aod::track::TPCrefit) : track.hasTPC()) : true);
 
       case TrackCuts::kITSNCls:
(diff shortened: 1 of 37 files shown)


njacazio:
@mario-krueger @mfaggin could you please have a look as well?

Common/Core/TrackSelection.h
```diff
@@ -52,26 +52,83 @@ class TrackSelection
   template <typename T>
   bool IsSelected(T const& track)
   {
-    const bool isRun2 = track.trackType() == o2::aod::track::Run2Track || track.trackType() == o2::aod::track::Run2Tracklet;
```
mfaggin, line 55: @njacazio maybe this line shall be left as it is, so that the class is still usable with Run2 converted data

Common/Core/TrackSelection.h
```diff
@@ -112,7 +169,7 @@ class TrackSelection
         return track.itsChi2NCl() <= mMaxChi2PerClusterITS;
 
       case TrackCuts::kITSRefit:
-        return (isRun2 && mRequireITSRefit) ? (track.flags() & o2::aod::track::ITSrefit) : true;
+        return (mRequireITSRefit ? (isRun2 ? (track.flags() & o2::aod::track::ITSrefit) : track.hasITS()) : true);
```
mfaggin, line 172: @njacazio maybe a similar adjustement can be done also for the `TrackCuts::kTPCRefit` case, asking for the `track.hasTPC()` requirement for Run3 tracks, what do you think?

iarsene:
Looks good from my side

Common/DataModel/TrackSelectionTables.h
```diff
@@ -19,12 +19,63 @@ namespace o2::aod
...
 {
 // Columns to store the DCA to the primary vertex
-DECLARE_SOA_COLUMN(DcaXY, dcaXY, float); //!
-DECLARE_SOA_COLUMN(DcaZ, dcaZ, float);   //!
+DECLARE_SOA_COLUMN(DcaXY, dcaXY, float); //! Impact parameter in XY of the track to the primary vertex
+DECLARE_SOA_COLUMN(DcaZ, dcaZ, float);   //! Impact parameter in Z of the track to the primary vertex
+
+struct TrackSelectionFlags {
+ public:
+  typedef uint16_t flagtype;
```
mfaggin, line 27: @njacazio here a simple curiosity: do you define a new `typedef` just to be flexible in case we need in the future to extend this `uint16_t` to a bigger data format, eg `uint32_t` oreven `uint64_t` in case we need to insert new cuts in the bitmap? Was it the idea? If yes, fine for me.

mfaggin (commented):
@njacazio thank you for the work, I just have a couple of suggestions, see my comments in line.

Common/DataModel/TrackSelectionTables.h
njacazio: exactly, this was my point

Common/Core/TrackSelection.h
njacazio: This should happen already here, before it was split in two occurrences, now it has been unified, have a look at line 138

Common/Core/TrackSelection.h
```diff
@@ -52,26 +52,83 @@ class TrackSelection
...
+    setFlag(TrackCuts::kITSRefit);
+    setFlag(TrackCuts::kITSHits);
+    setFlag(TrackCuts::kGoldenChi2);
+    setFlag(TrackCuts::kDCAxy);
+    setFlag(TrackCuts::kDCAz);
+
+    return flag;
   }
 
   // Temporary function to check if track passes a given selection criteria. To be replaced by framework filters.
```
njacazio, line 134: @mfaggin  Here it should be equivalent

Common/Core/TrackSelection.h
njacazio: Thanks Mattia, indeed, that's a good point, I implemented it

Common/Core/TrackSelection.h
mfaggin: you are right! Thank you

Common/TableProducer/trackselection.cxx
```diff
@@ -53,7 +57,8 @@ struct TrackSelectionTask {
   {
     for (auto& track : tracks) {
       filterTable((uint8_t)globalTracks.IsSelected(track),
```
mario-krueger, line 59: Since with your changes every cut is checked for every track, we could also derive the IsGlobalTrack decision from the flag that you create below, right? This would avoid doing the checks multiple times.
IsGlobalTrack requires all of the bits to be set, so one could just check if it is equal to `uint8_t(~uint8_t(0))`. 
Then in the future one can also easily define additional 'filterBitmasks' / decision columns where for instance the dca or pt cuts are omitted.

njacazio: Indeed one could define the global quality as a dynamic column! Good point Mario, thanks!

njacazio: @mario-krueger what do you think of this? https://github.com/AliceO2Group/O2Physics/pull/691/commits/6a364353d9a8f8ddce666ffd94268d12f8356765 Unfortunately to handle the SSD one still needs a separate bool

mario-krueger: I like it :)

Common/TableProducer/trackselection.cxx
```diff
@@ -56,8 +56,7 @@ struct TrackSelectionTask {
   void process(soa::Join<aod::FullTracks, aod::TracksExtended> const& tracks)
   {
     for (auto& track : tracks) {
-      filterTable((uint8_t)globalTracks.IsSelected(track),
-                  (uint8_t)globalTracksSDD.IsSelected(track),
+      filterTable((uint8_t)globalTracksSDD.IsSelected(track),
                   globalTracks.IsSelectedFlag(track));
```
mario-krueger, line 60: As an alternative to the dynamic columns which cannot be used in `Filters`, you could probably do it like this:
```
for (auto& track : tracks) {
  flagtype flags = globalTracks.IsSelectedFlag(track);
  bool isGlobalTrack = (flags & TrackSelectionFlags::kGlobalTrack) == TrackSelectionFlags::kGlobalTrack;
  filterTable((uint8_t)isGlobalTrack,
                    (uint8_t)globalTracksSDD.IsSelected(track),
                     flag);
}
```

njacazio: Indeed this is what I did at the end with the latest commit, thanks!

Common/TableProducer/trackselection.cxx
```diff
@@ -65,6 +72,6 @@ struct TrackSelectionTask {
 //****************************************************************************************
 WorkflowSpec defineDataProcessing(ConfigContext const& cfgc)
 {
-  WorkflowSpec workflow{adaptAnalysisTask<TrackSelectionTask>(cfgc, TaskName{"track-selection"})};
+  WorkflowSpec workflow{adaptAnalysisTask<TrackSelectionTask>(cfgc)};
```
jgrosseo, line 75: Why?

Common/DataModel/TrackSelectionTables.h
```diff
@@ -19,21 +19,79 @@ namespace o2::aod
...
+  static constexpr flagtype kDCAz = 1 << 14;
+  static constexpr flagtype kGlobalTrack = kTrackType | kPtRange | kEtaRange | kTPCNCls | kTPCCrossedRows | kTPCCrossedRowsOverNCls | kTPCChi2NDF | kTPCRefit | kITSNCls | kITSChi2NDF | kITSRefit | kITSHits | kGoldenChi2 | kDCAxy | kDCAz;
+};
 
 // Columns to store track filter decisions
-DECLARE_SOA_COLUMN(IsGlobalTrack, isGlobalTrack, uint8_t);       //!
-DECLARE_SOA_COLUMN(IsGlobalTrackSDD, isGlobalTrackSDD, uint8_t); //!
+DECLARE_SOA_COLUMN(IsGlobalTrack, isGlobalTrack, uint8_t);                     //!
+DECLARE_SOA_COLUMN(IsGlobalTrackSDD, isGlobalTrackSDD, uint8_t);               //!
+DECLARE_SOA_COLUMN(TrackCutFlag, trackCutFlag, TrackSelectionFlags::flagtype); //! Flag with the single cut passed flagged
```
jgrosseo, line 49: Isn't this duplicating information? If a certain set of those are 1, then isGlobalTrack is 1

Common/DataModel/TrackSelectionTables.h
```diff
@@ -19,21 +19,79 @@ namespace o2::aod
...
+  static constexpr flagtype kTPCChi2NDF = 1 << 6;
+  static constexpr flagtype kTPCRefit = 1 << 7;
+  static constexpr flagtype kITSNCls = 1 << 8;
+  static constexpr flagtype kITSChi2NDF = 1 << 9;
+  static constexpr flagtype kITSRefit = 1 << 10;
+  static constexpr flagtype kITSHits = 1 << 11;
+  static constexpr flagtype kGoldenChi2 = 1 << 12;
+  static constexpr flagtype kDCAxy = 1 << 13;
+  static constexpr flagtype kDCAz = 1 << 14;
+  static constexpr flagtype kGlobalTrack = kTrackType | kPtRange | kEtaRange | kTPCNCls | kTPCCrossedRows | kTPCCrossedRowsOverNCls | kTPCChi2NDF | kTPCRefit | kITSNCls | kITSChi2NDF | kITSRefit | kITSHits | kGoldenChi2 | kDCAxy | kDCAz;
```
jgrosseo, line 43: Will these be the same for Run 2 and Run 3?

Common/TableProducer/trackselection.cxx
njacazio: I thought that it was advised to remove `TaskName`, reverting this

Common/DataModel/TrackSelectionTables.h
njacazio: yes, the change is in the track selection class

jgrosseo: I mean, will we not have different cuts for Run 3?

Common/TableProducer/trackselection.cxx
jgrosseo: But then the name changes from track-selection to track-selection-task. Thanks for reverting.

Common/DataModel/TrackSelectionTables.h
njacazio: Yes, but the bits of the bitmask will be the same. Then the bitmask will be filled differently for Run2 and Run3

Common/DataModel/TrackSelectionTables.h
njacazio: Yes, this is definitively duplicated information. I tried having an expression column (https://github.com/AliceO2Group/O2Physics/pull/691/commits/9ead612d125e0d1b3cf820b871a2b8ef9f3b2cc9) based on the flag content but this was not working out at runtime. I can try to reproduce the issue I was observing

njacazio:
```
[17114:qa-event-track]: [12:14:10][STATE] Exiting FairMQ state machine
[17114:qa-event-track]: [12:14:10][FATAL] Unhandled o2::framework::runtime_error reached the top of main of o2-analysis-qa-event-track, device shutting down.
[17114:qa-event-track]:  Reason: Invalid combination of argument types 1 and 2
[17114:qa-event-track]:  Backtrace follow: 
[17114:qa-event-track]: 
```

This is why I reverted to having the duplicate information

jgrosseo: Some type cast seems to be missing. Can you send me the expression column in mattermost?

jgrosseo: But it is anyway in memory, so expression does not really help (it will use the same memory)...

njacazio:
@jgrosseo the duplicated info is gone but the filtering syntax is a bit changed. This affects many tasks, apologies ;)

njacazio:
@jgrosseo is we agree on this I would merge it or close it as the large number of changed file is large and the rate of conflicts is quite high

saganatt (dismissed):

jgrosseo:
@njacazio This is a tremendous work, indeed.

I wonder if `(aod::track::trackCutFlag & aod::track::TrackSelectionFlags::kGlobalTrack) == aod::track::TrackSelectionFlags::kGlobalTrack)` is not rather complicated for the standard analyzer... But I have no great proposal at this stage.

My second comment is if the getter names of the different bits are the good. E.g. `passedTPCCrossedRows` is not really logical. You mean `passedTPCCrossedRows` or so, right?

jgrosseo:
Couldn't one add one more bit for the SDD selection and it would be solved?

njacazio:
> @njacazio This is a tremendous work, indeed.
> 
> I wonder if `(aod::track::trackCutFlag & aod::track::TrackSelectionFlags::kGlobalTrack) == aod::track::TrackSelectionFlags::kGlobalTrack)` is not rather complicated for the standard analyzer... But I have no great proposal at this stage.
> 
Indeed that's a bit a long line. I now have defined a macro  that does this

> My second comment is if the getter names of the different bits are the good. E.g. `passedTPCCrossedRows` is not really logical. You mean `passedTPCCrossedRows` or so, right?

This is now fixed, good point!


> Couldn't one add one more bit for the SDD selection and it would be solved with the bitmask as well?

Yes, this is in the plans, however I would do this in the next PR

jgrosseo:
> > @njacazio This is a tremendous work, indeed.
> > I wonder if `(aod::track::trackCutFlag & aod::track::TrackSelectionFlags::kGlobalTrack) == aod::track::TrackSelectionFlags::kGlobalTrack)` is not rather complicated for the standard analyzer... But I have no great proposal at this stage.
> 
> Indeed that's a bit a long line. I now have defined a macro that does this. This now reads as:
> 
> ```
>   Filter trackFilter = requireGlobalTrackInFilter();
> ```
I agree with it. Are non-capital #ifdef allowed?

mario-krueger (approved):

saganatt (approved):

iarsene (approved):

vkucera (approved):

PR #697, opened 2022-04-22
Small fix in trigger task

vfeuilla:
Change in the track selection task so that the Selection flag table has always the same size as the track table.
Track selection does not need to run on both tracks and events

### DIFF
--- a/PWGDQ/Tasks/filterPP.cxx
+++ b/PWGDQ/Tasks/filterPP.cxx
@@ -222,29 +222,39 @@ struct DQBarrelTrackSelection {
   {
     uint32_t filterMap = uint32_t(0);
     trackSel.reserve(tracksBarrel.size());
+    int CollisionId = -1;
 
     VarManager::ResetValues(0, VarManager::kNBarrelTrackVariables);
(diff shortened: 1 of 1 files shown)


iarsene:
Hi @vfeuilla,
The point of having the events also available is to be able to run selections that depend also on event variables. For example, if we want to apply cuts on Lxy, we need the event vertex and covariances. Or we may want to apply some cuts depending on vertex position...  Anyway, this was the main reason i had the events available there. We could implement your PR, but we might have to add these back in when we will discuss the actual cuts which we will implement for the trigger or for when people would use the task on their on to filter events. Is there a significant gain in resource usage if we implement this PR ?

vfeuilla:
Hi @iarsene 
The problem by running on both the tracks and the events, in that tracks that are not linked to any event (which happens quite often in the Run3 AO2D) are not processed when doing the track selection, and therfore the "aod::tracks" tables and "aod::DQBarrelTrackCuts" table have different sizes and can't be joined in the "MyBarrelTracksSelected" table. The only way to avoid that is to run on the tracks only.

iarsene (approved):

PR #701, opened 2022-04-27
Forward vertexing for ambiguous tracks

robincaron13:
This task has been developed by @sarahherrmann and me, in order to associate a correct collision to an ambiguous track based on the calculation of the minimum transverse DCA. 
Right now this task is running on `aod::AmbiguousMFTTracks` but the code can be adapted to process the `aod::AmbiguousFwdTracks`. 
Some parts are temporary or in commented lines since we did not manage to use different o2 features (it is currently a bit time consuming, but we hope it could be improved later).

### DIFF
--- a/PWGMM/Tasks/vertexing-fwd.cxx
+++ b/PWGMM/Tasks/vertexing-fwd.cxx
@@ -31,7 +31,8 @@ using SMatrix5 = ROOT::Math::SVector<Double_t, 5>;
 struct vertexingfwd {
 
   // Configurable<int> rangeBC{"rangeBC", 10, "Range for collision BCId and BCglobalIndex correspondance"};
-  int rangeBC = 10;
+  int rangeBC = 2;
+
 
   HistogramRegistry registry{
     "registry",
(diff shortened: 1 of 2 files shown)


PWGMM/Tasks/vertexing-fwd.cxx
```diff
@@ -0,0 +1,198 @@
...
+        continue;
+      }
+      if (!tracks.size() | (tracks.size() <= 1)) {
+        continue;
+      }
+      if ((ambitrack.globalIndex() >= tracks.size()) | (ambitrack.globalIndex() < 0)) {
+        continue;
+      }
+
+      // soa::Join<o2::aod::FwdTracks, o2::aod::FwdTracksCov, aod::McFwdTrackLabels>::iterator extAmbiTrack;
```
aalkin, line 82: This is not needed, the index should be directly accessible like this: `auto track = ambitrack.mfttrack_as<soa::Join<o2::aod::MFTTracks, aod::McMFTTrackLabels>>();`

One can also introduce an alias `using MFTTracksLabeled = soa::Join<o2::aod::MFTTracks, aod::McMFTTrackLabels>;` and use it to save some typing.

robincaron13: Thanks!

PWGMM/Tasks/vertexing-fwd.cxx
```diff
@@ -0,0 +1,188 @@
...
+      // Fill the std::vector for ambiguous tracks with the quantities needed
+      vecAmbTrack.push_back(track.x());
+      vecAmbTrack.push_back(track.y());
+      vecAmbTrack.push_back(track.phi());
+      vecAmbTrack.push_back(track.tgl());
+      vecAmbTrack.push_back(track.signed1Pt());
+      vecAmbTrack.push_back(track.z());
+      vecAmbTrack.push_back(track.chi2());
+
+      for (auto& collision : collisions) {
```
aalkin, line 112: Looping over the whole collision table is a bit heavy, there is, however `Run3MatchedToBCSparse` table that is compatible with BCs table and can be joined to it, that also has an index to collisions. If you use ambiguous track's index that refers to it, you can directly retrieve a slice of the joined BC+Run3MatchedToBCSparse table and go directly to collisions (if any) from it.

aalkin: `bcIds().size()` returns the size of the _underlying index array_ which is always 2 - it contains two integers, the start and the end of a slice. The actual getter `bc_as<>` returns the subtable, size of it is what you need here.

sarahherrmann: Thanks for your answer Anton ! We then have a bigger problem, the code crashes when we try to get any information from collision1, such as collision1.posZ() or such ... Even though we checked that the collisionId needed to be >0 (-1 means no collision associated to this BC)
I really don't see why
Any idea of a security we could implement to avoid this issue ?

aalkin: Which version of code is crashing? There can be several reasons for the crash when dereferencing the index.

sarahherrmann: Hi Anton, Robin just pushed the version of the code which is crashing, it crashes when we try to access collision.posZ() (just after the LOG 1bis)
Any idea of how to fix this ? The error doesn't seem really helpful
Screenshot 2022-05-09 at 09 54 28

aalkin: I see now, thank you. You need to check if there is a collision for this bc, using `bcambi.has_collision()`.

sarahherrmann: Ok thanks a lot ! Will do

PR #710, opened 2022-05-02
[EMCAL-734] split EMCal cluster table and added cluster definition class

fjonasALICE:
This PR is the beginning of the efforts to solve two problems:
1. Allow multiple clusterizers to run at the same time, even though there is only a single table for emcal clusters
2. Deal with clusters that could not be matched to a collisions ID

The first problem is solved by introducing a EMCalClusterDefinition class that contains all the information about what is considered as a cluster. This information is then stored as an int in the cluster table. It has the added advantage that it can also be used for easy steering of the EMCal correction framework. In the current implementation, the correction framework can be given an arbitrary number of cluster definitions as a string separated by commas.

The second problem is solved by using two exclusive tables, where one table contains clusters which could not be matched to a collisionID, which uses the BCID as an index. For clusters that could be matched to a collision the table uses collisionID as an index.

The task [PWGJE/Tasks/emcclustermonitor.cxx] illustrates how to access clusters that were matched to a collision. The requested ClusterDefinition can be given by the user as a string. Once the framework evolves to allow backwards communication, the correctionframework could check what clusterdefinitions are requested by the various tasks.

In the future a second process function will be added to the emcclustermonitor to illustrate how to access ambigous clusters and cut on the BCIDs. In addition, some work is still needed to validate the clusterizer and matching to collisionID

### DIFF
--- a/PWGJE/DataModel/EMCALClusters.h
+++ b/PWGJE/DataModel/EMCALClusters.h
@@ -56,21 +56,21 @@ const EMCALClusterDefinition getClusterDefinitionFromString(const std::string& c
   }
 };
 
-DECLARE_SOA_INDEX_COLUMN(Collision, collision);                        //!
-DECLARE_SOA_INDEX_COLUMN(BC, bc);                                      //!
-DECLARE_SOA_COLUMN(ID, id, int);                                       //!
-DECLARE_SOA_COLUMN(Energy, energy, float);                             //!
-DECLARE_SOA_COLUMN(CoreEnergy, coreEnergy, float);                     //!
-DECLARE_SOA_COLUMN(Eta, eta, float);                                   //!
-DECLARE_SOA_COLUMN(Phi, phi, float);                                   //!
-DECLARE_SOA_COLUMN(M02, m02, float);                                   //!
-DECLARE_SOA_COLUMN(M20, m20, float);                                   //!
-DECLARE_SOA_COLUMN(NCells, nCells, int);                               //!
-DECLARE_SOA_COLUMN(Time, time, float);                                 //!
-DECLARE_SOA_COLUMN(IsExotic, isExotic, bool);                          //!
-DECLARE_SOA_COLUMN(DistanceToBadChannel, distanceToBadChannel, float); //!
-DECLARE_SOA_COLUMN(NLM, nlm, int);                                     //!
-DECLARE_SOA_COLUMN(Definition, definition, int);                       //!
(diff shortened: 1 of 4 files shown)


mfasDa:
Looks fine to me. @jgrosseo could you please merge?

PWGJE/DataModel/EMCALClusters.h
```diff
@@ -17,11 +17,46 @@
...
+  } else if (clusterDefinitionName == "kV3Variation1") {
+    return kV3Variation1;
+  } else if (clusterDefinitionName == "kV3Variation2") {
+    return kV3Variation2;
+  } else {
+    throw std::invalid_argument("Cluster definition name not recognized");
+  }
+};
+
+DECLARE_SOA_INDEX_COLUMN(Collision, collision);                        //!
```
jgrosseo, line 59: Could you add comments to all those lines behind the //!
This is parsed nightly and published in the documentation.

fjonasALICE: thanks for reviewing the code! I added the description accordingly

jgrosseo (approved):

PWGJE/DataModel/EMCALClusters.h
```diff
@@ -17,34 +17,77 @@
...
+DECLARE_SOA_COLUMN(NCells, nCells, int);                               //! number of cells in cluster
+DECLARE_SOA_COLUMN(Time, time, float);                                 //! cluster time (ns)
+DECLARE_SOA_COLUMN(IsExotic, isExotic, bool);                          //! flag to mark cluster as exotic
+DECLARE_SOA_COLUMN(DistanceToBadChannel, distanceToBadChannel, float); //! distance to bad channel
+DECLARE_SOA_COLUMN(NLM, nlm, int);                                     //! number of local maxima
+DECLARE_SOA_COLUMN(Definition, definition, int);                       //! cluster definition, see EMCALClusterDefinition.h
 
+} // namespace emcalcluster
+// table of clusters that could be matched to a collision
 DECLARE_SOA_TABLE(EMCALClusters, "AOD", "EMCALCLUSTERS", //!
```
jgrosseo, line 77: I only noticed now that this should not be plural in the in-file table name. Can you adjust to "EMCALCLUSTER", please.

jgrosseo: @fjonasALICE

PR #715, opened 2022-05-04
Multiplicity directory + multCalibrator first commit

ddobrigk:


ddobrigk:
I cannot find the whitespace problem at all... I did run clang-format on all files, and I tried bash scripts for removing trailing whitespace, but there isn't any problem anywhere to be found. Am I missing something? Sorry if so.

ddobrigk:
Turns out the reason for my confusion was that the trailing whitespaces the system complained about were not introduced by my changes - rather, they were already there before (I guess the checks changed with time). Should all be fine now!

CODEOWNERS
```diff
@@ -28,3 +28,5 @@
 /PWGLF @alibuild @njacazio
 /PWGMM @alibuild @aalkin
 /Tools/PIDML @alibuild @saganatt
+/Common/Tools/Multiplicity @alibuild @ddobrigk @victor-gonzalez
```
jgrosseo, line 31: Please move this up (behind Common)

Common/CMakeLists.txt
```diff
@@ -15,3 +15,4 @@ add_subdirectory(CCDB)
 add_subdirectory(Tasks)
 add_subdirectory(TableProducer)
 add_subdirectory(Tools)
+add_subdirectory(Tools/Multiplicity)
```
jgrosseo, line 18: Please add this to the CMakeLists.txt in Tools instead

ddobrigk: Thanks, @jgrosseo, done! (was busy preparing LHCP talk)

jgrosseo (approved):

PR #737, opened 2022-05-12
Using correct timestamp

jgrosseo:


jgrosseo:
Needs filling of MeanVertex

jgrosseo:
@chiarazampolli @victor-gonzalez

Common/TableProducer/trackPropagation.cxx
```diff
@@ -89,17 +90,22 @@ struct TrackPropagation {
...
+  void initCCDB(aod::BCsWithTimestamps::iterator const& bc)
+  {
+    if (runNumber == bc.runNumber()) {
+      return;
+    }
+    grpo = ccdb->getForTimeStamp<o2::parameters::GRPObject>(grpPath, bc.timestamp());
+    LOGF(info, "Setting magnetic field to %d kG for run %d from its GRP CCDB object", grpo->getNominalL3Field(), bc.runNumber());
+    o2::base::Propagator::initFieldFromGRP(grpo);
+    o2::base::Propagator::Instance()->setMatLUT(lut);
+    mVtx = ccdb->getForTimeStamp<o2::dataformats::MeanVertexObject>(mVtxPath, bc.timestamp());
```
victor-gonzalez, line 107: Not familiar with this but is the validity of the mean vertex the whole run?

victor-gonzalez: I saw that currently it is zero but I mean at mean term

victor-gonzalez (dismissed):
Apart from my comment the implementation looks correct to me

Common/TableProducer/trackPropagation.cxx
jgrosseo: Yes, see Ruben's comment "The MeanVertex is a run specific calibration object (should be constantly created in relatively short time intervals during the run) and should be used in the reconstruction for e.g. primary vertexing. " but this will anyway be replaced by the CCDB fetcher which will detect if the timestamp goes out of the objects validity.

alibuild (approved):

PR #739, opened 2022-05-12
reduce output even more

jgrosseo:


alibuild (approved):

PR #741, opened 2022-05-13
Add invariant mass selection

lbariogl:


PWGCF/FemtoDream/FemtoDreamV0Selection.h
```diff
@@ -264,6 +279,13 @@ bool FemtoDreamV0Selection::isSelectedMinimal(C const& col, V const& v0, T const
   const float dcaDaughv0 = v0.dcaV0daughters();
   const float cpav0 = v0.v0cosPA(col.posX(), col.posY(), col.posZ());
 
+  const float invMassLambda = v0.mLambda();
+  const float invMassAntiLambda = v0.mAntiLambda();
+
+  if ((invMassLambda < fInvMassLowLimit and invMassLambda > fInvMassUpLimit) and (invMassAntiLambda < fInvMassLowLimit and invMassAntiLambda > fInvMassUpLimit)) {
```
victor-gonzalez, line 285: It looks to me that this condition will never be true
Is it that the intention? Then, why asking about it?

victor-gonzalez (commented):
Check my comment
If you confirm it is your intention I'll approve it

PWGCF/FemtoDream/FemtoDreamV0Selection.h
lbariogl: No, you are right. I will change it now

PR #742, opened 2022-05-13
DPG: fix switcher for filter

mfaggin:


mfaggin:
@njacazio @mario-krueger here the small proposed fix

mario-krueger (approved):

PR #750, opened 2022-05-16
Run3 centrality/multiplicity tables

victor-gonzalez:


victor-gonzalez:
@ddobrigk No, actually this is the final implementation if everything works as expected
There remain the tests with the centralized CCDB

ekryshen:
@victor-gonzalez @ddobrigk thanks a lot for your efforts. A few comments/questions:
* It would be better to use FV0A notation instead of FV0M (keep M for the sum of forward and backward)

* I would avoid "Run3" in the estimators for Run3. We should try to keep the name as short as possible. Since Run3 is our default and we already have F in the name, it should be fine to use just centFV0A, centFT0M etc.

* I am not sure if having different centrality estimators in different tables is the best solution. I understand that it potentially saves some memory. But from the analysis point of view, one has to modify join<collisions, cent...> subscription of the process() function every time one wants to try a new estimator. It doesn't seem very convenient if one wants to change centrality estimators using some configurable of the analysis task. @jgrosseo what do you think?

* On FV0A multiplicity, FV0 (and FDD) signal may be long (several bcs) for high multiplicity events so we should be prepared to integrate multiplicity in several bcs. No need to fix it in this PR, just not to forget it in the future

victor-gonzalez:
@ekryshen Thanks for your comments/suggestions

- regarding the naming, we will get rid of `Run3` and move to `FV0A` in the next update
- regarding multiple tables, in here we don't agree. In O2 we have to enforce the analyzers to be aware of the 'limitations' we are addressing. It does not make any sense to fill the whole centrality table if only one of the estimators is being used. From the analysis point of view it only requires to have different `process` with different subscriptions and a common function for all of them. The different `process` can be activated from hyperloop or via json file if so is decided. No code need to be touched. I believe we should impose this kind of template whose usage the analyzers would benefit in its application to similar cases (data and MC reconstructed, MC generated, pp and PbPb analysis, etc) @jgrosseo What do you think?
- regarding the integration along several bcs, we'll keep track of it

ekryshen:
@victor-gonzalez regarding multiple tables, I don't insist, I am just thinking about user convenience. Regarding hyperloop, note that different users may need different centrality tables. Does it mean that we should foresee different centrality wagons with different est... configurables for different user tasks?

jgrosseo:
@victor-gonzalez @ekryshen The task figures out automatically which tables to fill, so on Hyperloop only one wagon is needed

Common/TableProducer/centralityTable.cxx
```diff
@@ -206,59 +224,190 @@ struct CentralityTable {
...
+          }
+        }
+        if (estRun3NTPV == 1) {
+          LOGF(info, "Getting new histograms with %d run number for %d run number", mRunNumber, bc.runNumber());
+          Run3NTPVInfo.mhMultSelCalib = getccdb("hCalibZeqNTracks");
+          if (Run3NTPVInfo.mhMultSelCalib != nullptr) {
+            Run3NTPVInfo.mCalibrationStored = true;
+          } else {
+            LOGF(fatal, "Calibration information from NTPV for run %d corrupted", bc.runNumber());
+          }
```
jgrosseo, line 349: This code looks very repetitive. Could it be solved by a function / lambda?

ekryshen:
@victor-gonzalez @jgrosseo Ok, thanks, I see.

victor-gonzalez:
@ekryshen @ddobrigk @jgrosseo The functionality is complete
It has to be still in draft until all CCDB information is in the production endpoint

ddobrigk:
Hi @victor-gonzalez, I think this is good to go, right? Note the conflict (due to my change making the vertex-Z optional) which must still be resolved. Thank you very much!

victor-gonzalez:
Conflict resolved!

victor-gonzalez:
Now everything is ready an all centrality information for pilot beam data is in the CCDB production endpoint
The PR is ready for merging after positive review

jgrosseo (approved):
This is fine with me.
The QA task could use the histogram registry in a future revision.

PR #756, opened 2022-05-18
PIDML: Use github repository to access model files

saganatt:
This will enable analysis on GRID


alibuild (approved):

PR #763, opened 2022-05-18
PWGEM/gammaConv add histos and cuts

sstiefel19:
- add collision histos for better control over processed events
- add cut on mc true eta and IsPhysicalPrimary


sstiefel19:
problem with clang-fromat, will reopen another

PR #766, opened 2022-05-19
Lambdakzerobuilder: auto B field, run2/3 TracksIU change

ddobrigk:


PWGLF/TableProducer/lambdakzerobuilder.cxx
```diff
@@ -148,13 +160,225 @@ struct lambdakzeroBuilder {
...
+                   ,
+                   aod::McParticles const& particlesMC
 #endif
   )
   {
+
+    float d_bz;
+    if (d_bz_input < -990) {
+      // Fetch magnetic field from ccdb for current collision
+      d_bz = getMagneticField(collision.bc_as<aod::BCsWithTimestamps>().timestamp());
```
njacazio, line 190: Won't this cause any overhead? I mean, maybe it could be useful to get the magnetic field only if the run number changed or something?

PWGLF/TableProducer/lambdakzerobuilder.cxx
```diff
@@ -148,13 +160,225 @@ struct lambdakzeroBuilder {
...
+      nTrack.rotateParam(fitter.getTrack(1).getAlpha());
+
+      // Retry closer to minimum with material corrections
+      o2::base::Propagator::MatCorrType matCorr = o2::base::Propagator::MatCorrType::USEMatCorrNONE;
+      if (useMatCorrType == 1)
+        matCorr = o2::base::Propagator::MatCorrType::USEMatCorrTGeo;
+      if (useMatCorrType == 2)
+        matCorr = o2::base::Propagator::MatCorrType::USEMatCorrLUT;
+
+      o2::base::Propagator::Instance()->propagateToX(pTrack, finalXpos, d_bz, 0.85f, 2.0f, matCorr);
```
njacazio, line 306: Can these `0.85f` and `2.0f` be defined as `constexpr` variables to have a definition?

PWGLF/TableProducer/lambdakzerofinder.cxx
```diff
@@ -133,17 +136,51 @@ struct lambdakzerofinder {
...
+    // using namespace analysis::lambdakzerofinder;
+
+    ccdb->setURL("https://alice-ccdb.cern.ch");
+    ccdb->setCaching(true);
+    ccdb->setLocalObjectValidityChecking();
+  }
+
+  float getMagneticField(uint64_t timestamp)
+  {
+    // TODO done only once (and not per run). Will be replaced by CCDBConfigurable
```
njacazio, line 158: If this is the same wrt the builder, maybe it is worth making a header file common to both

njacazio (commented):
Hi David, great job, just few minor suggestions that you could consider to streamline a bit

PWGLF/TableProducer/lambdakzerobuilder.cxx
```diff
@@ -59,24 +59,36 @@
...
 using namespace o2;
 using namespace o2::framework;
 using namespace o2::framework::expressions;
 using std::array;
 
-using FullTracksExt = soa::Join<aod::FullTracks, aod::TracksCov, aod::TracksExtended>;
+// custom configurable for switching between run2 and run3 track storing models
+void customize(std::vector<ConfigParamSpec>& workflowOptions)
+{
+  workflowOptions.push_back(ConfigParamSpec{"selection-run", VariantType::Int, 2, {"selection type: 2 - run 2, 3 - run 3"}});
```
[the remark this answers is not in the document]
njacazio: Actually, I realize that this might not be used at all? Could it be?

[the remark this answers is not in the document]
ddobrigk: You're right. This is technically a copy-paste from another code which implemented the run 2 / 3 switches which means that probably also that code has a surplus line... I'll remove! Thanks!!!

PWGLF/TableProducer/lambdakzerobuilder.cxx
ddobrigk: Very reasonable suggestion, will implement run-switch-detector! Thanks!

njacazio (approved):

PR #770, opened 2022-05-20
Adding a copy of TightGlobalTrack cut without ITS/TPC refit

feisenhu:
This is a PR to add two new cut setting in the CutsLibrary.h 

- lmee_GlobalTrackRun3

- lmee_GlobalTrackRun3_lowPt

both rely on a new cut called "TightGlobalTrackRun3" which is a copy of "TightGlobalTrack" without the `kIsITSrefit` and `kIsTPCrefit`. These two flags seem only to be available for converted data from Run2, hence I removed them to analyse Run3 data.


feisenhu:
Hello :)
@iarsene , @vfeuilla, @dsekihat 
I would appreciate if one of you could have a look at this PR.

thank you in advance.

iarsene (approved):

PR #787, opened 2022-05-24
Adding additional cuts to reduce the size of the output

feisenhu:
Adding two more cut settings for dielectric analysis:

- lmeePID_TPChadrejTOFrecRun3
- lmee_GlobalTrackRun3_TPC_ePID_lowPt

These apply loose and some stricter PID cuts to select electrons.

This should reduce the size of the output, necessary for a train.


feisenhu:
@iarsene

Hello :) could you have another look for this pull request?
Thank you very much.

iarsene (approved):

PR #798, opened 2022-05-26
[EMCAL-737] Added pi0 QC task for EMCal calibration

jokonig:
- QC task based on cluster monitor task in PWGJE
- Goal is to get the pi0 mass peak and width for calibration of the
EMCal. This task is not yet meant for physics
- Cluster cuts on timing, cluster shape, cluster energy etc. available
and settable via Configurables
- Same event as well as rotation background implemented.
- Various QA histograms for cluster properties


jokonig:
Could someone have a look at this PR?
@ktf @jgrosseo @iarsene

PR #809, opened 2022-05-30
DPG: use only process switches in EffQA

njacazio:


mfaggin:
> @mfaggin could you also have a look? Thanks!!

Ciao @njacazio , I had a look at it and it seems pretty clear to me, I think you can go on.

mario-krueger (approved):

PR #819, opened 2022-06-01
DPG: allow hyperloop running on derived data

njacazio:
- Separate split and non split tasks
- Add more histos for MC
- Fix process function of MC


njacazio:
@mfaggin this should take care of the issue + other improvements

PR #828, opened 2022-06-06
PID: Simplify workflow for Ev. Time

njacazio:
- Add TOF mass table


njacazio:
With this PR the Run3 vs Run2 selection is done in one single point and propagated to all the tasks using TOF PID having a single handle to decide how event time is used

jgrosseo (approved):

PR #829, opened 2022-06-06
Set output to standard instead of QualityObject

ercolessi:
Hi, @aimeric-landou since when we run this task on Hyperloop we don't get the QA outputs I thought to change it to the stardard ones. 
(many of the changes in the commit are just super minor clang fixes on comments ... apologies if they create a bit of entropy)

Could you have a look?


njacazio (approved):
I agree with the change, indeed the Quality Object is less useful when running over hyperloop, thanks Francesca!

PR #836, opened 2022-06-07
DPG: template process for data and MC.

mfaggin:


mario-krueger (approved):

mfaggin:
@njacazio here the implementation of template process function as I mentioned

njacazio:
This is fine, I would let Mario decide, thanks Mattia!

PR #837, opened 2022-06-07
Add customizable track selection

njacazio:


jgrosseo:
On Hyperloop as soon as someone customizes the track selection, we cannot join wagons of different people anymore... I wonder if we having a wrong strategy...

njacazio:
Would you have a set of different pT and Eta selection then? So that one can filter on those afterwards?

njacazio:
A solution would also be to remove the acceptance range from these cuts and leave it to the analyzers

jgrosseo:
Yes, I think this is better. And in that case analyzers have to put the pT and eta cut themselves.

iarsene (approved):

PR #846, opened 2022-06-09
fix awful bug (bz initialization order)

ddobrigk:


ddobrigk:
In this awful bug, the magnetic field initialization of the finders was done before the magnetic field was consulted in the CCDB. In my understanding this should only affect the very first event analysed by a certain job, but still, it is a bug to be fixed.

PWGLF/TableProducer/cascadebuilder.cxx
```diff
@@ -143,6 +143,20 @@ struct cascadeBuilder {
 
   void processRun2(aod::Collision const& collision, aod::V0sLinked const&, aod::V0Datas const& v0data, aod::Cascades const& cascades, FullTracksExt const&, aod::BCsWithTimestamps const&)
   {
+
+    hEventCounter->Fill(0.5);
+
```
njacazio, line 148: Can't this be put in a separate function so as to have a single breaking point?

PR #848, opened 2022-06-09
Add more masks for track cuts

njacazio:
With this we are more differential in the cuts and we build similar cuts with filter bits


njacazio:
@mario-krueger @mfaggin

njacazio:
Closed by mistake, apologies

PR #849, opened 2022-06-09
fix caching behaviour for TPC CCDB objects

jezwilkinson:
Previously the CCDB call made for the TPC PID response during init() was preventing newer CCDB calls during the track loop due to the caching. This patch removes the CCDB call during initialisation and the CCDB is only requested during the track loop so that the correct response object for the BC is loaded.

### DIFF
--- a/Common/TableProducer/PID/pidTPC.cxx
+++ b/Common/TableProducer/PID/pidTPC.cxx
@@ -136,12 +136,12 @@ struct tpcPid {
       const std::string path = ccdbPath.value;
       const auto time = ccdbTimestamp.value;
       ccdb->setURL(url.value);
-      ccdb->setTimestamp(time);
+    //  ccdb->setTimestamp(time);
       ccdb->setCaching(true);
       ccdb->setLocalObjectValidityChecking();
       ccdb->setCreatedNotAfter(std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch()).count());
-      response.SetParameters(ccdb->getForTimeStamp<o2::pid::tpc::Response>(path, time));
-      LOGP(info, "Loading TPC response from CCDB, using path: {} for ccdbTimestamp {}", path, time);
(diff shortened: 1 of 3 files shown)


jezwilkinson:
Hi @njacazio ,
I am pretty sure that the source of the problem was the following:
- the response is initialised with the default object, which has a validity limit from timestamp 0 up until the year 2100
- the process reaches the track loop and the CCDB call with the current event's timestamp
- Because the caching is enabled, and the already loaded default object is still valid for the current event's timestamp (because it is valid for every event), the CCDB handler doesn't attempt to make a request to the CCDB for an updated object, even if a newer one exists for that timestamp.

This is all within the logic of the timestamp validity limits of the CCDB objects. If that isn't what the caching is doing and it instead just takes the currently loaded object as long as the first collision was fine, then I don't see how we can make it update to take later collisions into account at all without disabling the cache.
I'll add another commit to re-enable calling a specific timestamp for the full run if defined as >0, but for the main issue here this is the only way I can see right now to deal with it.

njacazio:
It's ok, it was just a suggestion. I have nothing to add them

jgrosseo:
@jezwilkinson I agree. The local caching assumes that the objects have no overlap. So one should not have this ever-valid default object (which I think is a mistake anyway because you do not see when the object is missing)...

jgrosseo:
I can merge this but I would actually prefer that you remove the code instead of commenting, what do you think?

jezwilkinson:
@jgrosseo Agreed, the ever-valid object is not ideal and I think intended mostly as a fallback in case a specific spline doesn't exist yet for a given run. This we can/should revisit.

I re-implemented the fixed timestamp access as suggested by @njacazio which also gets rid of the redundant comments - now the behaviour should be as intended (timestamp configurable -1 = latest existing object; 0 = collision-based timestamp; >0 = anchored to a specific fixed timestamp for the whole run)

jezwilkinson:
@njacazio @jgrosseo Would you consider the changes OK to merge?

PR #851, opened 2022-06-09
Template filling function and tpc+tof signals

njacazio:


njacazio:
@mfaggin

njacazio:
@arossi81 this should simplify a bit the filling, hopefully it will not mess your workflow, you can open another PR and we merge it first so that we are in sync

mario-krueger (approved):

PR #854, opened 2022-06-10
Filling of gen TH3D + selection on zvtx (hard coded just for the time being)

arossi81:


arossi81:
Hi @njacazio ,
we probably can directly filter collisions for zvtx (and other selections), but for the time being I added an hard-coded cut.

Ciao
Andrea

DPG/Tasks/qaEventTrackLite.cxx
```diff
@@ -163,6 +163,14 @@ struct qaEventTrackLite {
   template <bool isMC, typename trackType>
   void fillHistograms(trackType const& track)
   {
+
+    if (TMath::Abs(track.dpgCollision().posZ() > 10.))
```
njacazio, line 167: Could this be a configurable cut?

DPG/Tasks/qaEventTrackLite.cxx
```diff
@@ -163,6 +163,14 @@ struct qaEventTrackLite {
...
+      return;
+    if constexpr (isMC) {
+      if (track.productionMode() == 0) {
+        histos.fill(HIST("Particle/selPtEtaPhiMCGenPrimary"), track.ptMC(), track.etaMC(), track.phiMC());
```
njacazio, line 171: Would this not be filled twice? (line 286)

DPG/Tasks/qaEventTrackLite.cxx
```diff
@@ -267,6 +275,9 @@ struct qaEventTrackLite {
     }
 
     for (const auto& particle : particles) {
+      if (TMath::Abs(particle.dpgCollision().posZ() > 10.))
```
njacazio, line 278: Maybe we could use here the same configurable

njacazio (commented):
Thanks Andrea! I have just minor suggestions

DPG/Tasks/qaEventTrackLite.cxx
arossi81: yes, of course, and also done via a Filter, but if you do not mind, I'll change it on Monday (or asap)

njacazio: Ok!

njacazio: I realized this was also wrong, the abs was on top of the bool selection, I fixed that in the last commit

arossi81: ah, right thanks! Sorry. Instead, for the filling of the histo, that's correct. It's not the same histo and must be filled both in the loop over tracks and in that over nonreco particles.

DPG/Tasks/qaEventTrackLite.cxx
arossi81: This is the histo counting gen particles, it has to be filled both when looping over recoParticles and nonRecoparticles. Isn't that correct?

njacazio: Ah yes, you're right, apologies!
Ok approved for me!

njacazio (approved):

njacazio:
@mario-krueger I don't think I can approve yet, some git magic has to happen yet. Could you also have a look pls?

PR #858, opened 2022-06-11
Move to helper for lambdakzerobuilder too

ddobrigk:


ddobrigk:
This is just code cleanup / uniformity with respect to the cascadebuilder: here, I now use a CheckAndUpdate helper method too.

PWGLF/TableProducer/lambdakzerobuilder.cxx
```diff
@@ -179,6 +179,19 @@ struct lambdakzeroBuilder {
     return output;
   }
 
+  void CheckAndUpdate(Int_t lRunNumber, Int_t lTimeStamp)
```
njacazio, line 182: Shouldn't the timestamp be a uint_64?

njacazio (commented):
Thanks David!

PWGLF/TableProducer/lambdakzerobuilder.cxx
ddobrigk: You're right! Thanks, changed!

njacazio: Maybe also in yesterday's commit then!

ddobrigk: Um yes, sure, I changed it in both already just a moment ago :-D

njacazio: Ok great!

njacazio (approved):

PR #866, opened 2022-06-11
Added AliFemto files

zchochul:
Introduction of AliFemto package based on correlation measurements.

### DIFF
--- a/PWGCF/AliFemto/AliFemtoDerived.h
+++ b/PWGCF/AliFemto/AliFemtoDerived.h
@@ -1,0 +1,187 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+#ifndef ANALYSIS_TASKS_PWGCF_ALIFEMTO_ALIFEMTODERIVED_H_
+#define ANALYSIS_TASKS_PWGCF_ALIFEMTO_ALIFEMTODERIVED_H_
+
+#include "Framework/ASoA.h"
+#include "MathUtils/Utils.h"
+#include "Framework/DataTypes.h"
+#include "Common/DataModel/Multiplicity.h"
+#include "Framework/AnalysisDataModel.h"
+#include "Framework/Expressions.h"
+#include "Common/DataModel/TrackSelectionTables.h"
+#include "Common/Core/PID/PIDResponse.h"
+#include <cmath>
+
+namespace o2::aod
+{
+/// FemtoDreamCollision
+namespace femtodreamcollision
+{
+DECLARE_SOA_COLUMN(MultV0M, multV0M, float);       //! V0M multiplicity
+DECLARE_SOA_COLUMN(Sphericity, sphericity, float); //! Sphericity of the event
+} // namespace femtodreamcollision
+
+DECLARE_SOA_TABLE(FemtoDreamCollisions, "AOD", "FEMTODREAMCOLS",
+                  o2::soa::Index<>,
+                  o2::aod::collision::PosZ,
+                  femtodreamcollision::MultV0M,
+                  femtodreamcollision::Sphericity,
+                  timestamp::Timestamp);
+using FemtoDreamCollision = FemtoDreamCollisions::iterator;
+
+/// FemtoDreamTrack
+namespace femtodreamparticle
+{
+/// Distinuishes the different particle types
+enum ParticleType {
+  kTrack,           //! Track
+  kV0,              //! V0
+  kV0Child,         //! Child track of a V0
+  kCascade,         //! Cascade
+  kCascadeBachelor, //! Bachelor track of a cascade
+  kNParticleTypes   //! Number of particle types
+};
+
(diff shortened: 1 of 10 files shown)


PWGCF/AliFemto/AliFemtoDetaDphiStar.h
```diff
@@ -0,0 +1,204 @@
...
+// or submit itself to any jurisdiction.
+
+/// \file AliFemtoDetaDphiStar.h
+/// \brief AliFemtoDetaDphiStar - Checks particles for the close pair rejection.
+/// \author Laura Serksnyte, TU München, laura.serksnyte@tum.de
+
+#ifndef ANALYSIS_TASKS_PWGCF_ALIFEMTO_ALIFEMTODETADPHISTAR_H_
+#define ANALYSIS_TASKS_PWGCF_ALIFEMTO_ALIFEMTODETADPHISTAR_H_
+
+//#include "AliFemtoDerived.h"
```
saganatt, line 19: You don't use it, so please remove it

PWGCF/AliFemto/AliFemtoDerived.h
```diff
@@ -0,0 +1,187 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
```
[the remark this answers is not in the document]
saganatt: I see that you are actually doing 2a), not using AliFemtoDerived at all. Can you then remove it from this PR, please?

PWGCF/AliFemto/AliFemtoEventHisto.h
```diff
@@ -0,0 +1,60 @@
...
+// or submit itself to any jurisdiction.
+
+/// \file AliFemtoEventHisto.h
+/// \brief AliFemtoEventHisto - Histogram class for tracks, V0s and cascades
+/// \author Andi Mathis, TU München, andreas.mathis@ph.tum.de
+
+#ifndef ANALYSIS_TASKS_PWGCF_O2ALIFEMTO_INCLUDE_O2ALIFEMTO_ALIFEMTOEVENTHISTO_H_
+#define ANALYSIS_TASKS_PWGCF_O2ALIFEMTO_INCLUDE_O2ALIFEMTO_ALIFEMTOEVENTHISTO_H_
+
+//#include "AliFemtoDerived.h"
```
saganatt, line 19: Please remove, and also in other files

PWGCF/AliFemto/aliFemtoPairTaskTrackTrack.cxx
```diff
@@ -0,0 +1,258 @@
...
+#include "Framework/HistogramRegistry.h"
+#include "Framework/ASoAHelpers.h"
+#include "Framework/RunningWorkflowInfo.h"
+#include "Framework/StepTHn.h"
+#include <CCDB/BasicCCDBManager.h>
+#include "DataFormatsParameters/GRPObject.h"
+
+//#include "AliFemtoDerived.h"
+#include "PWGCF/DataModel/FemtoDerived.h"
+#include "AliFemtoParticleHisto.h" //new AliFemto name
```
saganatt, line 27: Please remove these comments with "new AliFemto name", I think the code will be clear without them and it is easier to compare with FemtoDream :-)

saganatt (changes requested):

jgrosseo:
Is this a clone of the FemtoDream code?

I thought we had agreed *not* to do that in the discussion with Adam and Lukasz?

jgrosseo:
In particular the DataModel files should not be copied because you need the same format... Otherwise FemtoDream and AliFemto will have each 50% less space for derived data...

saganatt:
1. I agree on the data model, Zuzanna already removed AliFemtoDerived.h
2. We understood it is a bad code design (I agree!) but allowed as long as we stick to the common derived data format, as the data and memory size is more important than the code itself. Is that right?

jgrosseo:
You can have your own code if you have the common format, but why duplicate the entire FemtoDream codebase? You can make your own code which reads the same format but duplicating is a very bad design pattern and what is the motivation?
In this case just make your own code from scratch using the same input data...

saganatt:
We talked about it.

1. We could write from scratch if we have time, but this code would anyways do exactly the same things wrt applying cuts and the base for an analysis task. We didn't take everything from Femto Dream, only the cutculator + basic analysis task together with dependencies as the starting point.
2. Why then not to use it from Femto Dream? Because we would like to control the code as well. As said, Łukasz and Zuzanna have been trying to proceed with Femto Dream since March and it was difficult for them to follow because the codebase is fluid, constantly changing. We don't want to rely our analysis on something that will from some point work differently. Similarly, FemtoDream people don't want us to edit directly their codes so as their analysis results won't be impacted, that's perfectly understandable.
3. Also, we might want to have different cuts, for example. We will add more own stuff like other correlation functions and so on. So this starting code will soon deviate more and more from the FemtoDream version and it will no longer be the same.

jgrosseo:
OK, I follow your arguments. If we keep the copied code minimal and only start copying additional code when needed...

There is another point though: you have to change the name. Ali in AliFemto means AliRoot. But there is no AliRoot anymore. Please not replace it by O2 because we do not want to have O2 in the name of O2Physics either. So you can see what other folder name / file name you want, then we can go ahead.

jgrosseo:
If you change the name, we can merge this... @saganatt Is there still discussing ongoing on this?

saganatt:
Yes, Zuzanna will do it soon (at the weekend, I suppose)

victor-gonzalez:
@jgrosseo @saganatt We agreed a while ago in having the table producer tasks under `TableProducer` and the analysis tasks under  `Tasks`
Should we adhere to that? or should we create a similar layout under each framework?

jgrosseo:
You can recreate that structure under each PAG for example (HF has done this).

PR #881, opened 2022-06-15
Fix MC QA PID

njacazio:


njacazio:
Credits to @vkucera that kindly reported this.
This was a bug that prevented us to run with MC QA for PID, now it should be ok!
A) MC collision table was not subscribed to
B) particles of version 000 were used

njacazio:
@giovannimalfattore once this is merged, could you setup the dpg trains to include the PID QC on MC ? Thanks!

alibuild (approved):

PR #893, opened 2022-06-16
Fix mch match indices in data reduced tables

batoul-d:


PWGDQ/DataModel/ReducedInfoTables.h
```diff
@@ -258,7 +258,7 @@ DECLARE_SOA_TABLE(ReducedMuons, "AOD", "RTMUON", //!
 DECLARE_SOA_TABLE(ReducedMuonsExtra, "AOD", "RTMUONEXTRA", //!
                   fwdtrack::NClusters, fwdtrack::PDca, fwdtrack::RAtAbsorberEnd,
                   fwdtrack::Chi2, fwdtrack::Chi2MatchMCHMID, fwdtrack::Chi2MatchMCHMFT,
-                  fwdtrack::MatchScoreMCHMFT, fwdtrack::MFTTrackId, fwdtrack::MCHTrackId,
+                  fwdtrack::MatchScoreMCHMFT, fwdtrack::MFTTrackId, reducedmuon::MCHTrackId,
```
iarsene, line 261: Since its useless, and actually problematic when merging, please remove the MFTTrackId, as we discussed during the last meeting.

PWGDQ/TableProducer/tableMaker.cxx
```diff
@@ -350,10 +381,32 @@ struct TableMaker {
...
+          if (newMatchIndex.count(muon.index()) == 0) { // if an entry for this mch was not added it simply mean that non of the global tracks were matched to it
+            newMatchIndex[muon.index()] = -1;
+          }
+        }
+
         muonBasic(event.lastIndex(), trackFilteringTag, muon.pt(), muon.eta(), muon.phi(), muon.sign());
         muonExtra(muon.nClusters(), muon.pDca(), muon.rAtAbsorberEnd(),
                   muon.chi2(), muon.chi2MatchMCHMID(), muon.chi2MatchMCHMFT(),
-                  muon.matchScoreMCHMFT(), muon.matchMFTTrackId(), muon.matchMCHTrackId(), muon.mchBitMap(), muon.midBitMap(), muon.midBoards(), muon.trackType());
+                  muon.matchScoreMCHMFT(), muon.matchMFTTrackId(), newMatchIndex.find(muon.index())->second, muon.mchBitMap(), muon.midBitMap(), muon.midBoards(), muon.trackType());
```
iarsene, line 409: remove the matchMFTTrackId()

iarsene (dismissed):

PWGDQ/TableProducer/tableMaker.cxx
batoul-d: done

iarsene (approved):

PWGDQ/DataModel/ReducedInfoTables.h
```diff
@@ -242,7 +242,7 @@ DECLARE_SOA_DYNAMIC_COLUMN(MIDBoardCh3, midBoardCh3, //!
                            [](uint32_t midBoards) -> int { return static_cast<int>((midBoards >> 16) & 0xFF); });
 DECLARE_SOA_DYNAMIC_COLUMN(MIDBoardCh4, midBoardCh4, //!
                            [](uint32_t midBoards) -> int { return static_cast<int>((midBoards >> 24) & 0xFF); });
-
+DECLARE_SOA_SELF_INDEX_COLUMN_FULL(MCHTrack, matchMCHTrack, int, "Muons_MatchMCHTrack");
```
jgrosseo, line 245: I don' think this nomenclature will work with the aod merger which means that you cannot merge derived data (and therefore not produce it on hyperloop). This is because here you call it `Muons` and below it is called actually `RTMUON`.

Can you test this by producing two derived AODs (from two different input files) and the merge it using `o2-aod-merger`?

Can you test this by merging two

jgrosseo: @iarsene @batoul-d

iarsene: @jgrosseo  From my discussion with @batoul-d, she tested this, but can you confirm?

jgrosseo: A merging test is now also included in Hyperloop but this issue is not flagged as error, you should see a warning in the output though. If you made a test, you can send me the stdout.log and I can check (you find this at the end).

batoul-d: the name of this branch is indeed a problem for the merging. This pull request does not try to fix this issue but focuses on filling the index correctly. The compatibility of indices' names in the reduced files concerns all the indices (self- or not) so the next step is to fix that. From what I saw already, the change needs to be done on both sides, the branches and the tool itself. For example the tool explicitly adds the label"O2" when it tries to find the tree to which the index is pointing to (https://github.com/AliceO2Group/O2Physics/blob/master/Common/Core/aodMerger.cxx#L61) which is the naming convention used in aod but it's not used in the reduced files where instead we add the label "Reduced".

jgrosseo: O2 is the suffix of all derived tables, so that should be fine. It is added automatically. How did you get "Reduced"?

jgrosseo: You can follow the example here:
https://github.com/AliceO2Group/O2Physics/blob/master/PWGCF/DataModel/CorrelationsDerived.h

Important is simply that the first and third parameter are the same apart from the capitalization and the "s". Note that this limits the length of your classnames to 14 characters.

batoul-d: I see. Then I think the problem is when we want to save our tables we explicitly give them labels that do not follow this naming convention. Since I am still new to this I do not have all details but I will look into it and fix it in a new PR

PR #894, opened 2022-06-17
DPG: disable debug LOG(info) by default.

mfaggin:


njacazio:
Ciao Mattia, I'm not sure this is the best way, can't we have instead `info` moved to `debug`? then we move it back if needed. We will anyway not run soon on hyperloop so we might not need to read logs

njacazio:
i.e. `LOG(info)` -> `LOG(debug)`

njacazio:
Ah wait, I thought this was the lite task, I apologize!

njacazio (approved):

PR #905, opened 2022-06-21
Fixing default binning of a TH3 (too large size), implementing method…

arossi81:
…s for default selections Sel1,2,3.


DPG/Tasks/qaEventTrackLite.cxx
```diff
@@ -133,7 +152,7 @@ struct qaEventTrackLite {
     histos.add("Tracks/relativeResoPtMeanvsEtavsPt", "mean relative #it{p}_{T} resolution;", kTProfile2D, {axisPt, axisEta});
     histos.add("Tracks/reso1overPtMeanvsEtavs1overPt", "mean 1/#it{p}_{T} resolution;", kTProfile2D, {axis1overPt, axisEta});
     // track cuts map
-    histos.add("TrackCuts/selPtEtaPhiNoSel", "pt eta phi map no el; pt,eta,phi", kTH3D, {axisPt, {50, -1.2, 1.2, "#eta"}, {30, 0., 2 * M_PI, "#varphi"}});
```
njacazio, line 136: Hi @arossi81 maybe we could also use TH3F instead?

arossi81: Hi @njacazio,
that was not the histo causing problems. I think TH3D is ok for it.
Cheers,
Andrea

njacazio (approved):

PR #920, opened 2022-06-24
[PWGHF] Add code for cut-variation method 

fgrosa:
This PR includes:
- A new directory for the analysis macros
- A class for the computation of the data-driven (non-)prompt fraction of charm hadrons with the cut-variation method
- A script that uses the class for the cut-variation method
- An example of configuration file
- A module with helper functions to set the plotting style
- A brief README

The code was tested with Run2 inputs for the D+ in pp@13 TeV and it was verified to give the same results (see figures below for one pt interval).

@fcatalan92 @stefanopolitano please check the code and see if there is something that has to be changed / improved.
@xinyepeng @lvermunt @strogolo please check whether the structure for the macros is fine with you.
Thanks a lot!

image
image
image
image


lvermunt:
Thanks a lot @fgrosa! To me the PR and the structure of the macros look good. I would just specify "(non-)prompt fraction via cut-variation method" in the section header in the README, so (when this README gets long) it's also right away clear for non-experienced people

fgrosa:
> Thanks a lot @fgrosa! To me the PR and the structure of the macros look good. I would just specify "(non-)prompt fraction via cut-variation method" in the section header in the README, so (when this README gets long) it's also right away clear for non-experienced people

Hi @lvermunt thanks for your feedback! I have fixed the title in the README as you suggested 👍

fgrosa:
@xinyepeng @strogolo @fcatalan92 did you have time to look at this PR? Is it fine for you? Thanks!

xinyepeng:
@fgrosa Thanks a lot! It's very nice that the cut variation method can be included in O2Physics, I think it's fine as the first version, green light from my side

lvermunt (approved):

PR #922, opened 2022-06-26
O2femto code

zchochul:


jgrosseo:
Maybe my comment was not clear here ;-) but I wrote "Please not replace it by O2 because we do not want to have O2 in the name of O2Physics either": https://github.com/AliceO2Group/O2Physics/pull/866#issuecomment-1155076561

jgrosseo:
At the limit just call it "Femtoscopy"
Also your PR contains now both folders (AliFemto and O2Femto).

PR #927, opened 2022-06-27
Mean and resolution correction from neural networks for the TPC PID response

ChSonnabend:
This PR intends to bring the following features to O2Physics:
- Mean and resolution correction to a Bethe-Bloch parametrisation
- Automatic, runnumber-based fetching of the neural network files from AliEn

### DIFF
--- a/PWGLF/TableProducer/lambdakzerobuilder.cxx
+++ b/PWGLF/TableProducer/lambdakzerobuilder.cxx
@@ -115,8 +115,7 @@ struct lambdakzeroBuilder {
     "registry",
     {
       {"hEventCounter", "hEventCounter", {HistType::kTH1F, {{1, 0.0f, 1.0f}}}},
-      {"hV0Candidate", "hV0Candidate", {HistType::kTH1F, {{2, 0.0f, 2.0f}}}},
-      {"hGoodIndices", "hGoodIndices", {HistType::kTH1F, {{4, 0.0f, 4.0f}}}},
+      {"hV0Criteria", "hV0Criteria", {HistType::kTH1F, {{10, 0.0f, 10.0f}}}},
     },
   };
 
@@ -227,7 +226,8 @@ struct lambdakzeroBuilder {
 #endif
       MY_DEBUG_MSG(isK0SfromLc, LOG(info) << "V0 builder: found K0S from Lc, posTrack --> " << labelPos << ", negTrack --> " << labelNeg);
 
-      registry.fill(HIST("hGoodIndices"), 0.5);
+      //value 0.5: any considered V0
+      registry.fill(HIST("hV0Criteria"), 0.5);
       if (isRun2) {
         if (!(V0.posTrack_as<MyTracks>().trackType() & o2::aod::track::TPCrefit)) {
           MY_DEBUG_MSG(isK0SfromLc, LOG(info) << "posTrack " << labelPos << " has no TPC refit");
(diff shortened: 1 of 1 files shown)


ChSonnabend:
Hi @njacazio , the eggs are in the basket! :-)

Common/TableProducer/PID/pidTPC.cxx
```diff
@@ -68,18 +68,20 @@ struct tpcPid {
   o2::pid::tpc::Response* responseptr = nullptr;
   // Network correction for TPC PID response
   Network network;
+  std::string networkPathOnAlien = ""; // Will be set by the user or automatically if TPCPIDResponse object is fetched from CCDB
```
njacazio, line 71: Maybe you can consider using a configurable and having this set to override the choice via CCDB e.g. CCDB is used only if this is left empty

Common/TableProducer/PID/pidTPC.cxx
```diff
@@ -68,18 +68,20 @@ struct tpcPid {
...
 
   // Input parameters
   Service<o2::ccdb::BasicCCDBManager> ccdb;
   Configurable<std::string> paramfile{"param-file", "", "Path to the parametrization object, if emtpy the parametrization is not taken from file"};
   Configurable<std::string> url{"ccdb-url", "http://alice-ccdb.cern.ch", "url of the ccdb repository"};
   Configurable<std::string> ccdbPath{"ccdbPath", "Analysis/PID/TPC/Response", "Path of the TPC parametrization on the CCDB"};
   Configurable<long> ccdbTimestamp{"ccdb-timestamp", 0, "timestamp of the object used to query in CCDB the detector response. Exceptions: -1 gets the latest object, 0 gets the run dependent timestamp"};
+  o2::ccdb::CcdbApi ccdb_api; /// API to access CCDB headers
```
njacazio, line 79: So you need the api to access the headers? And you cannot use the `BasicCCDBManager` `ccdb` right?

njacazio: e.g. as in https://github.com/AliceO2Group/O2Physics/blob/master/Common/TableProducer/timestamp.cxx#L35

Common/TableProducer/PID/pidTPC.cxx
```diff
@@ -170,36 +199,50 @@ struct tpcPid {
     if (useNetworkCorrection) {
 
       auto start_overhead = std::chrono::high_resolution_clock::now();
-      std::vector<float> track_properties;
+
+      // Defining some network parameters
+      int inputDimensions = network.getInputDimensions();
+      int outputDimensions = network.getOutputDimensions();
+      const unsigned long track_prop_size = tracks.size() * 9;
```
njacazio, line 206: This is for every particle species, right? Would it require less space if one was interested in e.g. only pions and kaons?

njacazio (commented):
Hi @ChSonnabend this looks good to me, just minor suggestions/questions

Common/TableProducer/PID/pidTPC.cxx
ChSonnabend: Indeed that would be the case, however the correction is performed per track and since we don't know a priori which track has which species associated with it, we would need to produce the correction for each mass hypotheses. If we consider that the analyser might already trust the initial Bethe-Bloch parametrisation and assigns the mass hypotheses based on the distance to the nearest BB-curve, then indeed, we could save a lot of memory! The question is whether we want to make assumptions already before the correction is applied (this might lead to misidentification (on a small scale I guess), but would save a lot of memory of course). If this hypotheses would be used, then we would only calculate for each track a single mass hypotheses, thus reducing memory consumption by a factor 9.

Common/TableProducer/PID/pidTPC.cxx
ChSonnabend: Indeed this is exactly what is being done. This is just a global variable which is changed according to the configurable networkPathAlien. If the configurable is left empty, the objects are automatically fetched using the runnumber from the TPC PID response object (metadata). If this configurable is set however, the given path on AliEn is used to fetch the network file.

PR #931, opened 2022-06-28
Event counters in event selection

ekryshen:


jgrosseo (commented):
Thanks!
Do you know it this dynamic run number on the axis merge properly in ROOT? My understanding from the past is that it doesn't (it merges simply each first bin together, and in each run the first bin corresponds to a different run.)

ekryshen:
@jgrosseo I've just checked with hadd. The output was merged properly, new bin was added for additional run.

jgrosseo (approved):

PR #932, opened 2022-06-28
Use compilation job pool for linking as well

TimoWilken:
Instead of only using it for compilation. Hopefully, this will keep memory use during compilation within available limits.


jgrosseo (approved):

ktf (approved):

TimoWilken:
CI is green, so I've merged. Let's see how it fares in the daily build tonight.

PR #939, opened 2022-07-01
[PWGHF]: Fix search of mothers to work also at partonic level

fgrosa:
This PR fixes the search of the mother particles, in particular when needed at partonic level (i.e. non-prompt tagging for HF analyses). In the previous version, only the first element of the mothers vector was checked, however most of the times the beauty quark is not in the first element. This was the origin of the issue related to the non-prompt tagging found in [O2-2761](https://alice.its.cern.ch/jira/browse/O2-2761), now fixed with this PR.

### DIFF
--- a/Common/Core/RecoDecay.h
+++ b/Common/Core/RecoDecay.h
@@ -587,7 +587,8 @@ class RecoDecay
 
     // vector of vectors with mother indices; each line corresponds to a "stage"
     std::vector<std::vector<long int>> arrayIds{};
-    arrayIds.push_back(std::vector{particle.globalIndex()}); // the first vector contains the index of the original particle
+    std::vector<long int> initVec{particle.globalIndex()};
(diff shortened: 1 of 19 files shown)


vkucera:
Thanks for the fix. Can you please add comments in the code to help understand how you use the new variables?

fgrosa:
> Thanks for the fix. Can you please add comments in the code to help understand how you use the new variables?

Sure, I added them! Let me know if something is still unclear

vkucera (approved):

vkucera:
@jgrosseo @iarsene It is fine with me to merge this.

fgrosa:
> @jgrosseo @iarsene It is fine with me to merge this.

Thanks @vkucera! I noticed now that the test on MacOs became red, but the error is unrelated to this PR, can we merge it anyway? It would be good to have it merged because the current version on Run3 MC productions leads to a wrong result

jgrosseo (approved):

PR #940, opened 2022-07-01
[PWGHF] Add histogram of decay length at gen level

fgrosa:
This PR adds a TH2 with decay length vs hadron species, useful to debug the charm-hadron origin tag

### DIFF
--- a/PWGHF/Tasks/HFMCValidation.cxx
+++ b/PWGHF/Tasks/HFMCValidation.cxx
@@ -184,7 +184,7 @@ struct ValidationGenLevel {
 
           auto decayLength = RecoDecay::distance(vertexPrimary, vertexDau);
           if (whichHadron == 1 && origin == OriginType::Prompt && decayLength * 10000 > 100) {
-            RecoDecay::getMother(particlesMC, particle, kBottom, true, nullptr, -1, true);
+            RecoDecay::getMother(particlesMC, particle, kBottom, true);
           }
           //Filling histograms with per-component momentum conservation
           registry.fill(HIST("hMomentumCheck"), float(momentumCheck));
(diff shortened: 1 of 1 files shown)


vkucera:
Please add the PWGHF: prefix in the commit title when your PR has only one commit as the PR title is ignored in such a case.

PWGHF/Tasks/HFMCValidation.cxx
```diff
@@ -174,6 +178,14 @@ struct ValidationGenLevel {
           }
           double pDiff = RecoDecay::p(pxDiff, pyDiff, pzDiff);
           double ptDiff = RecoDecay::pt(pxDiff, pyDiff);
+          auto daughter0 = particle.daughters_as<aod::McParticles>().begin();
+          double vertexDau[3] = {daughter0.vx(), daughter0.vy(), daughter0.vz()};
+          double vertexPrimary[3] = {mccollision.posX(), mccollision.posY(), mccollision.posZ()};
+
+          auto decayLength = RecoDecay::distance(vertexPrimary, vertexDau);
+          if (whichHadron == 1 && origin == OriginType::Prompt && decayLength * 10000 > 100) {
+            RecoDecay::getMother(particlesMC, particle, kBottom, true, nullptr, -1, true);
```
vkucera, line 187: The last `true` does not match any argument of `getMother`.
Also you don't do anything with the returned index.

vkucera (changes requested):

PWGHF/Tasks/HFMCValidation.cxx
fgrosa: You are right, I added a feature for debugging and I forgot to remove it, now fixed 👍

vkucera (approved):

PR #944, opened 2022-07-03
Automatic fetching of neural networks from CCDB for the TPC PID Response correction

ChSonnabend:
The neural networks will be used for the TPC PID response correction. Automatic fetching is necessary to run jobs over multiple run numbers etc. and has to be done on the fly on Hyperloop. This PR brings that functionality.

### DIFF
--- a/Common/Core/CMakeLists.txt
+++ b/Common/Core/CMakeLists.txt
@@ -23,6 +23,7 @@ o2physics_target_root_dictionary(AnalysisCore
                         PID/DetectorResponse.h
                         PID/PIDTOF.h
                         PID/TOFReso.h
+                        PID/TPCPIDResponse.h
               LINKDEF AnalysisCoreLinkDef.h)
 
 o2physics_add_executable(merger
--- a/ALICE3/Tasks/alice3-cdeuteron.cxx
+++ b/ALICE3/Tasks/alice3-cdeuteron.cxx
@@ -172,13 +172,15 @@ struct Alice3CDeuteron {
 #undef MakeHistos
   }
 
+  Preslice<aod::McParticles_000> perMcCollision = aod::mcparticle::mcCollisionId;
+
   void process(const soa::Join<o2::aod::Collisions, o2::aod::McCollisionLabels>::iterator& coll,
                const o2::aod::McCollisions& Mccoll,
                const soa::Join<o2::aod::Tracks, o2::aod::McTrackLabels, o2::aod::TracksExtra, o2::aod::TracksCov,
                                aod::pidTOFFullPi, aod::pidTOFFullKa, aod::pidTOFFullDe>& tracks,
                const aod::McParticles_000& mcParticles)
   {
-    const auto particlesInCollision = mcParticles.sliceBy(aod::mcparticle::mcCollisionId, coll.mcCollision().globalIndex());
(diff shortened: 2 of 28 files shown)


ChSonnabend:
Whenever you have time, you can have a look @njacazio . This now fixes the behaviour for automatic fetching across different run number.

jgrosseo (changes requested):
Sorry, this is not allowed... Those objects have to go to CCDB. We do not allow direct alien accesses as this can lead to overloads if there are not sufficient replicas...

ChSonnabend:
But can the .onnx files be loaded into an object from CCDB?

njacazio:
Hi @jgrosseo first of all, of course we are aware and we agree with you. However I think that we cannot store now .onnx files as this is not supported yet on CCDB. What do you suggest then?

jgrosseo:
This cannot even be a c++ class, right? Hadn't we discussed in the past, I don't mean with the two of you but to solve this use case, to add a CCDB API function `CopyFile` which copies a file from a CCDB location to a local file and then it can be used? It may be that no one implemented this up to now, though. @sawenzel do you remember?
In principle CCDB is not bound to ROOT or C++ streamable objects I think. One can store a binary blob.

njacazio:
> This cannot even be a c++ class, right? Hadn't we discussed in the past, I don't mean with the two of you but to solve this use case, to add a CCDB API function `CopyFile` which copies a file from a CCDB location to a local file and then it can be used? It may be that no one implemented this up to now, though. @sawenzel do you remember?
> In principle CCDB is not bound to ROOT or C++ streamable objects I think. One can store a binary blob.

Indeed the storing part is without issues, the issues are when retrieving the object.
As you say we could download the blob and then open it, let's see what Sandro says

njacazio:
Hi, thanks a lot for your suggestions @jgrosseo ! Indeed with the CCDB API the blob can be retrieved and loaded in the workflow from a local file, this way we avoid both alien and CVMFS (that was another suggested way to go). I did a preliminary test and it seems to work, I let @ChSonnabend comment on the feasibility! Thanks again!

jgrosseo:
So it already works with the existing CCDB API? I was about to make a JIRA ticket for the needed functionality in the API, but if you say it works, I wait a bit...

Common/Core/CMakeLists.txt
```diff
@@ -23,6 +23,7 @@ o2physics_target_root_dictionary(AnalysisCore
                         PID/DetectorResponse.h
                         PID/PIDTOF.h
                         PID/TOFReso.h
+                        PID/TPCPIDResponse.h
```
jgrosseo, line 26: I think you need to rebase...

jgrosseo:
OK, great! I think you should also change the title then...

ChSonnabend:
Hi @jgrosseo and @njacazio,
I think now it should be done, a simple fetch & merge from master was enough.

jgrosseo (approved):

vkucera:
Merging this broke the macOS build for all subsequent PRs. @ChSonnabend Please fix it.

ChSonnabend:
> Merging this broke the macOS build for all subsequent PRs. @ChSonnabend Please fix it.

It is fixed already with the latest bug-fix PR I did

vkucera:
> > Merging this broke the macOS build for all subsequent PRs. @ChSonnabend Please fix it.
> 
> It is fixed already with the latest bug-fix PR I did

Thanks

PR #945, opened 2022-07-04
Implement setter function to check MC signal from mother to daughter

feisenhu:
Implemented a setter function in the MCprong files to give the option, when defining a MCsignal in the MCSignalLibrary.h to check the MC signal from the mother to the daughter (In time).

Therefore the MC signal needs the mother pdg code as first entry and its daughter as second:
e.g. 
```
  if (!nameStr.compare("LFQdecayToE")) {
    MCProng prong(2, {900, 11}, {true, true}, {false, false}, {0, 0}, {0, 0}, {false, false});
    prong.SetSignalInTime(0, true);                                                    // set direction to check for daughters (true, in time) or for mothers (false, back in time)
    signal = new MCSignal(name, "LF meson  + quarkonia decays into e", {prong}, {-1}); // pi0,eta,eta',rho,omega,phi,jpsi,psi2s mesons
    return signal;
  }
```

The CheckProng function in MCSignal.h then checks the boolean `fCheckGenerationsInTime` whether we have to check in the next generation for a mother or daughters.

If we look for daughters, we check if there is one, and if so if it has a pdg code of the next generation.
If it is true it uses the first daughter that fulfils that requirement and sets it as current particle to check for further generations.

(Note: There is no option currently to check multiple daughters in the same generation that would fulfil the MC signal.)

### DIFF
--- a/PWGDQ/Core/MCProng.cxx
+++ b/PWGDQ/Core/MCProng.cxx
@@ -104,7 +104,7 @@ void MCProng::SetUseANDonSourceBits(int generation, bool option /*=true*/)
 }
 
 //________________________________________________________________________________________________________________
-void MCProng::SetDirectionOfGeneration(int generation, bool intime /*=false*/)
+void MCProng::SetSignalInTime(int generation, bool intime /*=false*/)
 {
   if (generation < 0 || generation >= fNGenerations) {
     return;
--- a/PWGDQ/Core/MCProng.cxx
+++ b/PWGDQ/Core/MCProng.cxx
@@ -18,6 +18,7 @@ ClassImp(MCProng);
 
 //________________________________________________________________________________________________________________
 MCProng::MCProng() : fNGenerations(0),
+                     fCheckGenerationsInTime(0),
                      fPDGcodes({}),
                      fCheckBothCharges({}),
                      fExcludePDG({}),
@@ -28,7 +29,8 @@ MCProng::MCProng() : fNGenerations(0),
 }
 
 //________________________________________________________________________________________________________________
-MCProng::MCProng(int n) : fNGenerations(n)
+MCProng::MCProng(int n) : fNGenerations(n),
+                          fCheckGenerationsInTime(0)
 {
   fPDGcodes.reserve(n);
   fCheckBothCharges.reserve(n);
@@ -49,13 +51,14 @@ MCProng::MCProng(int n) : fNGenerations(n)
 //________________________________________________________________________________________________________________
 MCProng::MCProng(int n, const std::vector<int> pdgs, const std::vector<bool> checkBothCharges, const std::vector<bool> excludePDG,
                  const std::vector<uint64_t> sourceBits, const std::vector<uint64_t> excludeSource,
-                 const std::vector<bool> useANDonSourceBitMap) : fNGenerations(n),
-                                                                 fPDGcodes(pdgs),
-                                                                 fCheckBothCharges(checkBothCharges),
-                                                                 fExcludePDG(excludePDG),
-                                                                 fSourceBits(sourceBits),
-                                                                 fExcludeSource(excludeSource),
-                                                                 fUseANDonSourceBitMap(useANDonSourceBitMap){};
+                 const std::vector<bool> useANDonSourceBitMap, bool checkGenerationsInTime) : fNGenerations(n),
+                                                                                              fPDGcodes(pdgs),
+                                                                                              fCheckBothCharges(checkBothCharges),
+                                                                                              fExcludePDG(excludePDG),
+                                                                                              fSourceBits(sourceBits),
+                                                                                              fExcludeSource(excludeSource),
+                                                                                              fUseANDonSourceBitMap(useANDonSourceBitMap),
+                                                                                              fCheckGenerationsInTime(checkGenerationsInTime){};
 
 //________________________________________________________________________________________________________________
 void MCProng::SetPDGcode(int generation, int code, bool checkBothCharges /*= false*/, bool exclude /*= false*/)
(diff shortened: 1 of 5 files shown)


PWGDQ/Core/MCProng.h
```diff
@@ -79,14 +79,15 @@ class MCProng
...
-          std::vector<uint64_t> sourceBits, std::vector<uint64_t> excludeSource, std::vector<bool> useANDonSourceBitMap);
+          std::vector<uint64_t> sourceBits, std::vector<uint64_t> excludeSource, std::vector<bool> useANDonSourceBitMap, bool fCheckGenerationsInTime = false);
   MCProng(const MCProng& c) = default;
   virtual ~MCProng() = default;
 
   void SetPDGcode(int generation, int code, bool checkBothCharges = false, bool exclude = false);
   void SetSources(int generation, uint64_t bits, uint64_t exclude = 0, bool useANDonSourceBits = true);
   void SetSourceBit(int generation, int sourceBit, bool exclude = false);
   void SetUseANDonSourceBits(int generation, bool option = true);
+  void SetSignalInTime(int generation, bool intime = false); // set variable to check generations in time or back in time (default)
```
iarsene, line 90: This function does not need the first parameter (generation), since the rule to go back in time or in time, holds for the entire prong. So please remove it, only the "intime" parameter is used

PWGDQ/Core/MCProng.cxx
```diff
@@ -100,13 +103,22 @@ void MCProng::SetUseANDonSourceBits(int generation, bool option /*=true*/)
   fUseANDonSourceBitMap[generation] = option;
 }
 
+//________________________________________________________________________________________________________________
+void MCProng::SetSignalInTime(int generation, bool intime /*=false*/)
```
iarsene, line 107: The "generation" parameter is not needed since the "intime" holds for the whole prong history

PWGDQ/Core/MCProng.cxx
```diff
@@ -100,13 +103,22 @@ void MCProng::SetUseANDonSourceBits(int generation, bool option /*=true*/)
...
+{
+  if (generation < 0 || generation >= fNGenerations) {
+    return;
+  }
```
iarsene, line 111: these lines can be removed

PWGDQ/Core/MCSignal.h
```diff
@@ -147,17 +147,36 @@ bool MCSignal::CheckProng(int i, bool checkSources, const U& mcStack, const T& t
...
+      // make sure that a daughter exists in the stack before moving one generation younger
+      if (!currentMCParticle.has_daughters() && j < fProngs[i].fNGenerations - 1) {
+        return false;
+      }
+      if (currentMCParticle.has_daughters() && j < fProngs[i].fNGenerations - 1) {
+        const auto& daughtersSlice = currentMCParticle.template daughters_as<U>();
+        for (auto& d : daughtersSlice) {
+          if (abs(d.pdgCode()) == fProngs[i].fPDGcodes[j + 1]) {
+            currentMCParticle = d;
+            break;
```
iarsene, line 176: Here you check the absolute of the PDG code, however, the user has the option to choose between checking both charges (using the abs of the PDG code), or not. So the latter situation is overwritten here. Could you fix it? Ideally one should use here the MCProng::TestPDG() since that will take into account automatically these options.

PWGDQ/Core/MCSignal.h
```diff
@@ -202,23 +221,47 @@ bool MCSignal::CheckProng(int i, bool checkSources, const U& mcStack, const T& t
...
+        if (currentMCParticle.has_daughters() && j < fProngs[i].fNGenerations - 1) {
+          // currentMCParticle = currentMCParticle.template mothers_first_as<U>();
+          // currentMCParticle = mcStack.iteratorAt(currentMCParticle.mothersIds()[0]);
+        }
+        if (currentMCParticle.has_daughters() && j < fProngs[i].fNGenerations - 1) {
+          const auto& daughtersSlice = currentMCParticle.template daughters_as<U>();
+          for (auto& d : daughtersSlice) {
+            if (abs(d.pdgCode()) == fProngs[i].fPDGcodes[j + 1]) {
+              currentMCParticle = d;
+              break;
```
iarsene, line 260: same comment as above

feisenhu:
@iarsene 
Tank you for your feedback.
I have included your comments:
- removed the `generation` parameter in the SetSignalInTime() function
- added a comment for the else {} case, to point out that branch is changing to daughter prong which fulfils the pdg requirement first
- using the TestPDG() function to check if the daughter PDG is in the next generation
   I now use `fProngs[i].TestPDG(j + 1, d.pdgCode())` to check that daughter has correct pdg code.
   
In case there is something else that I should change please let me know.
kind regards

PWGDQ/Core/MCSignal.h
```diff
@@ -202,23 +221,48 @@ bool MCSignal::CheckProng(int i, bool checkSources, const U& mcStack, const T& t
...
+        // move one generation further in history
+        // pong history will be moved to the branch of the first daughter that matches the PDG requirement
+        // make sure that a daughter exists in the stack before moving one generation younger
+        if (!currentMCParticle.has_daughters() && j < fProngs[i].fNGenerations - 1) {
+          return false;
+        }
+        if (currentMCParticle.has_daughters() && j < fProngs[i].fNGenerations - 1) {
+          // currentMCParticle = currentMCParticle.template mothers_first_as<U>();
+          // currentMCParticle = mcStack.iteratorAt(currentMCParticle.mothersIds()[0]);
+        }
```
iarsene, line 255: This if statement doesn't do anything, should be removed

iarsene (approved):

iarsene:
Hi @feisenhu,
Thanks for the changes. There is just a small issue i still commented on, otherwise I am fine with the changes.

PWGDQ/Core/MCSignal.h
feisenhu: You are right. Thank you for pointing out. :)
I have removed these lines.

feisenhu:
> Hi @feisenhu,
Thanks for the changes. There is just a small issue i still commented on, otherwise I am fine with the changes.

You are right. Thank you for pointing out. :)
I have removed these lines.

iarsene (approved):

PR #948, opened 2022-07-04
Fix histogram names for processRun3

ercolessi:
Hi @ddobrigk, I think the changes of this commit should be propagated also to processRun3 otherwise it crashes https://github.com/AliceO2Group/O2Physics/commit/5ffe363208e4f66044be0852983e13d4dfa13a5f

Is that ok?


njacazio:
Hi @ddobrigk what should we do?

ddobrigk:
Hi @ercolessi , you're completely correct. However, I think we should use the same filling scheme for both processes (and not just change the names)... Could you please still adjust accordingly? Super thanks!!!

ercolessi:
Hi @ddobrigk you're right, now should be fixed!

ddobrigk:
Thanks a lot @ercolessi ! I think this looks good now!! Nicolo, can you please merge?

njacazio (approved):

PR #950, opened 2022-07-05
PWGMM: dndeta: transpose response; fill response only for reconstructed events

aalkin:


alibuild (approved):

PR #960, opened 2022-07-06
[Event filtering] Remove DCA cut from global track requirement for beauty triggers

fgrosa:


alibuild (approved):

PR #966, opened 2022-07-08
[Common, PWGHF, Event Filtering] Implement specific method to tag prompt/nonprompt charm hadrons

fgrosa:
This PR implements a specific method to tag prompt/non-prompt charm hadrons, as discussed in [Nonprompt_tagging_fgrosa_2022Jul04.pdf](https://indico.cern.ch/event/1178012/contributions/4948661/attachments/2474891/4246862/Nonprompt_tagging_fgrosa_2022Jul04.pdf). 

I just have a doubt regarding the output values of the `checkCharmHadronOrigin` function: they correspond to the `hf_cand::OriginType` enum, but are independently defined since the HF code should not be included in the common one. A possible solution to have it uniquely defined would be to move the `hf_cand::OriginType` enum in the Common code, or alternatively to move the method `checkCharmHadronOrigin` in the HF code (however it could also be useful for DQ, since it can be used to tag charmonia as well). @vkucera do you have suggestions?

### DIFF
--- a/Common/Core/RecoDecay.h
+++ b/Common/Core/RecoDecay.h
@@ -950,14 +950,8 @@ class RecoDecay
 
             if (!searchUpToQuark) {
               if (
-                (PDGParticleIMother / 100 == 5 ||            // b mesons
-                 PDGParticleIMother / 1000 == 5 ||           // b baryons
-                 (PDGParticleIMother - 10000) / 100 == 5 ||  // bbbar resonances
-                 (PDGParticleIMother - 20000) / 100 == 5 ||  // bbbar resonances
-                 (PDGParticleIMother - 30000) / 100 == 5 ||  // bbbar resonances
-                 (PDGParticleIMother - 100000) / 100 == 5 || // bbbar resonances
-                 (PDGParticleIMother - 200000) / 100 == 5 || // bbbar resonances
-                 (PDGParticleIMother - 300000) / 100 == 5)   // bbbar resonances
+                (PDGParticleIMother / 100 == 5 || // b mesons
+                 PDGParticleIMother / 1000 == 5)  // b baryons
               ) {
                 return 2;
               }
--- a/Common/Core/RecoDecay.h
+++ b/Common/Core/RecoDecay.h
@@ -937,7 +937,7 @@ class RecoDecay
   template <typename T>
   static int getCharmHadronOrigin(const T& particlesMC,
                                   const typename T::iterator& particle,
-                                  const bool& searchUpToQuark = false)
+                                  const bool searchUpToQuark = false)
   {
     int stage = 0; // mother tree level (just for debugging)
 
--- a/Common/Core/RecoDecay.h
+++ b/Common/Core/RecoDecay.h
@@ -41,20 +41,6 @@ using namespace o2::constants::math;
 /// - Monte Carlo matching of decays at track and particle level
 
 // mapping of charm-hadron origin type
-namespace o2
-{
-namespace aod
-{
-namespace hf_cand
-{
-enum OriginType { None = 0,
-                  Prompt,
-                  NonPrompt };
-} // namespace hf_cand
-} // namespace aod
-} // namespace o2
-
-using namespace o2::aod::hf_cand;
 
 class RecoDecay
 {
@@ -64,6 +50,10 @@ class RecoDecay
 
   /// Default destructor
   ~RecoDecay() = default;
+
+  enum OriginType { None = 0,
+                    Prompt,
+                    NonPrompt };
 
   // Auxiliary functions
 
--- a/Common/Core/RecoDecay.h
+++ b/Common/Core/RecoDecay.h
@@ -40,8 +40,6 @@ using namespace o2::constants::math;
 /// - calculation of topological properties of secondary vertices
 /// - Monte Carlo matching of decays at track and particle level
 
-// mapping of charm-hadron origin type
-
 class RecoDecay
 {
  public:
@@ -51,6 +49,7 @@ class RecoDecay
   /// Default destructor
   ~RecoDecay() = default;
 
+  // mapping of charm-hadron origin type
   enum OriginType { None = 0,
                     Prompt,
                     NonPrompt };
--- a/Common/Core/RecoDecay.h
+++ b/Common/Core/RecoDecay.h
@@ -913,6 +913,77 @@ class RecoDecay
     return true;
   }
 
+  /// Finds the origin (from charm hadronisation or beauty-hadron decay) of charm hadrons.
+  /// \param particlesMC  table with MC particles
+  /// \param particle  MC particle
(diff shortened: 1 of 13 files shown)


vkucera:
Hi @fgrosa , thanks for this implementation. I agree that the output of `checkCharmHadronOrigin` and `OriginType` should be synced independently of the HF code. I would be fine with moving `OriginType` into `RecoDecay`. It should not change anything about the includes because `RecoDecay.h` is already included in `HFSecondaryVertex.h`. What do you think?

Common/Core/RecoDecay.h
```diff
@@ -913,6 +913,71 @@ class RecoDecay
     return true;
   }
 
+  /// Finds the origin (from charm hadronisation or beauty-hadron decay) of charm hadrons.
+  /// \param particlesMC  table with MC particles
+  /// \param particle  MC particle
+  /// \param searchUpToQuark if true tag origin based on charm/beauty quark otherwise on b-hadron
+  /// \return an integer corresponding to the origin (0: none, 1: prompt, 2: nonprompt) as in hf_cand::OriginType (see PWGHF/DataModel/HFSecondaryVertex.h)
+  template <typename T>
+  static int checkCharmHadronOrigin(const T& particlesMC,
```
vkucera, line 922: Maybe a better name would be `getCharmHadronOrigin` since the result is directly returned.

Common/Core/RecoDecay.h
```diff
@@ -913,6 +913,71 @@ class RecoDecay
...
+                                    const typename T::iterator& particle,
+                                    const bool& searchUpToQuark = false)
```
vkucera, line 924: Why is `searchUpToQuark` passed by reference?

EventFiltering/PWGHF/HFFilter.cxx
```diff
@@ -982,10 +982,10 @@ struct HfFilter { // Main struct for HF triggers
       auto indexRec = RecoDecay::getMatchedMCRec(particlesMC, std::array{trackPos, trackNeg}, pdg::Code::kD0, array{+kPiPlus, -kKPlus}, true, &sign);
       if (indexRec > -1) {
         auto particle = particlesMC.rawIteratorAt(indexRec);
-        origin = (RecoDecay::getMother(particlesMC, particle, kBottom, true) > -1 ? OriginType::NonPrompt : OriginType::Prompt);
+        origin = RecoDecay::checkCharmHadronOrigin(particlesMC, particle, false);
```
vkucera, line 985: `false` can be omitted.

Common/Core/RecoDecay.h
```diff
@@ -913,6 +913,71 @@ class RecoDecay
...
+            }
+            auto mother = particlesMC.rawIteratorAt(iMother - particlesMC.offset());
+            // Check mother's PDG code.
+            auto PDGParticleIMother = std::abs(mother.pdgCode()); // PDG code of the mother
+            // printf("getMother: ");
+            // for (int i = stage; i < 0; i++) // Indent to make the tree look nice.
+            //   printf(" ");
+            // printf("Stage %d: Mother PDG: %d, Index: %d\n", stage, PDGParticleIMother, iMother);
+
+            if (!searchUpToQuark) {
```
vkucera, line 951: I think it would increase the readability if it was structured as:
```c++
if (searchUpToQuark) {
  ...
} else {
  ...
}
```

fgrosa:
> Hi @fgrosa , thanks for this implementation. I agree that the output of `checkCharmHadronOrigin` and `OriginType` should be synced independently of the HF code. I would be fine with moving `OriginType` into `RecoDecay`. It should not change anything about the includes because `RecoDecay.h` is already included in `HFSecondaryVertex.h`. What do you think?



> Hi @fgrosa , thanks for this implementation. I agree that the output of `checkCharmHadronOrigin` and `OriginType` should be synced independently of the HF code. I would be fine with moving `OriginType` into `RecoDecay`. It should not change anything about the includes because `RecoDecay.h` is already included in `HFSecondaryVertex.h`. What do you think?

Hi @vkucera, thanks for your feedback! Indeed I also think it's a good solution, so I implemented it in the last commit.

Common/Core/RecoDecay.h
fgrosa: Done!

EventFiltering/PWGHF/HFFilter.cxx
fgrosa: indeed, I removed it

Common/Core/RecoDecay.h
fgrosa: Done!

Common/Core/RecoDecay.h
vkucera: Well, copying a `bool` costs 1 B, copying its address costs 8 B. So in this case, passing by reference is counterproductive.

fgrosa: You are right, fixed!

Common/Core/RecoDecay.h
```diff
@@ -40,6 +40,22 @@ using namespace o2::constants::math;
...
+namespace hf_cand
+{
+enum OriginType { None = 0,
+                  Prompt,
+                  NonPrompt };
+} // namespace hf_cand
+} // namespace aod
+} // namespace o2
+
+using namespace o2::aod::hf_cand;
```
vkucera, line 57: This should not be needed here as it would be redundant in the files where it is needed and it will pollute the namespace in all other HF files.
How about moving the `OriginType` inside the `RecoDecay` class?

EventFiltering/PWGHF/HFFilter.cxx
vkucera: There are still some in `HFFilter.cxx`.

fgrosa: FIxed

Common/Core/RecoDecay.h
vkucera: Thanks. The additional includes of `RecoDecay.h` in `PWGHF/Tasks` are redundant though, as it is already included in `HFSecondaryVertex.h`.

Common/Core/RecoDecay.h
```diff
@@ -40,6 +40,8 @@ using namespace o2::constants::math;
 /// - calculation of topological properties of secondary vertices
 /// - Monte Carlo matching of decays at track and particle level
 
+// mapping of charm-hadron origin type
+
```
vkucera, line 44: Move to `enum`.

vkucera (changes requested):

Common/Core/RecoDecay.h
fgrosa: Removed 👍

vkucera (approved):
Thanks @fgrosa

fgrosa:
Since it is fine for @vkucera, @iarsene could you please approve it? Thanks a lot!

iarsene (approved):

PR #976, opened 2022-07-12
Add compatibility with new GRP scheme

njacazio:
@victor-gonzalez

### DIFF
--- a/Common/TableProducer/trackPropagation.cxx
+++ b/Common/TableProducer/trackPropagation.cxx
@@ -112,7 +112,7 @@ struct TrackPropagation {
       if (!grpo) {
         grpmag = ccdb->getForTimeStamp<o2::parameters::GRPMagField>(grpmagPath, bc.timestamp());
         if (!grpmag) {
-          LOG(fatal) << "Got nullptr from CCDB for path " << grpmagPath << " of object GRPMagField and " << grpPath << " of object GRPObject for timestamp " << timestamp;
+          LOG(fatal) << "Got nullptr from CCDB for path " << grpmagPath << " of object GRPMagField and " << grpPath << " of object GRPObject for timestamp " << bc.timestamp();
         }
         LOGF(info, "Setting magnetic field to current %d A for run %d from its GRPMagField CCDB object", grpmag->getL3Current(), bc.runNumber());
         o2::base::Propagator::initFieldFromGRP(grpmag);
--- a/Common/TableProducer/trackPropagation.cxx
+++ b/Common/TableProducer/trackPropagation.cxx
@@ -122,7 +122,7 @@ struct TrackPropagation {
       }
     } else {
       grpmag = ccdb->getForTimeStamp<o2::parameters::GRPMagField>(grpmagPath, bc.timestamp());
-      LOGF(info, "Setting magnetic field to current %d A for run %d from its GRPMagField CCDB object", grpmag->getL3Current(), bc.runNumber());
+      LOG(info) << "Setting magnetic field to current " << grpmag->getL3Current() << " A for run " << bc.runNumber() << " from its GRPMagField CCDB object";
       o2::base::Propagator::initFieldFromGRP(grpmag);
     }
     o2::base::Propagator::Instance()->setMatLUT(lut);
--- a/Common/TableProducer/trackPropagation.cxx
+++ b/Common/TableProducer/trackPropagation.cxx
@@ -25,6 +25,7 @@
 #include "CommonUtils/NameConf.h"
 #include "CCDB/CcdbApi.h"
 #include "DataFormatsParameters/GRPObject.h"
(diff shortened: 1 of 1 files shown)


njacazio:
Hi @ddobrigk could you have a look please?  I think that the approvers might be on leave, how should we merge this?

mfaggin:
@jgrosseo @iarsene @ktf are there news about this PR? This looks fundamental to work with on AO2Ds with LHC22e. Thank you.

Common/TableProducer/trackPropagation.cxx
```diff
@@ -93,16 +97,34 @@ struct TrackPropagation {
...
-    LOGF(info, "Setting magnetic field to %d kG for run %d from its GRP CCDB object", grpo->getNominalL3Field(), bc.runNumber());
-    o2::base::Propagator::initFieldFromGRP(grpo);
+    if (checkGRPObject) {
+      grpo = ccdb->getForTimeStamp<o2::parameters::GRPObject>(grpPath, bc.timestamp());
+      if (!grpo) {
+        grpmag = ccdb->getForTimeStamp<o2::parameters::GRPMagField>(grpmagPath, bc.timestamp());
+        if (!grpmag) {
+          LOG(fatal) << "Got nullptr from CCDB for path " << grpmagPath << " of object GRPMagField and " << grpPath << " of object GRPObject for timestamp " << bc.timestamp();
+        }
+        LOGF(info, "Setting magnetic field to current %d A for run %d from its GRPMagField CCDB object", grpmag->getL3Current(), bc.runNumber());
```
ktf, line 117: I would use the "detail" level for log by default, unless you really care about having the message during real production.

ktf:
This looks good on my side.

Common/TableProducer/trackPropagation.cxx
```diff
@@ -61,13 +62,16 @@ struct TrackPropagation {
...
+  o2::parameters::GRPMagField* grpmag = nullptr;
   o2::base::MatLayerCylSet* lut = nullptr;
 
   Configurable<std::string> ccdburl{"ccdb-url", "http://alice-ccdb.cern.ch", "url of the ccdb repository"};
   Configurable<std::string> lutPath{"lutPath", "GLO/Param/MatLUT", "Path of the Lut parametrization"};
   Configurable<std::string> geoPath{"geoPath", "GLO/Config/GeometryAligned", "Path of the geometry file"};
   Configurable<std::string> grpPath{"grpPath", "GLO/GRP/GRP", "Path of the grp file"};
+  Configurable<std::string> grpmagPath{"grpmagPath", "GLO/Config/GRPMagField", "CCDB path of the GRPMagField object"};
   Configurable<std::string> mVtxPath{"mVtxPath", "GLO/Calib/MeanVertex", "Path of the mean vertex file"};
+  Configurable<bool> checkGRPObject{"checkGRPObject", false, "Flag to check the CCDB for GRPObject before asking for the GRPMagField"};
```
ktf, line 74: actually, why not simply making it empty by default and use it if not?

njacazio: Because I want to have the feature that the CCDB will not be checked and will get directly the new structure.
This is because the old GRP structure will be translated to the new one

Common/TableProducer/trackPropagation.cxx
njacazio: I agree, however this is the same level of info that was present before, this is also important information that should be printed once per run i.e. once per job ideally

njacazio (commented):
Thanks @ktf for the suggestions, however, I would not change the code. Can we merge it as it is now? This is a blocker for other projects

mfaggin:
> Thanks @ktf for the suggestions, however, I would not change the code. Can we merge it as it is now? This is a blocker for other projects

thanks @njacazio , do you think we can proceed @ktf ?

mfaggin:
> > Thanks @ktf for the suggestions, however, I would not change the code. Can we merge it as it is now? This is a blocker for other projects
> 
> thanks @njacazio , do you think we can proceed @ktf ?

A gentle reminder

njacazio:
@jgrosseo @iarsene if you're back could you please check/approve?

jgrosseo (approved):

jgrosseo:
When will the old scheme be deprecated?

njacazio:
I think it might be already, I have to check with Victor

mfaggin:
> When will the old scheme be deprecated?

@victor-gonzalez can confirm, but if I am not wrong this will is already deprecated, apart for Run 2 converted data

victor-gonzalez:
For Run 2 converted datasets the old scheme will stay
For Run 3 it seems the new schema has been populated with OCT Pilot beam and with the simulations
JUN Pilot beam and Run 3 datasets are only under the new schema

jgrosseo:
Then I think the config parameter should be renamed to choose run 2 or 3...

PR #954, opened 2022-07-06
Add new task

chengtt0406:

PR #978, opened 2022-07-12
Add tool for postprocessing of TPC PID QA output

chengtt0406:


njacazio:
Hi @chengtt0406 thanks a lot for your work!
I have some requests, could you add some detailed instructions? Also I don't think that this is the correct place, @mfaggin what do you think? Maybe we could consider having some post processing directory, provided that we want this in the DPG

mfaggin:
> Hi @chengtt0406 thanks a lot for your work! I have some requests, could you add some detailed instructions? Also I don't think that this is the correct place, @mfaggin what do you think? Maybe we could consider having some post processing directory, provided that we want this in the DPG

Hi @njacazio , you were too fast, I was going to say the same thing :sweat_smile: . @chengtt0406 thanks a lot, this code can be useful but indeed I would think to a different place where to put it because this is not properly an analysis workflow, rather more an offline code written as a c++ program instead of a macro. I agree also with the request of adding a bit of documentation (even comments in the code may be enough) about how to launch the program. Thanks a lot.

njacazio:
Sounds good, if it's ok you can also rename the PR to something like post processing of TPCQA output or so, it would be more clear

jezwilkinson:
Hi, yes, we were also wondering if this was the correct place or if it belonged elsewhere (we were going to ask in the comments of this draft you got there first). Maybe a separate "tools" subdirectory instead of the tasks one?

As for the code itself - yes, this is basically an adapted version of the code that we used in the past to produce the PIDQA reports from the TPC spline generation, at present it takes as input the QAResults files, checks it for a predetermined list of possible histogram lists, and if they exist does the slice fits for the mean/sigma, storing the result in the defined output file name (PDF). We can add a bit more descriptive help text on this for sure.

mfaggin:
> Hi, yes, we were also wondering if this was the correct place or if it belonged elsewhere (we were going to ask in the comments of this draft but you got there first). Maybe a separate "tools" subdirectory instead of the tasks one?
> 
> As for the code itself - yes, this is basically an adapted version of the code that we used in the past to produce the PIDQA reports from the TPC spline generation, at present it takes as input the QAResults files, checks it for a predetermined list of possible histogram lists, and if they exist does the slice fits for the mean/sigma, storing the result in the defined output file name (PDF). We can add a bit more descriptive help text on this for sure.

Hi @jezwilkinson indeed I was thinking as well to a separate `tools` subdirectories or somethig similar. Thanks a lot.

njacazio:
It's also true that this post processes something which is in the DPG folder.. @mfaggin I trust you with this. We can also move the files at some point later

mfaggin:
> I'm not even sure that this should go under the DPG folder, maybe consider moving it to `Tools/TPC` or something as it will be under the responsability of the TPC. Then you could consider becoming approver for that folder. Maybe this can be discussed with others e.g. Jan Fiete or David, but I know many are on leave

if this folder here exists, I then tend to agree with Nicolò, maybe this code is more related to you work as TPC experts rather than DPG one.

mfaggin:
> We figured that since this is reading the output of the dpg tasks that it makes sense to keep it together somehow- possibly the other detectors could profit from something similar to read the outputs too. Also on a previous draft PR that Tiantian made recently, Jan Fiete suggested that dpg might be the right place (#954) . But if you think it's better off elsewhere then we can move it elsewhere

Thanks @jezwilkinson for pointing this out. We can then ask @jgrosseo if he agrees to move this code (that is not strictly a worflow) to somewherelike `Tools/TPC`, as suggested by Nicolò. If he prefers it to stay here, I would ask then to make a `Tools` subdirectory or similar.

PR #981, opened 2022-07-13
Add alternative TPC multiplicity producer/consumer

njacazio:
@jezwilkinson @AnnalenaKa @ChSonnabend this is what I suggest that ca be used to decouple from the central multiplicity task, which sometimes is not available. I tried with templating but it does not work

### DIFF
--- a/Common/TableProducer/PID/CMakeLists.txt
+++ b/Common/TableProducer/PID/CMakeLists.txt
@@ -33,6 +33,11 @@ o2physics_add_dpl_workflow(pid-tof-full
 
 # TPC
 
+o2physics_add_dpl_workflow(pid-tpc-base
+                    SOURCES pidTPCBase.cxx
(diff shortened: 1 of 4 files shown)


jezwilkinson:
Hi @njacazio  - thanks for the edit, we were initially told that the TPC multiplicity should belong to the mult table but it probably does make sense to have it separate after all.

If i'm understanding right, the only difference between the ProcessStandard() and ProcessCustom() is that one of them subscribes to this new AOD::TPCMult table and the other carries on using the old method for multTPC? If that's the case then i'd be more in favour of having a single process function that uses the new table and using that as standard, rather than two process functions doing the same thing via different routes

njacazio:
Hi, indeed this can be solved by removing the need for a calibration object in the mult task, however we have to make sure that this possibility is kept also in the future. Thanks Jeremy!

PR #985, opened 2022-07-13
[Event filtering] Fix match of decays including resonant channels

fgrosa:


alibuild (approved):

PR #986, opened 2022-07-13
Add protection for tracks without associated MC particles

ercolessi:


njacazio (approved):
Thanks Francesca, indeed this was highly needed

PR #988, opened 2022-07-13
CCDB update for event selection in Run1 converted data

ekryshen:


alibuild (approved):

PR #992, opened 2022-07-14
Add coordinators to codeowners

njacazio:
@lramona @alcaliva @lbariogl @BongHwi @lbarnby @mbombara @iravasen @njacazio I miss the users for Sourav and Chiara what do you think about this? I propose that we extend the merging to coordinators, but keep the same structure so that we avoid too much fragmentation, this PR is to discuss what to do


chiarapinto:
Hi, I agree with your proposal!

jgrosseo:
Don't you want to create PAG folders like the others have, right away?

njacazio:
No we will do it in a second run

PR #996, opened 2022-07-14
Update qaMatchEff.cxx

rosarioturrisi:
Changed process functions in order to loop only on tracks belonging to a collision


mfaggin (approved):
Thanks Rosario, grouping in collisions makes more sense in my opinion.

PR #1001, opened 2022-07-15
lambdakzerofinder - add process switch for run2 vs run3

aimeric-landou:


njacazio (dismissed):

njacazio:
Ah, you need to format probably

aimeric-landou:
Indeed there is a clang issue, unfortunately I cannot enter Clang/latest on my computer anymore. Sent an email to the analysis task force mailing list, hopefully someone can help me with that.

njacazio:
You could use lxplus until it is fixed

PR #1008, opened 2022-07-18
DPG: add collision grouping match. eff.

mfaggin:


alibuild (approved):

PR #1016, opened 2022-07-18
Fix small histogram typos

giovannimalfattore:


njacazio (dismissed):

njacazio:
Ah, it does not build, my bad

PR #934, opened 2022-06-29
Simplify TOF PID Response parameter usage

njacazio:
- bypass previous DetectorResponse common structure (will be removed)

- New structure holds only parameters
  - Implementation is left in code only
  - Add additional param getter
  - Using data member as mass defined at compile time
  - Fix warning of overriding streamers

- Simplify TOF param handler
  - Add performance plotting
  - Add validity based on run range
  - Add parameters to metadata

- Add PID simple mode w/o slicing
  - Add possibility to turn on and off slices

- Extend TOF PID tasks
  - Fix pidTOF switches
  - Update default par name to TOFResoParams, add FT0 only ev time
  - Clean includes

- Extend TOF QA plots
  - Add ev time reso
  - Fill ev time per class

PR #1020, opened 2022-07-19
Fix CCDB access in TPC PID response handler tool

jezwilkinson:
TPC PID response handler's CCDB access was broken by a previous commit that modified the timestamp initialisation in handleParamBase.h (#934). This PR fixes the incompatibility and removes redundancies with the options that were added to the base header.

Only impacts the parameter object handler tool; the analysis tasks are unaffected.


jezwilkinson:
Hi @jgrosseo @iarsene @ktf  - could you please check if the pull is OK? We can work around the issue by using the locally built version of the handler to push to CCDB, but it would be preferable to have the central version working properly.
Thanks, --Jeremy

iarsene (approved):

PR #1025, opened 2022-07-20
Improve pidTPCML.cxx

ktf:
* Out of line code to reduce memory usage while compiling pidTPC tasks
* Change to use an array for the intermediate tracks representation.
  It would probably be better to read directly the values from the
  arrow tables, but that will come in a different PR.

### DIFF
--- a/Common/TableProducer/PID/CMakeLists.txt
+++ b/Common/TableProducer/PID/CMakeLists.txt
@@ -11,6 +11,11 @@
 
 # TOF
 
+
+o2physics_add_library(PIDTPCCore
+             SOURCES pidTPCML.cxx
+             PUBLIC_LINK_LIBRARIES O2::Framework ONNXRuntime::ONNXRuntime
+)
 o2physics_add_dpl_workflow(pid-tof-base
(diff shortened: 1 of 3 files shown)


ktf:
@njacazio @ChSonnabend could you have a look? Besides the mechanical change of moving the inference code in a separate .cxx file (this is probably one of the causes why pidTPC takes so much memory to compile), I have also changed the intermediate track representation to be an std::array, which should already somewhat speedup the whole thing. A better approach would be to find a way to have ONNX runtime digest a columnwise representation so that it can bulk process the whole arrow table, rather than one track at the time.

ChSonnabend:
Hi Giulio,
Thanks a lot for the PR! I think this makes sense to put it in a separate .cxx file. It makes the code much cleaner.
Indeed I also mentioned the idea of running directly on the tables to Nicolo before, but I wasn't sure how to implement it. If you know how to do this, that would be very nice!

njacazio:
Hi, indeed, thanks a lot this was needed!

jgrosseo (approved):

PR #1026, opened 2022-07-20
[PWGMM] uploading some post-processing codes and legacy from AliPhysics

jikim1290:
uploading some post-processing codes and legacy from AliPhysics
This is request from Guillermo


aalkin:
This simply cannot be merged. We only allow small self-contained post-processing macros and binaries, but as a general rule anything that does not depend on O2/O2Physics code in any way should not be in the repository. It also concerns data files specifically - anything that is calibration-like should go in CCDB. 

Notifying @ddobrigk @jgrosseo

jikim1290:
@aalkin Thank you very much for commenting on this. Is it still impossible to merge two codes inside macros?

jgcn:
Hi Anton, David, Jan Fiete,

 I would like then to have your suggestions where to put this code.
The vdM analysis uses root macros (the macros in the directory https://github.com/alisw/AliPhysics/tree/master/PWGMM/Lumi/vdM, not in the subdirectories)
This code is pure root, so it does not need neither O2 nor AliPhysics.
For each analysis we get a new doctoral student doing her/his service task, so the code should be available for them so that they can use it and improve it for the next generation ...
The main reason why I want to move it to O2 is that  I understand  (may be wrongly?) that at some point AliPhysics will not be anymore mantained. And probably new students will install only O2 ... 
So, where do we put this analysis macros? To me the logical place is some where in the luminosity directory of O2 https://github.com/AliceO2Group/O2Physics/tree/master/PWGMM/Lumi but if I understand your comment above, this is not possible.

 So,suggestions on how to proceed?
 (no idea if this is the right place to have this discussion ... please advice if you want to move it to other communication channel ...)

thanks a lot and have a nice day
guillermo

jikim1290:
Posting @aalkin @ddobrigk  @jgrosseo  for the message from Guillermo
@jgcn

jgrosseo:
The AliPhysics repository will not disappear why we have to the legacy macros to O2?

PR #1036, opened 2022-07-21
PWGHF: Add Dplus-hadrons pairs tables

Simple-Shyam:
I added Dplus-hadrons pair tables for Dplus-hadron correlations.


vkucera:
Hi @Simple-Shyam , is there a reason to not commit the task as well?

Simple-Shyam:
> I am just trying simple case, the tasks are ready. I want to commit if everything works fine in this simple case. I seen checks are failed.

PR #1043, opened 2022-07-21
Add antiparticle histograms in light antinuclei analysis

giovannimalfattore:


PWGLF/Tasks/LFNucleiBATask.cxx
```diff
@@ -216,33 +234,57 @@ struct LFNucleiBATask {
...
+      if (std::abs(track.nsigTPC3He()) < nsigmaTPCcut) {
+        if (track.sign() > 0)
+          histos.fill(HIST("tracks/helium/h1HeliumSpectra"), track.pt());
+        else
+          histos.fill(HIST("tracks/helium/h1antiHeliumSpectra"), track.pt());
+      }
 
       if (track.hasTOF()) {
-        histos.fill(HIST("tracks/h2TOFbetaVsP"), track.p(), track.beta());
+        histos.fill(HIST("tracks/h2TOFbetaVsP"), track.p() / (1.f * track.sign()), track.beta());
```
njacazio, line 287: I think sign is just a `short` equal to `1` or `-1` this is ok but you could also consider `track.p() * track.sign()` as it is not the charge, it is really the sign

njacazio (approved):
Very good, highly needed!

PR #1050, opened 2022-07-25
DPG: add per-event qa plots for mismatched tracks (MC).

mfaggin:


mfaggin:
@fgrosa here the histos regarding the discussion of this morning, we'll keep monitoring this as well

njacazio (approved):

PR #1053, opened 2022-07-25
Add labelers

ddobrigk:


jgrosseo:
Would it make sense to have this part of the builder and switch it on/off depending on if the table is requested by another task?

ddobrigk:
(sorry for the slow response - was on night shifts earlier this week) Hi @jgrosseo, @njacazio, for merging with the builders, that's definitely a good idea! I will work on that. Nicolo, I think that, as for moving the strangeness tables, I guess it's a bit a matter of taste. It's true that these are general-purpose analysis tables, so I don't really have a strong feeling ...

ddobrigk:
Will incorporate into builders + make minor adjustments

PR #1054, opened 2022-07-26
fatal error for invalid configuration

jgrosseo:


jgrosseo:
@victor-gonzalez What do you think?
In addition the centrality table has both process functions (run2 and run3) off by default. The disadvantage is one cannot run it at all to get a dpl-config.json to then edit it. I would suggest to enable one of them by default, as we do for the multiplicity task. What do you think?

jgrosseo:
in my case the file is written but has an empty config entry for `centrality-table`. It is not the case for you?

Is this PR ok for you?

victor-gonzalez:
In the case of Malgorzata, which was the one which reported it, surprisingly, she got a complete `centrality-table` entry in the `json` file. Actually we used it for creating a configuration file which fixed the issue

The PR is OK for me

alibuild (approved):

PR #1058, opened 2022-07-26
DPG: kine correlation IU vs. DCA filtered tracks.

mfaggin:


mfaggin:
@njacazio not the most elegant solution probably, but this works. Trying to use `rawIteratorAt` with gloal indices breaks at runtime and this is because of grouping. I would go like this, if you agree.

PR #1064, opened 2022-07-27
Femtoworld - binningpolicy changes + debugging

zchochul:
After adding constants from `CommonConstants/MathConstants.h` I've encountered an error. Task is building properly locally but when I tried to run it I had errors, for example: `Error in <TExMap::Add>: key 139776835007120 is not unique` and `Error in <TCling::LoadPCM>: ROOT PCM /home/zuzanna/alice/sw/ubuntu2004_x86-64/ROOT/v6-26-04-patches-alice1-3/lib/libHist_rdict.pcm file does not exist`. <br>
I've also changed BinningPolicy -> ColumnBinningPolicy as @saganatt has requested.

### DIFF
--- a/PWGCF/FemtoWorld/Core/FemtoWorldContainer.h
+++ b/PWGCF/FemtoWorld/Core/FemtoWorldContainer.h
@@ -24,6 +24,9 @@
 #include "TMath.h"
 #include "TDatabasePDG.h"
 
+#include "CommonConstants/MathConstants.h"
+using namespace o2::constants::math;
+
 using namespace o2::framework;
 
 namespace o2::analysis::femtoWorld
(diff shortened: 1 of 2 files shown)


victor-gonzalez:
In which task you get `Error in <TExMap::Add>: key 139776835007120`?
This `<TCling::LoadPCM>: ROOT PCM /home/zuzanna/alice/sw/ubuntu2004_x86-64/ROOT/v6-26-04-patches-alice1-3/lib/libHist_rdict.pcm` point to a mismatch in your environment. Have you lately updated your `alidist`?

lgraczykCern:
Hi @victor-gonzalez . I think we somehow found the problem by sitting together with @zchochul . This error was observed only when the task was compiled with ninja. On the other hand, when compiled with aliBuild, it worked fine. Maybe ninja compilation is not always perfect.

victor-gonzalez (approved):

PR #1065, opened 2022-07-27
Adding purely timestamp-based CCDB-fetching of NN's

ChSonnabend:
This PR removes the runnumber-based fetching and implements a purely timestamp-based fetching of NN's from the CCDB. A call to CCDB is only made if a network runs out of its scope of validity of the start and end timestamps and at maximum once per call to the process-function in the tpc–pid and tpc-pid-full tasks, thus getting rid of unnecessary calls to CCDB.


ChSonnabend:
Hi @ktf, please let me know if you have any comments or suggestions, otherwise would it be possible to merge this quickly, please? :-)

ktf:
This is fine for me, however I think it should be vetted by @ddobrigk or @njacazio.

ddobrigk (approved):
Looks good to me. Thanks, @ChSonnabend !

njacazio (approved):

njacazio:
Thanks @ChSonnabend

PR #1067, opened 2022-07-28
PWGMM: dndeta: Use track selection flag instead of explicit cut

aalkin:
* remove obsolete configurables
* remove obsolete fill statements


alibuild (approved):

PR #1068, opened 2022-07-28
[EventFiltering] Automatically deal with models that need float and double precision in input

fgrosa:


alibuild (approved):

PR #1069, opened 2022-07-28
PWGCF: Add files for the first femto tutorial session

saganatt:
Hi,

@lbariogl @zchochul  and me prepared the first Femto PAG tutorial which is happening the next week.
There will be more tutorial sessions in the future, and this Tutorial directory will serve to contain all tutorial materials for easy reference.

### DIFF
--- a/PWGCF/CMakeLists.txt
+++ b/PWGCF/CMakeLists.txt
@@ -17,4 +17,5 @@ add_subdirectory(FemtoWorld)
 add_subdirectory(MultiparticleCorrelations)
 add_subdirectory(Tasks)
 add_subdirectory(TableProducer)
+add_subdirectory(Tutorial)
 
--- a/PWGCF/Tutorial/CFTutorialTask0.cxx
+++ b/PWGCF/Tutorial/CFTutorialTask0.cxx
@@ -1,0 +1,68 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+/// \author Luca Barioglio
+
+// O2 includes
+#include "Framework/AnalysisTask.h"
+#include "Framework/runDataProcessing.h"
+
+using namespace o2;
+using namespace o2::framework;
+using namespace o2::framework::expressions;
+
+// STEP 0
+// Example task illustrating how to create histograms and fill them with basic information.
+// A basic event selection is applied.
+
+struct CFTutorialTask0 {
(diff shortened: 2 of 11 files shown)


victor-gonzalez:
At some point this will become a femto tutorial, right?
As FemtoDream code have to move at some point to the standard layout for the PWGCF functional areas (as FemtoWorld already did it) I would suggest to incorporate the Tutorial directory into one of the Femto directories
I also would suggest to use the standard layout in the Tutorial directory so that the new analyzers get used to the layout structure and not at having everything in one place
@jgrosseo What do you think?

victor-gonzalez:
But the way, you have to rebase your code

victor-gonzalez:
I would suggest you to make it a draft PR for the time being

lbariogl:
Hi @victor-gonzalez, concerning your comments, I would suggest to keep the tutorial separeated, so that people can access it easily. I would prefer not to mix things, otherwise in the future it will become chaotic. Then yes, soon there will be a second tutorial dedicated to the use of FemtoDream. But first we needed a detailed introduction, so that users won't use the femto machinery blindly

lbariogl:
@victor-gonzalez now I see what you mean. The structure in FemtoDream will be changed later, in a dedicated PR

victor-gonzalez:
Hi @lbariogl, I've thinking around these tutorial files
The more I think about them the more I believe they should not be incorporated to the O2Physics tree
If this is established as the usual procedure in a near future the repository will be plagued of uncontrolled tutorials
I also think it is a good learning step to start things from the beginning not having them necessarily on the central repository

But as I said, my main concern is what is going to happen in the near term with all these tutorial files
@ddobrigk @jgrosseo What do you think?

lbariogl:
Hi @victor-gonzalez, it would be easier to have it in the official repository for the tutorial that will take place in two days. But if you are against this, we can provide future reference to the specific repository.

ddobrigk:
Hi @victor-gonzalez, I think we should not block tutorials for specific code, such as the femtodream, from being on the repository - tutorials are typically a part of software packages such as ROOT, etc.  Rather, I believe this goes the other way around: we should try and organise the tutorial files from each PWG inside subdirectories that are handled by PWG experts. I will anyway have a discussion about some related issues in this week's PB, so I will take this issue there. Hopefully, after this discussion, we can start to organise a bit and avoid what you called 'uncontrolled tutorials'.

victor-gonzalez:
@lbariogl @ddobrigk I would then suggest the following structure for the files in this PR

PWGCF
  CMakeLists.txt
  - FemtoDream
    CMakeLists.txt
    - Core (will come)
    - DataModel (will come)
    - TableProducer (will come)
    - Tasks (will come)
    - Tutorials
      CMakeLists.txt
      - Generic
        CMakeLists.txt
        README.md
        - Core
        - DataModel
        - TableProducer
        - Tasks
          TutorialTask0.cxx
          TutorialTask1.cxx
          TutorialTask2.cxx
          TutorialTask3.cxx
          TutorialTask4.cxx
          TutorialTask5.cxx
        - Macros
          run-config.json
          run.sh
      - Others (will come)

I believe it is important to show a structured way of organizing the code around O2Physics
We did not started with that because the priorities were others

victor-gonzalez:
As the time for the tutorial is coming and clear directives are not yet in place, I approve it for the time being

victor-gonzalez (approved):

PR #1072, opened 2022-07-28
Commented added template and cut of DCAxy

sarahherrmann:
I added some comments for readability, I put a template to avoid writing twice the same process function with just a small variation, and I implemented the DCAxy cut. The ambiguous tracks having a DCAxy > maxDCAXY will just stay ambiguous.

### DIFF
--- a/PWGMM/Mult/Tasks/vertexing-fwd.cxx
+++ b/PWGMM/Mult/Tasks/vertexing-fwd.cxx
@@ -9,13 +9,13 @@
 // granted to it by virtue of its status as an Intergovernmental Organization
 // or submit itself to any jurisdiction.
 
-///
-/// \file   vertexing-fwd.cxx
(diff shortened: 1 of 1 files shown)


PWGMM/Mult/Tasks/vertexing-fwd.cxx
```diff
@@ -8,6 +8,10 @@
 // In applying this license CERN does not waive the privileges and immunities
 // granted to it by virtue of its status as an Intergovernmental Organization
 // or submit itself to any jurisdiction.
+
+//Code written by Robin Caron and Sarah Herrmann
```
aalkin, line 12: There is standardized format for it, look for "author" in other tasks.

aalkin (dismissed):

aalkin (approved):

aalkin (approved):

PR #1078, opened 2022-07-29
[EventFiltering] Add option to run ML inference single thread

fgrosa:


alibuild (approved):

PR #1084, opened 2022-07-29
  PWGHF: Add first correlation task (c'ed from the common repository)

saganatt:
Dear all,

this is a continuation of https://github.com/AliceO2Group/O2Physics/pull/989 . I found the cause of the problem, something external has changed and the code for correlation container cannot use the additional axis without errors.

I am opening a new PR, as this is from a branch in our common repo, so Katarina and Gian can fix potential review comments and errors while I am offline.


Katka19:
We decided to update the name of the task, because we thought that the _correlation-d0_ did not express well what the task does. Its purpose is to eventually measure flow of HF hadrons, so: 1) we do not specify any particle name, and 2) we rather use _flow_ instead of _correlations_ to not confuse it with other correlation tasks in the same folder. Let us know if you have any suggestions.

vkucera (approved):

PR #1087, opened 2022-08-01
adding Daiki as codeowner for DQ

iarsene:


alibuild (approved):

PR #1089, opened 2022-08-02
FemtoWorld changes to the data structure, producer and track-track task

zchochul:


lgraczykCern:
Hi @victor-gonzalez  - this comes from the fact that when Laura (@lauraser) changed something in table definitions in FemtoDream data structures (the skimmed data format), she could not continue with her work as our code was depending on this and crashed the O2 compilation. We met on Monday and decided that for the time being, while we are developing other things, it would be much easier to copy the code so that we (both groups) are totally independent and develop it further on our own. When most functionalities will be in place, we can again come back and see if we can combine everything.

jgrosseo:
@lgraczykCern We are coming back to the same discussion. We agreed that the data format has to stay compatible (the reason is that the space for derived data is limited and otherwise the femto groups each get only 50%). There are ways to version data model changes if needed, but in this case I think one simply has to agree on the changes and apply it to both folders...

lgraczykCern:
@jgrosseo yes, this is true. We don't want to have our own skimmed data stored anywhere. We are basically still in the learning phase, that is why we are doing all these changes. @zchochul is our new student who just started, she has few months to get some very preliminary results for the BSc thesis, and we want to limit potential delays. This is also a way to practice and master the new system. Of course when skimmed data are to be produced we will use the ones agreed by everyone and adjust the code. But we are not there yet simply.

I also talked to @ddobrigk on Mattermost. Our code, as user code, does not have any further dependencies and there has to be some balance between the speed and quality. Otherwise we may have delays which may be frustrating, especially if student have tight deadlines.

majanik:
There are two issues: (1) code duplication - to develop the code and test it we need to put our files into the O2Physics; when we put the code in O2Physics it is said that it should be developed more since it is too similar to the one we started from. But it's very difficult to develop wihout code in O2. This way we are arraving at infinite loop of not being able to progress with code...
(2) We agreed that skimmed data should be common for both frameworks (I understand they should be common for all the analysis in the Femto PAG, or better - the whole CF?) But right now there is no common format of the data for CF, Femtodream team is changing things, and they get frustrated that after their change errors pop up in different places (understandable). We decided together that it would be best to disconnect the codes for the developement stage, but at the same time stay as similar to FemtoDream as possible (so that when we have to merge back it is as easy as possible). As soon as the dataformat in CF will be more-less finalized we will work on using this final format for FemtoWorld as well.

ddobrigk:
Hi everyone, I am checking this with conveners and will also touch base on this topic in today's physics board so that we can define a common policy whenever there is an issue such as this one. I will write back here later today! Thanks for your patience...

ddobrigk:
Thanks for your patience and for waiting until we clarified the situation. Following an internal discussion between PWG-CF conveners, physics coordination, and analysis coordination, we agreed to continue with this PR and merge it - with the explicit understanding that there will be an effort in merging the two data formats in the future. Please note that the absence of such an effort may mean that 50% of the full data is available for each of the two separate skimmed data formats.

lgraczykCern:
Thank you very much @ddobrigk . Of course we all want to use the same final format of derived data, when ready. @victor-gonzalez - if there are no more issues, can you please then approve the PR? Thanks!

victor-gonzalez:
@zchochul @lgraczykCern  I made four change requests and I don't see any of them implemented

lgraczykCern:
@victor-gonzalez hmm maybe could you repeat them here in the comments? We cannot see them anywhere here in the PR.

victor-gonzalez (changes requested):
Let see if now you see the requested changes

zchochul:
> Let see if now you see the requested changes

Now I see them, thank you!

saganatt (approved):

PWGCF/FemtoWorld/DataModel/FemtoWorldDerived.h
```diff
@@ -0,0 +1,249 @@
...
+/*
+namespace hash
+{
+DECLARE_SOA_COLUMN(Bin, bin, int); //! Hash for the event mixing
+} // namespace hash
+DECLARE_SOA_TABLE(Hashes, "AOD", "HASH", hash::Bin);
+using Hash = Hashes::iterator;
+
+} // namespace o2::aod
+*/
```
victor-gonzalez, line 240: Please get rid of these comments
Data models should be clean to not infer to wrong interpretations
You always have the original in femto dream in case you need it

PWGCF/FemtoWorld/TableProducer/CMakeLists.txt
```diff
@@ -8,3 +8,18 @@
 # In applying this license CERN does not waive the privileges and immunities
 # granted to it by virtue of its status as an Intergovernmental Organization
 # or submit itself to any jurisdiction.
+
+#o2physics_add_dpl_workflow(femto-world-producer-reduced
+#          SOURCES femtoWorldProducerReducedTask.cxx
+#          PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::AnalysisCore
+#          COMPONENT_NAME Analysis)
+
```
victor-gonzalez, line 16: Please get rid of these comments

PWGCF/FemtoWorld/TableProducer/CMakeLists.txt
```diff
@@ -8,3 +8,18 @@
...
+o2physics_add_dpl_workflow(femto-world-producer
+          SOURCES femtoWorldProducerTask.cxx
+          PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::AnalysisCore
+          COMPONENT_NAME Analysis)
+
+#o2physics_add_dpl_workflow(femto-world-producer-v0
+#          SOURCES femtoWorldProducerTaskV0Only.cxx
+#          PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::AnalysisCore
+#          COMPONENT_NAME Analysis)
```
victor-gonzalez, line 25: The same as above

PWGCF/FemtoWorld/TableProducer/femtoWorldProducerTask.cxx
```diff
@@ -0,0 +1,519 @@
...
+                                 v0.dcaV0daughters(),
+                                 v0.v0radius(),
+                                 v0.x(),
+                                 v0.y(),
+                                 v0.z(),
+                                 v0.mK0Short()); // QA for V0
+              }
+            }
+          }
+        }*/
```
victor-gonzalez, line 511: Please get rid off these commented lines

PWGCF/FemtoWorld/Tasks/CMakeLists.txt
```diff
@@ -9,7 +9,28 @@
...
+#o2physics_add_dpl_workflow(femto-world-debug-v0
+#          SOURCES femtoWorldDebugV0.cxx
+#          PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::AnalysisCore
+#          COMPONENT_NAME Analysis)
+
+#o2physics_add_dpl_workflow(femto-world-hash
+#          SOURCES femtoWorldHashTask.cxx
+#          PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::AnalysisCore
+#          COMPONENT_NAME Analysis)
+
```
victor-gonzalez, line 26: Please get rid of these commented lines

PWGCF/FemtoWorld/Tasks/CMakeLists.txt
```diff
@@ -9,7 +9,28 @@
...
 o2physics_add_dpl_workflow(femto-world-pair-track-track
           SOURCES femtoWorldPairTaskTrackTrack.cxx
           PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::AnalysisCore
           COMPONENT_NAME Analysis)
+
+#o2physics_add_dpl_workflow(femto-world-pair-track-v0
+#          SOURCES femtoWorldPairTaskTrackV0.cxx
+#          PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::AnalysisCore
+#          COMPONENT_NAME Analysis)
+
```
victor-gonzalez, line 36: Same comment as above

victor-gonzalez (changes requested):
Thanks for implementing the suggested change requests
Your code is going into production
Please implement the minor changes required and I will approve it

jgrosseo:
Thanks @victor-gonzalez for the detailed review. Even if already merged, please consider the comments which we posted.

PR #1090, opened 2022-08-02
str builders: Improvements + fixes + labelers

ddobrigk:
The following improvements, fixes and additions are in this PR: 

- **BUGFIX**: Two indexing issues were fixed in the cascade builder: one in which the collision index was taken from the bachelor and not the cascade and another in which the exclusive user table was filled with a global index to the V0Data table and not, as it should, to the V0 table. This should recover a large number of lost candidates. 
- **IMPROVEMENT**: All builders now use templatized functions to help with code serviceability in the future. 
- **ADDITION**: A new flag and process function was added to both V0 and cascade builders to generate two new tables: McV0Label and McCascLabel. These are joinable with V0Data and CascData and contain a single index to the McParticle, if any, corresponding to the V0 or cascade-originating particle (i.e. K0, Lambda, etc...). The labeler process also creates simplified pT-differential histograms to do basic bookkeeping of MC-associated K0, Lambda and Xi, Omega to allow for some quick checks of any situation. The labels will be -1 in case no association can be done (e.g. not the same mother, etc); filtering on a joined table might then be possible and will enable fast analysis of MC-associated candidates. 

As always, don't trust me :-D - please check carefully and complain in case of unexpected behaviour!


njacazio (approved):
Thanks David!

PWGLF/TableProducer/cascadebuilder.cxx
```diff
@@ -207,48 +204,53 @@ struct cascadeBuilder {
       auto v0 = v0index.v0Data(); //de-reference index to correct v0data in case it exists
 
       std::array<float, 3> pVtx = {v0.collision().posX(), v0.collision().posY(), v0.collision().posZ()};
+
+      auto bachTrackCast = casc.bachelor_as<TCascTracksTo>();
+      auto posTrackCast = v0.posTrack_as<TCascTracksTo>();
+      auto negTrackCast = v0.negTrack_as<TCascTracksTo>();
+
       hCascCandidate->Fill(1.5); //has matched V0
       if (tpcrefit) {
```
aalkin, line 213: This probably should be also ignored for Run 3

ddobrigk: @aalkin thanks for having a look, you're correct! fixing it now

ddobrigk: Okay, done, this should now be fine, I hope! Considering the importance of the fixes in the cascade side, it would be great if we could merge soon... Thanks!!!

njacazio (approved):

iarsene (approved):

PR #1095, opened 2022-08-03
[Event filtering] Tag as bad candidates for which BDT inference fails

fgrosa:


alibuild (approved):

PR #1097, opened 2022-08-04
Add dummy process

ddobrigk:


ddobrigk:
Added a dummy process function to labelers -> this should not be required, in principle, but for now it's needed. The necessity will be studied and potentially fixed later, but this will help with hanging workflows while using the strangeness builders before that fix.

mbombara (approved):

PR #1103, opened 2022-08-07
[EventFiltering] Move ahead preselections and simplify selections

fgrosa:


alibuild (approved):

PR #1104, opened 2022-08-07
[Event Filtering] Fix BDT application for 3-prong decays

fgrosa:


alibuild (approved):

PR #1115, opened 2022-08-09
[Event Filtering] Use enum for better clarity

fgrosa:


alibuild (approved):

PR #1124, opened 2022-08-10
Cascade analysis improvements

ddobrigk:


ddobrigk:
@aalkin - here is an interesting use case! In this code, I wanted to do is to have the ability to switch multiplicity selection on/off, PID on/off and track(IU) Run2/3 on/off. Since these are 3 on/off switches and correspond to different data subscriptions, this meant I did 2^3 = 8 process functions with different subscriptions. In principle, another way of doing similarly is to have the possibility of checking if a table has some data - for instance, one could query if a track contains "tpcNSigmaPr" by checking track.contains_tpcNSigmaPr()" or something like that, helper-wise. I have no clue if that's easy or hard or even feasible, but it's just a thought of something that could be useful...

njacazio (approved):

PR #1127, opened 2022-08-11
PWGMM: dndeta: refit ambiguous tracks

aalkin:
This is a first attempt to account for ambiguous tracks by trying to associate them to compatible collisions. The best collision is determined as one for which DCA is the smallest, out of all collisions (if any) associated to compatible BCs. 
A dedicated task produces a table, joinable with `AmbiguousTracks` that contains collision index and refitted track parameters. 

@fgrosa this may also work for you, you can access ambiguous tracks, re-associated to a (single) collision by subscribing to `soa::SmallGroups<soa::Join<aod::AmbiguousTracks, aod::BestCollisions>> const& atracks`. Note that this simple method does not create duplicated associations.


alibuild (approved):

PR #1137, opened 2022-08-15
[PWGMM] add magnet info and corresponding headers 

jikim1290:
add magnet info and corresponding headers

### DIFF
--- a/PWGMM/Lumi/Tasks/lumi.cxx
+++ b/PWGMM/Lumi/Tasks/lumi.cxx
@@ -41,9 +41,7 @@
 
 #include "CCDB/BasicCCDBManager.h"
 #include "DataFormatsParameters/GRPObject.h"
-#include "DetectorsBase/GeometryManager.h"
 
-#include "DetectorsBase/GeometryManager.h"
 #include "DetectorsBase/Propagator.h"
 
 #include "CommonUtils/NameConf.h"
@@ -56,7 +54,6 @@ using namespace o2::framework::expressions;
 
 struct lumiTask {
   Service<o2::ccdb::BasicCCDBManager> ccdb;
-  o2::base::MatLayerCylSet *lut;
   const char *ccdbpath_grp = "GLO/GRP/GRP";
   const char *ccdburl = "http://alice-ccdb.cern.ch";
   int mRunNumber;
--- a/PWGMM/Lumi/Tasks/lumi.cxx
+++ b/PWGMM/Lumi/Tasks/lumi.cxx
@@ -12,7 +12,7 @@
 /// \since November 2021
 // usage: o2-analysis-timestamp -b --aod-file AO2D.root --configuration
 // json://./config.json | o2-analysis-trackextension -b |
-// o2-analysis-trackselection -b --isRun3 <0, 1> | o2-analysis-mm-lumi -b
+// o2-analysis-trackselection -b --isRun3 0 | o2-analysis-mm-lumi -b
 // --configuration json://./config.json
 
 #include <array>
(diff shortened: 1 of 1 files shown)


PWGMM/Lumi/Tasks/lumi.cxx
```diff
@@ -116,6 +149,24 @@ struct lumiTask {
 
     std::vector<bool> vec_useTrk_PVrefit(vec_globID_contr.size(), true);
 
+    if (mRunNumber != bc.runNumber()) {
+      auto grpo = ccdb->getForTimeStamp<o2::parameters::GRPObject>(
+        ccdbpath_grp, bc.timestamp());
+      if (grpo != nullptr) {
+        o2::base::Propagator::initFieldFromGRP(grpo);
+        o2::base::Propagator::Instance()->setMatLUT(lut);
```
aalkin, line 157: I doubt you need material lookup table here, especially if you are doing this for Run2 data, all the tracks have parameters within the beam pipe. Same goes with geometry. And how does PVertexer uses magnetic field info here?

jikim1290: The code is designed to work with Run3 data. I'm now using converted Run2 data, whose structure is Run3 type.

aalkin: Converted Run2 data contains track with parameters that have been already propagated using material correction, there is absolutely no need to load a LUT. Run3 tracks, that are created by track-propagation task, also do not need any additional material correction. Vertexer may need GRP object, but you have no code that points the vertexer to it.

jikim1290: But this code is supposed to work with Run3 data, especially coming vdM scan at the end of Sep or Oct..

aalkin: Run3 data will have `TracksIU` table in AO2D files, to use `Tracks` you simply add `track-propagation` common task to the workflow, you do not apply material correction yourself.

jikim1290: If I add TracksIU task and track-propagation instead of trackselection and trackextention, the code is crushed  due to same tables were produced in hyperloop:
Error: The wagon "o2lumi" was disabled in "LHC18r_vdMscan". The following wagons have the same derived data outputs {TrackExtension_Run3,TrackPropagation:TracksDCA}

And I can run the code with TrackIU only. but in this case, I got the following errors from hyperloop system:
[45622:lumi]: [09:19:03][WARN] No Magnetic Field in TGeoGlobalMagField, checking legacy FairRunAna
[45622:lumi]: *** Program crashed (Segmentation fault, FPE, BUS, ABRT, KILL)

 Thus, I want to put those info into the task

aalkin: First of all, you should not use `TracksIU` directly, this is only for special cases. Second, you either use `track-extension` (for Run2 and older Run3 datasets) or `track-propagation` (for newer Run3 datasets). You will have to set up two different wagons with different dependencies on hyperloop for Run2 and 3 input. However, the task itself can be identical. And in the task you only need to load magnetic field, not geometry and material LUT. Tracks already have their parameters propagated inside the beam pipe, for vertex finding only the magnetic field is required.

jikim1290: Super thank! I now understand your point, no need to load lut and geometry!

aalkin (changes requested):
You no longer need to include the geometry manager and have `lut` pointer, otherwise it seems ready.

jikim1290:
Could you please approve this? Many thanks!

aalkin (approved):
GRPObject.h is included twice, but you can fix that later.

PR #1140, opened 2022-08-17
Silence warning

ktf:


jgrosseo:
This seems to fail clang...

PR #1153, opened 2022-08-22
Add customizable track selection

njacazio:
@mfaggin what do you think of this?

### DIFF
--- a/Common/Core/TrackSelectionDefaults.cxx
+++ b/Common/Core/TrackSelectionDefaults.cxx
@@ -26,7 +26,7 @@
 TrackSelection getGlobalTrackSelection()
 {
   TrackSelection selectedTracks;
-  selectedTracks.SetTrackType(o2::aod::track::Run2Track);// Run 2 track asked by default
+  selectedTracks.SetTrackType(o2::aod::track::Run2Track); // Run 2 track asked by default
   selectedTracks.SetPtRange(0.1f, 1e10f);
   selectedTracks.SetEtaRange(-0.8f, 0.8f);
   selectedTracks.SetRequireITSRefit(true);
--- a/Common/Core/TrackSelectionDefaults.cxx
+++ b/Common/Core/TrackSelectionDefaults.cxx
@@ -1,0 +1,79 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+///
+/// \file  TrackSelectionDefaults.cxx
+/// \brief Class for the definition of standard track selection objects
+/// \since 20-10-2020
+///
+
+#ifndef TrackSelectionDefaults_H
+#define TrackSelectionDefaults_H
+
+#include "Framework/DataTypes.h"
+#include "Common/Core/TrackSelection.h"
+#include "TrackSelectionDefaults.h"
+
+// Default track selection requiring one hit in the SPD
+TrackSelection getGlobalTrackSelection()
+{
+  TrackSelection selectedTracks;
+  selectedTracks.SetTrackType(o2::aod::track::Run2Track);// Run 2 track asked by default
+  selectedTracks.SetPtRange(0.1f, 1e10f);
+  selectedTracks.SetEtaRange(-0.8f, 0.8f);
+  selectedTracks.SetRequireITSRefit(true);
+  selectedTracks.SetRequireTPCRefit(true);
+  selectedTracks.SetRequireGoldenChi2(true);
+  selectedTracks.SetMinNCrossedRowsTPC(70);
+  selectedTracks.SetMinNCrossedRowsOverFindableClustersTPC(0.8f);
+  selectedTracks.SetMaxChi2PerClusterTPC(4.f);
+  selectedTracks.SetRequireHitsInITSLayers(1, {0, 1}); // one hit in any SPD layer
+  selectedTracks.SetMaxChi2PerClusterITS(36.f);
+  selectedTracks.SetMaxDcaXYPtDep([](float pt) { return 0.0105f + 0.0350f / pow(pt, 1.1f); });
+  selectedTracks.SetMaxDcaZ(2.f);
+  return selectedTracks;
+}
+
+// Default track selection requiring a particular Run 3 ITS matching
+TrackSelection getGlobalTrackSelectionRun3ITSMatch(int matching)
+{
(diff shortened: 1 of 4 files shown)


njacazio:
@mfaggin @mario-krueger suggestions are welcome

Common/TableProducer/trackselection.cxx
```diff
@@ -53,32 +56,29 @@ struct TrackSelectionTask {
         break;
       case 1:
         // Run 3 kAny on 3 IB layers of ITS
-        if (isRun3) {
-          globalTracks = getGlobalTrackSelectionITSMatch(TrackSelection::GlobalTrackRun3ITSMatching::Run3ITSibAny);
-        }
+        globalTracks = getGlobalTrackSelectionITSMatch(TrackSelection::GlobalTrackRun3ITSMatching::Run3ITSibAny);
```
mfaggin, line 59: This cut is for sure meaningful only for `Run 3` only analyses. With the idea to keep the `bool isRun3` and following your proposal, I suggest to write what follows:

`case 1:
     globalTracks = getGlobalTrackSelectionITSMatch(TrackSelection::GlobalTrackRun3ITSMatching::Run3ITSibAny);
     isRun3 = true;
     LOG(info) << ... ;`

where the `LOG(info)` is meant to advice that the `isRun3` boolean was just set to `true` because of the chosen ITS matching.

The same suggestion holds also for the other two cases below.

Common/TableProducer/trackselection.cxx
```diff
@@ -53,32 +56,29 @@ struct TrackSelectionTask {
...
         break;
-
       default:
         LOG(fatal) << "TrackSelectionTask with undefined cuts. Fix it!";
         break;
     }
-    globalTracksSDD = getGlobalTrackSelectionSDD();
-
-    if (isRun3) {
-      globalTracks.SetTrackType(o2::aod::track::TrackTypeEnum::Track);
```
mfaggin, line 80: I would keep this here

Common/Core/TrackSelectionDefaults.cxx
```diff
@@ -0,0 +1,79 @@
...
+  selectedTracks.SetMaxDcaXYPtDep([](float pt) { return 0.0105f + 0.0350f / pow(pt, 1.1f); });
+  selectedTracks.SetMaxDcaZ(2.f);
+  return selectedTracks;
+}
+
+// Default track selection requiring a particular Run 3 ITS matching
+TrackSelection getGlobalTrackSelectionRun3ITSMatch(int matching)
+{
+  TrackSelection selectedTracks = getGlobalTrackSelection();
+  selectedTracks.SetTrackType(o2::aod::track::TrackTypeEnum::Track); // Requiring that this is a Run 3 track
```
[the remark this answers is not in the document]
njacazio: Sounds good, let's keep the `bool`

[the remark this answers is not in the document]
njacazio: My point was that now we have two handles to configure the task for the similar information e.g. Run2 vs Run3. if you set the `bool` to false you and have the index at e.g. `2` you will end up without any cut object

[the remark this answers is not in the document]
mfaggin: yes this is what the code does and imho this is a good thing, because in that case the user should be aware that two incosistent settings were used, then the code crashes.
As it is now it looks ok, I think

Common/Core/TrackSelectionDefaults.cxx
```diff
@@ -0,0 +1,79 @@
...
+  selectedTracks.SetMaxChi2PerClusterTPC(4.f);
+  selectedTracks.SetRequireHitsInITSLayers(1, {0, 1}); // one hit in any SPD layer
+  selectedTracks.SetMaxChi2PerClusterITS(36.f);
+  selectedTracks.SetMaxDcaXYPtDep([](float pt) { return 0.0105f + 0.0350f / pow(pt, 1.1f); });
+  selectedTracks.SetMaxDcaZ(2.f);
+  return selectedTracks;
+}
+
+// Default track selection requiring a particular Run 3 ITS matching
+TrackSelection getGlobalTrackSelectionRun3ITSMatch(int matching)
```
mfaggin, line 46: In case of this renaming, please remember to update consistently the O2 docmentation [here](https://aliceo2group.github.io/analysis-framework/docs/basics-usage/HelperTasks.html#track-selection)

mfaggin (changes requested):

Common/TableProducer/trackselection.cxx
```diff
@@ -54,30 +58,35 @@ struct TrackSelectionTask {
...
+    globalTracks.SetEtaRange(etaMin, etaMax);
 
+    // Extra requirement on the ITS -> Run 2: asking for 1 hit SDD and no hit in SPD
+    globalTracksSDD = getGlobalTrackSelectionSDD();
+    globalTracksSDD.SetPtRange(ptMin, ptMax);
+    globalTracksSDD.SetEtaRange(etaMin, etaMax);
     if (isRun3) {
-      globalTracks.SetTrackType(o2::aod::track::TrackTypeEnum::Track);
+      globalTracks.SetTrackType(o2::aod::track::TrackTypeEnum::Track);    // Requiring that this is a Run 3 track
+      globalTracksSDD.SetTrackType(o2::aod::track::TrackTypeEnum::Track); // Requiring that this is a Run 3 track
```
jgrosseo, line 89: Is the correct? The SDD is for Run 2, isn't it?

jgrosseo:
Sorry for my ignorance but what is the purpose of splitting it into an extra file? In particular with respect to the title "customizable track selection". Where does it get customizable?

Common/TableProducer/trackselection.cxx
njacazio: That's true, but then it means that we have a column that is shipped always false

jgrosseo: You mean for Run3 ? Yes, I think this is how it is. We shouldn't even need to evaluate that one in Run 3 mode...

jgrosseo:
> The extra file significantly sped up the compilation of the code while doing changes so I added it. I personally think it does no harm but of course it can be removed. The customizable part is there https://github.com/AliceO2Group/O2Physics/pull/1153/files#diff-0d36bcdb8e8739940cb76fe7d3253f1b64e539a35238250dd52ab1d592fc3fd5R41 and it revolves only around the kinematic selection.

OK! I understand now. Thanks

Common/TableProducer/trackselection.cxx
njacazio: Ok, perfect!

njacazio: I fixed the requirement and skipped the computation of the flag for Run3

jgrosseo: Looks good.

jgrosseo (approved):

alibuild (approved):

njacazio:
I propose that in the future we could consider having the SDD together with all the ITS2 different requirements for the IB/OB in the mask already present and unify things. Also when running on Run2 data the evaluation is carried out twice.
@jgrosseo what do you think?

PR #1154, opened 2022-08-22
PWGHF: Introduce process switches in Λc → p K0S

pbuehler:


vkucera:
Hi @pbuehler , what is the reason for this change?

pbuehler:
Hi @vkucera, I can not switch off the execution of the MC task. The command line option '--doMC false' does not seem to work. On the other hand if the default is set to false, the MC task is switched on with '--doMC'. So with the default value set to false one has access to both options.

vkucera:
The MC option has been activated by default on purpose to allow for running on MC on AliHyperloop because the command line options are not available there. In order to provide both options on AliHyperloop, the next developer of this task should implement the process function switches as has been done for the other channels. I am not in favour of switching the configuration in the code. You can always do it locally.

Maybe it would help if you specify what exactly you want to do.

pbuehler:
Ok I see. We just stumbled across this while trying to run the task with and without MC option. If you agree I will modify the code to include process switches.

vkucera:
I agree. That would be great. Thanks. Please modify also https://github.com/AliceO2Group/O2Physics/blob/master/PWGHF/TableProducer/HFCandidateCreatorCascade.cxx to allow for the full functionality on AliHyperloop.

vkucera:
Hi @pbuehler , thanks for adding the switches. Please rename the PR accordingly.
I have a few questions.
* Why do you keep `HFCandidateCreatorCascadeExpressions` and `HFCandidateCreatorCascadeMC` separate?
* What is the purpose of `processDoNotMC`?
* Why don't you unify the configurables of the `TaskLcK0sP` task? (`selectionFlagLc`, `cutEtaCandMax`)
* Don't you want to replace the eta cut with a rapidity cut?

In general I think it would be good to be consistent with the structure of the 2/3-prong candidate creators and the Lc task.

pbuehler:
. Why do you keep HFCandidateCreatorCascadeExpressions and HFCandidateCreatorCascadeMC separate?
HFCandidateCreatorCascadeExpressions is only there to extend the HfCandCascBase table. The table is not used in HFCandidateCreatorCascadeMC. The table is extended regardless of whether the MC task is run or not. This is different in HFCandidateCreator3Prong, where the extended table is used in the MC task. To me it is not clear if the extension is needed at all (but please see my last comment below).

. What is the purpose of processDoNotMC?
It seems that if all PROCESS_SWITcHes of a task are set to false, the workflow does not work. This dummy process function is there to avoid this issue when switch processMC is set to false.

. Why don't you unify the configurables of the TaskLcK0sP task? (selectionFlagLc, cutEtaCandMax)?
There are different configurables in the original version of the task. But I think it makes sense to unify them and I will do so.

. Don't you want to replace the eta cut with a rapidity cut?
It tried not to change the logics of the tasks. So I propose to keep the eta cut for this PR and think about further changes which might include modifications of the processing logics.

vkucera:
@pbuehler The `processMC` function subscribes to the extended table in the other candidate creators as well as in this one and it should be done that way. Including the function into the extension task is just a convenient way of reducing the number of tasks.

The issue with process switches set to false is new to me. Has it been reported somewhere? AFAIK the other candidate creators never had that problem.

pbuehler:
@vkucera Ok then, I will do exactly as in the 3prong task.

pbuehler:
Hi @vkucera, do you know how to proceed with this?

vkucera:
Hi @pbuehler, MegaLinter is complaining about errors in some unrelated files. I think it could help if you rebase on top of the upstream/master and force push. But before that, it would be useful to try to understand why this is happening. @TimoWilken it seems that MegaLinter is not picking up the right list of changed files. Do you have an idea why? Could it be related to the fact that the PR is made from the master branch?

TimoWilken:
The `master` branch is the "main" branch in this repo, so that should be fine.

I think the problem is that MegaLinter looks for differences with the HEAD of the `master` branch in AliceO2Group/O2Physics *at the time of running*, not differences with the branch when the PR branch diverged from it. `git fetch upstream pull/1154/head pr-1154 && git diff --name-only pr-1154 nightly-20220826` shows the list of files that MegaLinter finds.

The easiest solution should be to rebase on top of the current master (or merge the current upstream master into this PR branch, but that seems less "clean").

pbuehler:
@vkucera Can this be merged?

PWGHF/Tasks/taskLcK0sP.cxx
```diff
@@ -49,14 +43,32 @@ struct TaskLcK0sP {
...
+      registry.add("hEtaRecBg", "cascade candidates (unmatched);#it{#eta};entries", {HistType::kTH1F, {{100, -2., 2.}}});
+      registry.add("hEtaGen", "MC particles (MC);#it{#eta};entries", {HistType::kTH1F, {{100, -2., 2.}}});
+    }
+  }
+
+  Filter filterSelectCandidates = (aod::hf_selcandidate_lc_k0sp::isSelLcK0sP >= selectionFlagLc || aod::hf_selcandidate_lc_k0sp::isSelLcK0sP >= selectionFlagLcbar);
 
-  void process(soa::Filtered<soa::Join<aod::HfCandCascExt, aod::HFSelLcK0sPCandidate>> const& candidates)
+  // -----------------------------------------------------------------------------
+  void processData(soa::Filtered<soa::Join<aod::HfCandCascExt, aod::HFSelLcK0sPCandidate>> const& candidates)
```
vkucera, line 69: Please keep the original name `process`.

vkucera (changes requested):
Hi @pbuehler , it looks already quite good to me, thanks. I think a little cleanup is still needed though. Please see my comments.

Also add yourself as an author in both files and add Chiara in the candidate creator.

Last but not least, propagate your changes in the validation framework.

pbuehler:
> Last but not least, propagate your changes in the validation framework.
@vkucera I am not familiar with the validation framework and need to study this more carefully. Could you approve the current changes and then there will be an extra PR for the adaption of the validation framework?

vkucera (approved):
Thanks, I can take care of the changes in the validation framework.

PR #1155, opened 2022-08-22
DPG: fix ccdb query for magnetic field with new GRPMagField.

mfaggin:


alibuild (approved):

PR #1156, opened 2022-08-23
Fix check on process functions

njacazio:


alibuild (approved):

PR #1168, opened 2022-08-26
Fix multiple CCDB queries in case first one fails

ddobrigk:


alibuild (approved):

PR #1169, opened 2022-08-26
Adjust path check

ddobrigk:
@AnnalenaKa this should do it! Note that it is a quick edit, though, so could you please check what I did? @njacazio, could you please also look at the code? Please be super critical and shout if it doesn't make sense! Thanks!

### DIFF
--- a/PWGLF/TableProducer/lambdakzerobuilder.cxx
+++ b/PWGLF/TableProducer/lambdakzerobuilder.cxx
@@ -191,22 +191,14 @@ struct lambdakzeroBuilder {
   {
     // TODO done only once (and not per run). Will be replaced by CCDBConfigurable
     static o2::parameters::GRPObject* grpo = nullptr;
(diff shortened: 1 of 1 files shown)


AnnalenaKa:
> @AnnalenaKa this should do it! Note that it is a quick edit, though, so could you please check what I did? @njacazio, could you please also look at the code? Please be super critical and shout if it doesn't make sense! Thanks!

Hi @ddobrigk thanks a lot for taking care of this. The changes look good to me and this should solve the problem. The path is now taken from the configurable and not hardcoded anywhere anymore. And this is the default path for Run 2 right? And if we use Run 3 data this GRP path has to be changed in the config file?

njacazio:
Hi @ddobrigk this looks ok! I'm wondering if for the sake of simplicity you want to get rid of the old grp object that is deprecated. New way should be viable with older data now

njacazio (approved):

PR #1173, opened 2022-08-29
LambdakzeroBuilder: Timestamp from bc

AnnalenaKa:
Hi @ddobrigk and @njacazio 

I tried to run the lambdakzerobuilder on the LHC22e data and the error about the grp object not being found still showed up. After another check and chat to Mattia we saw that the timestamp was also hardcoded and and this was leading to an error when getting the object. I changed that to the timestamp being taken from the bc. My test on the LHC22e set was running this time. But of course there could be some problem that I missed with Run 2 data. 
@ddobrigk of course this is your task and I don't mean to interfere. This is just a proposal to make the task working for the LHC22e and LHC22f datasets as well where we had some issues. Feel free to modify or delete my changes and suggestions as you wish.
I am sorry for not spotting this issue two days ago when going through your changes in the last PR on this task.


ddobrigk (approved):
Looks good to me. Thank you, @AnnalenaKa !

njacazio (approved):
Thanks @AnnalenaKa !

PR #1178, opened 2022-08-31
lambdakzerobuilder: add patch for type error Run 2

AnnalenaKa:
Hi, @ddobrigk and @njacazio 
this is the patch that recovers the error when running the lambdakzerobuilder on Run 2 converted data. There was a problem in the type of the timestamp.


ddobrigk:
Thanks, @AnnalenaKa ! @njacazio , can you please go ahead and merge?

njacazio (approved):

PR #1181, opened 2022-09-01
 Adjust GRP path check + take timestamp from bc

ercolessi:
Hi @ddobrigk, I realized that the cascadebuilder was also crashing on the GRP path check, so I tried to fix it following the most recent commits made on the lambdakzerobuilder in the past days.
This works for me now, could you have a look? Thanks!


ddobrigk (approved):
Super thanks, @ercolessi ! I knew this had to be done but as usual zero time from my side...

njacazio (approved):

PR #1192, opened 2022-09-03
Fix axes in LF nuclei task

giovannimalfattore:


njacazio (commented):
Hi, could you rebase to fix the failing tests?

giovannimalfattore:
@njacazio Hi, I don't know how. 
And I have no idea why every time I commit a new fixes and this become part of the code, still I can see all the list of the latter commits. Can all this commits become a new single one?

PR #1205, opened 2022-09-06
CF filter update to include pLL and LLL triplets

lauraser:


alibuild (approved):

PR #1207, opened 2022-09-06
[PWGMM]dndeta-mft - added ambiguous tracks fwd

sarahherrmann:


PWGMM/Mult/TableProducer/trackPropagation.cxx
```diff
@@ -185,6 +209,91 @@ struct AmbiguousTrackPropagation {
...
+              << " from its GRPMagField CCDB object";
+    o2::base::Propagator::initFieldFromGRP(grpmag);
+    runNumber = bc.runNumber();
+    o2::field::MagneticField* field = static_cast<o2::field::MagneticField*>(TGeoGlobalMagField::Instance()->GetField());
+    double centerMFT[3] = {0, 0, -61.4}; // Field at center of MFT
+    Bz = field->getBz(centerMFT);
+    printf("--------Bz = %f\n", Bz);
+  }
+
+  void process(aod::MFTTracks const&,
```
aalkin, line 243: This should probably have an option to exclude propagation of either central or forward ambiguous tracks when they are not needed.

sarahherrmann: I agree,will that do with a different process name and a process switch ? If inside a struct I just do smthg like`if (doProcessCentral){return;}` will that work ? 
Is that what you meant ?

aalkin: I would suggest doing this similar to other service tasks, both tables are produced, but they are filled by process functions that can be disabled, so if the table is not needed it would be empty.

sarahherrmann: Ok I did now :)

aalkin (approved):

PR #1209, opened 2022-09-07
adding and between all detectors involved in the mult estimator

aortizve:


alibuild (approved):

PR #1215, opened 2022-09-07
PWGHF: refinements to histogram ranges of taskD0

fcolamar:


vkucera (changes requested):
Hi @fcolamar , I'm afraid that some of the histograms are used for the validation against AliPhysics and changing their binning would break the comparison. If you think it might be a reasonable solution, one could rename the current validation histograms (e.g. `hValidateMass` etc.) to indicate they should not be changed and create new histograms with the optimised binning.

fcolamar:
Hi @vkucera, sure, seems a good solution. Does this hold for all the histograms I modified, of just for a subset of them?

fcolamar:
Ciao @vkucera, I have duplicated the histograms for which I wanted to update the binning, but renaming the new ones (at least until things are properly changed in the validation codes) so that using the current validation code won't create issues.
I also reverted, for the original version, the d0xd0 histogram binning that I had previously changed in another PR.

vkucera (approved):
Thanks @fcolamar

PR #1226, opened 2022-09-11
Fixed antiparticle check for PLL and LLL

lauraser:


alibuild (approved):

PR #1232, opened 2022-09-12
added processing of tracks w/o loop on collisions

rosarioturrisi:
two new process functions added - just to get rid of track-collision association

### DIFF
--- a/DPG/Tasks/AOTTrack/qaMatchEff.cxx
+++ b/DPG/Tasks/AOTTrack/qaMatchEff.cxx
@@ -637,6 +637,273 @@ struct qaMatchEff {
   PROCESS_SWITCH(qaMatchEff, processMC, "process MC", false);
   //
   //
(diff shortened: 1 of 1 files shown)


mfaggin (changes requested):
Thanks @rosarioturrisi 
In my opinion this task is getting quite overcrowded of functions that do the same things, probably a "cleaning" with templates will be necessary. But for the time being I would go ahead in this way.
However, before approving I would request a change: here the process functions with and w/o grouping are filling the same histograms, and this is not good. I would add a protection in the `init` that prevents the contemporary execution of both process functions w/ and w/o grouping (we can fix it tomorrow)

rosarioturrisi:
got the point, thanks Mattia.

mfaggin (approved):
Thanks @rosarioturrisi , this sounds good to me.

PR #1234, opened 2022-09-12
Add new default function with Run 3 ITS requirements.

mfaggin:


mfaggin:
This PR is meant to propose global track selections overcoming the actual default requirement on the ITS matching, which is a refuse from the Run 2 SPD. The poroposed solution requires to change a `Configurable` in the `trackselection.cxx` workflow (ie: in hyperloop a different core Service wagon for each possible configuration, potentially), however it seems to me quite smart and transparent for the users, since they do not need to touch at all their analysis code. Let me know what you think.

jgrosseo:
This looks good to me.

Just one comment: why the "k" in the middle of the enum word. E.g. Run3ITSIBkAny ? The "k" was a prefix value for enums in the past (for constant) but I don't think it is needed here. ;-)

mfaggin:
> This looks good to me.
> 
> Just one comment: why the "k" in the middle of the enum word. E.g. Run3ITSIBkAny ? The "k" was a prefix value for enums in the past (for constant) but as a first character (not needed by today's coding conventions anymore) and I don't think it is needed in the middle of the words. ;-)

Thanks @jgrosseo , I followed your suggestions (my bad, I always refer to `kAny` when talking about this kind of selections), let me know if this is fine for you

jgrosseo (approved):

jgrosseo:
For me this looks good. Do others have comments?
I think after this is merged, you can announce it and then we can put the default also to a Run 3 (instead of Run 2). @ddobrigk what do you think?

mfaggin:
> For me this looks good. Do others have comments? I think after this is merged, you can announce it and then we can put the default also to a Run 3 (instead of Run 2). @ddobrigk what do you think?

Before doing the announcement, I can also update the instructions in the documentation (here: https://aliceo2group.github.io/analysis-framework/docs/helperTasks/trackselection.html) below the table with the track selections

PR #1237, opened 2022-09-13
use EvSel and BcSel instead of default associations

pbuehler:


alibuild (approved):

PR #1249, opened 2022-09-15
PWGDQ: add option to pass string from config.json to tablemaker to add…

feisenhu:
… additional histograms

I have added an additional subgroup of histograms to the HistogramsLibrary.h
- Its subGroubStr starts with "track" to identify the corresponding histograms within the group as histograms that should be added to the subfolder of the track cuts

Within the TableMaker I added a Configurable called ``cfgAddHistogram`` initialised empty, which can be used and added by the user to pass a string to the TableMaker in order to select additional Histograms that should be written.

- The string functions as a comma separated list for multiple arguments
- For each histClass (`"Event"`,`"Track"` or `"Muons"`) the passed string is checked if it contains `"event"`, `"track"` or `"muons"` within its name. In this way the selected subGroub of the HistogramsLibrary can be added to the right histClass of the output file.


feisenhu:
Hello @iarsene,
I wanted to share the implementation of adding additional Histograms to the output of the table maker by passing a string.
In case you find time to look at it, I would be happy to receive some feedback on whether the implementation is fine like this or I should approach this in a different way. 
Kind regards, Florian

iarsene:
Hi @feisenhu,
Thanks for the quick action on this. This looks good, but I would suggest to actually go all the way and refactor all the configurables regarding QA histograming, which i think is now not very optimal. So at the moment we have one configurable which switches on/off the QA, and another configurable which specifies whether we want detailed QA in addition. So, I would leave by default just the histograms from the event, track and muon groups, and then add on top whatever new histogram groups the user wants to add.
So, we could just keep the fConfigNoQA configurable and remove the fConfigDetailedQA. Then, if the user specifies new histogram subgroups via your new configurables, then those will be filled, otherwise we will run just with the basic histograms.
All the QA should ofcourse be switchable on/off via the remaining fConfigNoQA configurabe.

PR #1250, opened 2022-09-16
PWGCF: Update multiparticle-correlations-ar task

ariedel-cern:
- add support for differential computation of multiparticle correlators


PWGCF/MultiparticleCorrelations/Tasks/multiparticle-correlations-ar.cxx
```diff
@@ -385,33 +610,505 @@ struct MultiParticleCorrelationsARTask {
...
+    fWeightsAll.push_back(weight);
+
+    // fill angles into bins of a track variable, for computing correlators as a function of track variable
+    std::array<double, AR::kLAST_CorTrackDep> TrackDep = {track.pt(), track.eta()};
+    int bin;
+    for (int track = 0; track < AR::kLAST_CorTrackDep; track++) {
+      bin = fTrackDepHists[track]->FindBin(TrackDep[track]) - 1; // in root, bins start at 1
+      fAzimuthalAnglesTrackDep[track].at(bin).push_back(angle);
+      fWeightsTrackDep[track].at(bin).push_back(weight);
+    }
```
victor-gonzalez, line 630: Do you need to do this loop on a per track basis?

PWGCF/MultiparticleCorrelations/Tasks/multiparticle-correlations-ar.cxx
```diff
@@ -385,33 +610,505 @@ struct MultiParticleCorrelationsARTask {
...
+      c2 += Recursion(nm1, harmonic, multp1, counter2);
+      --counter2;
+    }
+    harmonic[nm2] = harmonic[counter1];
+    harmonic[counter1] = hhold;
+    if (mult == 1)
+      return c - c2;
+    return c - double(mult) * c2;
+  }
+
```
victor-gonzalez, line 1111: Couldn't you use the Generic Framework already implemented?

victor-gonzalez (changes requested):

PWGCF/MultiparticleCorrelations/Tasks/multiparticle-correlations-ar.cxx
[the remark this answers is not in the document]
ariedel-cern: Fixed.

PWGCF/MultiparticleCorrelations/Tasks/multiparticle-correlations-ar.cxx
```diff
@@ -385,33 +610,505 @@ struct MultiParticleCorrelationsARTask {
...
   }
 
+  template <typename TrackObject>
+  void FillAzimuthalAngle(TrackObject track)
+  {
+    double angle = track.phi();
+    double weight = GetWeight<TrackObject>(track);
+
+    // all event angles, for computing correlators as a function of event variable
+    fAzimuthalAnglesAll.push_back(angle);
```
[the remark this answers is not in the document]
ariedel-cern: I store the angle of one track and the corresponding weight (at the moment) a total of three times.
Once in `fAzimuthalAnglesAll`, and twice in `fAzimuthalAnglesTrackDep`, once for each bin in pt and eta.

Then I can use the same methods for computing the Q-vectors and from that the correlators.
So the goal is that I have only one array holding the Q-vectors. I compute them once using all angles in the whole event (then compute the correlators and fill them in the corresponding TProfile) and then overwrite the values every time I compute them again for another bin in either pt or eta (and subsequently also the correlators and once again filling them in the corresponding TProfile).

I was thinking of other approaches, but this is the solution, which is, in my opinion, the most maintainable. Though, you are right, that I have to check if it is performant enough, but at least, testing the code locally, I did not have any issues.

I was also briefly thinking about using parititions, but was not able to figure out to change the number of partitions depending on my configurables (let's say I want to have a finer binning in pt, I do not want to change the code to make that possible).

PWGCF/MultiparticleCorrelations/Tasks/multiparticle-correlations-ar.cxx
ariedel-cern: I store a copy of the angle for each bin seperately (like I mentioned above).

When I first came up with this algorithm, instead of storing the value of angle, I stored the index of where that angle is in vector holding all the angles in the event. But whether I store the index (an integer) or the value (a double) should not make a difference.

PWGCF/MultiparticleCorrelations/Tasks/multiparticle-correlations-ar.cxx
ariedel-cern: To be honest, I have not really looked at their code. Ante and myself we would like to have a code seperate from theirs.

PWGCF/MultiparticleCorrelations/Tasks/multiparticle-correlations-ar.cxx
```diff
@@ -420,42 +1117,46 @@ struct MultiParticleCorrelationsARTask {
...
-    Angles.clear();
-    Weights.clear();
+    fAzimuthalAnglesAll.clear();
+    fWeightsAll.clear();
+    for (int track = 0; track < AR::kLAST_CorTrackDep; track++) {
+      for (std::size_t bin = 0; bin < fAzimuthalAnglesTrackDep[track].size(); bin++) {
+        fAzimuthalAnglesTrackDep[track].at(bin).clear();
+        fWeightsTrackDep[track].at(bin).clear();
+      }
+    }
```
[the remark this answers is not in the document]
ariedel-cern: Fixed.

PWGCF/MultiparticleCorrelations/Tasks/multiparticle-correlations-ar.cxx
victor-gonzalez: I'm not quite sure about this reason
O2Physics is not about personal opinions but about an effective Collaboration framework

PWGCF/MultiparticleCorrelations/Tasks/multiparticle-correlations-ar.cxx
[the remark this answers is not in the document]
victor-gonzalez: I approve it for the time being but consider the possibility of a task that pre-filters your collisions/tracks in a way that you don't need to replicate the azimuthal angle column in your inner structures. Probably the filter mechanism on the pre-filtered data will improve execution and the memory layout will be reduced

victor-gonzalez (approved):
I approve it for the time being but consider the possibility of a task that pre-filters your collisions/tracks in a way that you don't need to replicate the azimuthal angle column in your inner structures. Probably the filter mechanism on the pre-filtered data will improve execution and the memory layout will be reduced

PR #1251, opened 2022-09-16
Avoid double counting ambigtracks in dndeta-mft

sarahherrmann:


PWGMM/Mult/Tasks/dndeta-mft.cxx
```diff
@@ -92,6 +93,7 @@ struct PseudorapidityDensityMFT {
 
   void init(InitContext&)
   {
+    ambTrackIds.clear();
```
aalkin, line 96: This needs to be done for each dataframe, so in `process`, not in init

sarahherrmann: ok ! thx

aalkin (approved):

PR #1252, opened 2022-09-17
Two particle correlations skimming framework

victor-gonzalez:
Fully working for collisions and tracks
Collisions skimming implementing
- basic non configurable selection based on number of contributors, vertex quality and EventCuts pilup decisions
- fully configurable zvertex cut with default and variants
- fully configurable centrality/multiplicity cut with default and variants
- fully configurable advance pile-up rejection based on multiplicity correlators

Tracks skimming implementing
- full track characteristics with default and variants support
- Run 1 and 2 track types implemented (not all of them)
- PID information from TPC and TOF

The skimmed data now incorporates 
- the required multiplicities
according to the skimming configuration
- the run number 
- the timestamp

The skimmed tracks table now
- does not have the signed pT
- the sign of the charge will be extracted from the track mask once implemented

The PID skimmed tracks mask is now in a different table
- able to be joined to the skimmed tracks table
- now the mask incorporates also the bayesian PID probabilities cuts

Incorporated the MC labels tables 
- for skimmed collisions
- for skimmed tracks

The configuration has been slightly modified to better track the different active options (default analysis, systematic tests)
from the hyperloop GUI

Incorporated Magnetic Field polarity and track charge sign to the skimmed tables
Both have been incorporated as additional bits in the collision and track masks
The getting of the magnetic field has been also incorporated to the skimming task
Now it is possible to select via 'analysis configuration' (mask) if both magnetic field polarities are required or only just one
The track charge sign cannot be selected via 'analysis configuration' (mask). 
A dynamic column will be incorporated to extract it from the track mask

**The preserved skimmed tracks are from 0.1 to 50.0 GeV and
|eta| < 1.0 and they already include the PID mask 
(a 64b word per track)**

**Skimming ratio 100 -> 9**

The user task filters additionally the skimmed data but for
the selected configuration, default, systematic test, ...
with similar aim as the dptdptfilter task
- the filtered table does not replicate the skimmed table columns
- better naming for the filtered tables
- adapted the filtering task; now filters cannot be applied
to the skimming table because we are creating additional tables
to be joined with it
- extracting the armed index for a cut with variations which is
applied to getting the corresponding multiplicity index for
the multiplicities array column

Incorporated the cut bricks based on functions instead of values

Still pending
- Trigger selection (will increase the ratio if any)
- Skimming cuts signature storage in CCDB and check


PWGCF/TwoParticleCorrelations/DataModel/TwoParticleCorrelationsFiltered.h
```diff
@@ -0,0 +1,62 @@
...
+namespace o2
+{
+namespace aod
+{
+/* we have to change from int to bool when bool columns work properly */
+namespace twopfilter
+{
+DECLARE_SOA_COLUMN(TwoPCollisionCentMult, centmult, float); //! The centrality/multiplicity pecentile
+} // namespace twopfilter
+DECLARE_SOA_TABLE(TwoPAcceptedCollisions, "AOD", "TWOPACCCOLL", //! Accepted reconstructed collisions/events filtered table
```
jgrosseo, line 27: Attention here! The first and third argument must be alike (apart the `s`) to allow merging of DFs in the aod merger

PWGCF/TwoParticleCorrelations/DataModel/TwoParticleCorrelationsSkimmed.h
```diff
@@ -0,0 +1,91 @@
...
+#define O2_ANALYSIS_TWOPSKIMMED_H
+
+#include "Framework/ASoA.h"
+#include "Framework/AnalysisDataModel.h"
+
+namespace o2
+{
+namespace aod
+{
+/* we have to change from int to bool when bool columns work properly */
```
jgrosseo, line 34: I thought they are ok now. Aren't they?

PWGCF/TwoParticleCorrelations/DataModel/TwoParticleCorrelationsSkimmed.h
```diff
@@ -0,0 +1,91 @@
...
+namespace twopskim
+{
+DECLARE_SOA_COLUMN(TwoPSkimmedCollisionCentMult, centmult, std::vector<float>); //! The centrality/multiplicity pecentile
```
jgrosseo, line 37: Can we already use an array column as input for the event mixing? @saganatt ?

saganatt: Simplest way: can you describe in a lambda the logic of obtaining a bin number from such an array? Then you simply use LambdaBinningPolicy for mixing.
This is one array (set of values) per one collision, right? I don't have any universal logic for such case implemented.

jgrosseo: For a giving mixing one would use always the first entry of the array (or the second). So the lambda would probably do. WE will try it. Thanks!

saganatt: Great!
If later we will see there are more use cases and lambda code copies, and there will be such a need, I can try implementing a dedicated binning policy. But in a more distant future ;-)

PWGCF/TwoParticleCorrelations/DataModel/TwoParticleCorrelationsFiltered.h
victor-gonzalez: Oh! I didn't know it
I will fix it!

PWGCF/TwoParticleCorrelations/DataModel/TwoParticleCorrelationsSkimmed.h
victor-gonzalez: This is a legacy comment which does not apply here and which I will remove
But the current implementation of the dptdpt task still works without `bool` because last time I test them, by the end of January, they didn't work

PWGCF/TwoParticleCorrelations/DataModel/TwoParticleCorrelationsSkimmed.h
victor-gonzalez: You will always know to which index in the `centmult` column the current data refers to
The skimming framework will provide it

PWGCF/TwoParticleCorrelations/DataModel/TwoParticleCorrelationsSkimmed.h
jgrosseo: Yes, you are right. We should make a test case for this which we keep... (it didn't group at that time, right?)

victor-gonzalez:
Let's have it in draft while we keep progressing

victor-gonzalez:
@jgrosseo The initial comment contains what has been happening regarding functionality incorporated
There is one additional decision to take
Shall we keep it within the current place in the O2Physics tree or moving it somewhere else

jgrosseo:
From my point of view the location is fine. There is still the point about the table names (to allow merging)

PWGCF/TwoParticleCorrelations/DataModel/TwoParticleCorrelationsFiltered.h
victor-gonzalez: Done!

victor-gonzalez:
The table names are already corrected
Sorry for not closing the conversation

PWGCF/TwoParticleCorrelations/DataModel/TwoParticleCorrelationsSkimmed.h
```diff
@@ -0,0 +1,95 @@
...
+                  cfskim::Phi);
+using CFMCParticle = CFMCParticles::iterator;
+namespace cfskim
+{
+DECLARE_SOA_INDEX_COLUMN(CFMCParticle, mcparticle);
+}
+DECLARE_SOA_TABLE(CFMCCollisionLbls, "AOD", "CFMCCOLLISONLBL",
+                  cfskim::CFMCCollisionId,
+                  mccollisionlabel::McMask);
+DECLARE_SOA_TABLE(CFMCTrakLabels, "AOD", "CFMCTRACKLABEL",
```
jgrosseo, line 89: Typo in Track !

victor-gonzalez: Oh!
Thanks!!
I'll fix it!

alibuild (approved):

PR #1259, opened 2022-09-19
Small change in binning - Lightnuclei task

giovannimalfattore:


chiarapinto:
Ciao Giovanni, the check fails due to a space in L206, can you remove it?

giovannimalfattore:
> Ciao Giovanni, the check fails due to a space in L206, can you remove it?

Ciao Chiara, 
It should work now!

PR #1265, opened 2022-09-20
[EventFiltering] Split trigger flags between 2-prong and 3-prong charm candidates

fgrosa:


EventFiltering/PWGHF/HFFilterQC.cxx
```diff
@@ -36,10 +36,15 @@ namespace
 {
 
 enum HfTriggers {
```
mpuccio, line 38: Ciao @fgrosa ,
I now approve the PR to enable testing in production, but for the next iteration could you move common pieces of code such this one to a common header? Just to avoid unpleasant mismatches in the future.

mpuccio (approved):

EventFiltering/PWGHF/HFFilterQC.cxx
fgrosa: Ciao @mpuccio, indeed I can move these enums and also the definition of the produced tables in a header files. I will do it with another PR, thanks!

alibuild (approved):

PR #1269, opened 2022-09-21
Hf filter

AlexBigO:


EventFiltering/PWGHF/HFFilter.cxx
```diff
@@ -1128,6 +1149,17 @@ struct HfFilter { // Main struct for HF triggers
       auto pVec3Prong = RecoDecay::pVec(pVecFirst, pVecSecond, pVecThird);
       auto pt3Prong = RecoDecay::pt(pVec3Prong);
 
+      auto invMassDplus = RecoDecay::m(std::array{pVecFirst, pVecSecond, pVecThird}, std::array{massPi, massPi, massK});
```
fgrosa, line 1152: The kaon mass should be in the second position (since it's the opposite charge)
```suggestion
      auto invMassDplus = RecoDecay::m(std::array{pVecFirst, pVecSecond, pVecThird}, std::array{massPi, massK, massPi});
```

fgrosa (changes requested):
Hi @AlexBigO there is only a small typo, but apart from that the PR looks good to me 👍

AlexBigO:
Thanks @fgrosa !

fgrosa (approved):

PR #1271, opened 2022-09-21
export the cmake info for usage of O2Physics as dependency

adriansev:
For many reasons (including Open Access) would be required to be able to use O2Physics as (external) dependency for various downstream projects.
So far this is work in progress, as i do not understand the complete cmake structure of O2 which i tried to copy to O2Physics
At this moment the error that i get is:
```
2022-09-21@12:36:05:DEBUG:O2Physics:O2Physics:o2physics: -- Configuring done
2022-09-21@12:36:12:DEBUG:O2Physics:O2Physics:o2physics: CMake Error: INSTALL(EXPORT) given unknown export "O2PhysicsTargets"
```
and it's not clear to me where to look

@ktf @TimoWilken guys, if you have time to take a look maybe you can spot some obvious problem given your experience with the cmake insides of O2.
Thanks a lot!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -61,6 +61,9 @@ include(O2PhysicsAddWorkflow)
 include(O2PhysicsAddLibrary)
 include(O2PhysicsAddHeaderOnlyLibrary)
 include(O2PhysicsTargetRootDictionary)
+include(O2PhysicsSetROOTPCMDependencies)
+include(O2PhysicsDataFile)
+include(AddRootDictionary)
 
 include_directories("${CMAKE_CURRENT_SOURCE_DIR}")
 
@@ -81,3 +84,11 @@ add_subdirectory(PWGUD)
 add_subdirectory(Tools)
 add_subdirectory(Tutorials)
 add_subdirectory(EventFiltering)
+
+# Testing and packaging only needed if we are the top level directory
+if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
+  add_subdirectory(packaging)
+endif()
+
+set_root_pcm_dependencies()
+
--- a/cmake/O2PhysicsAddHeaderOnlyLibrary.cmake
+++ b/cmake/O2PhysicsAddHeaderOnlyLibrary.cmake
@@ -62,6 +62,6 @@ function(o2physics_add_header_only_library baseTargetName)
           DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
 
   install(TARGETS ${target}
-          EXPORT O2Targets
+          EXPORT O2PhysicsTargets
           INCLUDES DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
 endfunction()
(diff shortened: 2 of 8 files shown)


TimoWilken:
For your specific error: perhaps `cmake/O2PhysicsAddLibrary.cmake` L164 and `cmake/O2PhysicsAddHeaderOnlyLibrary.cmake` L65 need a `s/EXPORT O2Targets/EXPORT O2PhysicsTargets/`? These were probably just copy-pasted from O2 without changing the export name to something more appropriate.

adriansev:
@TimoWilken thank you!! indeed, i missed those .. now the compilation is successful and i get:
```
tree lib/cmake/O2Physics/
lib/cmake/O2Physics/
├── AddRootDictionary.cmake
├── CMakeLists.txt
├── FindKFParticle.cmake
├── O2PhysicsCompileFlags.cmake
├── O2PhysicsConfig.cmake
├── O2PhysicsDependencies.cmake
├── O2PhysicsTargets.cmake
├── O2PhysicsTargets-relwithdebinfo.cmake
└── rootcling_wrapper.sh.in
``` 
with this [content](https://asevcenc.web.cern.ch/asevcenc/o2phys_targets/o2physics_lib_cmake_O2Physics/)

does it look good to you?
and as a following step (to be able to use it): what would be the proper way to discover and use the O2 and O2Physics? (and is there a way to include and use the cmake files from O2physics)

Thanks a lot!!

TimoWilken:
I don't know if that's good or not; depends what you need to compile against! As for building against it, I expect the following in the `CMakeLists.txt` of whatever project depends on O2Physics should do it:

```cmake
find_package(O2 CONFIG REQUIRED)
find_package(O2Physics CONFIG REQUIRED)
```

Then you should be able to refer to `O2::*` and `O2Physics::*` targets as e.g. `PUBLIC_LINK_LIBRARIES` in `*_add_dpl_workflow`. Does that work for you?

adriansev:
@TimoWilken Thanks a lot! So, after adding `CONFIG` to the line with O2Physics [here](https://github.com/mgheata/O2openaccess/blob/main/CMakeLists.txt#L41) the cmake configuration is succesful (but the compilation fails as Common/DataModel/EventSelection.h file from O2PHysics is not installed/copied in the install directory of O2Physics, but i will open a different issue/PR for that)

for the moment this seems to work, so i would say is ready for review! Thanks a lot!!

TimoWilken:
Great, that's good to hear! :)

jgrosseo (approved):

adriansev:
@ktf @TimoWilken Hi guys! the PR is stuck in a failed test with:
```
pidOnnxModel.h:249:7: error: private field 'mPid' is not used [-Werror,-Wunused-private-field]
  int mPid;
      ^
```

but i'm quite sure that my PR does not touch that ... could you please help to move forward? thanks a lot!

adriansev:
@ktf @alibuild @iarsene @ddobrigk @AliceO2Group/core 
Hi guys! Now that the tests are green, could you guys please review this? Thanks a lot!

jgrosseo (approved):

PR #1272, opened 2022-09-21
pass by reference

jgrosseo:


alibuild (approved):

ktf:
@aalkin maybe we should find a way to avoid this kind of mistakes, after all tables are immutable once created.

aalkin:
@ktf 
We have
```cpp
static_assert(std::is_lvalue_reference_v<T>, "Argument to process needs to be a reference (&).");
```
in `adaptAnalysisTask`, but it only checks the process functions, of course.

ktf:
I was more thinking about not allowing copies at all. After all there is no case in which that would matter, no?

aalkin:
I can try explicitly deleting copy constructor and see if everything compiles.

PR #1273, opened 2022-09-21
PID ML: Adjust ONNX model class to the generalized interface

saganatt:


alibuild (approved):

PR #1276, opened 2022-09-21
DPG: update ITS-TPC match. eff. workflow to improve the comparison with QC.

mfaggin:
- dcaXY selection added;
 - possibility to use pt(TPC), i.e. calculated from the inner wall of TPC.

CAVEATS:
 - selections on TPC kinematic variables not possible in AO2Ds. Only pt(TPC) cut is applied "by hand" when pt(TPC) is enabled;
 - pt(TPC) calculated using tan(lambda) from tracking (value at the inner wall of TPC not available).

### DIFF
--- a/DPG/Tasks/AOTTrack/qaMatchEff.cxx
+++ b/DPG/Tasks/AOTTrack/qaMatchEff.cxx
@@ -36,10 +36,14 @@ struct qaMatchEff {
   // Track selections
   Configurable<bool> b_useTrackSelections{"b_useTrackSelections", false, "Boolean to switch the track selections on/off."};
   // kinematics
+  Configurable<float> ptMinCutInnerWallTPC{"ptMinCutInnerWallTPC", 0.1f, "Minimum transverse momentum calculated at the inner wall of TPC (GeV/c)"};
   Configurable<float> ptMinCut{"ptMinCut", 0.1f, "Minimum transverse momentum (GeV/c)"};
   Configurable<float> ptMaxCut{"ptMaxCut", 100.f, "Maximum transverse momentum (GeV/c)"};
   Configurable<float> etaMinCut{"etaMinCut", -2.0f, "Minimum pseudorapidity"};
   Configurable<float> etaMaxCut{"etaMaxCut", 2.0f, "Maximum pseudorapidity"};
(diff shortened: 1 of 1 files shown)


mfaggin:
@njacazio @rosarioturrisi  this is the development of what discussed before. Please have a look if eveything is good.

DPG/Tasks/AOTTrack/qaMatchEff.cxx
```diff
@@ -377,6 +390,15 @@ struct qaMatchEff {
     //
     //
     for (auto& jT : jTracks) {
+
+      float trackPt = jT.pt();
+      if (b_useTPCinnerWallPt) {
+        /// Using pt calculated at the inner wall of TPC
+        /// Caveat: tgl still from tracking: this is not the value of tgl at the inner wall of TPC
+        const float tgl = jT.tgl();
+        trackPt = jT.tpcInnerParam() / sqrt(1 + tgl * tgl);
```
njacazio, line 399: 1 + could become 1.f +

DPG/Tasks/AOTTrack/qaMatchEff.cxx
```diff
@@ -644,6 +666,15 @@ struct qaMatchEff {
...
```
njacazio, line 675: Same here

DPG/Tasks/AOTTrack/qaMatchEff.cxx
```diff
@@ -911,6 +942,15 @@ struct qaMatchEff {
...
```
njacazio, line 951: Same here

DPG/Tasks/AOTTrack/qaMatchEff.cxx
```diff
@@ -987,6 +1027,15 @@ struct qaMatchEff {
...
```
njacazio, line 1036: Same here

njacazio (commented):
Looking good Mattia, I'm wondering if you would not prefer to have `trackPt` as a data member and have it set via a unique function to minimize code duplication, but so far it looks good!

mfaggin:
@njacazio thank you, I applied your suggestions

mfaggin:
> @njacazio thank you, I applied your suggestions

@njacazio I approve it by myself, so that tomorrow morning I can enable HL trains, in case of issues we will fix it with a new PR.

alibuild (approved):

PR #1277, opened 2022-09-21
Optimisation tree

AlexBigO:
Hi @fgrosa , @zhangbiao-phy ! I send you this first shot so you can tell me if I am going in the right direction or not because I am not so confident about the lines I added ^^
I still need to determine the type of `scores` so I can define `candidates_vector`. Maybe this issue could be overcome by rebuilding an array made of `scores[0]`, `scores[1]` and `scores[2]` (that are floats I believe) and pushing it in the vector, but it would be better to have the type of `scores` I think.

PS: I don't know why but I had issues when I tried rebasing my branch with the master branch (that's why there are remnants of the previous PR for the invariant masses ^^)

### DIFF
--- a/EventFiltering/PWGHF/HFFilter.cxx
+++ b/EventFiltering/PWGHF/HFFilter.cxx
@@ -249,6 +249,21 @@ DECLARE_SOA_TABLE(HFTrigTrain3P, "AOD", "HFTRIGTRAIN3P", //!
                   hftraining3p::FlagOrigin,
                   hftraining3p::Channel,
                   hftraining3p::HFSelBit);
+
+namespace hfoptimisationTree
+{
+DECLARE_SOA_COLUMN(CollisionIndex, collisionIndex, int);           //!
+DECLARE_SOA_COLUMN(BkgBDT, bkgBDT, float);           //!
+DECLARE_SOA_COLUMN(PromptBDT, promptBDT, float);           //!
+DECLARE_SOA_COLUMN(NonpromptBDT, nonpromptBDT, float);           //!
+DECLARE_SOA_COLUMN(DCAXY, dcaXY, float);           //!
+} // namespace hfoptimisationTree
+DECLARE_SOA_TABLE(HFOptimisationTree, "AOD", "HFOPTIMTREE", //!
+                  hfoptimisationTree::CollisionIndex,
+                  hfoptimisationTree::BkgBDT,
+                  hfoptimisationTree::PromptBDT,
+                  hfoptimisationTree::NonpromptBDT,
+                  hfoptimisationTree::DCAXY);
 } // namespace o2::aod
 
 struct AddCollisionId {
@@ -337,6 +352,8 @@ struct HfFilter { // Main struct for HF triggers
   Configurable<std::string> url{"ccdb-url", "http://alice-ccdb.cern.ch", "url of the ccdb repository"};
   Configurable<std::string> mlModelPathCCDB{"mlModelPathCCDB", "Analysis/PWGHF/ML/HFTrigger/", "Path on CCDB"};
   Configurable<long> ccdbTimestamp{"ccdb-timestamp", 0, "timestamp of the ONNX file for ML model used to query in CCDB. Exceptions: > 0 for the specific timestamp, 0 gets the run dependent timestamp"};
+  // parameter for Optimisation Tree
+  Configurable<bool> applyOptimisation{"applyOptimisation", false, "Flag to enable or disable optimisation"};
 
   // array of ONNX config and BDT thresholds
   std::array<std::string, kNCharmParticles> onnxFiles;
@@ -858,6 +875,7 @@ struct HfFilter { // Main struct for HF triggers
 
       bool isCharmTagged{true}, isBeautyTagged{true};
 
+      float scores[3] = {-1, -1, -1}; // initialize BDT scores array outside ML loop
       // apply ML models
       if (applyML && onnxFiles[kD0] != "") {
         isCharmTagged = false;
@@ -883,7 +901,9 @@ struct HfFilter { // Main struct for HF triggers
           assert(outputTensorD0.size() == outputNamesML[kD0].size() && outputTensorD0[1].IsTensor());
           auto typeInfo = outputTensorD0[1].GetTensorTypeAndShapeInfo();
           assert(typeInfo.GetElementCount() == 3); // we need multiclass
-          auto scores = outputTensorD0[1].GetTensorMutableData<float>();
+          scores[0] = outputTensorD0[1].GetTensorMutableData<float>()[0];
+          scores[1] = outputTensorD0[1].GetTensorMutableData<float>()[1];
+          scores[2] = outputTensorD0[1].GetTensorMutableData<float>()[2];
 
           if (applyML && activateQA) {
             hBDTScoreBkg[kD0]->Fill(scores[0]);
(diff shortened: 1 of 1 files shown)


fgrosa:
> Hi @fgrosa , @zhangbiao-phy ! I send you this first shot so you can tell me if I am going in the right direction or not because I am not so confident about the lines I added ^^ I still need to determine the type of `scores` so I can define `candidates_vector`. Maybe this issue could be overcome by rebuilding an array made of `scores[0]`, `scores[1]` and `scores[2]` (that are floats I believe) and pushing it in the vector, but it would be better to have the type of `scores` I think.
> 
> PS: I don't know why but I had issues when I tried rebasing my branch with the master branch (that's why there are remnants of the previous PR for the invariant masses ^^)

Hi @AlexBigO by default in O2Physics the PR are "squashed and merged", which means that all the commits in the PR are squashed in a single one. Once merged this makes diverge your fork branch with the upstream master and therefore to "realign" the two histories you first have to do `git reset --hard upstream/master`. You can do it and reopen the PR, but be careful that it will overwrite all your developments (so make sure that you saved it somewhere first 😄)

EventFiltering/PWGHF/HFFilter.cxx
```diff
@@ -249,8 +249,60 @@ DECLARE_SOA_TABLE(HFTrigTrain3P, "AOD", "HFTRIGTRAIN3P", //!
...
+DECLARE_SOA_COLUMN(DCAXYD0, dcaXYD0, float);           //!
+} // namespace hfoptimisationTree2Prong
+DECLARE_SOA_TABLE(HFOptimisationTree2Prong, "AOD", "HFOPTIMTREE2P", //!
+                  hfoptimisationTree2Prong::CollisionIndex,
+                  hfoptimisationTree2Prong::BkgBDTD0,
+                  hfoptimisationTree2Prong::PromptBDTD0,
+                  hfoptimisationTree2Prong::NonpromptBDTD0,
+                  hfoptimisationTree2Prong::DCAXYD0);
+
+namespace hfoptimisationTree3Prong
```
fgrosa, line 268: If we don't have 2-prong or 3-prong specific cuts, we do not need to define 2 tables: it can be only one (and we can have in one column the "identity" of the candidate, using for example the `beautyParticles` enum)

EventFiltering/PWGHF/HFFilter.cxx
```diff
@@ -1121,6 +1197,9 @@ struct HfFilter { // Main struct for HF triggers
               }
             }
           }
+          if (applyOptimisation) {
+            optimisationTree3Prong(collision.globalIndex(), myscores[1][0], myscores[1][1], myscores[1][2], dcaXYCharm[1]);
+          }
```
fgrosa, line 1202:
```suggestion
```

EventFiltering/PWGHF/HFFilter.cxx
```diff
@@ -1011,6 +1074,9 @@ struct HfFilter { // Main struct for HF triggers
       std::array<int8_t, kNCharmParticles - 1> isCharmTagged = is3Prong;
       std::array<int8_t, kNCharmParticles - 1> isBeautyTagged = is3Prong;
 
+      const int scoresSize = kNCharmParticles;
```
fgrosa, line 1077:
```suggestion
      const int scoresSize = kNCharmParticles - 1;
```

EventFiltering/PWGHF/HFFilter.cxx
```diff
@@ -1043,6 +1109,9 @@ struct HfFilter { // Main struct for HF triggers
             auto typeInfo = outputTensor[1].GetTensorTypeAndShapeInfo();
             assert(typeInfo.GetElementCount() == 3); // we need multiclass
             auto scores = outputTensor[1].GetTensorMutableData<float>();
+            myscores[iCharmPart +1][0] = scores[0];
+            myscores[iCharmPart +1][1] = scores[1];
+            myscores[iCharmPart +1][2] = scores[2];
```
fgrosa, line 1114:
```suggestion
            myscores[iCharmPart][0] = scores[0];
            myscores[iCharmPart][1] = scores[1];
            myscores[iCharmPart][2] = scores[2];
```

EventFiltering/PWGHF/HFFilter.cxx
```diff
@@ -1096,6 +1165,10 @@ struct HfFilter { // Main struct for HF triggers
         }
       } // end high-pT selection
 
+
+      float dcaXYCharm[scoresSize];
+      std::fill_n(dcaXYCharm, scoresSize, -99.);
+
```
fgrosa, line 1171:
```suggestion
```

EventFiltering/PWGHF/HFFilter.cxx
```diff
@@ -1113,6 +1186,9 @@ struct HfFilter { // Main struct for HF triggers
               auto massCandB = RecoDecay::m(std::array{pVec3Prong, pVecFourth}, std::array{massCharmHypos[iHypo], massPi});
               if (std::abs(massCandB - massBeautyHypos[iHypo]) <= deltaMassHypos[iHypo]) {
                 keepEvent[kBeauty4P] = true;
+                if (applyOptimisation) {
+                  dcaXYCharm[iHypo+1] = track.dcaXY();
+                }
```
fgrosa, line 1191:
```suggestion
                if (applyOptimisation) {
                  optimisationTree2Prong(collision.globalIndex(), myscores[iHypo-1][0], scores[iHypo-1][1], scores[iHypo-1][2], track.dcaXY());
                }
```

fgrosa (changes requested):
Hi @AlexBigO thanks a lot for the implementation! As a general comment I would fill only one table (with both 2 prongs and 3prongs) adding a column to keep track of the identity of the charm particle. In addition, you should also fill the table for the B0 -> D*+pi channel

EventFiltering/PWGHF/HFFilter.cxx
fgrosa: Yes this would be great!

AlexBigO: In the loops, we have access to the iterator index (`iHypo`) but I don't know how to get the string inside `beautyParticles` enum from this.
Actually, it depends on the kind of ID we want:

- an `int`: in this case I just need to drop the index corresponding to the index of the particle in `beautyParticles` (but it would require to know which index corresponds to which particle afterwards)
- a `std::string` but this would require (in the case of Bplus) to be able to get the `kBplus` from `beautyParticles` if the index is `0`.

EventFiltering/PWGHF/HFFilter.cxx
AlexBigO: @fgrosa , I don't understand why we should put `iHypo - 1` instead of `iHypo`. 
In a general manner, I am a bit confused with the `++iteratorIndex` (because I am more used to `iteratorIndex++` I guess ^^). 
For instance, in the the loop over charm particles, with index `iCharmPart`, we use `++iCharmPart`. As it has been initialized to `0`, its value should be `1` for the first iteration. Here, we write `myscores[iCharmPart][0] = scores[0]` (while I would have gone for  `myscores[iCharmPart - 1][0] = scores[0]` so that we would have `myscores[0][0 = scores[0]` for the first iteration).
Now, in the loop over beauty particles, the index is `iHypo` (also initialized at `0` and incremented via `++iHypo`). There, we fille the optimisation tree with `myscores[iHypo - 1][0]`. 
I think I am missing something because I don't see why we put `iCharmPart` in one case and `iHypo - 1` in the other :)

fgrosa: That's a typo from my side actually, it should simply be `myscores[iHypo]`, sorry for the confusion! :)

AlexBigO: No problem. Actually, it made me realized that I understood the `++i` vs `i++` thing wrong ^^ For some reason I thought `++i` (`i++`) incremented `i` at the beginning (end) of the loop but it is not the case. I did some tests and it gives the same iteration values for `i` ^^

AlexBigO: Actually, I believe that my initial belief is true when we don't manipulate a `for` loop iterator :)

fgrosa: > Actually, I believe that my initial belief is true when we don't manipulate a `for` loop iterator :)

indeed ;)

fgrosa (approved):

PR #1286, opened 2022-09-23
Adding fwdtrack-extension task

mcoquet642:
Adding fwdtrack-extension task, similar to the track-extension task, to compute forward track DCA then stored in PWGDQ data model.


jgrosseo (changes requested):
Hi!
Thanks for this code. This is indeed close to the `trackextension` and not the `trackpropagation`. Can you confirm? The latter is producing also updated track parameters, but here you only store the DCA (which is fine of course). Will a trackpropagation be needed in the future? If yes, shall we right away go this direction?

Do you really need the cov matrix for the present calculation? For the trackextension, it is not needed. Do you know where the difference comes from?

iarsene:
Hi @mcoquet642, I suggest that you also add corresponding variables in the VarManager, and fill them, so that we can already define histograms and cuts based on this. According to what we have now for the muons, these can e.g. kMuonDCAx and kMuonDCAy

mcoquet642:
Hi @iarsene, thanks for the comment. I added the MuonDCA variables to the VarManager.

iarsene (dismissed):

Common/TableProducer/fwdtrackextension.cxx
```diff
@@ -0,0 +1,70 @@
...
+#include "Math/SMatrix.h"
+#include "ReconstructionDataFormats/TrackFwd.h"
+
+using namespace o2;
+using namespace o2::framework;
+using namespace o2::framework::expressions;
+
+using SMatrix55 = ROOT::Math::SMatrix<double, 5, 5, ROOT::Math::MatRepSym<double, 5>>;
+using SMatrix5 = ROOT::Math::SVector<double, 5>;
+using MyMuonsWithCov = aod::FwdTracks;
```
jgrosseo, line 36: I propose to rename or drop this alias because it does not include `Cov`

Common/DataModel/TrackSelectionTables.h
```diff
@@ -120,6 +126,10 @@ DECLARE_SOA_TABLE(TrackSelection, "AOD", "TRACKSELECTION", //! Information on th
                   track::IsGlobalTrackWoPtEta<track::TrackCutFlag>,
                   track::IsGlobalTrackWoDCA<track::TrackCutFlag>);
 
+DECLARE_SOA_TABLE(FwdTracksDCA, "AOD", "FWDTRACKDCA", //! DCA information for the track
```
jgrosseo, line 129:
```suggestion
DECLARE_SOA_TABLE(FwdTracksDCA, "AOD", "FWDTRACKDCA", //! DCA information for the forward track
```

Common/DataModel/TrackSelectionTables.h
```diff
@@ -16,6 +16,12 @@
 
 namespace o2::aod
 {
+namespace fwdtrack
+{
+DECLARE_SOA_COLUMN(FwdDcaX, fwddcaX, float); //! Impact parameter in X of forward track to the primary vertex
```
jgrosseo, line 21:
```suggestion
DECLARE_SOA_COLUMN(FwdDcaX, fwdDcaX, float); //! Impact parameter in X of forward track to the primary vertex
```
(Usually this is the schema we follow, we only change the capitalization of the first character

Common/DataModel/TrackSelectionTables.h
```diff
@@ -16,6 +16,12 @@
...
+DECLARE_SOA_COLUMN(FwdDcaY, fwddcaY, float); //! Impact parameter in X of forward track to the primary vertex
```
jgrosseo, line 22:
```suggestion
DECLARE_SOA_COLUMN(FwdDcaY, fwdDcaY, float); //! Impact parameter in X of forward track to the primary vertex
```

PWGDQ/DataModel/ReducedInfoTables.h
```diff
@@ -248,6 +248,8 @@ DECLARE_SOA_COLUMN(Pt, pt, float);   //!
 DECLARE_SOA_COLUMN(Eta, eta, float); //!
 DECLARE_SOA_COLUMN(Phi, phi, float); //!
 DECLARE_SOA_COLUMN(Sign, sign, int); //!
+DECLARE_SOA_COLUMN(FwdDcaX, fwddcaX, float);       //!
+DECLARE_SOA_COLUMN(FwdDcaY, fwddcaY, float);       //!
```
jgrosseo, line 252: Follow the naming changes above (if you accept) and also add the comment here

jgrosseo (commented):
Thanks a lot for the updates! I made a few minor suggestions, apart from that it can go in

Common/TableProducer/fwdtrackextension.cxx
mcoquet642: Fixed, thanks.

mcoquet642:
Hi @jgrosseo, thanks again for the review. I fixed your suggested points.

jgrosseo (approved):

jgrosseo:
This also needs approval from DQ

iarsene (approved):

PR #1292, opened 2022-09-26
PID ML: Add readme for users

saganatt:
Hi @ddobrigk , let me know if this README is fine or whether you prefer adding such information to the official O2 Analysis documentation.


fcatalan92:
Hi @saganatt, I would keep this README here for the moment and then move it to the official O2 Analysis documentation when this feature is ready for "production". On top of that, in the official documentation, I would only provide the description of how to use the ML PID while the details on the inner working could be in this README.

alibuild (approved):

PR #1300, opened 2022-09-28
Fix TOF PID issue and update initializer

BongHwi:
the hasTOF is not mandatory.


alibuild (dismissed):

njacazio (approved):

PR #1303, opened 2022-09-30
DPG: update code to count ambiguous tracks

mfaggin:
- Execution for procesTrackMatch alone fixed
 - BugF fix: retrieve the track from the amb. track table via trackId(), not globalIndex()
 - New algorithm based simply on iterator and indices comparison, instead of the slow std::find


mfaggin:
ciao @njacazio , this PR for sure solves a bug that I originally introduced (I was wrongly comparing the `track.globalIndex()` with `amb_track.globalIndex()` instead of `amb_track.trackId()`) and, hoperfully, this shall avoid the huge memory consumption deriving from information storing into a vector and the usage of `std::find`

PR #1321, opened 2022-10-03
[DPG] Widen K0s production radius range

alessandrobalbino:


mfaggin (approved):
ciao @alessandrobalbino I approve this PR, but what about using `ConfigurableAxis`? This will let you choose at run time the range of your axes. See an example here https://github.com/AliceO2Group/O2Physics/blob/master/DPG/Tasks/AOTTrack/qaEventTrack.cxx#L80 
But this for next time, at the moment go on in this way

PR #1322, opened 2022-10-03
PWGDQ: Remove usage of ccdb in tableReader

lucamicheletti93:


lucamicheletti93:
The ccdb path is not used in the code and it is an example for the users. Locally it works properly but on HyperLoop it provides errors since the file is not found. For this reason I would keep it as commentes.

feisenhu (approved):

PR #1324, opened 2022-10-04
move/switch Common/DataModel includes to header_only_library

adriansev:
At this moment the includes from Common/DataModel are not copied/deployed into installation of O2Physics (so cannot be used by another application that have as dependency O2Physics). This is just NOT-WORKING initial tryout...
@TimoWilken could you please take a look at this when you have time? My immediate problem is that
pid-ml-producer use the specified includes so i tried to add to the pid-ml-producer dpl-workflow what i thought that i defined as target header only library O2Physics::CommonDataModel .. 
but cmake complained that the target was not found..
What did i missed?
Thanks a lot!!


TimoWilken:
Hi @adriansev, I can take a look next week.

adriansev:
superseded by https://github.com/AliceO2Group/O2Physics/pull/1353

PR #1329, opened 2022-10-05
PWGHF: fix B-field Run 3 GRP object for PV refit.

mfaggin:
Hi @vkucera @cterrevo @ginnocen @fcolamar ,
I fixed the GRP object query for the magnetic field used for the PV refit, since from the first implementation the data type changed for Run 3. I have tested it locally and the execution succeded.

I put the PR in draft for the moment, as usual. Let me know what you think.

### DIFF
--- a/PWGHF/TableProducer/HFTrackIndexSkimsCreator.cxx
+++ b/PWGHF/TableProducer/HFTrackIndexSkimsCreator.cxx
@@ -27,12 +27,13 @@
 #include "Common/DataModel/TrackSelectionTables.h"
 #include "ReconstructionDataFormats/V0.h"
 #include "PWGHF/Utils/UtilsDebugLcK0Sp.h"
-#include "DetectorsVertexing/PVertexer.h"     // for PV refit
-#include "ReconstructionDataFormats/Vertex.h" // for PV refit
-#include "CCDB/BasicCCDBManager.h"            // for PV refit
-#include "DataFormatsParameters/GRPObject.h"  // for PV refit
-#include "DetectorsBase/Propagator.h"         // for PV refit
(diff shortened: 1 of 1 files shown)


vkucera:
Hi @mfaggin , thanks for the update. I have a few questions/comments:
- I think it makes sense to rename `isRun2GRP` to `isRun2` and use it as a general switch between Run2/3 code or configuration, since further differences may be needed in future.
- The default values of the new configurables `ccdb...` look like something quite generic and stable. Do they need to be configurable? If so, I would think these should be set somewhere in the CCDB manager code instead. 
- Last but not least, you can add yourself in the list of authors, since you have contributed significantly to this code.

mfaggin:
> Hi @mfaggin , thanks for the update. I have a few questions/comments:
> 
> * I think it makes sense to rename `isRun2GRP` to `isRun2` and use it as a general switch between Run2/3 code or configuration, since further differences may be needed in future.
> * The default values of the new configurables `ccdb...` look like something quite generic and stable. Do they need to be configurable? If so, I would think these should be set somewhere in the CCDB manager code instead.
> * Also please propagate the new parameters in the validation framework.
> * Last but not least, you can add yourself in the list of authors, since you have contributed significantly to this code.

Hi @vkucera I have implemented the small change you asked for `isRun2`. For what concerns the `ccdb` object path I would stay conservative and leave them `Configurables` for the moment. I experienced in the recent past a few issues when running code in some data samples because some files where not in the default ccdb path (in that case it was an issue for the TOF PID), but in general I would stay like this to be flexible and not prevent anaylyses because of it.

Let me know if everything is fine.

vkucera:
> 

OK, I see. Thanks for your explanation. Let's keep it then.

vkucera (approved):

PR #1347, opened 2022-10-11
Suppress the Bayesian PID for the time being

victor-gonzalez:


alibuild (approved):

PR #1350, opened 2022-10-12
Change type Run2 trigger aliaslabels

lvermunt:
Change of type `aliasLabels` from `const char*` to `std::string` is required to use it as a lookup table to find the corresponding enum value from a configurable string (which is the strategy in HF).

To me this looks as a trivial change, but maybe I'm missing something why it should be a `char*` array? @ekryshen @jgrosseo ?

If it should remain a `char*` array, we should think of a better strategy to do the trigger selection in HF, since right now events are selected based on the condition that the **next** event has a kINT7 trigger enabled (independent of what trigger you set in the json file). @vkucera

### DIFF
--- a/PWGHF/TableProducer/HFTrackIndexSkimsCreator.cxx
+++ b/PWGHF/TableProducer/HFTrackIndexSkimsCreator.cxx
@@ -109,12 +109,14 @@ struct HfTagSelCollisions {
   Configurable<double> chi2Max{"chi2Max", 0., "max. chi^2 of primary-vertex reconstruction"};
   Configurable<std::string> triggerClassName{"triggerClassName", "kINT7", "trigger class"};
   Configurable<bool> useSel8Trigger{"useSel8Trigger", false, "use sel8 trigger condition, for Run3 studies"};
-  int triggerClass = std::distance(aliasLabels, std::find(aliasLabels, aliasLabels + kNaliases, triggerClassName.value.data()));
+  int triggerClass;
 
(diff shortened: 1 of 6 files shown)


ekryshen:
Fine with me to change char* to std::string

vkucera (approved):
Hi @lvermunt , thanks a lot for fixing this. The HF part looks fine to me.

vkucera:
I noticed that you fill the primary vertex histograms only in the `processTrigSel` function. I guess we should fill them in `processNoTrigSel` as well, right?

jgrosseo (dismissed):

PWGHF/TableProducer/HFTrackIndexSkimsCreator.cxx
```diff
@@ -205,6 +211,9 @@ struct HfTagSelCollisions {
     // selected events
     if (fillHistograms && statusCollision == 0) {
       registry.fill(HIST("hEvents"), 2);
+      registry.fill(HIST("hPrimVtxX"), collision.posX());
+      registry.fill(HIST("hPrimVtxY"), collision.posY());
+      registry.fill(HIST("hPrimVtxZ"), collision.posZ());
```
vkucera, line 216: Repeat in `processNoTrigSel`.

vkucera (changes requested):

vkucera (approved):
Thanks

jgrosseo (approved):

lvermunt:
@iarsene, @dsekihat, or @feisenhu, could one of you also approve? Had to touch the DQ code since I changed the type of aliasLabels from char to std::string. Thanks!

iarsene (approved):

PR #1351, opened 2022-10-12
Fix for Mac not accepting not override marked overriding methods

victor-gonzalez:


alibuild (approved):

jgrosseo:
I disabled the auto merge because in this way you can see the macos CI output

victor-gonzalez:
Thanks!

victor-gonzalez:
This one is a bit strange because I thought `uint64_t` was an `unsigned long`
I will try the suggested format and see if it goes through Ubuntu

victor-gonzalez:
@jgrosseo So with the suggested formatting from the Mac compiler Ubuntu gives this warning
```
/localone/alice/O2Physics/PWGCF/TwoParticleCorrelations/Core/SelectionFilterAndAnalysis.cxx:63:37: warning: format '%llx' expects argument of type 'long long unsigned int', but argument 2 has type 'long unsigned int' [-Wformat=]
   63 |     str += TString::Format("0x%016llx", option);
      |                               ~~~~~~^   ~~~~~~
      |                                     |   |
      |                                     |   long unsigned int
      |                                     long long unsigned int
      |                               %016lx
```
and suggests the opposite to Mac
Any suggestion?

victor-gonzalez:
Will an explicit cast `u_long(option)` work in Mac?

jgrosseo:
This seems to work for both: https://github.com/jgrosseo/O2Physics/blob/master/Tutorials/src/compatibleBCs.cxx#L36

victor-gonzalez:
But I guess the issue happens when requiring the hexadecimal `x` format

alibuild (approved):

PR #1352, opened 2022-10-12
PWGDQ: fix error exception on vector when running on different datasets

robincaron13:


robincaron13:
Hi @iarsene,
This is the fix that we discussed already. I hope it should be fine now on hyperloop.

iarsene (approved):

PR #1353, opened 2022-10-12
Some CMake fixes/improvements

aphecetche:
@hristov @TimoWilken  A few fixes / improvements (hopefully) in some CMake files for your consideration : 

- the `o2physics_add_header_only_library` was creating libraries in the `O2::` namespace instead of `O2Physics::' one (which might explain the failed attempt by @adriansev #1324) 
- with the fix above DataModel can now be a proper (header-only) library
- the two added CMake find modules are not strictly necessary for a stack completely driven by aliBuild, but helps in case e.g. ONNXRuntime is installed by other means (and should not hurt 100% alibuild stacks)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,7 +9,7 @@
 # granted to it by virtue of its status as an Intergovernmental Organization
 # or submit itself to any jurisdiction.
 
-cmake_minimum_required(VERSION 3.17 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.19 FATAL_ERROR)
 
 project(O2Physics
         VERSION 0.0.1
@@ -18,8 +18,6 @@ project(O2Physics
 
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 set_property(GLOBAL PROPERTY REPORT_UNDEFINED_PROPERTIES)
-
-include(GNUInstallDirs)
 
 cmake_host_system_information(RESULT _totalmem QUERY TOTAL_PHYSICAL_MEMORY)
 math(EXPR _total_analysis_jobs "(${_totalmem}-4096)/4096")
@@ -35,12 +33,6 @@ if(NOT DEFINED CMAKE_CXX_STANDARD)
   set(CMAKE_CXX_STANDARD 17)
   set(CMAKE_CXX_STANDARD_REQUIRED TRUE)
 endif()
-
-# Dependencies
-find_package(O2 CONFIG REQUIRED)
-find_package(ONNXRuntime::ONNXRuntime CONFIG)
-
-feature_summary(WHAT ALL FATAL_ON_MISSING_REQUIRED_PACKAGES)
 
 include(O2PhysicsDefineOptions)
 o2physics_define_options()
@@ -64,8 +56,6 @@ include(O2PhysicsTargetRootDictionary)
 include(O2PhysicsSetROOTPCMDependencies)
 include(O2PhysicsDataFile)
 include(AddRootDictionary)
-
-include_directories("${CMAKE_CURRENT_SOURCE_DIR}")
 
 # Main targets of the project in various subdirectories. Order matters.
 add_subdirectory(Common)
--- a/Common/Core/CMakeLists.txt
+++ b/Common/Core/CMakeLists.txt
@@ -14,7 +14,7 @@ o2physics_add_library(AnalysisCore
                        OrbitRange.cxx
                        PID/ParamBase.cxx
                        TrackSelectionDefaults.cxx
-               PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::AnalysisCore ROOT::EG O2::CCDB ROOT::Physics)
+               PUBLIC_LINK_LIBRARIES O2::Framework ROOT::EG O2::CCDB ROOT::Physics)
 
 o2physics_target_root_dictionary(AnalysisCore
               HEADERS   TrackSelection.h
--- a/Common/DataModel/CMakeLists.txt
+++ b/Common/DataModel/CMakeLists.txt
@@ -9,9 +9,4 @@
 # granted to it by virtue of its status as an Intergovernmental Organization
 # or submit itself to any jurisdiction.
 
-add_subdirectory(DataModel)
-add_subdirectory(Core)
-add_subdirectory(CCDB)
-add_subdirectory(Tasks)
-add_subdirectory(TableProducer)
-add_subdirectory(Tools)
+o2physics_add_header_only_library(DataModel)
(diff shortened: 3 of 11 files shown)


aphecetche:
obviously the `fjcontrib` change is less innocent that what I thought from my tests on my Mac and a dev FLP.
Investigating...

adriansev:
@aphecetche Thanks a lot for looking at this!!
looking at cmake/O2PhysicsAddHeaderOnlyLibrary.cmake it seems to me that it requires to move all includes to a `include` sub-directory (but of course this will break other tasks)
Looking at O2 usage of o2_add_header_only_library seems that this is how it was used ..

jgrosseo:
What is a o2physics_add_header_only_library?

We do not need the DataModel headers in a library. They are only includes for somewhere else.

aphecetche:
@adriansev first, note that I've just force-pushed a new version of the `o2_add_header_only_library` (in case you've looked at the previous one), as the previous version failed at install stage.

As far as I can tell (but I'm no O2Physics expert) the location of the (installed) include is different from O2 (where a XXX target includes get installed under `INSTALLROOT/include/XXX/include`. For `O2Physics` the installed includes are in a single flat top directory. Looking e.g. on lxplus I get : 

```
[O2Physics/nightly-20220926-2] ~ > ls $O2PHYSICS_ROOT/include
AnalysisCompositeCut.h	    CorrelationContainer.h  EventSelectionParams.h  GFWCumulant.h	JetFinder.h  MixingHandler.h  ProfileSubset.h		TriggerAliases.h  multGlauberNBDFitter.h
AnalysisConfigurableCuts.h  DGCutparHolder.h	    FlowContainer.h	    GFWWeights.h	MCProng.h    OrbitRange.h     TrackSelection.h		VarManager.h	  multMCCalibrator.h
AnalysisCut.h		    DGPIDSelector.h	    GFW.h		    HistogramManager.h	MCSignal.h   ParamBase.h      TrackSelectionDefaults.h	multCalibrator.h  pidTPCML.h
```   

I believe my (latest) change in `o2_add_header_only_library` does that.

aphecetche:
@jgrosseo the idea of the `o2physics_add_header_only_library` is to define a CMake target that only contains, as the name implies, only headers. In retrospect maybe the proper name would be `o2physics_add_header_only_target`, but I guess the name (and the code) was derived from the O2 version...  It allows to properly express in some `CMakeLists.txt` that some other libraries do depend on those includes, and to avoid the usage of the `include_directories` command in the top `CMakeLists.txt` which is not using the concept of targets at all and should be avoided.  

Whether we install or not those includes (for instance to allow usage from something depending on `O2Physics` is another story, and I must confess I've not checked whether or not that's working).

jgrosseo:
> @jgrosseo the idea of the `o2physics_add_header_only_library` is to define a CMake target that only contains, as the name implies, only headers. In retrospect maybe the proper name would be `o2physics_add_header_only_target`, but I guess the name (and the code) was derived from the O2 version... It allows to properly express in some `CMakeLists.txt` that some other libraries do depend on those includes, and to avoid the usage of the `include_directories` command in the top `CMakeLists.txt` which is not using the concept of targets at all and should be avoided.
> 
> Whether we install or not those includes (for instance to allow usage from something depending on `O2Physics` is another story, and I must confess I've not checked whether or not that's working).

OK, thanks! So in a library is not created. Then this is good of course.

I also confirm that the header layout is flat in O2Physics. We also decided to install all headers but just those which are needed downstream.

adriansev:
> @adriansev first, note that I've just force-pushed a new version of the `o2physics_add_header_only_library` (in case you've looked at the previous one), as the previous version failed at install stage.
> 
> As far as I can tell (but I'm no O2Physics expert) the location of the (installed) include is different from O2 (where a XXX target includes get installed under `INSTALLROOT/include/XXX/include`. For `O2Physics` the installed includes are in a single flat top directory. Looking e.g. on lxplus I get :
> 
> ```
> [O2Physics/nightly-20220926-2] ~ > ls $O2PHYSICS_ROOT/include
> AnalysisCompositeCut.h	    CorrelationContainer.h  EventSelectionParams.h  GFWCumulant.h	JetFinder.h  MixingHandler.h  ProfileSubset.h		TriggerAliases.h  multGlauberNBDFitter.h
> AnalysisConfigurableCuts.h  DGCutparHolder.h	    FlowContainer.h	    GFWWeights.h	MCProng.h    OrbitRange.h     TrackSelection.h		VarManager.h	  multMCCalibrator.h
> AnalysisCut.h		    DGPIDSelector.h	    GFW.h		    HistogramManager.h	MCSignal.h   ParamBase.h      TrackSelectionDefaults.h	multCalibrator.h  pidTPCML.h
> ```
> 
> I believe my (latest) change in `o2physics_add_header_only_library` does that.

@aphecetche Thanks a lot!! Yes, i did not see it, checking it out right away

adriansev:
> > > > @jgrosseo the idea of the `o2physics_add_header_only_library` is to define a CMake target that only contains, as the name implies, only headers. In retrospect maybe the proper name would be `o2physics_add_header_only_target`, but I guess the name (and the code) was derived from the O2 version... It allows to properly express in some `CMakeLists.txt` that some other libraries do depend on those includes, and to avoid the usage of the `include_directories` command in the top `CMakeLists.txt` which is not using the concept of targets at all and should be avoided.
> > > > Whether we install or not those includes (for instance to allow usage from something depending on `O2Physics` is another story, and I must confess I've not checked whether or not that's working).
> > > 
> > > 
> > > OK, thanks! So in a library is not created. Then this is good of course.
> > > I also confirm that the header layout is flat in O2Physics. We also decided to install all headers but just those which are needed downstream.
> > 
> > 
> > @jgrosseo well, if one want/need O2Physics to be upstream dependency (for downstream software that dependends on O2Physics) then all (or by selection) includes are needed to be installed in $INSTALLDIR/include
> 
> Yes, selected ones should be installed. On request. We have also includes which are only used by some executable, they should not be exposed.

@jgrosseo ok, and as you can see in this PR, the target is first to be able to do it, then it's about the Common/DataModel includes which are of general usage in analyses .. are there any objections to this?

TimoWilken:
Sorry, the Mac builder has a bit of a queue at the moment. It'll pick this PR up eventually.

jgrosseo:
> Sorry, the Mac builder has a bit of a queue at the moment. It'll pick this PR up eventually.

Good. We should certainly wait for this

aphecetche:
All green (including Mac builder), so ready to be reviewed (and/or merged :wink:)

adriansev:
@jgrosseo @TimoWilken Could you review this please? Thanks a lot!!

TimoWilken:
Looks like a good effort to me in general, thank you!

One thing though: it looks like the new explicit "FastJet::Contrib" dependency relies on aliBuild loading fastjet transitively for O2. This should probably be specified explicitly in alidist as well: https://github.com/alisw/alidist/pull/4612.

adriansev:
gentle ping @TimoWilken IMHO if does not break anything it could be a situation of good enough (the intended jobs is done), and having a step forward where one can then improve further .. would it be ok to just merge it and then build upon this? Thanks a lot!!

ktf:
This is fine with me, however the fact the analysis data model is header only is just an accident at the model. In principle it might happen in the future that a full fledged library will be needed.

PR #1354, opened 2022-10-13
PWGMM: change namespace to avoid clash with O2 changes

aalkin:


alibuild (approved):

PR #1355, opened 2022-10-13
PWGDQ: fixing ambiguous histogram filling

lucamicheletti93:
The ambiguous tracks histograms were not filled when Tracks / MuonsWithCov were selected. The fix allows to process both the track types

### DIFF
--- a/PWGDQ/TableProducer/tableMaker.cxx
+++ b/PWGDQ/TableProducer/tableMaker.cxx
@@ -148,13 +148,13 @@ struct TableMaker {
     if (enableBarrelHistos) {
       if (fConfigDetailedQA) {
         histClasses += "TrackBarrel_BeforeCuts;";
-        if (fIsAmbiguous) {
+        if (fIsAmbiguous && !fIsRun2) {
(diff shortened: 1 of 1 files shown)


robincaron13:
Hi Luca, I'm wondering if you have tried to run the new code also on Run 2 converted data? to prevent to have error on the missing table of ambiguous tracks, because you add it for every table-maker processes.

lucamicheletti93:
Dear @robincaron13 , thanks for pointing the issue! Indeed this could be problematic for Run2 converted data. I have done a another commit, but I want to double-check if this fix the issue. I'll comment soon.

iarsene:
Hi Luca,
Fine with me but please fix the conflicts and the checks you mentioned.

lucamicheletti93:
Dear @iarsene and @robincaron13 , following your suggestions I have modified the tableMaker code in order to run it also with Run2 converted data. I have prepared a draft pull request so that you can have a look at the code modifications. The main difference is that now the ambiguous tracks are passed to the fullSkimming only in 2 dedicated process functions, while in the other cases the tables are not provided. Here's the link to the PR: https://github.com/AliceO2Group/O2Physics/pull/1380

PR #1356, opened 2022-10-13
PWG-LF: spectra task, add compatibility with Run2

njacazio:


alibuild (approved):

PR #1358, opened 2022-10-15
add protection from missing mcCollision

galocco:
In Run3 there can be "fake collisions". That means one might encounter collisions where there exists no corresponding MC Collision.


PWGDQ/TableProducer/tableMakerMC.cxx
```diff
@@ -244,6 +244,9 @@ struct TableMakerMC {
     uint64_t trackFilteringTag = 0;
     uint8_t trackTempFilterMap = 0;
     for (auto& collision : collisions) {
+      if (!collision.has_mcCollision()) {
```
iarsene, line 247: As we discussed, please add a //TODO:  statement here, so we can comeback to this in the future

iarsene (approved):

PR #1360, opened 2022-10-16
Made variables const where possible

pbuehler:


alibuild (approved):

PR #1361, opened 2022-10-16
Add helper header file and optimisation trees for charm and femto triggers

fgrosa:


alibuild (approved):

PR #1364, opened 2022-10-17
Add qa task for V0 analysis using derived data 

ercolessi:


njacazio (approved):
Thanks Francesca, Indeed this is the good tool to use for the job

PR #1365, opened 2022-10-18
PWGMM: dndeta: use core BC-to-collisions index

aalkin:


alibuild (approved):

PR #1375, opened 2022-10-19
Add the option to save the V0 candidates able maker and table reader

XiaozhiBai:
HI Ionut

We add the option to save the V0 candidates on the table maker and fill the ND histogram via the table reader.

We validate the code locally. 
Cheers
Xiaozhi

### DIFF
--- a/PWGDQ/Core/HistogramsLibrary.h
+++ b/PWGDQ/Core/HistogramsLibrary.h
@@ -321,4 +321,42 @@ void o2::aod::dqhistograms::DefineHistograms(HistogramManager* hm, const char* h
     hm->AddHistogram(histClass, "DeltaEta_DeltaPhi", "", false, 20, -2.0, 2.0, VarManager::kDeltaEta, 50, -8.0, 8.0, VarManager::kDeltaPhi);
     hm->AddHistogram(histClass, "DeltaEta_DeltaPhiSym", "", false, 20, -2.0, 2.0, VarManager::kDeltaEta, 50, -8.0, 8.0, VarManager::kDeltaPhiSym);
   }
+
+  if (groupStr.Contains("postcalib")) {
+    const int kNvarsPID = 4;
+    const int kTPCnsigmaNbins = 70;
+    double tpcNsigmaBinLims[kTPCnsigmaNbins + 1];
+    for (int i = 0; i <= kTPCnsigmaNbins; ++i)
+      tpcNsigmaBinLims[i] = -7.0 + 0.2 * i;
+
+    const int kPinEleNbins = 10;
+    double pinEleBinLims[kPinEleNbins + 1] = {0.2, 0.4, 0.6, 0.8, 1.0, 1.2, 1.5, 2.0, 3.0, 4.0, 6.0};
+
+    const int kEtaNbins = 9;
+    double etaBinLimsI[kEtaNbins + 1] = {-0.9, -0.7, -0.5, -0.3, -0.1, 0.1, 0.3, 0.5, 0.7, 0.9};
+
+    const int kTPCnClusterbins = 16;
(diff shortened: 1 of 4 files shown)


PWGDQ/Core/HistogramsLibrary.h
```diff
@@ -311,4 +311,42 @@ void o2::aod::dqhistograms::DefineHistograms(HistogramManager* hm, const char* h
     hm->AddHistogram(histClass, "DeltaEta_DeltaPhi", "", false, 20, -2.0, 2.0, VarManager::kDeltaEta, 50, -8.0, 8.0, VarManager::kDeltaPhi);
     hm->AddHistogram(histClass, "DeltaEta_DeltaPhiSym", "", false, 20, -2.0, 2.0, VarManager::kDeltaEta, 50, -8.0, 8.0, VarManager::kDeltaPhiSym);
   }
+
+  if (groupStr.Contains("postcalib")) {
```
iarsene, line 315: You do not need to make a new group in the library. Just make the "postcalib" as a subgroup of the "track" group. The reason for this is that it should be more or less clear from the group name, where these histograms are being filled. So for example, histograms in "track" are filled in the track loops, etc.
Then i think you do not need the "electron", pion and proton subgroups. When we postcalibrate, we normally do all the species together, so when you switch "postcalib" on, you should include all those histograms.

PWGDQ/Tasks/tableReader.cxx
```diff
@@ -999,5 +999,17 @@ void DefineHistograms(HistogramManager* histMan, TString histClasses)
     if (classStr.Contains("DileptonHadronCorrelation")) {
       dqhistograms::DefineHistograms(histMan, objArray->At(iclass)->GetName(), "dilepton-hadron-correlation");
     }
+
+    if (classStr.Contains("PIDCalib")) {
```
iarsene, line 1003: I see that you add here the histogram classes, but you did not enabled these histogram classes in the init() and I also don't see where you are filling them in the process function. You still need to add some code?

PWGDQ/Core/HistogramsLibrary.h
XiaozhiBai: HI, Ionut
Now, I moved the "postcalib" to the track histogram as you suggested, 
Cheers
Xiaozhi

PWGDQ/Tasks/tableReader.cxx
XiaozhiBai: Hi Ionut, 
Now we moved the histogram to the track group, see above. we don't need anything  init().
Cheers
Xiaozhi

PWGDQ/Tasks/tableReader.cxx
```diff
@@ -966,6 +966,15 @@ void DefineHistograms(HistogramManager* histMan, TString histClasses)
     if (classStr.Contains("Track")) {
       if (classStr.Contains("Barrel")) {
         dqhistograms::DefineHistograms(histMan, objArray->At(iclass)->GetName(), "track", "its,tpcpid,dca,tofpid");
+        if (classStr.Contains("PIDCalibElectron")) {
+          dqhistograms::DefineHistograms(histMan, objArray->At(iclass)->GetName(), "track", "electron");
```
iarsene, line 970: this would not work, because what you have now in the HistogramsLibrary is now requiring the subgroup name to contain both "postcalib" and "electron". So you should use something like "postcalib_electron".

XiaozhiBai: Hi,Ionut 
Yes, right we have, in the cut selection "cfgTrackCuts": "PIDCalibElectron,PIDCalibPion,PIDCalibProton",
Cheers
Xiaozhi

iarsene (approved):

PR #1381, opened 2022-10-21
Added configurable to select TRD presence

rosarioturrisi:


DPG/Tasks/AOTTrack/qaMatchEff.cxx
```diff
@@ -404,6 +407,20 @@ struct qaMatchEff {
     //
     for (auto& jT : jTracks) {
 
+      // choose if we keep the track according to the TRD presence requirement
+      if (b_TRDisThere && !jT.hasTRD())
+        continue;
+      if (!b_TRDisThere && jT.hasTRD())
```
mfaggin, line 413: In my opinion we do need to add a case in which we do require the TRD, but not a case in which we explicitly remove it. In other words, in my opinion we should not ask explicitly to remove the TRD, but just to see what changes if we do explicitly require it or not. Then, in my opinion this second `if` here shall be removed (here and below as well)

rosarioturrisi:
Indeed I got from the discussion that it could be enlightening the direct comparison ‘on/off’. What is missing is the default case, i.e. when you do not make a choice. I’ll add and resubmit.On 21 Oct 2022, at 11:56, Mattia Faggin ***@***.***> wrote:﻿
@mfaggin commented on this pull request.



In DPG/Tasks/AOTTrack/qaMatchEff.cxx:
> @@ -404,6 +407,20 @@ struct qaMatchEff {
     //
     for (auto& jT : jTracks) {
 
+      // choose if we keep the track according to the TRD presence requirement
+      if (b_TRDisThere && !jT.hasTRD())
+        continue;
+      if (!b_TRDisThere && jT.hasTRD())

In my opinion we do need to add a case in which we do require the TRD, but not a case in which we explicitly remove it. In other words, in my opinion we should not ask explicitly to remove the TRD, but just to see what changes if we do explicitly require it or not. Then, in my opinion this second if here shall be removed (here and below as well)

mfaggin:
apparently this PR is getting a bit dirty. @rosarioturrisi I close it and you will opena a new one when you are ready.

PR #1383, opened 2022-10-21
Bump MegaLinter and enable cpplint for C++ code

vkucera:
- Bumpt MegaLinter to v6
- enable cpplint: enforce line length 160, disable some filters
- pyright: disable some filters
- Print all processed files to check what files were picked up by the CI

cpplint enforces the Google coding guidelines which the O2 guidelines are derived from.
I disabled some of the filters that conflicted with the formatting style and a few others that should not be applied.

I propose to set the line width limit to 160 characters which is the width of the GitHub frame to guarantee easy web browsing through the code (even though the O2 guidelines prescribe 120). If agreed, this should be propagated to `ColumnLimit` in `.clang-format`.


TimoWilken:
Looks good to me in principle, thanks Vit!

I have no opinion on the line length.

There are some failures in the check now. Will those also be there in every other PR when this PR is merged, or will Megalinter only check the files changed in each PR?

aalkin:
I would avoid setting a limit on line length, what seems like a reasonable default now may become restrictive later. Majority of editors are capable of wrapping lines above certain threshold, which usually can be toggled.

vkucera:
The errors are there because of the weird way the workflow is getting the list of changed files (diffing against upstream/master at the time of running?). I'm not sure how this can be fixed. But it's unrelated to this PR. When the PR branch is up to date with upstream/master, only the changed files are linted. This does not change with this PR.

I kept the errors there at this stage on purpose, to illustrate the kind of errors cpplint reports. Once we agree on the configuration, I will rebase to get rid of them.

Concerning the line length I tend to agree that we should not enforce a limit on it. I tried to format some files with ClangFormat and the limit and it grouped some lines which actually look better broken.

vkucera:
> > The errors are there because of the weird way the workflow is getting the list of changed files (diffing against upstream/master at the time of running?). I'm not sure how this can be fixed. But it's unrelated to this PR. When the PR branch is up to date with upstream/master, only the changed files are linted. This does not change with this PR.
> > I kept the errors there at this stage on purpose, to illustrate the kind of errors cpplint reports. Once we agree on the configuration, I will rebase to get rid of them.
> 
> Hm, that's a little unfortunate. Presumably this means that anyone sending a PR will get spurious errors in files they didn't touch, if the `master` branch moves ahead of the commit they based their PR on.
> 
> How many cpp-lint errors (unrelated to the line length) are there in total? Is it feasible to fix all of them at once now?

I agree that is a problem but it is not a new one. It's been happening for some time that the MegaLinter check fails because of errors in unrelated files that have already been fixed in the upstream in the meantime. If it is a genuine bug, I can report it to the developers. But till then, one can always rebase.

Now, after formatting the repo with updated `.clang-format`, cpplint finds 2191 errors:
[build/c++11] 7
[build/header_guard] 281
[build/include] 2
[build/include_order] 15
[build/include_subdir] 3
[build/include_what_you_use] 118
[readability/alt_tokens] 235
[readability/braces] 315
[readability/casting] 445
[readability/inheritance] 41
[readability/namespace] 1
[readability/todo] 218
[runtime/arrays] 12
[runtime/explicit] 37
[runtime/int] 109
[runtime/string] 12
[runtime/threadsafe_fn] 2
[whitespace/blank_line] 102
[whitespace/empty_if_body] 2
[whitespace/ending_newline] 38
[whitespace/line_length] 5
[whitespace/newline] 28
[whitespace/parens] 62
[whitespace/semicolon] 88
[whitespace/todo] 13

It's definitely not feasible to fix them all at once. Most of them are related to good coding practice and cannot be fixed with formatting.

TimoWilken:
Hm, fair enough. The last thing I can think of: maybe this behaviour (with unrelated files being checked) is because the workflow has `on: pull_request_target`? In that case it runs on the `HEAD` commit of the PR. If it had `on: pull_request` instead, then it'd run on the merge commit between `HEAD` and `master` instead, which shouldn't show unrelated files (from `master`) as changed.

vkucera:
> Hm, fair enough. The last thing I can think of: maybe this behaviour (with unrelated files being checked) is because the workflow has `on: pull_request_target`? In that case it runs on the `HEAD` commit of the PR. If it had `on: pull_request` instead, then it'd run on the merge commit between `HEAD` and `master` instead, which shouldn't show unrelated files (from `master`) as changed.

Running it on `pull_request` did not make a difference the list of picked files.

vkucera:
Numbers of found errors with the updated filters:
[build/c++11] 7
[build/header_guard] 281
[build/include] 31
[build/include_order] 231
[build/include_subdir] 3
[build/include_what_you_use] 118
[readability/alt_tokens] 235
[readability/braces] 315
[readability/casting] 445
[readability/inheritance] 41
[readability/namespace] 1
[runtime/arrays] 12
[runtime/explicit] 37
[runtime/int] 109
[runtime/string] 12
[runtime/threadsafe_fn] 2
[whitespace/empty_if_body] 2
[whitespace/ending_newline] 38
[whitespace/newline] 28
[whitespace/parens] 62

vkucera:
Ready!

TimoWilken (approved):

PR #1395, opened 2022-10-26
[Event filtering] Add collision table and fix loading BDT models from CCDB

fgrosa:


alibuild (approved):

PR #1415, opened 2022-11-02
Add a script to find dependencies for producing tables and running workflows

vkucera:
OBSOLETE:

This script helps to quickly figure out which dependency is missing when a tree cannot be found.
It finds the source files where a provided table is produced or spawned and the names of the corresponding executables.
It basically does automatically the search that one is advised to do when following the [link](https://aliceo2group.github.io/analysis-framework/docs/troubleshooting/#tree-not-found) in the error message and the instructions.

EDIT: Replaced with a proper Python script that processes JSON files.

Proposed documentation paragraph:

This scripts finds out which workflows can produce a given table and which tables are needed to run a given workflow. These dependencies are reported in the terminal in the form of a nested tree and optionally visualised graphically in a figure.
- Depending on the specified maximum number of levels of the workflow tree, one can request any dependency depth, from the direct dependencies to the full workflow topology leading to the provided table(s) and/or workflow(s).
- By default, all possible ways of producing a table and all possible ways of running a workflow are considered. This can be customised by providing a list of tables and workflows that should be excluded from the search.
- Since the "Couldn't get TTree" error message in O2 reports the lowercase name of the missing table, table names are treated as case-insensitive by default so that one can just copy-paste the name from the error message.

```
[O2Physics/latest] ~/alice/O2Physics/Scripts $> ./find_dependencies.py -w o2-analysis-hf-task-lc-to-k0s-p -g png -l -1 -x o2-analysis-alice3-trackextension o2-analysis-mc-converter o2-analysis-trackextension -h
usage: find_dependencies.py [-h] [-t TABLE [TABLE ...]] [-w WORKFLOW [WORKFLOW ...]] [-c] [-g {pdf,svg,png}] [-x EXCLUDE [EXCLUDE ...]] [-l LEVELS]

Find dependencies required to produce a given table or to run a given workflow.

optional arguments:
  -h, --help            show this help message and exit
  -t TABLE [TABLE ...]  table(s)
  -w WORKFLOW [WORKFLOW ...]
                        workflow(s)
  -c                    be case-sensitive with table names
  -g {pdf,svg,png}      make a topology graph in a given format
  -x EXCLUDE [EXCLUDE ...]
                        tables and workflows to exclude
  -l LEVELS             maximum number of workflow tree levels (default = 0, include all if < 0)
```
Example:
```
[O2Physics/latest] ~/alice/O2Physics/Scripts $> ./find_dependencies.py -w o2-analysis-hf-task-d0 -t tracksdca -g png -l 1 -x o2-analysis-alice3-trackextension o2-analysis-mc-converter o2-analysis-trackextension

Table: tracksdca

tracksdca <- ['o2-analysis-track-propagation']

Workflow dependency tree:

o2-analysis-track-propagation <- ['BCs', 'Collisions', 'Timestamps', 'TracksCov_IU', 'Tracks_IU']
  Timestamps <- ['o2-analysis-timestamp']
    o2-analysis-timestamp <- ['BCs']

Workflow: o2-analysis-hf-task-d0

o2-analysis-hf-task-d0 <- ['HfCand2ProngBase', 'HfCand2ProngExtExtension', 'HfCand2ProngMcGen', 'HfCand2ProngMcRec', 'HfSelD0', 'McParticles', 'McTrackLabels', 'Tracks', 'TracksCov', 'TracksExtra']
  HfCand2ProngBase <- ['o2-analysis-hf-candidate-creator-2prong']
    o2-analysis-hf-candidate-creator-2prong <- ['BCs', 'Collisions', 'Hf2Prongs', 'HfPvRefit2Prong', 'Timestamps', 'Tracks', 'TracksCov', 'TracksExtra', 'HfCand2ProngBase', 'McParticles', 'McTrackLabels']
  HfCand2ProngExtExtension <- ['o2-analysis-hf-candidate-creator-2prong']
  HfCand2ProngMcGen <- ['o2-analysis-hf-candidate-creator-2prong']
  HfCand2ProngMcRec <- ['o2-analysis-hf-candidate-creator-2prong']
  HfSelD0 <- ['o2-analysis-hf-candidate-selector-d0']
    o2-analysis-hf-candidate-selector-d0 <- ['HfCand2ProngBase', 'HfCand2ProngExtExtension', 'Tracks', 'TracksCov', 'TracksDCA', 'TracksExtra', 'pidTOFFullEl', 'pidTOFFullKa', 'pidTOFFullMu', 'pidTOFFullPi', 'pidTOFFullPr', 'pidTPCFullEl', 'pidTPCFullKa', 'pidTPCFullMu', 'pidTPCFullPi', 'pidTPCFullPr']
  Tracks <- ['o2-analysis-track-propagation']
  TracksCov <- ['o2-analysis-track-propagation']

Making dot file in: tracksdca_o2-analysis-hf-task-d0.gv
Making graph in: tracksdca_o2-analysis-hf-task-d0.png
```
tracksdca_o2-analysis-hf-task-d0


aalkin:
Good idea! You may want to add `Builds` as well, for completeness.

vkucera:
> Good idea! You may want to add `Builds` as well, for completeness.

Thanks, I was wondering whether I was missing some keywords.
I also extended the `grep` pattern for the table string.

Could you check whether this now covers all the cases?

jgrosseo:
Wouldn't it be better to base this script on the json files of the workflows that are produced during compilation? The list the tables that are produces and this would cover all cases, e.g. also if someone has some macros which create the Builds statement etc.

vkucera:
> Wouldn't it be better to base this script on the json files of the workflows that are produced during compilation? They list the tables that are produces and this would cover all cases, e.g. also if someone has some macros which create the Builds statement etc.

Agreed. I will make a better version, processing the dpl json files.

ddobrigk:
Hi @vkucera, did you manage to play a bit more with this? As I discussed with you in the ALICE week dinner, this is a super nice concept and in fact we should even consider creating a table producer "map" (perhaps automatically) to be printed out perhaps at runtime error time or to be uploaded to the documentation for tree_not_found (in a similar way as the auto-generated documentation). Thank you very much!

vkucera:
> 

Hi @ddobrigk , I had a look at the way of doing it with the JSON files and it works fine for whatever can run on AliHyperloop. The only drawback is that it does not work for tasks that are pushed into workflows using command line options which are false by default and therefore do not appear inside the JSON files generated during the compilation. But these cannot run on AliHyperloop anyway, so maybe it would be fine.

ddobrigk (approved):
Many many thanks for the initiative, Vit!

PR #1419, opened 2022-11-02
PID: Add table for TOF only event time, add event time QA

njacazio:


njacazio:
This PR also fixes the units of the FT0 that are used for the TOF PID

ddobrigk (approved):
Looks good as far as I can tell, thanks!

alibuild (approved):

PR #1426, opened 2022-11-04
PWGHF: Lc validation histograms

AnnalenaKa:
Adding histograms for the Lc validation


lvermunt:
Hi @AnnalenaKa, thanks a lot! I repeat also here what I wrote to you in chat so others are also aware: I don't agree with adding the `Lc` suffix in the configurable names. This is something one should fix in the json loading in the corresponding AliPhyics validation task. We shouldn't change O2Physics for this.

More feedback later.

PWGHF/Tasks/taskLc.cxx
```diff
@@ -13,6 +13,7 @@
 /// \brief Λc± → p± K∓ π± analysis task
 /// \note Extended from taskD0
 ///
+/// \author Annalena Kalteyer <annalena.sophie.kalteyer@cern.ch>, GSI Darmstadt
```
vkucera, line 16: I would say it is clearer to add authors in chronological order.

PWGHF/Tasks/taskLc.cxx
```diff
@@ -108,67 +259,191 @@ struct TaskLc {
...
       if (cutYCandMax >= 0. && std::abs(YLc(candidate)) > cutYCandMax) {
         continue;
       }
+
       if (std::abs(candidate.flagMCMatchRec()) == 1 << DecayType::LcToPKPi) {
         // Get the corresponding MC particle.
         auto indexMother = RecoDecay::getMother(particlesMC, candidate.index0_as<aod::BigTracksMC>().mcParticle_as<soa::Join<aod::McParticles, aod::HfCandProng3MCGen>>(), pdg::Code::kLambdaCPlus, true);
         auto particleMother = particlesMC.rawIteratorAt(indexMother);
-        registry.fill(HIST("hPtGenSig"), particleMother.pt()); // gen. level pT
-        auto ptRec = candidate.pt();
```
vkucera, line 155: Why did you remove `ptRec`?

vkucera (changes requested):
In general, it would make sense to declare variables for the quantities that are repeatedly filled in many histograms.

PWGHF/Tasks/taskLc.cxx
AnnalenaKa: Sure. I moved the line to the end of the list. I was not sure what the convention was.

AnnalenaKa:
> In general, it would make sense to declare variables for the quantities that are repeatedly filled in many histograms.

Okay, I replaced the repeated calls of the pt when willing the histograms.

PWGHF/Tasks/taskLc.cxx
```diff
@@ -107,68 +258,193 @@ struct TaskLc {
...
-        registry.fill(HIST("hCPARecSig"), candidate.cpa());
-        registry.fill(HIST("hDecLengthRecSig"), candidate.decayLength());
-        registry.fill(HIST("hDca2RecSig"), candidate.chi2PCA());
-        registry.fill(HIST("hEtaRecSig"), candidate.eta());
-      } else {
-        registry.fill(HIST("hPtRecBg"), candidate.pt());
-        registry.fill(HIST("hCPARecBg"), candidate.cpa());
-        registry.fill(HIST("hDecLengthRecBg"), candidate.decayLength());
-        registry.fill(HIST("hDca2RecBg"), candidate.chi2PCA());
-        registry.fill(HIST("hEtaRecBg"), candidate.eta());
```
vkucera, line 171: These histograms should not be removed.

PR #1430, opened 2022-11-04
TOF: add more plots for the event time + more utils

njacazio:


ddobrigk (approved):
Also looks reasonable, thanks for the extra work also on the QA part!

alibuild (approved):

PR #1432, opened 2022-11-04
[PWGHF] Fix histogram name in Ds task

rebeccacerri:


fcatalan92:
@stefanopolitano @vkucera @lvermunt Rebecca and I fixed a couple of typos that prevent from running the Ds task on MC.

lvermunt (approved):

PR #1422, opened 2022-11-03
PWGEM: added lmee-lf-cocktail task

DanielSamitz:
Added first version of the lmee-lf-cocktail task.
(corresponds to https://github.com/alisw/AliPhysics/blob/master/PWGDQ/dielectron/LMEE/AliAnalysisTaskLMeeCocktailMC.cxx in Run2)

PR #1435, opened 2022-11-04
Fix Compilation on macOS

ktf:


ktf:
https://github.com/AliceO2Group/O2Physics/pull/1422 broke macOS builds.

Notice also that lmeeLFCocktail.cxx seems to open local files. Is that the intended behaviour? I think that will not work in the trains, no?

@ddobrigk @jgrosseo

ktf:
@TimoWilken is it normal that the O2Physics tests do not have macOS builds anymore? What happened?

TimoWilken:
@ktf Nothing's changed. It's just that the Mac check isn't marked as required, so GitHub doesn't show it as "expected" before the CI picks it up. It'll only show up as yellow when the CI first sets its status as pending.

ivorobye:
hi @ktf thanks a lot for quick fix, indeed it was weird that build on macOS was not checked automatically (can't we just mark the Mac check as required?) As for the access of the local files, this indeed won't work atm on train, but we are aware of it and @DanielSamitz will continue to improve the task

ivorobye (approved):

dsekihat:
Hi @DanielSamitz , let's discuss where to store pT parameterizations for LF hadrons. I guess CCDB is the best place to store them.

DanielSamitz:
@dsekihat sure. I have no experience with that, so wherever you think is suitable is fine for me.

PR #1446, opened 2022-11-06
added the first version of the TPC post-calibration maps for the global testing

XiaozhiBai:
Hi,
We added the first version of the TPC post-calibration maps for the global… testing, this is temporary testing.
Xiaozhi


iarsene:
Hi @XiaozhiBai,
I will approve these changes so that you can continue your checks. However, as I mentioned a few times in the meetings, we will get the postcalibration maps via the CCDB, which is time sensitive, so we should not need to hardcode data periods in the VarManager. So, for the next developments, we have to load the calibration histograms from the CCDB, with the timestamp corresponding to the data one analyzes. If no calibration hisotgrams are available yet for some data periods, we should define good defaults.

iarsene (approved):

PR #1454, opened 2022-11-10
cmake:: tryout for fixing _add_header_only_library

adriansev:
@TimoWilken @ktf @aphecetche Hi guys! this is another tryout for making _add_header_only_library function to work. I mirrored the exact content from O2 but then there are a lots of dataflows that no longer find the corresponding includes from Common/DataModel where i try to apply this function. To reiterate, the purpose is to have the headers from Common/DataModel also in INSTALLDIR/include directory.
@aphecetche I mentioned you that i seen that you were the original committer from O2
Thanks a lot for any help!


aphecetche:
@adriansev well, I thought the includes of DataModel were actually installed. Will have a look and let you know.

PR #1459, opened 2022-11-11
PWGMM: dndeta: update

aalkin:
* Separate reassignment table and refitted parameters table
* Use original track parameters in histograms
* Add separate DCA histograms for reassigned tracks


alibuild (approved):

PR #1461, opened 2022-11-11
PWGCF: fix compiler warning in FemtoDream framework

ariedel-cern:
Warning occurs if the loop object is constant, but we we create a copy of it and do not use a reference. This is fixed with this patch. 

Also containing some changes to pass MegaLinter checks.


ariedel-cern:
adding @lauraser and @jgrosseo , who requested the patch

ariedel-cern:
MegaLinter failed because it did not manage to pull the docker image. Can the check be restarted?

victor-gonzalez (approved):

PR #1465, opened 2022-11-14
Extend V0 tracking QC

njacazio:
- catch runtime errors of secondary vertexer


njacazio:
@alessandrobalbino I added the check on the IU, could you have a look please?
@ddobrigk I catch the exception of invalid cov matrix, what do you think?

alibuild (approved):

PR #1471, opened 2022-11-14
Policy entity for orchestrating skimming and analysis

victor-gonzalez:
Now the signature of the skimming configuration is stored in the CCDB and the check of the consistency between skimmed data configuration and analysis configuration is supported


victor-gonzalez:
Check https://alice.its.cern.ch/jira/browse/PWGCF-176# for evolution and pending functionality

alibuild (approved):

PR #1473, opened 2022-11-15
Adding Configurable to set number of threads for ONNX

ChSonnabend:
It was noticed that upon running within singularity containers on a SLURM cluster, ONNX has difficulties to allocate the correct resources if the number of CPU's is limited. A manually set value for the number of available threads fixed this issue. This should not affect the code any further, since the default value lets ONNX choose the number of threads automatically.

### DIFF
--- a/Common/TableProducer/PID/pidTPC.cxx
+++ b/Common/TableProducer/PID/pidTPC.cxx
@@ -156,13 +156,12 @@ struct tpcPid {
 
       /// Testing hyperloop core settings
       const char* alien_cores = gSystem->Getenv("ALIEN_JDL_CPUCORES");
-      if(alien_cores!=NULL){
+      if (alien_cores != NULL) {
         std::string msg_cores = "Hyperloop-run detected! Number of cores = ";
         LOGP(info, msg_cores + std::string(alien_cores));
         activeThreads = 1;
-      }
-      else{
-        if(networkSetNumThreads>0){
+      } else {
+        if (networkSetNumThreads > 0) {
           LOGP(info, "Not running on Hyperloop. Threads for neural network inference are fixed: {} threads", std::to_string(networkSetNumThreads));
         }
       }
--- a/Common/TableProducer/PID/pidTPC.cxx
+++ b/Common/TableProducer/PID/pidTPC.cxx
@@ -24,9 +24,9 @@
 #include "TSystem.h"
 
 // O2 includes
+#include <CCDB/BasicCCDBManager.h>
 #include "Framework/AnalysisTask.h"
 #include "ReconstructionDataFormats/Track.h"
-#include <CCDB/BasicCCDBManager.h>
 #include "CCDB/CcdbApi.h"
 #include "Common/DataModel/PIDResponse.h"
 #include "Common/Core/PID/TPCPIDResponse.h"
@@ -78,7 +78,7 @@ struct tpcPid {
   Configurable<std::string> paramfile{"param-file", "", "Path to the parametrization object, if empty the parametrization is not taken from file"};
   Configurable<std::string> url{"ccdb-url", "http://alice-ccdb.cern.ch", "url of the ccdb repository"};
   Configurable<std::string> ccdbPath{"ccdbPath", "Analysis/PID/TPC/Response", "Path of the TPC parametrization on the CCDB"};
-  Configurable<long> ccdbTimestamp{"ccdb-timestamp", 0, "timestamp of the object used to query in CCDB the detector response. Exceptions: -1 gets the latest object, 0 gets the run dependent timestamp"};
+  Configurable<uint64_t> ccdbTimestamp{"ccdb-timestamp", 0, "timestamp of the object used to query in CCDB the detector response. Exceptions: -1 gets the latest object, 0 gets the run dependent timestamp"};
   // Parameters for loading network from a file / downloading the file
   Configurable<bool> useNetworkCorrection{"useNetworkCorrection", 0, "(bool) Wether or not to use the network correction for the TPC dE/dx signal"};
   Configurable<bool> autofetchNetworks{"autofetchNetworks", 1, "(bool) Automatically fetches networks from CCDB for the correct run number"};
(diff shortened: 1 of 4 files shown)


jgrosseo:
Hi Christian. How do we avoid that this is misused on the Grid?

ChSonnabend:
> Hi Christian. How do we avoid that this is misused on the Grid?

Hi Jan-Fiete,
That's a very good point indeed. I haven't taken that into consideration yet. I think we could add a flag "runOnHyperloop" which could disable the usage of the number of threads automatically (just another condition in the if-statement in the header .cxx file). What do you think about this?
Potentially we wouldn't even need to use this feature at all on hyperloop, I only noticed this behaviour on the GSI SLURM cluster.
Are there maybe already some variables automatically set when an analysis is run on hyperloop? Then I could maybe use these...

jgrosseo:
There are some environment variables present in a Grid job which one could use but at the moment we do not provide them on Hyperloop (could be done of course) so that it looks the same.

I flagged this because we had issues with ONNX Grid jobs at some point because they were pinning to the wrong CPU and then the threads were disabled. But I do not remember the details @saganatt, @mpuccio do you?

jgrosseo:
If we could use the inverse trick for you, we may not even need a configurable...

jgrosseo:
Discussing with Costin, on the Grid there is the variable `ALIEN_JDL_CPUCORES` which should be respected. We can also set it on Hyperloop. I guess if that variable is set, you should ignore the setting of the number of threads. What do you think?

ChSonnabend:
> Discussing with Costin, on the Grid there is the variable `ALIEN_JDL_CPUCORES` which should be respected. We can also set it on Hyperloop. I guess if that variable is set, you should ignore the setting of the number of threads. What do you think?

Indeed, I think this is a good idea. I also feel much better if there is an automatic switch like this rather than the user having to specify it themselves. Just to be sure, this variable is only set on hyperloop right?
I will look up the variable and adjust the code.
Thank you for your help!

jgrosseo:
> > Discussing with Costin, on the Grid there is the variable `ALIEN_JDL_CPUCORES` which should be respected. We can also set it on Hyperloop. I guess if that variable is set, you should ignore the setting of the number of threads. What do you think?
> 
> Indeed, I think this is a good idea. I also feel much better if there is an automatic switch like this rather than the user having to specify it themselves. Just to be sure, this variable is only set on hyperloop right? I will look up the variable and adjust the code. Thank you for your help!

It will be set in the Hyperloop test and for all Grid jobs, also those which go to GSI. Your observation above was I guess with locally submitting to the cluster and not through the Grid, right? In case of your local submissions, the variable is of course not set.

I propose that if that variable is set, you ignore the user configurable and if the user configurable is not the default value print a LOGP(warning) to inform the user that the variable is ignored.

What do you think?

Common/TableProducer/PID/pidTPC.cxx
```diff
@@ -142,9 +148,25 @@ struct tpcPid {
...
+
     if (!useNetworkCorrection) {
       return;
     } else {
+
+      /// Testing hyperloop core settings
+      const char* alien_cores = gSystem->Getenv("ALIEN_JDL_CPUCORES");
+      if (alien_cores != NULL) {
+        std::string msg_cores = "Hyperloop-run detected! Number of cores = ";
+        LOGP(info, msg_cores + std::string(alien_cores));
```
jgrosseo, line 161:
```suggestion
        LOGP(info, "Hyperloop test/Grid job detected! Number of cores = {}. Setting threads anyway to 1.", alien_cores);
```

Common/TableProducer/PID/pidTPCFull.cxx
```diff
@@ -142,9 +148,25 @@ struct tpcPidFull {
...
+
     if (!useNetworkCorrection) {
       return;
     } else {
+
+      /// Testing hyperloop core settings
+      const char* alien_cores = gSystem->Getenv("ALIEN_JDL_CPUCORES");
+      if (alien_cores != NULL) {
+        std::string msg_cores = "Hyperloop-run detected! Number of cores = ";
+        LOGP(info, msg_cores + std::string(alien_cores));
```
jgrosseo, line 161: Same as above if you like it ;)

ChSonnabend: Yes, thank you, will add that ;-)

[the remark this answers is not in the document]
jgrosseo: Makes it easier ;)

jgrosseo (approved):

fcatalan92:
> I flagged this because we had issues with ONNX Grid jobs at some point because they were pinning to the wrong CPU and then the (additional?) threads were disabled. But I do not remember the details @saganatt, @mpuccio do you?

@jgrosseo The issue was for the treelite library and its interaction with jobs running on singularity containers. We do not use this library anymore in O2/O2Physics.
However, also ONNXRuntime has its own thread pool and similar issues could arise. IMHO, a central interface to ONNXRuntime is needed to enforce optimisation and implement eventual fixes in a single place.

fgrosa:
> > I flagged this because we had issues with ONNX Grid jobs at some point because they were pinning to the wrong CPU and then the (additional?) threads were disabled. But I do not remember the details @saganatt, @mpuccio do you?
> 
> @jgrosseo The issue was for the treelite library and its interaction with jobs running on singularity containers. We do not use this library anymore in O2/O2Physics. However, also ONNXRuntime has its own thread pool and similar issues could arise. IMHO, a central interface to ONNXRuntime is needed to enforce optimisation and implement eventual fixes in a single place.

Hi @fcatalan92 @jgrosseo indeed the issue was spotted for the HFFilters (at least, might be not only). The CPU time was 600% without adding
```
https://github.com/AliceO2Group/O2Physics/blob/a8b591be5a0498dd1b443c40242261d88582ca98/EventFiltering/PWGHF/HFFilter.cxx#L234-L235
```
while it is close to 100% after setting them to 1.

jgrosseo (approved):

jgrosseo:
Thanks for your replies. @ChSonnabend and me agreed to have the threads fixed to 1 for now on the Grid and we will discuss the strategy at a WP14 meeting.

PR #1476, opened 2022-11-16
add the some histograms for the TPC and ITS matching study

XiaozhiBai:
HI, 
i added some histograms for the TPC and ITS matching study, and TPC clusters for positive and negative tracks.

Cheers
Xiaozhi

### DIFF
--- a/PWGDQ/Core/HistogramsLibrary.h
+++ b/PWGDQ/Core/HistogramsLibrary.h
@@ -150,6 +150,13 @@ void o2::aod::dqhistograms::DefineHistograms(HistogramManager* hm, const char* h
       hm->AddHistogram(histClass, "IsSPDany", "", false, 2, -0.5, 1.5, VarManager::kIsSPDany);
       hm->AddHistogram(histClass, "IsSPDfirst", "", false, 2, -0.5, 1.5, VarManager::kIsSPDfirst);
       hm->AddHistogram(histClass, "ITSClusterMap", "", false, 128, -0.5, 127.5, VarManager::kITSClusterMap);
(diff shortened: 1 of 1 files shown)


PWGDQ/Core/VarManager.h
```diff
@@ -750,6 +751,9 @@ void VarManager::FillTrack(T const& track, float* values)
         if (fgUsedVars[kTrackDCAresZ]) {
           values[kTrackDCAresZ] = std::sqrt(track.cZZ());
         }
+        if (fgUsedVars[kPt_in]) {
+          values[kPt_in] = values[kPin] / std::sqrt(1.0 + track.tgl() * track.tgl());
+        }
```
iarsene, line 756: Two comments.
1) The kPt_IN is not properly computed here, since tgl does not correspond to the track parameter at the inner TPC wall. Why is this needed?
2) Why is this needed here ? This code block is for the TrackDCA, so the kPt_in variable should be computed below, at L794, where you have it a second time.

PWGDQ/Core/VarManager.h
```diff
@@ -787,6 +791,7 @@ void VarManager::FillTrack(T const& track, float* values)
     values[kTrackCSnpSnp] = track.cSnpSnp();
     values[kTrackCTglTgl] = track.cTglTgl();
     values[kTrackC1Pt21Pt2] = track.c1Pt21Pt2();
+    values[kPt_in] = values[kPin] / std::sqrt(1.0 + track.tgl() * track.tgl());
```
iarsene, line 794: Since this variable require some computation, please add an "if(fgUsedVars[kPt_in])" around it.
But first we need to understand why is this variable useful.

PR #1477, opened 2022-11-16
PWGDQ: Add task for dalitz selection

glegras:
Adding DalitzSelection.cxx for selecting electrons from dalitz decays (running on framework)
Keeping information from the dalitz Selections in tableMaker and tableMakerMC
AnalysisPrefilterSelection in the tableReader to run dalitz selection and prefilter

### DIFF
--- a/PWGDQ/DataModel/ReducedInfoTables.h
+++ b/PWGDQ/DataModel/ReducedInfoTables.h
@@ -477,6 +477,13 @@ DECLARE_SOA_TABLE(V0Bits, "AOD", "V0BITS", //!
 // iterators
 using V0Bit = V0Bits::iterator;
 
+namespace DalBits
+{
+DECLARE_SOA_COLUMN(DALITZBits, dalitzBits, int); //!
+} // namespace DalBits
+
+// bit information for particle species.
+DECLARE_SOA_TABLE(DalitzBits, "AOD", "DALITZBITS", DalBits::DALITZBits);
 } // namespace o2::aod
 
 #endif // O2_Analysis_ReducedInfoTables_H_
--- a/PWGDQ/DataModel/ReducedInfoTables.h
+++ b/PWGDQ/DataModel/ReducedInfoTables.h
@@ -479,7 +479,7 @@ using V0Bit = V0Bits::iterator;
 
 namespace DalBits
 {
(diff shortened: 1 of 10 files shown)


iarsene:
Hi @glegras, you need to rebase and merge your changes again. I sent you some instructions over mattermost.

PWGDQ/DataModel/ReducedInfoTables.h
```diff
@@ -477,6 +477,13 @@ DECLARE_SOA_TABLE(V0Bits, "AOD", "V0BITS", //!
 // iterators
 using V0Bit = V0Bits::iterator;
 
+namespace DalBits
+{
+DECLARE_SOA_COLUMN(DALITZBits, dalitzBits, int); //!
```
iarsene, line 482: The type here should be unsigned. I suggest we limit to uint8_t, similar to the v0Bits table

PWGDQ/TableProducer/tableMaker.cxx
```diff
@@ -349,7 +358,11 @@ struct TableMaker {
             }
           }
         }
+
         trackFilteringTag |= (uint64_t(trackTempFilterMap) << 7); // BIT7-14:  user track filters
+        if constexpr (static_cast<bool>(TTrackFillMap & VarManager::ObjTypes::DalitzBits)) {
+          trackFilteringTag |= (uint64_t(track.dalitzBits()) << 15); // BIT15-...: Dalitz
+        }
```
iarsene, line 365: I suggest that we inject the Dalitz bits right after the V0 bits. The user track filters should be at the end.
We can probably reserve 8 bits just for the Dalitz. The user cuts might be many more after we discuss the cuts in the DQ for the skimming strategy, so they would fill up the space up to 64. You need to do this change also in the tableMakerMC and in the VarManager, where you have to correct for the bit position in the variables you defined.

iarsene (approved):

iarsene (approved):

iarsene:
@ktf  The failing macOS test don't seem to be related to this PR, but since its required now, it prevents the merging.
Could you have a look?

iarsene (approved):

iarsene (approved):

PR #1478, opened 2022-11-16
PWGHF: Fix the cascade skimmer subscription

vkucera:


alibuild (approved):

aalkin:
I'm not sure this will work with duplicated subscription to BCs, did you try it like this?

vkucera:
> 

It doesn't throw any error at least. But definitely better to replace the `BCs` subscription.

alibuild (approved):

PR #1479, opened 2022-11-16
Give all BCs compatible with collisions in output

mpuccio:


mpuccio:
MegaLinter thinks fmt format and rapidjson are c system libraries, and btw, it also suggested to me the change that triggered the failure of clang-format and the last commit fixing it. I think I am going to ignore it in the future.

alibuild (approved):

PR #1489, opened 2022-11-18
Rationalization of the skimming task when no derived data stored

victor-gonzalez:


alibuild (approved):

PR #1492, opened 2022-11-19
TOF: Add first data skimmer for recursive calib.

njacazio:
This adds a derived dataset for TOF


njacazio:
@chiarazampolli @noferini this adds a feature to write derived datasets for the recursive calibration of the TOF response. This is a prototype, do you think we should add also the directory into the codeowner file?

noferini:
> @chiarazampolli @noferini this adds a feature to write derived datasets for the recursive calibration of the TOF response. This is a prototype, do you think we should add also the directory into the codeowner file?

Hi @njacazio, I think that at the moment it is not necessary but the case you want to add other developers which are not owners in the top dirs.

chiarazampolli (approved):

PR #1493, opened 2022-11-19
LF: Catch fitter exceptions

njacazio:
- Add histos to count them
@ddobrigk


alibuild (approved):

PR #1500, opened 2022-11-21
[EventFiltering] Add possibility to apply PID preselection for all species

fgrosa:


alibuild (approved):

PR #1512, opened 2022-11-22
Bug-fix: Neural network, using local file for timestamp -1 and autofetch 0 and CCDB timestamp uint64_t -> int64_t

ChSonnabend:
Local files of neural networks were not usable if timestamp was set to -1 (default) and autofetch to 0. This PR fixes this bug.


jgrosseo (commented):
The changes look OK but can you squash locally? There are so many commits in this PR...

PR #1517, opened 2022-11-23
Quering orbitReset timestamp with EOR timestamp instead of SOR

ekryshen:
Sometimes CTP resets orbit after SOR (but before SOX). So ccdb query with SOR timestamp provides wrong orbitReset reference. It usually affects the first run in the fill. It is more reliable to use EOR timestamp to request orbitReset.

### DIFF
--- a/Common/TableProducer/timestamp.cxx
+++ b/Common/TableProducer/timestamp.cxx
@@ -16,14 +16,13 @@
 /// \brief  A task to fill the timestamp table from run number.
 ///         Uses headers from CCDB
 ///
-
-#include "Framework/runDataProcessing.h"
-#include "Framework/AnalysisTask.h"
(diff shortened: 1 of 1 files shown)


ddobrigk (dismissed):
thanks for the quick fix!!!!

ddobrigk (approved):
Thanks for taking care of this!

jgrosseo:
@ekryshen This seems to brake Run 2 analysis, see https://alice.its.cern.ch/jira/browse/O2-3416
Do you want to revert it or have a better solution?

ekryshen:
@jgrosseo Hm, then we could simply check SOR for Run2 and EOR for Run3.

jgrosseo:
Yes, let's do that. Or (SOR+EOR) / 2 for both cases. Could you make a PR with the solution you prefer?

PR #1538, opened 2022-11-27
Strangeness builder: first pr

ddobrigk:
This is a first commit of a unified `strangeness builder` that does the job of both the lambdakzero and the cascade builders in a more CPU-efficient and modular way. As a consequence, cascade building in particular will become significantly faster, as the number of propagation calls has been much reduced. 

No nested loops are done in this code; instead, std::multimaps and table functionality are employed to reduce CPU usage. Optionally, the strangeness builder also provides V0 and cascade covariance matrices that could be used to initialize TrackParCovs for V0s and cascades when reconstructing decay chains. 

For the time being, this task will coexist with the `lambdakzerobuilder` and `cascadebuilder` for testing and performance comparisons. Once all testing has been concluded, the `strangenessbuilder` task should become de de facto standard.


ddobrigk:
As far as I can tell, the linter errors are completely unrelated to the PR... :-(

njacazio (approved):

PR #1542, opened 2022-11-28
Remove useless configurables (already moved in qaEventTrackLiteProduc…

mfaggin:
…er.cxx).


mfaggin:
@njacazio I remove these configurables.They are not only useless, but also putting some non-sense large value creates non-sense issues in hyperloop tests.

alibuild (approved):

PR #1544, opened 2022-11-28
[Event Filtering] Add protection for not filled score arrays

fgrosa:


alibuild (approved):

PR #1545, opened 2022-11-28
MegaLinter: Make concurrency groups unique per PR

vkucera:
@TimoWilken I suspect that this has been causing cancellation of the action runs due to ongoing checks in other PRs.
Could you please check?

### DIFF
--- a/.github/workflows/mega-linter.yml
+++ b/.github/workflows/mega-linter.yml
@@ -11,7 +11,7 @@ env:
   PR_NUMBER: ${{ github.event.pull_request.number }}
 
 concurrency:
-  group: ${{ github.ref }}-${{ github.workflow }}
+  group: ${{ github.ref }}-${{ github.workflow }}-${{ env.PR_NUMBER }}
   cancel-in-progress: true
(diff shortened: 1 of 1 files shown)


TimoWilken:
Hi @vkucera, nice catch, thank you!

A couple of points:

- [The docs][docs] say you can only use the `github` context, so I don't know if using values via `env.*` is supported.
- As we're always running on `pull_request_target`, `github.ref` will basically always be `master`. You can probably get rid of it.
- I'd consider adding a fallback for the PR number in case we run this workflow on push in the future, for instance.

I'd suggest something like:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}
```

[docs]: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency

TimoWilken:
Cheers @vkucera!

The MegaLinter failure is only due to cpplint errors, thus unrelated; merged.

vkucera:
Thanks

PR #1547, opened 2022-11-28
Developing general framework for the application of ONNX models for analysis

ChSonnabend:
This PR intends to bring:
- A common header file for the application of ONNX models in O2 at analysis time
- Settings for the ONNX models (like CPU-thread setting on a SLURM cluster) with full flexibility for optimization in each task
- Showcasing the application in the tpc-pid tasks


ChSonnabend:
Pinging @fcatalan92 @saganatt
Can someone please add Hannah? I didn't find here GitHub account.

Tools/ML/ONNX/model.cxx
```diff
@@ -0,0 +1,213 @@
...
+  std::stringstream ss("");
+  for (size_t i = 0; i < v.size() - 1; i++)
+    ss << v[i] << "x";
+  ss << v[v.size() - 1];
+  return ss.str();
+}
+
+OnnxModel& OnnxModel::operator=(OnnxModel& inst)
+{
+
```
saganatt, line 40: excessive newline

Tools/ML/ONNX/model.cxx
```diff
@@ -0,0 +1,213 @@
...
+  mOutputNames = inst.mOutputNames;
+  mOutputShapes = inst.mOutputShapes;
+
+  modelPath = inst.modelPath;
+  ccdbUrl = inst.ccdbUrl;
+  activeThreads = inst.activeThreads;
+  valid_from = inst.valid_from;
+  valid_until = inst.valid_until;
+
+  LOG(debug) << "Model copied!";
```
saganatt, line 55: ok, so you assume a copy of an instance doesn't have the same session options?

Tools/ML/ONNX/model.cxx
```diff
@@ -0,0 +1,213 @@
...
+  }
+
+  LOG(info) << "--- Model initialized! ---";
+
+  if (checkHyperloop()) {
+    sessionOptions.SetIntraOpNumThreads(activeThreads);
+  }
+}
+
+bool OnnxModel::fetchFromCCDB(std::string path_from, int64_t ccdbTimestamp, std::string path_to = "model.onnx")
```
saganatt, line 106: You can avoid copying the code if you call `downloadToFile()` inside `fetchFromCCDB()`

Tools/ML/ONNX/model.cxx
```diff
@@ -0,0 +1,213 @@
...
+{
+
+  LOG(info) << "--- ONNX-ML model ---";
+
+  // e.g. path_from = "Analysis/PID/TPC/ML"
+
+  o2::ccdb::CcdbApi ccdbApi;
```
saganatt, line 113: Did you test that it works with this CcdbApi outside an analysis task? Did you already have such a situation with your TPC ONNX before?
I am thinking especially about the case when an analyzer defines themselves a CCDB API in the analysis task -- probably the 2 API instances won't interfere with each other, but I simply don't know for sure.

Tools/ML/ONNX/model.cxx
```diff
@@ -0,0 +1,213 @@
...
+  if (headers.count("Valid-Until") == 0) {
+    LOG(info) << "Valid-Until not found in metadata";
+  } else {
+    LOG(info) << "Timestamp, valid until: " << headers["Valid-Until"].c_str();
+  }
+
+  return retrieve_success;
+}
+
+float* OnnxModel::evalModel(std::vector<Ort::Value> input)
```
saganatt, line 167: Would it be more intuitive for the users to get float instead of a pointer?
Do you have multiple output values (a vector) in your network?

In the future, we can extend it to return any type, for now ok, float is fine.

saganatt: also, what is the use case for calling `evalModel` with an ONNX vector?

saganatt:
Thanks a lot Christian!

I put some detailed comments. In general, this looks good for a first, simplest solution. We can later consider what will be next and next steps for extending.

One general note: I would put all ONNX code in a header file, it could be more efficient that keeping cxx executable.

Tools/ML/ONNX/model.cxx
ChSonnabend: > Would it be more intuitive for the users to get float instead of a pointer? Do you have multiple output values (a vector) in your network?

Indeed, I usually have multiple output values since I call the run function on a vector containing multiple entries (e.g. tracks). The float* returns a float array in this case. For a single return value, a float would also be fine, but for optimization (calling the ONNX Run()-function once instead of N-times) I considered returning a float.

> In the future, we can extend it to return any type, for now ok, float is fine.

That would be great! Maybe a template instance would be best here...

ChSonnabend: > also, what is the use case for calling `evalModel` with an ONNX vector?

For now, it's only a feature if someone wants to make modifications to the tensor before passing it to the model. It's not needed now, but might be helpful in the future.

ChSonnabend:
> Thanks a lot Christian!
> 
> I put some detailed comments. In general, this looks good for a first, simplest solution. We can later consider what will be next and next steps for extending.
> 
> One general note: I would put all ONNX code in a header file, it could be more efficient that keeping cxx executable.

Thank you for the comments, Maja!
I also had it in one header initially, but later on a PR was made suggesting me to move it to a separate .cxx file. For me both is fine, maybe we need to discuss with the experts here about what they would prefer.

Tools/ML/ONNX/model.cxx
ChSonnabend: Technically, the = operator does not exist for two models in ONNX (I think it is explicitly deleted). That's why I created a new one here. Maybe there is some better way of doing it, but for now this should work. Of course this should be extended in the future!

Tools/ML/ONNX/model.cxx
ChSonnabend: Indeed I just tested and it does work 👍

Tools/ML/ONNX/model.cxx
saganatt: If operator= is deleted, then maybe there is a good reason in ONNX. Performance or logic-wise. Where do you need it?

Tools/ML/ONNX/model.cxx
saganatt: Ok, I see your point.
Long time ago I could not get multiple outputs from ONNX. So, it's great it's possible!

Can you point me how do you then push multiple tracks into ONNX? You probably need to manually extract track columns and copy values for ONNX?

saganatt:
> Thank you for the comments, Maja! I also had it in one header initially, but later on a PR was made suggesting me to move it to a separate .cxx file. For me both is fine, maybe we need to discuss with the experts here about what they would prefer.

Ah, I see... We strongly preferred headers in the O2/Framework, maybe O2Physics has other rules...

Tools/ML/ONNX/model.cxx
ChSonnabend: Sure, you can find it e.g. here: https://github.com/AliceO2Group/O2Physics/blob/7234c430df8a1e90b893ef6da7b60edd586d3bd2/Common/TableProducer/PID/pidTPCFull.cxx#L278-L317
One has to be a bit careful then not to mess up the indexing, but otherwise it should work.

saganatt: Oh, so it is a flat array? And onnx itself can make it a 2D tensor or something like that, for the input?

ChSonnabend: Sort of, yeah.
So ONNX (I think) checks the input dimension of the model and then passes as many values from the array into it as possible. So if your model has k inputs and you generate them for N tracks, then your array is k x N long. ONNX then produces a total N outputs for this input. So it automatically applies everything based on its input dimensions.

saganatt: I see, thanks! I see now that you yourself create the 2D tensor inside `evalModel()`, that makes sense now!
For sure, after the first PRs we will need to test this on more cases (what if `mInputShapes[0][1]==0`?), for now it is fine. I am getting a bit spammy, using this GitHub conversation to save any potential extension ideas for the future ;-)

Common/TableProducer/PID/pidTPCFull.cxx
```diff
@@ -199,11 +178,14 @@ struct tpcPidFull {
...
           }
           LOG(info) << "Using local file [" << networkPathLocally.value << "] for the TPC PID response correction.";
-          Network temp_net(networkPathLocally.value,
-                           enableNetworkOptimizations.value,
-                           activeThreads);
+          OnnxModel temp_net(networkPathLocally.value);
           network = temp_net;
-          network.evalNetwork(std::vector<float>(network.getInputDimensions(), 1.)); // This is an initialisation and might reduce the overhead of the model
+          if (enableNetworkOptimizations) {
+            Ort::SessionOptions* sessOpts = network.getSessionOptions();
```
saganatt, line 184: another point for the future - maybe we can wrap this optimization part as well into one-liner for the users.

saganatt:
Thanks again, Christian! After the next reading of the code, I don't see anything problematic for this first PR. Just fix the MegaLinter complaint.
We can later, for other PRs, think about the points mentioned in the comments.
Let's see also @fcatalan92 and Hannah's opinions.

ChSonnabend:
> Thanks again, Christian! After the next reading of the code, I don't see anything problematic for this first PR. Just fix the MegaLinter complaint. We can later, for other PRs, think about the points mentioned in the comments. Let's see also @fcatalan92 and Hannah's opinions.

Thanks for all the comments! Indeed, let's see what Fabio and @hjbossi think and then it would be fine for me to merge it.

Common/TableProducer/PID/pidTPCFull.cxx
ChSonnabend: That could be done easily, yes. I also had it in my header as a boolean in the constructor. I just thought that for full flexibility with all the options ONNX provides (thread / CPU / GPU settings or model optimizations), it is easiest if one can modify the sessionOptions and then reset the sessions with the optimizations without being limited by a function. But we can surely also put it in a function to make it a bit easier for the users.

saganatt: ok, yeah, that is open for discussion. There can be a wrapper with an optional argument to select optimization type, for example. That could balance flexibility and simplicity.

Tools/ML/ONNX/model.h
```diff
@@ -0,0 +1,100 @@
...
+ public:
+  OnnxModel();
+  explicit OnnxModel(std::string);
+  ~OnnxModel() = default;
+  OnnxModel& operator=(OnnxModel&);
+
+  // IO
+  std::map<std::string, std::string> fetchHeaders(std::string, int64_t, bool);
+  bool downloadToFile(std::string, int64_t, std::string, bool);
+  bool fetchFromCCDB(std::string, int64_t, std::string, bool);
```
fcatalan92, line 55: Do these functions need to be exposed to the user? In general, I would keep the interface as simple as possible.

Common/TableProducer/PID/pidTPC.cxx
```diff
@@ -163,33 +164,23 @@ struct tpcPid {
...
-          bool retrieve_success = ccdbApi.retrieveBlob(networkPathCCDB.value, ".", metadata, ccdbTimestamp.value, false, networkPathLocally.value);
-          if (retrieve_success) {
-            std::map<std::string, std::string> headers = ccdbApi.retrieveHeaders(networkPathCCDB.value, metadata, ccdbTimestamp.value);
-            if (headers.count("Valid-From") == 0) {
-              LOG(fatal) << "Valid-From not found in metadata";
+          bool fetchSuccess = network.fetchFromCCDB(networkPathCCDB, ccdbTimestamp.value, networkPathLocally, true);
+          if (fetchSuccess) {
+            if (enableNetworkOptimizations) {
+              Ort::SessionOptions* sessOpts = network.getSessionOptions();
+              sessOpts->SetGraphOptimizationLevel(GraphOptimizationLevel::ORT_ENABLE_EXTENDED);
```
fcatalan92, line 180: I would add a setter in the new class and avoid to expose the pointer

fcatalan92 (commented):
Hi @ChSonnabend, thank you a lot for this effort (and @saganatt too). It already looks quite good. I added some comments to the code and I have a few more general ones too:

- try to keep the interface as simple as possible. I would try to have a few steps/functions to load/create the model and perform the inference, hiding the complexity. For example you could try to have all the information for the model initialisation set in the constructor, then an init function to load and validate the model, and finally the inference;
- have a single way to load a local model or a CCDB one inferring what to do from the path or maybe a flag, and perform the checks on the validity of the model already in the class;
- generalise the inference to work for matrix input (this can be done at a later stage).

Tools/ML/ONNX/model.cxx
fcatalan92: Indeed, I don't see good reasons for copying the model. I would remove the = operator from the class.

Common/TableProducer/PID/pidTPC.cxx
ChSonnabend: I did this explicitly to have the full flexibility of the ONNX options available for the user. But maybe you have an idea how this could be done with a function/class which still maintains the flexibility?

Tools/ML/ONNX/model.h
ChSonnabend: Not necessarily, but sometimes a user might want to download a model without loading it into the class immediately, that's why I added these two functions. But it depends on what the goal for the class would be.

Tools/ML/ONNX/model.cxx
```diff
@@ -0,0 +1,225 @@
...
+    LOGP(info, "Not running on Hyperloop.");
+  }
+
+  return alien_cores_found;
+}
+
+OnnxModel::OnnxModel()
+{
+  ccdbApi.init(ccdbUrl);
+}
```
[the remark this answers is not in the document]
ChSonnabend: The point is that the ccdb should always be initialised (for fetching and loading the models) which cannot be guaranteed with the fetch and download functions (or we have more, redundant initialisations later in the code).

Tools/ML/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
...
+# All rights not expressly granted are reserved.
+#
+# This software is distributed under the terms of the GNU General Public
+# License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+#
+# In applying this license CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+add_subdirectory(ONNX)
```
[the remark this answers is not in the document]
ChSonnabend: Sounds good, I just thought about putting it in a subdirectory because in the future we might also want to investigate TMVA, but I could also just rename the files with e.g. model_ONNX.h

Tools/ML/ONNX/model.cxx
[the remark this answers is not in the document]
fcatalan92: This is a good reason to have an `Init` function to initialise the ccdb, fetch and load the model, and validate it, as I was mentioning in the more general comment. It is also worth thinking if having an `o2::ccdb::CcdbApi` object for each model is fine.

Tools/ML/ONNX/model.h
fcatalan92: I would try to enforce a one-to-one relation between an ML model and an instance of this class. In which cases do you think it would be helpful to separate the fetching of the model from creating the class object?

Common/TableProducer/PID/pidTPC.cxx
fcatalan92: Maybe we could expose some setters to enable specific options (btw this can be done at a later stage), not sure that having full flexibility is a good idea.

fcatalan92 (dismissed):

hjbossi:
Hi Christian, 

Thanks for all of the work on this! I agree with Fabio that it looks good to go! One small comment is that for organizational purposes, it would be good to squash the commits into a single commit if possible! Other than that looks good to go! 

Cheers, 

Hannah

ddobrigk (approved):
approved after in-person discussion. Thanks!

saganatt:
Hi,
somehow re-building O2Physics fails for me after this update.

I get an error:
```
/home/mkabus/alice/sw/SOURCES/O2Physics/qa-all/0/Tools/ML/model.cxx: In member function 'float* o2::ml::OnnxModel::evalModel(std::vector<Ort::Value>)':
/home/mkabus/alice/sw/SOURCES/O2Physics/qa-all/0/Tools/ML/model.cxx:107:60: error: no match for 'operator==' (operand types are 'std::vector<long int>'
 and 'int')
  107 |     assert(input[0].GetTensorTypeAndShapeInfo().GetShape() == getInputDimensions());
      |            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^~ ~~~~~~~~~~~~~~~~~~~~
      |                                                         |                       |
      |                                                         std::vector<long int>   int
```

I updated today alidist, O2, O2Physics. It is strange given that your PR build checks passed...

PR #1550, opened 2022-11-29
PWGCF: Add column to FemtoDreamCollision table for charge particle multiplicity

ariedel-cern:
Use multNtrPV to estimate the charge particle multiplicity of an event and add it as a column to FemtoDreamCollision table.

The V0M signal, which has been used so far, might not be needed in the future, but it will be kept for now.


ariedel-cern:
adding @lauraser

victor-gonzalez (approved):

PR #1556, opened 2022-11-30
Task for analysis with ZDC in UE-PAG

coppedis:
This is the first task to analyze ZDC data and ZDC correlations with less forward detectors signals in Pb-Pb collisions


PWGMM/UE/Tasks/ue-zdc-analysys.cxx
```diff
@@ -0,0 +1,294 @@
...
+
+#include "TH1F.h"
+#include "TH2F.h"
+#include "TObjArray.h"
+
+using namespace o2;
+using namespace o2::framework;
+using namespace o2::framework::expressions;
+
+void customize(std::vector<o2::framework::ConfigParamSpec>& workflowOptions)
```
aalkin, line 34: Please note, that these options cannot be used on hyperloop. The new approach is to use multiple process functions with switches to enable/disable them, that can be used on hyperloop.

coppedis: I am not sure I fully get you point, Anton. Which options are you referring to? Those in defineDataProcessing?
Thanks a lot for this input (there are several things I am not aware of, evidently)

aalkin: Yes, the customize function introduces options that can only be provided via command line, they are not supported on hyperloop, which means you would not be able to configure your task there. Instead of having several tasks that are controlled by these options, you can instead have a single task with multiple process functions that are controlled by the process switches defined in the task itself. These switches can be changed through configuration JSON, which is how the tasks are configured on hyperloop. 

We have an example for this: https://github.com/AliceO2Group/O2Physics/blob/master/Tutorials/src/multiProcess.cxx

coppedis: @aalkin 
I updated my task following your suggestions and I tested it locally.

aalkin:
Thank you, now only some formatting issues remain. Note that for clang-format you can simply merge the PR in your own fork of O2Physics on github that was automatically created after failed check, it will apply necessary changes.

aalkin:
There is still a small issue from Megalinter
`/github/workspace/PWGMM/UE/Tasks/ue-zdc-analysys.cxx:226:  Could not find a newline character at the end of the file.  `

The rest are spurious and can be ignored.

aalkin (approved):

PR #1558, opened 2022-11-30
PWGDQ: Add prefilter in tableReader

glegras:


iarsene:
Hi @glegras, 
Could you have a look in the failed tests? I still have to look in the PR carefully.

ctolon:
If possible, I have a small request. Is it possible to use the naming convention for same event pairing as processDecayToEEPrefilterSkimmed instead of processDecayToEESkimmedPrefilter?

I can handle this situation much easier as I use Skimmed as suffix for same event pairing in python scripts. Of course, even if it stays like this, I will try to handle it somehow. Thanks for your attention.

glegras:
@ctolon Sure, I can do it

ctolon:
@glegras Thanks a lot!

PWGDQ/Tasks/tableReader.cxx
```diff
@@ -54,12 +54,15 @@ DECLARE_SOA_COLUMN(MixingHash, mixingHash, int);
 DECLARE_SOA_COLUMN(IsEventSelected, isEventSelected, int);
 DECLARE_SOA_COLUMN(IsBarrelSelected, isBarrelSelected, int);
 DECLARE_SOA_COLUMN(IsMuonSelected, isMuonSelected, int);
+DECLARE_SOA_COLUMN(IsBarrelSelectedDalitz, isBarrelSelectedDalitz, int);
+DECLARE_SOA_COLUMN(IsDalitzLeg, isDalitzLeg, int);
```
iarsene, line 58: This is a general comment. We need to remove Dalitz from the naming. This is about prefiltering, which is mainly about removing conversions. The prefiltering is also the usual name which was used in DQ since the very beginning. Please remove Dalitz from names everywhere.

PWGDQ/Tasks/tableReader.cxx
```diff
@@ -351,6 +363,196 @@ struct AnalysisMuonSelection {
...
+  Configurable<std::string> fConfigDalitzPairCuts{"cfgDalitzPairCuts", "", "Dalitz pair selection cuts"};
+  Configurable<std::string> fConfigAddTrackPairHistogram{"cfgAddTrackPairHistogram", "", "Comma separated list of histograms"};
+  Configurable<bool> fQA{"cfgQA", false, "QA histograms"};
+
+  Filter filterBarrelTrackSelected = aod::dqanalysisflags::isBarrelSelectedDalitz > 0; // Dalitz candidates
+
+  OutputObj<THashList> fOutputList{"output"}; //! the histogram manager output list
+  OutputObj<TList> fStatsList{"Statistics"};  //! skimming statistics
+
+  std::map<int, uint8_t> dalitzmap;
```
iarsene, line 381: We should respect some conventions for the naming of variables, for easier code readability. Usually, the class data members start with an "f" followed by a capital. So in this case, "dalitzmap" -> "fDalitzMap". But find a better name, since Dalitz should be removed.

PWGDQ/Tasks/tableReader.cxx
```diff
@@ -351,6 +363,196 @@ struct AnalysisMuonSelection {
...
+
+  std::vector<int> fTrackCutsId;
+  std::vector<AnalysisCompositeCut> fPairCuts;
+  int nCuts = 0;
```
iarsene, line 385: nCuts -> fNCuts

PWGDQ/Tasks/tableReader.cxx
```diff
@@ -351,6 +363,196 @@ struct AnalysisMuonSelection {
...
+    if (!cutNamesPairStr.IsNull()) {
+      std::unique_ptr<TObjArray> objArray(cutNamesPairStr.Tokenize(","));
+      for (int icut = 0; icut < objArray->GetEntries(); ++icut) {
+        fPairCuts.push_back(*dqcuts::GetCompositeCut(objArray->At(icut)->GetName()));
+      }
+    }
+
+    if (fTrackCutsId.size() != fPairCuts.size()) {
+      std::cout << "WARNING: YOU SHOULD PROVIDE THE SAME NUMBER OF TRACK AND PAIR CUTS" << std::endl;
+    }
```
iarsene, line 411: Here we should create a Fatal actually. The warning would just go unnoticed.

PWGDQ/Tasks/tableReader.cxx
```diff
@@ -351,6 +363,196 @@ struct AnalysisMuonSelection {
...
+    nCuts = std::min(fTrackCutsId.size(), fPairCuts.size());
+
+    VarManager::SetUseVars(AnalysisCut::fgUsedVars); // provide the list of required variables so that VarManager knows what to fill
+    VarManager::SetDefaultVarNames();
+    fHistMan = new HistogramManager("analysisHistos", "aa", VarManager::kNVars);
+    fHistMan->SetUseDefaultVariableNames(kTRUE);
+    fHistMan->SetDefaultVarNames(VarManager::fgVariableNames, VarManager::fgVariableUnits);
```
iarsene, line 418: if the fQA configurable is false, you don't need to initialize the var manager anymore. The same with the other blocks below. If no QA is needed, we should not create histograms or output directories.

PWGDQ/Tasks/tableReader.cxx
```diff
@@ -351,6 +363,196 @@ struct AnalysisMuonSelection {
...
+    // Dalitz selection statistics: one bin for each (track,pair) selection
+    TH1I* histTracks = new TH1I("TrackStats", "Dalitz selection statistics", nCuts, -0.5, nCuts - 0.5);
+    for (int icut = 0; icut < nCuts; icut++) {
+      int trackCutId = fTrackCutsId.at(icut);
+      AnalysisCompositeCut pairCut = fPairCuts.at(icut);
+      histTracks->GetXaxis()->SetBinLabel(icut + 1, Form("track%d_%s", trackCutId, pairCut.GetName()));
+    }
+    if (fQA) {
+      fStatsList->Add(histTracks);
+    }
```
iarsene, line 446: I would say this is not needed. You have already filled histograms for prefiltered tracks and pairs build in the prefiltering. is there any new information here beyond what I could extract by looking in the "TrackBarrel" classes above and check the number of entries? My suggestion is to remove this from the task.

PWGDQ/Tasks/tableReader.cxx
```diff
@@ -351,6 +363,196 @@ struct AnalysisMuonSelection {
...
+    VarManager::SetUseVars(fHistMan->GetUsedVars()); // provide the list of required variables so that VarManager knows what to fill
+    fOutputList.setObject(fHistMan->GetMainHistogramList());
+  }
+
+  template <uint32_t TTrackFillMap, typename TTracks>
+  void runDalitzPairing(TTracks const& tracks1, TTracks const& tracks2)
+  {
+    const int TPairType = VarManager::kDecayToEE;
```
iarsene, line 454: This is harcoded here. Eventually one would want to run this over different pair types. So either use the VarManager::kDecayToEE directly at the place it is needed, or make the TPairType a template argument

PWGDQ/Tasks/tableReader.cxx
```diff
@@ -351,6 +363,196 @@ struct AnalysisMuonSelection {
...
+    } // end of tracksP,N loop
+  }
+
+  void processBarrelSkimmed(MyEventsSelected const& events, soa::Filtered<MyBarrelTracksSelected> const& filteredTracks, MyBarrelTracks const& tracks)
+  {
+    dalitzmap.clear();
+    for (auto& event : events) {
+      if (event.isEventSelected()) {
+        auto groupedFilteredTracks = filteredTracks.sliceBy(perCollision, event.globalIndex());
+        runDalitzPairing<gkTrackFillMap>(groupedFilteredTracks, groupedFilteredTracks);
```
iarsene, line 539: Here you run a double loop using the tracks from the prefilter group. This is not the way we should do this. The pairing should be done between the "good" tracks and the tracks from the prefiltering group, which are usually selected with more open cuts to increase the probability of finding a conversion. 
Also, the prefiltering should have an option of running over same event pairs. This is not typical, however, sometimes we can diagnose various problems in analysis, such as track duplicates, etc.

PWGDQ/Tasks/tableReader.cxx
```diff
@@ -234,25 +239,32 @@ struct AnalysisTrackSelection {
...
-
       iCut = 0;
       for (auto cut = fTrackCuts.begin(); cut != fTrackCuts.end(); cut++, iCut++) {
         if ((*cut).IsSelected(VarManager::fgValues)) {
-          filterMap |= (uint32_t(1) << iCut);
+          if (iCut != fConfigPrefilterCutId) {
+            filterMap |= (uint32_t(1) << iCut);
+          }
+          if (iCut == fConfigPrefilterCutId) {
+            filterMapPrefilter |= (uint8_t(1) << (iCut - fConfigPrefilterCutId));
```
iarsene, line 259: I think this command is equivalent to just: filterMapPrefilter = uint_8(1);

PWGDQ/Tasks/tableReader.cxx
```diff
@@ -613,11 +694,13 @@ struct AnalysisSameEventPairing {
   Configurable<string> ccdbPath{"ccdb-path", "Users/lm", "base path to the ccdb object"};
   Configurable<int64_t> nolaterthan{"ccdb-no-later-than", std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch()).count(), "latest acceptable timestamp of creation for the object"};
   Configurable<std::string> fConfigAddSEPHistogram{"cfgAddSEPHistogram", "", "Comma separated list of histograms"};
+
   Service<o2::ccdb::BasicCCDBManager> ccdb;
   Filter filterEventSelected = aod::dqanalysisflags::isEventSelected == 1;
   // NOTE: the barrel filter map contains decisions for both electrons and hadrons used in the correlation task
   Filter filterBarrelTrackSelected = aod::dqanalysisflags::isBarrelSelected > 0;
   Filter filterMuonTrackSelected = aod::dqanalysisflags::isMuonSelected > 0;
+  Filter prefilter = aod::dqanalysisflags::isPrefiltered == 0;
```
iarsene, line 703: Does this work for the process functions where the prefilter is not enabled?

PR #1561, opened 2022-11-30
Implement bitmap for CEFP output

strogolo:
Updates in the CEFP:
- add CefpSelected bitmap to the CEFP output table to keep track of selected triggers
- add CefpTrigger bitmap to the CEFP output table to keep track of fired triggers
- add globalBCId to the output table 
- modify the output table for future filtering per BCs

### DIFF
--- a/EventFiltering/cefpTask.cxx
+++ b/EventFiltering/cefpTask.cxx
@@ -355,9 +355,9 @@ struct centralEventFilterTask {
       }
     }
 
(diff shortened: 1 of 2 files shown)


strogolo:
@mpuccio here is the PR we discussed. Could you please check it carefully? Especially the part where we access the BC column.
Thanks and regards

mpuccio (approved):

PR #1564, opened 2022-12-01
PWGHF: B+ Candidate Creator update

apalasciano:
- Process Switch introduced
- Load Magnetic Field from CCDB
- Mass selection window in B+ creator (B+ task: histogram range changed accordingly)


vkucera (changes requested):
Hi @apalasciano , can you please group and sort the `struct` members as precised in the HF coding conventions?

vkucera (approved):

PR #1570, opened 2022-12-02
PWG-LF: update pid tables to be LF specific

njacazio:


alibuild (approved):

PR #1575, opened 2022-12-03
First-generation topological selection autodetect

ddobrigk:
- separation into single and multi-strange builders
- general improvements in configurability
- more to follow


njacazio (approved):
Thanks David

PR #1585, opened 2022-12-05
PWGLF: Decouple Run2 from Run3 spectra task

njacazio:
- Add run2 spectra task
- Add header for spectra task


alibuild (approved):

PR #1589, opened 2022-12-05
PWGHF: Adding final D0 reco histograms

DelloStritto:


vkucera (changes requested):
Hi @DelloStritto , I think it is fine to fill the `Pt` and `PtVsY` histograms in the same block.

DelloStritto:
> 

You are totally right @vkucera. I modified the task following your suggestion. Thanks!

vkucera (approved):

PR #1591, opened 2022-12-05
PWGLF: Minor fix in NucleiTask

giovannimalfattore:


njacazio (approved):
Indeed, good catch

njacazio:
I think this will need this fix https://github.com/AliceO2Group/O2Physics/pull/1592

PR #1593, opened 2022-12-05
PWGHF: Add Omegac0 code

ZFederica:


vkucera (changes requested):
Hi @ZFederica , thanks for this addition. There a several things to fixed. Please check the HF coding and naming conventions.
Also use `PROCESS_SWITCH` instead of the command line `doMC` option (which would not be usable on AliHyperloop ).

ZFederica:
Thanks @vkucera for your reply. I'll change to PROCESS_SWITCH. As for the HF conventions, where can I find some indications?

PR #1597, opened 2022-12-06
PWGEM/PhotonMeson: Add Background calculation to pi0QC task

jokonig:
- Add mixed events and rotation background calculation
- Fix loop over collisions instead of whole timeframes
- Added opening angle cut


jokonig:
@mikesas @m-c-danisch could you approve this PR?

mikesas (approved):

PR #1605, opened 2022-12-07
DPG: TPC Skimming and Skimming Light

AnnalenaKa:


AnnalenaKa:
Hi @chiarazampolli and @noferini, 
I have made some changes to the skimming task we need for the TPC to provide the initial Bethe Bloch parameters and the neural network correction. The light version of the skimming task is needed to do a first selection of tracks and it produces a tree to obtain the Bethe Bloch parameters and the larger version of the task is run at a later stage and uses reconstructed and selected V0s to train the neural network. 
Could you approve the PR if you are happy with the changes I have made? 
Thank you :)

noferini:
Dear @AnnalenaKa ,
there is an error in o2 check but I didn't undestand why it is complaining.
Do your change compile in your local test?

A second question: do you need to trace all those commits or can you squash some of them into a single one?
Cheers,
Francesco

AnnalenaKa:
Hi @noferini,

everything was compiling well locally, but then the MegaLinter complained about the order of my header files and I changed those. This introduced my building error. It is now solved and the tests should be fine. Sorry about that. I did not expect that. We also do not need the single commits. I think there should be a button you can click when merging to squash all of them into one commit. That is also absolutely fine for us. 

Cheers,
Annalena

noferini (approved):
It is fine.
I'll merge once the last test is passed

PR #1610, opened 2022-12-07
added QA histos for EMC bc and col matching

fjonasALICE:


mfasDa:
Since EMCAL is a triggered detector for which the detector information is related to the BC ID we need to monitor how many BCs are uniquely matched to a vertex, for how many BCs we get in-bunch pileup which we are unable to separate, and for how many BCs we fail to find a vertex.

PR #1611, opened 2022-12-07
MegaLinter: Checkout the merge commit instead of PR HEAD

vkucera:
@TimoWilken This should avoid including unrelated files in the list of files checked by MegaLinter when the differences w.r.t. upstream are due to the PR branch not being up to date.


TimoWilken:
Thanks @vkucera !

PR #1629, opened 2022-12-10
Update hypertriton QA for tests

ddobrigk:
* adds simple 2d QA plots for relevant variables
* Removes unnecessary cov matrix subscription for reduction of memory footprint


njacazio (approved):

mpuccio:
Hi @njacazio @ddobrigk, do you need help with the tests on the hypertriton? A student is already working on it: he fixed the SVertexer, had the first O2Physics code for that, and presented it at the PAG... it is a bit strange that we do not discuss these things in the PAG.

ddobrigk:
Hi @mpuccio, I assumed that the lack of code in O2Physics meant that there was nobody who tested - sorry for that assumption. It is in fact also a bit weird if there were results also not in the main code repository... And about presenting, sure, I am in touch with the PAG conveners and will in fact [present this Monday ](https://indico.cern.ch/event/1225943/). I would be happy to work together!

mpuccio:
The code was not there for the issue we got with the magnetic field, for which we sent you an email back in May.
Only recently, this was fixed in the strangeness builder, and we wanted to add the code.
OK, I will then talk to the conveners, it seems strange that they knew about both the developments and we did not get in the loop.

PR #1637, opened 2022-12-12
PWGCF: Update Femtodream framework

ariedel-cern:
- add more capabilities for processing Monte Carlo data
- update/fix the reduced FemtoDreamProcuder task and cutculator for Pair Track-Track analysis
- fix formating and linting


PWGCF/DataModel/FemtoDerived.h
```diff
@@ -185,25 +185,36 @@ namespace femtodreamparticleMC
...
+  kMaterial,   //! Particle from a material
+  kNotPrimary, //! Particle from a material
+  kFake,
+  kDaughterLambda,
+  kDaughterSigmaplus
 };
 
-DECLARE_SOA_INDEX_COLUMN(FemtoDreamParticle, femtoDreamParticle);
+// DECLARE_SOA_INDEX_COLUMN(FemtoDreamCollision, femtoDreamCollision);
+// DECLARE_SOA_INDEX_COLUMN(FemtoDreamParticle, femtoDreamParticle);
```
victor-gonzalez, line 198: Please, do not leave code commented

PWGCF/DataModel/FemtoDerived.h
```diff
@@ -185,25 +185,36 @@ namespace femtodreamparticleMC
...
 // debug variables
 DECLARE_SOA_COLUMN(MotherPDG, motherPDG, int); //! Checks mother PDG, where mother is the primary particle for that decay chain
 } // namespace femtodreamparticleMC
 
 DECLARE_SOA_TABLE(FemtoDreamParticlesMC, "AOD", "FEMTODREAMPSMC",
-                  o2::soa::Index<>,
-                  femtodreamparticleMC::FemtoDreamParticleId,
+                  // o2::soa::Index<>,
+                  // femtodreamparticleMC::FemtoDreamParticleId,
+                  // femtodreamparticleMC::FemtoDreamCollisionId,
```
victor-gonzalez, line 212: Please do not leave code commented

PWGCF/FemtoDream/FemtoDreamContainer.h
```diff
@@ -126,15 +131,84 @@ class FemtoDreamContainer
...
+    framework::AxisSpec femtoObsAxis = {kstarBins, femtoObs.c_str()};
+    framework::AxisSpec kTAxis = {kTBins, "#it{k}_{T} (GeV/#it{c})"};
+    framework::AxisSpec mTAxis = {mTBins, "#it{m}_{T} (GeV/#it{c}^{2})"};
+    std::string folderName = static_cast<std::string>(mFolderSuffix[mEventType]);
+    mHistogramRegistry->add((folderName + "relPairDist_truth").c_str(), ("; " + femtoObs + "; Entries").c_str(), kTH1F, {femtoObsAxis});
+    /*
+    mHistogramRegistry->add((folderName + "relPairkT_truth").c_str(), "; #it{k}_{T} (GeV/#it{c}); Entries", kTH1F, {kTAxis});
+    mHistogramRegistry->add((folderName + "relPairkstarkT_truth").c_str(), ("; " + femtoObs + "; #it{k}_{T} (GeV/#it{c})").c_str(), kTH2F, {femtoObsAxis, kTAxis});
+    mHistogramRegistry->add((folderName + "relPairkstarmT_truth").c_str(), ("; " + femtoObs + "; #it{m}_{T} (GeV/#it{c}^{2})").c_str(), kTH2F, {femtoObsAxis, mTAxis});
+    */
```
victor-gonzalez, line 153: Please do not leave code commented

PWGCF/FemtoDream/FemtoDreamContainer.h
```diff
@@ -126,15 +131,84 @@ class FemtoDreamContainer
...
+    }
+    // const float kT = FemtoDreamMath::getkTMC(part1, mMassOne, part2, mMassTwo);
+    // const float mT = FemtoDreamMath::getmTMC(part1, mMassOne, part2, mMassTwo);
+    if (mHistogramRegistry) {
+      mHistogramRegistry->fill(HIST(mFolderSuffix[mEventType]) + HIST("relPairDist_truth"), femtoObsMC);
+      /*
+      mHistogramRegistry->fill(HIST(mFolderSuffix[mEventType]) + HIST("relPairkT_truth"), kT);
+      mHistogramRegistry->fill(HIST(mFolderSuffix[mEventType]) + HIST("relPairkstarkT_truth"), femtoObsMC, kT);
+      mHistogramRegistry->fill(HIST(mFolderSuffix[mEventType]) + HIST("relPairkstarmT_truth"), femtoObsMC, mT);
+      */
```
victor-gonzalez, line 188: Please do not leave code commented

PWGCF/FemtoDream/FemtoDreamContainer.h
```diff
@@ -126,15 +131,84 @@ class FemtoDreamContainer
...
+  template <typename T>
+  void setPairMC(T const& part1, T const& part2, const int mult)
+  {
+    float femtoObs, femtoObsMC;
+    if constexpr (mFemtoObs == femtoDreamContainer::Observable::kstar) {
+      femtoObs = FemtoDreamMath::getkstar(part1, mMassOne, part2, mMassTwo);
+      femtoObsMC = FemtoDreamMath::getkstarMC(part1, mMassOne, part2, mMassTwo);
+    }
+    // const float kT = FemtoDreamMath::getkTMC(part1, mMassOne, part2, mMassTwo);
+    // const float mT = FemtoDreamMath::getmTMC(part1, mMassOne, part2, mMassTwo);
```
victor-gonzalez, line 181: Please do not leave code commented

PWGCF/FemtoDream/FemtoDreamContainer.h
```diff
@@ -126,15 +131,84 @@ class FemtoDreamContainer
...
+      mHistogramRegistry->fill(HIST(mFolderSuffix[mEventType]) + HIST("relPairkstarmT_truth"), femtoObsMC, mT);
+      */
+      mHistogramRegistry->fill(HIST(mFolderSuffix[mEventType]) + HIST("relPairkstarMult_truth"), femtoObsMC, mult);
+      /*
+      mHistogramRegistry->fill(HIST(mFolderSuffix[mEventType]) + HIST("kstarPtPart1"), femtoObsMC, part1.ptTruth());
+      mHistogramRegistry->fill(HIST(mFolderSuffix[mEventType]) + HIST("kstarPtPart2"), femtoObsMC, part2.ptTruth());
+      mHistogramRegistry->fill(HIST(mFolderSuffix[mEventType]) + HIST("MultPtPart1"), part1.ptTruth(), mult);
+      mHistogramRegistry->fill(HIST(mFolderSuffix[mEventType]) + HIST("MultPtPart2"), part2.ptTruth(), mult);
+      mHistogramRegistry->fill(HIST(mFolderSuffix[mEventType]) + HIST("PtPart1PtPart2"), part1.pt(), part2.pt());
+      */
```
victor-gonzalez, line 196: Please do not leave code commented

PWGCF/FemtoDream/FemtoDreamCutculator.h
```diff
@@ -47,7 +47,16 @@ class FemtoDreamCutculator
...
       LOG(fatal) << "Failed to read JSON config file " << configFile << " (" << e.what() << ")";
     }
-    mConfigTree = root.get_child("femto-dream-producer-task");
+
+    // check the config file for all known producer task
+    std::vector<const char*> ProducerTasks = {"femto-dream-producer-task", "femto-dream-producer-reduced-task"};
+    for (auto& Producer : ProducerTasks) {
+      if (root.count(Producer) > 0) {
+        mConfigTree = root.get_child(Producer);
+        std::cout << "Found " << Producer << " in " << configFile << std::endl;
```
victor-gonzalez, line 56: Please do not use `cout`, use instead `LOGF`

PWGCF/FemtoDream/FemtoDreamMath.h
```diff
@@ -135,8 +135,58 @@ class FemtoDreamMath
...
+  /// \tparam T type of tracks
+  /// \param part1 Particle 1
+  /// \param mass1 Mass of particle 1
+  /// \param part2 Particle 2
+  /// \param mass2 Mass of particle 2
+  template <typename T>
+  static float getmTMC(const T& part1, const float mass1, const T& part2, const float mass2)
+  {
+    return std::sqrt(std::pow(getkTMC(part1, mass1, part2, mass2), 2.) + std::pow(0.5 * (mass1 + mass2), 2.));
+  }
```
victor-gonzalez, line 187: This is what I meant
You are replicating code which does not need to

PWGCF/DataModel/FemtoDerived.h
ariedel-cern: fixed.

PWGCF/DataModel/FemtoDerived.h
ariedel-cern: fixed.

PWGCF/FemtoDream/FemtoDreamContainer.h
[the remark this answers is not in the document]
ariedel-cern: fixed.

PWGCF/FemtoDream/FemtoDreamContainer.h
```diff
@@ -126,15 +131,84 @@ class FemtoDreamContainer
...
+    */
+    mHistogramRegistry->add((folderName + "relPairkstarMult_truth").c_str(), ("; " + femtoObs + "; Multiplicity").c_str(), kTH2F, {femtoObsAxis, multAxis});
+    /*
+    mHistogramRegistry->add((folderName + "kstarPtPart1_truth").c_str(), ("; " + femtoObs + "; #it{p} _{T} Particle 1 (GeV/#it{c})").c_str(), kTH2F, {femtoObsAxis, {375, 0., 7.5}});
+    mHistogramRegistry->add((folderName + "kstarPtPart2_truth").c_str(), ("; " + femtoObs + "; #it{p} _{T} Particle 2 (GeV/#it{c})").c_str(), kTH2F, {femtoObsAxis, {375, 0., 7.5}});
+    mHistogramRegistry->add((folderName + "MultPtPart1_truth").c_str(), "; #it{p} _{T} Particle 1 (GeV/#it{c}); Multiplicity", kTH2F, {{375, 0., 7.5}, multAxis});
+    mHistogramRegistry->add((folderName + "MultPtPart2_truth").c_str(), "; #it{p} _{T} Particle 2 (GeV/#it{c}); Multiplicity", kTH2F, {{375, 0., 7.5}, multAxis});
+    mHistogramRegistry->add((folderName + "PtPart1PtPart2_truth").c_str(), "; #it{p} _{T} Particle 1 (GeV/#it{c}); #it{p} _{T} Particle 2 (GeV/#it{c})", kTH2F, {{375, 0., 7.5}, {375, 0., 7.5}});
+    */
```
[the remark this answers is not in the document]
ariedel-cern: fixed.

PWGCF/FemtoDream/FemtoDreamContainer.h
ariedel-cern: fixed.

PWGCF/FemtoDream/FemtoDreamContainer.h
ariedel-cern: fixed.

PWGCF/FemtoDream/FemtoDreamCutculator.h
ariedel-cern: fixed.

PWGCF/FemtoDream/femtoDreamProducerReducedTask.cxx
```diff
@@ -153,28 +167,66 @@ struct femtoDreamProducerReducedTask {
...
+          particleOrigin = aod::femtodreamparticleMC::ParticleOriginMCTruth::kMaterial;
+        }
+      } else {
+        particleOrigin = aod::femtodreamparticleMC::ParticleOriginMCTruth::kFake;
+      }
+      outputPartsMC(particleOrigin, pdgCode, particleMC.pt(), particleMC.eta(), particleMC.phi());
+      // fill with correct values, this is currently placeholder
+      // outputDebugPartsMC(-999);
+    } else {
+      outputPartsMC(-999, -999, -999, -999, -999);
```
victor-gonzalez, line 221: Should not be safer to set `aod::femtodreamparticleMC::ParticleOriginMCTruth::kFake` instead of `-999`

victor-gonzalez (commented):
Please, consider my comments

PWGCF/DataModel/FemtoDerived.h
```diff
@@ -185,25 +185,36 @@ namespace femtodreamparticleMC
...
 };
 
-DECLARE_SOA_INDEX_COLUMN(FemtoDreamParticle, femtoDreamParticle);
+// DECLARE_SOA_INDEX_COLUMN(FemtoDreamCollision, femtoDreamCollision);
+// DECLARE_SOA_INDEX_COLUMN(FemtoDreamParticle, femtoDreamParticle);
 DECLARE_SOA_COLUMN(PartOriginMCTruth, partOriginMCTruth, uint8_t); //! Origin of the particle, according to femtodreamparticle::ParticleOriginMCTruth
 DECLARE_SOA_COLUMN(PDGMCTruth, pdgMCTruth, int);                   //! Particle PDG
+DECLARE_SOA_COLUMN(PtTruth, ptTruth, float);                       //! p_T of the generated particle (GeV/c)
+DECLARE_SOA_COLUMN(EtaTruth, etaTruth, float);                     //! Eta of the generated particle
+DECLARE_SOA_COLUMN(PhiTruth, phiTruth, float);                     //! Phi of the generated particle
```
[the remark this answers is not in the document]
ariedel-cern: The intention of the original author was to add columns containing MC truth to the table holding information about the reconstructed particle (I am refering to the tables of the femtodream particle table). Since in femtoscopy we only care about purity, we do not need the information about generated particles that are not being reconstructed (or reconstructed particles that do not pass minimal cuts). As far as I understand, each column needs a unique name, so that is why we have pt-pttruth, eta-etatruth.

If I understand your suggestions correctly, we should implement a standalone table holding the MC truth and pass both tables through the framework instead of joining the MC information to the existing table?

PWGCF/FemtoDream/FemtoDreamMath.h
ariedel-cern: Yeah, you are right.

PWGCF/FemtoDream/femtoDreamProducerReducedTask.cxx
ariedel-cern: So far the code was only meant to be used for the assumption that the paritcle is a proton and the default values have been set accordingly. That will be updated in the future.

PWGCF/DataModel/FemtoDerived.h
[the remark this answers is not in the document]
victor-gonzalez: No. What I'm suggesting is to get rid of these new column you are defining in here
The for the MC table declaration in L.209 use the columns declared in L.73-75, i.e.
```
DECLARE_SOA_COLUMN(Pt, pt, float);                    //! p_T (GeV/c)
DECLARE_SOA_COLUMN(Eta, eta, float);                  //! Eta
DECLARE_SOA_COLUMN(Phi, phi, float);                  //! Phi
```
Then, as far a you use only this columns in the routines, you don't need to replicate any code, the template mechanism ensure that things are done correctly. 
Of course you keep populating the MC table in the same way you do it now. Nothing needs to be changed.
The only things
- remove the additional columns
- use the same columns as for reco and data
- remove the replicated routines
- call the same templated routines as for reco and data

As said, this will work for the cases shown in this PR
It should work also in the general case but when something specific is need which differentiates the MC table from the Reco/Data table

PWGCF/FemtoDream/FemtoDreamMath.h
victor-gonzalez: No. You should get rid of these routines
You use a `constexpr` (not necessarily a template argument) when you need really differentiate between both cases but, pT is pT, eta is eta and phi is phi, i.e. no need in this case

PWGCF/DataModel/FemtoDerived.h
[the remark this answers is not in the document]
ariedel-cern: Sorry, Victor, I think we are talking passed each other or maybe the lableing of our columns is missleading (or there is something else I am missing, because implementing your changes would, I think, require quite some changes to the framework)

The columns ptTruth, etaTruth and phiTruth, which we use to compose the FemtoDreamParticlesMC table, are not meant to be used standalone, but must be joined with the FemtoDreamParticles table when we do MC processing. That means one row in the joined table (i.e. one reconstructed particle) will have columns containing values for pt, eta, phi (on the reconstructed level) and ptTruth, etaTruth and phiTruth (on the generator level), which will in general be different values. Like I said before, in femtoscopy we only care about the MC truth of particles that are reconstructed and who pass the minimal selections. In our producer task we only ever iterate over reconstructed particles, extract the information we need (including MC truth, if available) and write them into our FemtoDreamParticleTable. In principle we could add just 3 more columns to FemtoDreamParticleTable for the MC truth and just set them to some default value when there is no MC truth, but we thought it would be smarter if write these values into seperate columns (or table) which we can join later on demand. (Strictly speaking, we would need 4 columns, since we also need to the know the particle's origin).

That was also the reason why we orignially implemented two versions for the getkstar method, since we used different columns (of the same particle) for the computation of kstar and get different values (a lazy solution, I admit).

I think a straight forward solution to the code duplication problem would be go get rid off the template of the getkstar and similar methods and don't pass the particles and just pass the kinematic information of both particles (i.e. pt,eat,phi,mass per particle) directly as floats. I already implemented a patch for that, if you are fine with that solution.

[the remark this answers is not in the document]
victor-gonzalez: Well, in my opinion both are bad design decisions
- joining the reco and gen tables forces you to replicate code
- getting rid of the templating forces you to store table values in other places in memory which penalize the execution and is highly not recommended 

The right way of doing it should be the one use in the overall data model, i.e. having a table which addresses the MC particles and which is joinable to the reconstructed tracks table. For the kinematic variables the MC and the tracks tables share the column names so that the functions that manipulate them are easily templated. Of course, in the equivalent MC table in your case you only put what you want, i.e. only MC particles associated to reconstructed tracks. The advantage is that if you run on MC productions on

PR #1648, opened 2022-12-13
HF offline trigger selections for D mesons with gamma

mikesas:


EventFiltering/PWGHF/HFFilter.cxx
```diff
@@ -985,6 +1032,25 @@ struct HfFilter { // Main struct for HF triggers
...
+            if (isGamma) {
+              std::array<float, 3> gammaVec = {gamma.px(), gamma.py(), gamma.pz()};
+              auto massGammaCharm = RecoDecay::m(std::array{pVec3Prong, gammaVec}, std::array{massCharmHypos[iHypo], massGamma});
+              if (massGammaCharm < 3.) { // remove candidates with invariant mass above some value
+                keepEvent[kGammaCharm3P] = true;
+              }
+            }
+          }
+        }
+      } // end gamma selection
```
fgrosa, line 1052: There is no need to loop over all the 3-prong species; for sure we do not need the D+, while for the baryons there were no states decaying to Lc/Xic gamma so far, so I would also let them out (at least for the moment)
```suggestion
      if (!keepEvent[kGammaCharm3P] && isCharmTagged[kDs - 1]) {
        for (auto& gamma : theV0s) {
          float V0CosinePA = gamma.v0cosPA(collision.posX(), collision.posY(), collision.posZ());
          bool isGamma = isSelectedGamma(gamma, V0CosinePA);
          if (isGamma) {
            std::array<float, 3> gammaVec = {gamma.px(), gamma.py(), gamma.pz()};
            auto massGammaCharm = RecoDecay::m(std::array{pVec3Prong, gammaVec}, std::array{massDs, 0.});
            if (massGammaCharm < 3.) { // remove candidates with invariant mass above some value (TODO: set me as a configurable)
              keepEvent[kGammaCharm3P] = true;
            }
          }
        }
      } // end gamma selection
```

fgrosa (changes requested):
Ciao @mikesas thanks a lot for the development! I just have one comment (see below), then we can merge it 👍

fgrosa (approved):

fgrosa:
Hi @mpuccio could you approve this PR? We implemented a small change in `filterTables.h` and I have no permissions to approve it. Thanks!

mpuccio (approved):

PR #1651, opened 2022-12-13
[EMCAL-670, EMCAL-734, EMCAL-850] Dedicated QA task for EMCAL event selection + more histograms in cluster task

mfasDa:
- [EMCAL-670, EMCAL-850] Outsource event selection histograms to dedicated task
  + global BC % 3564 for all triggered BCs, BCs with
     EMCAL readout, BCs with EMCAL cell content and
     collisions
  + Adapt EMCAL readout check for run2 data

- [EMCAL-734] Add more histograms to cluster monitor
  + Number of clusters per event / bc total and per
     supermodule
  + Cluster energy spectrum vs. supermodule ID
  + Per supermodule:
     * Time vs. energy
     * Number of cells vs. energy
     * Shower shape parameters vs. energy

### DIFF
--- a/PWGJE/TableProducer/emcalCorrectionTask.cxx
+++ b/PWGJE/TableProducer/emcalCorrectionTask.cxx
@@ -14,6 +14,7 @@
 // Author: Raymond Ehlers & Florian Jonas
 
 #include <algorithm>
+#include <iostream>
 #include <cmath>
 
 #include "CCDB/BasicCCDBManager.h"
@@ -39,8 +40,6 @@
 using namespace o2;
 using namespace o2::framework;
 using namespace o2::framework::expressions;
-
-using bcEvSelIt = o2::soa::Join<o2::aod::BCs, o2::aod::BcSels>::iterator;
 
 struct EmcalCorrectionTask {
   Produces<o2::aod::EMCALClusters> clusters;
@@ -79,8 +78,6 @@ struct EmcalCorrectionTask {
   OutputObj<TH2I> hCellRowCol{"hCellRowCol"};
   OutputObj<TH1F> hClusterE{"hClusterE"};
   OutputObj<TH2F> hClusterEtaPhi{"hClusterEtaPhi"};
-  OutputObj<TH1F> hCollisionMatching{"hCollisionMatching"};
-  OutputObj<TH1F> hCollisionMatchingReadout{"hCollisionMatchingReadout"};
 
   void init(InitContext const&)
   {
@@ -142,8 +139,6 @@ struct EmcalCorrectionTask {
     hCellRowCol.setObject(new TH2I("hCellRowCol", "hCellRowCol;Column;Row", 97, 0, 97, 600, 0, 600));
     hClusterE.setObject(new TH1F("hClusterE", "hClusterE", 200, 0.0, 100));
     hClusterEtaPhi.setObject(new TH2F("hClusterEtaPhi", "hClusterEtaPhi", 160, -0.8, 0.8, 72, 0, 2 * 3.14159));
-    hCollisionMatching.setObject(new TH1F("hCollisionMatching", "hCollisionMatching", 3, -0.5, 2.5));        // 0, no vertex,1 vertex found , 2 multiple vertices found
-    hCollisionMatchingReadout.setObject(new TH1F("hCollisionMatching", "hCollisionMatching", 3, -0.5, 2.5)); // 0, no vertex,1 vertex found , 2 multiple vertices found, only filled for BCs with kTVXinEMC
   }
 
   // void process(aod::Collision const& collision, soa::Filtered<aod::Tracks> const& fullTracks, aod::Calos const& cells)
@@ -151,7 +146,7 @@ struct EmcalCorrectionTask {
   // void process(aod::BCs const& bcs, aod::Collision const& collision, aod::Calos const& cells)
 
   //  Appears to need the BC to be accessed to be available in the collision table...
-  void process(bcEvSelIt const& bc, aod::Collisions const& collisions, aod::Tracks const& tracks, aod::Calos const& cells)
+  void process(aod::BC const& bc, aod::Collisions const& collisions, aod::Tracks const& tracks, aod::Calos const& cells)
   {
     LOG(debug) << "Starting process.";
     // Convert aod::Calo to o2::emcal::Cell which can be used with the clusterizer.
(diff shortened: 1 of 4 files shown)


mfasDa:
@TimoWilken The CI fails in build/O2Physics/o2  with `Please resolve merge conflicts` which I don't understand, since the branch was rebased to master before pushing and also github should have complained itself if there was a merge conflict. There can't be a merge conflict - something must be going on with the check. Could you please have a look?

TimoWilken:
@mfasDa I'm not sure what's going wrong here -- maybe GitHub is having some issues? `cd O2Physics && git fetch upstream +pull/1649/head:test` and any number below 1649 works fine, but if I use 1650 or 1651 instead, I get `fatal: couldn't find remote ref pull/1650/head`. I'll see if we can open a ticket with GitHub...

ekryshen:
Hi Markus,

In general should be fine for isolated-bunch runs. I noticed two issues:
* collision.bc() is not good. It gives you a rough estimate of the collision bc so you will often see it shifted wrt actual colliding bc. I would suggest to process join of collision and evsel and use collision.foundBC() instead where foundBC is produced by evsel task via matching with FT0 entries. Then you can use collision.foundBC().globalBC().

* I am not sure if void process(bcEvSelIt const& bc, aod::Collisions const& collisions, aod::Tracks const& tracks, aod::Calos const& cells) is what you want. As far as I understand you effectively apply grouping on collisions and calos tables to select entries from the same bc. It might be safer to use process(bcs, collisions ...) and perform loops over bcs in the process function.

Evgeny

mfasDa:
Hi @ekryshen, I have adapted your first comment in the way you suggested. Could you please have a look? Unfortunately accessing `collision.foundBC().globalBC()` (same for ` collision.bc().globalBC()` I get a segfault, so I had to comment out the line (for the moment). Do you have an idea what could be the problem? I was trying on AODs from LHC22s apass4.

ekryshen:
Hi @mfasDa, actually my first comment is closely related to the second one. You are trying to access foundBC which is different from the current one, therefore you need to subscribe to the full BC table. See example here:
https://github.com/AliceO2Group/O2Physics/blob/master/Common/Tasks/eventSelectionQa.cxx#L995

And since your bcs are actually a join with bcsels, you need to use foundBC_as<>(), see example here:
https://github.com/AliceO2Group/O2Physics/blob/master/Common/Tasks/eventSelectionQa.cxx#L1016

mfasDa:
Hi @ekryshen, Thanks for your help! Based on your comment we think it is easier to outsource the event selection monitoring to a dedicated QA task, in order to minimise the dependencies. Therefore I have reverted the changes in the EMCAL correction task, which is now again a simple table producer, and moved the histograms, including your second suggestion, to the task emcaleventselectionqa.cxx. Does this monitoring task now look OK? I have tested it locally and now I see also the BC-collision histogram filled.

ekryshen:
Hi @mfasDa,
This is not exactly what I suggested. In particular I would remove bcEvSelIt const& bc from the subscription and would do a loop over bcs in the process function. Actually I am surprized that this double subscription to the bc table iterator and to the full bc table does work without crashes:
process(bcEvSelIt const& bc, collEventSels const& collisions, bcEvSels const& bcs,...

Why don't you use this approach?

process(collEventSels const& collisions, bcEvSels const& bcs,...){
  for (auto & bc: bcs) {
    ...
  }
  for (auto & collision : collisions) {
    ...
  }  
According to Anton, if one uses iterator in the first argument of the process function, then other subscribed tables should be filtered based on the index of the first argument. I wonder if you actually see all collisions or only some selection (can you check the size of the collision array?).

mfasDa:
Hi @ekryshen, I rewrote the task as you suggested above, though I don't fully understand why this complication is needed. Please have a look whether this is now in line with what you are suggesting.

For the matching of collisions to the BC do we need to use
``` c++
for (const auto& col : collisions) {
  if (col.bc_as<bcEvSels>() == bc) {
     ...
  }
}
```

or 

``` c++
for (const auto& col : collisions) {
  if (col.foundBC_as<bcEvSels>() == bc) {
     ...
  }
}
```
? I am wondering here also what to do in the correction task where we need to assign objects in the EMCAL - using BC as column - to collisions. In case foundBC would be needed I understand we need to use the internal matching as well, because collisions matched to some extend to wrong BCs by the automatic matching of the framework?

ekryshen:
@mfasDa 
To clarify a bit more. See example with tracks associated to collisions:
https://aliceo2group.github.io/analysis-framework/docs/basics-tasks/SubscribingToData.html#navigating-data-associations
Similarly, if you use bc iterator in the first argument, you'll preselect only those collisions which are associated to the current bc. While we need all collisions to search for correct one via foundBC index.

Regarding your code examples, you should use this one:
for (const auto& col : collisions) {
  if (col.foundBC_as<bcEvSels>() == bc) {
     ...
  }
}

Actually to avoid double loops, one can first loop over collisions before bc loop and create a multimap with (foundBCId, collisionIndex) pairs (the reason for multimap is that one may have several collisions per bc). You can access those indices via col.foundBCId() and col.globalIndex(). Then loop over bcs and check if current bc (with index bc.globalIndex()) is found in the map. If found, you can access corresponding collisions by index using collisions.iteratorAt(collisionIndex).

I am afraid these complications are unavoidable in your case.

ekryshen:
@mfasDa 
I've just realized that O2 analysis framework provides a nice tool that can be used instead of multimap to select collisions matched to the current bc via foundBCId. You should declare preslice before process function and use sliceBy:

  Preslice<collEventSels> perFoundBC = aod::evsel::foundBCId;

  void process(bcEvSels const& bcs, collEventSels const& collisions){
    for (auto const& bc : bcs) {
      auto collisionsGrouped = collisionss.sliceBy(perFoundBC, bc.globalIndex());
      if (collisionsGrouped.size()) {
        printf("%lu\n",collisionsGrouped.size());
      }
      for (auto const& col : collisionsGrouped){
       ...
      }
    }

mfasDa:
Hi @ekryshen, thanks a lot for the hint! Very useful. I have updated the emcal event selection QA task. If the task is fine I would like to ask to merge the code so that we can start using it on hyperloop. The change in the correction task will be done in a separate commit. We are now aware of this and will take action as soon as possible, but I would like to have this PR merged now and not wait again for another implementation - we need it now.

ekryshen:
@mfasDa This if is still trying to select collisions based on approximate bc.
```
if (col.bc_as<bcEvSels>() == bc) {
```
As I mentioned, you would have to use col.foundBC_as<bcEvSels>(). But in fact you don't need ifs, since you've already selected collisions matched with current bc according to foundBC. You can also use bcID:
```
      for (const auto& col : collisionsGrouped) {
          mHistManager.fill(HIST("hBCCollisions"), bcID);
      }
```

mfasDa:
Ooops, that line should have gone out. Thanks for spotting! And indeed the bcID is enough, no need to determine it again as foundBC was already used in the filter.

nzardosh (approved):

PR #1652, opened 2022-12-13
change to use autodetect feature

ddobrigk:
* the builder autodetect feature is picking up still _default_ cuts and not json-specific cuts from other tasks (in contact with Giulio and Anton about it, might be a framework bug, not sure) but in anticipation of it working, I am committing this change also for the MC tasks


njacazio (approved):
Thanks David!

PR #1654, opened 2022-12-14
DPG: extend TOF PID QA with TRD

njacazio:
- Add resolution plot in qaEventTrack

### DIFF
--- a/DPG/Tasks/AOTTrack/qaEventTrack.cxx
+++ b/DPG/Tasks/AOTTrack/qaEventTrack.cxx
@@ -120,7 +120,7 @@ struct qaEventTrack {
       return;
     }
     const AxisSpec axisPt{binsPt, "#it{p}_{T} [GeV/c]"};
-    const AxisSpec axisInvPt{100, 0, 100, "#it{p}_{T} [GeV/c]"};
+    const AxisSpec axisInvPt{100, 0, 100, "1/#it{p}_{T}_{gen} [GeV/c]^{-1}"};
     const AxisSpec axisEta{180, -0.9, 0.9, "#it{#eta}"};
     const AxisSpec axisPhi{180, 0., 2 * M_PI, "#it{#varphi} [rad]"};
     const AxisSpec axisVertexNumContrib{200, 0, 200, "Number Of contributors to the PV"};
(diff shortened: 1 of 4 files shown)


njacazio:
@mfaggin for the resolution what do you think?

njacazio:
> @mfaggin for the resolution what do you think?

i.e. here: https://github.com/AliceO2Group/O2Physics/pull/1654/files#diff-220de43dd7fbde6f6e1678306b6d4e4100c9e7e1ac1d0f21cf4e87e16684bccfR1111

DPG/Tasks/AOTTrack/qaEventTrack.cxx
```diff
@@ -119,6 +120,7 @@ struct qaEventTrack {
       return;
     }
     const AxisSpec axisPt{binsPt, "#it{p}_{T} [GeV/c]"};
+    const AxisSpec axisInvPt{100, 0, 100, "#it{p}_{T} [GeV/c]"};
```
mfaggin, line 123: Here the title should be changed, since you want 1/pt rather than pt

DPG/Tasks/AOTTrack/qaEventTrack.cxx
```diff
@@ -184,6 +186,10 @@ struct qaEventTrack {
     histos.add("Tracks/Kine/phivspt", "#varphi vs #it{p}_{T}", kTH2F, {axisPt, axisPhi});
     if (doprocessMC || doprocessRun2ConvertedMC) {
       histos.add<TH2>("Tracks/Kine/resoPt", "", kTH2D, {axisDeltaPt, axisPt});
+      auto hreso = histos.add<TH2>("Tracks/Kine/resoInvPt", "", kTH2D, {axisDeltaPt, axisInvPt});
```
mfaggin, line 189: here again: `axisDeltaPt` defines the residuals between the recon and gen pt, not 1/pt, right? If yes, one should change this axis as well

mfaggin (dismissed):
Thanks @njacazio , if I am not wrong a couple of fixes on the axes are required. Apart from this, we can then go ahead.

DPG/Tasks/AOTTrack/qaEventTrack.cxx
njacazio: Good catch

DPG/Tasks/AOTTrack/qaEventTrack.cxx
njacazio: I will use the same binning but different title

alibuild (approved):

PR #1658, opened 2022-12-14
PWGDQ: add forward q-vector process, configurables and small fix

robincaron13:
- Make the task able to compute q-vectors using forward MFT tracks (even if centrality is not yet available). 
- Add a small fix to prevent to fill nan values for dilepton into FillPairVn function.
- Add more configurables for minimal/maximal eta values for selecting ranges and gaps.


robincaron13:
Hi @iarsene @dsekihat,
Here I think the MegaLinter error should not be related

dsekihat:
Hi Robin, can you change the order of "include" and replace long with int16 or int64 in dqFlow.cxx? Mega Linter complains these points.

iarsene (approved):

iarsene:
Hi @dsekihat and @robincaron13, 
I agree that in principle we should fix the issues with megaLinter otherwise these will be more and more. There are many which are not related with your PR indeed. We should at some point take care of all of them. But if there are a few related to your PR, would be good if you fix them. Here i mean the ones in dqFlow. I would leave out those in VarManager, which i will try to fix.

PR #1659, opened 2022-12-14
PID: fix definition of expected sigma

jezwilkinson:
Small bugfix to the TPC response to use the expected signal, not the measured signal, to define the sigma


njacazio (approved):
Superb Jeremy! Thanks a lot!

ddobrigk (approved):

PR #1662, opened 2022-12-14
PWGHF: add check on PDG code for Lc 1st daughter in MC.

mfaggin:
Hi,
looking at the code I see that when filling the `MC/reconstructed` histograms for the Lc mass we do not check explicitly the real mass hypothesis, but we do only rely on the response from the reconstructon cuts (`isSelLcToPKPi` and `isSelLcToPiKP`). If we do not do it, we risk to do double counting if for our reconstructed candidate both `isSelLcToPKPi` and `isSelLcToPiKP` are true, even if we know that this is a real Lc from here: https://github.com/AliceO2Group/O2Physics/blob/master/PWGHF/Tasks/taskLc.cxx#L321
Am I right or is there something that I am missing?

### DIFF
--- a/PWGHF/Tasks/taskLc.cxx
+++ b/PWGHF/Tasks/taskLc.cxx
@@ -325,11 +325,11 @@ struct HfTaskLc {
         registry.fill(HIST("MC/generated/signal/hPtGenSig"), particleMother.pt()); // gen. level pT
         auto pt = candidate.pt();
         /// MC reconstructed signal
-        if (candidate.isSelLcToPKPi() >= selectionFlagLc) {
+        if (candidate.isSelLcToPKPi() >= selectionFlagLc && abs(candidate.prong0_as<aod::BigTracksMC>().mcParticle().pdgCode()) == kProton) {
           registry.fill(HIST("MC/reconstructed/signal/hMassRecSig"), invMassLcToPKPi(candidate));
           registry.fill(HIST("MC/reconstructed/signal/hMassVsPtRecSig"), invMassLcToPKPi(candidate), pt);
         }
-        if (candidate.isSelLcToPiKP() >= selectionFlagLc) {
+        if (candidate.isSelLcToPiKP() >= selectionFlagLc && abs(candidate.prong0_as<aod::BigTracksMC>().mcParticle().pdgCode()) == kPiPlus) {
           registry.fill(HIST("MC/reconstructed/signal/hMassRecSig"), invMassLcToPiKP(candidate));
           registry.fill(HIST("MC/reconstructed/signal/hMassVsPtRecSig"), invMassLcToPiKP(candidate), pt);
         }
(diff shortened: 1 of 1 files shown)


mfaggin:
@vkucera @ginnocen @cterrevo @fcolamar

vkucera:
Hi @mfaggin , thanks for pointing this out. What you suggest makes sense to me. Shouldn't we then do the same check for the `hMassRecSig` and `invMassLcToPKPi` histograms?

mfaggin:
> Hi @mfaggin , thanks for pointing this out. What you suggest makes sense to me. Shouldn't we then do the same check for the `hMassRecSig` and `invMassLcToPKPi` histograms?

Hi @vkucera , yu are totally right, my bd. I was forgetting the non-prompt case as well as the inclusive one. Now it should be correct.

PWGHF/Tasks/taskLc.cxx
```diff
@@ -325,11 +325,11 @@ struct HfTaskLc {
         registry.fill(HIST("MC/generated/signal/hPtGenSig"), particleMother.pt()); // gen. level pT
         auto pt = candidate.pt();
         /// MC reconstructed signal
-        if (candidate.isSelLcToPKPi() >= selectionFlagLc) {
+        if (candidate.isSelLcToPKPi() >= selectionFlagLc && abs(candidate.prong0_as<aod::BigTracksMC>().mcParticle().pdgCode()) == kProton) {
```
vkucera, line 328:
```suggestion
        if (candidate.isSelLcToPKPi() >= selectionFlagLc && std::abs(candidate.prong0_as<aod::BigTracksMC>().mcParticle().pdgCode()) == kProton) {
```

PWGHF/Tasks/taskLc.cxx
```diff
@@ -367,11 +367,11 @@ struct HfTaskLc {
 
         /// reconstructed signal prompt
         if (candidate.originMcRec() == RecoDecay::OriginType::Prompt) {
-          if (candidate.isSelLcToPKPi() >= selectionFlagLc) {
+          if ((candidate.isSelLcToPKPi() >= selectionFlagLc) && abs(candidate.prong0_as<aod::BigTracksMC>().mcParticle().pdgCode()) == kProton) {
```
vkucera, line 370:
```suggestion
          if (candidate.isSelLcToPKPi() >= selectionFlagLc && abs(candidate.prong0_as<aod::BigTracksMC>().mcParticle().pdgCode()) == kProton) {
```
And everywhere else

vkucera (changes requested):
Thanks, I added a few comments.

PWGHF/Tasks/taskLc.cxx
mfaggin: done

PWGHF/Tasks/taskLc.cxx
mfaggin: done

vkucera (approved):

PR #1682, opened 2022-12-17
PID: remove QA from TPC task to reduce memory usage

njacazio:
@jezwilkinson I propose this change to reduce the memory usage in compilation, I did the same for TOF few weeks ago, what do you think? The feature would be nice to have but it's more important to keep compilation ongoing


jezwilkinson:
If it impacts the compilation then sure, this seems fine. And makes sense to have TPC and TOF consistent here. What's the change in the workflow to attach the QA task after this?

njacazio:
> If it impacts the compilation then sure, this seems fine. And makes sense to have TPC and TOF consistent here. What's the change in the workflow to attach the QA task after this?
The change is that `--add-qa 1` will be not producing QA output anymore (it anyway worked only when running locally), the changes will be to add the `o2-analysis-pid-tpc-qa` as we do on the hyperloop so no loss of features apart from the possibility to have it implicitly done in the PID task

ddobrigk (approved):
thanks, nicolo!

alibuild (approved):

PR #1683, opened 2022-12-17
[PWG-DQ] Apply TPC postcalibration via CCDB

iarsene:
1) Apply the TPC postcalibration via CCDB
- postcalibration can be applied either in table-maker (produced skims will already include postcalibrated TPC n-sigmas) or at the analysis level. 
- calibration parameters are loaded from CCDB (path can be specified via configurables) based on timestamp.
2) timestamp was added in the datamodel. Currently, BC quantities via BCsWithTimestamps are not computed in the VarManager due to problems at compilation or running which are not yet understood. So for now, the BC related quantities are computed directly in the table-maker. Note that this does not affect the skimmed data model, which includes the BC info in a different table.


alibuild (approved):

PR #1690, opened 2022-12-19
Replace old builders with newer ones, following full validation

ddobrigk:


ddobrigk:
* cross-checks on physics performance have shown that these builders have the same efficiencies as the old ones 
* CPU benchmarking has shown that these builders have super computing performance by something around 2x-5x, depending on system. The larger increases take place in high-multiplicity events (e.g. Pb-Pb). 
* Once this is merged, the newer builder tasks (that were placed side-by-side with the older ones for checking) will be officially deprecated and removed from the repository.

njacazio (approved):
Thanks David, we will remove the duplicate builders in a next PR right?

ddobrigk:
> Thanks David, we will remove the duplicate builders in a next PR right?

Yes, right! I will do that once this is merged (and after I send a message around with this replacement)

PR #1693, opened 2022-12-19
[Event Filtering] Add QA histo with pT vs mass for D*0 and Ds*+

fgrosa:
@mikesas I added two histos with the Dstar0 and Dstar_s+ masses to check that we actually trigger on real signals


alibuild (approved):

PR #1697, opened 2022-12-19
PWGHF: fix process_switch candidateCreatorOmegac

ZFederica:


vkucera:
Hi @ZFederica , why are these additional functions needed? I don't see why their absence would prevent the workflow from running.

vkucera (approved):

PR #1700, opened 2022-12-19
LF: spectra task, add centrality FT0C

njacazio:
- Remove subscription to tiny tables
- Add parametrizations for negative particles


alibuild (approved):

PR #1703, opened 2022-12-20
Add per BC QA, add Nch profiles for MC centrality

ddobrigk:


alibuild (approved):

PR #1705, opened 2022-12-20
Fix to allow that the extended track-selection table must is produced…

arossi81:
… also when running on run2 converted data.


mfaggin:
thanks @arossi81 . I approve it. @saganatt this should fix the issue you spotted on run2 converted data.
@pzhristov can you please allow the merging once the compilation tests are ok? Thanks.

njacazio (approved):
Thanks Andrea, indeed this was needed

ddobrigk (approved):

PR #1713, opened 2022-12-22
midrap Nch profiles: add switch to use Zeq or not

ddobrigk:


alibuild (approved):

PR #1726, opened 2022-12-27
cross-check if all is fine, otherwise fatal

ddobrigk:


ddobrigk:
I am provisionally merging this during the holiday break to make sure everything continues as planned. Note that this is an extremely simple fix, and the most important aspect of this is, in fact, the error message explaining people what to do.

alibuild (approved):

PR #1732, opened 2022-12-29
additional PV study functionality

ddobrigk:


PWGLF/Tasks/QC/straRecoStudy.cxx
```diff
@@ -79,27 +79,80 @@ struct preProcessMCcollisions {
...
+
+  struct collisionStats { // This structure is named "myDataType"
+    float nContribsWithTPC;
+    float nContribsWithTRD;
+    float nContribsWithTOF;
+    float nContribsWithITS;
+  };
+
+  template <typename T>
+  std::vector<std::size_t> sort_indices(const std::vector<T>& v)
```
aalkin, line 92: Why not use `std::sort` directly?

PWGLF/Tasks/QC/straRecoStudy.cxx
```diff
@@ -79,27 +79,80 @@ struct preProcessMCcollisions {
...
-      for (auto& collision : collisions) {
-        if (collision.has_mcCollision()) {
-          if (mccollision.globalIndex() == collision.mcCollision().globalIndex()) {
-            lExists++;
-          }
+    std::vector<collisionStats> collisionStats;
+    std::vector<int> collisionNContribs;
+    collisionStats.reserve(collisions.size());
+    // collisionNContribs.reserve(collisions.size());
+    for (Int_t ic = 0; ic < collisions.size(); ic++) {
```
aalkin, line 117: It would be easier to use the default zeros in the struct definition and instead of reserving the vector you can fill it with N default-constructed structs on construction.

PWGLF/Tasks/QC/straRecoStudy.cxx
ddobrigk: Because sort sorts the information in a vector but I need the indices...

PWGLF/Tasks/QC/straRecoStudy.cxx
ddobrigk: Ah sure, this I can do, fair enough :-)

PWGLF/Tasks/QC/straRecoStudy.cxx
aalkin: If you have sorted the vector, the index is also sorted. Meaning you can iterate from 0 to size-1 knowing that elements would be ordered as they should.

ddobrigk: No, that's not what I wanted to accomplish. I really need the original index to ordered index map, because I am going to go over the other vectors (number of contributors who have certain detector signals) with that map. Like this, I can be sure that I am accessing the information coming from exactly the right collision. I don't think there is an easy way to do that with a single use of std::sort, unless I miss something...and imho having this in a neat function that gives indices after sorting is quite a clean way of doing that :-D

aalkin: You can probably use some better data structure and fill things directly at numContribs position to have the ordering by construction. The current code does not really require the ordering, however, I assume this is for later additions.

ddobrigk: Sure, there are always different ways of doing things, and I am aware that I could have made the sorting on somewhat different data structures that could be more convenient in the future, but this works fine for now indeed ;-) 

The current code does this ordering because I want to store the collision statistics and properties from largest to smallest, assuming that there may be a pattern in terms of how different contribution sources lead to split vertices (which I do see in MC). Since this assumption is reasonable for a first study, then yes, what I wanted to accomplish here did require the ordering... Because otherwise, the information could come in garbled (ordered in ways that are not indicative of splitting causes) as first and second collision could be swapped with regards to those properties. That was my reasoning. Does that make sense?

aalkin: I mean that ordering is not important if you fill the information into a histogram, because it is not preserved there. Or you can achieve automatic ordering, adding a numContrib axis. Am I missing something?

ddobrigk: The histogram I fill is actually ordered from greater to smaller collision (X axis) and then it contains information for collisions being reconstructed N times (Y axis). Yeah, I could do a numcontrib axis instead of the X axis, but I should then also pack the information regarding PV index (n-th time the same MC collision has been reconstructed), and there's actually the difficulty. If I just replaced the X axis with Ncontribs, I would lose that information. 

In the broadest sense, what is interesting is actually the correlation between number of times a MC collision has been reconstructed and the number of corresponding contributors - and the order in which these PVs are stored could play a confusing role if we aren't careful. That's what I wanted to study, but I still wanted to understand if there is some sort of pattern there. For instance: maybe we have a bigger Ncontribs collision with a subset of detectors providing timing information and then a smaller Ncontribs collision with another smaller subset providing displaced information. Finding a pattern like this would be best for us to understand, but to be sure, this code that I am doing is only a very first and rough and approximate attempt at that. For sure it will have to be elaborated more. But one has to start somewhere and see what comes out :-)

ddobrigk: @aalkin the latest PR should take care of your comments + adds new features too :-D

And here comes a spoiler: the split vertices that I see are quite numerous, and it seems that the biggest vertex always aggregates most tracks with TOF timing and the smallest vertex always aggregates most tracks with TRD timing. This is an indication that the timing information isn't completely consistent. I'll prepare something about that for us all to discuss early next year... Enjoy the holidays!

PWGLF/Tasks/QC/straRecoStudy.cxx
```diff
@@ -79,27 +79,129 @@ struct preProcessMCcollisions {
...
-            lExists++;
-          }
+    int lNumberOfXi = 0;
+    for (auto& mcp : mcParticles) {
+      // mimic triggering strategy precisely
+      if (TMath::Abs(mcp.eta()) < 0.8 && mcp.pdgCode() == 3312)
+        lNumberOfXi++;
+    }
+
+    std::vector<collisionStats> collisionStatAggregator;
```
aalkin, line 151: `std::vector` has a constructor with a single int argument, that creates a vector of that size filled with default-constructed items of the vector element type, no need to emplace them in a loop.

ddobrigk: thanks, done!

njacazio (approved):

PR #1734, opened 2023-01-01
increase tolerance for CYY check

ddobrigk:
Interestingly, in MC, a very very small (order of 1 per million or less) fraction of PVs seem to have very-close-to-zero negative CYY values. This is likely a numerical precision problem (to be checked with Ruben after the holiday period). In the meantime, this PR increases the tolerance so that it doesn't trigger on that. Precision check was CYY < -1e-8, is now CYY < -1e-6.


alibuild (approved):

PR #1736, opened 2023-01-02
Added BC filter diffractionBCFilter

pbuehler:
This is an event filter task which selects BCs instead of Collisions


mpuccio (approved):

alibuild (approved):

PR #1745, opened 2023-01-09
[PWGMM] addition of headers

jikim1290:
addition of headers


PWGMM/Lumi/Tasks/lumi.cxx
```diff
@@ -42,13 +43,18 @@
 #include "CCDB/BasicCCDBManager.h"
 #include "DataFormatsParameters/GRPObject.h"
```
aalkin, line 44: This header and `BasicCCDBManager.h` are included twice.

aalkin (changes requested):
Please remove duplicated includes (this is what causes megalinter test to fail).

aalkin (approved):

PR #1746, opened 2023-01-09
[PWG-DQ] reducing binning in histograms for tpcpid and tofpid

amerzlay:


feisenhu:
Hello @amerzlay,
I kindly want to ask if you made these changes of the binning in agreement with someone or if there is a specific reason/need for the reduction of the number of bins?
For some people in PWG-EM (including me) who also use these histograms need a finer binning of these histograms. 
Kind regards, Florian.

iarsene:
Hi @feisenhu,
I asked Anastasia to have a look into this and implement some reductions in binning. Ofcourse if detailed binning is required we need to find a solution. But what triggered this is that when one switches on several histogram subdirectories and runs e.g. over MC with several signals as well, we get a crash because we allocate too many bins. This is a ROOT issue which until now we ignored because we had relatively few histograms defined in the library. So one solution is to reduce binning in 2D or higher dimension histograms if its not really needed. We have to discuss on this and decide on some policy, because I am quite sure many of us will hit this very soon. One solution, is that very granular histograms should go into separate histogram sub-classes which are for some targeted purposes. For example, not everyone needs 2D histograms where the eta dimension has a granularity of 0.01, or the pIN a granularity of 10 MeV.
We could also define some special sub-classes for simple QA purposes.
So for this PR, is there some specific binning for which these changes are too coarse? What is the granularity that you need?
Cheers, Ionut

feisenhu:
Hey @iarsene ,

Thank you for clarifying.
As we have discussed, I made some changes to the HistogramsLibrary in which I introduced two new sub-classes for the TPC and TOF PID (`tpcpid_fine`, `torpid_fine`) in which I used a variable binning. If the sub-classes are not used, it uses the plots defined here. This should help to reduce the number of bins for both cases.

So from my side it is ok to merge this PR. Afterwards I can create a new PR and apply my changes.
Cheers Florian

iarsene (approved):

PR #1749, opened 2023-01-09
PWGDQ: standardize task, histo names and make centrality available

robincaron13:
- Standardize the task name for DQ framework, also for qvector histograms
- Make available the centrality values from the FT0C detector for this task by creating a new `ObjType` called `CollisionCentRun3` in order to fill the associated centrality values within the `VarManager` 
- Modify the qvector processes to add the centrality and add the associated histograms 
- Reorganize the libraries and change of `int` values into `uint8_t`

### DIFF
--- a/PWGDQ/Core/HistogramsLibrary.h
+++ b/PWGDQ/Core/HistogramsLibrary.h
@@ -77,6 +77,8 @@ void o2::aod::dqhistograms::DefineHistograms(HistogramManager* hm, const char* h
     if (subGroupStr.Contains("cent")) {
       hm->AddHistogram(histClass, "CentV0M", "CentV0M", false, 100, 0., 100., VarManager::kCentVZERO);
       hm->AddHistogram(histClass, "CentV0M_vtxZ", "CentV0M vs Vtx Z", false, 60, -15.0, 15.0, VarManager::kVtxZ, 20, 0., 100., VarManager::kCentVZERO);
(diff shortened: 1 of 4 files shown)


robincaron13:
Hi @iarsene @dsekihat @feisenhu,
1. As you see, I hope that creating a dedicated `ObjType` for filling the `event.centFT0C()` in the `VarManager` is fine for now. Afterwards we will have to handle with different centrality values coming from different detectors, as well for filling the histograms with these different types.
2. It seems that MegaLinter is complaining about the cast of TH3F in the `VarManager` (e.g. this line 864: `TH3F* calibMean = (TH3F*)fgCalibs[kTPCElectronMean];`) so not really related to these changes.

iarsene:
Hi Robin,
So, i see that indeed you capture the centFT0C just in the VarManager. So you will have it available for the dqFlow task and other task which would run on the Framework data model. The variable is then lost if one want to run a flow analysis using the derived DQ data model. We need the FIT centrality and other event wise quantities in the data model and to be skimmed. But I kind of started working on this with one of my students, so I think that for now we can approve this PR. 
Regarding the second issue, that is fixed in a PR by Florian.

robincaron13:
Hi Ionut,
Thanks a lot for reviewing. Indeed I did not add the variable into the derived DQ tables, so in a sense it is just a quick start to be continued (e.g. later it could be added with other centrality values to the `ReducedEventsExtended`).

iarsene (approved):

PR #1753, opened 2023-01-10
PWGMM: dndeta: various improvements

aalkin:
* Only add histograms to registry when the corresponding process is enabled
* Use `THnSparse` in place of `TH3` to save memory


alibuild (approved):

PR #1755, opened 2023-01-10
PWGHF: First attempt on accounting for ambiguous tracks in HF vertexing

fgrosa:
In this draft PR @mfaggin and I implemented a code that accounts for the ambiguous tracks in the HF vertexing.

In this implementation, an additional table of tracks is created cloning all the ambiguous tracks by propagating them to all the other collisions to which they are compatible. The same is done for PV contributors in case of multiple collisions per BC (in-bunch pileup).

An improved version is also under development with @jgrosseo but, in case it is decided to move ahead with this one, before merging it will require optimisation.

### DIFF
--- a/PWGHF/TableProducer/trackIndexSkimCreator.cxx
+++ b/PWGHF/TableProducer/trackIndexSkimCreator.cxx
@@ -394,7 +394,7 @@ struct HfTrackIndexSkimCreatorProduceAmbTracks {
               /// Fill the table with this track propagated to the new collisions
               collIndices.push_back(collision.globalIndex());
               trackIndices.push_back(track.globalIndex());
-              trackTypes.push_back(BIT(hf_amb_tracks::Ambiguous));
+              trackTypes.push_back(BIT(hf_amb_tracks::PVContributor));
               trackX.push_back(trackParCov.getX());
               trackAlpha.push_back(trackParCov.getAlpha());
               trackY.push_back(trackParCov.getY());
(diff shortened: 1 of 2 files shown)


PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -252,10 +252,201 @@ struct HfTrackIndexSkimCreatorTagSelCollisions {
...
+              dcaInfoCov.set(999, 999, 999, 999, 999);
+              auto trackParCov = getTrackParCov(track);
+              vtx.setPos({collision.posX(), collision.posY(), collision.posZ()});
+              vtx.setCov(collision.covXX(), collision.covXY(), collision.covYY(), collision.covXZ(), collision.covYZ(), collision.covZZ());
+              o2::base::Propagator::Instance()->propagateToDCABxByBz(vtx, trackParCov, 2.f, matCorr, &dcaInfoCov);
+
+              /// Fill the table with this track propagated to the new collisions
+              collIndices.push_back(collision.globalIndex());
+              trackIndices.push_back(track.globalIndex());
+              trackTypes.push_back(BIT(hf_amb_tracks::Ambiguous));
```
mfaggin, line 397: @fgrosa here probably we wrongly changed the flag, it should be `hf_amb_tracks::PVContributor`

mfaggin (commented):
@fgrosa here we are with the fix on CCDB init (tested locally as well, it seems it works)

PR #1757, opened 2023-01-10
PWGDQ: Introduce hisogram for PdgCode for mc tracks and their mothers…

feisenhu:
…, implemented MotherPDG in VarManager

I have introduced a new `void` called `FillTrackMC` in which the previous MC track variables are moved to and further more now the pdg code of the mother particle is calculated.
In order to get access to the mother PdgCode the MCParticle stack has to be passed to the function.

Furthermore, I introduced the corresponding histograms in the HistogramsLibrary.h

### DIFF
--- a/PWGDQ/Core/VarManager.h
+++ b/PWGDQ/Core/VarManager.h
@@ -860,8 +860,8 @@ void VarManager::FillTrack(T const& track, float* values)
 
     // compute TPC postcalibrated electron nsigma based on calibration histograms from CCDB
     if (fgUsedVars[kTPCnSigmaEl_Corr] && fgRunTPCPostCalibration[0]) {
-      TH3F* calibMean = (TH3F*)fgCalibs[kTPCElectronMean];
-      TH3F* calibSigma = (TH3F*)fgCalibs[kTPCElectronSigma];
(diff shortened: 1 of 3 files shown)


PWGDQ/Core/VarManager.h
```diff
@@ -963,6 +968,18 @@ void VarManager::FillTrack(T const& track, float* values)
...
+
+template <uint32_t fillMap, typename U, typename T>
+void VarManager::FillTrackMC(const U& mcStack, T const& track, float* values)
+{
+  if (!values) {
+    values = fgValues;
+  }
+
+  // Quantities based on the mc particle table
   if constexpr ((fillMap & ParticleMC) > 0) {
```
iarsene, line 983: Since now this function has only one purpose (fill quantities for the ParticleMC), the template argument fillMap is not really necessary at the moment. I suggest to drop it from the function definition, unless you have plans in the short term to include more bits in the fillMap which would apply for this function.

iarsene:
Hi @feisenhu, 
The PR is fine for me. I have only a small comment, see above.

iarsene (approved):

PR #1758, opened 2023-01-10
DPG: add correlation PV contr. with TOF and with TRD.

mfaggin:
@ddobrigk this should implement your suggestion


alibuild (approved):

PR #1759, opened 2023-01-11
PWGHF: load and apply the TPC postcalibration from CCDB

zhangbiao-phy:


fgrosa:
Ciao @zhangbiao-phy it looks good to me, thanks! If you have tested it and it works we can merge it. I was just wondering whether it would not be convenient to create a table with postcalibrated nSigma that can be used for the different triggers (e.g. for c-hadron daughters, but then also for the proton in the `kFemto` one to avoid the recomputation), but for the moment we can stay with this and improve it later.

zhangbiao-phy:
> Ciao @zhangbiao-phy it looks good to me, thanks! If you have tested it and it works we can merge it. I was just wondering whether it would not be convenient to create a table with postcalibrated nSigma that can be used for the different triggers (e.g. for c-hadron daughters, but then also for the proton in the `kFemto` one to avoid the recomputation), but for the moment we can stay with this and improve it later.



> 

Ciao @fgrosa, Thanks for the comments and suggestions! Yes, I have tested it with the data (LHC22m and LHC22f), It works. Then Also we can create a table with postcalibrated nSigma and add it to the optimisationTree. I will make a new PR later. Thanks a lot!

EventFiltering/PWGHF/HFFilter.cxx
```diff
@@ -138,6 +138,9 @@ struct HfFilter { // Main struct for HF triggers
 
   // parameter for Optimisation Tree
   Configurable<bool> applyOptimisation{"applyOptimisation", false, "Flag to enable or disable optimisation"};
+  Configurable<bool> fConfigComputeTPCpostCalib{"cfgTPCpostCalib", false, "If true, compute TPC post-calibrated n-sigmas"};
+  Configurable<string> fConfigCcdbPathTPC{"ccdb-path-tpc", "Users/i/iarsene/Calib/TPCpostCalib", "base path to the ccdb object"};
+  int fCurrentRun; // needed to detect if the run changed and trigger update of calibrations etc.
```
vkucera, line 143:
```suggestion
  int currentRun; // needed to detect if the run changed and trigger update of calibrations etc.
```

EventFiltering/PWGHF/HFFilter.cxx
```diff
@@ -157,6 +160,12 @@ struct HfFilter { // Main struct for HF triggers
   std::array<std::shared_ptr<TH1>, kNCharmParticles> hBDTScorePrompt{};
   std::array<std::shared_ptr<TH1>, kNCharmParticles> hBDTScoreNonPrompt{};
 
+  // Histogram of TPC postcalibration map for pion and proton
+  TH3F* mapPionMean;
+  TH3F* mapPionSigma;
+  TH3F* mapProtonMean;
+  TH3F* mapProtonSigma;
```
vkucera, line 167: It is not clear from the names that these are histograms.

EventFiltering/PWGHF/HFFilter.cxx
```diff
@@ -235,6 +244,48 @@ struct HfFilter { // Main struct for HF triggers
...
   }
+
+  /// compute TPC postcalibrated nsigma based on calibration histograms from CCDB
+  /// \param objectMean calibration histograms of mean from CCDB
+  /// \param objectSigma calibration histograms of sigma from CCDB
+  /// \param track is the track
+  /// \param chargePar is the PID species
+  /// \return the corrected Nsigma value for the PID species
+  template <typename P, typename T>
+  double GetPostTPCCalib(const P& objectMean, const P& objectSigma, const T& track, const int chargePar)
```
vkucera, line 255:
```suggestion
  double getTPCPostCalib(const P& objectMean, const P& objectSigma, const T& track, const int chargePar)
```
It is not clear from the names that `objectMean` and `objectSigma` are histograms.
Also what is the point of templating the histogram type if the function can digest only TH3F?

EventFiltering/PWGHF/HFFilter.cxx
```diff
@@ -235,6 +244,48 @@ struct HfFilter { // Main struct for HF triggers
...
+  {
+    TH3F* calibMean = objectMean;
+    TH3F* calibSigma = objectSigma;
```
vkucera, line 258: Why is this needed? Can you not use `objectMean` and `objectSigma` directly?

EventFiltering/PWGHF/HFFilter.cxx
```diff
@@ -235,6 +244,48 @@ struct HfFilter { // Main struct for HF triggers
...
+
+    double kTPCncls = track.tpcNClsFound();
+    double kTPCPin = track.tpcInnerParam();
+    double kEta = track.eta();
+    double kTPCnSigma = 0.;
```
vkucera, line 263: These are not constants but variables and should be named accordingly.

EventFiltering/PWGHF/HFFilter.cxx
```diff
@@ -235,6 +244,48 @@ struct HfFilter { // Main struct for HF triggers
...
+
+    if (chargePar == kKa)
+      kTPCnSigma = track.tpcNSigmaKa();
+    if (chargePar == kPi)
+      kTPCnSigma = track.tpcNSigmaPi();
+    if (chargePar == kPr)
+      kTPCnSigma = track.tpcNSigmaPr();
```
vkucera, line 270: Use `else if` or `switch` for faster execution.

EventFiltering/PWGHF/HFFilter.cxx
```diff
@@ -235,6 +244,48 @@ struct HfFilter { // Main struct for HF triggers
...
+
+    int binTPCncls = calibMean->GetXaxis()->FindBin(kTPCncls);
+    binTPCncls = (binTPCncls == 0 ? 1 : binTPCncls);
+    binTPCncls = (binTPCncls > calibMean->GetXaxis()->GetNbins() ? calibMean->GetXaxis()->GetNbins() : binTPCncls);
+    int binPin = calibMean->GetYaxis()->FindBin(kTPCPin);
+    binPin = (binPin == 0 ? 1 : binPin);
+    binPin = (binPin > calibMean->GetYaxis()->GetNbins() ? calibMean->GetYaxis()->GetNbins() : binPin);
+    int binEta = calibMean->GetZaxis()->FindBin(kEta);
+    binEta = (binEta == 0 ? 1 : binEta);
+    binEta = (binEta > calibMean->GetZaxis()->GetNbins() ? calibMean->GetZaxis()->GetNbins() : binEta);
```
vkucera, line 280: You can use `std::min`, `std::max` instead of the conditions.

EventFiltering/PWGHF/HFFilter.cxx
```diff
@@ -235,6 +244,48 @@ struct HfFilter { // Main struct for HF triggers
...
+
+    double mean = calibMean->GetBinContent(binTPCncls, binPin, binEta);
+    double width = calibSigma->GetBinContent(binTPCncls, binPin, binEta);
+    double kTPCnSigma_Corr = (kTPCnSigma - mean) / width;
+
+    return kTPCnSigma_Corr;
```
vkucera, line 286: Name the variable properly.

EventFiltering/PWGHF/HFFilter.cxx
```diff
@@ -138,6 +138,9 @@ struct HfFilter { // Main struct for HF triggers
 
   // parameter for Optimisation Tree
   Configurable<bool> applyOptimisation{"applyOptimisation", false, "Flag to enable or disable optimisation"};
+  Configurable<bool> fConfigComputeTPCpostCalib{"cfgTPCpostCalib", false, "If true, compute TPC post-calibrated n-sigmas"};
```
vkucera, line 141:
```suggestion
  Configurable<bool> computeTPCPostCalib{"computeTPCPostCalib", false, "If true, compute TPC post-calibrated n-sigmas"};
```

EventFiltering/PWGHF/HFFilter.cxx
```diff
@@ -325,6 +376,10 @@ struct HfFilter { // Main struct for HF triggers
     float NSigmaTOF = track.tofNSigmaPr();
     float NSigma;
 
+    if (fConfigComputeTPCpostCalib) {
+      NSigmaTPC = GetPostTPCCalib(mapProtonMean, mapProtonSigma, track, kPr);
+    }
```
vkucera, line 381:
```suggestion
    if (fConfigComputeTPCpostCalib) {
      nSigmaTPC = GetPostTPCCalib(mapProtonMean, mapProtonSigma, track, kPr);
    }
```

vkucera (changes requested):

EventFiltering/PWGHF/HFFilter.cxx
zhangbiao-phy: > Why is this needed? Can you not use `objectMean` and `objectSigma` directly?

Yes, you are right! I know, I know, I hate myself. Thanks a lot!

EventFiltering/PWGHF/HFFilter.cxx
zhangbiao-phy: I would keep it here for the moment.

vkucera: I mean
```c++
(binTPCncls > calibMean->GetXaxis()->GetNbins() ? calibMean->GetXaxis()->GetNbins() : binTPCncls)
```
is equivalent to 
```c++
std::min(calibMean->GetXaxis()->GetNbins(), binTPCncls)
```
which is much easier to read and less prone to errors.

EventFiltering/PWGHF/HFFilter.cxx
vkucera: OK, please make one then.

EventFiltering/PWGHF/HFFilter.cxx
vkucera: Please fix the names.

EventFiltering/PWGHF/HFFilter.cxx
vkucera: Please fix the names.

EventFiltering/PWGHF/HFFilter.cxx
```diff
@@ -157,6 +160,12 @@ struct HfFilter { // Main struct for HF triggers
   std::array<std::shared_ptr<TH1>, kNCharmParticles> hBDTScorePrompt{};
   std::array<std::shared_ptr<TH1>, kNCharmParticles> hBDTScoreNonPrompt{};
 
+  // Histogram of TPC postcalibration map for pion and proton
+  TH3F* hMapPionMean;
+  TH3F* hMapPionSigma;
+  TH3F* hMapProtonMean;
+  TH3F* hMapProtonSigma;
```
vkucera, line 167: Initialise

EventFiltering/PWGHF/HFFilter.cxx
```diff
@@ -679,6 +749,19 @@ struct HfFilter { // Main struct for HF triggers
...
 
+    auto bc = collision.template bc_as<aod::BCsWithTimestamps>();
+    if (computeTPCPostCalib && currentRun != bc.runNumber()) {
+
+      auto calibList = ccdb->getForTimeStamp<TList>(fConfigCcdbPathTPC.value, bc.timestamp());
+
+      hMapPionMean = (TH3F*)calibList->FindObject("mean_map_pion");
+      hMapPionSigma = (TH3F*)calibList->FindObject("sigma_map_pion");
+      hMapProtonMean = (TH3F*)calibList->FindObject("mean_map_proton");
+      hMapProtonSigma = (TH3F*)calibList->FindObject("sigma_map_proton");
```
vkucera, line 760: You should check that all these are found and throw a fatal otherwise.

vkucera (changes requested):

EventFiltering/PWGHF/HFFilter.cxx
```diff
@@ -235,6 +244,48 @@ struct HfFilter { // Main struct for HF triggers
...
+    auto kTPCNSigma = 0.;
+
+    if (pidSpecies == kKa)
+      kTPCNSigma = track.tpcNSigmaKa();
+    else if (pidSpecies == kPi)
+      kTPCNSigma = track.tpcNSigmaPi();
+    else if (pidSpecies == kPr)
+      kTPCNSigma = track.tpcNSigmaPr();
+    else
+      LOG(fatal) << "Wrong PID Species be selected, please check!";
```
vkucera, line 270: MIssing braces

EventFiltering/PWGHF/HFFilter.cxx
```diff
@@ -161,10 +161,10 @@ struct HfFilter { // Main struct for HF triggers
...
 
   // Histogram of TPC postcalibration map for pion and proton
-  TH3F* hMapPionMean;
-  TH3F* hMapPionSigma;
-  TH3F* hMapProtonMean;
-  TH3F* hMapProtonSigma;
+  TH3F* hMapPionMean = new TH3F("hMapPionMean", "Map of TPC pion mean; TPCncls; Pin; Eta", 160, 0., 160, 100, 0., 5, 100, -0.8, 0.8);
+  TH3F* hMapPionSigma = new TH3F("hMapPionSigma", "Map of TPC pion sigma; TPCncls; Pin; Eta", 160, 0., 160, 100, 0., 5, 100, -0.8, 0.8);
+  TH3F* hMapProtonMean = new TH3F("hMapProtonMean", "Map of TPC proton mean; TPCncls; Pin; Eta", 160, 0., 160, 100, 0., 5, 100, -0.8, 0.8);
+  TH3F* hMapProtonSigma = new TH3F("hMapProtonSigma", "Map of TPC proton Sigma; TPCncls; Pin; Eta", 160, 0., 160, 100, 0., 5, 100, -0.8, 0.8);
```
vkucera, line 167: You have to initialise with `nullptr`.
What you do here instead is a memory leak, where you create new objects that will be left behind in memory later on when you load the histograms from files with `calibList->FindObject`.

EventFiltering/PWGHF/HFFilter.cxx
```diff
@@ -255,35 +255,35 @@ struct HfFilter { // Main struct for HF triggers
   double getTPCPostCalib(const TH3F* hCalibMean, const TH3F* hCalibSigma, const T& track, const int pidSpecies)
   {
 
-    auto kTPCncls = track.tpcNClsFound();
+    auto kTPCNCls = track.tpcNClsFound();
     auto kTPCPin = track.tpcInnerParam();
     auto kEta = track.eta();
-    auto kTPCnSigma = 0.;
+    auto kTPCNSigma = 0.;
```
vkucera, line 261: Maybe I wasn't clear. The problem here is with the prefix `k` which is reserved for constant named `enum` values (e.g. `kNCharmParticles`, `kPi`).
Wat you have here are variables which should not have the `k` as a prefix.
You can use e.g. `tpcNClsFound`, `etaTrack`,...

EventFiltering/PWGHF/HFFilter.cxx
```diff
@@ -255,35 +255,35 @@ struct HfFilter { // Main struct for HF triggers
...
 
-    auto mean = hCalibMean->GetBinContent(binTPCncls, binPin, binEta);
-    auto width = hCalibSigma->GetBinContent(binTPCncls, binPin, binEta);
-    auto kTPCnSigma_Corr = (kTPCnSigma - mean) / width;
+    auto mean = hCalibMean->GetBinContent(binTPCNCls, binPin, binEta);
+    auto width = hCalibSigma->GetBinContent(binTPCNCls, binPin, binEta);
+    auto kTPCNSigmaCorr = (kTPCNSigma - mean) / width;
 
-    return kTPCnSigma_Corr;
+    return kTPCNSigmaCorr;
```
vkucera, line 286: Same problem with the `k` prefix.

EventFiltering/PWGHF/HFFilter.cxx
```diff
@@ -752,12 +752,17 @@ struct HfFilter { // Main struct for HF triggers
...
-      hMapProtonSigma = (TH3F*)calibList->FindObject("sigma_map_proton");
+      if (!calibList->FindObject("mean_map_pion") || !calibList->FindObject("sigma_map_pion") || !calibList->FindObject("mean_map_proton") || !calibList->FindObject("sigma_map_proton"))
+        LOG(fatal) << "Can not find histograms!";
+      else {
+
+        hMapPionMean = (TH3F*)calibList->FindObject("mean_map_pion");
+        hMapPionSigma = (TH3F*)calibList->FindObject("sigma_map_pion");
+        hMapProtonMean = (TH3F*)calibList->FindObject("mean_map_proton");
+        hMapProtonSigma = (TH3F*)calibList->FindObject("sigma_map_proton");
+      }
```
vkucera, line 765: You are loading everything twice there unnecessarily. First load all the histograms and then check the validity of the pointers `hMapPionMean`,...
Also you should check that `calibList` right after loading it.

vkucera (changes requested):

EventFiltering/PWGHF/HFFilter.cxx
vkucera: Still missing

EventFiltering/PWGHF/HFFilter.cxx
```diff
@@ -679,6 +749,32 @@ struct HfFilter { // Main struct for HF triggers
...
+
+      if (!hMapPionMean || !hMapPionSigma || !hMapProtonMean || !hMapProtonSigma) {
+        LOG(fatal) << "Can not find histograms!";
+      } else {
+
+        hMapPionMean = (TH3F*)calibList->FindObject("mean_map_pion");
+        hMapPionSigma = (TH3F*)calibList->FindObject("sigma_map_pion");
+        hMapProtonMean = (TH3F*)calibList->FindObject("mean_map_proton");
+        hMapProtonSigma = (TH3F*)calibList->FindObject("sigma_map_proton");
+      }
```
vkucera, line 773: Why do you load all again?

vkucera (changes requested):

EventFiltering/PWGHF/HFFilter.cxx
```diff
@@ -235,6 +244,48 @@ struct HfFilter { // Main struct for HF triggers
...
+    binPin = std::min(hCalibMean->GetYaxis()->GetNbins(), binPin);
+    auto binEta = hCalibMean->GetZaxis()->FindBin(eta);
+    binEta = (binEta == 0 ? 1 : binEta);
+    binEta = std::min(hCalibMean->GetZaxis()->GetNbins(), binEta);
+
+    auto mean = hCalibMean->GetBinContent(binTPCNCls, binPin, binEta);
+    auto width = hCalibSigma->GetBinContent(binTPCNCls, binPin, binEta);
+    auto tpcNSigmaCorr = (tpcNSigma - mean) / width;
+
+    return tpcNSigmaCorr;
```
vkucera, line 286: You can directly return the expression `(tpcNSigma - mean) / width`.

vkucera (changes requested):

EventFiltering/PWGHF/HFFilter.cxx
zhangbiao-phy: > Why do you load all again?
my mistake

vkucera (approved):

vkucera:
Thanks for the fixes.

PR #1771, opened 2023-01-12
move to o2 physics constants

ddobrigk:


chiarapinto (approved):

ddobrigk:
This changes from using RecoDecay mass queries to the more lightweight o2::constants::physics functionality to reduce memory consumption

PR #1773, opened 2023-01-12
PWGHF: fix ccdb initialization for B field.

mfaggin:


mfaggin:
@fcolamar @cterrevo @ginnocen @vkucera this PR should fix the issue with CCDB that was creating giant headaches since the O2 tag of 11th January. From my tests, now this should fix it.

Please approve it asap so that also the code on Hyperloop can get back to work soon.

fcolamar (approved):

fcolamar:
Ciao Mattia, I'm approving the PR, thanks for the big debugging work of these two days :)

vkucera:
Thanks a lot @mfaggin for this tedious works.
I see that you moved the `ccdb` lines out of the the `if (doPvRefit) {` condition block in the `HfTrackIndexSkimCreator` task while it stays inside the block in the `HfTrackIndexSkimCreatorTagSelTracks` task. Is that intended?

mfaggin:
> Thanks a lot @mfaggin for this tedious works. I see that you moved the `ccdb` lines out of the the `if (doPvRefit) {` condition block in the `HfTrackIndexSkimCreator` task while they stay inside the block in the `HfTrackIndexSkimCreatorTagSelTracks` task. Is that intended?

Hi @vkucera yes, it is intended for the following reason: in `HfTrackIndexSkimCreator` the `ccdb` object must be always correctly initialized, because from it we retrieve the value of the B-field to be used by the dcaFitters. In the case of `HfTrackIndexSkimCreatorTagSelTracks`, instead, the `ccdb` object should matter only for the PV refit and, in particular, when repropagating the track to the refitted PV.
From my tests with `doPvRefit==false` everything seems to work now. The interesting thing is that eveything was working w/o this changes before the O2 tag nightly-20230110-1.
I'm in contact with @davidrohr and @ktf to understand better why.

vkucera:
> > Thanks a lot @mfaggin for this tedious works. I see that you moved the `ccdb` lines out of the the `if (doPvRefit) {` condition block in the `HfTrackIndexSkimCreator` task while they stay inside the block in the `HfTrackIndexSkimCreatorTagSelTracks` task. Is that intended?
> 
> Hi @vkucera yes, it is intended for the following reason: in `HfTrackIndexSkimCreator` the `ccdb` object must be always correctly initialized, because from it we retrieve the value of the B-field to be used by the dcaFitters. In the case of `HfTrackIndexSkimCreatorTagSelTracks`, instead, the `ccdb` object should matter only for the PV refit and, in particular, when repropagating the track to the refitted PV. From my tests with `doPvRefit==false` everything seems to work now. The interesting thing is that eveything was working w/o this changes before the O2 tag nightly-20230110-1. I'm in contact with @davidrohr and @ktf to understand better why.

Ok, thanks for the detailed explanation.

PR #1777, opened 2023-01-13
PWGDQ: Introduce variable binning for TOF and TPC PID hists

feisenhu:
I have introduced a new sub-group for TOF and TPC PID histograms with a finer variable binning.
- tpcpid_fine
- tofpid_fine
If those are not used the histograms have the standard fixed binning.


iarsene (approved):

iarsene:
@feisenhu 
Thanks a lot for fixing the warnings in the histogram manager

feisenhu:
I have missed using the brackets after each cast function. This should be fixed now.

PR #1782, opened 2023-01-13
do not load TGeo if using LUT or no correction

ddobrigk:
In this PR, the builders have been changed such that the TGeoVolume etc loading only happens if the TGeo correction is requested and LUT loading only happens if the LUT correction is requested. This will be tested just in case before going ahead, to be sure all is fine


ddobrigk:
Tested and all good to go ahead!

njacazio (approved):

PR #1784, opened 2023-01-13
remove loading of full geometry (not req. by LUT)

ddobrigk:
This will remove the unnecessary loading of the full ALICE geometry in the track propagator (not required when propagating with the LUT approach). It will still be tested before going ahead


ddobrigk:
Testing concluded, results are valid -> ready to review

alibuild (approved):

ddobrigk:
Approving, will lead to a significant reduction of memory usage of the task (and will not have any impact on any results). Confirmed with Ruben that the O2 base propagator does not require full geometry in case the LUT option is used. In the past (2021) the propagator actually checked (unnecessarily) if the geo was initialized, but this is no longer the case.

PR #1791, opened 2023-01-16
WIP: PWG-HF finding with collision-to-track association for ambigous tracks

jgrosseo:


PWGHF/DataModel/CandidateReconstructionTables.h
```diff
@@ -40,18 +40,34 @@ DECLARE_SOA_TABLE(HfSelCollision, "AOD", "HFSELCOLLISION", //!
...
                   hf_sel_track::PzProng);
 
+namespace hf_track_association
+{
+DECLARE_SOA_INDEX_COLUMN(Collision, collision);                      //! Collision index
+DECLARE_SOA_INDEX_COLUMN(Track, track);                              //! Track index
+DECLARE_SOA_SELF_ARRAY_INDEX_COLUMN(CompatibleColl, compatibleColl); //! Track index
+} // namespace hf_track_association
+
+DECLARE_SOA_TABLE(HfTrackAssoc, "AOD", "HFTRACKASSOC", //! Table for track-to-collision association for HF vertex finding - tracks can appear for several collisions
```
jgrosseo, line 64:
```suggestion
DECLARE_SOA_TABLE(HfTrackAssoc, "AOD", "HFTRACKASSOC", //! Table for collision-to-track association for HF vertex finding - tracks can appear for several collisions
```

PWGHF/DataModel/CandidateReconstructionTables.h
```diff
@@ -40,18 +40,34 @@ DECLARE_SOA_TABLE(HfSelCollision, "AOD", "HFSELCOLLISION", //!
...
+                  hf_track_association::CollisionId,
+                  hf_track_association::TrackId);
+
+DECLARE_SOA_TABLE(HfTrackCompColls, "AOD", "HFTRKCOMPCOLLS", //! Table with vectors of collision indices stored per track
```
jgrosseo, line 68: I'd call it HfCompColls so that we stay within the 14 char limit. Otherwise we can never store this table on disk (in case we want to)

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -1452,667 +1486,748 @@ struct HfTrackIndexSkimCreator {
...
-    bool cutStatus2Prong[n2ProngDecays][nCuts2Prong];
-    bool cutStatus3Prong[n3ProngDecays][nCuts3Prong];
-    int nCutStatus2ProngBit = BIT(nCuts2Prong) - 1; // bit value for selection status for each 2-prong candidate where each selection is one bit and they are all set to 1
-    int nCutStatus3ProngBit = BIT(nCuts3Prong) - 1; // bit value for selection status for each 3-prong candidate where each selection is one bit and they are all set to 1
+        // // retrieve the selection flag that corresponds to this collision
+        int isSelProngPos1 = trackPos1.isSelProng();
+        for (auto iCollId{0u}; iCollId < trackPos1.compatibleCollIds().size(); ++iCollId) {
+          if (trackPos1.compatibleCollIds()[iCollId] == collision.globalIndex()) {
+            isSelProngPos1 = trackPos1.isSelProngAllColls()[iCollId];
+            break;
```
jgrosseo, line 1586: I think it is fine and cannot be done faster for such a short list
(we would need to keep an iterator per track if we want to cash the previous position (the collisions are ordered)). I don't think that is worth it.

jgrosseo: Maybe we can make a lambda, to avoid the copy-paste with the inner track loop

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -2230,193 +2350,207 @@ struct HfTrackIndexSkimCreatorCascades {
...
+        // selections on the bachelor
+
+        // // retrieve the selection flag that corresponds to this collision
+        int isSelProngBach = bach.isSelProng();
+        for (auto iCollId{0u}; iCollId < bach.compatibleCollIds().size(); ++iCollId) {
+          if (bach.compatibleCollIds()[iCollId] == collision.globalIndex()) {
+            isSelProngBach = bach.isSelProngAllColls()[iCollId];
+            break;
+          }
+        }
```
jgrosseo, line 2403: Yes, a lambda would be good :)

PWGHF/TableProducer/trackToCollisionAssociator.cxx
```diff
@@ -0,0 +1,148 @@
...
+  void processAssocWithTime(Collisions const& collisions,
+                            TracksWithSel const& tracksUnfiltered,
+                            soa::Filtered<TracksWithSel> const& tracks,
+                            BCs const& bcs)
+  {
+    std::vector<int>** collsPerTrack = new std::vector<int>*[tracksUnfiltered.size()];
+    memset(collsPerTrack, 0x0, sizeof(collsPerTrack) * tracksUnfiltered.size());
+
+    // loop over collisions to find time-compatible tracks
+    float nSigmaForTimeCompat2 = nSigmaForTimeCompat * nSigmaForTimeCompat;
```
jgrosseo, line 47:
```suggestion
    const float nSigmaForTimeCompat2 = nSigmaForTimeCompat * nSigmaForTimeCompat;
```

PWGHF/TableProducer/trackToCollisionAssociator.cxx
```diff
@@ -0,0 +1,148 @@
...
+    auto trackBegin = tracks.begin();
+    const auto bOffsetMax = 241;
+    for (const auto& collision : collisions) {
+      float collTime = collision.collisionTime();
+      float collTimeRes2 = collision.collisionTimeRes() * collision.collisionTimeRes();
```
jgrosseo, line 52:
```suggestion
      const float collTime = collision.collisionTime();
      const float collTimeRes2 = collision.collisionTimeRes() * collision.collisionTimeRes();
```

PWGHF/TableProducer/trackToCollisionAssociator.cxx
```diff
@@ -0,0 +1,148 @@
...
+      uint64_t collBC = collision.bc().globalBC();
+      bool iteratorMoved = false;
+      for (auto track = trackBegin; track != tracks.end(); ++track) {
+        if (!track.has_collision()) {
+          continue;
+        }
+        int64_t bcOffset = (int64_t)track.collision().bc().globalBC() - (int64_t)collBC;
```
jgrosseo, line 59:
```suggestion
        const int64_t bcOffset = (int64_t)track.collision().bc().globalBC() - (int64_t)collBC;
```

PWGHF/TableProducer/trackToCollisionAssociator.cxx
```diff
@@ -0,0 +1,148 @@
...
+        float trackTime{0.};
+        float trackTimeRes2{0.};
+        if (track.isPVContributor()) {
+          trackTime = track.collision().collisionTime(); // if PV contributor, we assume the time to be the one of the collision
+          trackTimeRes2 = 25.f;                          // 1 BC
```
jgrosseo, line 64: I think this is not correct because it should be `25 * 25`

PWGHF/TableProducer/trackToCollisionAssociator.cxx
```diff
@@ -0,0 +1,148 @@
...
+        LOGP(debug, "collision time={}, collision time res={}, track time={}, track time res={}, bc collision={}, bc track={}, delta time={}", collTime, collision.collisionTimeRes(), track.trackTime(), track.trackTimeRes(), collBC, track.collision().bc().globalBC(), deltaTime);
+        if (!iteratorMoved && bcOffset > -bOffsetMax) {
+          trackBegin.setCursor(track.filteredIndex());
+          iteratorMoved = true;
+          LOGP(debug, "Moving iterator begin {}", track.globalIndex());
+        } else if (bcOffset > bOffsetMax) {
+          LOGP(debug, "Stopping iterator {}", track.globalIndex());
+          break;
+        }
+        if (deltaTime * deltaTime < nSigmaForTimeCompat2 * sigmaTimeRes2) {
```
jgrosseo, line 80: As discussed in case of the 25, it should use 1 sigma here. So one needs a case. This case should also be used if the flag `TrackTimeResIsRange` is set 
https://github.com/AliceO2Group/AliceO2/blob/dev/Framework/Core/include/Framework/DataTypes.h

PWGHF/TableProducer/trackToCollisionAssociator.cxx
```diff
@@ -0,0 +1,148 @@
...
+        float trackTime{0.};
+        float trackTimeRes2{0.};
+        if (track.isPVContributor()) {
+          trackTime = track.collision().collisionTime(); // if PV contributor, we assume the time to be the one of the collision
+          trackTimeRes2 = 25.f;                          // 1 BC
+        } else {
+          trackTime = track.trackTime();
+          trackTimeRes2 = track.trackTimeRes() * track.trackTimeRes();
+        }
+        float deltaTime = trackTime - collTime + bcOffset * 25.f;
```
jgrosseo, line 69:
```suggestion
        const float deltaTime = trackTime - collTime + bcOffset * 25.f;
```

PWGHF/TableProducer/trackToCollisionAssociator.cxx
```diff
@@ -0,0 +1,148 @@
...
+        float sigmaTimeRes2 = collTimeRes2 + trackTimeRes2;
+        LOGP(debug, "collision time={}, collision time res={}, track time={}, track time res={}, bc collision={}, bc track={}, delta time={}", collTime, collision.collisionTimeRes(), track.trackTime(), track.trackTimeRes(), collBC, track.collision().bc().globalBC(), deltaTime);
+        if (!iteratorMoved && bcOffset > -bOffsetMax) {
```
jgrosseo, line 72: Add:
`// optimization to avoid looping over the full track list each time. This assumes that tracks are sorted by BCs (which they should be because collisions are sorted by BCs)

PWGHF/TableProducer/trackToCollisionAssociator.cxx
```diff
@@ -0,0 +1,148 @@
...
+          trackBegin.setCursor(track.filteredIndex());
+          iteratorMoved = true;
+          LOGP(debug, "Moving iterator begin {}", track.globalIndex());
+        } else if (bcOffset > bOffsetMax) {
+          LOGP(debug, "Stopping iterator {}", track.globalIndex());
+          break;
+        }
+        if (deltaTime * deltaTime < nSigmaForTimeCompat2 * sigmaTimeRes2) {
+          auto collIdx = collision.globalIndex();
+          auto trackIdx = track.globalIndex();
```
jgrosseo, line 82:
```suggestion
          const auto collIdx = collision.globalIndex();
          const auto trackIdx = track.globalIndex();
```

PWGHF/TableProducer/trackToCollisionAssociator.cxx
```diff
@@ -0,0 +1,148 @@
...
+    for (const auto& track : tracksUnfiltered) {
+      LOGP(debug, "Track id {}", track.globalIndex());
+      if (collsPerTrack[track.globalIndex()] == nullptr) {
+        reverseIndices(empty);
+      } else {
+        // for (const auto& collId : *collsPerTrack[track.globalIndex()]) {
+        //   LOGP(info, "  -> Coll id {}", collId);
+        // }
+        reverseIndices(*collsPerTrack[track.globalIndex()]);
+      }
```
jgrosseo, line 104: We can do the delete already here

PWGHF/TableProducer/trackToCollisionAssociator.cxx
```diff
@@ -0,0 +1,148 @@
...
+
+  PROCESS_SWITCH(HfTrackToCollisionAssociation, processAssocWithTime, "Use track-to-collision association based on time", false);
+
+  Preslice<Tracks> perCollision = o2::aod::track::collisionId;
+
+  void processStandardAssoc(Collisions const& collisions,
+                            Tracks const& tracks)
+  {
+    for (const auto& collision : collisions) { // we do it for all tracks, to be compatible with Run2 analyses
+      uint64_t collIdx = collision.globalIndex();
```
jgrosseo, line 121:
```suggestion
      const uint64_t collIdx = collision.globalIndex();
```

PWGHF/TableProducer/trackIndexSkimCreator.cxx
aalkin: Why not use `std::find_if`?

PWGHF/TableProducer/trackToCollisionAssociator.cxx
```diff
@@ -0,0 +1,148 @@
...
+
+  using TracksWithSel = soa::Join<Tracks, TracksExtra, TrackSelection>;
+  Filter trackFilter = requireGlobalTrackWoDCAInFilter();
+
+  void processAssocWithTime(Collisions const& collisions,
+                            TracksWithSel const& tracksUnfiltered,
+                            soa::Filtered<TracksWithSel> const& tracks,
+                            BCs const& bcs)
+  {
+    std::vector<int>** collsPerTrack = new std::vector<int>*[tracksUnfiltered.size()];
```
aalkin, line 43: It would be better to just nest vectors: `std::vector<std::vector<int>>`, `new` should be avoided.

jgrosseo: But we only have it sparsely filled. Why avoid new?

aalkin: The vector of vectors can have empty vectors as entries, there is a negligible overhead. It is better to avoid direct memory allocations simply because it would protect from possible memory issues.

jgrosseo: Yes, I understand that we can have empty vectors but we need to allocate all of them and here we process the tracks table. Given that we have a large fraction empty (we only process tracks which are accepted by the track selection), I think the approach with allocation is better. It is a service task and we can make sure we don't produce leaks I think :)

aalkin: Then one can do `std::vector<std::unique_ptr<std::vector<int>>>`, it would be still be sparse but the allocations are better managed.

PWGHF/DataModel/CandidateReconstructionTables.h
fgrosa: Done

PWGHF/TableProducer/trackIndexSkimCreator.cxx
fgrosa: Done 😄

PWGHF/TableProducer/trackToCollisionAssociator.cxx
fgrosa: Good catch, thanks!

PWGHF/TableProducer/trackToCollisionAssociator.cxx
fgrosa: I removed the `delete` it given that I implemented Anton's suggestion with the `unique_ptr`

PWGHF/TableProducer/trackToCollisionAssociator.cxx
fgrosa: I have implemented it and I directly initialised it to the number of tracks in the definition (I checked and indeed it initialises each element to `nullptr`)

fgrosa:
Hi @vkucera @fcolamar @cterrevo @ginnocen @jgrosseo ,
This code is ready for review. We solved the issue of the extra CPU time in case of "standard association", which is now much below the one of the track-propagation task
```
track-propagation                                 : CPU(abs) = 40.910259,       CPU(perc) =    2022.12
hf-track-index-skim-creator-tag-sel-collisions    : CPU(abs) = 0.709226,       CPU(perc) =      41.69
track-selection                                   : CPU(abs) = 4.215740,       CPU(perc) =     217.64
hf-track-to-collision-association                 : CPU(abs) = 0.917257,       CPU(perc) =      51.92
hf-track-index-skim-creator-tag-sel-tracks        : CPU(abs) = 4.608081,       CPU(perc) =     245.52
hf-track-index-skim-creator                       : CPU(abs) = 13.762280,       CPU(perc) =     685.17
hf-track-index-skim-creator-cascades              : CPU(abs) = 0.459317,       CPU(perc) =      30.60
```
while with the previous version was significantly higher. Together with @mfaggin we checked that the output is expected, for all the options. We will give a dedicated presentation in one of the next HF O2 meetings.

In particular, if one wants to keep the old behaviour of the code, it is sufficient to use `processStandardAssoc`, which is the default. Currently the `processAssocWithTime` is meant only for the HFFilters (and is quite urgent, as you already know).

Once merged it is important to propagate to all the HF analysers that the `o2-analysis-hf-track-to-collision-associator` workflow has to be added everytime the `o2-analysis-hf-track-index-skim-creator` is executed.

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -1033,20 +1057,33 @@ struct HfTrackIndexSkimCreator {
...
+  template <typename T, typename U>
+  void getPxPyPz(T const& trackPars, U& pVec)
+  {
+    auto pt = 1.f / std::abs(trackPars.getQ2Pt());
+    float cs = cosf(trackPars.getAlpha()), sn = sinf(trackPars.getAlpha());
+    auto r = std::sqrt((1.f - trackPars.getSnp()) * (1.f + trackPars.getSnp()));
+    pVec[0] = pt * (r * cs - trackPars.getSnp() * sn);
+    pVec[1] = pt * (trackPars.getSnp() * cs + r * sn);
+    pVec[2] = pt * trackPars.getTgl();
+  }
```
vkucera, line 1075: This is a nice candidate for addition in [trackUtilities.h](https://github.com/AliceO2Group/O2Physics/blob/master/Common/Core/trackUtilities.h).

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -1033,20 +1057,33 @@ struct HfTrackIndexSkimCreator {
...
     lut = o2::base::MatLayerCylSet::rectifyPtrFromFile(ccdb->get<o2::base::MatLayerCylSet>(ccdbPathLut));
-    // if (!o2::base::GeometryManager::isGeometryLoaded()) {
-    //   ccdb->get<TGeoManager>(ccdbPathGeo);
-    // }
     runNumber = 0;
   }
 
+  /// Method to compute px, py, and px from the track parameters
+  /// \param trackPars are the track parameters
+  /// \param hfTrack1 is the momentum array to be filled
```
vkucera, line 1065:
```suggestion
  /// \param pVec is the momentum array to be filled
```

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -1033,20 +1057,33 @@ struct HfTrackIndexSkimCreator {
     ccdb->setCaching(true);
     ccdb->setLocalObjectValidityChecking();
     lut = o2::base::MatLayerCylSet::rectifyPtrFromFile(ccdb->get<o2::base::MatLayerCylSet>(ccdbPathLut));
-    // if (!o2::base::GeometryManager::isGeometryLoaded()) {
-    //   ccdb->get<TGeoManager>(ccdbPathGeo);
-    // }
     runNumber = 0;
   }
 
+  /// Method to compute px, py, and px from the track parameters
```
vkucera, line 1063:
```suggestion
  /// Method to compute px, py, pz from the track parameters
```

PWGHF/DataModel/CandidateReconstructionTables.h
```diff
@@ -41,16 +41,29 @@ DECLARE_SOA_TABLE(HfSelCollision, "AOD", "HFSELCOLLISION", //!
...
+namespace hf_track_association
+{
+
+enum eTrackType { Regular = 0,
+                  Ambiguous,
+                  PVContributor };
+
+DECLARE_SOA_INDEX_COLUMN(Collision, collision); //! Collision index
+DECLARE_SOA_INDEX_COLUMN(Track, track);         //! Track index
+DECLARE_SOA_COLUMN(TrackType, trackType, int);  //! Track type
```
vkucera, line 58: Could `TrackType` be a smaller type than `int`?

PWGHF/DataModel/CandidateReconstructionTables.h
```diff
@@ -41,16 +41,29 @@ DECLARE_SOA_TABLE(HfSelCollision, "AOD", "HFSELCOLLISION", //!
...
-                  hf_sel_track::IsSelProng,
-                  hf_sel_track::PxProng,
-                  hf_sel_track::PyProng,
-                  hf_sel_track::PzProng);
+                  hf_sel_track::IsSelProng);
+
+namespace hf_track_association
+{
+
+enum eTrackType { Regular = 0,
```
vkucera, line 52:
```suggestion
enum TrackType { Regular = 0,
```

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -380,21 +378,17 @@ struct HfTrackIndexSkimCreatorTagSelTracks {
...
   }
 
   /// Single-track cuts for 2-prongs or 3-prongs
-  /// \param hfTrack is a track
+  /// \param trackPt is the track pt
   /// \param dca is a 2-element array with dca in transverse and longitudinal directions
   /// \return true if track passes all cuts
-  template <typename T>
-  bool isSelectedTrack(const T& hfTrack, const array<float, 2>& dca, const int candType)
+  bool isSelectedTrackDCA(const float& trackPt, const array<float, 2>& dca, const int candType)
```
vkucera, line 389: Please add documentation of `candType`.

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -408,6 +402,184 @@ struct HfTrackIndexSkimCreatorTagSelTracks {
     return true;
   }
 
+  /// Single-track cuts for 2-prongs or 3-prongs
+  /// \param hfTrack is a track
+  /// \param trackPt is the track pt
+  /// \param trackEta is the track eta
+  /// \param dca is a 2-element array with dca in transverse and longitudinal directions
+  /// \param statusProng is the selection flag
+  /// \return true if track passes all cuts
```
vkucera, line 411:
```suggestion
```

PWGHF/DataModel/CandidateReconstructionTables.h
vkucera: Does this collide with the `TrackType` column name?

PWGHF/TableProducer/trackIndexSkimCreator.cxx
fgrosa: Done!

PWGHF/DataModel/CandidateReconstructionTables.h
fgrosa: an `int8_t` is enough, modified

PWGHF/TableProducer/trackIndexSkimCreator.cxx
fgrosa: Done

vkucera (approved):

ddobrigk (approved):
@fgrosa thanks for the extra utility, looks useful! I have some other comments but those are on unrelated parts so I'll leave that for another moment. Thanks again

fgrosa:
> @fgrosa thanks for the extra utility, looks useful! I have some other comments but those are on unrelated parts so I'll leave that for another moment. Thanks again

Thanks to you!

PR #1793, opened 2023-01-16
Adding configurable to use the TPC post calibration from the CCDB

vfeuilla:
Adding a configurable to allow the possibility to use the postcalibrated maps for the TPC once they are uploaded in the CCDB. Current location for those maps is set  to "Users/i/iarsene/Calib/TPCpostCalib"


iarsene (approved):

iarsene (approved):

iarsene (dismissed):

vfeuilla:
Hi @iarsene 
I figured out what the issue was :
- To use bc_as I need to use the `MyEvents::iterator const& collision`
- If I use `MyEvents::iterator const& collision` instead of `MyEvents const& collision` then I cannot use the `collision_as` to determine if 2 tracks belong to the same event, so I removed this part and put `VarManager::FillEvent<TEventFillMap>(collision);` out of the track loop
But if I use the iterator, then I don't need to do this check anymore, is that correct? So everything should be fine now

PR #1796, opened 2023-01-17
PWGDQ: adding histograms for QA

lucamicheletti93:
Adding histograms for alignment studies


PWGDQ/Core/HistogramsLibrary.h
```diff
@@ -339,6 +339,8 @@ void o2::aod::dqhistograms::DefineHistograms(HistogramManager* hm, const char* h
       hm->AddHistogram(histClass, "Mass_Rapidity", "", false, 750, 0.0, 15.0, VarManager::kMass, 200, 2.5, 4.0, VarManager::kRap);
       hm->AddHistogram(histClass, "Mass_VtxZ", "", true, 30, -15.0, 15.0, VarManager::kVtxZ, 750, 0.0, 15.0, VarManager::kMass);
       hm->AddHistogram(histClass, "cosThetaHE", "", false, 100, -1., 1., VarManager::kCosThetaHE);
+      hm->AddHistogram(histClass, "DeltaPtotTracks", "", false, 2000, -100., 100., VarManager::kDeltaPtotTracks);
+      hm->AddHistogram(histClass, "Mass_DeltaPtotTracks", "", false, 750, 0.0, 15.0, VarManager::kMass, 2000, -100., 100., VarManager::kDeltaPtotTracks);
```
iarsene, line 343: This 2D histogram is using about 2 million bins, which, multiplied with the number of histogram directories in which it might be replicated, can lead to a significant amount of memory usage. Ofcourse that if you need it like this you should have it. But otherwise, my suggestion is to decrease the amount of bins to the level you really need. This can be done by either making coarser bins everywhere, or, if you need fine bins in some regions and can be coarser in other, you can always create histograms with variable bin widths.

iarsene (approved):

PR #1800, opened 2023-01-18
Allow O2Physics to compile using GCC 12.X.

ktf:


ktf:
This PR fixes some unintialised variables in [PWGCF/FemtoUniverse/Core/FemtoUniverseObjectSelection.h](https://github.com/AliceO2Group/O2Physics/pull/1800#diff-cba98f07d0dce3cc03f7c4c0949347da3ba48fe02bc0d3e1e5e9bba4af64ef58). Can the authors please check this does not affect results?

saganatt:
Thanks for fixing the double copy in femto codes. I think Zuzanna could check later when she has time.

Just for the other analyzers' information the initialization to check is in [PWGEM/Dilepton/Tasks/lmeeLFCocktail.cxx](https://github.com/AliceO2Group/O2Physics/pull/1800/files#diff-b1927bb8e47c02f180aaf7390e1d57bae9ea39ae917dc76538e101e60322a306).

ivorobye:
Thanks for the fix, @DanielSamitz can you have a quick look at the changes in LF cocktail file? I think it's ok to initialise the DCA variables with 0, but maybe some other value like -999 can be also considered.

DanielSamitz:
I don't see a problem with initializing DCA variables with 0 (negative values would be an option, but then they enter quadratically in the pairDCA, so it wouldn't help there). OK for me.

PR #1803, opened 2023-01-18
PWGMM: dndeta: reduce code duplication

aalkin:
Using `Join::contains<>()` code for inclusive and centrality-binned versions is unified. 

Depends on https://github.com/AliceO2Group/AliceO2/pull/10578.

@ddobrigk as discussed


alibuild (approved):

PR #1808, opened 2023-01-19
DPG AOT: add efficiency without collision grouping

njacazio:


njacazio:
@rbailhac this is my attempt, a new process function will be available that skips the collisions

DPG/Tasks/AOTTrack/qaEfficiency.cxx
```diff
@@ -591,9 +590,12 @@ struct QaEfficiency {
 
   void initMC(const AxisSpec& axisSel)
   {
-    if (!doprocessMC) {
+    if (!doprocessMC && !doprocessMCWithoutCollisions) {
```
mfaggin, line 593: I'd add also here a `LOG(fatal)`, so that one can understand immediately that he/she needs to switch on one of the `process` functions

mfaggin (approved):
It looks fine to me. The issue that I still se with this solution is that we risk to normalize our efficiency by MC particles that do not match with any of the reco collision we take tracks from, but as level 0 check in my opinion it's fine

alibuild (approved):

PR #1810, opened 2023-01-19
fix for mixed event counter

jgrosseo:


alibuild (approved):

PR #1811, opened 2023-01-19
[DPG] Add task for simple MC QC to be used on alibi testing

fgrosa:
@chiarazampolli @noferini this is a simple task to have AOD testing on alibi (https://alice.its.cern.ch/jira/browse/O2-3501). I think that it should be under the hood of the DPG, but I did not find an appropriate directory so I created a new one. Let me know if you think it should go somewhere else, I will  move it. Thanks!


noferini:
Hi @fgrosa ,
if I understood correctly the purpose of this task is to monitor specific observables and trigger alarms on alibi in case of strange behaviors.
"MC" looks to me a bit generic and I would not use it as a top directory inside DPG. Otherwise we should have also a "data" dir.
I would suggest to create a new dir with something which recalls what tasks inside focus on.
checks, monitor, ....
whatever you think matches better your purpose.
Then inside it no problem to have a "MC" dir
Cheers

fgrosa:
> Hi @fgrosa , if I understood correctly the purpose of this task is to monitor specific observables and trigger alarms on alibi in case of strange behaviors. "MC" looks to me a bit generic and I would not use it as a top directory inside DPG. Otherwise we should have also a "data" dir. I would suggest to create a new dir with something which recalls what tasks inside focus on. checks, monitor, .... whatever you think matches better your purpose. Then inside it no problem to have a "MC" dir Cheers

Hi @noferini your proposal makes very sense for me, I changed it accordingly! 👍

noferini (approved):

PR #1812, opened 2023-01-19
Tools KF Particle package

AnnalenaKa:
This header can be used by various PWGs when using the KF particle package in the analysis. 
It implements basic functionalities that were also used in AliPhysics.


ddobrigk (approved):
Thanks! Approved for further testing. Location in O2Physics is also reasonable

PR #1815, opened 2023-01-19
PWGHF: process_switch to enable/disable HfTrackIndexSkimCreatorCascades

DanielSamitz:
Suggestion for a `PROCESS_SWITCH` for HfTrackIndexSkimCreatorCascades, instead of the `--doCascades` option.

The motivation is to allow to enable the cascade task also on HyperLoop.

The empty dummy `processNoCascades` function seems to be necessary when switching off the main `processCascades`  (the workflow gets stuck without it).

In principle one could also build a similar process switch for HfTrackIndexSkimCreator (with default=true), to switch it off when running with HfTrackIndexSkimCreatorCascades instead.

I don't know how these changes should be transferred to https://github.com/AliceO2Group/Run3Analysisvalidation/blob/master/codeHF/workflows.yml


vkucera (approved):
Please propagate the configuration changes in the validation framework.

PR #1819, opened 2023-01-20
Even skipped collision require an entry for CEFP

mpuccio:
@aortizve, this should fix the issue with the CEFP.


aortizve:
Hello @mpuccio, thanks, but I already implemented a solution, see https://github.com/AliceO2Group/O2Physics/pull/1818/files

PR #1820, opened 2023-01-20
skipped events are assigned a value

aortizve:


alibuild (approved):

PR #1824, opened 2023-01-20
Always fill the trigger decision

mpuccio:
@ChiaraDeMartin95 @ercolessi: it is working now as the default vertex cut is at 20, but formally the code needs this change


alibuild (approved):

PR #1827, opened 2023-01-22
Separate labelers for memory reduction

ddobrigk:


ddobrigk:
@ChiaraDeMartin95 @ercolessi @njacazio this PR separates the label builders from the main v0/cascade builders, so that two less unnecessary DPL devices are spawned when doing analyses. This should save some further memory!

njacazio (approved):

PR #1829, opened 2023-01-23
Add @ChiaraDeMartin95 to the PWGLF triggers

mpuccio:


ddobrigk (approved):
very logical, of course, thanks!

PR #1831, opened 2023-01-24
PWGDQ: Add KFParticle to DQ tableReader

lupengzhong:


iarsene:
@lupengzhong 
So as we discussed in mattermost, we should refactor these changes. First I suggest that you just prepare changes to implement the calcution of vertexing using KF in the VarManager, as an alternative to the DCAFitter. After we commision this, we can then add eventually other changes, like the tables you want to produce in the tableReader.

PR #1834, opened 2023-01-24
Integration test task

ddobrigk:
Added a task to serve as Run 3 integration tester. Checks data model, collision and BC associated basics, and allows for basic output. Optionally includes also 2D correlations between track counters.


ddobrigk:
@jgrosseo this we could maybe use for the Run 3 integration test. It's a quick and minimal task that not only checks the sizes of all tables (a la check data model) but also checks some basic collisions/BC and tracks||v0s||cascades/BC... and optionally also plots basic correlations between Ntrack counters to spot any unusual changes in reconstruction quickly (or that's the hope). It could run on the nightlies. Any thoughts?

ddobrigk:
Will approve, further tuning can still happen as we move forward with the integration tests

alibuild (approved):

PR #1835, opened 2023-01-24
[EMCAL-733] Add histograms for time distribution in different BC phases

mfasDa:
Histogram cell time vs. cell abs ID for each of the 4
BC phases


mfasDa:
Including fixes for MegaLinter

ddobrigk (approved):

PR #1433, opened 2022-11-04
DPG: Add track sel QA task

njacazio:
- In track selection task add compatibility mode for tracks without IU

PR #1839, opened 2023-01-25
Add warning for global track selection setting.

mfaggin:
This PR just aims to dump at screen a few words to help the user understanding which global track selection set is used in the analysis. In particular:
-  some `LOG(info)` are added mentioning which function is called
-  a `LOG(warning)` is added for the case `itsMatching==0` and `isRun3==true`. By construction, since PR #1433 this configuration is equal to `itsMatching==1` and `isRun3==true`, but it was never explicitly mentioned.
Tagging @njacazio and @ChiaraDeMartin95 , just for their info

### DIFF
--- a/Common/TableProducer/trackselection.cxx
+++ b/Common/TableProducer/trackselection.cxx
@@ -60,24 +60,29 @@ struct TrackSelectionTask {
       case 0:
         // Run 2 SPD kAny
         if (!isRun3) {
+          LOG(info) << "setting up globalTracks = getGlobalTrackSelection();";
           globalTracks = getGlobalTrackSelection();
           break;
         }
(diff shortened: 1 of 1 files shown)


mfaggin:
@ddobrigk @jgrosseo @ktf @iarsene a genlte reminder

ktf:
I would use the "detail" level which is lower than info, but still gets always compiled in (contrary to debug). That said, maybe it's better to simply print statistics at the end?

mfaggin:
> I would use the "detail" level which is lower than info, but still gets always compiled in (contrary to debug). That said, maybe it's better to simply print statistics at the end?

hi @ktf what do you mean with `statistics`?

ktf:
Nevermind, I understood you were doing it per track.

mfaggin:
> I would use the "detail" level which is lower than info, but still gets always compiled in (contrary to debug). That said, maybe it's better to simply print statistics at the end?

about the `LOG(info)`: I think it has a negligible impact, since it's called just once per execution, but it helps the analyser to understand which cuts are really used. What do you think? Can we leave it?

ddobrigk (approved):
discussion seems ok, ready to go, approving. Thanks!

PR #1840, opened 2023-01-25
DPG: update PID qa

njacazio:
- Remove integrated charge if split is required
- Add TOF map


alibuild (approved):

PR #1848, opened 2023-01-26
PIDML producer: Drop backward compability with old data, protect MC cases from crash

saganatt:


alibuild (approved):

PR #1851, opened 2023-01-26
Improve strangeness builder preselectors

ddobrigk:
Based on needs for software triggering and the content shown at the [AOT meeting of the 12th](https://indico.cern.ch/event/1236818/contributions/5203341/attachments/2574614/4439297/DDChinellato-2023-StrangenessReconstructionInO2-02.pdf). 

This PR adds the possibility to pre-filter TPC Ncrossedrows at the preselector level, avoiding loops on candidates with ITS-only tracks if not required (and with the filtering being arrow-based). It also adds an extra flag such that TPC requirements (dE/dx and Ncrossedrows) can optionally only be applied to baryon/heavy daughters. To preserve standard operation, the new flag defaults to false; using it will make it possible to also explore strangeness measurements with the meson prong not having dE/dx information. @ChiaraDeMartin95 and @ercolessi : this is relevant for you :-D

@strogolo @mpuccio @jgrosseo this could potentially also help a bit with the CEFP performance but let's proceed cautiously. I will still investigate further before suggesting any change in those configurations, of course.

### DIFF
--- a/PWGLF/TableProducer/cascadebuilder.cxx
+++ b/PWGLF/TableProducer/cascadebuilder.cxx
@@ -753,7 +753,7 @@ struct cascadePreselector {
   }
   //*+-+*+-+*+-+*+-+*+-+*+-+*+-+*+-+*+-+*+-+*
   /// This process function ensures that all cascades are built. It will simply tag everything as true.
-  void processBuildAll(aod::Cascades const& cascades, TracksWithExtra const&)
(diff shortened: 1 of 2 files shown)


njacazio (dismissed):

ddobrigk:
Just realised a minor fix is still necessary. Will do it in a moment

njacazio (approved):

PR #1855, opened 2023-01-26
DPG: add collision numContrib in the derived data.

mfaggin:


alibuild (approved):

PR #1710, opened 2022-12-21
PWGHF: implement SigmaC in Lc->pKpi creator and task

mfaggin:
Hi, here a first implementation of SigmaC code @vkucera

PR #1856, opened 2023-01-27
PWGHF: Move D2H tasks to dedicated PAG folder

lvermunt:
As agreed with the PWGHF conveners several months ago (sorry for the delay) and recently revisited, we move the D2H tasks from `PWGHF/Tasks` to the dedicated `PWGHF/D2H/Tasks` directory.

To not confuse future D2H analysers on which task to use, I did not touch the (outdated?) `task[D0/Lc]ParametrizedPID.cxx`, `taskLcCentrality.cxx`, and dedicated ALICE3 tasks. I also kept the QA tasks in the general directory, since I think these are useful for multiple PAGs.

This Sigmac task from @mfaggin (of open PR #1710) should be moved as well before being merged.


fcolamar (approved):
Ciao Luuk, thanks for moving the D2H tasks in the dedicated folder, I'm approving the PR since this had been already agreed in the past.

PR #1857, opened 2023-01-27
Remove unnecessary DetectorsVertexing linking

ddobrigk:
@ktf following our conversation... this is only the part in which I cleaned up the unnecessary linking. 

@mpuccio @jgrosseo @strogolo this is ONLY the "simple" part of not linking unnecessary libraries. In principle if this compiles and passes all tests, and therefore merges OK, this should be completely harmless. But of course please jump up in case you have objections :-)


ddobrigk:
Some background: an abnormally high memory consumption was found recently for a variety of tasks. After investigation (thanks a lot to Giulio and Sandro!) it was found that the DetectorsVertexing library linked GlobalTracking, which in turn linked MFTTracking, and MFT tracking increased in size very significantly on the 18th January. The large memory usage of MFTTracking is being investigated, but this incident has also made us aware that DetectorsVertexing was being linked to many tasks that, in fact, have no requirement of that library. This PR removes the linking to DetectorsVertexing from the 71 tasks that do not need it and should result in memory savings for those tasks. 

As a further step: in those cases in which DetectorsVertexing is indeed required, that's because of the DCAFitter. However, loading the entire GlobalTracking library because of that is also a bit unnecessary, so in continuation, the next logical step is to modularize the DCAFitter (under e.g. Common/DCAFitter) such that it becomes a standalone library. This will then improve the memory consumption of tasks such as strangeness and HF candidate building and will be done in a subsequent PR in the future.

ktf:
IMHO, we should also segregate all the libraries we want to allow from O2 into a single entrypoint (say O2::ReconstructionCore), and then forbid any other explicit usage of O2 packages in analysis.

ddobrigk:
@ktf yes, I think here the problem was that everyone (including myself on occasion) copy-pasted the vertexing when in doubt and having a single entry point for stuff that could be useful would even make it simpler for people to know if what they are doing is ok or not, so I fully agree. One thing only, then, is that even within the single entry point it is maybe good to keep good modularity of sub-items to be loaded to memory instead of just having one bigger library of utilities, I guess.

ktf:
Well, if you end up loading them all in one place or another, it does not really matter, but yes, we could have a few of those entrypoints. The important thing to me is that not everything which is in O2 is considered to be "linkable".

jgrosseo (approved):

ddobrigk:
Could we please merge this? I don't have permissions to override the many needed approvals but perhaps @shahor02 or someone else has permissions to merge? Thank you!!!

njacazio (approved):

PR #1860, opened 2023-01-27
Change CEFP to deal with schema evolution

mpuccio:
- Fix schema evolution treatment in CEFP
- Fix bin label assignment
- Remove collision id from output table


mpuccio:
Hi @fkrizek, I also removed the collision id from your output table, as it is not necessary.

alibuild (approved):

PR #1862, opened 2023-01-27
PWGHF: Add variables for Ds selection + update Ds selector and task

fcatalan92:


lvermunt (approved):
Hi @fcatalan92, looks good to me, I approve (but you need one more).

Just one thing, it seems to me that for the `KKPi` case you are using the other Kaon in the `cosPiKPhiRestFrame()` function with respect to the AliPhysics equivalent. I let it to you to judge if this is fine/it's better to keep it consistent with AliPhysics/I overlooked something and it's fine after all.

AliPhysics line: https://github.com/alisw/AliPhysics/blob/6015b235c21bc8b9d00af9a764be1fb58f7bb32d/PWGHF/vertexingHF/AliAODRecoDecayHF3Prong.cxx#L308

lvermunt:
I believe this can be merged @vkucera, @cterrevo, @ginnocen, @fcolamar 

(the minor point I mentioned can be addressed later if needed)

PWGHF/TableProducer/candidateSelectorDsToKKPi.cxx
```diff
@@ -43,47 +44,92 @@ struct HfCandidateSelectorDsToKKPi {
...
       return false;
     }
-    // cut on daughter pT
+
     if (trackKaon1.pt() < cuts->get(pTBin, "pT K") || trackKaon2.pt() < cuts->get(pTBin, "pT K") || trackPion.pt() < cuts->get(pTBin, "pT Pi")) {
       return false;
     }
-    // invariant-mass cut
-    if (std::abs(invMassDsToKKPi(candidate) - RecoDecay::getMassPDG(pdg::Code::kDS)) > cuts->get(pTBin, "m") && (std::abs(invMassDsToPiKK(candidate) - RecoDecay::getMassPDG(pdg::Code::kDS)) > cuts->get(pTBin, "m"))) {
+    if (std::abs(invMassDsToKKPi(candidate) - RecoDecay::getMassPDG(pdg::Code::kDS)) > cuts->get(pTBin, "deltaM")) {
```
vkucera, line 97: Why did you drop the cut on `invMassDsToPiKK`?

vkucera (changes requested):

PWGHF/D2H/Tasks/taskDs.cxx
```diff
@@ -90,117 +101,188 @@ struct HfTaskDs {
...
-    for (auto& candidate : recoFlagDsCandidates) {
-      // not possible in Filter since expressions do not support binary operators
-      if (!(candidate.hfflag() & 1 << DecayType::DsToKKPi)) {
+    for (auto& candidate : recodDsToKKPiCand) {
+      if (yCandMax >= 0. && std::abs(yDs(candidate)) > yCandMax) {
         continue;
       }
+      if (std::abs(candidate.flagMcMatchRec()) == 1 << DecayType::DsToKKPi &&
+          std::abs(candidate.prong0_as<aod::BigTracksMC>().mcParticle_as<candDsMcGen>().pdgCode()) == kKPlus) {
+        auto indexMother = RecoDecay::getMother(particlesMC, candidate.prong0_as<aod::BigTracksMC>().mcParticle_as<candDsMcGen>(), pdg::Code::kDS, true);
```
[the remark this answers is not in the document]
fcatalan92: I had a second look at the implementation and I refactored the code removing the separate loop for the KKPi e PiKK mass hypotheses. I also use only one call to `candidate.prong0_as<aod::BigTracksMC>().mcParticle_as<candDsMcGen>()`

PWGHF/TableProducer/candidateSelectorDsToKKPi.cxx
fcatalan92: I separated the selections depending on the mass hypothesis into two different functions, see L116. 
Actually, the previous implementation was incorrect, since the candidate had to fail the checks for both mass hypotheses to be rejected and one could imagine (although very unlikely) that a candidate could pass the mass selection for one mass hypothesis and the rest of the selections for the other one.

fcatalan92:
@vkucera I addressed your comments

@lvermunt Indeed, I switched which kaon is used w.r.t. the AliPhysics implementation, which used different tracks for the `PiKK` and `KKPi` hypotheses. Now, always the opposite signed kaon is used.
On the practical side, this should not have any effect. Since the `KK` pair is back-to-back in the phi meson rest frame, the difference w.r.t. the old code is at most a minus sign and we are using the absolute value of this quantity to cut (I will in any case check this when validating the code w.r.t. AliPhysics).

lvermunt (approved):

vkucera (approved):

PR #1866, opened 2023-01-28
First tests of separate DCAFitter library

ddobrigk:
for much reduced memory footprints + partial cleanup


ddobrigk:
Requires newer O2 build than the tester uses. Will re-test tomorrow

PR #1871, opened 2023-01-29
boolean enabling 'dummy' vals (keep def operation)

ddobrigk:


alibuild (approved):

PR #1872, opened 2023-01-29
[Event Filtering] improve memory and move helper methods in dedicated file

fgrosa:


alibuild (approved):

PR #1880, opened 2023-01-30
strangeness builders adjustments / fixes

ddobrigk:
- removes setMatLUT call from first init to prevent a spurious crash due to an implict (and not-up-to-date) magnetic field initialization inside the propagator 
- preselector processBuildValiddEdx changes from TPCLf to TPCFull tables to be compliant with CEFP operations


njacazio (approved):

alibuild (approved):

PR #1883, opened 2023-01-31
Validating and updating the HF-Jet framework

nzardosh:


ddobrigk (approved):
Looks good! One quick comment only: I guess you need all the libs you are linking to in the CMake, but if ever there is a lib that you don't need, remember to remove it because it reduces the memory footprint of the analysis framework when running these tasks. Thanks!

PR #1886, opened 2023-01-31
Move ALICE 3 to dedicated DCAFitter library

ddobrigk:
This PR moves the existing ALICE 3 code to the new standalone DCAFitter library to avoid depending on DetectorsVertexing unnecessarily.


njacazio (approved):
Good catch, thanks David

PR #1887, opened 2023-01-31
PWGDQ: Streamline linked libraries

ddobrigk:
* remove unnecessary linking to DetectorsBase
* Move to link to DCAFitter standalone library instead of DetectorsVertexing to reduce memory consumption

### DIFF
--- a/PWGDQ/Core/CMakeLists.txt
+++ b/PWGDQ/Core/CMakeLists.txt
@@ -21,7 +21,7 @@ o2physics_add_library(PWGDQCore
                         AnalysisCompositeCut.cxx
                         MCProng.cxx
                         MCSignal.cxx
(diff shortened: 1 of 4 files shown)


ddobrigk:
@feisenhu @iarsene @dsekihat - just to give some context, this PR is part of the cleanup effort to reduce the memory consumption of our analysis framework. What has been done here is to change the library that provides the `DCAFitter` from the old `DetectorsVertexing` to the new standalone `DCAFitter` library, since for now (temporarily) the `DCAFitter` is provided by both libraries. This is a completely transparent change that will make no difference in any physics performance, and the next step after this is to remove the `DCAFitter` from the `DetectorsVertexing` library in O2 (which can of course only be done after this PR gets merged). In addition, `DetectorsBase` was removed from analysis since that library isn't actually required. Please do let me know in case of any issues or questions and thanks a lot!

iarsene (approved):

iarsene:
Thanks @ddobrigk

PR #1888, opened 2023-01-31
PWGEM: streamline library linking

ddobrigk:
* remove unnecessary links to DetectorsBase
* Move to linking DCAFitter dedicated library instead of DetectorsVertexing

### DIFF
--- a/PWGEM/PhotonMeson/TableProducer/CMakeLists.txt
+++ b/PWGEM/PhotonMeson/TableProducer/CMakeLists.txt
@@ -11,7 +11,7 @@
 
 o2physics_add_dpl_workflow(skimmergammaconversions
                     SOURCES skimmerGammaConversions.cxx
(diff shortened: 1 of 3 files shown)


ddobrigk:
@mikesas @rbailhac @m-c-danisch @novitzky just to give some context, this PR is part of the cleanup effort to reduce the memory consumption of our analysis framework. What has been done here is to change the library that provides the `DCAFitter` from the old `DetectorsVertexing` to the new standalone `DCAFitter` library, since for now (temporarily) the `DCAFitter` is provided by both libraries. This is a completely transparent change that will make no difference in any physics performance, and the next step after this is to remove the `DCAFitter` from the `DetectorsVertexing` library in O2 (which can of course only be done after this PR gets merged). Please do let me know in case of any issues or questions and thanks a lot!

rbailhac (approved):

PR #1889, opened 2023-01-31
PWGHF: Streamline HF library dependencies

ddobrigk:
* remove `ROOT::EG` as linked library from table builders (not necessary, not used) 
* fully remove `DetectorsVertexing` or `DCAFitter` dependency from tree creators, not used
* move to `DCAFitter` standalone library for those tasks that use the `DCAFitter`
* kept `DetectorsVertexing` dependency for the [trackIndexSkimCreator.cxx](https://github.com/AliceO2Group/O2Physics/blob/master/PWGHF/TableProducer/trackIndexSkimCreator.cxx), which uses seemingly `pvertexer` classes

### DIFF
--- a/PWGHF/TableProducer/CMakeLists.txt
+++ b/PWGHF/TableProducer/CMakeLists.txt
@@ -11,7 +11,7 @@
 
 o2physics_add_dpl_workflow(track-index-skim-creator
                     SOURCES trackIndexSkimCreator.cxx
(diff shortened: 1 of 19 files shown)


ddobrigk:
So, there was one final step missing: the trackIndexSkimsCreator needs to depend on both the `DCAFitter` and the `DetectorsVertexing` library such that we can then proceed to (adiabatically) remove the `DCAFitter` from the `DetectorsVertexing` library and have everything work properly afterwards with less memory consumption. The rest stands as is. Note that none of this will have any impact in the results except for the memory consumption when getting those results :-D Please let me know if you spot something that looks suspicious and thanks!

vkucera (approved):

PR #1890, opened 2023-01-31
Tools: KFParticle QA tasks

AnnalenaKa:
In this PR two QA tasks for the KF particle will me introduced. 
One task can be used for basic checks on collisions and tracks and the second task does a D0 reconstruction using the KF package.


ddobrigk (approved):
Thanks! Approving following private message exchange

PR #1894, opened 2023-01-31
Multiplicity in Run2 and DCA binning

gmantzar:
1)
We added if statements to use the correct multiplicity estimator when running on converted data. For that we included new collumns in the FemtoDreamTable. We will test the them and evaluate after a hyperloop run, which colums to keep for the multiplicity.

2)
we made the binning of the TempFitVar histogramm (in the currently used scenario the DCA) into a configurable. This way it can also be used for future analysis involving also V0s and Cascades

### DIFF
--- a/PWGCF/DataModel/FemtoDerived.h
+++ b/PWGCF/DataModel/FemtoDerived.h
@@ -27,12 +27,12 @@ namespace o2::aod
 /// FemtoDreamCollision
 namespace femtodreamcollision
 {
-DECLARE_SOA_COLUMN(MultV0M, multV0M, float);               //! V0M multiplicity
-DECLARE_SOA_COLUMN(MultNtrPV, multNtrPV, int);             //! multiplicity of charged tracks contributing to the primary vertex
-DECLARE_SOA_COLUMN(MultNtrWithTPC, multNtrWithTPC, int);   //! multiplicity of charged tracks with TPC information
-DECLARE_SOA_COLUMN(MultNtrlets, multNtrlets, int);         //! multiplicity of tracklets (used in Run2)
-DECLARE_SOA_COLUMN(Sphericity, sphericity, float);         //! Sphericity of the event
-DECLARE_SOA_COLUMN(MagField, magField, float);             //! Magnetic field of the event
+DECLARE_SOA_COLUMN(MultV0M, multV0M, float);             //! V0M multiplicity
+DECLARE_SOA_COLUMN(MultNtrPV, multNtrPV, int);           //! multiplicity of charged tracks contributing to the primary vertex
+DECLARE_SOA_COLUMN(MultNtrWithTPC, multNtrWithTPC, int); //! multiplicity of charged tracks with TPC information
+DECLARE_SOA_COLUMN(MultNtrlets, multNtrlets, int);       //! multiplicity of tracklets (used in Run2)
+DECLARE_SOA_COLUMN(Sphericity, sphericity, float);       //! Sphericity of the event
+DECLARE_SOA_COLUMN(MagField, magField, float);           //! Magnetic field of the event
 
 } // namespace femtodreamcollision
 
--- a/PWGCF/DataModel/FemtoDerived.h
+++ b/PWGCF/DataModel/FemtoDerived.h
@@ -27,10 +27,12 @@ namespace o2::aod
 /// FemtoDreamCollision
 namespace femtodreamcollision
 {
-DECLARE_SOA_COLUMN(MultV0M, multV0M, float);       //! V0M multiplicity
-DECLARE_SOA_COLUMN(MultNtrPV, multNtrPV, float);   //! multiplicity of charged tracks contributing to the primary vertex
-DECLARE_SOA_COLUMN(Sphericity, sphericity, float); //! Sphericity of the event
-DECLARE_SOA_COLUMN(MagField, magField, float);     //! Sphericity of the event
+DECLARE_SOA_COLUMN(MultV0M, multV0M, float);               //! V0M multiplicity
+DECLARE_SOA_COLUMN(MultNtrPV, multNtrPV, int);             //! multiplicity of charged tracks contributing to the primary vertex
+DECLARE_SOA_COLUMN(MultNtrWithTPC, multNtrWithTPC, int);   //! multiplicity of charged tracks with TPC information
+DECLARE_SOA_COLUMN(MultNtrlets, multNtrlets, int);         //! multiplicity of tracklets (used in Run2)
+DECLARE_SOA_COLUMN(Sphericity, sphericity, float);         //! Sphericity of the event
+DECLARE_SOA_COLUMN(MagField, magField, float);             //! Magnetic field of the event
 
 } // namespace femtodreamcollision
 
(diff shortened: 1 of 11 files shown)


PWGCF/FemtoDream/femtoDreamPairTaskTrackTrack.cxx
```diff
@@ -113,17 +118,24 @@ struct femtoDreamPairTaskTrackTrack {
...
     MixQaRegistry.add("MixingQA/hSECollisionBins", ";bin;Entries", kTH1F, {{120, -0.5, 119.5}});
     MixQaRegistry.add("MixingQA/hMECollisionBins", ";bin;Entries", kTH1F, {{120, -0.5, 119.5}});
 
-    sameEventCont.init(&resultRegistry, CfgkstarBins, CfgMultBins, CfgkTBins, CfgmTBins);
+    if (ConfAnalyseRun3) {
+      sameEventCont.init(&resultRegistry, CfgkstarBins, CfgMultBins, CfgkTBins, CfgmTBins);
+      mixedEventCont.init(&resultRegistry, CfgkstarBins, CfgMultBins, CfgkTBins, CfgmTBins);
+    } else {
+      // sameEventCont.init(&resultRegistry, CfgkstarBins, {{16384, 0, 32768}}, CfgkTBins, CfgmTBins);
+      // mixedEventCont.init(&resultRegistry, CfgkstarBins, {{16384, 0, 32768}}, CfgkTBins, CfgmTBins);
```
victor-gonzalez, line 134: Please, don't leave code commented

PWGCF/DataModel/FemtoDerived.h
```diff
@@ -27,10 +27,12 @@ namespace o2::aod
...
-DECLARE_SOA_COLUMN(MultV0M, multV0M, float);       //! V0M multiplicity
-DECLARE_SOA_COLUMN(MultNtrPV, multNtrPV, float);   //! multiplicity of charged tracks contributing to the primary vertex
-DECLARE_SOA_COLUMN(Sphericity, sphericity, float); //! Sphericity of the event
-DECLARE_SOA_COLUMN(MagField, magField, float);     //! Sphericity of the event
+DECLARE_SOA_COLUMN(MultV0M, multV0M, float);             //! V0M multiplicity
+DECLARE_SOA_COLUMN(MultNtrPV, multNtrPV, int);           //! multiplicity of charged tracks contributing to the primary vertex
+DECLARE_SOA_COLUMN(MultNtrWithTPC, multNtrWithTPC, int); //! multiplicity of charged tracks with TPC information
+DECLARE_SOA_COLUMN(MultNtrlets, multNtrlets, int);       //! multiplicity of tracklets (used in Run2)
+DECLARE_SOA_COLUMN(Sphericity, sphericity, float);       //! Sphericity of the event
+DECLARE_SOA_COLUMN(MagField, magField, float);           //! Magnetic field of the event
```
victor-gonzalez, line 35: Are you analyzing for different multiplicity estimators?
It  looks to me that you are storing everything and deciding what to use later
This is not a good design practice

PWGCF/FemtoDream/FemtoDreamEventHisto.h
```diff
@@ -47,8 +49,10 @@ class FemtoDreamEventHisto
   {
     if (mHistogramRegistry) {
       mHistogramRegistry->fill(HIST("Event/zvtxhist"), col.posZ());
-      mHistogramRegistry->fill(HIST("Event/MultNTracksPV"), col.multNtrPV());
       mHistogramRegistry->fill(HIST("Event/MultV0M"), col.multV0M());
+      mHistogramRegistry->fill(HIST("Event/MultNTracksPV"), col.multNtrPV());
+      mHistogramRegistry->fill(HIST("Event/MultNTracksWithTPC"), col.multNtrWithTPC());
+      mHistogramRegistry->fill(HIST("Event/MultNTracklets"), col.multNtrlets());
```
victor-gonzalez, line 55: Are you analyzing for different multiplicity estimators?
It  looks to me that you are storing everything and deciding what to use later
This is not a good design practice

PWGCF/FemtoDream/femtoDreamProducerTask.cxx
```diff
@@ -369,13 +371,13 @@ struct femtoDreamProducerTask {
     // in case of trigger run - store such collisions but don't store any particle candidates for such collisions
     if (!colCuts.isSelected(col)) {
       if (ConfIsTrigger) {
-        outputCollision(vtxZ, mult, multNtr, spher, mMagField);
+        outputCollision(vtxZ, mult, multNtr, multNtrwithTPC, multNtracklets, spher, mMagField);
```
victor-gonzalez, line 374: Same comment as before

PWGCF/FemtoDream/femtoDreamProducerTaskV0Only.cxx
```diff
@@ -228,31 +228,36 @@ struct femtoDreamProducerTaskV0Only {
       mRunNumber = bc.runNumber();
     }
 
+    const auto vtxZ = col.posZ();
+    const auto spher = colCuts.computeSphericity(col, tracks);
+    const auto multNtr = col.multNTracksPV();
+    const auto multNtrwithTPC = col.multTPC();
+    const auto multNtracklets = col.multTracklets();
```
victor-gonzalez, line 235: Same comment as before
Do you plan to simultaneously compare results from different estimators on the same analysis?

victor-gonzalez (commented):
Please have a look at my comments

gmantzar:
Dear Victor,
thank you for your comments. I saw that the BinningPolicy was not updated to use the multiplicity estimator we wanted, so we had to include some more changes.

Concerning your comment about the different estimators: We wanted to use an estimator which gives us directly the number of tracks, thats why we need the `MultNtrPV` column. For Run2, this information does not exist and therefore we want to use the Tracklets in the ITS, which we store in the column `MultNtrlets`.
The third estimator `MultNTrWithTPC` is for particles that also have TPC information and it is available for Run2 and 3. We want to include it for now and test if we can use this for both Runs. But for that we need to be able to see both distributions at the same time and run over a full dataset.
Actually, it would be also nice to be able to switch the binning policy via a configurable to the TPC based estimator. I just included this in the addition to the current pull request.
Finally, the V0M signal, we have to keep because other tasks are depending on it.

victor-gonzalez:
The point is that you will produce your final results with just one multiplicity estimator, right?
It makes no sense to have the multiplicities for a sizable set of estimators in your data model.
If you wanted to check if the multiplicities from different estimators match or not, that you can do it at your first level in your analysis without the need of having the whole set of multiplicities in your data model
One of the side effects that you are already suffering is that you need to add everywhere which multiplicity you have to chose (is Run2 or not? is Run3 or not?) 
Your data model should hide as much as possible such details in a way that the tasks that use it (the data model) don't need to implement again the details that were used to populate your data model

gmantzar:
Ah ok now I understand your point. Yes indeed, this makes much more sense. I changed it now such that we select our multiplicity estimator in the producer task.
I kept the V0M as a separate columns because other tasks seem to be using it. But with respect to the current upstream version we have no additional columns.
Thanks a lot for the feedback. Is it now like you meant it?

victor-gonzalez (dismissed):

gmantzar:
Dear Victor,
The build error happened because I missinterpreted a cpp-linter suggestion. It should be fixed now

victor-gonzalez (approved):
For the next time, please, don't push a PR with code that does not compile locally

PR #1897, opened 2023-02-01
add ITS cluster map study for secondaries

ddobrigk:
@ercolessi @ChiaraDeMartin95 @njacazio @mfaggin let's see. This will be fun :-D


njacazio (dismissed):
Thanks David

mfaggin:
hi @ddobrigk referring to the ITS custer maps things that we discussed together, I think you are on track here.

ddobrigk:
> hi @ddobrigk referring to the ITS custer map things that we discussed together, I think you are on track here.

Yes, actually the idea of the histograms I did is to explicitly find out the behaviour of the afterburning versus the V0 or cascade decay radius. Specifically, I aim to understand why it is that we afterburn V0s from Xi much worse than primary V0s. Presumably, it's the decay radius only - if that's the case, then a V0 that decayed at a certain radius will have its ITS profile be the same regardless of if primary or if secondary. We'll see!

njacazio (approved):

PR #1899, opened 2023-02-01
PID: Decouple multiplicity from TPC PID

njacazio:
@jezwilkinson @chiarazampolli @noferini @strogolo  here's my proposal of decoupling, we should add a dependency on o2-analysis-pid-tpc-base to both PID tasks, I tested it and it works, what do you think? We could also make sure that the value in the other table is picked up from the one of the TPC, of course it creates another dependency, I don't know if this is what we want

### DIFF
--- a/Common/TableProducer/PID/pidTPCBase.cxx
+++ b/Common/TableProducer/PID/pidTPCBase.cxx
@@ -10,9 +10,9 @@
 // or submit itself to any jurisdiction.
 
 ///
-/// \file   pidTOFBase.cxx
+/// \file   pidTPCBase.cxx
 /// \author Nicolò Jacazio nicolo.jacazio@cern.ch
-/// \brief  Base to build tasks for TOF PID tasks.
+/// \brief  Base to build tasks for TPC PID tasks.
 ///
 
 #include <utility>
@@ -21,7 +21,6 @@
 
 // O2 includes
 #include "CCDB/BasicCCDBManager.h"
-#include "TOFBase/EventTimeMaker.h"
 #include "Framework/AnalysisTask.h"
 #include "ReconstructionDataFormats/Track.h"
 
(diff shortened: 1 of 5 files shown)


njacazio:
I'm wondering if we should just replace fully the TPC multiplicity computation in another task

Common/TableProducer/PID/pidTPCBase.cxx
```diff
@@ -0,0 +1,71 @@
...
+
+  //For vertex-Z corrections in calibration
+  Partition<soa::Join<aod::Tracks, aod::TracksExtra>> tracksWithTPC = (aod::track::tpcNClsFindable > (uint8_t)0);
+  bool enableTable = false;
+
+  void init(InitContext& initContext)
+  {
+    LOG(info) << "Initializing PID Mult Task";
+    // Checking that the table is requested in the workflow and enabling it
+    enableTable = isTableRequiredInWorkflow(initContext, "TOFSignal");
```
jezwilkinson, line 52: should this one be `initContext, "PIDMults"`? (i'm not 100% sure of the syntax for `isTableRequiredInWorkflow` and which "table name" is needed)

jezwilkinson:
Hi @njacazio - thanks! The determination and access look consistent with before, this should in principle be fine. Just one doubt I had about the "isTableRequired" in the new base task, i commented directly on the line

Common/TableProducer/PID/pidTPCBase.cxx
njacazio: That's true, I copied it from the TOF base, thanks for spotting this!

njacazio:
Thanks @jezwilkinson fixed!
Since this revolves around the multiplicity also @ddobrigk should be concerned

ddobrigk (approved):

PR #1903, opened 2023-02-01
don't try every event in case dummy requested

ddobrigk:


alibuild (approved):

njacazio (approved):

PR #1909, opened 2023-02-02
PWGHF: bugfix for v0 indices in HfCascades

DanielSamitz:
This seems to be a bug in which V0 index is written into the HfCascades table, because the wrong V0s are then picked up in the candidateCreatorCascade.

Compare to how the V0s are loaded in the candidateCreatorCascade:

https://github.com/AliceO2Group/O2Physics/blob/27d0817f4c5ea57083ed56b53c47fe66b3ac9ca1/PWGHF/TableProducer/candidateCreatorCascade.cxx#L111


ddobrigk:
Hi @DanielSamitz, I think this is indeed a super-bug that you just fixed (a super-bug because this one is slippery :-) ). Great work :-D

vkucera (approved):
Thanks @DanielSamitz

PR #1910, opened 2023-02-02
[EMCAL-850] Use fiter for selection of EMCAL cells

mfasDa:
- Use filter for selection of EMCAL cells (caloType=1)
- Cache number of cells per BC ID in unordered map
  in order to aviod internal loop over cells for each
  collision for speed reason


mfasDa:
A small modification is needed for counting BCs with payload, because the internal loop to find cells per collision was too slow. Instead, since we only need the number of cells, a lookup table with the global BCs is used to determine whether the collision has cell payload.

nzardosh (approved):

PR #1684, opened 2022-12-17
DPG: remove efficiency computed for all particle

njacazio:
- Add length vs last TRD cluster plot

PR #1930, opened 2023-02-06
Add efficiencies for primaries and all particle species

saganatt:
Unfortunately, this PR requires many code changes though it intends to fix just a few things:
- added histograms for "all" = all selected particles together
- added histograms and efficiencies for primaries for eta and phi
- remove a few redundant brackets in `isInAcceptance()`
- allow variable binning (line 1003)

@ginnocen FYI

### DIFF
--- a/DPG/Tasks/AOTTrack/qaEfficiency.cxx
+++ b/DPG/Tasks/AOTTrack/qaEfficiency.cxx
@@ -36,9 +36,9 @@ using namespace o2::framework;
 
 struct QaEfficiency {
   // Particle information
-  static constexpr int nSpecies = o2::track::PID::NIDs; // One per PDG
-  static constexpr const char* particleTitle[nSpecies] = {"e", "#mu", "#pi", "K", "p", "d", "t", "^{3}He", "#alpha"};
-  static constexpr int PDGs[nSpecies] = {kElectron, kMuonMinus, kPiPlus, kKPlus, kProton, 1000010020, 1000010030, 1000020030, 1000020040};
+  static constexpr int nSpecies = o2::track::PID::NIDs + 1; // One per PDG + 1 for all
+  static constexpr const char* particleTitle[nSpecies] = {"e", "#mu", "#pi", "K", "p", "d", "t", "^{3}He", "#alpha", "All"};
+  static constexpr int PDGs[nSpecies] = {kElectron, kMuonMinus, kPiPlus, kKPlus, kProton, 1000010020, 1000010030, 1000020030, 1000020040, 0};
   // Track/particle selection
   Configurable<bool> noFakesHits{"noFakesHits", false, "Flag to reject tracks that have fake hits"};
   Configurable<float> maxProdRadius{"maxProdRadius", 9999.f, "Maximum production radius of the particle under study"};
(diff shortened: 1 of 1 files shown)


njacazio:
But didn't we decide to remove the efficiency for all particles because it was confusing?

njacazio:
https://github.com/AliceO2Group/O2Physics/pull/1684 here we remove it, do you want to add it back? @mfaggin what do you think?

saganatt:
> But didn't we decide to remove the efficiency for all particles because it was confusing?

Yes, I remember. I didn't fully understand the previous definition of "all".
Now my "all" are "all selected". E.g., if we have `do-el`, `do-pr` set to true, all others false, then "all" = el + pr.
I added it as Zaida had the same "all together" plot from AliPhysics.
If it would more confusing than necessary, I can for now remove that part from this PR, shall I do it?

DPG/Tasks/AOTTrack/qaEfficiency.cxx
```diff
@@ -850,7 +1026,25 @@ struct QaEfficiency {
     static_assert(charge == 0 || charge == 1);
     static_assert(id > 0 || id < nSpecies);
 
-    // Selecting a specific PDGs
+    if constexpr (PDGs[id] == 0) { // All PDGs
+      if constexpr (charge == 0) {
+        for (int i = 0; i < nSpecies - 1; i++) {
+          if (PDGCfgs[i] && mcParticle.pdgCode() == PDGs[i]) {
```
njacazio, line 1032: I would remove this, as it can be computed from the single efficiencies of the PDGs offline

DPG/Tasks/AOTTrack/qaEfficiency.cxx
```diff
@@ -817,14 +989,18 @@ struct QaEfficiency {
...
     LOG(info) << "Set doPtEta to: " << (doPtEta ? "true" : "false");
 
+    PDGCfgs = {doEl.value, doMu.value, doPi.value, doKa.value, doPr.value, doDe.value, doTr.value, doHe.value, doAl.value};
+    LOG(info) << "Set PDGCfgs to: " << doEl.value << ", " << doMu.value << ", " << doPi.value << ", " << doKa.value << ", " << doPr.value << ", " << doDe.value << ", " << doTr.value << ", " << doHe.value << ", " << doAl.value;
+
     auto doLimits = [&](double& min, double& max, const ConfigurableAxis& binning) {
       const AxisSpec a{binning, "dummy"};
       min = a.binEdges[0];
-      max = a.binEdges[1];
+      max = a.binEdges[a.binEdges.size() - 1];
```
njacazio, line 1003: Is this needed to work with variable width bins?

njacazio (commented):
Thanks a lot for the work @saganatt very nice! Indeed I would remove the _all_ case as it can be computed offline by summing the single PDGs, this was one of the reasons why we removed it in the first place. What do you think?

DPG/Tasks/AOTTrack/qaEfficiency.cxx
saganatt: this formula works with both fixed-width and variable-width bins. Specifically, for pt we might have variable width.

DPG/Tasks/AOTTrack/qaEfficiency.cxx
saganatt: Ok, I will adjust.

PR #1935, opened 2023-02-06
EventFiltering: adapt HF task to new method for track and collision associations

fgrosa:


alibuild (approved):

PR #1942, opened 2023-02-07
Make labelers Run2/3 agnostic

ddobrigk:
This tiny - but still very cool! - change is made possible by https://github.com/AliceO2Group/AliceO2/pull/10694 and makes the labelers de-reference track indices straight to McTrackLabels, making it fully unnecessary to specify if the data is Run 2 or Run 3.


ddobrigk:
Made mistake, redoing

PR #1951, opened 2023-02-08
Fix unused variables

ktf:


ktf:
Breaks the build currently. Merging.

PR #1962, opened 2023-02-09
[EventFiltering] Small adjustments to HF filter

fgrosa:


alibuild (approved):

PR #1964, opened 2023-02-09
PWGHF: apply minor adjustments

fgrosa:
- do not create histograms if the process function is not enabled
- fix name of flag for track preselection in track-to-collision-associator


vkucera (approved):
Thanks @fgrosa !

PR #1974, opened 2023-02-10
Fix warning

ktf:


ktf:
@TimoWilken @ddobrigk   This is the second time in two days I have to fix a warning because it's breaking the some of the build. Did anything change? In principle we should have consistent checks between alidist checks and O2Physics.

@TimoWilken I also noticed that while reporting errors in https://github.com/AliceO2Group/O2Physics/pull/1922 logs, the PR is marked green. Is that expected?

PR #1975, opened 2023-02-10
Install python scripts

benedikt-voelkel:
* since anyway only useful in alienv, they can now be run via ${O2PHYSICS_ROOT}/share/scripts/<script>.py --> no need to run from SOURCE dir (which is not well-defined)

* can be used via cvmfs --> GRID/Hyperloop

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -75,6 +75,8 @@ add_subdirectory(Tools)
 add_subdirectory(Tutorials)
 add_subdirectory(EventFiltering)
 
+add_subdirectory(Scripts)
+
 # Testing and packaging only needed if we are the top level directory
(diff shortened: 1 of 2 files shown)


benedikt-voelkel:
@vkucera What do you think?

vkucera:
> @vkucera What do you think?

I really like that feature. Thank @benedikt-voelkel . I cannot approve it though.

benedikt-voelkel:
Thanks!
Yes, I find it reasonable and I would like to incorporate some functionality of the dependency finder in nightly testing in order to further ensure the sanity of analyses versus new MC productions on a regular basis.
This will be important/used for GRID productions as well since this will give us a first insight in whether MC productions are good to be analysed.

iarsene (approved):

PR #1982, opened 2023-02-13
[WIP] Implementing background subtraction in the JE framework

hahassan7:
Implementing background subtraction methods in the O2Physics PWGJE framework. The methods that were added are as follows:
1. Median background estimator (as in AliPhysics)
2. Rho sparse estimator, which is used for sparse events like in pp or p-Pb.
3. The perpendicular cone method (used in few papers).
4. The fastjet builtin median background estimator (it's a little bit different from the AliPhysics one).
5. The fastjet builtin constituent subtraction method.

Also the JetFinder class has been clean, and all the stuff related to Bkg. Sub has been moved to a new class.

### DIFF
--- a/PWGJE/Core/JetBkgSubUtils.cxx
+++ b/PWGJE/Core/JetBkgSubUtils.cxx
@@ -14,6 +14,7 @@
 // Author: Hadi Hassan
 #include "PWGJE/Core/JetBkgSubUtils.h"
 #include "Framework/Logger.h"
+#include "TVector2.h"
 
 JetBkgSubUtils::JetBkgSubUtils(float jetBkgR, float bkgPhiMin, float bkgPhiMax, float bkgEtaMin, float bkgEtaMax, float constSubAlpha, float constSubRMax, fastjet::GhostedAreaSpec ghostAreaSpec) : mJetBkgR(jetBkgR),
                                                                                                                                                                                                      mBkgPhiMin(bkgPhiMin),
@@ -26,11 +27,11 @@ JetBkgSubUtils::JetBkgSubUtils(float jetBkgR, float bkgPhiMin, float bkgPhiMax, 
 
 {
   mJetDefBkg = fastjet::JetDefinition(mAlgorithmBkg, mJetBkgR, mRecombSchemeBkg, fastjet::Best);
-  mAreaDefBkg = fastjet::AreaDefinition(fastjet::active_area, mGhostAreaSpec);
+  mAreaDefBkg = fastjet::AreaDefinition(fastjet::active_area_explicit_ghosts, mGhostAreaSpec);
   mSelRho = fastjet::SelectorRapRange(mBkgEtaMin, mBkgEtaMax) && fastjet::SelectorPhiRange(mBkgPhiMin, mBkgPhiMax) && !fastjet::SelectorNHardest(2); // here we have to put rap range, to be checked!
 }
 
-double JetBkgSubUtils::estimateRhoAreaMedian(std::vector<fastjet::PseudoJet>& inputParticles, bool doAreaSparse)
+double JetBkgSubUtils::estimateRhoAreaMedian(std::vector<fastjet::PseudoJet>& inputParticles)
 {
 
   if (inputParticles.size() == 0) {
@@ -60,7 +61,7 @@ double JetBkgSubUtils::estimateRhoAreaMedian(std::vector<fastjet::PseudoJet>& in
   // calculate Rho as the median of the jet pT / jet area
   double rho = TMath::Median<double>(rhovector.size(), rhovector.data());
 
-  if (doAreaSparse) {
+  if (mDoSparseSub) {
     // calculate The ocupancy factor, which the ratio of covered area / total area
     double occupancyFactor = totalAreaCovered > 0 ? totaljetAreaPhys / totalAreaCovered : 1.;
     rho *= occupancyFactor;
@@ -79,30 +80,30 @@ double JetBkgSubUtils::estimateRhoPerpCone(const std::vector<fastjet::PseudoJet>
   double perpPtDensity1 = 0;
   double perpPtDensity2 = 0;
 
-  fastjet::Selector selLeadingJet = fastjet::SelectorEtaRange(mBkgEtaMin, mBkgEtaMax) && fastjet::SelectorPhiRange(mBkgPhiMin, mBkgPhiMax);
+  fastjet::Selector selectJet = fastjet::SelectorEtaRange(mBkgEtaMin, mBkgEtaMax) && fastjet::SelectorPhiRange(mBkgPhiMin, mBkgPhiMax);
 
-  std::vector<fastjet::PseudoJet> leadingJets = fastjet::sorted_by_pt(selLeadingJet(jets));
+  std::vector<fastjet::PseudoJet> selectedJets = fastjet::sorted_by_pt(selectJet(jets));
 
-  if (leadingJets.size() == 0) {
+  if (selectedJets.size() == 0) {
     return 0.0;
   }
 
-  fastjet::PseudoJet leadingJet = leadingJets[0];
+  fastjet::PseudoJet leadingJet = selectedJets[0];
 
   double dPhi1 = 999.;
   double dPhi2 = 999.;
   double dEta = 999.;
   double PerpendicularConeAxisPhi1 = 999., PerpendicularConeAxisPhi2 = 999.;
   // build 2 perp cones in phi around the leading jet (right and left of the jet)
-  PerpendicularConeAxisPhi1 = ((leadingJet.phi() + (M_PI / 2.)) > (2 * M_PI)) ? leadingJet.phi() - ((3. / 2.) * M_PI) : leadingJet.phi() + (M_PI / 2.);
-  PerpendicularConeAxisPhi2 = ((leadingJet.phi() - (M_PI / 2.)) < 0) ? leadingJet.phi() + ((3. / 2.) * M_PI) : leadingJet.phi() - (M_PI / 2.);
+  PerpendicularConeAxisPhi1 = TVector2::Phi_0_2pi(leadingJet.phi() + (M_PI / 2.));
+  PerpendicularConeAxisPhi2 = TVector2::Phi_0_2pi(leadingJet.phi() - (M_PI / 2.));
 
   for (auto& particle : inputParticles) {
     // sum the momentum of all paricles that fill the two cones
-    dPhi1 = TMath::Abs(particle.phi() - PerpendicularConeAxisPhi1);
-    dPhi1 = (dPhi1 > M_PI) ? 2 * M_PI - dPhi1 : dPhi1;
-    dPhi2 = TMath::Abs(particle.phi() - PerpendicularConeAxisPhi2);
-    dPhi2 = (dPhi2 > M_PI) ? 2 * M_PI - dPhi2 : dPhi2;
+    dPhi1 = particle.phi() - PerpendicularConeAxisPhi1;
+    dPhi1 = TVector2::Phi_mpi_pi(dPhi1);
+    dPhi2 = particle.phi() - PerpendicularConeAxisPhi2;
+    dPhi2 = TVector2::Phi_mpi_pi(dPhi2);
     dEta = leadingJet.eta() - particle.eta();
     if (TMath::Sqrt(dPhi1 * dPhi1 + dEta * dEta) <= mJetBkgR) {
       perpPtDensity1 += particle.perp();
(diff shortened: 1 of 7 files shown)


nzardosh (changes requested):
Hi Haadi, thanks for the nice work. I have some change requests which I will attach to the review. A general one would be to remove m from the beginning of variable names to keep it consistent with the rest of the code

PWGJE/Core/JetBkgSubUtils.cxx
```diff
@@ -0,0 +1,182 @@
...
+      perpPtDensity1 += particle.perp();
+    }
+
+    if (TMath::Sqrt(dPhi2 * dPhi2 + dEta * dEta) <= mJetBkgR) {
+      perpPtDensity2 += particle.perp();
+    }
+  }
+
+  // Caculate rho as the ratio of average pT of the two cones / the cone area
+  Double_t perpPtDensity = (perpPtDensity1 + perpPtDensity2) / (2 * M_PI * mJetBkgR * mJetBkgR);
```
nzardosh, line 117: What happens if one if the cones is outside the detector acceptance? can this happen or are they always in since phi is the only variable changed? if one can be outside then we should make sure the denominator here doesnt have both cone areas in that case

PWGJE/Core/JetBkgSubUtils.cxx
```diff
@@ -0,0 +1,182 @@
...
+}
+
+/// Sets the background subtraction estimater pointer
+void JetBkgSubUtils::setBkgE()
+{
+  bkgE = decltype(bkgE)(new fastjet::JetMedianBackgroundEstimator(mSelRho, mJetDefBkg, mAreaDefBkg));
+}
+
+/// Sets the background subtraction object
+fastjet::Subtractor JetBkgSubUtils::setSub(std::vector<fastjet::PseudoJet>& inputParticles, float& rhoParam, BkgSubMode bkgSubMode, const std::vector<fastjet::PseudoJet>& jets)
```
nzardosh, line 129: you can put the jet vector after the particle one in case we want to set default values for the other quantities later

PWGJE/Core/JetBkgSubUtils.cxx
```diff
@@ -0,0 +1,182 @@
...
+{
+  fastjet::Subtractor sub;
+
+  if (bkgSubMode == BkgSubMode::rhoMedianAreaSub) {
+
+    rhoParam = estimateRhoAreaMedian(inputParticles);
+    sub = fastjet::Subtractor(rhoParam);
+
+  } else if (bkgSubMode == BkgSubMode::rhoSparseSub) {
```
nzardosh, line 138: since here the rhomedianareasun and the rhosparsesub have different enums then we can remove the doAreaSparse input variable in line 33 and track this with a internal class member which is set if the correct enum is chosen. In this way if someone sets the enum they dont need to remember to set the other variable too

PWGJE/Core/JetBkgSubUtils.cxx
```diff
@@ -0,0 +1,182 @@
...
+
+    rhoParam = estimateRhoAreaMedian(inputParticles, true);
+    sub = fastjet::Subtractor(rhoParam);
+
+  } else if (bkgSubMode == BkgSubMode::rhoPerpConeSub) {
+
+    if (jets.size() == 0) {
```
nzardosh, line 145: can this happen and should we still return a valid object. In the case that we do we should initialise the sub object with some well controlled default value

PWGJE/Core/JetBkgSubUtils.h
```diff
@@ -0,0 +1,106 @@
...
+};
+
+class JetBkgSubUtils
+{
+
+ public:
+  // Default contructor
+  JetBkgSubUtils() = default;
+
+  JetBkgSubUtils(float jetBkgR_out, float bkgPhiMin_out, float bkgPhiMax_out, float bkgEtaMin_out, float bkgEtaMax_out, float constSubAlpha_out, float constSubRMax_out, fastjet::GhostedAreaSpec ghostAreaSpec_out);
```
nzardosh, line 50: I think all these inputs can have default values in the constructor since things like phi will not change for charge jet analyses at least. And then these inputs can have setters which can be set if needed, also the same for the protected members at the bottom of the task.

PWGJE/Core/JetBkgSubUtils.h
```diff
@@ -0,0 +1,106 @@
...
+#include "fastjet/tools/JetMedianBackgroundEstimator.hh"
+#include "fastjet/tools/Subtractor.hh"
+#include "fastjet/contrib/ConstituentSubtractor.hh"
+
+#include "Framework/Logger.h"
+
+enum class BkgSubMode { none,
+                        rhoAreaSub,
+                        constSub,
+                        rhoSparseSub,
```
nzardosh, line 38: we should decide if this needs to be its own enum or a setter in rhoAreaSub

PWGJE/Core/JetFinder.cxx
```diff
@@ -14,82 +14,49 @@
...
   jets.clear();
 
-  if (bkgE) {
-    bkgE->set_particles(inputParticles);
-    setSub();
-  }
-  if (constituentSub) {
-    inputParticles = constituentSub->subtract_event(inputParticles);
-  }
+  sub = mSubUtils->setSub(inputParticles, mRho, bkgSubMode);
```
nzardosh, line 51: you can put a comment here saying in case of constituentsub the subtraction is already done here in this line. We can also add an extra constituent sub option (separate to this one) like there used to be where the constituent subtraction is only applied to the clustered jet constituents and not to the whole event. This could be implemented like the perpendicular cone method which takes in the jets

PWGJE/Core/JetFinder.cxx
```diff
@@ -14,82 +14,49 @@
...
+
   fastjet::ClusterSequenceArea clusterSeq(inputParticles, jetDef, areaDef);
-  jets = sub ? (*sub)(clusterSeq.inclusive_jets()) : clusterSeq.inclusive_jets();
+
+  if (bkgSubMode == BkgSubMode::rhoPerpConeSub) {
+    sub = mSubUtils->setSub(inputParticles, mRho, bkgSubMode, clusterSeq.inclusive_jets());
+  }
+
+  jets = (mRho > DBL_EPSILON && bkgSubMode != BkgSubMode::constSub) ? (sub)(clusterSeq.inclusive_jets()) : clusterSeq.inclusive_jets();
```
nzardosh, line 59: what is DBL_EPSILON?

PWGJE/TableProducer/jetfinder.cxx
```diff
@@ -66,13 +66,22 @@ struct JetFinderTask {
...
   OutputObj<TH2F> hJetEta{"h_jet_eta"};
   OutputObj<TH2F> hJetN{"h_jet_n"};
+  OutputObj<TH2F> hJetRho{"h_jet_rho"};
 
   Configurable<float> vertexZCut{"vertexZCut", 10.0f, "Accepted z-vertex range"};
   Configurable<float> trackPtCut{"trackPtCut", 0.1, "minimum constituent pT"};
   Configurable<float> trackEtaCut{"trackEtaCut", 0.9, "constituent eta cut"};
-  Configurable<bool> DoRhoAreaSub{"DoRhoAreaSub", false, "do rho area subtraction"};
+
+  Configurable<bool> DoRhoAreaSub{"DoRhoAreaSub", false, "do rho area subtraction (fastjet)"};
```
nzardosh, line 75: we can only select one of these at a time right? I guess the later would override each other if multiple are selected? In that case instead of having multiple configurable you can have one configurable which takes in the enum values for the subtraction types and one can set the subtraction type through that. This would cause less errors by not allowing accidentally to have multiple on at the same time

nzardosh (changes requested):

PWGJE/Core/JetBkgSubUtils.cxx
```diff
@@ -0,0 +1,182 @@
...
+    // Full area
+    totalAreaCovered += ijet.area();
+  }
+  // calculate Rho as the median of the jet pT / jet area
+  double rho = TMath::Median<double>(rhovector.size(), rhovector.data());
+
+  if (doAreaSparse) {
+    // calculate The ocupancy factor, which the ratio of covered area / total area
+    double occupancyFactor = totalAreaCovered > 0 ? totaljetAreaPhys / totalAreaCovered : 1.;
+    rho *= occupancyFactor;
```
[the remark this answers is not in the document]
hahassan7: No, rho has to multiplied by the occupancy factor, so it can correct for covered/total area.

PWGJE/Core/JetBkgSubUtils.cxx
```diff
@@ -0,0 +1,182 @@
...
+  mJetDefBkg = fastjet::JetDefinition(mAlgorithmBkg, mJetBkgR, mRecombSchemeBkg, fastjet::Best);
+  mAreaDefBkg = fastjet::AreaDefinition(fastjet::active_area, mGhostAreaSpec);
+  mSelRho = fastjet::SelectorRapRange(mBkgEtaMin, mBkgEtaMax) && fastjet::SelectorPhiRange(mBkgPhiMin, mBkgPhiMax) && !fastjet::SelectorNHardest(2); // here we have to put rap range, to be checked!
+}
+
+double JetBkgSubUtils::estimateRhoAreaMedian(std::vector<fastjet::PseudoJet>& inputParticles, bool doAreaSparse)
+{
+
+  if (inputParticles.size() == 0) {
+    return 0.0;
```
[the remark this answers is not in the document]
hahassan7: if we put it non zero, it will be assumed that there is a rho and it might be subtracted from the jet. So I think it's safer to keep 0.

PWGJE/Core/JetBkgSubUtils.cxx
```diff
@@ -0,0 +1,182 @@
...
+{
+
+  if (inputParticles.size() == 0 || jets.size() == 0) {
+    return 0.0;
+  }
+
+  double perpPtDensity1 = 0;
+  double perpPtDensity2 = 0;
+
+  fastjet::Selector selLeadingJet = fastjet::SelectorEtaRange(mBkgEtaMin, mBkgEtaMax) && fastjet::SelectorPhiRange(mBkgPhiMin, mBkgPhiMax);
```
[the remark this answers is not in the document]
hahassan7: Working on it.

PWGJE/Core/JetBkgSubUtils.cxx
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+  fastjet::PseudoJet leadingJet = leadingJets[0];
+
+  double dPhi1 = 999.;
+  double dPhi2 = 999.;
+  double dEta = 999.;
+  double PerpendicularConeAxisPhi1 = 999., PerpendicularConeAxisPhi2 = 999.;
+  // build 2 perp cones in phi around the leading jet (right and left of the jet)
+  PerpendicularConeAxisPhi1 = ((leadingJet.phi() + (M_PI / 2.)) > (2 * M_PI)) ? leadingJet.phi() - ((3. / 2.) * M_PI) : leadingJet.phi() + (M_PI / 2.);
```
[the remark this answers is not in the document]
hahassan7: Working on it.

PWGJE/Core/JetBkgSubUtils.cxx
hahassan7: The perpendicular cone is draw perp in the phi direction and the same eta as the leading jet, and our detector is 360° in phi so the cone will be for sure inside the detector acceptance.

PWGJE/Core/JetBkgSubUtils.cxx
hahassan7: rhoParam is given by reference so it can't have a default, but jets vector is needed only in the perp cone method so it needs to be empty by default.

PWGJE/Core/JetBkgSubUtils.cxx
hahassan7: Working on it.

PWGJE/Core/JetBkgSubUtils.cxx
hahassan7: if jets vector was empty, then rho will be set to 0, so bkgsub will be done in this case.

PWGJE/Core/JetBkgSubUtils.h
hahassan7: Working on it.

PWGJE/Core/JetBkgSubUtils.h
hahassan7: I don't fully understand, you mean BkgSubMode enum should be a member in the BkgSubUtils class ?
In principle it's possible, but the problem referring to it will be a bit long `BkgSubUtils::BkgSubMode::rhoAreaSub`  instead of just using `BkgSubMode::rhoAreaSub`.

PWGJE/Core/JetFinder.cxx
hahassan7: I see, yes I will do it ASAP.

PWGJE/Core/JetFinder.cxx
hahassan7: DBL_EPSILON is the0 for floating number since mRho is float. So what I am doing here is that if Rho=0 (which means bkgsub is not set) then don't do any subtraction. The problem here is that sub is no longer a pointer in which we can just say (!sub), but now it's just a normal object. Any idea on how to do it better than this ?

PWGJE/TableProducer/jetfinder.cxx
hahassan7: I didn't know we can do this. I will do it ASAP.

PR #1985, opened 2023-02-14
replace hard coded table names by metadata information

pbuehler:


pbuehler:
not all commits needed

PR #1989, opened 2023-02-14
TOF dev: move to next parameter container for pass handling

njacazio:
Depends on https://github.com/AliceO2Group/AliceO2/pull/10760


ddobrigk (approved):
thanks Nicolò!

PR #1997, opened 2023-02-14
[EMCAL-670] Cluster-collision matching using foundBC

mfasDa:
Match clusters to collision according to their found
BC and not the BC from tracks from the reconstruction.
As such the framework matching has to be disabled
and in order to subscribe to all BC Change in logics:
- Pre-sort cells and global indices before BC loop
- In BC loop access pre-sorted cells objects
- Match collision(s) to cells using PreSlice with
  foundBCId
- If cells found for BC run processing inside BC loop
  as done previoisly

### DIFF
--- a/PWGJE/TableProducer/emcalCorrectionTask.cxx
+++ b/PWGJE/TableProducer/emcalCorrectionTask.cxx
@@ -15,6 +15,7 @@
 
 #include <algorithm>
 #include <iostream>
+#include <memory>
 #include <cmath>
 
 #include "CCDB/BasicCCDBManager.h"
@@ -42,6 +43,9 @@ using namespace o2;
 using namespace o2::framework;
 using namespace o2::framework::expressions;
 using myGlobTracks = o2::soa::Join<o2::aod::FullTracks, o2::aod::TrackSelection>;
+using bcEvSels = o2::soa::Join<o2::aod::BCs, o2::aod::BcSels>;
+using collEventSels = o2::soa::Join<o2::aod::Collisions, o2::aod::EvSels>;
+using filteredCells = o2::soa::Filtered<aod::Calos>;
 
 struct EmcalCorrectionTask {
   Produces<o2::aod::EMCALClusters> clusters;
@@ -58,6 +62,9 @@ struct EmcalCorrectionTask {
   Configurable<float> maxMatchingDistance{"maxMatchingDistance", 0.4f, "Max matching distance track-cluster"};
   Configurable<bool> hasPropagatedTracks{"hasPropagatedTracks", false, "temporary flag, only set to true when running over data which has the tracks propagated to EMCal/PHOS!"};
 
+  // Require EMCAL cells (CALO type 1)
+  Filter emccellfilter = aod::calo::caloType == selectedCellType;
+
   // CDB service (for geometry)
   Service<o2::ccdb::BasicCCDBManager> mCcdbManager;
 
@@ -67,9 +74,6 @@ struct EmcalCorrectionTask {
   std::vector<std::unique_ptr<o2::emcal::Clusterizer<o2::emcal::Cell>>> mClusterizers;
   std::vector<std::unique_ptr<o2::emcal::ClusterFactory<o2::emcal::Cell>>> mClusterFactories;
   // Cells and clusters
-  std::vector<o2::emcal::Cell> mEmcalCells;
-  // map of cellId (local in BC) to global cell index in cell table in AO2D
-  std::map<int, int64_t> mCellIdToCellGlobalIndex;
   std::vector<o2::emcal::AnalysisCluster> mAnalysisClusters;
 
   std::vector<o2::aod::EMCALClusterDefinition> mClusterDefinitions;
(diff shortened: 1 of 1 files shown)


mfasDa:
@fjonasALICE @mhemmer-cern Please have a look!

This implements the proper matching to BCs based using the found BCs. As I pre-sort the cells to BCs before iteration over the BCs the performance loss should not be as severe.

mfasDa:
Removing WIP in order to start the CI, but please don't merge yet!

PWGJE/TableProducer/emcalCorrectionTask.cxx
```diff
@@ -174,238 +178,265 @@ struct EmcalCorrectionTask {
...
-
-          std::vector<double> clusterPhi;
-          std::vector<double> clusterEta;
+            auto&& [clusterToTrackIndexMap, trackToClusterIndexMap] = JetUtilities::MatchClustersAndTracks(clusterPhi, clusterEta, trackPhi, trackEta, maxMatchingDistance, 10);
+            // we found a collision, put the clusters into the none ambiguous table
+            clusters.reserve(mAnalysisClusters.size());
+            int cellindex = -1;
+
+            unsigned int k = 0;
+            for (const auto& cluster : mAnalysisClusters) {
```
mhemmer-cern, line 361: Could we implement a JetUtilities::MatchClustersAndTracks function that only uses a single cluster or just always only provide a single cluster in the array so we could merge the two loops over the mAnalysisClusters? @fjonasALICE 

Other than that, this PR looks good from my side! 👍

PWGJE/TableProducer/emcalCorrectionTask.cxx
```diff
@@ -174,238 +178,265 @@ struct EmcalCorrectionTask {
...
-        hBC->Fill(5);
+      if (hasEMCCell) {
+        hBC->Fill(0);
+        if (collisionsInFoundBC.size() == 0) {
+          hBC->Fill(1);
+        } else if (collisionsInFoundBC.size() == 1) {
+          hBC->Fill(2);
+        } else {
+          hBC->Fill(3);
+        }
```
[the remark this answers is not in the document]
mfasDa: Sure. In a second iteration I want to add a request on the EMCAL trigger but this we can only do when the EMCAL trigger status is in place (apass3 pp and later), in order to skip BCs without trigger and speed up the code.

mfasDa:
For the record: This is needed for high-IR runs: Up to now we were using the automatic matching by the framework for selecting the collision to which we match the EMCAL triggers. As discussed in https://github.com/AliceO2Group/O2Physics/pull/1651 the BC of the collision is determined from tracks which has some imprecision in the BC due to the vertex time assigned from tracks, while for with the trigger we need to use the BC index from the matched TVX trigger, stored in the column foundBC. Unfortunately since it is a dynamic column it cannot be matched by the framework so we have to match ourselves, for which we need to subscribe in the process functions to all BCs in timeframe, otherwise the framework would match the collisions itself and consequently drop collisions which are not matching the BC from tracks. 

For low-IR with single bunches or single trains this is not of relevance since the vertex time is not ambiguous, but for the nominal interaction rate (500 kHz) and the high-IR this matters, otherwise vertices might be randomly assigned.

mfasDa:
Tagging @mpuccio @lietava @ekryshen

PWGJE/TableProducer/emcalCorrectionTask.cxx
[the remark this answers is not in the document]
mfasDa: Done

nzardosh (dismissed):

PWGJE/TableProducer/emcalCorrectionTask.cxx
```diff
@@ -174,238 +178,262 @@ struct EmcalCorrectionTask {
...
+      auto foundBC = cellContainer.find(bcGlobal);
+      std::shared_ptr<std::vector<o2::emcal::Cell>> cellsBC;
+      std::shared_ptr<std::vector<int64_t>> cellIndicesBC;
+      if (foundBC != cellContainer.end()) {
+        cellsBC = foundBC->second.first;
+        cellIndicesBC = foundBC->second.second;
+      } else {
+        cellsBC = std::make_shared<std::vector<o2::emcal::Cell>>();
+        cellIndicesBC = std::make_shared<std::vector<int64_t>>();
+        cellContainer[bcGlobal] = {cellsBC, cellIndicesBC};
```
mhemmer-cern, line 206: Maybe this can be optimised by using try_emplace and reserve:
```c++
std::map<int, std::pair<std::shared_ptr<std::vector<o2::emcal::Cell>>, std::shared_ptr<std::vector<int64_t>>>> cellContainer;
for (auto& cell : cells) {
  auto bcGlobal = cell.bc_as<bcEvSels>().globalBC();
  auto [cellsBC, cellIndicesBC] = cellContainer.try_emplace(bcGlobal, std::make_shared<std::vector<o2::emcal::Cell>>(), std::make_shared<std::vector<int64_t>>())->second;
  cellsBC->reserve(cellsBC->size() + 1);
  cellIndicesBC->reserve(cellIndicesBC->size() + 1);
  cellsBC->emplace_back(cell.cellNumber(),
                        cell.amplitude(),
                        cell.time(),
                        o2::emcal::intToChannelType(cell.cellType()));
  cellIndicesBC->emplace_back(cell.globalIndex());
}
```

mfasDa: The try_emplace syntax looks indeed nice (apart from that value must be std::pair I think (two containers). The reserve not sure what it is needed for, emplace_back should enlarge the container in any case.

PWGJE/TableProducer/emcalCorrectionTask.cxx
fjonasALICE: no I think this is not possible or at least i think we should not do it for single clusters. The system works by knowing positions of call clusters and track. Important is also (however not used yet) that the system works in both directions! It will not only return the closest tracks for each cluster but also all closest clusters for each track

fjonasALICE: in any case, since number of clusters is not too large, i think the second loop does not hurt our performance

mfasDa: Even if such an implementation would be possible and combine loops into 1 running MatchClustersAndTracks per cluster would come with a huge overhead, as the KD-Tree with tracks would have to be created for every cluster separately and would make the code extremely slow. I had similar performance losses in other tasks. Sometimes building maps and processing all in one go is performance-wise better even if it needs an extra loop for building the map.

PWGJE/TableProducer/emcalCorrectionTask.cxx
mhemmer-cern: I am not sure about the evaluation order. If make_shared is indeed evaluated before you are right, this would be an expensice memory allocation that we don't need that often.
The reserve with size +1 should reduce the number of memory allocations that would be needed if we just use emplace_back, if I remember correctly.

nzardosh (approved):

PR #1998, opened 2023-02-15
EventFiltering: Add the kaon PID case and fix the names

zhangbiao-phy:


zhangbiao-phy:
Hi @fgrosa, here is the PR for the kaon case

fgrosa (approved):

PR #2003, opened 2023-02-15
Tools: KF particle task add TOF PID for D0

AnnalenaKa:
In this PR the following points are addressed:
-  Removing unnecessary dependencies as discussed with @ddobrigk .
- Use functions from RecoDecay.h for KFUtils.h in order to avoid errors and duplications.
- Add TOF PID in the D0 KF task
- Add collision time to the event tree.


AnnalenaKa:
Hi @ddobrigk ,
whenever you have time, could you please take a look at the changes in this PR? 
Following your suggestion, I removed dependencies that were not needed. This should improve the performance of the task.
The other changes were implemented in the D0 task, addressing comments I have received in the D2H meeting. 
Now the PID with TPC and TOF is included and a copy of some central functions is removed, but instead the central function is now called in the KFUtils.h.

ddobrigk (approved):
Thanks a lot, looks okay!

PR #2006, opened 2023-02-15
DPG: fix dca recalculation after PV refit.

mfaggin:
thanks a lot to @galocco for spotting this issue!


alibuild (approved):

PR #2009, opened 2023-02-15
Scripts: Add a debugging script to extract useful information from an O2 log file.

vkucera:
This script extracts the following information from an O2 log file:
- devices (omitting `internal-dpl...`)
- input files
- warnings
- errors

If a missing tree is reported, the `find_dependencies.py` script is executed if possible,
otherwise corresponding instructions are provided.

```
Usage: debug_info.sh -f LOGFILE [-d] [-i] [-w] [-e] [-h] 
-d  Show devices.
-i  Show input files.
-w  Show warnings.
-e  Show errors.
```

### DIFF
--- a/Scripts/CMakeLists.txt
+++ b/Scripts/CMakeLists.txt
@@ -9,7 +9,8 @@
 # granted to it by virtue of its status as an Intergovernmental Organization
 # or submit itself to any jurisdiction.
 
-install(FILES find_dependencies.py
+install(FILES debug_info.sh
+              find_dependencies.py
               update_ccdb.py
         PERMISSIONS GROUP_READ GROUP_EXECUTE OWNER_EXECUTE OWNER_WRITE OWNER_READ WORLD_EXECUTE WORLD_READ
         DESTINATION share/scripts/)
--- a/Scripts/debug_info.sh
+++ b/Scripts/debug_info.sh
@@ -1,0 +1,108 @@
+#!/bin/bash
+
+# Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+# See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+# All rights not expressly granted are reserved.
+#
+# This software is distributed under the terms of the GNU General Public
+# License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+#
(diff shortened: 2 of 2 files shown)


ktf:
Please let's avoid having yet another wrapper script and let's address the issues centrally.

vkucera:
> Please let's avoid having yet another wrapper script and let's address the issues centrally.

Hi @ktf , can you be more specific about what you dislike about it?
- What other wrapper script(s) are you referring to?
- Are you suggesting that a tool with these features is not needed?
- Are you suggesting there are already better tools with these features?
- Are you objecting against the call of `find_dependencies.py`?

ktf:
I do not see the point of having a Scripts/debug_info.sh which is a glorified grep taking into accounts different capitalisations (which we should fix) or the fact that some errors do not use Framework/Logger (which we should use). If something in the log makes it difficult to grep for interesting things, we should fix it not add another layer of complexity to fix the complexity below.

vkucera:
@ktf The point of having such a script is simply to have a helper tool that extracts useful debugging information from the O2 log in a way that is faster, easier and more reliable than if the user was scrolling through the log or searching for several keywords manually.
That, in my opinion, is valuable enough to have it in the repository, accessible to anyone.
Unifying the format of reported errors would definitely help but I don't think it is relevant to the question whether this script is useful or not.
I also don't see why analysers should be waiting for it to happen if they need to debug their workflows now.
And even if every single error in the log starts with "[ERROR]" (which is impossible because of fatals and errors coming from outside of O2), it will still be useful to have this script to avoid manual search or grepping, just for the mere reason that the log is long and hard to read.

vkucera:
@ktf @ddobrigk , while I do appreciate Giulio's suggestion to improve the logging, I don't see why it would make this script less useful. Therefore I would like to merge this PR to give other users access to this tool.

vkucera:
@ddobrigk @ktf I am reopening the PR and asking again for merging.

I would like to point out a few reasons:
* Analysers keep posting irrelevant log lines when asking for help in the Mattermost channel.
* The tutorial is approaching and both participants and helpers will need to analyse the logs in case of problems.
* This script is not supposed to replace a proper investigation of the log or the use of debugging tools. It just makes it easier and quicker to spot the problems and the relevant lines of the log.

PR #2012, opened 2023-02-16
Fix unused variable

ktf:


njacazio (approved):

ktf:
Actually, we need to drop almost all of them. I suggest we move to explicitly handled errors on warning, so that we do not get hit by different defaults with different compilers. If you want some warning to become an error, first fix all the instances and then enable it.

PR #2014, opened 2023-02-16
[EMCAL-670] Filter cells with Preslice

mfasDa:
- Use framework Preslice instead of custom
  lookup table for Cells
- Handle QA histograms with histogram registry


mfasDa:
@mhemmer-cern @fjonasALICE

nzardosh (approved):

PR #2030, opened 2023-02-20
PWGHF: Move to using chi2 in vertexer

fcatalan92:
@vkucera I implemented what was discussed at the last HF O2 meeting. 
To not bloat the configuration too much, I have hardcoded the settings of the vertexer to the default minimisation (which is the chi2 one) everywhere unless for the `HfTrackIndexSkimCreator`, `HfCandidateCreator2Prong`, and `HfCandidateCreator3Prong` which could be used for further testing.

### DIFF
--- a/PWGHF/TableProducer/candidateCreator2Prong.cxx
+++ b/PWGHF/TableProducer/candidateCreator2Prong.cxx
@@ -37,6 +37,8 @@ struct HfCandidateCreator2Prong {
   Configurable<bool> doPvRefit{"doPvRefit", false, "do PV refit excluding the candidate daughters, if contributors"};
   // Configurable<double> bz{"bz", 5., "magnetic field"};
   Configurable<bool> propagateToPCA{"propagateToPCA", true, "create tracks version propagated to PCA"};
+  Configurable<bool> useAbsDCA{"useAbsDCA", false, "Minimise abs. distance rather than chi2"};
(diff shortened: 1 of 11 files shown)


vkucera (changes requested):
Thanks @fcatalan92 . Can you please also update [candidateCreatorToXiPi.cxx](https://github.com/AliceO2Group/O2Physics/blob/master/PWGHF/TableProducer/candidateCreatorToXiPi.cxx#L116)?

fcatalan92:
For the [candidateCreatorToXiPi.cxx](https://github.com/AliceO2Group/O2Physics/blob/master/PWGHF/TableProducer/candidateCreatorToXiPi.cxx#L116), I thought these settings were used. Maybe @ZFederica could comment on that.

ZFederica:
Hi @fcatalan92 and @vkucera , what you would change in the candidateCreatorToXiPi.cxx? Only the default value for useAbsDCA, right? (both these options are already implemented in my task, but the defaut value for useAbsDCA is true)

fcatalan92:
Hi @ZFederica, we would remove the configuration parameters `useAbsDCA`, `useWeightedPCA`, and `refitWithMatCorr`, hardcoding the settings to their default values.

ZFederica:
As for the candidateCreatorToXiPi.cxx, it's okay to remove the configuration parameters `useAbsDCA` and `useWeightedPCA`, but I prefer to leave the configurable `refitWithMatCorr` (if you want to change the default value of that configurable is fine)

vkucera (approved):

PR #2034, opened 2023-02-21
Extend jet substructure output and clean up

qgp:
- additional output tables for detector- and generator-level matched jets
- removal of command line options used for different jet finding workflows


vkucera (dismissed):
Approving the HF part

PR #2043, opened 2023-02-21
Add missing configurables of FullJet&PHOS triggers

mpuccio:
@strogolo, I think we will need to recreate the wagon tomorrow﻿


alibuild (approved):

PR #2045, opened 2023-02-22
fix an issue related with mean FV0 mult extraction

aortizve:


alibuild (approved):

PR #2054, opened 2023-02-22
Store global BC id in the output

mpuccio:


alibuild (approved):

PR #2061, opened 2023-02-23
PID: Using TrackExtra instead of tracks

njacazio:


njacazio:
@ddobrigk this is in line with what we discussed. @jezwilkinson This should reduce the dependency and speed up the process, let me know if it's ok with you.

ddobrigk (approved):
Thanks!

PR #2062, opened 2023-02-23
Avoid track selections redefinition for each track

mpuccio:


alibuild (approved):

PR #2068, opened 2023-02-24
Add ZDC table to UD data model

nburmaso:


pbuehler (approved):
This looks good to me. The new table will be joinable with UDCollisions.

PR #2076, opened 2023-02-24
PWGHF: Lc selector with BDT

saganatt:
Lc candidate selector that pre-selects with PID, mass and CPA cuts, and then applies a BDT model from the trigger framework for the final selection.

### DIFF
--- a/PWGHF/TableProducer/candidateSelectorLcMl.cxx
+++ b/PWGHF/TableProducer/candidateSelectorLcMl.cxx
@@ -19,9 +19,11 @@
 
 #include "Framework/runDataProcessing.h"
 #include "Framework/AnalysisTask.h"
+#include "CCDB/CcdbApi.h"
 
 #include "PWGHF/DataModel/CandidateReconstructionTables.h"
 #include "PWGHF/DataModel/CandidateSelectionTables.h"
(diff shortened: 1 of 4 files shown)


PWGHF/TableProducer/candidateSelectorLcMl.cxx
```diff
@@ -0,0 +1,325 @@
...
+            if (scores[0] > thresholdBDTScoreLcToPiKP.get(0u, "BDTbkg")) {
+              // background
+              statusLcToPKPi = 0;
+              statusLcToPiKP = 0;
+            }
+            if (scores[1] > thresholdBDTScoreLcToPiKP.get(0u, "BDTprompt")) {
+              // prompt
+            }
+            if (scores[2] > thresholdBDTScoreLcToPiKP.get(0u, "BDTnonprompt")) {
+              // non-prompt
```
saganatt, line 303: How do we want to treat non-prompt particles, esp. in the case a particle gets classified as both prompt and non-prompt?

saganatt:
@vkucera @DelloStritto FYI

vkucera (approved):

ddobrigk (approved):

PR #2087, opened 2023-02-27
[EMCAL-670] Add cluster nonlinearity and shaper saturation correction

mfasDa:
- Correction for nonlinearity based on
  NonlinearityHandler
- Nonlinearity function selectable
- Energy corrected for nonlinearity as additional
  column in cluster table

### DIFF
--- a/PWGJE/DataModel/EMCALClusters.h
+++ b/PWGJE/DataModel/EMCALClusters.h
@@ -13,9 +13,10 @@
 //
 // Author: Raymond Ehlers
 
-#ifndef O2_ANALYSIS_DATAMODEL_EMCALCLUSTERS
-#define O2_ANALYSIS_DATAMODEL_EMCALCLUSTERS
+#ifndef PWGJE_DATAMODEL_EMCALCLUSTERS_H_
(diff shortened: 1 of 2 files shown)


mfasDa:
Needs https://github.com/AliceO2Group/AliceO2/pull/10845

mfasDa:
@fjonasALICE @jokonig @mhemmer-cern for reference

Shall we use the default energy column for the energy corrected for the nonlinearity? We should keep in mind that we should make the code also easy to use, and most users will care about the corrected energy. The uncorrected energy should be there as core energy in any case.

fjonasALICE:
i would use calibrated energy as default and make uncorrected energy available in a different field. Also, correct nonlin should be automatically applied by default without needing to specify in the configuration of the CF

mfasDa:
@fjonasALICE I also think the default should be the corrected energy. I have changed the pull request accordingly.

For what concerns the selection of the function: I agree with you that it should be automatised. For the first iteration however I would like to keep the Configurable, and we care about this it further iterations.

nzardosh (approved):

mfasDa:
Retrigger CI after update of O2 version in the o2 CI.

nzardosh (approved):

PR #2091, opened 2023-02-28
[PWGJE]: Add option for switching on/off ghost recalculation

mfasDa:
Option for ghost recalculation in the the GhostAreaSpec steered with Configurable
- 1: Ghosts recalculated for each ClusterSequenceArea
- Anything else (preferably 0): One set of ghosts used for all ClusterSequenceAreas

### DIFF
--- a/PWGJE/TableProducer/jetfinder.cxx
+++ b/PWGJE/TableProducer/jetfinder.cxx
@@ -78,6 +78,7 @@ struct JetFinderTask {
   Configurable<bool> DoRhoAreaSub{"DoRhoAreaSub", false, "do rho area subtraction"};
   Configurable<bool> DoConstSub{"DoConstSub", false, "do constituent subtraction"};
   Configurable<float> jetPtMin{"jetPtMin", 10.0, "minimum jet pT"};
+  Configurable<int> ghostRepeat{"ghostRepeat", 1, "set to 0 to gain speed if you dont need area calculation"};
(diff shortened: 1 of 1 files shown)


mfasDa:
@qgp @mverwe @nzardosh Please review!

mfasDa:
Profiling of the jet finder tasks with the macOS time profiler has shown that the dominant amount of time is spent in the initialisation of the ClusterSequenceArea (99.4%) and here in particular in the initialisation of the ghosts (95.4%). With a ghostRepeatN parameter of 1 the ClusterSequenceArea will create a new set of ghosts for each ClusterSequenceArea (i.e. every collision and R), making the jet finder incredibly slow. With any other value of ghostRepeatN the ClusterSequenceArea will create the ghost sample only once and reuse this, resulting in a significant speedup. The ghost recalculation must be made settable, in particular at trigger level we will not be able to afford calculating for each event as it consumes by far too much CPU.

mfasDa:
Hi @nzardosh ,

Thanks for your comment! I have changed the description indicating that setting ghostRepeat to 0 leads to a performance boost and should be used in case area calculation is not needed.

It should be tested however whether a single set of ghost might be sufficient for the area calculation also in high density environments: In case you run over 50 kHz PbPb in regular trains the same problem will persist.

Cheers

Markus

nzardosh (approved):

PR #2094, opened 2023-02-28
depend on TracksIU in Run 3

ddobrigk:
@mpuccio as discussed just now!

@ekryshen, this changes the evsel processRun3 to depend on TracksIU and not Tracks for the track counting. It will mean that the track propagation workflow isn't necessary at all in some cases (strangeness analyses etc). Please jump up in case I am missing something


alibuild (approved):

PR #2095, opened 2023-02-28
Track propag experiment

ddobrigk:
* this PR is needed to do some tests with propagation on hyperloop 
* please be advised this is only temporary: the task will be deleted once all studies are converged
* A better solution for configurability will have to be discussed; will result in updates to code


alibuild (approved):

PR #2107, opened 2023-03-01
CPV index protection added

peressounko:


peressounko:
Dear @ktf, could you please approve this PR? Thanks, Dmitri

ktf:
For the record, I always recommend to have the negation condition and use continue, rather than adding an extra indentation level, but that's of course preference. Merging.

PR #2109, opened 2023-03-01
PWGHF: Clarify and extend PID selections

lvermunt:
- Clarify by renaming `getStatusTrackPIDAll` to `getStatusTrackPIDTpcOrTof` and initialising `nSigma..Cond..` to `0` (which means disabled) instead of `±1000` (which means a completely open PID selection)
- Adding the D2H Run2 PID strategy: TpcAndTof or TPC only in case TOF not available (and vice versa when TPC not available) as `getStatusTrackPIDTpcAndTof`

@vkucera, what do you think?

### DIFF
--- a/Common/Core/TrackSelectorPID.h
+++ b/Common/Core/TrackSelectorPID.h
@@ -94,7 +94,7 @@ class TrackSelectorPID
     mNSigmaTPCMax = nsMax;
   }
 
-  /// Set TPC nσ range in which a track should be conditionally accepted if combined with TOF.
+  /// Set TPC nσ range in which a track should be conditionally accepted if combined with TOF. Set to 0 to disable.
   void setRangeNSigmaTPCCondTOF(float nsMin, float nsMax)
   {
     mNSigmaTPCMinCondTOF = nsMin;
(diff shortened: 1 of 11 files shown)


vkucera:
Hi @lvermunt , thanks for the contribution.
The function `getStatusTrackPIDTpcAndTof` returns `PIDNotApplicable` if "(`NotApplicable` for one detector) and (`NotApplicable` or Conditional for the other)" which is the same behaviour as in the `getStatusTrackPIDTpcAndTof` functions and it's fine, but it also returns `PIDNotApplicable` if (`PIDAccepted` for one detector and `PIDConditional` for the other). Shouldn't such cases be accepted?

lvermunt:
@AnnalenaKa, I thought it was better to do it in two stages (also involving the relevant analysers). So this PR is intended to not change anything on the PID selection in the tasks yet.

@vkucera, indeed I overlooked that case. I added it now. Thanks!

Also fixed the megalinter issue and fixed the newly added LcMl selector

vkucera:
> @AnnalenaKa, I thought it was better to do it in two stages (also involving the relevant analysers). So this PR is intended to not change anything on the PID selection in the tasks yet.
> 
> @vkucera, indeed I overlooked that case. I added it now. Thanks!
> 
> Also fixed the megalinter issue and fixed the newly added LcMl selector

Thanks @lvermunt for the fixes. I agree that this PR should not change the current behaviour of the selectors.

vkucera (approved):

vkucera:
@ddobrigk @jgrosseo @iarsene For me this is ready for merging.

ddobrigk (approved):

PR #2130, opened 2023-03-03
PID TOF add limit for ev time

njacazio:
- update pidtofbase

### DIFF
--- a/Common/TableProducer/PID/pidTOFBase.cxx
+++ b/Common/TableProducer/PID/pidTOFBase.cxx
@@ -60,8 +60,8 @@ struct tofSignal {
       LOG(info) << "Table TOFSignal enabled!";
     }
   }
-  using Trks = o2::soa::Join<aod::Tracks, aod::TracksExtra>;
-  void process(Trks const& tracks)
+  using Trks = o2::soa::Join<aod::TracksIU, aod::TracksExtra>;
+  void processRun3(Trks const& tracks)
   {
     if (!enableTable) {
       return;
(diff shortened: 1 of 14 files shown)


njacazio:
@ddobrigk I also introduced the use of IU tracks for run3 processing

Common/TableProducer/PID/pidTOFBase.cxx
```diff
@@ -60,8 +60,8 @@ struct tofSignal {
       LOG(info) << "Table TOFSignal enabled!";
     }
   }
-  using Trks = o2::soa::Join<aod::Tracks, aod::TracksExtra>;
-  void process(Trks const& tracks)
+  using Trks = o2::soa::Join<aod::TracksIU, aod::TracksExtra>;
```
njacazio, line 63: @jezwilkinson with this PR there will be two process functions, one that works with Run3 and Run2 and one that works only with Run3 (but faster) what do you think?

Common/TableProducer/PID/pidTPCBase.cxx
```diff
@@ -48,16 +50,30 @@ struct PidMultiplicity {
     if (enableTable) {
       LOG(info) << "Table TPC PID Multiplicity enabled!";
     }
+    if (doprocessStandard == true && doprocessIU == true) {
+      LOG(fatal) << "Both processStandard and processIU are enabled, pick one!";
+    }
+  }
+
+  void processIU(aod::Collision const& collision, TrksIU const& tracksExtra)
```
njacazio, line 58: @jezwilkinson sorry, wrong line ;) I wanted to point you at this one

jezwilkinson: Thanks @njacazio for the heads-up - just wondering, is there any way we are able to do this that avoids separate process functions? For TPCBase it's not such a big deal, since that's a very small and simple task, but I was looking into making the same Tracks->TracksIU change for the pidTPC/pidTPCFull table creators, and this gets messy to maintain if we have to duplicate the entire process function for just the table type

njacazio: Maybe we can merge, this PR in principle does not change the default behaviour

ddobrigk (approved):

vkucera:
@njacazio since this PR changes the configuration of the workflow, which affects many analyses, this should be announced in the dedicated channel as soon as the PR is merged.

njacazio:
Hi @vkucera the new configuration is the default, this should transparent for the user, isn't this the case?

PR #2135, opened 2023-03-05
Separated EMCAL and DCAL triggers, option for fiducial cut in jet trigger, fixed jet finder filters

GijsvWeelden:


ddobrigk (dismissed):
Thanks, this looks okay as far as I can tell. I can approve the changes to the directory JE exceptionally, but for the triggering it would be crucial to get @mpuccio / @strogolo to look and provide feedback (they are also the codeowners there, so therefore this also needs formal approval from them on git). @mpuccio @strogolo please take a look when you get a moment! Thanks!

mpuccio (approved):

ddobrigk (approved):

PR #2136, opened 2023-03-05
Expanding upon track selection request class

ddobrigk:
* still experimental, will still change and be tuned
* includes track selection configuration combiner (logical OR)
* missing: explicit ITS map requirement (i.e. L0||L1, etc etc) 
* further logic / algorithmic development needed


alibuild (approved):

PR #2139, opened 2023-03-06
Dev: Add automode for cov mat filling

njacazio:
@mfaggin @iouribelikov


njacazio:
@ddobrigk what do you think of this? This will solve our issue in the DPG that the qaImpactParameter requires the Cov Mat while the PID don't and therefore we ask for 2 separate trains

PR #2149, opened 2023-03-07
Workaround for triggerBC in apass2/3

ekryshen:
* Workaround for triggerBC in apass2/3 (shifted by -294 bcs)
* Use TracksIU instead of Tracks in the evselQA
* latest macro for uploads of trigger aliases

### DIFF
--- a/Common/TableProducer/eventSelection.cxx
+++ b/Common/TableProducer/eventSelection.cxx
@@ -189,13 +189,25 @@ struct BcSelectionTask {
                    aod::FT0s const&,
                    aod::FDDs const&)
   {
+    // map from GlobalBC to BcId needed to find triggerBc
+    std::map<uint64_t, int32_t> mapGlobalBCtoBcId;
+    for (auto& bc : bcs) {
+      mapGlobalBCtoBcId[bc.globalBC()] = bc.globalIndex();
+    }
+
     for (auto bc : bcs) {
       EventSelectionParams* par = ccdb->getForTimeStamp<EventSelectionParams>("EventSelection/EventSelectionParams", bc.timestamp());
       TriggerAliases* aliases = ccdb->getForTimeStamp<TriggerAliases>("EventSelection/TriggerAliases", bc.timestamp());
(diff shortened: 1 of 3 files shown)


Common/TableProducer/eventSelection.cxx
```diff
@@ -189,13 +189,25 @@ struct BcSelectionTask {
...
     for (auto bc : bcs) {
       EventSelectionParams* par = ccdb->getForTimeStamp<EventSelectionParams>("EventSelection/EventSelectionParams", bc.timestamp());
       TriggerAliases* aliases = ccdb->getForTimeStamp<TriggerAliases>("EventSelection/TriggerAliases", bc.timestamp());
       int32_t alias[kNaliases] = {0};
-      uint64_t triggerMask = bc.triggerMask();
-      for (auto& al : aliases->GetAliasToTriggerMaskMap()) {
-        alias[al.first] |= (triggerMask & al.second) > 0;
+
+      // workaround for pp2022 apass2-apass3 (trigger info is shifted by -294 bcs)
+      int32_t triggerBcId = mapGlobalBCtoBcId[bc.globalBC() - 294];
```
mfasDa, line 204: For runs with recalculated CTP digits (LHC22m and part of LHC22o) the BC of the recalculated digits will be at LM, as they are calculated from the detector CTF data which was already corrected to LM during the synchronous reconstruction. Could one therefore maybe make the shift dynamic, and use 0 for runs with recalculated CTP digits and -294 for runs with CTP digits from runs with CTP readout?

ekryshen: @mfasDa Thanks for the suggestion, I've added relevant configurable.

mfasDa:
@ekryshen Do you think this is ready for merging? How do you think about the proposal from @mpuccio to take the shift from the CCDB instead?

ekryshen:
This is just a temporary workaround so CCDB solution might be too heavy. I've just hardcoded relevant run ranges and kept possibility for custom triggerBcShift settings.

ddobrigk (approved):

PR #2166, opened 2023-03-09
some update for runbyrun QA

zjxiongOvO:


iarsene:
Hi Zhenjun,
As I also mentioned to Xiaozhi at some point, I would like to avoid this kind of implementation for the run lists. However, since for now we have no alternative, I will allow it temporarily. But we should move to a CCDB based solution.

iarsene (approved):

XiaozhiBai:
HI Ionut

I know,  this is a temporary solution。
Best
Xiaozhi

PR #2175, opened 2023-03-09
Specify minimal set of GH Action permissions

TimoWilken:
For security and to silence errors from MegaLinter.

Cc: @vkucera


TimoWilken:
This passes MegaLinter, merging.

PR #2181, opened 2023-03-09
Adding an alternative filterPP task which uses collision-track associations

iarsene:
This is done in a similar fashion as the collision-track associator task in PWG-HF, but adapted to DQ.
Once this task is commisioned, it will replace the filterPP task.


dsekihat (approved):

iarsene:
Hi Jan, i will have a look in the common task now. But the reason i did not used directly the hf task was because in DQ i can still optimize quite a bit because we filter tracks a lot. But i will think how can all this be done in common, since the algorithm is very similar

PR #2184, opened 2023-03-10
PWGHF: Fix B0 workflow and modify PID selections in Dplus selector

AlexBigO:
Hey @fgrosa @iouribelikov  !

Here are the fixes applied to the B0 workflow, everything works (I get filled histograms in output of the execution command line) but there still might be some mistakes :)

One of the main modifications I applied to the candidate creator is removing the double loop on D+/D-: I now check the sign of `sign(D) * sign(Pi)` to only keep pairs that have opposite electric charge.

Also, the `processMc` functions work as well.

-----------------------------------------------------------------------------------------
List of fixes:

- `candidateSelectorDpluToPiKPi.cxx`
  - modify PID selection (put in a function): TpcOrTof --> TpcAndTof + add configurable to accept PIDNotApplicable status (default value set to `true`)
- `candidateCreatorB0.cxx`
  - [loop over D- & loop over Pi+] + [loop over D+ & loop over Pi-] --> reject DPi pairs with same electric charge signs + loop over pions
  - subscribe to `HfCand3Prong` in `processMc` and bind indices
- `candidateSelectorB0ToDPi.cxx`
  - pT cut on D candidate: "pT D^{#minus}" --> "pT D" (defined in `SelectorCuts.h`)
  - subscribe to `TracksWithSel` instead of `BigTracksPID`
  - add PID selection (TpcAndTof) + add configurable to accept `PIDNotApplicable` status (default value set to `true`)
  - sync settings of `selectionFlagD` (from B0 creator) and `usePid` [temporary solution using AOD table]
- `taskB0.cxx`
  -  subscribe to `TracksWithSel` instead of `BigTracksPID`
  - fix condition on `candidate.hfflag()` in `processMc` (a `!` was missing in front of `TESTBIT` in the if loop)

### DIFF
--- a/PWGHF/D2H/Tasks/taskB0.cxx
+++ b/PWGHF/D2H/Tasks/taskB0.cxx
@@ -60,61 +60,7 @@ struct HfTaskB0 {
     registry.add("hDecLenXYErr", "B^{0} candidates;B^{0} candidate decay length xy error (cm);entries", {HistType::kTH2F, {{100, 0., 1.}, {(std::vector<double>)binsPt, "#it{p}_{T} (GeV/#it{c})"}}});
     registry.add("hIPProd", "B^{0} candidates;B^{0} candidate impact parameter product;entries", {HistType::kTH2F, {{100, -0.5, 0.5}, {(std::vector<double>)binsPt, "#it{p}_{T} (GeV/#it{c})"}}});
     registry.add("hInvMassD", "B^{0} candidates;prong0, D^{#minus} inv. mass (GeV/#it{c}^{2});entries", {HistType::kTH2F, {{500, 0, 5}, {(std::vector<double>)binsPt, "#it{p}_{T} (GeV/#it{c})"}}});
-  }
 
-  void process(aod::Collision const& collision, soa::Filtered<soa::Join<aod::HfCandB0, aod::HfSelB0ToDPi>> const& candidates, soa::Join<aod::HfCand3Prong, aod::HfSelDplusToPiKPi> const&, aod::BigTracks const&)
-  {
-    for (auto const& candidate : candidates) {
-      if (TESTBIT(candidate.hfflag(), hf_cand_b0::DecayType::B0ToDPi)) {
-        continue;
-      }
-      if (yCandMax >= 0. && std::abs(yB0(candidate)) > yCandMax) {
-        continue;
-      }
-
-      auto candD = candidate.prong0_as<soa::Join<aod::HfCand3Prong, aod::HfSelDplusToPiKPi>>();
-      auto candPi = candidate.prong1_as<aod::BigTracks>();
-
-      registry.fill(HIST("hMass"), invMassB0ToDPi(candidate), candidate.pt());
-      registry.fill(HIST("hPtCand"), candidate.pt());
-      registry.fill(HIST("hPtProng0"), candidate.ptProng0());
-      registry.fill(HIST("hPtProng1"), candidate.ptProng1());
-      registry.fill(HIST("hIPProd"), candidate.impactParameterProduct(), candidate.pt());
-      registry.fill(HIST("hDecLength"), candidate.decayLength(), candidate.pt());
-      registry.fill(HIST("hDecLengthXY"), candidate.decayLengthXY(), candidate.pt());
-      registry.fill(HIST("hd0Prong0"), candidate.impactParameter0(), candidate.pt());
-      registry.fill(HIST("hd0Prong1"), candidate.impactParameter1(), candidate.pt());
-      registry.fill(HIST("hCPA"), candidate.cpa(), candidate.pt());
-      registry.fill(HIST("hEta"), candidate.eta(), candidate.pt());
-      registry.fill(HIST("hRapidity"), yB0(candidate), candidate.pt());
-      registry.fill(HIST("hImpParErr"), candidate.errorImpactParameter0(), candidate.pt());
-      registry.fill(HIST("hImpParErr"), candidate.errorImpactParameter1(), candidate.pt());
-      registry.fill(HIST("hDecLenErr"), candidate.errorDecayLength(), candidate.pt());
-      registry.fill(HIST("hDecLenXYErr"), candidate.errorDecayLengthXY(), candidate.pt());
-      if (candPi.sign() < 0) {
-        registry.fill(HIST("hInvMassD"), invMassDplusToPiKPi(candD), candidate.pt());
-      }
-    } // candidate loop
-  }   // process
-};    // struct
-
-/// B0 MC analysis and fill histograms
-struct HfTaskB0Mc {
-  Configurable<int> selectionFlagB0{"selectionFlagB0", 1, "Selection Flag for B0"};
-  Configurable<double> yCandMax{"yCandMax", 0.8, "max. cand. rapidity"};
-  Configurable<std::vector<double>> binsPt{"binsPt", std::vector<double>{hf_cuts_b0_to_d_pi::vecBinsPt}, "pT bin limits"};
-
-  Filter filterSelectCandidates = (aod::hf_sel_candidate_b0::isSelB0ToDPi >= selectionFlagB0);
-
-  HistogramRegistry registry{
-    "registry",
-    {{"hPtRecSig", "B0 candidates (matched);candidate #it{p}_{T} (GeV/#it{c});entries", {HistType::kTH1F, {{300, 0., 30.}}}},
-     {"hPtRecBg", "B0 candidates (unmatched);candidate #it{p}_{T} (GeV/#it{c});entries", {HistType::kTH1F, {{300, 0., 30.}}}},
-     {"hPtGenSig", "B0 candidates (matched);candidate #it{p}_{T}^{gen.} (GeV/#it{c});entries", {HistType::kTH1F, {{300, 0., 10.}}}},
-     {"hPtGen", "MC particles (matched);candidate #it{p}_{T} (GeV/#it{c});entries", {HistType::kTH1F, {{300, 0., 30.}}}}}};
-
-  void init(o2::framework::InitContext&)
-  {
     registry.add("hEtaGen", "MC particles (matched);B^{0} candidate #it{#eta}^{gen};entries", {HistType::kTH2F, {{100, -2., 2.}, {(std::vector<double>)binsPt, "#it{p}_{T} (GeV/#it{c})"}}});
     registry.add("hYGen", "MC particles (matched);B^{0} candidate #it{y}^{gen};entries", {HistType::kTH2F, {{100, -2., 2.}, {(std::vector<double>)binsPt, "#it{p}_{T} (GeV/#it{c})"}}});
     registry.add("hPtProng0Gen", "MC particles (matched);prong 0 (D^{#minus}) #it{p}_{T}^{gen} (GeV/#it{c});entries", {HistType::kTH2F, {{100, 0., 10.}, {(std::vector<double>)binsPt, "#it{p}_{T} (GeV/#it{c})"}}});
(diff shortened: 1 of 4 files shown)


PWGHF/D2H/Tasks/taskB0.cxx
```diff
@@ -159,28 +105,71 @@ struct HfTaskB0Mc {
...
+    for (auto const& candidate : candidates) {
+      if (!TESTBIT(candidate.hfflag(), hf_cand_b0::DecayType::B0ToDPi)) {
+        continue;
+      }
+      if (yCandMax >= 0. && std::abs(yB0(candidate)) > yCandMax) {
+        continue;
+      }
+
+      auto candD = candidate.prong0_as<soa::Join<aod::HfCand3Prong, aod::HfSelDplusToPiKPi>>();
+      auto candPi = candidate.prong1_as<TracksWithSel>();
```
iouribelikov, line 127: This line generates a warning, that is treated as an error on my Mac (and so the compilation breaks).
Because, candPi is never used in the code below.  Should we (for the moment) comment this line out ?

PWGHF/TableProducer/candidateCreatorB0.cxx
```diff
@@ -116,220 +115,146 @@ struct HfCandidateCreatorB0 {
     df3.setMaxDZIni(maxDZIni);
     df3.setMinParamChange(minParamChange);
     df3.setMinRelChi2Change(minRelChi2Change);
-    df3.setUseAbsDCA(useAbsDCA);
-    df3.setWeightedFinalPCA(useWeightedFinalPCA);
```
fgrosa, line 120: I would let it with the configurables; in fact, it would be better to have the flexibility to use the same configuration used in the `track-index-skim-creator` for the D+ (while I agree to hard code it for the B0, since this is the only working configuration for vertices built with "pseudo" tracks)

fgrosa (commented):
Ciao @AlexBigO, thanks a lot! Also for me it's fine, I just have a minor comment (see below)

PWGHF/D2H/Tasks/taskB0.cxx
AlexBigO: Thanks for the comment, I commented the line :)

PWGHF/TableProducer/candidateCreatorB0.cxx
AlexBigO: Indeed, thanks. I must have been working with a former version that did not take into account the PR of Fabio from 2 weeks ago.

AlexBigO:
I had a question I forgot to ask. In the selector, I currently subscribe to `aod::HfCandB0` via ` aod::HfCandB0 const& hfCandB0s` in `process(...)` and then loop over the candidates. Do you think it would be better if I simply subscribed to an iterator of the table: `aod::HfCandB0::iterator const& hfCandB0` which should automatically loop over all the candidates?

PWGHF/D2H/Tasks/taskB0.cxx
```diff
@@ -159,28 +105,71 @@ struct HfTaskB0Mc {
...
+      if (!TESTBIT(candidate.hfflag(), hf_cand_b0::DecayType::B0ToDPi)) {
         continue;
       }
       if (yCandMax >= 0. && std::abs(yB0(candidate)) > yCandMax) {
         continue;
       }
-      auto candD = candidate.prong0_as<aod::HfCand3Prong>();
+      // auto candD = candidate.prong0_as<aod::HfCand3Prong>();
+      auto candD = candidate.prong0_as<soa::Join<aod::HfCand3Prong, aod::HfCand3ProngMcRec>>();
       if (TESTBIT(std::abs(candidate.flagMcMatchRec()), hf_cand_b0::DecayType::B0ToDPi)) {
```
vkucera, line 165: Are you expecting to use `candD` also for unmatched candidates? If not, I suggest to move it inside the `TESTBIT` block, since otherwise it is executed unnecessarily taking some extra CPU spent on the casting, AFAIK.
In any case, it is an unused variable now and will not pass the compilation checks.

PWGHF/TableProducer/candidateCreatorB0.cxx
```diff
@@ -120,216 +120,143 @@ struct HfCandidateCreatorB0 {
...
+        }
+        // reject pi and D with same sign
+        if (trackPion.sign() * track0.sign() > 0) {
+          continue;
+        }
 
-          hMassB0ToDPi->Fill(massDPi);
+        hPtPion->Fill(trackPion.pt());
+        array<float, 3> pVecPion = {trackPion.px(), trackPion.py(), trackPion.pz()};
+        ;
```
vkucera, line 183: This looks like a remnant.

vkucera:
Could you please a brief list of the fixed bugs in the PR description?

fgrosa:
> I had a question I forgot to ask. In the selector, I currently subscribe to `aod::HfCandB0` via ` aod::HfCandB0 const& hfCandB0s` in `process(...)` and then loop over the candidates. Do you think it would be better if I simply subscribed to an iterator of the table: `aod::HfCandB0::iterator const& hfCandB0` which should automatically loop over all the candidates?

Indeed if I did not overlook anything, we can directly pass the iterator, there is no need to have the loop inside the `process` function. Thanks!

PWGHF/D2H/Tasks/taskB0.cxx
AlexBigO: `candD` is called in both `if` (signal) and `else` (background) blocks, that's why it is defined above.

PWGHF/D2H/Tasks/taskB0.cxx
```diff
@@ -223,7 +212,7 @@ struct HfTaskB0Mc {
     for (auto const& particle : particlesMC) {
       if (TESTBIT(std::abs(particle.flagMcMatchGen()), hf_cand_b0::DecayType::B0ToDPi)) {
 
-        auto yParticle = RecoDecay::y(array{particle.px(), particle.py(), particle.pz()}, RecoDecay::getMassPDG(pdg::Code::kB0));
+        auto yParticle = RecoDecay::y(array{particle.px(), particle.py(), particle.pz()}, RecoDecay::getMassPDG(particle.pdgCode()));
```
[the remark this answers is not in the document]
AlexBigO: I checked and it does not make any difference indeed. It must be a remnant of some debugging tests.

PWGHF/TableProducer/candidateCreatorB0.cxx
AlexBigO: Thanks, the cut on invariant mass window disappeared for some reason. I think I had different versions of the code that overlapped for some reason, I guess some bad manipulation in git.

PWGHF/TableProducer/candidateCreatorB0.cxx
```diff
@@ -344,7 +271,7 @@ struct HfCandidateCreatorB0Expressions {
       flag = 0;
       origin = 0;
       debug = 0;
-      auto candD = candidate.prong0();
+      auto candD = candidate.prong0_as<aod::HfCand3Prong>();
```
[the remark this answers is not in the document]
AlexBigO: You are right, it also works without this extra information.

AlexBigO:
I remembered that I had another question. In `candidateCreatorB0.cxx`, I put two `FIXME` comments:

- line 125:
```
if (!TESTBIT(candD.hfflag(), hf_cand_3prong::DecayType::DplusToPiKPi)) continue;
```
This condition should be embedded in the filter applied on `candsD` because it gives a `flag=0` if not verified in [`candidateSelectorDplusToPiKPi.cxx`](https://github.com/AliceO2Group/O2Physics/blob/master/PWGHF/TableProducer/candidateSelectorDplusToPiKPi.cxx#L123). Am I right?
- line 130: 
```
if (candD.isSelDplusToPiKPi() >= selectionFlagD) { ...}
```
This `if` loop is a remnant and is obviously not needed as `candsD` are already filtered (I will remove it in the next commit).

vkucera:
> 

AFAIK, bit operations are not supported in filters. (It might not be the case anymore.)

fgrosa:
> > 
> 
> AFAIK, bit operations are not supported in filters. (It might not be the case anymore.)

Ciao @vkucera we filter D+ candidates according to `aod::hf_sel_candidate_dplus::isSelDplusToPiKPi >= selectionFlagD`, and the flag is `0` if the `TESTBIT` is false in the D+ selector (see https://github.com/AliceO2Group/O2Physics/blob/977f7cae6181b1c674be1ea878397c9f615a9bda/PWGHF/TableProducer/candidateSelectorDplusToPiKPi.cxx#L123). So there is no need to retest it, unless one sets `selectionFlagD==0` in the configurable. But I guess that there is no reason to do it, that's what @AlexBigO meant I guess

vkucera:
> > > 
> > 
> > 
> > AFAIK, bit operations are not supported in filters. (It might not be the case anymore.)
> 
> Ciao @vkucera we filter D+ candidates according to `aod::hf_sel_candidate_dplus::isSelDplusToPiKPi >= selectionFlagD`, and the flag is `0` if the `TESTBIT` is false in the D+ selector (see
> 
> https://github.com/AliceO2Group/O2Physics/blob/977f7cae6181b1c674be1ea878397c9f615a9bda/PWGHF/TableProducer/candidateSelectorDplusToPiKPi.cxx#L123
> ). So there is no need to retest it, unless one sets `selectionFlagD==0` in the configurable. But I guess that there is no reason to do it, that's what @AlexBigO meant I guess

Indeed.

AlexBigO:
Indeed, it is what I meant. Thanks, I will remove these two `if` loops in the next commit.

AlexBigO:
Hi @lvermunt , thanks for the feedback.

- Alright, I will change it both in B0 and Dplus selectors. Should I also modify [this condition](https://github.com/AliceO2Group/O2Physics/blob/master/PWGHF/TableProducer/candidateSelectorDplusToPiKPi.cxx#L155) by checking ` != TrackSelectorPID::Status::PIDAccepted` instead of `==TrackSelectorPID::Status::PIDRejected` (as there is also the [`Status::PIDNotApplicable`](https://github.com/AliceO2Group/O2Physics/blob/master/Common/Core/TrackSelectorPID.h#L487)?
- I agree, it is better this way.

lvermunt (dismissed):

PWGHF/TableProducer/candidateSelectorB0ToDPi.cxx
```diff
@@ -121,40 +125,79 @@ struct HfCandidateSelectorB0ToDPi {
...
+        hfSelB0ToDPiCandidate(statusB0ToDPi);
+        // LOGF(info, "B0 candidate selection failed at topology selection");
         continue;
       }
+      SETBIT(statusB0ToDPi, aod::SelectionStep::RecoTopol); // RecoTopol = 1 --> statusB0ToDPi = 3
+
+      // track-level PID selection
+      if (usePid) {
+        if (!TESTBIT(candD.isSelDplusToPiKPi(), aod::SelectionStep::RecoPID)) { // safety
+          LOG(warning) << "PID not enabled in DplusToPiKPi selector. Set selectionFlagD=7 in hf-candidate-creator-b0";
```
vkucera, line 181: I don't understand this part. The condition tests whether the D+ candidate passed the PID selection. Why does the warning mention enabling PID in the D+ selector and adjusting the selection flag in the B0 candidate creator?

PWGHF/TableProducer/candidateSelectorDplusToPiKPi.cxx
vkucera, line 1: This should ideally go in a separate PR since it is not touching the B0 workflow directly. If you prefer to keep it here, please update the PR title and description and also @fcatalan92 should be asked to check the changes.

PWGHF/D2H/Tasks/taskB0.cxx
```diff
@@ -152,69 +98,112 @@ struct HfTaskB0Mc {
...
-    registry.add("hThetaStarRecSig", "B^{0} candidates (matched);B^{0} #cos(#theta^{*});entries", {HistType::kTH2F, {{110, -1.1, 1.1}, {(std::vector<double>)binsPt, "#it{p}_{T} (GeV/#it{c})"}}});
-    registry.add("hThetaStarRecBg", "B^{0} candidates (unmatched);B^{0} #cos(#theta^{*});entries", {HistType::kTH2F, {{110, -1.1, 1.1}, {(std::vector<double>)binsPt, "#it{p}_{T} (GeV/#it{c})"}}});
+
+    registry.add("hPtRecSig", "B0 candidates (matched);candidate #it{p}_{T} (GeV/#it{c});entries", {HistType::kTH1F, {{300, 0., 30.}}});
+    registry.add("hPtRecBg", "B0 candidates (unmatched);candidate #it{p}_{T} (GeV/#it{c});entries", {HistType::kTH1F, {{300, 0., 30.}}});
+    registry.add("hPtGenSig", "B0 candidates (gen+rec);candidate #it{p}_{T}^{gen.} (GeV/#it{c});entries", {HistType::kTH1F, {{300, 0., 10.}}});
+    registry.add("hPtGen", "MC particles (generated);candidate #it{p}_{T} (GeV/#it{c});entries", {HistType::kTH1F, {{300, 0., 30.}}});
   }
 
+  using TracksWithSel = soa::Join<aod::BigTracksExtended, aod::TrackSelection>;
```
vkucera, line 113: `using` should be declared at the top, not between functions. See the guidelines.

PWGHF/D2H/Tasks/taskB0.cxx
```diff
@@ -152,69 +98,112 @@ struct HfTaskB0Mc {
...
+      registry.fill(HIST("hDecLenErr"), candidate.errorDecayLength(), ptCandB0);
+      registry.fill(HIST("hDecLenXYErr"), candidate.errorDecayLengthXY(), ptCandB0);
+      registry.fill(HIST("hInvMassD"), invMassDplusToPiKPi(candD), ptCandB0);
+    } // candidate loop
+  }   // process
+
+  /// B0 MC analysis and fill histograms
   void processMc(soa::Filtered<soa::Join<aod::HfCandB0, aod::HfSelB0ToDPi, aod::HfCandB0McRec>> const& candidates,
-                 soa::Join<aod::McParticles, aod::HfCandB0McGen> const& particlesMC, aod::BigTracksMC const& tracks, aod::HfCand3Prong const&)
+                 soa::Join<aod::McParticles, aod::HfCandB0McGen> const& particlesMC,
```
vkucera, line 151:
```suggestion
                 soa::Join<aod::McParticles, aod::HfCandB0McGen> const& particlesMc,
```

PWGHF/D2H/Tasks/taskB0.cxx
```diff
@@ -152,69 +98,112 @@ struct HfTaskB0Mc {
...
+                 aod::BigTracksMC const& tracks,
```
vkucera, line 152:
```suggestion
                 aod::BigTracksMC const&,
```
At least I don't see it used.

PWGHF/TableProducer/candidateSelectorB0ToDPi.cxx
```diff
@@ -19,6 +19,7 @@
 #include "PWGHF/Core/SelectorCuts.h"
 #include "PWGHF/DataModel/CandidateReconstructionTables.h"
 #include "PWGHF/DataModel/CandidateSelectionTables.h"
+#include "Common/Core/TrackSelectorPID.h"
```
vkucera, line 22: This should go up (sort).

vkucera (changes requested):

PWGHF/TableProducer/candidateSelectorB0ToDPi.cxx
vkucera: OK, I think I see what you mean now. You want to sync the application of PID between the B0 candidate creator (`selectionFlagD`) and the B0 candidate selector (`usePid`), but it has nothing to do with the PID selection in the D+ selector (which cannot be enabled or disabled).
If you run both B0 workflows (creator and selector), you could inspect the value of `selectionFlagD` from the selector and compare it with `usePid`. Of course this only works if you run both workflows.
Otherwise, one has to either apply both selections in the same workflow or just cross-check the settings.

vkucera: In any case, the warning message should be rephrased and not refer to the D+ selector.

AlexBigO: Yes, I agree. I had a poor choice of words. Indeed, whatever is the value of `selectionFlagD`, the PID selection is applied in Dplus selector, `selectionFlagD` only affects the `Filter` of Dplus candidates in the B0 creator.

So, do you think we should keep this approach or simply for for option 2 (where there is no more conflict between `selectionFlagD` and `usePid`)?

PWGHF/TableProducer/candidateSelectorDplusToPiKPi.cxx
AlexBigO: Thanks, I put the default value to `true`.

PWGHF/TableProducer/candidateSelectorB0ToDPi.cxx
vkucera: The values in `HFSELDPLUS` should definitely be used and I guess the point of using it in the B0 creator instead of in the selector is to reduce the number of candidates stored in the table and processed in the next steps. All that makes sense. Therefore, if you want to sync the settings of `selectionFlagD` and `usePid` (instead of keeping them independent), I would be in favour of adding the value inspection. I would suggest putting it all inside `init`, including the warning message, such that it is checked (and reported in the log) only once instead of for each candidate.

PWGHF/TableProducer/candidateSelectorDplusToPiKPi.cxx
vkucera: @AlexBigO please add a summary of these changes in the PR description.

PWGHF/TableProducer/candidateSelectorB0ToDPi.cxx
```diff
@@ -121,40 +125,85 @@ struct HfCandidateSelectorB0ToDPi {
...
+    }
+    if (acceptPIDNotApplicable && pidTrackPi == TrackSelectorPID::Status::PIDRejected) {
+      return false;
+    }
 
-      int statusB0 = 0;
+    return true;
+  }
+
+  using TracksPIDWithSel = soa::Join<aod::BigTracksPIDExtended, aod::TrackSelection>;
```
vkucera, line 144: This should also go up.

vkucera (changes requested):

PWGHF/TableProducer/candidateSelectorB0ToDPi.cxx
AlexBigO: Indeed, I missed this one, thanks!

PWGHF/TableProducer/candidateSelectorB0ToDPi.cxx
AlexBigO: Thanks for the reply @vkucera . But I don't know the syntax to get the value of the configurable `selectionFlagD` (defined in `candidateCreatorB0.cxx`) inside the `init` function of `candidateSelectorB0ToDPi.cxx` (I did not find it in the documentation).

vkucera: Can you try to take inspiration from this snippet? https://github.com/AliceO2Group/O2Physics/commit/80b870054a0480d1b01e70e97532b0db030975ce#diff-f39bae092160ecb233c90fe7d2e546e3d12b169c54dddc3db216aa27a400ea95L582-L587

AlexBigO: I am pretty sure because I changed the value of `selectionFlagD` in my `.json` and it still displayed the hardcoded "default" value.

vkucera: Hi @iouribelikov , thanks for the link. I didn't know. That is unfortunate. In that case, we cannot guarantee that the two configurations are synced.
@AlexBigO how do you want to detect the cases `usePid=false` and `selectionFlagD=7`?

vkucera: Indeed, as a temporary solution, it might work. Please add a `FIXME` comment with the link to the JIRA ticket at the relevant place in the B0 selector.

vkucera: In which function do you fill the table with `myflag`? I suppose that even filling in `init` should work, since there the actual values are used for sure.

AlexBigO: - in `candidateCreatorB0.cxx`
First, I made a mistake by defining `int myflag` before `init` or `process`. If I define it inside `init` or `process` I get the configurable value set in `.json`. So far so good.
Now, if I try filling my table in `init`, `hf-candidate-creator-b0` crashes during the execution
But it works if I fill the table in `process` (I put a condition so it only gets filled  only once).
- in `candidateSelectorB0ToDPi.cxx`
If I try to subscribe to my AOD table in `init` (on top of `InitContext const&`), O2Physics does not compile. But, if I subscribe to the table in `process` (as it is usually done), then it works.
- Recap
I fill and read my table in both `process` functions of B0 creator and selector, respectively. Is that acceptable as a temporary solution? Or is there a way to do it via `init`?

vkucera: Thanks for the tests. I think what you describe in the recap is acceptable for now.

vkucera (changes requested):
Thanks, looks good to me except for a few little things.
Please do not force pushing into your branch when the review is ongoing. It destroys the commit history.

vkucera (approved):

PR #2185, opened 2023-03-10
Secondary vertexing in dileptonHadron added to tableReader

torkjellsdatter:
**The purpose of this PR is to add FillDileptonTrackVertexing to the tableReader for the dileptonHadron-trask.** Main alterations is in this very same task (using the index of the legs to get the full track info of the electrons, in order to do the secondary vertexing, and to check that the hadron candidate is not the same as the legs). Therefore a secondary list DileptonsExtra is a parameter to dileptonHadron, and the covariance is added in order to do the secondary vertexing. I altered the default cuts used in the tableReader. The default cut cannot be empty as the hadron cut must be in the last bit, as the code is currently structured.

### DIFF
--- a/PWGDQ/TableProducer/tableMaker.cxx
+++ b/PWGDQ/TableProducer/tableMaker.cxx
@@ -115,7 +115,7 @@ struct TableMaker {
   HistogramManager* fHistMan;
 
   Configurable<std::string> fConfigEventCuts{"cfgEventCuts", "eventStandard", "Event selection"};
-  Configurable<std::string> fConfigTrackCuts{"cfgBarrelTrackCuts", "jpsiO2MCdebugCuts", "Comma separated list of barrel track cuts"};
+  Configurable<std::string> fConfigTrackCuts{"cfgBarrelTrackCuts", "jpsiO2MCdebugCuts2", "Comma separated list of barrel track cuts"};
   Configurable<std::string> fConfigMuonCuts{"cfgMuonCuts", "muonQualityCuts", "Comma separated list of muon cuts"};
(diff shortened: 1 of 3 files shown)


PWGDQ/TableProducer/tableMaker.cxx
```diff
@@ -115,7 +115,7 @@ struct TableMaker {
   HistogramManager* fHistMan;
 
   Configurable<std::string> fConfigEventCuts{"cfgEventCuts", "eventStandard", "Event selection"};
-  Configurable<std::string> fConfigTrackCuts{"cfgBarrelTrackCuts", "jpsiPID1", "Comma separated list of barrel track cuts"};
+  Configurable<std::string> fConfigTrackCuts{"cfgBarrelTrackCuts", "jpsiO2MCdebugCuts,kaonPID", "Comma separated list of barrel track cuts"};
```
iarsene, line 118: remove the kaonPID from the default

PWGDQ/Tasks/tableReader.cxx
```diff
@@ -743,7 +744,7 @@ struct AnalysisSameEventPairing {
   int fCurrentRun; // needed to detect if the run changed and trigger update of calibrations etc.
 
   OutputObj<THashList> fOutputList{"output"};
-  Configurable<string> fConfigTrackCuts{"cfgTrackCuts", "", "Comma separated list of barrel track cuts"};
+  Configurable<string> fConfigTrackCuts{"cfgTrackCuts", "jpsiPID1", "Comma separated list of barrel track cuts"};
```
iarsene, line 747: use the same default as in the previous tasks

iarsene (approved):

PR #2187, opened 2023-03-10
Cleaner logic for nSigma selection

fmazzasc:
@lbariogl


lbariogl (approved):

njacazio:
Great work. @fmazzasc could you please move this in the table producer in the next pr? Thanks and have a nice weekend

mpuccio:
Ciao @njacazio, why? This produces a table to be stored as an output, not a table to be consumed by other tasks

njacazio:
Ciao @mpuccio in principle there is no difference between the two cases

njacazio:
Actually, you have a good point here, we might consider moving the table definition to the datamodel

PR #2188, opened 2023-03-10
Protect from OOB iterator increment

mpuccio:


alibuild (approved):

PR #2201, opened 2023-03-13
Ml model update

saganatt:
Hi @fcatalan92 @ChSonnabend 

Recently I developed a Lc selector that uses BDT models developed by Fabrizio. I used the ML framework, but I needed some adjustments for BDT:
- different output types, not only floats
- BDT has output values in tensor[1], not tensor[0]. Currently, my solution is to use the outputs from the last tensor in a vector. In case someone else will have a yet different model with different outputs, we can again make adjustments.

Let me know if the changes there are fine for you.

### DIFF
--- a/Tools/ML/model.h
+++ b/Tools/ML/model.h
@@ -58,7 +58,7 @@ class OnnxModel
 
     try {
       auto outputTensors = mSession->Run(mInputNames, input, mOutputNames);
-      LOG(info) << "Number of output tensors: " << outputTensors.size();
+      LOG(debug) << "Number of output tensors: " << outputTensors.size();
       if (outputTensors.size() != mOutputNames.size()) {
         LOG(fatal) << "Number of output tensors: " << outputTensors.size() << " does not agree with the model specified size: " << mOutputNames.size();
       }
--- a/Tools/ML/model.h
+++ b/Tools/ML/model.h
@@ -48,8 +48,45 @@ class OnnxModel
 
   // Inferencing
   void initModel(std::string, bool = false, int = 0, uint64_t = 0, uint64_t = 0);
-  float* evalModel(std::vector<Ort::Value>);
-  float* evalModel(std::vector<float>);
+
+  // template methods -- best to define them in header
+  template <typename T>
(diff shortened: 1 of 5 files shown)


Tools/ML/model.h
```diff
@@ -48,8 +48,45 @@ class OnnxModel
...
+  // template methods -- best to define them in header
+  template <typename T>
+  T* evalModel(std::vector<Ort::Value>& input)
+  {
+    LOG(debug) << "Input tensor shape: " << printShape(input[0].GetTensorTypeAndShapeInfo().GetShape());
+    // assert(input[0].GetTensorTypeAndShapeInfo().GetShape() == getNumInputNodes()); --> Fails build in debug mode, TODO: assertion should be checked somehow
+
+    try {
+      auto outputTensors = mSession->Run(mInputNames, input, mOutputNames);
+      LOG(info) << "Number of output tensors: " << outputTensors.size();
```
ChSonnabend, line 61: Maybe this could also be changed to LOG(debug)

ChSonnabend:
Hi Maja,
very nice, indeed. The templated output types was something that I didn't think of anymore but that makes of course a lot of sense (maybe in the future someone wants). Also the error handling is much better handled now 👍  For the output values, I'm fine with what there is now (I hope everything works the same as before for the networks ;-) ). I guess we might never find a perfect solution which fits everyone, but for now it's good as it is :-)

Tools/ML/model.h
saganatt: ah, true, I was just printing it for myself locally.

saganatt:
Hi @jgrosseo @ddobrigk @iarsene could any of you approve the Common/ part?
These are only minor changes in the ML-related code after I forbade passing vector by value (copying) in the network `evalModel()` method. Christian already expressed his positive opinion above.

ddobrigk (approved):
thanks!

saganatt:
Oh, and I see that the `Tools/` part is also up to some of you, not Christian - perhaps @ddobrigk ?

ChSonnabend (approved):

saganatt:
Thanks a lot!

PR #2208, opened 2023-03-14
PWGHF: Fixes to Lc ML selector

saganatt:
Hi @vkucera 

These are the follow-up updates to make the selector work well with BDT.

I slightly changed the condition for rejecting candidates. Based on the threshold values, the conditions in lines 313 and 319-320 are more or less equivalent. Let's see on Hyperloop what we get, then we can think more how to use the BDT output.


saganatt:
Hi @vkucera can we merge this? :)

vkucera (commented):
Thanks @saganatt 
I just spotted two little things to fix:
- Histogram names should start with `h`.
- The named PDG codes should be used instead of literal numbers.
I leave it up to you if you prefer to merge this now as it is and make another PR then or add the fixes in this PR.

vkucera (approved):

PR #2209, opened 2023-03-14
[EMCAL-903] Select only clusters of either subdetector in max cluster vs. cluster plot

mfasDa:


mfasDa:
When looking at cluster pt/E vs max cluster pt/E the subdetector (EMCAL/DCAL) needs to be distinguished, otherwise one gets entries for cluster pt/E larger the max. cluster pt/E from the other side detector.

nzardosh (approved):

PR #2211, opened 2023-03-14
PID TPC, fix IU process function

njacazio:


ddobrigk (approved):

njacazio:
Thanks @ddobrigk with this one we should have full compatibility with IU tracks for PID base tasks (if requested) and for the TOF we can also use IU tracks to compute the NSigma values

alibuild (approved):

PR #2221, opened 2023-03-15
[EMCAL-903] Fill selected jet histograms also in case of neutral jet …

mfasDa:
…trigger


nzardosh (approved):

nzardosh (dismissed):

mfasDa:
Fixes in this PR:
- Histograms for selected jets were not filled in case the trigger type is neutral jet trigger
- Correlation histogram between triggers added for overlap studies

mfasDa:
The problem is that adding one histogram as output object exceed the amount of possible objects to handle by the framework  - the histograms need to be converted to the histogram registry.

nzardosh (dismissed):

PR #2227, opened 2023-03-16
Allow IU usage

njacazio:


ddobrigk (approved):
Test confirmed successful with Nicolò (thanks!), approving

PR #2231, opened 2023-03-16
Tutorials: a pair of example tasks where the configuration is inherited

aalkin:
A tutorial example of how a task can adjust its configuration based on the requirements of other tasks in the workflow.

@ddobrigk

### DIFF
--- a/Tutorials/CMakeLists.txt
+++ b/Tutorials/CMakeLists.txt
@@ -212,3 +212,13 @@ o2physics_add_dpl_workflow(hep-mc
                   SOURCES src/hepMC.cxx
                   PUBLIC_LINK_LIBRARIES O2::Framework
                   COMPONENT_NAME AnalysisTutorial)
+
+o2physics_add_dpl_workflow(preprocess
+                  SOURCES src/preprocess.cxx
+                  PUBLIC_LINK_LIBRARIES O2::Framework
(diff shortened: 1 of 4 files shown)


aalkin:
@TimoWilken I see that MegaLinter often has an error from `secretlint` check due to missing `branch_protection_rules.json`. Do we need the `secretlint` at all for O2Physics?

TimoWilken:
Good point @aalkin, I've disabled it (you might have to rebase or push to this PR to make GitHub pick up the updated MegaLinter workflow).

Cc: @vkucera

vkucera:
> Good point @aalkin, I've disabled it (you might have to rebase or push to this PR to make GitHub pick up the updated MegaLinter workflow).
> 
> Cc: @vkucera

Hi @TimoWilken , I cannot judge how serious the `secretlint` errors are, but please move the `DISABLE` line from the workflow file into the configuration file.

https://github.com/AliceO2Group/O2Physics/blob/master/.mega-linter.yml#L13

TimoWilken:
> Hi @TimoWilken , I cannot judge how serious the `secretlint` errors are, but please move the `DISABLE` line from the workflow file into the configuration file.

Done, thanks for the hint Vit!

ddobrigk (approved):
Many many thanks!

PR #2232, opened 2023-03-16
Bugfix; fixed variable types in table.

RafRaf11:


chiarapinto (approved):

PWGLF/Tasks/NucleiHistTask.cxx
```diff
@@ -42,7 +42,7 @@ namespace o2::aod
 namespace NucleiTableHist
 {
```
njacazio, line 43: Also here, consider in the future to split the table producer part in a task that is in the table producer and the data model accordingly. Thanks!

PR #2237, opened 2023-03-17
PWGHF: Move trackToCollision associator to common

jgrosseo:
Moving the task to Common as it is of general interest
Also:
- removing processAssocWithAmb (was there for cross-check)
- simplifying processStandardAssoc
- allowing processing of ambiguous tracks (not yet supported by candidate creator)
- use collision index stored with candidates in creators
- small update to validation task


jgrosseo:
@fgrosa @vkucera As discussed

(tagging also @mfaggin for the removal of the `amb` part)

mfaggin:
> @fgrosa @vkucera As discussed
> 
> (tagging also @mfaggin for the removal of the `amb` part)

Hi @jgrosseo , just to be sure: the thing is that, on top moving the amb. track - to - coll association outside the HF to make it common (absolutely reasonable in my opinion), the `processAssocWithAmb` functionalities are removed and the association now is only base on time (ie what `processAssocWithTime` does). Right?
It should be all, but please let me know if I'm neglecting something.

jgrosseo:
> > @fgrosa @vkucera As discussed
> > (tagging also @mfaggin for the removal of the `amb` part)
> 
> Hi @jgrosseo , just to be sure: the thing is that, on top moving the amb. track - to - coll association outside the HF to make it common (absolutely reasonable in my opinion), the `processAssocWithAmb` functionalities are removed and the association now is only base on time (ie what `processAssocWithTime` does). Right? It should be all, but please let me know if I'm neglecting something.

Correct! In addition, there is the traditional (Run 2) method.

Common/DataModel/CollisionAssociation.h
```diff
@@ -0,0 +1,30 @@
...
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+#ifndef COMMON_DATAMODEL_COLLISIONASSOCIATION_H_
+#define COMMON_DATAMODEL_COLLISIONASSOCIATION_H_
```
vkucera, line 12: Please add Doxygen documentation lines.

Common/TableProducer/trackToCollisionAssociator.cxx
```diff
@@ -0,0 +1,171 @@
...
+/// \author Fabrizio Grosa <fgrosa@cern.ch>, CERN
+/// \author Mattia Faggin <mfaggin@cern.ch>, University and INFN Padova
+
+#include "CommonConstants/LHCConstants.h"
+#include "Framework/AnalysisDataModel.h"
+#include "Framework/AnalysisTask.h"
+#include "Framework/ASoAHelpers.h"
+#include "Framework/runDataProcessing.h"
+#include "Common/DataModel/CollisionAssociation.h"
+#include "Common/DataModel/TrackSelectionTables.h"
```
vkucera, line 24: Please sort

Common/TableProducer/trackToCollisionAssociator.cxx
```diff
@@ -0,0 +1,171 @@
...
+using namespace o2::aod;
+
+struct TrackToCollisionAssociation {
+  Produces<TrackAssoc> association;
+
+  Configurable<float> nSigmaForTimeCompat{"nSigmaForTimeCompat", 4.f, "number of sigmas for time compatibility"};
+  Configurable<float> timeMargin{"timeMargin", 0.f, "time margin in ns added to uncertainty because of uncalibrated TPC"};
+  Configurable<bool> applyTrackSelForRun2{"applyMinimalTrackSelForRun2", false, "flag to apply minimal track selection for Run 2 in case of standard association"};
+  Configurable<bool> applyIsGlobalTrack{"applyIsGlobalTrackWoDCA", true, "flag to apply global track w/o DCA selection"};
+  Configurable<bool> usePVAssociation{"usePVAssociation", true, "if the track is a PV contributor, use the collision time for it"};
```
vkucera, line 38: Use the same name for the struct member and for the JSON string.

Common/TableProducer/trackToCollisionAssociator.cxx
```diff
@@ -0,0 +1,171 @@
...
+  Configurable<bool> includeUnassigned{"includeUnassigned", false, "consider also tracks which are not assigned to any collision"};
+  Configurable<bool> debug{"debug", false, "fill a table with flag to keep track of kind of track (PV contributor or ambiguous)"};
```
vkucera, line 40: Where does this enable/disable filling a table?

Common/TableProducer/trackToCollisionAssociator.cxx
```diff
@@ -0,0 +1,171 @@
...
+  using TracksWithSel = soa::Join<Tracks, TracksExtra, TrackSelection>;
+  using TracksWithSelFilter = soa::Filtered<TracksWithSel>;
+
+  void init(InitContext const&)
+  {
+    std::array<int, 2> doProcess = {doprocessAssocWithTime, doprocessStandardAssoc};
+    int doProcessSum = std::accumulate(doProcess.begin(), doProcess.end(), 0);
+    if (doProcessSum != 1) {
+      LOGP(fatal, "Exactly one process function should be enabled! Exit");
+    }
```
vkucera, line 52: Wouldn't just `if (doprocessAssocWithTime == doprocessStandardAssoc)` be enough?

Common/TableProducer/trackToCollisionAssociator.cxx
```diff
@@ -0,0 +1,171 @@
...
+
+    if (applyIsGlobalTrack && applyTrackSelForRun2) {
+      LOGP(fatal, "You cannot apply simultaneously Run 2 track selections and isGlobalWoDCA! Exit");
```
vkucera, line 55: Maybe `isGlobalWoDCA` can be replaced with the correct name appearing in the code?

Common/TableProducer/trackToCollisionAssociator.cxx
```diff
@@ -0,0 +1,171 @@
...
+  PROCESS_SWITCH(TrackToCollisionAssociation, processStandardAssoc, "Use standard track-to-collision association", true);
+};
+
+//________________________________________________________________________________________________________________________
+WorkflowSpec defineDataProcessing(ConfigContext const& cfgc)
+{
+  WorkflowSpec workflow{};
+  workflow.push_back(adaptAnalysisTask<TrackToCollisionAssociation>(cfgc));
+
+  return workflow;
```
vkucera, line 170: Can be a one-liner `return WorkflowSpec{...`.

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -40,6 +40,7 @@
 #include "ReconstructionDataFormats/Vertex.h" // for PV refit
 #include "ReconstructionDataFormats/V0.h"
 
+#include "Common/DataModel/CollisionAssociation.h"
```
vkucera, line 43: Please move up to keep alphabetical order.

vkucera (changes requested):
Hi @jgrosseo , please see my comments.

Common/TableProducer/trackToCollisionAssociator.cxx
vkucera: I guess the documentation line needs to be updated after removing the table.

Common/DataModel/CollisionAssociation.h
jgrosseo: Is this OK? 

/// \file CollisionAssociation.h
/// \brief Table definitions for track to collision associator

jgrosseo:
https://github.com/AliceO2Group/O2Physics/compare/master...jgrosseo:O2Physics:master

Common/TableProducer/trackToCollisionAssociator.cxx
vkucera: Isn't `(false == false)` equal to `true` (as well as `(true == true)`)?

jgrosseo: yes and `false == false` is invalid, too ;-)

vkucera: So the condition `if (doprocessAssocWithTime == doprocessStandardAssoc)` would throw a fatal in both invalid cases, right?

jgrosseo: Yes, you are right. Sorry...

alibuild (approved):

PR #2241, opened 2023-03-17
Fix public link libraries for EM filter

strogolo:
Update PUBLIC_LINK_LIBRARIES for EM filter to be used in the CEFP


alibuild (approved):
Auto-approving on behalf of @strogolo.

PR #2252, opened 2023-03-19
Cut away secondaries from material

mpuccio:


alibuild (approved):

PR #2254, opened 2023-03-19
Cut on TPC momentum instead of pt

mpuccio:


alibuild (approved):

PR #2270, opened 2023-03-23
Counters should use double

mpuccio:
The issue is the following:
```
root [0] float n = 4.1e9
(float) 4.10000e+09f
root [1] float newN = n + 1.f
(float) 4.10000e+09f
root [2] newN - n
(float) 0.00000f
```
for large runs, this is happening systematically. With doubles this is still fine until 10e15
```
root [0] double n = 4.1e15
(double) 4.1000000e+15
root [1] double newN = n + 1.
(double) 4.1000000e+15
root [2] newN - n
(double) 1.0000000
```

### DIFF
--- a/Common/TableProducer/eventSelection.cxx
+++ b/Common/TableProducer/eventSelection.cxx
@@ -23,7 +23,7 @@ using namespace o2::framework;
 #include "CommonConstants/LHCConstants.h"
 #include "Framework/HistogramRegistry.h"
 #include "DataFormatsFT0/Digit.h"
-#include "TH1F.h"
+#include "TH1D.h"
 using namespace evsel;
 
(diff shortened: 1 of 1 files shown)


ddobrigk (approved):
Looks very reasonable, thanks!

ktf:
Maybe I am naive, but counters should use 64bit integers, no?

ddobrigk:
> Maybe I am naive, but counters should use 64bit integers, no?

Yes, in principle you are totally right - the only drawback is that AFAIK ROOT does not have a 64-bit-integer-based histogram and usually that's why one resorts to doubles instead in these cases, when histogramming is involved.

ktf:
Yes, indeed... I forgot TH1I is 32 bit...

PR #2273, opened 2023-03-23
[DPG]: updates on qaMatchEff.cxx

mfaggin:
1. write a unique template function to handle all the `process`;
2. add process functions with IU tracks.

    CAVEAT: in the current version, the track-propagation workflow is still necessary, because the `makethn` is not `constexpr` and therefore the table `aod::TracksDCA` is always necessary. Use `minPropagationRadius==0` in this case.
    
    POSSIBLE IMPROVEMENT: put `makethn` as a `constexpr` variable and add further process functions to enable the `THnSparse` filling.
    ---> not a priority now!
    
3. remove useless newlines (I hope no more are there);
4. fix some names;
5. switch TH1F to TH1D to avoid floating point issues;

### DIFF
--- a/DPG/Tasks/AOTTrack/qaMatchEff.cxx
+++ b/DPG/Tasks/AOTTrack/qaMatchEff.cxx
@@ -21,6 +21,7 @@
 #include "CommonConstants/MathConstants.h"
 #include "Framework/AnalysisTask.h"
 #include "Framework/runDataProcessing.h"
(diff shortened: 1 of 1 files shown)


mfaggin:
@rosarioturrisi this is the PR I was mentioning before

mfaggin:
@njacazio have a look at it, if you like I've completed the developments for the moment

njacazio (approved):

PR #2280, opened 2023-03-27
PWGHF: Handle ambiguous tracks in B0 candidate creator and get bz from CCDB

AlexBigO:
Hi @fgrosa ! Here is the new structure of the code:

- Loop over `aod::Collision const& collisions`
- Loop over D candidates associated to this collision
- Repropagate the daughter(s) to this collision if needed
- Propagate D daughters to D decay vertex and get `pVec` of these daughters at this vertex (previsouly, I was getting the impulsion of the `track0,1,2` and not the `pVec` values (modified by `getPxPyPz`). 
Is it better to call `getPxPyPz` (from [trackUtilities.h](https://github.com/AliceO2Group/O2Physics/blob/master/Common/Core/trackUtilities.h#L97)) or `getPxPyPzGlo` (from [TrackParametrization](https://github.com/AliceO2Group/AliceO2/blob/dev/DataFormats/Reconstruction/src/TrackParametrization.cxx#L127)) ?
- Loop over tracks associated to this collision
- Propagate D and Pion candidates to B0 vertex (to get their impulsion at this vertex)
- Propagate D and Pion candidates to primary vertex to get their imapct parameter

Also, I had two remarks:
- I don't initialize`pVec0,1,2`  as I update it via `getPxPyPz`;
- I define `dca0,1,2` but don't initialize it because it is modified by `propagateToDCA` (when called) and because I don't use the dca of the D daughters anyway.
If this is okay, I will remove the comments on these lines.

---------------------------------------------
New modifications applied:
- choose `getPxPyPzGlo`
- modify initialization of `o2::dataformat::DCA` objects for D daughters
- get `bz` value from CCDB

### DIFF
--- a/PWGHF/TableProducer/candidateCreatorB0.cxx
+++ b/PWGHF/TableProducer/candidateCreatorB0.cxx
@@ -19,6 +19,7 @@
 #include "Framework/AnalysisTask.h"
 #include "DCAFitter/DCAFitterN.h"
 #include "Common/Core/trackUtilities.h"
+#include "Common/DataModel/CollisionAssociation.h"
 #include "ReconstructionDataFormats/DCA.h"
 #include "ReconstructionDataFormats/V0.h"
 #include "PWGHF/DataModel/CandidateReconstructionTables.h"
@@ -76,8 +77,18 @@ struct HfCandidateCreatorB0 {
 
   using TracksWithSel = soa::Join<aod::BigTracksExtended, aod::TrackSelection>;
 
+  // Filter filterSelectCollisions = (aod::hf_sel_collision::whyRejectColl == 0); // select only triggered events
+  // using SelectedCollisions = soa::Filtered<soa::Join<aod::Collisions, aod::HfSelCollision>>;
   Filter filterSelectTracks = (!usePionIsGlobalTrackWoDCA) || ((usePionIsGlobalTrackWoDCA) && requireGlobalTrackWoDCAInFilter());
   Filter filterSelectCandidates = (aod::hf_sel_candidate_dplus::isSelDplusToPiKPi >= selectionFlagD);
+
+  using TracksWithSelFiltered = soa::Filtered<TracksWithSel>;
+
+  using CandsDFiltered = soa::Filtered<soa::Join<aod::HfCand3Prong, aod::HfSelDplusToPiKPi>>;
+  Preslice<CandsDFiltered> candsDPerCollision = aod::track_association::collisionId;
+
+  //using TrackAssocSelFiltered = soa::Filtered<aod::TrackAssoc>; // soa::Filtered<soa::Join<aod::TrackAssoc, aod::HfSelTrack>>;
+  Preslice<aod::TrackAssoc> trackIndicesPerCollision = aod::track_association::collisionId;
 
   OutputObj<TH1F> hMassDToPiKPi{TH1F("hMassDToPiKPi", "D^{#minus} candidates;inv. mass (p^{#minus} K^{#plus} #pi^{#minus}) (GeV/#it{c}^{2});entries", 500, 0., 5.)};
   OutputObj<TH1F> hPtD{TH1F("hPtD", "D^{#minus} candidates;D^{#minus} candidate #it{p}_{T} (GeV/#it{c});entries", 100, 0., 10.)};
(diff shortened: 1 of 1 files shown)


PWGHF/TableProducer/candidateCreatorB0.cxx
```diff
@@ -140,119 +149,166 @@ struct HfCandidateCreatorB0 {
...
-      for (const auto& trackPion : tracksPion) {
-        // minimum pT selection
-        if (trackPion.pt() < ptPionMin || !isSelectedTrackDCA(trackPion)) {
-          continue;
+    for (const auto& collision : collisions) {
+      auto primaryVertex = getPrimaryVertex(collision);
+      auto covMatrixPV = primaryVertex.getCov();
+
+      if (ncol % 10000 == 0) {
+        LOG(info) << ncol << " collisions parsed";
```
fgrosa, line 159:
```suggestion
        LOG(debug) << ncol << " collisions parsed";
```

PWGHF/TableProducer/candidateCreatorB0.cxx
```diff
@@ -79,6 +80,13 @@ struct HfCandidateCreatorB0 {
   Filter filterSelectTracks = (!usePionIsGlobalTrackWoDCA) || ((usePionIsGlobalTrackWoDCA) && requireGlobalTrackWoDCAInFilter());
   Filter filterSelectCandidates = (aod::hf_sel_candidate_dplus::isSelDplusToPiKPi >= selectionFlagD);
 
+  using TracksWithSelFiltered = soa::Filtered<TracksWithSel>;
```
fgrosa, line 83: unfortunately we cannot filter them because otherwise the lookup of the track index (see `auto trackPion = trackId.track_as<TracksWithSelFiltered>();`) might fail if the track has been filtered out

PWGHF/TableProducer/candidateCreatorB0.cxx
```diff
@@ -140,119 +149,166 @@ struct HfCandidateCreatorB0 {
...
+        auto trackParCov0 = getTrackParCov(track0);
+        auto trackParCov1 = getTrackParCov(track1);
+        auto trackParCov2 = getTrackParCov(track2);
+
+        std::array<float, 3> pVec0; // = {track0.px(), track0.py(), track0.pz()};
+        std::array<float, 3> pVec1; // = {track1.px(), track1.py(), track1.pz()};
+        std::array<float, 3> pVec2; // = {track2.px(), track2.py(), track2.pz()};
+
+        o2::dataformats::DCA dca0;
+        // dca0.set(track0.dcaXY(), track0.dcaZ());
```
fgrosa, line 184: I would keep the initialisation (it will change only if `track.collisionId() != thisCollId`, otherwise the default one is the good one)

PWGHF/TableProducer/candidateCreatorB0.cxx
```diff
@@ -140,119 +149,166 @@ struct HfCandidateCreatorB0 {
...
+        auto track0 = candD.prong0_as<TracksWithSel>();
+        auto track1 = candD.prong1_as<TracksWithSel>();
+        auto track2 = candD.prong2_as<TracksWithSel>();
+        auto trackParCov0 = getTrackParCov(track0);
+        auto trackParCov1 = getTrackParCov(track1);
+        auto trackParCov2 = getTrackParCov(track2);
+
+        std::array<float, 3> pVec0; // = {track0.px(), track0.py(), track0.pz()};
+        std::array<float, 3> pVec1; // = {track1.px(), track1.py(), track1.pz()};
+        std::array<float, 3> pVec2; // = {track2.px(), track2.py(), track2.pz()};
```
fgrosa, line 181: I would keep the initialisation (it will change only if `track.collisionId() != thisCollId`, otherwise the default one is the good one)

PWGHF/TableProducer/candidateCreatorB0.cxx
```diff
@@ -140,119 +149,166 @@ struct HfCandidateCreatorB0 {
...
+
+        int indexTrack0 = track0.globalIndex();
+        int indexTrack1 = track1.globalIndex();
+        int indexTrack2 = track2.globalIndex();
+
+        auto trackIdsThisCollision = trackIndices.sliceBy(trackIndicesPerCollision, thisCollId);
+
+        for (const auto& trackId : trackIdsThisCollision) { // start loop over track indices associated to this collision
+          auto trackPion = trackId.track_as<TracksWithSelFiltered>();
+
```
fgrosa, line 236: Here we should add the selections that we cannot have in the filter anymore

fgrosa (changes requested):
Thanks @AlexBigO! There is only one thing to change (see comments below), for the rest is good for me

fgrosa (approved):

PWGHF/TableProducer/candidateCreatorB0.cxx
```diff
@@ -107,10 +137,11 @@ struct HfCandidateCreatorB0 {
     return true;
   }
 
-  void process(aod::Collision const& collision,
-               soa::Filtered<soa::Join<aod::HfCand3Prong, aod::HfSelDplusToPiKPi>> const& candsD,
+  void process(aod::Collisions const& collisions,
```
vkucera, line 140: What is preventing grouping by collision?

vkucera: Is it just the subscription to `TracksWithSel`?

vkucera (approved):

PR #2281, opened 2023-03-27
Adapting to new DPL arrow table slicing cache

aalkin:
This replaces the use of old `sliceByCached` method call, that uses internal cache of the table object, by a new call that relies on external cache. The external cache is maintained by the framework and automatically includes all cases that are defined by grouping signatures and `Preslice` declarations. 

This should increase performance, especially so for the tasks that use several partitions. As usual for such changes, it has to be validated by testing these tasks and comparing the output with the one produced by older version.


njacazio (dismissed):
Thanks a lot Anton!

saganatt (dismissed):
Thanks Anton!

aalkin:
I've updated this PR with changes to GroupedCombinations using the new cache API. Please, review the changes in the tasks that are relevant to you. This change will provide considerable performance boost to the tasks that rely on several groupings.

mhemmer-cern (approved):
PWGEM/PhotonMeson changes look good
Thanks a lot for this!

vkucera (approved):

ddobrigk (approved):
Thank you!

njacazio (approved):
Thanks Anton!

ddobrigk:
Can we maybe get @ktf to merge this? @ktf what do you think? (I think you got the permissions for that AFAIK). Thanks!!

ddobrigk:
Thanks @ktf - and have a nice easter break!

PR #2285, opened 2023-03-28
Fix case of 0 selected BCs

mpuccio:


alibuild (approved):

PR #2293, opened 2023-03-29
PWGDQ: Updating DQ fit library

lucamicheletti93:
Update of the python fit library


iarsene (approved):

TimoWilken:
@ktf No, I don't know, they should have run... But I've had a quick look at the changes and I can't see any obvious formatting issues, and the PR doesn't change any C++ code at all.

ktf:
Ah, maybe that's the issue? Do we handle correctly the formatting in case there are not any changes in C++ files?

TimoWilken:
As far as I can tell, the workflow should still run even if no C++ files are changed, it's just that some of the steps won't do anything (like the clang-format one).

PR #2296, opened 2023-03-30
PWGEM/PhotonMeson: quick fix to wrong return in pcm QC

dsekihat:
quick fix to wrong return in pcm QC


mhemmer-cern (approved):
Looks good!

m-c-danisch (approved):

PR #2307, opened 2023-04-01
PWGEM/PhotonMeson: add photon selection

dsekihat:
add photon selection


mhemmer-cern (approved):
Looks good from my side.
Maybe you could add some comments for some of the functions so it's easier to understand what they are doing.

mikesas (approved):

PR #2048, opened 2023-02-22
[PWGJE] Adding the functionality of hf-jets for inclusive jets

nzardosh:

PR #2315, opened 2023-04-02
[PWGJE: Preparation Validation Framework] Implementing reduced set of trackcuts as filtBit5 

jloemker:
Hello,
for the jet validation (performed on ESD's) from AliPhysics to O2Physics we want to recreate the Run2 jet-hybrid-tracks. They require less, and even rougher dca cuts. Once this PR is approved and merged the missing code for the jetfinder validation can be pushed to #2048 (or directly to O2Physics if both PR's are approved). 
These changes would facilitate and speed up our work - Many thanks in advance !
Johanna

*Ps @njacazio , @mfaggin I told you about this before and hope it is still okey (even if the filterbits are replaced at some point).

### DIFF
--- a/Common/Core/TrackSelectionDefaults.cxx
+++ b/Common/Core/TrackSelectionDefaults.cxx
@@ -107,4 +107,24 @@ TrackSelection getGlobalTrackSelectionRun3HF()
   return selectedTracks;
 }
 
+// Reduced default track selection for jet validation based on hybrid cuts for converted (based on ESD's from run 2) A02D's
+TrackSelection getJEGlobalTrackSelectionRun2()
+{
+  TrackSelection selectedTracks;
+  selectedTracks.SetTrackType(o2::aod::track::Run2Track); // Run 2 track asked by default
+  selectedTracks.SetMinNCrossedRowsTPC(70);
+  selectedTracks.SetMinNCrossedRowsOverFindableClustersTPC(0.8f);
+  selectedTracks.SetMaxChi2PerClusterTPC(4.f);
+  selectedTracks.SetRequireTPCRefit(true);
+  selectedTracks.SetRequireITSRefit(true);
+  selectedTracks.SetRequireHitsInITSLayers(1, {0, 1}); // one hit in any SPD layer
+  selectedTracks.SetMaxChi2PerClusterITS(36.f);
+  selectedTracks.SetPtRange(0.1f, 1e10f); // slightly lower initial cut for maximal flexibility
+  selectedTracks.SetEtaRange(-0.9f, 0.9f);
+  selectedTracks.SetMaxDcaXY(2.4f);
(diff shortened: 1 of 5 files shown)


Common/Core/TrackSelectionDefaults.cxx
```diff
@@ -107,4 +107,24 @@ TrackSelection getGlobalTrackSelectionRun3HF()
...
 
+// Reduced default track selection for jet validation based on hybrid cuts for converted (based on ESD's from run 2) A02D's
+TrackSelection getJEGlobalTrackSelectionRun2()
+{
+  TrackSelection selectedTracks;
+  selectedTracks.SetTrackType(o2::aod::track::Run2Track); // Run 2 track asked by default
+  selectedTracks.SetMinNCrossedRowsTPC(70);
+  selectedTracks.SetMinNCrossedRowsOverFindableClustersTPC(0.8f);
+  selectedTracks.SetMaxChi2PerClusterTPC(4.f);
+  selectedTracks.SetRequireTPCRefit(true);
```
njacazio, line 118: Here I would still use the main method and highlight what are the differences, i.e. you can start with   `TrackSelection selectedTracks = getGlobalTrackSelection();` and then build on that, we should do the same for the hf cut btw

jloemker:
@njacazio , I added the proposed change and it was build like a charm on my laptop. @nzardosh if this gets approved and merged I will open the PR for jets this evening. Many thanks !

njacazio (approved):
@arossi81 FYI

mfaggin:
Hi @jloemker , beware that for Run 3 you have to use the track-propagation workflow, not the track-extension. May you try again with it?

mfaggin:
> Hi @mfaggin I am testing the code for the run2 jet validation on converted AO2D's from run2. For my first attempt I could obtain a AnalysisResults.root that had exactly the cuts I wanted it to have. After changing the TrackSelectionDefaults.cxx (= last commit) the workflow crushes at the trackextension (and trackselection; both some issues in the terminal). But putting a value on the cut a wanted to avoid (locally `SetMaxDcaXYPtDep([](float` pt) { return 0.0105f + 0.0350f / pow(pt, 1.1f); })`) I am able to get a results file - But with 250 jets less. **help please**

Hi @johanna if you are running only on converted data then the track extension is correct.
Trying to understand your problem, if you want to mask the pt-dependent cut on dcaXY you can try to put something like

selectedTracks.SetMaxDcaXYPtDep([](float pt) { return 1e+10; });
selectedTracks.SetMaxDcaXY(2.4);

Let us know if this helps. However, if you do not manage to make the code work properly using the suggestions from @njacazio I'd suggest to go back to your 1st implementation, namely with all the cuts set by hand. Nicolò 's suggestion is valid and should be followed, however in the end the important thing is that the code works as you expect.

PR #2316, opened 2023-04-03
Update DPG codeowners

njacazio:
Reflects change in DPG coord. @mfaggin @iouribelikov

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -17,7 +17,7 @@
 /ALICE3 @alibuild @njacazio @ginnocen @vkucera
 /DPG @alibuild @chiarazampolli @noferini
 /DPG/Tasks/AOTEvent @alibuild @ekryshen @strogolo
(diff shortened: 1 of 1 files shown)


mfaggin:
Ciao @njacazio , actually I've nothing against keeping you in the busines, especially for the `qaEfficiency.cxx` task and the `PID folder. What do you think?`

iouribelikov:
> Ciao @njacazio , actually I've nothing against keeping you in the busines, especially for the `qaEfficiency.cxx` task and the `PID folder. What do you think?`

@mfaggin @njacazio Sure. I fully agree with Mattia.

njacazio:
I understand with this that I will never leave this business ;) for now I can stay but I will not approve PRs unless asked for or in the PID

mfaggin:
> I understand with this that I will never leave this business ;) for now I can stay but I will not approve PRs unless asked for or in the PID

Fair enough, thanks @njacazio

ddobrigk (approved):

PR #2321, opened 2023-04-03
dn-deta distribution for mft with mid-rapidity

hyungjun-lee:
dn-deta distribution for mft with mid-rapidity


aalkin:
I'm not sure why you are using MFT tracks here, especially with the eta cut - MFT is a forward detector, it will not have any tracks in central rapidity.

hyungjun-lee:
> I'm not sure why you are using MFT tracks here, especially with the eta cut - MFT is a forward detector, it will not have any tracks in central rapidity.

I ran my code locally, but the eta distribution in forward is normally tagged. the eta cut is not applied in forward (mfttrack). does it have different result when I ran this code in hyper loop?

hyungjun-lee:
> In the line 439 you have `Partition<aod::MFTTracks> tSample2 = nabs(aod::fwdtrack::eta) < estimatorEta;`
> 
> MFT acceptance is one-sided, so the symmetric cut is not good idea. You are also using the same cut for the central tracks, so you would only be able to run for either central, or MFT tracks, and not the two together.

oh, I missed the line. I will modify! thanks!

aalkin (approved):

PR #2330, opened 2023-04-04
PWGEM adding rights for Marvin Hemmer

mikesas:
@mhemmer-cern added to CODEOWNERS


mikesas:
Dear @ddobrigk , @jgrosseo , we would like to add Marvin Hemmer as a code approver for the sub-folder of the Photons and Mesons PAG of PWG-EM. I think he still has to be added to the project (if this is correct). Thanks a lot!

jgrosseo (approved):

PR #2369, opened 2023-04-11
Check also the evSel BC

mpuccio:


alibuild (approved):

PR #2372, opened 2023-04-11
[PWGJE] Validation task for Jetfinder on run2 data

jloemker:
Hi @nzardosh , @fkrizek, 

here is the basic jet-validation-qa task, performed on **global instead of hybrid tracks**. This should allow Filip to test the validation tools before I extend the task to handle MC from run2 (and run3 ?). I propose to change the cuts in AliPhysics accordingly, or to invite experts from tracks to a discussion to find a 'hybrid track solution' for ambiguous tracks in O2Physics. 

Cheerio,
Johanna

### DIFF
--- a/PWGJE/Tasks/jetvalidationqa.cxx
+++ b/PWGJE/Tasks/jetvalidationqa.cxx
@@ -12,8 +12,8 @@
 //  \author
 //	Johanna Lömker
 //	\since Dec 2022
-// The goal would be to set up the jet, track and event/collision QA's for the validation framework 
-// Staring with the hybrid tracks 
+// The goal would be to set up the jet, track and event/collision QA's for the validation framework
+// Staring with the hybrid tracks
 
 #include "Framework/runDataProcessing.h"
 #include "Framework/AnalysisTask.h"
@@ -31,9 +31,9 @@ using namespace o2;
 using namespace o2::framework;
 using namespace o2::framework::expressions;
 
-using TracksJE = soa::Join<aod::Tracks, aod::TracksExtra,  aod::TracksDCA, aod::TrackSelection>;
+using TracksJE = soa::Join<aod::Tracks, aod::TracksExtra, aod::TracksDCA, aod::TrackSelection>;
 
-struct jetTrackCollisionQa{
+struct jetTrackCollisionQa {
 
   HistogramRegistry mHistManager{"JetCollisionQAHistograms"};
   Configurable<int> nBins{"nBins", 200, "N bins in histos"};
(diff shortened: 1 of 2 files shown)


PWGJE/Tasks/jetvalidationqa.cxx
```diff
@@ -0,0 +1,191 @@
...
+    mHistManager.add("MCjetConstTrackEta", "MC inclusive jet constituent #eta ; #eta ", HistType::kTH1F, {{nBinsEta, -0.9, 0.9}});
+    // cross check the cuts from Run2Hybrid selection
+    mHistManager.add("MCselectedTrackPt", "MC hybrid track Pt ; p_{T} (GeV/#it{c})", HistType::kTH1F, {{nBinsPt, 0, 100}});
+    mHistManager.add("MCselectedTrackPhi", "MC hybrid track #phi ; #phi ", HistType::kTH1F, {{nBinsPhi, 0, 6.4}});
+    mHistManager.add("MCselectedTrackEta", "MC hybrid track #eta ; #eta ", HistType::kTH1F, {{nBinsEta, -0.9, 0.9}});
+  }
+  // add another process for MC studies !
+  void processData(soa::Join<aod::Collisions, aod::EvSels>::iterator const& collision, soa::Join<aod::Jets, aod::JetConstituents> const& jets, TracksJE const& tracks)
+  {
+    if (collision.posZ() > 10) {
```
nzardosh, line 113: This cut exists in the jet finder already so having it twice might cause confusion. If the value is tighter in the jet finder then the cut here will have no effect

PWGJE/Tasks/jetvalidationqa.cxx
```diff
@@ -0,0 +1,191 @@
...
+        leadingJetPhi = j.phi();
+        leadingJetEta = j.eta();
+      }
+
+      double leadingConstTrackPt = -1;
+      double leadingConstTrackPhi = -1;
+      double leadingConstTrackEta = -1;
+      // access jet constituents as tracks
+      for (auto& jct : j.tracks_as<TracksJE>()) {
+        if (!validationTracks.IsSelected(jct)) {
```
nzardosh, line 137: I am not sure about this if statement. The track is already clustered into the jet so removing it here wont change any jet proprties. Of course one can check if the jet finder really is making the correct cut but if you want to find the pt, eta and phi distributions of the jet constituents then this rejection would be out of place

nzardosh (changes requested):

jloemker:
@nzardosh , I included your changes and the fast formatting checks passed. For the other tests I need a first approval of the commit. Many thanks !

nzardosh (approved):

PR #2377, opened 2023-04-12
OTF: Add DelphesO2 LUT to onTheFlySmearer

njacazio:
@ddobrigk


alibuild (approved):

ALICE3/Core/DelphesO2Smearer.h
vkucera, line 1: The file and the class implemented in it should have the same name.

ALICE3/TableProducer/onTheFlyTracker.cxx
```diff
@@ -119,28 +149,48 @@ struct OnTheFlyTracker {
...
   {
     o2::dataformats::DCA dcaInfoCov;
     o2::dataformats::VertexBase vtx;
-
+    // First we compute the number of charged particles in the event
+    dNdEta = 0.f;
     for (const auto& mcParticle : mcParticles) {
-      auto pdg = std::abs(mcParticle.pdgCode());
-      if (pdg != kElectron && pdg != kMuonMinus && pdg != kPiPlus && pdg != kKPlus && pdg != kProton)
+      if (TMath::Abs(mcParticle.eta()) > multEtaRange) {
```
vkucera, line 160:
```suggestion
      if (std::abs(mcParticle.eta()) > multEtaRange) {
```

vkucera (changes requested):
Also please fix the C++ errors.

ALICE3/Core/DelphesO2Smearer.h
njacazio: Hi Vit, indeed you spotted a problem here. This is really a porting of the DelphesO2 class to O2Physics, already adding the Copyright is a change of pace wrt to the original code. Maybe we should think at another way of putting this into the O2Physics. @ddobrigk what do you think?

njacazio:
Hi Vit, as always  thanks for the review. For the c++ issue, I think it might be a false positive, I see also other PR's CI fail. For the file I will need to think about it

vkucera:
> Hi Vit, as always thanks for the review. For the c++ issue, I think it might be a false positive, I see also other PR's CI fail. For the file I will need to think about it

Hi @njacazio , I didn't mean the compilation check. I meant the `cpplint` errors (missing header guard, incorrect casting, redundant `;`).

njacazio:
Hi Vit, ok yes. Let's do it!

ALICE3/TableProducer/onTheFlyTracker.cxx
```diff
@@ -76,19 +77,12 @@ struct OnTheFlyTracker {
   o2::base::Propagator::MatCorrType matCorr = o2::base::Propagator::MatCorrType::USEMatCorrNONE;
 
   // Track smearer
-  TrackSmearer mSmearer;
+  o2::delphes::TrackSmearer mSmearer;
 
   void init(o2::framework::InitContext& initContext)
   {
     // Checking if the tables are requested in the workflow and enabling them
-    auto& workflows = initContext.services().get<RunningWorkflowInfo const>();
```
njacazio, line 71: @ddobrigk I use the central way of checking that the table gets used, hope it's ok

njacazio:
@vkucera I should have fixed all the issues with the PR, let me know

vkucera:
> 

Hi @njacazio , 

> @vkucera I should have fixed all the issues with the PR, let me know

Thanks @njacazio . What shall we do about the naming? Now the file is called `onTheFlyTracker` and the class is called `TrackSmearer`.

njacazio:
Hi Vit, indeed the header should be for the task struct, would you call it  `TrackSmearer` ?

vkucera:
> Hi Vit, indeed the header should be for the task struct, would you call it `TrackSmearer` ?

We should not mix tasks with class definitions. It makes it impossible to reuse the class anywhere else.
The task `OnTheFlyTracker` should be implemented in `TableProducer/onTheFlyTracker.cxx` and the class `TrackSmearer` should be a standalone class defined in `Core/TrackSmearer.{h|cxx}`.
Btw, why is `onTheFlyTracker.h` needed?

njacazio:
Ok, I did this with last commit
> Btw, why is `onTheFlyTracker.h` needed?
it was to separate the imported class from the rest, now it's in `Core`

ALICE3/Core/DelphesO2TrackSmearer.h
```diff
@@ -0,0 +1,211 @@
...
+// #include <map>
+
+using O2Track = o2::track::TrackParCov;
+
+namespace o2
+{
+namespace delphes
+{
+
+class TrackSmearer
```
vkucera, line 153:
```suggestion
class DelphesO2TrackSmearer
```

vkucera (changes requested):

vkucera:
> Ok, I did this with last commit
> 
> > Btw, why is `onTheFlyTracker.h` needed?
> 
> it was to separate the imported class from the rest, now it's in `Core` For the name I would like to keep to `DelphesO2` in the name of the file so that we avoid confusion

Thanks. The name of the file and of the class must match to make it obvious which class is defined inside that file. Furthermore, if, as you say, missing `DelphesO2` in the file name might cause confusion, then it should be in the class name as well for the same reason.

njacazio:
> > Ok, I did this with last commit
> > > Btw, why is `onTheFlyTracker.h` needed?
> > 
> > 
> > it was to separate the imported class from the rest, now it's in `Core` For the name I would like to keep to `DelphesO2` in the name of the file so that we avoid confusion
> 
> Thanks. The name of the file and of the class must match to make it obvious which class is defined inside that file. Furthermore, if, as you say, missing `DelphesO2` in the file name might cause confusion, then it should be in the class name as well for the same reason.

I see what you mean, but then it's not a port but it involves changing the imported code significantly. Maybe I could add a `using DelphesO2TrackSmearer`

njacazio:
> > > Ok, I did this with last commit
> > > > Btw, why is `onTheFlyTracker.h` needed?
> > > 
> > > 
> > > it was to separate the imported class from the rest, now it's in `Core` For the name I would like to keep to `DelphesO2` in the name of the file so that we avoid confusion
> > 
> > 
> > Thanks. The name of the file and of the class must match to make it obvious which class is defined inside that file. Furthermore, if, as you say, missing `DelphesO2` in the file name might cause confusion, then it should be in the class name as well for the same reason.
> 
> I see what you mean, but then it's not a port but it involves changing the imported code significantly. Maybe I could add a `using DelphesO2TrackSmearer`, what do you think?

e.g. like in https://github.com/AliceO2Group/O2Physics/pull/2377/files#diff-b4419c90b0ece9f1d49dd1260a78eecdf0c12723f3d4db3ed0a2e7ef0e19f6a1R211

vkucera:
> 

`using` would works as well of course, but what is the point of keeping the name `TrackSmearer`? Sorry if I am missing something.

njacazio:
> `TrackSmearer`

Ok then I would go for the `using`
Doing so we will be able to import from 
```
///                 DelphesO2/src/lutCovm.hh https://github.com/AliceO2Group/DelphesO2/blob/master/src/lutCovm.hh
///                 DelphesO2/src/TrackSmearer.cc https://github.com/AliceO2Group/DelphesO2/blob/master/src/TrackSmearer.cc
///                 DelphesO2/src/TrackSmearer.hh https://github.com/AliceO2Group/DelphesO2/blob/master/src/TrackSmearer.hh
```
with minimal changes

vkucera:
> > `TrackSmearer`
> 
> Ok then I would go for the `using` Doing so we will be able to import from
> 
> ```
> ///                 DelphesO2/src/lutCovm.hh https://github.com/AliceO2Group/DelphesO2/blob/master/src/lutCovm.hh
> ///                 DelphesO2/src/TrackSmearer.cc https://github.com/AliceO2Group/DelphesO2/blob/master/src/TrackSmearer.cc
> ///                 DelphesO2/src/TrackSmearer.hh https://github.com/AliceO2Group/DelphesO2/blob/master/src/TrackSmearer.hh
> ```
> 
> with minimal changes

OK, fine with me.

njacazio:
Fantastic, thanks a lot Vit! Done!

vkucera (approved):

alibuild (approved):

PR #2385, opened 2023-04-13
PWGMM: dndeta-mft: update to new API

aalkin:
I have missed a commit when preparing #2281, here it is as a separate PR.


alibuild (approved):

PR #2394, opened 2023-04-13
added ITS tagged histos and Q/pt

rosarioturrisi:


DPG/Tasks/AOTTrack/qaMatchEff.cxx
```diff
@@ -532,28 +674,61 @@ struct qaMatchEff {
       }
       //
       // all tracks, no conditions
+      if (track.hasITS() && isTrackSelectedITSCuts(track)) {
+        if constexpr (IS_MC) {
+          if (track.signed1Pt() > 0)
```
mfaggin, line 679: Here you can completely avoid the two `if`, because the data model provides already a method to obtain `q/pt`. Here (and everywhere below as well) you shall fill the histogram using `track.signed1Pt()` instead of `+-1./trackPt` (see for example here: https://github.com/AliceO2Group/O2Physics/blob/master/DPG/Tasks/AOTTrack/qaEventTrack.cxx#L1121)

mfaggin (changes requested):

mfaggin:
ciao @rosarioturrisi many thanks for the implementations. I've just asked a small improvement, easy to implement. Once ths is done, we can go ahead.

mfaggin (approved):
Thank you @rosarioturrisi ! I approve it

PR #2422, opened 2023-04-17
 [FT0FDD] task to analyse VdM data @Arvindiiti

arvindkhuntia:


PWGMM/Lumi/Tasks/LumiFDDFT0.cxx
```diff
@@ -0,0 +1,305 @@
...
+    ccdb->setURL(ccdburl);
+    ccdb->setCaching(true);
+    ccdb->setLocalObjectValidityChecking();
+    uint64_t now = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch()).count();
+    ccdb->setCreatedNotAfter(now);
+    mRunNumber = 0;
+  }
+
+  void process(soa::Join<aod::Collisions, aod::EvSels>::iterator const& collision, aod::FDDs const& fdds, aod::FT0s const& ft0s, aod::BCsWithTimestamps const&,
+               o2::soa::Join<o2::aod::Tracks, o2::aod::TrackSelection,
```
aalkin, line 148: There is no reason to subscribe to Tracks twice, especially since no filtering is actually applied and only `unfiltered_tracks` is used in the process function.

arvindkhuntia: This is done in the new PR. Many thanks

PR #2425, opened 2023-04-17
Add CI check for code layout of table declarations

TimoWilken:
Add a simple CI check to make sure that table declarations are only under `*/DataModel`, not in other subdirectories.

Unchanged files are not checked to avoid errors from existing declarations outside `*/DataModel`.

For now, this applies to *all* PWGs and other subdirs; let me know if it should be narrowed down to only check in particular PWGs.

Cc: @ddobrigk

### DIFF
--- a/.github/workflows/code-layout.yml
+++ b/.github/workflows/code-layout.yml
@@ -1,0 +1,38 @@
+---
+name: Code layout
+
+'on':
+  - pull_request_target
+
+permissions: {}
+
+jobs:
+  table-decls:
+    name: table declarations
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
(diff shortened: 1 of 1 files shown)


ddobrigk:
Hi Timo, thanks a lot for the draft! Let's keep it as draft for the time being, since I think more discussion is needed. For the per-PWG rules, this is exactly what I think should happen IMHO: I think the best way to go is that the coordination from each PWG decides on this being implemented as CI or not. System-wise, we will just offer the _option_ to do that, since this is what was requested, and we then keep PWG directory self-management autonomy. I will let you know once I have more info on what is desired for each PWG. Thanks again!

TimoWilken:
Sounds good, thanks @ddobrigk!

aalkin:
Occasionally table declarations are made directly in .cxx files, when the intent is to, for example, save a tree for local non-O2-based analysis. Such tables are not used by any tasks except those that write them. Should we enforce moving those to the data model headers?

PR #2429, opened 2023-04-18
TOF: Next PID TOF

njacazio:
- move to V2 parameter containers
- extend parameters usage, split per particle, charge effect


njacazio:
@noferini

njacazio:
@ddobrigk with this PR the data model gets simplified, I removed legacy columns that were introduced at the beginning of the PID work (and that are not used)

ddobrigk (approved):
Thanks a lot! Approved

njacazio:
Hi Vit, sure thanks for the reminder!

vkucera:
Thanks Nicolo!

njacazio:
Thanks to you! Let us know if A) it works and B) if it improves!

PR #2437, opened 2023-04-19
Add extension to TraCascData table

ddobrigk:
* adds an extension `Spawns` command to the cascade builder, required to extend the tracked cascade table with additional standard LF variables
* thanks a lot for finding this @qgp !


njacazio (approved):
Super, thx!

PR #2445, opened 2023-04-20
PWGHF: Add tutorial skeleton

vkucera:


vkucera:
Hi @TimoWilken , is clang format supposed to format also JSON files? It converts 4-space indentation (which is the style of the dumped O2 configuration) into 2-space indentation.

vkucera:
> Hi @vkucera, according to `clang-format --help`, it also supports JSON. At the moment, we don't restrict the files we pass to clang-format; we let it pick which ones it formats.
> 
> Are JSON files in this repo generally supposed to be human-readable/editable? If so, I'd tend towards letting clang-format reformat them, unless you have a strong opinion either way?

Formatting is fine but it would be better if there was a way to tell clang format to use 4 spaces for indentation so that the JSON in the repository can be directly compared with the one dumped by O2.

TimoWilken:
Fair enough @vkucera, see #2448.

vkucera:
> Fair enough @vkucera, see #2448.

Thanks a lot!

ddobrigk (approved):

alibuild (approved):

PR #2448, opened 2023-04-20
Use O2-like 4-space indents in JSON files

TimoWilken:
Configure clang-format to format JSON files like O2 would dump them, i.e. with 4-space indents. This way, we can track O2-generated JSON files in the repo directly, without having to reformat them.

Tested with an existing JSON file in the O2Physics repo; seems to work.

This came up in #2445.


vkucera:
Thanks @TimoWilken !

ddobrigk (approved):

PR #2456, opened 2023-04-21
PWGHF: Add possibility to split B0 workflow in two separate workflows

AlexBigO:
Hi @fgrosa @iouribelikov , here is the result of the "workflow splitting". The goal is to split the whole B0 workflow in two parts:

(usual workflow up to DPlusPi pair creation) `workflow1` --> `reducedAODs` --> `workflow 2` (B0 creator, selector and task)

`workflow 1` would run on Hyperloop, produce `reducedAODs` files that shall be consumed by `workflow 2`. This `workflow 2` could then run locally (or on Hyperloop) with way lower computation time (allowing faster testing/modifications of the B0-dedicated tasks).

Only collisions containing at least one DPi pair are saved in `reducedAODs`, this requires a careful handling of the total number of collisions studied for normalization purposes. In this spirit, we store the size of each `COLLISION` table called in `process` function of `ReducedDataCreatorDplusPi.cxx`.

New files introduced:
- `ReducedDataModel.h` : header file with all needed table and methods 
- `dataCreatorDplusPiReduced.cxx` : task to produce DPi pairs and fill `reducedAODs`
- `candidateCreatorB0Reduced.cxx` : task to create B0 candidates only consuming `reducedAODs`
- `candidateSelectorB0ToDPiReduced.cxx` : task to select B0 candidates consuming B0 creator output and `reducedAODs`
- `taskB0Reduced.cxx` : analysis task consuming B0 creator&selector outputs, along with  `reducedAODs`

---------------------------------------------------------------
New modifications:
- move files to D2H/ directory
- use AOD columns already defined 
- change naming conventions

Future (?) modification:
- define header containing methods that are common to the "standard" and "reduced" workflows


PWGHF/DataModel/ReducedDataModel.h
```diff
@@ -0,0 +1,247 @@
...
+                  pidtpc::TPCNSigmaEl,
+                  pidtpc::TPCNSigmaMu,
+                  pidtpc::TPCNSigmaPi,
+                  pidtpc::TPCNSigmaKa,
+                  pidtpc::TPCNSigmaPr,
+                  pidtof::TOFNSigmaEl,
+                  pidtof::TOFNSigmaMu,
+                  pidtof::TOFNSigmaPi,
+                  pidtof::TOFNSigmaKa,
+                  pidtof::TOFNSigmaPr);
```
fgrosa, line 156: Maybe we do not need all the hypotheses, i.e. Pi, K, and Pr might be enough (I fear that in our momentum range anyway we are not able to distinguish pions from e/mu)

PWGHF/TableProducer/reducedCandidateCreatorB0.cxx
```diff
@@ -0,0 +1,214 @@
...
+    df2.setWeightedFinalPCA(useWeightedFinalPCA);
+  }
+
+  void process(aod::HfReducedCollisions const& collisions,
+               aod::HfReducedCand3Prong const& candsD,
+               aod::HfReducedTracksWithSel const& tracksPion,
+               aod::HfOriginalCollisionsCounter const& collisionsCounter)
+  {
+    for (const auto& collisionCounter : collisionsCounter) {
+      registry.fill(HIST("hEvents"), 1, collisionCounter.originalAODSize());
```
fgrosa, line 85:
```suggestion
      registry.fill(HIST("hEvents"), 1, collisionCounter.originalCollisionCount());
```

PWGHF/TableProducer/reducedCandidateCreatorB0.cxx
```diff
@@ -0,0 +1,214 @@
...
+      if (ncol % 10000 == 0) {
+        LOG(info) << ncol << " collisions parsed";
+      }
+      ncol++;
+
+      /// Set the magnetic field from ccdb
+      bz = collision.bz();
+      df2.setBz(bz);
+
+      // FIXME : partitioning is slower than using if loops
```
fgrosa, line 104: Did you try with `Preslice` as well?
In general, we could consider to define an index for our reduced collisions that does not correspond to the original collision index. However this would require to store as `collisionId` in the pion and D+ tracks the one of the reduced collision and not the one of the original collision.

PWGHF/DataModel/ReducedDataModel.h
```diff
@@ -0,0 +1,247 @@
...
+namespace hf_reduced_collision
+{
+DECLARE_SOA_COLUMN(Bz, bz, float); //! Magnetic field in z-direction
+// keep track of the number of studied events (for normalization purposes)
+DECLARE_SOA_COLUMN(OriginalAODSize, originalAODSize, int); //! Size of COLLISION table processed
+
+} // namespace hf_reduced_collision
+
+DECLARE_SOA_TABLE(HfReducedCollisions, "AOD", "HFREDCOLLISION", //! Table with collision for reduced workflow
+                  track::CollisionId,                           // FIXME : soa::Index<> does not work event with DECLARE_EQUIVALENT_FOR_INDEX(aod::Collisions, aod::HfReducedCollisions)
```
fgrosa, line 47: We could consider to define an index for our reduced collisions that does not correspond to the original collision index. However this would require to store as `collisionId` in the pion and D+ tracks the one of the reduced collision and not the one of the original collision.

fgrosa (commented):
Hi @AlexBigO thanks a lot for the huge amount of work! I have still a couple of minor comments, on top of what we discussed offline.

PWGHF/TableProducer/reducedCandidateCreatorB0.cxx
```diff
@@ -0,0 +1,214 @@
...
+      auto primaryVertex = getPrimaryVertex(collision);
+      auto covMatrixPV = primaryVertex.getCov();
+
+      if (ncol % 10000 == 0) {
+        LOG(info) << ncol << " collisions parsed";
+      }
+      ncol++;
+
+      /// Set the magnetic field from ccdb
+      bz = collision.bz();
```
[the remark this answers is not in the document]
AlexBigO: Because I did not want to store the `BCsWithTimestamps` table needed to call `auto bc = collision.bc_as<aod::BCsWithTimestamps>();`.

PWGHF/DataModel/ReducedDataModel.h
AlexBigO: In our case, I only need the Pi information actually, as I only use PID on the Pion in B0 selector.

fgrosa: Ok! We can keep it like this for the moment and possibly improve it in a next PR

iouribelikov:
@fgrosa @vkucera  There seems to no more comments and suggestions. Also, all the checks have passes.  Could this PR be merged, please ?

fgrosa:
> @fgrosa @vkucera There seems to be no more comments and suggestions. Also, all the checks have passed. Could this PR be merged, please ?

Hi @iouribelikov I have no other comments beyond those that Alexandre already implemented so all good from my side! (But I also don't have approval permissions :))

fcolamar (commented):
Ciao @AlexBigO,
I don't know if Vit is currently reviewing the PR, but otherwise I would go forward so you can start with the tests and checks also on the Hyperloop and in case implement possible additional comments with a subsequent PR.
Anyway I have two general questions:
- Is the ReducedDataModel.h header intended to be used only for this channel (in such a case I'd change the filename making it more specific), or your plan is to keep it general and have other analyzers working on similar implementations for other particles/channels add their tables in this file?
- Is this new workflow intended to be implemented in addition to the existing one (which would remain in place, at least for a first stage to allow cross-checks and comparisons between the two approaches), or would it substitute the existing workflow which produces the B0 all-at-once? I see that with this PR you don't remove anything existing, so this goes in the former direction, but I don't know if you have further plans later on.
Thanks!

vkucera:
> 

I will check today.

AlexBigO:
>
Ciao @fcolamar !
> Ciao @AlexBigO, I don't know if Vit is currently reviewing the PR, but otherwise I would go forward so you can start with the tests and checks also on the Hyperloop and in case implement possible additional comments with a subsequent PR. Anyway I have a couple of general questions:
> 
>     * Is the ReducedDataModel.h header intended to be used only for this channel (in such a case I'd change the filename making it more specific), or your plan is to keep it general and have other analyzers working on similar implementations for other particles/channels add their tables in this file?
> 
Indeed, the plan is to make it general so other analyzers can "split" their workflow. I did not use a "general naming convention" just yet so it is more understandable and easier to debug for a first implementation :)
But, for instance, the MC tables can be generalized by removing `B0` in the name and one could add a `HfReducedCand2Prong` table (for `Bplus` for instance) and I guess this could do the job (along with other modifications).
>     * Is this new workflow intended to be implemented in addition to the existing one (which would remain in place, at least for a first stage to allow cross-checks and comparisons between the two approaches), or would it substitute the existing workflow which produces the B0 all-at-once? I see that with this PR you don't remove anything existing, so this goes in the former direction, but I don't know if you have further plans later on.
> 
The idea is to provide an alternative (lighter) workflow (ligher once `workflow1` is accomplished of course) without removing the current workflow that might be better for some dataset.
>     * If I got it correctly, you effectively store the full information of the collisions containing suitable candidates, so you retain the full flexibility of changing any kind of topological/PID selection (or, in a future, a ML model) when running on the reduced data, as if you were running the original workflow, is this correct?
> 
Yes that's it. We need to be careful with the (pre)-selections used (in our case, on `Dplus`) during `workflow1` as they will impact the content of our `reducedAODs`. But, once `workflow1` is accomplished, we have access to every information needed to vary selections (topological and PID) during `workflow2` dedicated to `B0` (without having to go through `workflow1` again).
> 
> Thanks!

PWGHF/TableProducer/reducedDataCreatorDplusPi.cxx
```diff
@@ -0,0 +1,483 @@
...
+    }
+    runNumber = 0;
+  }
+
+  /// Pion selection (D Pi <-- B0)
+  /// \param trackPion is a track with the pion hypothesis
+  /// \param track0 is prong0 of selected D candidate
+  /// \param track1 is prong1 of selected D candidate
+  /// \param track2 is prong2 of selected D candidate
+  /// \param indexD is an array of globalIndex of each D prong
```
vkucera, line 150: There is no argument `indexD`.

PWGHF/TableProducer/reducedDataCreatorDplusPi.cxx
```diff
@@ -0,0 +1,483 @@
...
+  void process(aod::Collisions const& collisions,
+               CandsDFiltered const& candsD,
+               aod::TrackAssoc const& trackIndices,
+               TracksPIDWithSel const&,
+               aod::BCsWithTimestamps const&)
+  {
+    // store configurables needed for B0 workflow
+    if (!isHfCandB0ConfigFilled) {
+      int mySelectionFlagD = selectionFlagD;
+      rowCandidateConfig(mySelectionFlagD);
```
vkucera, line 203: Why is this not filled directly?

PWGHF/TableProducer/reducedDataCreatorDplusPi.cxx
```diff
@@ -0,0 +1,483 @@
...
+      isHfCandB0ConfigFilled = true;
+    }
+
+    static int aodNumber = 0;
+    aodNumber++;
+
+    // handle normalization by the right number of collisions
+    hfCollisionCounter(collisions.tableSize());
+
+    static int ncol = 0;
```
vkucera, line 213:
```suggestion
    static int nCol = 0;
```

PWGHF/TableProducer/reducedDataCreatorDplusPi.cxx
```diff
@@ -0,0 +1,483 @@
...
+
+// event types
+enum Event : uint8_t {
+  Processed = 0,
+  noDPiSelected,
+  DPiSelected,
+  kNEvent
+};
+
+/// Selection of Dplus and Pions
```
vkucera, line 43: This is not a selection task

PWGHF/TableProducer/reducedDataCreatorDplusPi.cxx
```diff
@@ -0,0 +1,483 @@
...
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+/// \file reducedDataCreatorDplusPi.cxx
+/// \brief Selection of Dplus and Pions tracks
+/// \note
```
vkucera, line 14: This is not a selection workflow.

PWGHF/DataModel/ReducedDataModel.h
```diff
@@ -0,0 +1,249 @@
...
+                  hf_reduced_track_index::HfReducedCollisionId,
+                  HFTRACKPARCOV_COLUMNS);
+
+namespace hf_track_pid_with_sel
+{
+DECLARE_SOA_COLUMN(Pt, pt, float); //! Transverse momentum of the track in GeV/c
+} // namespace hf_track_pid_with_sel
+
+// table with all attributes needed to call getStatusTrackPIDTpcAndTof() in the selector task
+DECLARE_SOA_TABLE(HfReducedTracksPIDWithSel, "AOD", "HFREDTRACKPID", //! Table with PID track information for reduced workflow
```
vkucera, line 143:
```suggestion
DECLARE_SOA_TABLE(HfReducedTracksPidWithSel, "AOD", "HFREDTRACKPID", //! Table with PID track information for reduced workflow
```

PWGHF/TableProducer/reducedDataCreatorDplusPi.cxx
```diff
@@ -0,0 +1,483 @@
...
+  kNEvent
+};
+
+/// Selection of Dplus and Pions
+struct HfReducedDataCreatorDplusPi {
+  // Produces AOD tables to store track information
+  Produces<aod::HfReducedCollisions> hfReducedCollision;
+  Produces<aod::HfOriginalCollisionsCounter> hfCollisionCounter;
+  Produces<aod::HfReducedTracksWithSel> hfTrackPion;
+  Produces<aod::HfReducedTracksPIDWithSel> hfTrackPIDPion;
```
vkucera, line 49:
```suggestion
  Produces<aod::HfReducedTracksPIDWithSel> hfTrackPidPion;
```

PWGHF/TableProducer/reducedDataCreatorDplusPi.cxx
```diff
@@ -0,0 +1,483 @@
...
+      hfReducedCollision(collision.posX(), collision.posY(), collision.posZ(),
+                         collision.covXX(), collision.covXY(), collision.covYY(),
+                         collision.covXZ(), collision.covYZ(), collision.covZZ(),
+                         bz);
+    } // collision
+  }   // process
+};    // struct
+
+/// Extends the base table with expression columns and performs MC matching.
+struct HfReducedDataCreatorDplusPiExpressions {
```
vkucera, line 382: Fix the description and the task name. There are no expressions involved.

PWGHF/DataModel/ReducedDataModel.h
```diff
@@ -0,0 +1,249 @@
...
+DECLARE_SOA_COLUMN(PtProng0, ptProng0, float);   //! Transverse momentum of the track's prong0 in GeV/c
+DECLARE_SOA_COLUMN(YProng0, yProng0, float);     //! Rapidity of the track's prong0
+DECLARE_SOA_COLUMN(EtaProng0, etaProng0, float); //! Pseudorapidity of the track's prong0
+DECLARE_SOA_COLUMN(PtProng1, ptProng1, float);   //! Transverse momentum of the track's prong1 in GeV/c
+DECLARE_SOA_COLUMN(YProng1, yProng1, float);     //! Rapidity of the track's prong1
+DECLARE_SOA_COLUMN(EtaProng1, etaProng1, float); //! Pseudorapidity of the track's prong1
+} // namespace hf_b0_mc
+
+// table with results of reconstruction level MC matching
+DECLARE_SOA_TABLE(HfReducedDPiMcRec, "AOD", "HFREDDPIMCREC", //! Table with reconstructed MC information on DPi(<-B0) pairs for reduced workflow
```
vkucera, line 202:
```suggestion
DECLARE_SOA_TABLE(HfDPiMcRecReduced, "AOD", "HFDPIMCRECRED", //! Table with reconstructed MC information on DPi(<-B0) pairs for reduced workflow
```

PWGHF/DataModel/ReducedDataModel.h
```diff
@@ -0,0 +1,249 @@
...
+                  hf_cand_b0::Prong0Id,
+                  hf_track_index::Prong1Id,
+                  hf_cand_b0::FlagMcMatchRec,
+                  hf_cand_b0::OriginMcRec,
+                  hf_cand_b0::DebugMcRec,
+                  hf_b0_mc::PtMother);
+
+// Table with same size as HFCANDB0
+DECLARE_SOA_TABLE(HfReducedB0McRec, "AOD", "HFREDB0MCREC", //! Reconstruction-level MC information on B0 candidates for reduced workflow
```
vkucera, line 211:
```suggestion
DECLARE_SOA_TABLE(HfB0McRecReduced, "AOD", "HFB0MCRECRED", //! Reconstruction-level MC information on B0 candidates for reduced workflow
```

PWGHF/DataModel/ReducedDataModel.h
```diff
@@ -0,0 +1,249 @@
...
+                  hf_cand_b0::FlagMcMatchRec,
+                  hf_cand_b0::OriginMcRec,
+                  hf_cand_b0::DebugMcRec,
+                  hf_b0_mc::PtMother);
+
+DECLARE_SOA_TABLE(HfReducedB0McGen, "AOD", "HFREDB0MCGEN", //! Generation-level MC information on B0 candidates for reduced workflow
```
vkucera, line 217:
```suggestion
DECLARE_SOA_TABLE(HfB0McGenReduced, "AOD", "HFB0MCGENRED", //! Generation-level MC information on B0 candidates for reduced workflow
```

PWGHF/TableProducer/reducedDataCreatorDplusPi.cxx
```diff
@@ -0,0 +1,483 @@
...
+        }
+      }
+
+      // save information for B0 task
+      if (!TESTBIT(std::abs(flag), hf_cand_b0::DecayType::B0ToDPi)) {
+        continue;
+      }
+
+      auto ptParticle = particle.pt();
+      auto yParticle = RecoDecay::y(array{particle.px(), particle.py(), particle.pz()}, RecoDecay::getMassPDG(pdg::Code::kB0));
```
vkucera, line 455:
```suggestion
      auto yParticle = RecoDecay::y(std::array{particle.px(), particle.py(), particle.pz()}, RecoDecay::getMassPDG(pdg::Code::kB0));
```

vkucera (changes requested):
Hi @AlexBigO , in general, it would be useful to identify the common parts between the standard and the reduced workflows and avoid code duplication, especially in the candidate selection.

Another thing to consider is the location of this code. One could consider moving the entire code of the reduced workflows under the D2H directory, as the definition of the data format of the derived AO2Ds for the standalone analyses should be also handled by the PAG coordinators.

PWGHF/D2H/Tasks/reducedTaskB0.cxx
```diff
@@ -0,0 +1,263 @@
...
+using namespace o2::aod::hf_cand_b0; // from CandidateReconstructionTables.h
+
+/// B0 analysis task
+struct HfReducedTaskB0 {
+  Configurable<int> selectionFlagB0{"selectionFlagB0", 1, "Selection Flag for B0"};
+  Configurable<double> yCandMax{"yCandMax", -1, "max. cand. rapidity"};
+  Configurable<std::vector<double>> binsPt{"binsPt", std::vector<double>{hf_cuts_b0_to_d_pi::vecBinsPt}, "pT bin limits"};
+
+  float etaMaxAcceptance = 0.8;
+  float ptMinAcceptance = 0.1;
```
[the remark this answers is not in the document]
AlexBigO: I put them as configurables.

PWGHF/DataModel/ReducedDataModel.h
```diff
@@ -0,0 +1,249 @@
...
+DECLARE_SOA_COLUMN(SigSnp2, cSnpSnp, float);    //! Covariance matrix
+DECLARE_SOA_COLUMN(SigTglY, cTglY, float);      //! Covariance matrix
+DECLARE_SOA_COLUMN(SigTglZ, cTglZ, float);      //! Covariance matrix
+DECLARE_SOA_COLUMN(SigTglSnp, cTglSnp, float);  //! Covariance matrix
+DECLARE_SOA_COLUMN(SigTgl2, cTglTgl, float);    //! Covariance matrix
+DECLARE_SOA_COLUMN(SigQ2PtY, c1PtY, float);     //! Covariance matrix
+DECLARE_SOA_COLUMN(SigQ2PtZ, c1PtZ, float);     //! Covariance matrix
+DECLARE_SOA_COLUMN(SigQ2PtSnp, c1PtSnp, float); //! Covariance matrix
+DECLARE_SOA_COLUMN(SigQ2PtTgl, c1PtTgl, float); //! Covariance matrix
+DECLARE_SOA_COLUMN(SigQ2Pt2, c1Pt21Pt2, float); //! Covariance matrix
```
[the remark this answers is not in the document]
AlexBigO: Column names adapted to getters names.

PWGHF/DataModel/ReducedDataModel.h
```diff
@@ -0,0 +1,249 @@
...
+namespace hf_track_par_cov
+{
+// CAREFUL: the getters names shall be the same as the ones of the getTrackParCov method in Common/Core/trackUtilities.h
+DECLARE_SOA_COLUMN(X, x, float);            //! X of track evaluation
+DECLARE_SOA_COLUMN(Alpha, alpha, float);    //! Track frame angle
+DECLARE_SOA_COLUMN(Y, y, float);            //! Y of track evaluation
+DECLARE_SOA_COLUMN(Z, z, float);            //! Z of track evaluation
+DECLARE_SOA_COLUMN(Snp, snp, float);        //! sin(phi)
+DECLARE_SOA_COLUMN(Tgl, tgl, float);        //! tan(lambda)
+DECLARE_SOA_COLUMN(Q2Pt, signed1Pt, float); //! (sign of charge)/Pt in c/GeV
```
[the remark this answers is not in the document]
AlexBigO: Remnant from a previous version. I deleted this block as I can call the columnds from `track`.

PWGHF/DataModel/ReducedDataModel.h
```diff
@@ -0,0 +1,249 @@
...
+                  hf_b0_mc::EtaProng0,
+                  hf_b0_mc::PtProng1,
+                  hf_b0_mc::YProng1,
+                  hf_b0_mc::EtaProng1);
+
+// store all configurables values used in the first part of the workflow
+// so we can use them in the B0 part
+namespace hf_cand_b0_config
+{
+DECLARE_SOA_COLUMN(MySelectionFlagD, mySelectionFlagD, int); //! Flag to filter selected D+ mesons
```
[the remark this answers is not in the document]
AlexBigO: I put `int8_t`.

PWGHF/TableProducer/reducedDataCreatorDplusPi.cxx
AlexBigO: Remnant from a previous version, thanks.

PWGHF/TableProducer/reducedDataCreatorDplusPi.cxx
AlexBigO: I remember trying this in the past and I got errors because I was filling a table with a configurable. That's why I define a new variable of type `int` and not `Configurable<int>` to fill the table.

PWGHF/DataModel/ReducedDataModel.h
AlexBigO: I realized I could give simpler names to the Tracks tables: `HfTracksReduced` and `HfTracksPidReduced`. I initially added `WithSel` because I use `TracksPIDWithSel = soa::Join<aod::BigTracksPIDExtended, aod::TrackSelection>` in `dataCreatorDplusPiReduced.cxx` but I don't use any column of `aod::TrackSelection` beyond this task.

AlexBigO:
> Hi @AlexBigO , in general, it would be useful to identify the common parts between the standard and the reduced workflows and avoid code duplication, especially in the candidate selection.
> 
I totally agree. Is it okay if I implement this in the next PR? This way we could keep this current PR for the implementation of the new workflow and we would be able to start using it earlier (which would be very helpful especially for tuning the cuts).
> Another thing to consider is the location of this code. One could consider moving the entire code of the reduced workflows under the D2H directory, as the definition of the data format of the derived AO2Ds for the standalone analyses should be also handled by the PAG coordinators.
>
I see, I will then move the code to the D2H directory.

PWGHF/TableProducer/reducedCandidateSelectorB0ToDPi.cxx
```diff
@@ -0,0 +1,261 @@
...
+#include "PWGHF/DataModel/ReducedDataModel.h"
+
+using namespace o2;
+using namespace o2::aod;
+using namespace o2::framework;
+using namespace o2::aod::hf_cand_b0; // from CandidateReconstructionTables.h
+using namespace o2::analysis;
+using namespace o2::analysis::hf_cuts_b0_to_d_pi; // from SelectorCuts.h
+
+struct HfReducedCandidateSelectorB0ToDPi {
```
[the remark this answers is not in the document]
vkucera: Not addressed yet

PWGHF/TableProducer/reducedDataCreatorDplusPi.cxx
vkucera: Have you tried `selectionFlagD.value`?

PWGHF/TableProducer/reducedCandidateSelectorB0ToDPi.cxx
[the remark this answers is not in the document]
AlexBigO: I thought it was done, thanks.

PWGHF/TableProducer/reducedDataCreatorDplusPi.cxx
AlexBigO: No, I had not tried it and it works.

PWGHF/TableProducer/reducedCandidateSelectorB0ToDPi.cxx
```diff
@@ -0,0 +1,261 @@
...
+    }
+  }
+
+  /// Apply topological cuts as defined in SelectorCuts.h
+  /// \param hfCandB0 is the B0 candidate
+  /// \param hfCandD is prong1 of B0 candidate
+  /// \param trackPi is prong1 of B0 candidate
+  /// \return true if candidate passes all selections
+  template <typename T>
+  bool selectionTopol(const T& hfCandB0)
```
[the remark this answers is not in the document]
vkucera: To follow up in https://github.com/AliceO2Group/O2Physics/issues/2594

PR #2458, opened 2023-04-21
PWGDQ:  Add CosPA to VarManager and fix small bugs when setting electron PDG code to KFParticle

lupengzhong:
1. Added cos pointing angle to varmanager
2. fix small bugs when setting electron PDG code to a KFParticle
3. change the location of the kVtxNcontribReal in the VarManager and HistogramsLibrary


PWGDQ/Core/VarManager.h
```diff
@@ -1622,7 +1622,7 @@ void VarManager::FillPairVertexing(C const& collision, T const& t1, T const& t2,
       values[kVertexingTauzErr] = values[kVertexingLzErr] * KFGeoTwoProngBarrel.GetMass() / (TMath::Abs(KFGeoTwoProngBarrel.GetPz()) * o2::constants::physics::LightSpeedCm2NS);
 
       values[kKFChi2OverNDFGeo] = KFGeoTwoProngBarrel.GetChi2() / KFGeoTwoProngBarrel.GetNDF();
-      values[kKFCosPA] = calculateCosPA(KFGeoTwoProngBarrel,KFPV);
+      values[kKFCosPA] = calculateCosPA(KFGeoTwoProngBarrel, KFPV);
```
iarsene, line 1625: I would like to have these quantities calculated ONLY if they are required (used in a histogram or in a cut). So please add if (fgUsedVars[...]) statements for them. I suggest that you have a look at all the quantities in this function. I see that everything is computed independent on whether they are used or not. 
Also, you should keep in mind that sometimes a quantity might be needed to be computed because it is required by another one (which is user required). In such cases, we have the VarManager::SetVariableDependencies() function, which can be used to make sure that such variables are always computed.
We have to use these things in order to keep our CPU as low as possible since we will add many more variables over time.

lupengzhong: Dear Ionut,

lupengzhong: thanks so much for the nice and detailed comments. I updated the PR accordingly. hope I understand you correctly

PR #2459, opened 2023-04-21
Remove dependencies from NucleiSpectraTask

mpuccio:
- Remove dependency from the track propagation
- Remove TOF pid full dependency
- Final touches

### DIFF
--- a/PWGLF/Tasks/CMakeLists.txt
+++ b/PWGLF/Tasks/CMakeLists.txt
@@ -63,7 +63,7 @@ o2physics_add_dpl_workflow(spectra-tpc-tiny-pikapr
 
 o2physics_add_dpl_workflow(nuclei-spectra
                     SOURCES NucleiSpectraTask.cxx
-                    PUBLIC_LINK_LIBRARIES O2Physics::AnalysisCore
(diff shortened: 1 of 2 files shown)


njacazio (approved):
Hi @mpuccio thanks! Indeed this is a use case that we should bring to the attention of the community and discuss at the WP4+14 so that we can improve how the tracking is done for all tasks, in any case it would be needed for long trains with multiple wagons

mpuccio:
Yes, I think that when the development of the on-demand track cuts from @ddobrigk becomes available, such tricks could be avoided. But unfortunately, last time, we needed approval from the PB just because the track propagation was taking too long, and with these modifications in the future, we can avoid that.

ddobrigk:
> Yes, I think that when the development of the on-demand track cuts from @ddobrigk becomes available, such tricks could be avoided. But unfortunately, last time, we needed approval from the PB just because the track propagation was taking too long, and with these modifications in the future, we can avoid that.

Fully agreed :-) now Anton fixed the bug that was holding me back, so I have to get back to that because it will make a good impact on CPU. I will have more time after the O2 tutorials and this is one of the things I want to finish! And for now, having this done in your task is great as it will allow us to save CPU immediately.

njacazio:
sure, let's make this point then!

chiarapinto (approved):

Issue #2156, opened 2023-03-07
PWGHF: Subscribe to the lite PID tables instead of the full ones


vkucera:
**Problem:**

Candidate selectors are using the full PID tables with a fine n_sigma granularity which is not needed.
Using the lite PID tables with reduced granularity might reduce the memory consumption of jobs including the selectors.

**Solution:**

- [ ] Replace `pidTPCFull...` with `pidTPC...` in [`BigTracksPID`](https://github.com/AliceO2Group/O2Physics/blob/master/PWGHF/DataModel/CandidateReconstructionTables.h#L102)
- [ ] Replace `o2-analysis-pid-tpc-full` with `o2-analysis-pid-tpc` in [config_tasks.sh](https://github.com/AliceO2Group/Run3Analysisvalidation/blob/master/codeHF/config_tasks.sh), [workflows.yml](https://github.com/AliceO2Group/Run3Analysisvalidation/blob/master/codeHF/workflows.yml)
- [ ] Update device configuration in [dpl-config_run3.json](https://github.com/AliceO2Group/Run3Analysisvalidation/blob/master/codeHF/dpl-config_run3.json).
- [ ] Replace dependencies in Hyperloop wagons
- [ ] Same for TOF

vkucera:
It would probably also help to split the hadron PID tables (p K pi) and the lepton PID tables (e mu).

AlexBigO:
> It would probably also help to split the hadron PID tables (p K pi) and the lepton PID tables (e mu).

Hi @vkucera ! Sorry, I missed this comment. Do you mean the following:
```
using BigTracksHadronPID = soa::Join<BigTracks,
                                  aod::pidTPCFullPi, aod::pidTPCFullKa, aod::pidTPCFullPr,
                                  aod::pidTOFFullPi, aod::pidTOFFullKa, aod::pidTOFFullPr>;
using BigTracksLeptonPID = soa::Join<BigTracks,
                                  aod::pidTPCFullEl, aod::pidTPCFullMu,
                                  aod::pidTOFFullEl, aod::pidTOFFullMu>;
```
?

PR #2460, opened 2023-04-21
PWGHF: Follow issue #2156

AlexBigO:
Following [Vit's issue](https://github.com/AliceO2Group/O2Physics/issues/2156).


AlexBigO:
Hi @vkucera ! By grepping `tofExp` or `tpcExp` in `PWGHF/`, I didn't get any output so I supposed there should be no difference between full and restricted PID workflows (as we currently only call `NSigma` getters in `PWGHF/`).
However, after testing it on Dplus and B0 selectors, and I get a small difference between the 2 workflows. For instance, I go from `6779`  selected D+ with full PID to `6852` selected D+ with restricted PID.
Could it come from the difference of column definition between [restricted](https://github.com/AliceO2Group/O2Physics/blob/master/Common/DataModel/PIDResponse.h#L510) and [full ](https://github.com/AliceO2Group/O2Physics/blob/master/Common/DataModel/PIDResponse.h#L470)PID?

vkucera:
@AlexBigO , did you manage to identify the source of the differences?

PR #2466, opened 2023-04-22
added O2pdg dependency lines 

neelimaagrawal:


neelimaagrawal:
@victor-gonzalez Here you find the changes as per the suggestion from Anton to fix the pdg code of Deuteron. I hope it is okay now. Many thanks in advance for approving it.

PR #2469, opened 2023-04-23
instructions and scripts

mwinn2:
Contains a set of scripts and json configuration files to be used for the hands-on tutorials


iarsene (dismissed):

iarsene:
@ktf, @TimoWilken 
Hi, I don't understand why the clang-format fails. I rerun the test with debugging enabled but I don't seem to get an obvious error. Moreover, @mwinn2 run clang on his commits and there was no change. Can you please have a look? These are needed for the tutorial sessions.

TimoWilken:
Hi @iarsene, where do the `Tutorials/PWGDQ/readerConfiguration_*.json` files come from?

We have clang-format configured to format JSON files with 4-space indents, to match what O2 produces, and the above files have 2-space indents.

As always with the clang-format check, it creates a PR in the PR author's fork, in this case here: https://github.com/mwinn2/O2Physics/pull/1. @mwinn2 can merge this PR to fix the clang-format errors.

iarsene (approved):

iarsene:
> Hi @iarsene, where do the `Tutorials/PWGDQ/readerConfiguration_*.json` files come from?
> 
> We have clang-format configured to format JSON files with 4-space indents, to match what O2 produces, and the above files have 2-space indents.
> 
> As always with the clang-format check, it creates a PR in the PR author's fork, in this case here: [mwinn2#1](https://github.com/mwinn2/O2Physics/pull/1). @mwinn2 can merge this PR to fix the clang-format errors.

Hi Timo, the json files are taken from a private repository which we use in PWG-DQ. I have no enabled checks on the PRs from that repo unfortunately. I will check if I can enable them so that for the future we don't run into similar problems.

PR #2470, opened 2023-04-23
Monte Carlo in FemtoDream

gmantzar:
Dear Victor,
(@AntonRiedel and @lauraser if you want to follow)

thanks for your input and suggestions in the past for how to best include Monte Carlo in femtodream. With this pull request, we considered your comments and we would kindly like to ask you to approve it, if you agree with the changes. Here a summary of the changes:

**Datamodel:**
As you suggested, the FemtoDreamMCParticles table reuses columns from the `FemtoDreamParticle` table and has additional ones only for the PDG code and the origin.
The index table `FemtoDreamMCLabals` is linkable with the `FemtoDreamParticles` table in order to access Monte Carlo truth information.

**Producer(Reduced)Task:**
The old process function in the `ProducerReducedTask` is now inside the `fillCollisionsAndTracks` function, similar as it is already in the `ProducerTask`. Inside the `fillCollisionsAndTracks`, the `fillMCParticle` function is called, which fills the `FemtoDreamMCParticles` and `FemtoDreamMCLabels` tables. The function which classifies the mother of the particle is `checkDaughterType` and can be found in `FemtoUtils.h`. We decided to do it this way, because we want to be able to consider different mother particles in future analysis, and this function could grow in terms of if statements.
There is still a bit code doublication between the two producer tasks (as there was already before). But we plan to unify them after the O2 tutorial next week.

**PairTaskTrackTrack:**
We restructured the task similar to the producer tasks: The main functionalities of the former `processSameEvent` and `processMixedEvent` functions are now in the templated `doSameEvent` and `doMixedEvent` functions. They are called by the now four process functions, depending on the use cases (processSameEvent, processSameEventMC, processMixedEvent, processMixedEventMC).
In the function body itself, almost nothing changed: The initialization and filling of the histogramms is now done with an extra parameter `isMC`. The `FemtoDreamParticleHisto` and the `FemtoDreamContainer` classes take care of the filling of MC truth or reconstructed data correctly.

**FemtoDreamParticleHisto and FemtoDreamContainer class:**
When they are initialized or their fill functions are called, they always execute the `_base` functions, which take care of the histograms common for Monte Carlo truth and MC reconstructed. If MC truth information is wanted, the additional `_MC` functions are called.

Please let us know if you have other questions or suggestions :)

### DIFF
--- a/PWGCF/DataModel/FemtoDerived.h
+++ b/PWGCF/DataModel/FemtoDerived.h
@@ -186,12 +186,12 @@ namespace femtodreamMCparticle
 {
 /// Distinuishes the different particle origins
 enum ParticleOriginMCTruth {
-  kPrimary,          //! Primary track or V0
-  kDaughter,         //! Particle from a decay
-  kMaterial,         //! Particle from a material
-  kNotPrimary,       //! Not primary particles (kept for compatibility reasons with the FullProducer task. will be removed, since we look at "non primaries" more differentially now)
-  kFake,             //! particle, that has NOT the PDG code of the current analysed particle
-  kDaughterLambda,   //! Daughter from a Lambda decay
+  kPrimary,           //! Primary track or V0
+  kDaughter,          //! Particle from a decay
+  kMaterial,          //! Particle from a material
+  kNotPrimary,        //! Not primary particles (kept for compatibility reasons with the FullProducer task. will be removed, since we look at "non primaries" more differentially now)
+  kFake,              //! particle, that has NOT the PDG code of the current analysed particle
+  kDaughterLambda,    //! Daughter from a Lambda decay
   kDaughterSigmaplus, //! Daughter from a Sigma^plus decay
   kNOriginMCTruthTypes
 };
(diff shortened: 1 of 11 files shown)


PWGCF/DataModel/FemtoDerived.h
```diff
@@ -181,34 +182,60 @@ DECLARE_SOA_TABLE(FemtoDreamDebugParticles, "AOD", "FEMTODEBUGPARTS",
...
+  kNotPrimary,        //! Not primary particles (kept for compatibility reasons with the FullProducer task. will be removed, since we look at "non primaries" more differentially now)
+  kFake,              //! particle, that has NOT the PDG code of the current analysed particle
+  kDaughterLambda,    //! Daughter from a Lambda decay
+  kDaughterSigmaplus, //! Daughter from a Sigma^plus decay
+  kNOriginMCTruthTypes
 };
 
-DECLARE_SOA_INDEX_COLUMN(FemtoDreamParticle, femtoDreamParticle);
+static constexpr std::string_view ParticleOriginMCTruthName[kNOriginMCTruthTypes] = {"_Primary", "_Daughter", "_Material", "_NotPrimary", "_Fake", "_DaughterLambda", "DaughterSigmaPlus"}; //! Naming of the different OriginMCTruth types
+
```
victor-gonzalez, line 200: This is a cosmetic suggestion 
Use the same layout as for `ParticleOriginMCTruth`
Return after the open curl bracket, one line for each input and a new line for the closing curl bracket

victor-gonzalez: It will help the readability

gmantzar: ok, done

victor-gonzalez (approved):
Thanks!

PR #2472, opened 2023-04-24
simplification which makes it work also on older MC AO2D files

ChiaraDeMartin95:


alibuild (approved):

njacazio (approved):
Thx Chiara

PR #2473, opened 2023-04-24
PWGJE: removing global variables from jet finder

nzardosh:


alibuild (approved):

PR #2489, opened 2023-04-26
Extend ST QA

qgp:


vkucera (changes requested):

ddobrigk:
> Hi @ddobrigk, discussing with @mpuccio we considered `Common/DataModel` the better place in order to avoid inter-PWG dependencies as the tables would also be used from PWGHF. Would you prefer to include `PWGLF` headers from `PWGHF` analyses?

Hi @qgp , this has been discussed quite extensively in the early beginnings of the directory substructure and after lengthy discussions, we ultimately decided to keep the strangeness data model inside LF. This has been the decision since 2020, with the beginning of the strangeness data model, and yes, it's not only HF code but others (DQ, EM) that make use of the LF data model like this - this hasn't been a problem to date. If this is to change, I would appreciate having a broader discussion also with the PWG conveners involved.

njacazio:
Hi @qgp,
This has to be developed in line with what was already done, and it should stay under the responsibility of the PWG-LF, and this is where it should be discussed.
So far (since 2020 or so), this has worked very well, in complete symbiosis with all the PWG (not only HF, btw).
Also, we might be mixing two separate discussions within this PR (HF and LF). Consider splitting the PR into 2 if you want to discuss it separately.
Cheers,
Alberto & Nicolo'

qgp:
Dear @vkucera, thanks a lot for your review and the comments. I have updated the task and hope not to have overlooked anything.

qgp:
Hi @ddobrigk, @njacazio, concerning where the tables should go, I might have missed some previous discussions. The tables in question extend the strangeness tracking tables - which are part of `AnalysisDataModel.h` - by adding the collision ID which is not explicitly stored to save space, nothing more, to allow grouping with collisions. To me this seems a generic extension of what is in the general data model and I find them well placed in `Common`. If, however, everyone thinks that they are better placed in `PWGLF` I can move them there.

Common/DataModel/StrangenessTracking.h
```diff
@@ -0,0 +1,33 @@
...
+
+DECLARE_SOA_TABLE(TrackedV0Colls, "AOD", "TRAV0COLL", //! Table joinable with TrackedV0s containing collision ids
+                  track::CollisionId, o2::soa::Marker<2>);
+using TrackedV0Coll = TrackedV0Colls::iterator;
+
+DECLARE_SOA_TABLE(Tracked3BodyColls, "AOD", "TRA3BODYCOLL", //! Table joinable with Tracked3Bodys containing collision ids
+                  track::CollisionId, o2::soa::Marker<3>);
+using Tracked3BodyColl = Tracked3BodyColls::iterator;
+} // namespace o2::aod
+
```
mpuccio, line 32: Thanks a lot, @qgp!
Shall we also have the `Join`s and their iterators predefined here?
If you think it makes sense, it can be also done at the next iteration.

fmazzasc:
Hi @ddobrigk and @njacazio , 
as I personally wrote the data model in O2 and I plan to do analysis based on the strangeness tracking here are my two cents.

> This has to be developed in line with what was already done

This is exactly what it is done here: a common data model in O2 should go in the corresponding directory in O2Physics, as it is meant to be employed by almost all the PWGs. As David said, I think it would be useful to have a broader discussion in the PWG about that and consider to make some changes in the policy, as the argumentation "we have always done like this since 2020 " (three years only) sounds a bit weak to me.
Cheers,
Francesco

jgrosseo:
I guess it is known but just for completeness: cross-dependencies requiring data model headers included by tasks do not cause a specific order in which the PWG* folders have to be compiled or their libraries loaded. Because the data model headers do not become part of the library.

ddobrigk:
> Hi @qgp, @ddobrigk and @njacazio , as I personally wrote the data model in O2 and I plan to do analysis based on the strangeness tracking here are my two cents.
> 
> > This has to be developed in line with what was already done
> 
> This is exactly what it is done here: a common data model in O2 should go in the corresponding directory in O2Physics, as it is meant to be employed by almost all the PWGs. As David said, I think it would be useful to have a broader discussion in the PWG about that and consider to make some changes in the policy, as the argumentation "we have always done like this since 2020 " (three years only) sounds a bit weak to me. Cheers, Francesco

Hi Francesco,  

I feel like I should clarify something: the point I raised, that of three years of development, is not meant to exclude any further discussion, and I remain open to changing paths. However, I believe that we should have strangeness tracking be a natural evolution of the existing LF data model because we will all benefit if the many already-existing tasks that utilize the existing strangeness data model can easily migrate, if they so wish, to use the latest in reconstruction software (in this case, tracked cascades). This is why, in the existing strangeness analysis data model, I have now mirrored essentially all strangeness model getters that are there for cascades also in the tracked cascade analysis table (which btw includes a collision ID column, meaning there is actually no need for that to be generated by an additional table producer), and all this is already fully functional for tracked cascades. This is what we had discussed in person at CERN last year: that strangeness tracking would be incorporated into the cascade builders, and now it already is.

The point of uniform development inside the *existing* PWGLF data model and modus operandi is exactly to potentialize the benefits of your great work in the reconstruction by making it as simple as possible for others to also use this technology with existing tasks that use cascades. I think that's a very, very strong argument: we should be working together to improve the analyses and data models that are already in place and used in O2Physics. This is what I care about most, much much more than in which directory which code ends up - though even on this, it feels more appropriate, in my mind, that LF experts and approvers (and not the 'Common’ approvers) approve strangeness tracking analysis work. Plus, as Jan Fiete says, there's actually no reason from the technical point of view to change the location of the strangeness-analysis-related data model declarations. But anyway: this discussion is maybe better continued in another venue! 😉

PWGHF/D2H/Tasks/taskOmegacSt.cxx
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  void process(aod::Collision const& collision,
+               TrackedCascades const& trackedCascades, aod::Cascades const& cascades,
+               aod::V0s const& v0s, TracksExt const& tracks, aod::McParticles const& mcParticles, aod::BCsWithTimestamps const&)
+  {
+    const auto bc = collision.bc_as<aod::BCsWithTimestamps>();
+    if (mRunNumber != bc.runNumber()) {
+      mRunNumber = bc.runNumber();
+      auto run3grp_timestamp = bc.timestamp();
```
vkucera, line 76: Names of variables should be in camelCase.

PWGHF/D2H/Tasks/taskOmegacSt.cxx
```diff
@@ -0,0 +1,166 @@
...
+
+      if (o2::parameters::GRPObject* grpo = ccdb->getForTimeStamp<o2::parameters::GRPObject>(grpPath, run3grp_timestamp)) {
+        o2::base::Propagator::initFieldFromGRP(grpo);
+      } else {
+        if (o2::parameters::GRPMagField* grpmag = ccdb->getForTimeStamp<o2::parameters::GRPMagField>(grpmagPath, run3grp_timestamp)) {
+          o2::base::Propagator::initFieldFromGRP(grpmag);
+        } else {
+          LOG(fatal) << "Got nullptr from CCDB for path " << grpmagPath << " of object GRPMagField and " << grpPath << " of object GRPObject for timestamp " << run3grp_timestamp;
+        }
+      }
```
vkucera, line 86: Is this nesting needed?

PWGHF/D2H/Tasks/taskOmegacSt.cxx
```diff
@@ -0,0 +1,166 @@
...
+      if (ptrack.mcParticle().has_mothers() && ntrack.mcParticle().has_mothers() &&
+          ptrack.mcParticle().mothersIds()[0] == ntrack.mcParticle().mothersIds()[0]) {
+        const auto v0part = ptrack.mcParticle().mothers_as<aod::McParticles>()[0];
+        LOG(debug) << "v0 with PDG code: " << v0part.pdgCode();
+        if (v0part.has_mothers() && bachelor.mcParticle().has_mothers() &&
+            v0part.mothersIds()[0] == bachelor.mcParticle().mothersIds()[0]) {
+          const auto mother = v0part.mothers_as<aod::McParticles>()[0];
+          const auto pdgCode = mother.pdgCode();
+          LOG(debug) << "cascade with PDG code: " << pdgCode;
+          if (std::abs(pdgCode) == 3334) {
```
vkucera, line 129: Do not use magic numbers. Use a named PDG value. Define it in `https://github.com/AliceO2Group/O2Physics/blob/master/PWGHF/Core/SelectorCuts.h#L29` if needed.

PWGHF/D2H/Tasks/taskOmegacSt.cxx
```diff
@@ -0,0 +1,166 @@
...
+            LOG(debug) << "found Omega, looking for pions";
+            std::array<double, 2> masses{RecoDecay::getMassPDG(3334), RecoDecay::getMassPDG(kPiPlus)};
+            std::array<std::array<float, 3>, 2> momenta;
+            trackParCovTrk.getPxPyPzGlo(momenta[0]);
+            // momenta[0] = {trackCasc.px(), trackCasc.py(), trackCasc.pz()};
+            for (const auto& track : tracks) {
+              if (!track.has_mcParticle())
+                continue;
+              const auto mcpart = track.mcParticle();
+              if (mcpart.pdgCode() == std::copysign(211, pdgCode)) {
```
vkucera, line 139: Same here

PWGHF/D2H/Tasks/taskOmegacSt.cxx
```diff
@@ -0,0 +1,166 @@
...
+          LOG(debug) << "cascade with PDG code: " << pdgCode;
+          if (std::abs(pdgCode) == 3334) {
+            LOG(debug) << "found Omega, looking for pions";
+            std::array<double, 2> masses{RecoDecay::getMassPDG(3334), RecoDecay::getMassPDG(kPiPlus)};
+            std::array<std::array<float, 3>, 2> momenta;
+            trackParCovTrk.getPxPyPzGlo(momenta[0]);
+            // momenta[0] = {trackCasc.px(), trackCasc.py(), trackCasc.pz()};
+            for (const auto& track : tracks) {
+              if (!track.has_mcParticle())
+                continue;
```
vkucera, line 137: Missing braces

PWGHF/D2H/Tasks/taskOmegacSt.cxx
```diff
@@ -0,0 +1,166 @@
...
+
+      registry.fill(HIST("hDca"), std::sqrt(impactParameterTrk.getR2()));
+      registry.fill(HIST("hDcaXY"), impactParameterTrk.getY());
+      registry.fill(HIST("hDcaZ"), impactParameterTrk.getZ());
+      registry.fill(HIST("hDcaVsPt"), impactParameterTrk.getY(), trackCasc.pt());
+      registry.fill(HIST("hDcaVsR"), impactParameterTrk.getY(), RecoDecay::sqrtSumOfSquares(trackCasc.x(), trackCasc.y()));
+      registry.fill(HIST("hMassVsPt"), trackedCascade.omegaMass(), trackCasc.pt());
+
+      if (!ptrack.has_mcParticle() || !ntrack.has_mcParticle() || !bachelor.has_mcParticle())
+        continue;
```
vkucera, line 112: Missing braces.

vkucera (changes requested):

PWGHF/D2H/Tasks/taskOmegacSt.cxx
```diff
@@ -0,0 +1,166 @@
...
+
+struct HfTaskOmegacSt {
+  Configurable<double> bz{"bz", 50., "magnetic field"};
+  Configurable<int> cfgMaterialCorrection{"cfgMaterialCorrection", static_cast<int>(o2::base::Propagator::MatCorrType::USEMatCorrNONE), "Type of material correction"};
+  Configurable<std::string> ccdburl{"ccdb-url", "http://alice-ccdb.cern.ch", "url of the ccdb repository"};
+  Configurable<std::string> grpmagPath{"grpmagPath", "GLO/Config/GRPMagField", "CCDB path of the GRPMagField object"};
+  Configurable<std::string> grpPath{"grpPath", "GLO/GRP/GRP", "Path of the grp file"};
+
+  Service<o2::ccdb::BasicCCDBManager> ccdb;
+  int mRunNumber;
```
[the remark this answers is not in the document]
qgp: I don't understand the comment here, this is a member variable.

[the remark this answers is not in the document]
vkucera: It is a struct member, not a class member.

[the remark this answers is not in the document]
qgp: In C++, there is no difference between a class and a struct other than members default to public in a struct.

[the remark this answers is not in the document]
vkucera: Correct. Yet, the conventions clearly distinguish the two in this case.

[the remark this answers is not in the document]
qgp: Fine even though I don't think this makes any sense. If we were to switch from `struct` to `class` (when the analysis framework can handle it) we would have to change all variable names.

Common/DataModel/StrangenessTracking.h
qgp: Added

vkucera (approved):

vkucera (approved):
Thanks @qgp for addressing the comments.

vkucera:
I let @lvermunt @fcatalan92 give their approval for the D2H.

ddobrigk:
Not for me but I am not an approver of LF :-D

ddobrigk (approved):
(not sure if I need to approve, since I have rights only to a directory 'Common' that isn't ultimately touched by this, but I am approving anyway :-) )

fcatalan92 (approved):

njacazio (approved):
Hi @qgp apologies, I missed the notification, thanks a lot for your work!

PR #2490, opened 2023-04-26
Update MC PDG particle list to keep

BongHwi:
Set configurable for event mixing pool
And optimise the nBins of QA histograms


alibuild (approved):

njacazio (approved):

PR #2496, opened 2023-04-27
PWGHF: adding daughter track min/max eta selection

fcolamar:
Needed for urgent QA checks on new data.
The only potential problem that I see is that analysers should be warned to set also etaMinTrack2Prong/3Prong/Bach/V0Daug (depending on what objects they use) in their json files, otherwise the default values will be taken, and if they change the default eta cut by only modifying etaMaxTrack2Prong/3Prong/Bach/V0Daug (and only that) with respect to the default, the behavior will not be what they would expect.

I had tried to set the default min values to a specific unphysical value (-99999.), and then, in the init, set etaMinTrack2Prong to -etaMaxTrack2Prong in case the value of the min eta configurable is still -99999. (meaning that it was not set in the json), but it seems that a configurable cannot have its value reassigned at a later stage, with a etaMinTrack2Prong = -etaMaxTrack2Prong instruction.
Another possibility to avoid any problem would be to set the default min eta configurable values to the opposite of the max eta configurable, but I don't know if this kind of definition is doable.

But if you have a different working solution to avoid this potential problem, I can implement it.


vkucera:
Hi @fcolamar , changing the configurable in `init`, as you suggest, should work. If it is not the case, it indicates a problem.

vkucera:
Maybe @aalkin can help?

aalkin:
Indeed, the configurable does not have an `operator=` defined. It should work like this `etaMinV0Daugh.value = -etaMaxV0Daugh`

fcolamar:
Hi @aalkin, @vkucera, it's working now, thanks!
With this information I could define the new min configurables in a fully backward-compatible way.
Cheers,
Fabio

vkucera (approved):

PR #2501, opened 2023-04-27
PWGDQ/EM: TableReader SameEventPairing: Change format of required cfgPairCut. Optimise CPU, Memory.

feisenhu:
Update the SameEventPairing process function within the tableReader to be more CPU and memory friendly.

Currently: all  provided TrackCuts are combined with all provided PairCuts. (Computing intensive and large memory use for output)
This pull request: Provide a string by cfgPairCuts, to specify which Track/Muon cuts should be combined with which PairCut. (only provided combinations are calculated and stored in output)

How to use:
Ask for comma separated list of cuts of format "TrackCut":"MuonCut":"PairCut"
If one of them is not provided it has to be replaced by " " (a blank space).

Note: If "TrackCut" or "MuonCut" cut is used in cfgPairCuts, they have to be listed in the cfgTrackCuts, cfgMuonCuts as well.


iarsene:
Hi @feisenhu, thanks for this commit. Although the functionally of your commit is doing for the pairing task something more or less in the direction i would like to go, I am wondering if we can't go all the way and implement a "full functionality". What I mean is that now we are using the "convention" that in the same event pairing task we must give the track cut names in the same order as in the barrel selection task (and the track cuts computed in the track selection task meant to be used in the same event pairing must be the first ones). This is however very error prone and i implement it like this a while ago because I did not know at the time how to do it otherwise. If I understand the commit correctly, your commit does not help much in this direction. However, now I think we can access the framework init context for all tasks, and in principle we can access from the same event pairing task the names of track cuts calculated in the barrel selection task. In this way, the order will not be important anymore, and we can also do a check that the track cut requested in the same event pairing, was actually computed in the barrel selection task. Similarly for the muon workflow part. So the user should be able to run an arbitrary list of cuts in the track selection task (for some cuts you may not want to run the pairing) and then in the same event pairing you can ask for an arbitrary combination of trackCut:pairCut and the corresponding bit for the track cut can be deduced automatically (by accessing the configurable of the track selection task). This is important also for the DileptonHadron task, where we currently have the additional convention that the track cut used to correlate with the dileptons, is the last one in the barrel selection task configuration.

Another comment I have is that I would argue that if the user is specifying a list of n track cuts and m pairing cuts, one usually wants a systematic check, where all combinations are usually wanted. However, i think that in most analysis cases, in the same event pairing task you don't need to run more than one dilepton cut because you can actually create histograms with the quantity on which you want to cut as an axis. You can do most cuts "offline". So, providing just the names of the track cuts for which you want to do same event pairing should be enough in most cases. However, if you think that in most analysis cases you need to run some track cuts with a given pair cut and other cuts with another pair cut, we could go with your changes. However, people that really want to do all combinations will then have to fill the configurable with a long and error prone coma separated list of string triplets.

Another comment is that we should do something similar also for the event mixing task, and for the DileptonHadron task. For the last one, I know its not your responsability, but it would be great if you can do it.

feisenhu:
@iarsene Thank you for your comment.

I agree, using the provided barrel cut strings from the barrel selection task would help a lot to reduce the error-proneness.
This sounds reasonable and I could try to implement this.

On your second comment:
For systematic studies of combinations between n track cuts and m pair cuts, the current version of the same event pairing is doing exactly this. But there is no option to switch this on or off right now. So if you provide a list of track cuts and a list of pair cuts, then it combines automatically all cuts. The problem which I see in this is when you run electron and muon analysis together which do not require the same track or pair cuts. Then the output would have several output lists with cuts corresponding to the muons and thus would not be meaningful. (I don't think this would be a usual case but should be considered. Something similar would hold for the processElectronMuonSkimmed function or the processAllSkimmed functions) 
I guess we should go for a general solution on how to handle this and apply it for other process functions as well. 

I'm aware that the event mixing would need a similar (if not the same) functionality but I have not used it that much yet and would need time to implement it. The DileptonHadron task is still unknown to me I would need to check this.

PR #2506, opened 2023-04-28
MegaLinter: Disable trivy

vkucera:
The trivy check does not run correctly and is probably not even needed.
@TimoWilken


TimoWilken:
Currently broken; merged. Thanks @vkucera!

PR #2512, opened 2023-04-29
Avoid two process functions for the MC

mpuccio:
Also, add the z vertex position to the output tree and fix the integer sign


mpuccio:
@ivorobye with this, we do not need two process functions and we get the right output

njacazio (approved):

PR #2528, opened 2023-05-02
Add histogram for TPC checks vs vertex z

fcatalan92:
@mfaggin I added a TH3 for the check I showed at the last AOT-Track meeting. 
I decided not to modify `Tracks/TPC/tpcNClsFoundVsEta`, which now holds the same information as the new histogram, to avoid the need for post-processing to have the TPC cls vs eta plot. Let me know if you prefer the two histograms to be merged.

### DIFF
--- a/DPG/Tasks/AOTTrack/qaEventTrack.cxx
+++ b/DPG/Tasks/AOTTrack/qaEventTrack.cxx
@@ -76,7 +76,7 @@ struct qaEventTrack {
   ConfigurableAxis binsPt{"binsPt", {VARIABLE_WIDTH, 0.0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0, 1.1, 1.2, 1.3, 1.4, 1.5, 2.0, 5.0, 10.0, 20.0, 50.0}, ""};
   ConfigurableAxis binsDeltaPt{"binsDeltaPt", {100, -0.495, 0.505}, ""};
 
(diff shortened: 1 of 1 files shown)


DPG/Tasks/AOTTrack/qaEventTrack.cxx
```diff
@@ -76,7 +76,7 @@ struct qaEventTrack {
   ConfigurableAxis binsPt{"binsPt", {VARIABLE_WIDTH, 0.0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0, 1.1, 1.2, 1.3, 1.4, 1.5, 2.0, 5.0, 10.0, 20.0, 50.0}, ""};
   ConfigurableAxis binsDeltaPt{"binsDeltaPt", {100, -0.495, 0.505}, ""};
 
-  ConfigurableAxis binsVertexPosZ{"binsVertexPosZ", {100, -20., 20.}, ""}; // TODO: do we need this to be configurable?
+  ConfigurableAxis binsVertexPosZ{"binsVertexPosZ", {80, -20., 20.}, ""};  // TODO: do we need this to be configurable?
```
mfaggin, line 79: Any specific reason to change it? I'd keep it as it is (it's useful for us for an easy comparison with old plots). In case you need exactly this binning for your plot, consider to do you own axis (but consider this option only if really necessary).

mfaggin (approved):
thanks a lot @fcatalan92 !

PR #2543, opened 2023-05-04
PWGMM: dndeta: fix incorrect bin filled for event selection

aalkin:


alibuild (approved):

PR #2547, opened 2023-05-04
Add QA histograms with mass of charm hadrons vs phi

fgrosa:


alibuild (approved):

vkucera:
Hi @fgrosa , if I remember well, there is a problem with the order of headers, but besides this there are other errors can could be fixed.

fgrosa:
> Hi @fgrosa , if I remember well, there is a problem with the order of headers, but besides this there are other errors that could be fixed.

Ciao @vkucera, sure I will do it in another PR (I hope this one will enter in the daily tag so that I can use it on hyperloop)

PR #2592, opened 2023-05-10
Fixes to h-Strangeness code

ddobrigk:
* adds a self-correlation safeguard for collision mixing
* fixes a mass region check problem in data model 
* changes subscription of multiplicity into centrality percentiles


njacazio (dismissed):

njacazio:
That's weird, the CI does not start

PR #2595, opened 2023-05-10
Tools: Add eta and phi to D0 and single daughter of the output tables

lupengzhong:
Add eta and phi information to D0 and its single daughter in qaKFParticle.cxx and qaKFParticle.h


ddobrigk (approved):
thanks!

iarsene (approved):

PR #2596, opened 2023-05-11
DPG: Update matching efficiency code

mfaggin:
- fix TOF PID usage (only tracks with `hasTOF` true)
 - remove useless new lines


alibuild (approved):

Issue #2594, opened 2023-05-10
PWGHF: Factor out common functions used in B0 selectors


vkucera:
> Hi @AlexBigO , in general, it would be useful to identify the common parts between the standard and the reduced workflows and avoid code duplication, especially in the candidate selection.
> 
I totally agree. Is it okay if I implement this in the next PR? This way we could keep this current PR for the implementation of the new workflow and we would be able to start using it earlier (which would be very helpful especially for tuning the cuts).

_Originally posted by @AlexBigO in https://github.com/AliceO2Group/O2Physics/issues/2456#issuecomment-1535336794_

vkucera:
Addressed in https://github.com/AliceO2Group/O2Physics/pull/2598

PR #2598, opened 2023-05-11
PWGHF: Factor out common functions used in B0 selectors

AlexBigO:
Hi @vkucera, I put the selection functions in `CandidateSelectionTables.h` (the same way the reconstruction functions are implemented in `CandidateReconstructionTables.h`).

Following issue #2594 .

------------------
Modifications applied:
- move `selectionTopol` and `selectionPID` to `CandidateSelectionTables.h` under `hf_sel_candidate_b0` namespace and add the required headers
- add arguments for the configurables of the tasks
- specific issue linked to `acceptPIDNotApplicable`. This is a boolean configurable, so the operators `!` and `&&` lead to compilation errors. Solution proposed: use `acceptPIDNotApplicable.value` of type `bool` as argument of `selectionPID`
- modify the code in `candidateSelectorB0ToDPi.cxx` to adapt it to the latest version (similar to the one of `candidateSelectorB0ToDPiReduced.cxx`):
  - no need to create an instance of D candidate as all the information we need is stored in B0 candidate table --> no need to subscribe to `HfCand3Prong` anymore
  - instead of passing `hfCandB0`, `candD` and `trackPi` in `selectionTopol`, only pass `hfCandB0` and use it to build `ptD`, `ptPi`

### DIFF
--- a/PWGHF/D2H/TableProducer/candidateSelectorB0ToDPiReduced.cxx
+++ b/PWGHF/D2H/TableProducer/candidateSelectorB0ToDPiReduced.cxx
@@ -88,101 +88,6 @@ struct HfCandidateSelectorB0ToDPiReduced {
     }
   }
 
-  /// Apply topological cuts as defined in SelectorCuts.h
-  /// \param hfCandB0 is the B0 candidate
-  /// \param hfCandD is prong1 of B0 candidate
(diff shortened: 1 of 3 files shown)


vkucera (approved):
Thanks @AlexBigO

PR #2603, opened 2023-05-11
Small example task for dielectron. Prep for OTF MC

hscheid:


njacazio (approved):
Thanks Sebastian, supernice

PR #2605, opened 2023-05-12
Analytical length calculator added

ddobrigk:
* this adds a calculation of analytical track length to a TOF layer, solving a problem that appears due to the ALICE track parametrization and leading to the following comparison between numerical (previous) and analytical values of track lengths (given here as an example for B = 2T and rTOF = 1m): 
```
Pt:	0.000	Length:	0.000000,	 analytical: -100.000000,	delta = -100.0000000000
Pt:	0.100	Length:	28.060884,	 analytical: -100.000000,	delta = -128.0608825684
Pt:	0.200	Length:	58.917965,	 analytical: -100.000000,	delta = -158.9179687500
Pt:	0.300	Length:	90.278923,	 analytical: 171.579880,	delta = 81.3009567261
Pt:	0.400	Length:	81.894432,	 analytical: 126.392189,	delta = 44.4977569580
Pt:	0.500	Length:	119.894913,	 analytical: 119.895638,	delta = 0.0007247925
Pt:	0.600	Length:	117.071625,	 analytical: 117.071938,	delta = 0.0003128052
Pt:	0.700	Length:	115.538307,	 analytical: 115.538567,	delta = 0.0002593994
Pt:	0.800	Length:	114.600685,	 analytical: 114.600822,	delta = 0.0001373291
Pt:	0.900	Length:	113.981468,	 analytical: 113.981583,	delta = 0.0001144409
Pt:	1.000	Length:	113.549644,	 analytical: 113.549706,	delta = 0.0000610352
```
* notably, the numerical values below 100cm for the numerical length - the first column - have to be wrong as they are shorter than the TOF radius; they were a result of spurious results of the `getXatLabR` function when the necessary angle traveled in the track helix is above pi/2. 
* the calculation is such that it should also operate correctly in case a track doesn't reach the TOF layer - in which case no TOF signal will be generated, which was not necessarily the case before due to the `getXatLabR` finding an incorrect X. 
* For the future, something like this could be used as a preparatory stage in a `getXatLabR` improvement, too.

### DIFF
--- a/ALICE3/TableProducer/onTheFlyTOFPID.cxx
+++ b/ALICE3/TableProducer/onTheFlyTOFPID.cxx
@@ -127,60 +127,59 @@ struct OnTheFlyTOFPID {
   {
     // don't make use of the track parametrization
     float length = -100;
-    
+
     o2::math_utils::CircleXYf_t trcCircle;
     float sna, csa;
-    track.getCircleParams(magneticField, trcCircle, sna, csa );
-    
+    track.getCircleParams(magneticField, trcCircle, sna, csa);
+
     // distance between circle centers (one circle is at origin -> easy)
     float centerDistance = std::hypot(trcCircle.xC, trcCircle.yC);
-    
+
     // condition of circles touching - if not satisfied returned length will be -100
-    if(centerDistance < trcCircle.rC + radius && centerDistance > fabs(trcCircle.rC - radius)){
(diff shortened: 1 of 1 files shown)


ddobrigk:
@hscheid this one's for you, my friend! :-D

njacazio (approved):

PR #2607, opened 2023-05-12
add LightTask for hyperloop

BongHwi:
Avoid to use `Partition`


alibuild (approved):

PR #2612, opened 2023-05-14
Add Light task not using partition

BongHwi:
Partition was introduced to have the same format as the event-mixing process.
But it's revealed that it uses too much CPU time, so I decided to avoid using it.


alibuild (approved):

alibuild (approved):

PR #2614, opened 2023-05-15
Macro to upload trigger class aliases for 2023

ekryshen:


alibuild (approved):

PR #2618, opened 2023-05-15
Monte Carlo: fixes and additional Histogramms

gmantzar:
**Fixes:**
- We removed the entry `kFake` in the enum `ParticleOriginMCTruth` since we can distinguish fake and real candidates in Monte Carlo directly via their PDG code.
- The TempFitVar plots (for tracks the DCA) are filled only in cathegories according to their origin and not according to their species (i.e. if they are fake or not)

**New histogramms:**
In order to estimate the purity in a pT and kstar dependent way, we added the histogramms with the suffix `_NoFakeCandidates`. They are filled with the same values as the histogramms with Monte Carlo reconstructed, but only if the track candidate (or both candidates in the kstar plots) have the correct PDG code.

**Some rearrangements** in the `SetPair` functions in `FemtoDreamContainer` in order to calculate values only once and passing them where they are needed to fill the histogramm instead of recalculating every time.


PWGCF/FemtoDream/FemtoDreamContainer.h
```diff
@@ -151,8 +151,8 @@ class FemtoDreamContainer
   /// \param part1 Particle one
   /// \param part2 Particle two
   /// \param mult Multiplicity of the event
-  //template <o2::aod::femtodreamMCparticle::MCType mc, typename T>
-  //void setPair_base(T const& part1, T const& part2, const int mult)
+  // template <o2::aod::femtodreamMCparticle::MCType mc, typename T>
+  // void setPair_base(T const& part1, T const& part2, const int mult)
```
victor-gonzalez, line 155: Please, don't leave code commented

victor-gonzalez (commented):
Please, consider my comments

gmantzar:
Ok done
Sorry, I missed that

victor-gonzalez (approved):

PR #2620, opened 2023-05-16
Tools/KFparticle: Fix event selection

AnnalenaKa:
Fix event selection with sel8 trigger as filter:
Previously, the event selection was not working because a retrun was used. In the PR the event selection is replaced by a filter and the filtered events are passed to the process function. Like this, the event selection is always applied properly.
Also, the simple track selections are replaced by a filter.
@lupengzhong, @ddobrigk


ddobrigk (approved):
Thank you!

PR #2623, opened 2023-05-16
PWGLF: Adding 'TRD only' check for TOF

jloemker:
Hey @njacazio,

this is the additional configurable to check for tracks with only TRD information. 

Cheers, 
Johanna


njacazio (approved):
Thx Johanna!

PR #2626, opened 2023-05-16
Update to use centrality FT0M

BongHwi:
It will affect all resonance tasks
One should add the centrality-table task.


alibuild (approved):

PR #2644, opened 2023-05-19
PWGEM/PhotonMeson: update on 2 photon analyses

dsekihat:
This PR contains following commits:
PWGEM/PhotonMeson: re-activate EMC track matching veto
PWGEM/PhotonMeson: add v0leg-cluster matching
PWGEM/PhotonMeson: add cuts for PCM


mhemmer-cern (approved):
Looks good, but hopefully this can be included in the emcCorrectionTask at some point so the track matching is focused in one place and not the job of the analysers 😅

PR #2647, opened 2023-05-19
[PWGLF]  spectraTOF: requirements for no TRD and evTime histograms 

jloemker:
Hi @njacazio,

the changes to monitor the event time selection for TOF with and without TRD.
Here are the configurables for TRD only and no TRD, as well as the event time selection.
We also enabled the centrality histograms from FT0M and FT0A.

Cheers,
Johanna


njacazio (approved):
Thanks Johanna!

PR #2658, opened 2023-05-22
Move to C++20

ktf:


ktf:
Let's see how much it breaks.

ddobrigk:
Okay, I think we missed merging this - will merge and then if we get troubles it wil be from tomorrow's tag (Monday). Let's see....

ddobrigk (approved):

ddobrigk:
Will wait until end of Sunday to merge to master, stop me if not ok!

PR #2663, opened 2023-05-23
PWGHF: Enabling AND logic for TPC+TOF PID for D0 selection

fcolamar:
Activates the possibility of using TPC and TOF approach, as done in Run2, following the general implementation introduced in [https://github.com/AliceO2Group/O2Physics/pull/2109](PR#2109).
Option enabled via a dedicated configurable, set to false as defaults to preserve backward compatibility


vkucera:
> Ciao @vkucera, ok, it makes sense indeed, PR updated. Fabio

Thanks @fcolamar

PR #2668, opened 2023-05-23
Efficient not stored derived data handling

victor-gonzalez:
Also 
- support for identified two-particle correlations analyses
- initial support for optional mixed events two-particle correlations


alibuild (approved):

PR #2695, opened 2023-05-27
Remove unnecessarily linked libs from core services

ddobrigk:
Should improve memory consumption (to be checked)

### DIFF
--- a/Common/TableProducer/CMakeLists.txt
+++ b/Common/TableProducer/CMakeLists.txt
@@ -28,42 +28,42 @@ o2physics_add_dpl_workflow(trackselection
 
 o2physics_add_dpl_workflow(event-selection
                     SOURCES eventSelection.cxx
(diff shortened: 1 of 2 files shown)


alibuild (approved):

ddobrigk:
@ekryshen it is possible that this was also inflating the memory footprint of the `event-selection` task. I will test this tomorrow on hyperloop and let you know...

ktf (approved):

alibuild (approved):

njacazio (dismissed):

ddobrigk:
Okay, this error is very confusing since it seems it also appears in Linux and I in fact tested this PR in Linux without issues. I will investigate more and try to reproduce this issue....

alibuild (approved):

ddobrigk:
> Okay, this error is very confusing since it seems it also appears in Linux and I in fact tested this PR in Linux without issues. I will investigate more and try to reproduce this issue....

My bad, was a simple mistake I did while editing the file. Should be fine now...

PR #2718, opened 2023-05-31
TPC PID: Add deuteron and triton selection in skimmer task

creetz16:


jezwilkinson:
@chiarazampolli @noferini Could this pull please be marked for merging ? We would like to have it included in production as soon as possible so that it can be used for the apass4 NN training

noferini (approved):

PR #2722, opened 2023-06-01
Analysis task for cascade correlations

rspijkers:
An analysis task for angular correlations between two cascades

### DIFF
--- a/PWGLF/Tasks/CMakeLists.txt
+++ b/PWGLF/Tasks/CMakeLists.txt
@@ -160,3 +160,8 @@ o2physics_add_dpl_workflow(antimatter-abs-hmpid
                     SOURCES AntimatterAbsorptionHMPID.cxx
                     PUBLIC_LINK_LIBRARIES O2Physics::AnalysisCore
                     COMPONENT_NAME Analysis)
+
+o2physics_add_dpl_workflow(strangecorrelations
+                    SOURCES strangecorrelations.cxx
+                    PUBLIC_LINK_LIBRARIES O2::Framework O2::DetectorsBase O2Physics::AnalysisCore O2::DetectorsVertexing
+                    COMPONENT_NAME Analysis)
--- a/PWGLF/Tasks/CMakeLists.txt
+++ b/PWGLF/Tasks/CMakeLists.txt
@@ -163,5 +163,5 @@ o2physics_add_dpl_workflow(antimatter-abs-hmpid
 
 o2physics_add_dpl_workflow(cascadecorrelations
                     SOURCES cascadecorrelations.cxx
-                    PUBLIC_LINK_LIBRARIES O2::Framework O2::DetectorsBase O2Physics::AnalysisCore O2::DetectorsVertexing
+                    PUBLIC_LINK_LIBRARIES O2Physics::AnalysisCore
                     COMPONENT_NAME Analysis)
--- a/PWGLF/Tasks/CMakeLists.txt
+++ b/PWGLF/Tasks/CMakeLists.txt
@@ -161,7 +161,7 @@ o2physics_add_dpl_workflow(antimatter-abs-hmpid
                     PUBLIC_LINK_LIBRARIES O2Physics::AnalysisCore
                     COMPONENT_NAME Analysis)
 
-o2physics_add_dpl_workflow(strangecorrelations
-                    SOURCES strangecorrelations.cxx
+o2physics_add_dpl_workflow(cascadecorrelations
+                    SOURCES cascadecorrelations.cxx
                     PUBLIC_LINK_LIBRARIES O2::Framework O2::DetectorsBase O2Physics::AnalysisCore O2::DetectorsVertexing
                     COMPONENT_NAME Analysis)
--- a/PWGLF/Tasks/cascadecorrelations.cxx
+++ b/PWGLF/Tasks/cascadecorrelations.cxx
@@ -75,6 +75,11 @@ struct cascadeSelector {
     },
   };
 
+  // Configurables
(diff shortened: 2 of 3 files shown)


PWGLF/Tasks/CMakeLists.txt
```diff
@@ -155,3 +155,8 @@ o2physics_add_dpl_workflow(v0postprocessing
                     SOURCES v0postprocessing.cxx
                     PUBLIC_LINK_LIBRARIES O2Physics::AnalysisCore
                     COMPONENT_NAME Analysis)
+
+o2physics_add_dpl_workflow(cascadecorrelations
+                    SOURCES cascadecorrelations.cxx
+                    PUBLIC_LINK_LIBRARIES O2::Framework O2::DetectorsBase O2Physics::AnalysisCore O2::DetectorsVertexing
```
ChiaraDeMartin95, line 161: Are you sure that O2Physics::AnalysisCore is not sufficient?

PWGLF/Tasks/cascadecorrelations.cxx
```diff
@@ -0,0 +1,234 @@
...
+      auto bachTrack = casc.bachelor_as<FullTracksExtIUWithPID>();
+      auto posTrack = v0data.posTrack_as<FullTracksExtIUWithPID>();
+      auto negTrack = v0data.negTrack_as<FullTracksExtIUWithPID>();
+
+      // PID
+      // Lambda check
+      if (casc.sign() < 0) {
+        // Proton check:
+        if (TMath::Abs(posTrack.tpcNSigmaPr()) > 3) {
+          cascflags(0);
```
ChiaraDeMartin95, line 99: I suggest using a configurable for the cut of the nSigmaTPC (instead of 3 as hard-coded number). In this way you will be able to modify your selection without recompiling the code.

ChiaraDeMartin95: The configurable could be different for different particles species to allow for more flexibility

PWGLF/Tasks/cascadecorrelations.cxx
```diff
@@ -0,0 +1,234 @@
...
+
+      if (casc.sign() < 0) { // FIXME: could be done better...
+        registry.fill(HIST("hMassXiMinus"), casc.mXi());
+        registry.fill(HIST("hMassOmegaMinus"), casc.mOmega());
+      } else {
+        registry.fill(HIST("hMassXiPlus"), casc.mXi());
+        registry.fill(HIST("hMassOmegaPlus"), casc.mOmega());
+      }
+
+      registry.fill(HIST("hPhi"), RecoDecay::phi(casc.px(), casc.py()));
```
ChiaraDeMartin95, line 173: To get the phi, you could use the method phi() implemented in the LFStrangenessTable (which by the way does exactly what you are doing)

PWGLF/Tasks/cascadecorrelations.cxx
```diff
@@ -0,0 +1,234 @@
...
+      auto lambda1 = c1.v0_as<o2::aod::V0sLinked>();
+      if (!(lambda0.has_v0Data()) || !(lambda1.has_v0Data())) {
+        continue; // reject if no v0data in either of the lambda's
+      }
+      auto v0data0 = lambda0.v0Data();
+      auto v0data1 = lambda1.v0Data();
+
+      double phi0 = RecoDecay::phi(c0.px(), c0.py());
+      double phi1 = RecoDecay::phi(c1.px(), c1.py());
+      double dphi = std::fmod(phi0 - phi1 + 2.5 * PI, 2 * PI) - 0.5 * PI;
```
ChiaraDeMartin95, line 187: Is this analogous to the method ComputeDeltaPhi(Double_t phi1, Double_t phi2) used in https://github.com/AliceO2Group/O2Physics/blob/master/PWGLF/Tasks/hStrangeCorrelation.cxx?

PWGLF/Tasks/cascadecorrelations.cxx
```diff
@@ -0,0 +1,234 @@
...
+        }
+        cascflags(1);
+        continue;
+      } else if (TMath::Abs(bachTrack.tpcNSigmaKa()) < 3) {
+        cascflags(2);
+        continue;
+      }
+      // if we reach here, the bachelor was neither pion nor kaon
+      cascflags(0);
+    } // cascade loop
```
ChiaraDeMartin95, line 134: Do you plan to implement some topological selections in this part of the code?

ChiaraDeMartin95 (changes requested):

PWGLF/Tasks/cascadecorrelations.cxx
rspijkers: Yes, in the future I will implement that!

PWGLF/Tasks/cascadecorrelations.cxx
rspijkers: Yes, it is in principle the same. If desirable I can change it to their format method, or maybe put the method somewhere such that people can easily call it from their analysis task

vkucera: There is already a configurable [RecoDecay::constrainAngle](https://github.com/AliceO2Group/O2Physics/blob/master/Common/Core/RecoDecay.h#L225) function for this, which can be directly fed with `phi0 - phi1`.
Please let's not make thousands versions of this.

rspijkers: Hi Vit, I wasn't aware of this method, I've implemented it now, thanks for pointing it out!

ChiaraDeMartin95 (approved):

PR #2724, opened 2023-06-01
PWGLF: tasks update for the efficiency calculation

nepeivodaRS:
cascqaanalysis.cxx: 

1) Some topological + kinematic pre-selections and event selections are moved to separate standalone functions AcceptCascCandidate and AcceptEvent
2) 2 new processes: processMCrec and processMCgen are added to treat MC sim data separately from real data (processData function) 1) processMCrec is basically the same as processData but with calculations of PDGcode and isPrimary for the mycascades table. In processData those columns are filled with -1. 2) processMCgen process is for creating TH3F hists of mc generated Xi and Omega after the same event selection as applied for data (axes: pt, rapidity and multFT0M (which is filled with 0s since there's no centrality calibration for MC collisions, so MB eff. will be used in analysis)).

cascpostprocessing.cxx
1) del. of 2 configurables (sel8 and cutzvertex), which can't be used at this step
2) hPtCascMinusTrueRec and hPtCascPlusTrueRec histos for the eff. calculation


PWGLF/Tasks/cascqaanalysis.cxx
```diff
@@ -140,44 +165,155 @@ struct cascqaanalysis {
...
+      return false;
+    }
+  }
 
-  void process(soa::Join<aod::Collisions, aod::EvSels, aod::Mults, aod::CentFT0Ms, aod::CentFV0As>::iterator const& collision, soa::Filtered<aod::CascDataExt> const& Cascades, aod::V0sLinked const&, aod::V0Datas const&, DauTracks const& tracks)
+  template <typename TCollision>
+  bool AcceptEvent(TCollision const& coll, bool isFillEventSelectionQA)
   {
-    // Event selection
-    if (sel8 && !collision.sel8()) {
```
njacazio, line 148: Hi @nepeivodaRS I would keep the same notation to keep changes to the bare bone

ercolessi: Hi @njacazio you can merge I have no comments

njacazio (dismissed):

PWGLF/Tasks/cascqaanalysis.cxx
njacazio: let's merge then!

PR #2732, opened 2023-06-01
PWGJE : adding charged trigger info to QA task

nzardosh:


alibuild (approved):

PR #2733, opened 2023-06-01
PWGHF: Add tree creator for DplusToPiKPi

AlexBigO:
Ciao @fgrosa , here is a first proposition for the tree creator of Dplus.

The goal would be to use  `HfCand3ProngFull` as an input for BDT training. It should contain for information than needed for this purpose (there are all the topological variables used in the Dplus selector + PID Nsigma), with the `OriginMcRec` flag used to label data for the BDT.


fgrosa:
Ciao @AlexBigO thanks a lot!
In general it looks good to me (also because it's very similar to the D0 one). I would however suggest to add a configurable that allows to store a "reduced" version of this table, since it is unlikely that we need all these variables. For instance, in the light version I would include only:
```cpp
                  hf_cand::Chi2PCA,
                  full::DecayLength,
                  full::DecayLengthXY,
                  full::DecayLengthNormalised,
                  full::DecayLengthXYNormalised,
                  full::PtProng0,
                  full::PtProng1,
                  full::PtProng2,
                  hf_cand::ImpactParameter0,
                  hf_cand::ImpactParameter1,
                  hf_cand::ImpactParameter2,
                  full::NSigTpcPi0,
                  full::NSigTpcKa0,
                  full::NSigTofPi0,
                  full::NSigTofKa0,
                  full::NSigTpcPi1,
                  full::NSigTpcKa1,
                  full::NSigTofPi1,
                  full::NSigTofKa1,
                  full::NSigTpcPi2,
                  full::NSigTpcKa2,
                  full::NSigTofPi2,
                  full::NSigTofKa2,
                  full::CandidateSelFlag,
                  full::M,
                  full::Pt,
                  full::Cpa,
                  full::CpaXY,
                  full::MaxNormalisedDeltaIP,
                  full::Eta,
                  full::Phi,
                  full::Y,
                  hf_cand_3prong::FlagMcMatchRec,
                  hf_cand_3prong::OriginMcRec
```  
If you think there are other variables to consider you can add them, but I would avoid to add many more.
Cheers, Fabrizio

AlexBigO:
Ciao @fgrosa ! Thanks for the detailed reply. I ended up writing a whole new `process` function instead of including a configurable (as we are only interested in this new table for our ML training purposes).

PWGHF/TableProducer/treeCreatorDplusToPiKPi.cxx
```diff
@@ -0,0 +1,430 @@
...
+          RecoDecay::y(array{particle.px(), particle.py(), particle.pz()}, RecoDecay::getMassPDG(particle.pdgCode())),
+          particle.flagMcMatchGen(),
+          particle.originMcGen());
+      }
+    }
+  }
+
+  PROCESS_SWITCH(HfTreeCreatorDplusToPiKPi, processMc, "Process MC", false);
+
+  void processForMLTraining(soa::Filtered<soa::Join<aod::HfCand3Prong, aod::HfCand3ProngMcRec, aod::HfSelDplusToPiKPi>> const& candidates,
```
fcatalan92, line 408:
```suggestion
  void processReducedTableMC(soa::Filtered<soa::Join<aod::HfCand3Prong, aod::HfCand3ProngMcRec, aod::HfSelDplusToPiKPi>> const& candidates,
```
I would use *reduced table* instead of *train table* and *ML training* which is in general more clear IMHO. As you state in the task description, these tables can be used for other things than ML training.

PWGHF/TableProducer/treeCreatorDplusToPiKPi.cxx
```diff
@@ -0,0 +1,430 @@
...
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+/// \file treeCreatorDplusToPiKPi.cxx
+/// \brief Writer of D+ → π+ K- π+ candidates in the form of flat tables to be stored in TTrees.
+///        Intended for debug, local optimization of analysis on small samples or BDT training.
```
fcatalan92, line 14:
```suggestion
///        Intended for debug, local optimization of analysis on small samples or ML training.
```

fcatalan92 (commented):
@AlexBigO I have just a couple of minor comments. On top of that, I think you should also implement the production of the reduced table for the data. Usually, the background sample for the ML training is obtained from real data, so such a table is needed.

PWGHF/TableProducer/treeCreatorDplusToPiKPi.cxx
AlexBigO: Ciao @fcatalan92 ! Thanks for the feedback.

I chose `Lite` instead of `Reduced` (as the latter is already used in the reduced workflow).

Indeed, as we need to implement it for data and for MC, I defined a configurable (which was Fabrizio's initial suggestion :D) to fill this lite table.

fcatalan92: Now with the configurable, you are always producing the large table even if you want only the small one. I would make this exclusive filling only one table at a time (this should have benefits in terms of memory and disk consumption).

AlexBigO: I fully agree, thanks :)

fcatalan92 (approved):

PWGHF/TableProducer/treeCreatorDplusToPiKPi.cxx
```diff
@@ -0,0 +1,429 @@
...
+      candidate.eta(),
+      candidate.phi(),
+      y,
+      flagMc,
+      origin);
+  }
+
+  template <typename T, typename U>
+  void fillTable(const T& candidate, const U& prong0, const U& prong1, const U& prong2, int selectionFlag, double invMass,
+                 double ct, double y, double e, int8_t flagMc, int8_t origin)
```
vkucera, line 271: Are you passing all these arguments from outside to avoid code duplication inside `fillTable` and `fillLiteTable`?

vkucera (changes requested):

PWGHF/TableProducer/treeCreatorDplusToPiKPi.cxx
vkucera: I mean that all the arguments you pass can be accessed from `candidate`.

vkucera: But now I see that there is also a big overlap between the two `process` functions which could also be factored out into a single filling function.

vkucera: > > I mean that all the arguments you pass can be accessed from `candidate`.
> 
> Indeed, some of them can be. However, for `prong0,1,2`, it does not compile if I define them inside the `fillTable` function.

OK, fair enough. Let's keep the prongs where they are then. :-)

vkucera (approved):

PR #2741, opened 2023-06-02
Split efficiency histos according to PID detector

rosarioturrisi:


DPG/Tasks/AOTTrack/qaMatchEff.cxx
```diff
@@ -1160,6 +1169,15 @@ struct qaMatchEff {
               histos.get<TH1>(HIST("data/phihist_tpc_piminus"))->Fill(track.phi());
               histos.get<TH1>(HIST("data/etahist_tpc_piminus"))->Fill(track.eta());
             }
+            if (pionPIDwithTPC && !pionPIDwithTOF) {
```
mfaggin, line 1172:
```suggestion
            if (pionPIDwithTPC) {
```

DPG/Tasks/AOTTrack/qaMatchEff.cxx
```diff
@@ -1237,6 +1255,15 @@ struct qaMatchEff {
                 histos.get<TH1>(HIST("data/phihist_tpcits_piminus"))->Fill(track.phi());
                 histos.get<TH1>(HIST("data/etahist_tpcits_piminus"))->Fill(track.eta());
               }
+              if (pionPIDwithTPC && !pionPIDwithTOF) {
```
mfaggin, line 1258:
```suggestion
              if (pionPIDwithTPC) {
```

rosarioturrisi:
Hi,
I got the point, nonetheless I would add the conditions you suggest, leaving the stricter conditions I originally set in order to disentangle more clearly possible effects depending on the particle or the detector. Of course K,P can be added.

mfaggin:
> Hi, I got the point, nonetheless I would add the conditions you suggest, leaving the stricter conditions I originally set in order to disentangle more clearly possible effects depending on the particle or the detector. Of course K,P can be added.

Hi @rosarioturrisi ok it's fine for me if you add what I suggest as an additional case

DPG/Tasks/AOTTrack/qaMatchEff.cxx
```diff
@@ -1206,6 +1263,21 @@ struct qaMatchEff {
...
+              histos.get<TH1>(HIST("data/pthist_tpc_pr_PIDTPC_O"))->Fill(trackPt);
+            }
+	    if (protonPIDwithTPC) {
+              histos.get<TH1>(HIST("data/pthist_tpc_pr_PIDTPC"))->Fill(trackPt);
+            }	    
+            if (!protonPIDwithTPC && protonPIDwithTOF) {
+              histos.get<TH1>(HIST("data/pthist_tpc_pr_PIDTOF_O"))->Fill(trackPt);
+            }
+            if (protonPIDwithTOF) {
+              histos.get<TH1>(HIST("data/pthist_tpc_pr_PIDTOF"))->Fill(trackPt);
```
mfaggin, line 1276:
```suggestion
            if (protonPIDwithTOF) {
              histos.get<TH1>(HIST("data/pthist_tpc_pr_PIDTOF"))->Fill(trackPt);
              if(!protonPIDwithTPC) {
                   histos.get<TH1>(HIST("data/pthist_tpc_pr_PIDTOF_O"))->Fill(trackPt);
              }
            }
```

mfaggin (changes requested):
Hi @rosarioturrisi , 2 things:

1.  I suggest another way of writing the histo filling for particles identified with at least 1 detector or with only 1 detector. The one I picked-up is just an example, please consider to apply to all cases (if you have strong opinions against it please tell me);
2. I think there one thing to fix. The histo fillings as in point 1. are inside the `if(isPion, isKaon, isProton)` statements, which are always defined as follows: `bool isPion = (isPIDPionRequired && pionPIDwithTPC && (!trkWithTOF || pionPIDwithTOF));` (here just the pion example). This means that you always identify tracks with TPC+TOF when available, but instead in point 1. you also want cases in which eg the particle is identified by TOF but not by TPC. This needs to be fixed in my opinon.

rosarioturrisi:
Hi Mattia, thanks (as usual) for the comments, I cleaned the logic and leaved some redundancy in a few conditions. Curiously, just adding a few lines to the committed file triggers the code checks, had to clang again with some weird results...but worked.

mfaggin (dismissed):
Thanks @rosarioturrisi , now it looks fine to me

mfaggin:
hi @TimoWilken in this PR we have an error on macOS compilation that, from a quick look, does not sound to be related to our changes. Is it related to the migrations you mentioned via mail?

mfaggin (approved):

PR #2742, opened 2023-06-02
[EventFiltering] Add scripts for BDT trainings and add info for particles reco in wrong collision

fgrosa:


alibuild (approved):

vkucera:
There were several Python errors reported by `pyright` that would be worth fixing.

PR #2754, opened 2023-06-05
[EMCAL-850] Use unsorted preslice to slice according to foundBC

mfasDa:


alibuild (approved):
Auto-approving on behalf of @mfasDa.

PR #2765, opened 2023-06-06
PWGDQ: Adding collision index for dilepton -hadron combination

lucamicheletti93:
- Adding collisionId and (x,y,z) coordinates to DileptonsExtra
- Fixing the Matching of the collisions among DQ and HF candidates tables
- Updating the labels of 2D histograms


iarsene:
Ciao Luca, maybe we can talk in more details also on mattermost, but would be good to understand first why you need the collision id in the skimmed data model and whether there are alternatives. This addition affects all the workflows, where its mostly not used.

iarsene (approved):

iarsene (dismissed):

iarsene (approved):

PR #2766, opened 2023-06-06
[PWGJE Validation] Adding configurable for flexible usage of evSel7,8

jloemker:
Hi @fkrizek, @archita-dash,

in order to get more jets for the comparison to AliPhysics I included a configurable for the event selection 7 (or 8 for run3). 
As we apply a jet-to-collision association we restricted the amount of jets in O2Physics to jets from collisions that passed the evSel7. Knowing that the Jetfinder from O2Physics does not use the evSel7 for run2 we should have the ability to consider collisions (and thus collision associated jets) with and without the sel7.

**Excluding the evSel7 leads to slightly (when tested on one single AOD.root) more jets in the jetvalidation output !** 

Lets hope that this PR gets merged soon such that we can make a proper comparison for the output with and without the evSel7.

Cheerio,
Johanna

### DIFF
--- a/PWGJE/TableProducer/jetfinder.cxx
+++ b/PWGJE/TableProducer/jetfinder.cxx
@@ -38,6 +38,7 @@ struct JetFinderTask {
   Configurable<float> trackPhiMin{"trackPhiMin", -999, "minimum track phi"};
   Configurable<float> trackPhiMax{"trackPhiMax", 999, "maximum track phi"};
   Configurable<std::string> trackSelections{"trackSelections", "globalTracks", "set track selections"};
+  Configurable<std::string> evSel{"evSel", "evSel8", "choose event selection"};
 
   // cluster level configurables
   Configurable<std::string> clusterDefinitionS{"clusterDefinition", "kV3Default", "cluster definition to be selected, e.g. V3Default"};
@@ -108,7 +109,7 @@ struct JetFinderTask {
   void processChargedJets(soa::Filtered<soa::Join<aod::Collisions, aod::EvSels>>::iterator const& collision,
                           JetTracks const& tracks)
   {
-    if (!selectCollision(collision)) {
+    if (!selectCollision(collision, evSel)) {
       return;
     }
     inputParticles.clear();
--- a/PWGJE/TableProducer/jetfinder.h
+++ b/PWGJE/TableProducer/jetfinder.h
@@ -66,7 +66,7 @@ bool selectTrack(T const& track, std::string trackSelection)
     return false;
   } else if (trackSelection == "QualityTracks" && !track.isQualityTrack()) {
     return false;
-  } else if (trackSelection == "hybridTracksJE" && !track.trackCutFlagFb5()) { // isQualityTrack
+  } else if (trackSelection == "hybridTracksJE" && !track.trackCutFlagFb5()) {
     return false;
   } else {
     return true;
@@ -237,9 +237,12 @@ void analyseParticles(std::vector<fastjet::PseudoJet>& inputParticles, float par
 }
 
 template <typename T>
-bool selectCollision(T const& collision)
+bool selectCollision(T const& collision, std::string evSel)
 {
-  if (!collision.sel8()) {
+  if (evSel == "evSel8" & !collision.sel8()) {
+    return false;
+  }
+  if(evSel == "evSel7" & !collision.sel7()){
     return false;
   }
   return true;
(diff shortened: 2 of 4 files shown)


nzardosh:
Hi @jloemker . If I understand correctly your aim is to get jets with sel7() instead of sel8(). However the validation task does not contain any jet finding and takes jets from the main jet finder. Therefore in order to do jet finding with this selection you need to change the jetfinder. 
this is the function you need to add the selection to : https://github.com/AliceO2Group/O2Physics/blob/586ab7f44760f3124ccdbd1d2d491204ba1e7bb1/PWGJE/TableProducer/jetfinder.h#L240 
you can add it in a similar way as is done for tracks:
https://github.com/AliceO2Group/O2Physics/blob/586ab7f44760f3124ccdbd1d2d491204ba1e7bb1/PWGJE/TableProducer/jetfinder.h#L63

You then put the selction configurable in the inclusive and Hf jet finders.
Thanks,
Nima

jloemker:
Hi @nzardosh , @fkrizek, @archita-dash,

I included the suggestion from Nima to the jetfinder and I added the jet-to-collision association in the validation task. On a local check over one AOD file it turns out to be a real game changer !
With the configurable for the evSel7 in the jetfinder we obtain **twice as much jetQA entries** in the validation task.

Many thanks to Nima !

Cheerio,
Johanna

mfasDa:
> Hi @jloemker . If I understand correctly your aim is to get jets with sel7() instead of sel8(). However the validation task does not contain any jet finding and takes jets from the main jet finder. Therefore in order to do jet finding with this selection you need to change the jetfinder. Currently with your PR the produced jet population doesnt change and if you add a sel7() cut you will have it mixing with the sel8() this is the function you need to add the selection to :
> 
> https://github.com/AliceO2Group/O2Physics/blob/586ab7f44760f3124ccdbd1d2d491204ba1e7bb1/PWGJE/TableProducer/jetfinder.h#L240
> 
> 
> you can add it in a similar way as is done for tracks:
> https://github.com/AliceO2Group/O2Physics/blob/586ab7f44760f3124ccdbd1d2d491204ba1e7bb1/PWGJE/TableProducer/jetfinder.h#L63
> 
> You then put the selction configurable in the inclusive and Hf jet finders. Thanks, Nima

Hi @nzardosh, 

not sure it is so good to have a trigger selection in the jet finder. For run2 data it does not harm, since all triggers are a subset of FIT (TVX), however in run2 triggers downscaled independently and a trigger selection in the jet finder make analysis of i.e. EMCAL-triggered data on converted AODs impossible. Also I don't see the benefit a trigger selection provides at producer level, the design goal is normally to have the most flexible selection at producer level so that it doesn't bias analyses that consume objects from the producer. Please consider dropping this! 

Thanks, Markus

jloemker:
Hello @mfasDa , @nzardosh ,

maybe this **additional option evSel=="None" is a good compromise** for all of us ?
This way everyone can select the evSel7, evSel8 or no no requirement on the evSel in the Jetfinder.

Else I would reset the commit to the original PR, which means that we keep the standard evSel8 in the jetfinder for run2 and run3. This should be disfavored as any analysis on converted run2 AOD files will have an evSel mismatch (Nima pointed that out) and the run3 analyzers might bias their jetselection (what Markus mentioned).

I hope this proposed solution satisfies everyone.

Cheerio,
Johanna

mfasDa:
> Hello @mfasDa , @nzardosh ,
> 
> maybe this **additional option evSel=="None" is a good compromise** for all of us ? This way everyone can select the evSel7, evSel8 or no no requirement on the evSel in the Jetfinder.
> 
> Else I would reset the commit to the original PR, which means that we keep the standard evSel8 in the jetfinder for run2 and run3. This should be disfavored as any analysis on converted run2 AOD files will have an evSel mismatch (Nima pointed that out) and the run3 analyzers might bias their jetselection (what Markus mentioned).
> 
> I hope this proposed solution satisfies everyone.
> 
> Cheerio, Johanna

Hi @jloemker ,

I would be in favour of this, it kind of reflects the condition "kAny" which we applied in run2 analyses for service wagons. Service wagons should never select events, this is the task of the user wagons, otherwise you introduce biases. 

Concerning INT8 and run2 analyses: Also in run2 we were running TVX-based min bias triggers, which however see approx half of the luminosity compared to the V0M cross section, so for analyses on run2 data it is not really advantageous. Furthermore with the event selection you apply you require an AND of V0M and TVX, which however should not lead to a significant loss on the TVX side.

Cheers

Markus

mfasDa (approved):

PR #2767, opened 2023-06-07
EventFiltering/PWGHF: Add safety on test fraction

AlexBigO:


EventFiltering/PWGHF/Macros/train_hf_triggers.py
```diff
@@ -192,10 +192,12 @@ def data_prep(config):
     )
 
     labels_array = np.array([0]*n_bkg + [1]*n_prompt + [2]*n_nonprompt)
-    if test_f < 1:
+    if 0 < test_f < 1:
         train_set, test_set, y_train, y_test = train_test_split(
             df_tot, labels_array, test_size=test_f, random_state=seed_split
         )
+    else:
+        print("ERROR: test_fraction must belong to ]0,1[")
```
fgrosa, line 200: I would also suggest to exit the execution if this error is raised:
```suggestion
        print("ERROR: test_fraction must belong to ]0,1[")
        sys.exit(0)
```

fgrosa (commented):
Thanks @AlexBigO! I just have a small comment (see below)

EventFiltering/PWGHF/Macros/train_hf_triggers.py
AlexBigO: Thanks!

fgrosa (approved):

vkucera:
@fgrosa @AlexBigO please silence the `pylint` errors related to importing non-standard modules, so that the PR with automatic fixes can be made.
There is also a warning reported by `bandit` concerning unsafe loading of yaml files which should be fixed.

fgrosa:
> @fgrosa @AlexBigO please silence the `pylint` errors related to importing non-standard modules, so that the PR with automatic fixes can be made. There is also a warning reported by `bandit` concerning unsafe loading of yaml files which should be fixed.

Hi @vkucera, I think that there is something misconfigured in the MegaLinter test for pylint, infact several errors for not found modules should not be there. For instance, it raises errors for `numpy` and `pandas` which should be included (see https://github.com/alisw/alidist/blob/58384186dcd5dc5d10d21a685157f4bea257f0b0/python-modules-list.sh). So before adding exceptions to `pylint` because it does raises errors, I think that it should be properly configured. Regarding `safe_load`, it is a non-blocking error of a not-required test of the code and it does not work with our `yaml` files because it does not support python tuples
```
yaml.constructor.ConstructorError: could not determine a constructor for the tag 'tag:yaml.org,2002:python/tuple'
```
Also, the only advantage to use `safe_load` instead of `load` is to read `yaml` files from unreliable and untrusted sources, which is not our case.

PR #2769, opened 2023-06-07
Change in hStrangeCorrelation

lhusova:
implementation of h-Pion correlations


ddobrigk:
Thanks a lot, @lhusova ! The megalinter I think we can go on without for now - I still want to do some small cleanup and maybe I can take care of that in the next round. In fact, this also reminds me we did not yet fully converge on a DCAxy selection for the triggers (and then also for the pions!), but also this we can implement in a next PR. Perhaps we can merge this as is and then I try to merge my local changes and then we can go on with testing also the h-Pi in the next train :-D

(tagging @ChiaraDeMartin95 @kcui1 for completeness)

njacazio (approved):

PR #2772, opened 2023-06-07
updated eventSelections Configurable

Archita-Dash:


Archita-Dash:
Hi @nzardosh. 
As suggested by you, I changed 

1. the event selection configurable from "evSel" to "eventSelections", 
2.  cased the above configurable as a string inside of **init**,
3. removed the "None" case for the event selection as if neither sel7 or sel8 is selected then it'll return true any way.

nzardosh (approved):

PR #2774, opened 2023-06-07
PWGEM/PhotonMeson: add conversion point Rxy vs. phi

dsekihat:
PWGEM/PhotonMeson: add conversion point Rxy vs. phi


mhemmer-cern (approved):
looks good

PR #2777, opened 2023-06-08
PWGHF: Remove Lc selection when filling the MC tree

saganatt:


saganatt:
For MC files, only signal candidates are stored in the tree. This is not the expected behavior, the PR fixes that.

saganatt:
@vkucera MegaLinter errors are for the unchanged code, should I fix them as well?

PR #2778, opened 2023-06-08
Drop unneeded rootmap files

ktf:


ddobrigk:
Hi @ktf , seems like there is still some whitespace issue. Can you please take a look? (Btw sometimes the CI picks up on tabs in commented lines (yes, comments) and the alibot does not actually offer to fix those. If this is the case here, you have to look for a tab in your favorite editor and fix it manually ... Just in case it helps). Have a nice Sunday!

ktf:
Does not actually work. Closing.

ktf:
I think I managed to fix the issue. Let's see if it's a single one or there is more. Apparently it was indeed a missing dependency.

cmake/AddRootDictionary.cmake
```diff
@@ -111,7 +110,6 @@ function(add_root_dictionary target)
   set(dictionaryFile ${CMAKE_CURRENT_BINARY_DIR}/${dictionary}.cxx)
   set(pcmBase ${dictionary}_rdict.pcm)
   set(pcmFile ${lib_output_dir}/${pcmBase})
-  set(rootmapFile ${lib_output_dir}/lib${basename}.rootmap)
```
ktf, line 113:
```suggestion
```

PR #2783, opened 2023-06-09
PWG-UD: Correct copy of arrays from table to table

pbuehler:


alibuild (approved):

PR #2787, opened 2023-06-09
Updates to QA the skimmed data

mpuccio:
- Add global-TOF match correlation monitoring
- Add macro to directly cross check the skimming
- Additional check of missing BCs in AO2D


alibuild (approved):

PR #2809, opened 2023-06-14
Tools/KFParticle: Fix small bug on the processMC

lupengzhong:
Fix a small bug on the processMC on qaKFParticle.cxx and qaKFEventTrack.cxx when running on MC


ddobrigk:
Be cautious with `abs` : if I remember correctly, `abs` is for integers and typecasts to integers. You probably want to make sure to use `fabs` for DCAs instead: can you please check?

lupengzhong:
Thank you David for your reminder. I changed "abs" accordingly to "fabs".

PR #2813, opened 2023-06-15
EventFiltering: Add missing if statement before filling QA histos for V0s

fgrosa:


alibuild (approved):

PR #2816, opened 2023-06-15
EventFiltering: Add missing condition for QA histos

fgrosa:


alibuild (approved):

PR #2821, opened 2023-06-16
Remove TGeo load from track extension

ddobrigk:


ktf:
Glad it was that easy!

jgrosseo (approved):

PR #2822, opened 2023-06-16
PWGHF: Remove TGeo load from HF creators

fgrosa:
@vkucera @fcolamar @ginnocen this PR removes the unnecessary load of the geometry, which should also allow us to reduce by a good amount the memory usage on hyperloop.


vkucera:
Thanks a lot @fgrosa ! Can you please also fix it in `candidateCreatorToXiPi.cxx`?

fgrosa:
> `candidateCreatorToXiPi`

Thanks, I missed that file indeed! Now fixed

vkucera:
> > `candidateCreatorToXiPi`
> 
> Thanks, I missed that file indeed! Now fixed

Thanks!

vkucera (approved):

PR #2826, opened 2023-06-17
Add missing references

ktf:
Without it only works if the passed argument is a reference itself


ktf:
Needed by https://github.com/AliceO2Group/AliceO2/pull/11517.

ddobrigk (approved):

PR #2837, opened 2023-06-19
EventFiltering: switch from mass to delta mass for D resonances

fgrosa:


alibuild (approved):

PR #2841, opened 2023-06-19
PWGHF: Implement ML inference class

AlexBigO:
Hello @fcatalan92 @fgrosa @iouribelikov !

Here are the modifications applied to the base structure provided by Fabio: the core of the inference class in the header `HFMLResponse.h` and an example of application in the Dplus selector task.

The functionalities are the following:

- configure `HFMLResponse` instance with configurables provided in JSON. There are two `config` methods, one when CCDB access (that does not change `mPaths` attribute) and one when no CCDB access (that sets `mPaths` to local paths provided)
- - possibility to retrieve model from CCDB (dedicated method for this)
- initialize `HFMLResponse` instance: this will initialize the OnnxModel instances (one for each model/pt bin) that we'll work with inside the methods defined
- compute combined Nsigma for TPC and TOF --> move it to `TrackSelectorPID` in a next PR
- method returning a boolean to know whether the ML output passes the selections pre-defined in JSON (+ possibility to save this score as `std::vector` in AOD)
- new bin in QA histogram

The ML model is applied after Skim+Topol+PID selections. One can tune these selections (that are preselections for the ML models) via JSON.


PWGHF/TableProducer/candidateSelectorDplusToPiKPi.cxx
```diff
@@ -71,12 +87,18 @@ struct HfCandidateSelectorDplusToPiKPi {
...
+      labels[1 + aod::SelectionStep::RecoML] = "ML selection";
       static const AxisSpec axisSelections = {kNBinsSelections, 0.5, kNBinsSelections + 0.5, ""};
       registry.add("hSelections", "Selections;;#it{p}_{T} (GeV/#it{c})", {HistType::kTH2F, {axisSelections, {(std::vector<double>)binsPt, "#it{p}_{T} (GeV/#it{c})"}}});
       for (int iBin = 0; iBin < kNBinsSelections; ++iBin) {
         registry.get<TH2>(HIST("hSelections"))->GetXaxis()->SetBinLabel(iBin + 1, labels[iBin].data());
       }
     }
+
+    if (applyML) {
+      hfMLResponse.init(pTBinsML, cutsML, cutDirML, modelPathsML, true);
```
fgrosa, line 99: Is the optimisation actually doing something for the BDT models?

PWGHF/Core/HFMLResponse.h
```diff
@@ -0,0 +1,214 @@
...
+  {
+    // import configurables
+    mBinsLimits = binsLimits;
+    mCuts = cuts;
+    mCutDir = cutDir;
+
+    mNetworks = std::vector<o2::ml::OnnxModel>(mNModels);
+    mPaths = std::vector<std::string>(mNModels);
+    mNModels = binsLimits.size() - 1;
+    mNClasses = (cutDir.size() >= 3) ? cutDir.size() : 2;
```
fgrosa, line 123: I would avoid to infer the number of classes from this, because the risk is to always get `3` from the default value of the `hf_cuts_ml`. It would be better in my opinion to have an explicit argument, set by the user.

fgrosa (commented):
Hi @AlexBigO thanks a lot! To me this looks good enough for a first version (only a couple of minor comments). As an improvement (not needed for this iteration IMHO), I would suggest to have a derived class of the `HFMLResponse` specific for the D+ meson (or for the 3-prongs, to be seen) with a method that takes a row of the `aod::HfCand3Prong` table, builds the vector of features and feeds it to the `onnxModel`. If so, one could use it very simply in the [candidateSelectorDplusToPiKPi.cxx](https://github.com/AliceO2Group/O2Physics/blob/master/PWGHF/TableProducer/candidateSelectorDplusToPiKPi.cxx) task, reducing the code for the user.

PWGHF/TableProducer/candidateSelectorDplusToPiKPi.cxx
AlexBigO: Actually I don't know, I had trouble finding its actual purpose. But it was a typo when modifying the code, I meant to set it to `false` by default.

PWGHF/Core/HFMLResponse.h
AlexBigO: I am putting a cofigurable to choose the number of classes. But I had made tests and, if I initialize the `HFMLResponse` object inside the `void init(InitContext const&) ...` function, then it is the JSON value that is selected.

PWGHF/Core/HFMLResponse.h
```diff
@@ -0,0 +1,212 @@
...
+      ++iClass;
+    }
+    return true;
+  }
+
+ private:
+  std::vector<o2::ml::OnnxModel> mNetworks;       // OnnxModel objects, one for each bin
+  uint8_t mNModels = 1;                           // number of bins
+  uint8_t mNClasses = 3;                          // number of model classes
+  std::vector<double> mBinsLimits = {};           // bin limits of the variable (e.g. pT) used to select which model to use
```
fcatalan92, line 204: This variable is not used at the moment since the index to the model is provided by the "user". Maybe the selection of which model to use could be integrated in this class?

fcatalan92 (commented):
Hi @AlexBigO, nice job! I added a few comments, most of the minor

PWGHF/Core/HFMLResponse.h
```diff
@@ -0,0 +1,212 @@
...
+    }
+  }
+
+  // Initialize OnnxModels when paths already defined
+  /// \param paths is a vector of onnx model paths
+  /// \param enableOptimizations is a switch no enable optimizations
+  /// \param threads
+  void init(bool enableOptimizations = false, int threads = 0)
+  {
+    return init(mPaths, enableOptimizations, threads);
```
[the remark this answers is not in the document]
AlexBigO: Yes, I also noticed it and fixed it :)

PWGHF/Core/HFMLResponse.h
```diff
@@ -0,0 +1,212 @@
...
+    return std::sqrt((nSigTPC * nSigTPC + nSigTOF * nSigTOF) / 2);
+  }
+  if (TESTBIT(tagPID, PIDStatus::TPCOnly)) {
+    return std::abs(nSigTPC);
+  }
+
+  if (TESTBIT(tagPID, PIDStatus::TOFOnly)) {
+    return std::abs(nSigTOF);
+  }
+}
```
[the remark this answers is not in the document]
AlexBigO: Noted, I will move these methods in `TrackSelectorPID` in a next PR.

PWGHF/Core/HFMLResponse.h
AlexBigO: You mean feeding `candidate.pt()` to `isSelectedML` instead of the `pTBin`?

fcatalan92: Yes

AlexBigO: Done :)

PWGHF/Core/HFMLResponse.h
fgrosa: The JSON value is the one of the configurable, did I get properly?

AlexBigO: Yes, that's it :)

fcatalan92 (approved):

fcatalan92:
@vkucera, @fcolamar This PR is ready to be merged IMHO

PWGHF/Core/HFMLResponse.h
```diff
@@ -0,0 +1,225 @@
...
+#include <vector>
+#include <string>
+#include <map>
+
+#include "Tools/ML/model.h"
+
+#include "Framework/Array2D.h"
+#include "PWGHF/Core/SelectorCuts.h"
+
+#include "CCDB/CcdbApi.h"
```
vkucera, line 30: Sort

PWGHF/Core/HFMLResponse.h
```diff
@@ -0,0 +1,225 @@
...
+
+namespace o2::analysis
+{
+
+enum PIDStatus : uint8_t {
+  TPCAndTOF = 0,
+  TPCOnly,
+  TOFOnly
```
vkucera, line 38: Capitalise only the first letter of each "sub-word".

PWGHF/Core/HFMLResponse.h
```diff
@@ -0,0 +1,225 @@
...
+  }
+  if (TESTBIT(tagPID, PIDStatus::TPCOnly)) {
+    return std::abs(nSigTPC);
+  }
+
+  return std::abs(nSigTOF); // if (TESTBIT(tagPID, PIDStatus::TOFOnly))
+}
+
+template <typename T = float>
+class HFMLResponse
```
vkucera, line 77:
```suggestion
class HfMlResponse
```
and elsewhere

PWGHF/Core/HFMLResponse.h
```diff
@@ -0,0 +1,225 @@
...
+  /// Default destructor
+  virtual ~HFMLResponse() = default;
+
+  /// Configure class instance (import configurables)
+  /// \param binsLimits is a vector containing bins limits
+  /// \param cuts is a LabeledArray containing selections per bin
+  /// \param cutDir is a vector telling whether to reject score values greater or smaller than the threshold
+  /// \param nClasses is the number of classes for each model
+  /// \param paths is a vector of onnx model paths
+  void config(const std::vector<double>& binsLimits, const o2::framework::LabeledArray<double>& cuts, const std::vector<int>& cutDir,
```
vkucera, line 91:
```suggestion
  void configure(const std::vector<double>& binsLimits, const o2::framework::LabeledArray<double>& cuts, const std::vector<int>& cutDir,
```

PWGHF/Core/HFMLResponse.h
```diff
@@ -0,0 +1,225 @@
...
+              const uint8_t& nClasses)
+  {
+    mBinsLimits = binsLimits;
+    mCuts = cuts;
+    mCutDir = cutDir;
+    mNClasses = nClasses;
+    mNModels = binsLimits.size() - 1;
+    mModels = std::vector<o2::ml::OnnxModel>(mNModels);
+    mPaths = std::vector<std::string>(mNModels);
```
vkucera, line 117: This could be done also in the first version of `config` if `paths` is empty.

PWGHF/Core/HFMLResponse.h
```diff
@@ -0,0 +1,225 @@
...
+  /// Get vector with model predictions
+  /// \param input a vector containing the values of features used in the model
+  /// \param nModel is the model index
+  /// \return model prediction for each class and the selected model
+  template <typename T1, typename T2>
+  std::vector<T> getModelOutput(T1& input, const T2& nModel)
+  {
+    T* outputPtr = mModels[nModel].evalModel(input);
+    std::vector<T> output(outputPtr, outputPtr + mNClasses);
+    return output;
```
vkucera, line 161: You can return the expression directly.

PWGHF/Core/HFMLResponse.h
```diff
@@ -0,0 +1,225 @@
...
+namespace o2::analysis
+{
+
+enum PIDStatus : uint8_t {
+  TPCAndTOF = 0,
+  TPCOnly,
+  TOFOnly
+};
+
+template <typename T1>
```
vkucera, line 41: Put this line lower just before the method.

PWGHF/Core/HFMLResponse.h
```diff
@@ -0,0 +1,225 @@
...
+    std::vector<T> output(outputPtr, outputPtr + mNClasses);
+    return output;
+  }
+
+  /// ML selections
+  /// \param input is the input features
+  /// \param pt is the candidate transverse momentum
+  /// \return boolean telling if model predictions pass the cuts
+  template <typename T1, typename T2>
+  bool isSelectedML(T1& input, const T2& pt)
```
vkucera, line 169:
```suggestion
  bool isSelectedMl(T1& input, const T2& pt)
```

PWGHF/Core/HFMLResponse.h
```diff
@@ -0,0 +1,225 @@
...
+    return true;
+  }
+
+  /// ML selections
+  /// \param input is the input features
+  /// \param pt is the candidate transverse momentum
+  /// \param output is a container to be filled with model output
+  /// \return boolean telling if model predictions pass the cuts
+  template <typename T1, typename T2>
+  bool isSelectedML(T1& input, const T2& pt, std::vector<T>& output)
```
vkucera, line 194:
```suggestion
  bool isSelectedMl(T1& input, const T2& pt, std::vector<T>& output)
```

PWGHF/Core/HFMLResponse.h
```diff
@@ -0,0 +1,225 @@
...
+  {
+    auto nModel = o2::analysis::findBin(&mBinsLimits, pt);
+    output = getModelOutput(input, nModel);
+    uint8_t iClass{0};
+    for (const auto& outputValue : output) {
+      uint8_t dir = mCutDir.at(iClass);
+      if (dir != hf_cuts_ml::CutDirection::CutNot) {
+        if (dir == hf_cuts_ml::CutDirection::CutGreater && outputValue > mCuts.get(nModel, iClass)) {
+          return false;
+        } else if (dir == hf_cuts_ml::CutDirection::CutSmaller && outputValue < mCuts.get(nModel, iClass)) {
```
vkucera, line 204: Looks like `else` is not needed here.

PWGHF/Core/SelectorCuts.h
```diff
@@ -126,6 +126,42 @@ static const std::vector<std::string> labelsPt{};
...
+  1000};
+auto binsPt_v = std::vector<double>{binsPt, binsPt + nBinsPt + 1};
+
+// default values for the ML model paths, one model per pT bin
+static const std::vector<std::string> modelPaths = {
+  ""};
+
+// default values for the cut directions
+constexpr int cutDir[nCutScores] = {CutGreater, CutSmaller, CutSmaller};
+auto cutDir_v = std::vector<int>{cutDir, cutDir + nCutScores};
```
vkucera, line 152:
```suggestion
auto vecCutDir = std::vector<int>{cutDir, cutDir + nCutScores};
```

PWGHF/DataModel/CandidateSelectionTables.h
```diff
@@ -15,8 +15,10 @@
 #ifndef PWGHF_DATAMODEL_CANDIDATESELECTIONTABLES_H_
 #define PWGHF_DATAMODEL_CANDIDATESELECTIONTABLES_H_
 
-#include "Common/Core/RecoDecay.h"
+#include <vector>
+
 #include "Common/Core/TrackSelectorPID.h"
+#include "Common/Core/RecoDecay.h"
```
vkucera, line 21: Why the swap?

PWGHF/TableProducer/CMakeLists.txt
```diff
@@ -116,7 +116,7 @@ o2physics_add_dpl_workflow(candidate-selector-d0
 
 o2physics_add_dpl_workflow(candidate-selector-dplus-to-pi-k-pi
                     SOURCES candidateSelectorDplusToPiKPi.cxx
-                    PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::AnalysisCore
+                    PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::AnalysisCore O2::CCDB O2Physics::MLCore
```
vkucera, line 119: Is `O2::CCDB` needed?

vkucera:
> @vkucera, @fcolamar This PR is ready to be merged IMHO

Thanks, I am checking.

PWGHF/TableProducer/candidateSelectorDplusToPiKPi.cxx
vkucera, line 1: `ML` -> `Ml`

PWGHF/TableProducer/candidateSelectorDplusToPiKPi.cxx
```diff
@@ -39,15 +41,34 @@ struct HfCandidateSelectorDplusToPiKPi {
...
   Configurable<bool> activateQA{"activateQA", false, "Flag to enable QA histogram"};
+  // ML inference
+  Configurable<bool> applyML{"applyML", false, "Flag to apply ML selections"};
+  Configurable<std::vector<double>> binsPtML{"binsPtML", std::vector<double>{hf_cuts_ml::binsPt_v}, "pT bin limits for ML application"};
+  Configurable<std::vector<std::string>> modelPathsML{"modelPathsML", std::vector<std::string>{hf_cuts_ml::modelPaths}, "Paths of the ML models, one for each pT bin"};
+  Configurable<std::vector<int>> cutDirML{"cutDirML", std::vector<int>{hf_cuts_ml::cutDir_v}, "Whether to reject score values greater or smaller than the threshold"};
+  Configurable<LabeledArray<double>> cutsML{"ml_cuts", {hf_cuts_ml::cuts[0], hf_cuts_ml::nBinsPt, hf_cuts_ml::nCutScores, hf_cuts_ml::labelsPt, hf_cuts_ml::labelsCutScore}, "ML selections per pT bin"};
+  Configurable<int8_t> nClassesML{"nClassesML", (int8_t)hf_cuts_ml::nCutScores, "Number of classes in ML model"};
+  // CCDB configuration
+  o2::ccdb::CcdbApi ccdbApi;
```
vkucera, line 63: This is not a `Configurable`. Respect the order of members.

PWGHF/TableProducer/candidateSelectorDplusToPiKPi.cxx
```diff
@@ -39,15 +41,34 @@ struct HfCandidateSelectorDplusToPiKPi {
...
+  Configurable<std::string> url{"ccdb-url", "http://alice-ccdb.cern.ch", "url of the ccdb repository"};
```
vkucera, line 64: Follow the naming conventions.
```suggestion
  Configurable<std::string> ccdbUrl{"ccdbUrl", "http://alice-ccdb.cern.ch", "url of the ccdb repository"};
```

PWGHF/TableProducer/candidateSelectorDplusToPiKPi.cxx
```diff
@@ -216,13 +258,62 @@ struct HfCandidateSelectorDplusToPiKPi {
...
       }
 
+      if (applyML) {
+        // tracks without TPC nor TOF already rejected at PID selection level
+        // tag tracks with TPC and TOF, only TPC, only TOF
+        auto tagPIDPos1 = o2::analysis::tagPID(trackPos1);
+        auto tagPIDNeg = o2::analysis::tagPID(trackNeg);
+        auto tagPIDPos2 = o2::analysis::tagPID(trackPos2);
+
+        auto combinedNSigmaPiPos1 = getCombinedNSigma(trackPos1.tpcNSigmaPi(), trackPos1.tofNSigmaPi(), tagPIDPos1);
```
vkucera, line 278: How come you don't need `o2::analysis::getCombinedNSigma` as you do for `tagPID`?
I don't see `using namespace o2::analysis`.

vkucera (changes requested):

PWGHF/Core/HFMLResponse.h
```diff
@@ -0,0 +1,225 @@
...
+}
+
+template <typename T = float>
+class HFMLResponse
+{
+ public:
+  /// Default constructor
+  HFMLResponse() = default;
+  /// Default destructor
+  virtual ~HFMLResponse() = default;
```
vkucera, line 83: Why is the destructor virtual?

AlexBigO: It was a remnant. I will remove it.

PWGHF/Core/HFMLResponse.h
AlexBigO: Indeed, I only need one `configure` method then, thanks.

PWGHF/TableProducer/candidateSelectorDplusToPiKPi.cxx
AlexBigO: It woked without it but this is strange indeed. I added the prefix.

PWGHF/TableProducer/CMakeLists.txt
AlexBigO: This is not needed, I removed it :)

PWGHF/Core/HfMlResponse.h
```diff
@@ -0,0 +1,212 @@
...
+/// \file HFMLResponse.h
+/// \brief Class to compute the ML response for HF-analysis selections
+/// \author Fabio Catalano <fabio.catalano@cern.ch>, CERN
+/// \author Alexandre Bigot <alexandre.bigot@cern.ch>, IPHC Strasbourg
+
+#ifndef PWGHF_CORE_HFMLRESPONSE_H_
+#define PWGHF_CORE_HFMLRESPONSE_H_
+
+#include <onnxruntime/core/session/experimental_onnxruntime_cxx_api.h>
+#include <map>
```
vkucera, line 21:
```suggestion
#include <onnxruntime/core/session/experimental_onnxruntime_cxx_api.h>

#include <map>
```

PWGHF/Core/HfMlResponse.h
```diff
@@ -0,0 +1,212 @@
...
+#include <string>
+#include <vector>
+
+#include "CCDB/CcdbApi.h"
+#include "Framework/Array2D.h"
+#include "Tools/ML/model.h"
```
vkucera, line 27:
```suggestion
#include "Framework/Array2D.h"

#include "Tools/ML/model.h"
```
`Array2D.h` is from O2, `model.h` is from O2Physics

PWGHF/Core/HfMlResponse.h
```diff
@@ -0,0 +1,212 @@
...
+template <typename T1>
+uint8_t tagPID(const T1& track)
+{
+  uint8_t tag{0};
+  if (track.hasTPC() && track.hasTOF()) {
+    SETBIT(tag, PIDStatus::TpcAndTof);
+  } else if (track.hasTPC() && !track.hasTOF()) {
+    SETBIT(tag, PIDStatus::TpcOnly);
+  } else {
+    SETBIT(tag, PIDStatus::TofOnly);
```
vkucera, line 52: Shouldn't you check `!track.hasTPC() && track.hasTOF()` as a condition of setting `TofOnly`?

PWGHF/Core/HfMlResponse.h
```diff
@@ -0,0 +1,212 @@
...
+T1 getCombinedNSigma(const T1& nSigTpc, const T1& nSigTof, const T2& tagPID)
+{
+  if (TESTBIT(tagPID, PIDStatus::TpcAndTof)) {
+    return std::sqrt((nSigTpc * nSigTpc + nSigTof * nSigTof) / 2);
+  }
+  if (TESTBIT(tagPID, PIDStatus::TpcOnly)) {
+    return std::abs(nSigTpc);
+  }
+
+  return std::abs(nSigTof); // if (TESTBIT(tagPID, PIDStatus::TofOnly))
```
vkucera, line 72: Shouldn't there be a condition of `TofOnly` for returning `nSigTof`?

vkucera (changes requested):

PWGHF/Core/HfMlResponse.h
AlexBigO: Actually, if a track has no TPC and no TOF, it does not pass the PID selections performed before the ML ones. That's why I did not write it explicitely. But I can modify it to be more explicit.

PWGHF/Core/HfMlResponse.h
AlexBigO: As `tagPID` can only take 3 values, I only parsed the 2 first conditions. But again, this is the same philosophy as the comment above, I will make it more explicit.

PWGHF/Core/HfMlResponse.h
```diff
@@ -0,0 +1,216 @@
...
+#include <onnxruntime/core/session/experimental_onnxruntime_cxx_api.h>
+
+#include <map>
+#include <string>
+#include <vector>
+
+#include "CCDB/CcdbApi.h"
+#include "Tools/ML/model.h"
+
+#include "Framework/Array2D.h"
```
vkucera, line 29: CCDB and Framework are from O2, Tools is from O2Physics. Please fix the grouping.

PWGHF/Core/HfMlResponse.h
```diff
@@ -62,14 +64,16 @@ uint8_t tagPID(const T1& track)
...
-    return std::abs(nSigTpc);
+    combinedNSigma = std::sqrt((nSigTpc * nSigTpc + nSigTof * nSigTof) / 2);
+  } else if (TESTBIT(tagPID, PIDStatus::TpcOnly)) {
+    combinedNSigma = std::abs(nSigTpc);
+  } else if (TESTBIT(tagPID, PIDStatus::TofOnly)) {
+    combinedNSigma = std::abs(nSigTof);
   }
 
-  return std::abs(nSigTof); // if (TESTBIT(tagPID, PIDStatus::TofOnly))
+  return combinedNSigma;
```
vkucera: Why did you change the logic from returning the value right away to assigning it to a local variable?

vkucera (changes requested):

PWGHF/Core/HfMlResponse.h
AlexBigO: Because I get a Warning at compilation time if I only have `return` inside `if loops`.

vkucera: You can add `return 0.;` after the last `if`.

AlexBigO: Okay, thanks :)

vkucera (approved):

PR #2853, opened 2023-06-20
strangeness: TOF PID considering decay topology

ddobrigk:
* first commit of attempted test of TOF PID considering decay topology 
* further refinement definitely necessary: length calculation could benefit from analytical calculator developed for ALICE 3 OTF TOF, length as stored in AO2D 
* ...still, this PR will allow for some first tests of the K0 and especially Lambda use case


njacazio (approved):
Thanks David

PR #2859, opened 2023-06-21
PWGHF: Fix use of wrong abs

vkucera:
The default abs function may return int.
Using std::abs should always be safe for any number type.


vkucera:
@ddobrigk @fcolamar These are the only occurrences I found in PWGHF.

alibuild (approved):

PR #2862, opened 2023-06-21
PWGHF: Fix use of PI

vkucera:
Make proper use of constants `PIHalf` and `TwoPI` instead of dividing/multiplying `PI`.


alibuild (approved):

PR #2866, opened 2023-06-22
PWGHF: split configurable for generated and reconstructed rapidity cut

fgrosa:
In general, the cut on generated and reconstructed rapidity might be different for efficiency calculation. In particular, if `-1` is set (no cut) it has a much different (and maybe undesired) effect on reconstructed (limited by acceptance anyway) and generated (limited by the MC settings)

### DIFF
--- a/PWGHF/D2H/Tasks/taskB0Reduced.cxx
+++ b/PWGHF/D2H/Tasks/taskB0Reduced.cxx
@@ -33,7 +33,8 @@ using namespace o2::aod::hf_cand_b0; // from CandidateReconstructionTables.h
 /// B0 analysis task
 struct HfTaskB0Reduced {
   Configurable<int> selectionFlagB0{"selectionFlagB0", 1, "Selection Flag for B0"};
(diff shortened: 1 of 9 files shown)


PWGHF/D2H/Tasks/taskB0Reduced.cxx
```diff
@@ -169,7 +170,7 @@ struct HfTaskB0Reduced {
       if (!TESTBIT(candidate.hfflag(), hf_cand_b0::DecayType::B0ToDPi)) {
         continue;
       }
-      if (yCandMax >= 0. && std::abs(yB0(candidate)) > yCandMax) {
+      if (yCandGenMax >= 0. && std::abs(yB0(candidate)) > yCandGenMax) {
```
fcatalan92, line 173:
```suggestion
      if (yCandRecoMax >= 0. && std::abs(yB0(candidate)) > yCandRecoMax) {
```
This looks like a typo

fcatalan92 (changes requested):

fcatalan92 (approved):

PR #2873, opened 2023-06-22
PWGHF: Add D+ and Ds+ to MC efficiency task

fgrosa:
@jgrosseo @mfaggin I have added the possibility to run the MC efficiency task with D+ and Ds. Could you please have a look if the implementation is fine with you? Thanks a lot!

### DIFF
--- a/PWGHF/Tasks/taskMcEfficiency.cxx
+++ b/PWGHF/Tasks/taskMcEfficiency.cxx
@@ -513,12 +513,28 @@ struct HfTaskMcEfficiency {
         if (absPdg != pdgCode) { /// abs. value because only "kLambdaCPlus" is defined, not "kAntiLambdaCPlus"
           continue;
         }
+
+        std::array<int, 3> pdgDaughters;
+        if (pdgCode == pdg::kDPlus) {
+          pdgDaughters[0] = +kPiPlus;
+          pdgDaughters[1] = -kKPlus;
+          pdgDaughters[2] = +kPiPlus;
+        } else if (pdgCode == pdg::kDS) {
+          pdgDaughters[0] = +kKPlus;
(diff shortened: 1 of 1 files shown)


PWGHF/Tasks/taskMcEfficiency.cxx
```diff
@@ -177,12 +199,18 @@ struct HfTaskMcEfficiency {
...
+          } else if (pdgCode == pdg::kDS) {
+            massHypo1 = invMassDsToKKPi(candidate);
           }
           hCandidates->Fill(kHFStepTracked, pt, massHypo1, pdgCode, cpa, collisionMatched, origin);
         }
         if (isHypoMass2TrackStep) {
           if (pdgCode == pdg::kLambdaCPlus) {
             massHypo2 = invMassLcToPiKP(candidate);
+          } else if (pdgCode == pdg::kDS) {
+            massHypo1 = invMassDsToPiKK(candidate);
```
mfaggin, line 213: @fgrosa shouldn't it be `massHypo2` here?

fgrosa: It was a typo indeed, fixed, thanks!

mfaggin (commented):
hi @fgrosa from a quick look it seems really good, extending what we thought together. I just have 1 question, see the other comment.

Fine for me for the rest.
Best.
Mattia

vkucera:
@jgrosseo Can you please review this?

jgrosseo (approved):
This is fine for me. We should see if we find a better way for the additional template arguments (`bool hasDplus, bool hasDs, bool hasLc`) but this is not a showstopper...

vkucera (approved):

PR #2874, opened 2023-06-22
PWGHF: Move ALICE 3 code to a dedicated directory

vkucera:


alibuild (approved):

PR #2875, opened 2023-06-22
PWGHF: add candidate collIision Id to treeCreatorToXiPi

ZFederica:
- add candidate collIision Id to treeCreatorToXiPi for rejection factor studies

### DIFF
--- a/PWGHF/TableProducer/treeCreatorToXiPi.cxx
+++ b/PWGHF/TableProducer/treeCreatorToXiPi.cxx
@@ -32,6 +32,7 @@ namespace o2::aod
 namespace full
 {
 // from creator
+DECLARE_SOA_INDEX_COLUMN(Collision, collision);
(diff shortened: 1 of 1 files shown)


vkucera:
Hi @ZFederica , can you please elaborate on how the collision index is supposed to be used in the studies you mention?

ZFederica:
Hi @vkucera, if I had this information available offline I could play around with the selections on the saved variables, making them more or less tight, and checking how many different collision indexes I have in the end. This would allow me to trace back to how many events I would save with which selections.

vkucera:
I assume the fraction of collisions with more than one Omega_c is negligible so I would expect that whatever candidate selection you vary, the resulting change in the number of collisions will be equal to the change in the number of candidates. Am I missing something?

vkucera:
Yes, I understand, I meant the candidates. But OK, give it a try.

vkucera (approved):

PR #2905, opened 2023-06-27
remove unnecessary TGeo load

ddobrigk:
These instances of TGeoManager load seem unnecessary - @lupengzhong , can you please take a look? It seems that the material LUT is loaded right on top, and the `matCorr` is set to be material LUT and the TGeo is never used. Please note that not loading the TGeo saves about 250MB of memory.


jgrosseo (approved):

lupengzhong:
Hi,

Sorry for the late reply! I checked locally and the PR can work fine. Thanks a lot!

PR #2908, opened 2023-06-28
add D0 candidate type mask

Mingyu3360715:


vkucera:
Hi @Mingyu3360715 , first please fix the PR title, make it more descriptive and add more details about the code changes in the PR description, as the title does not give enough information about the several important changes included in this PR.
Second, since you change the content and the format of the output table, I would like to ask whether these changes have been discussed with other D0 analysers.

PR #2920, opened 2023-06-29
PWGHF: Remove unnecessary linked libraries and included headers

vkucera:
- Remove unnecessary linked libraries.
- Remove clearly unrelated header includes.
- Add a comment for each C++ header listing what is used.
- Use angle brackets for includes external to O2/O2Physics.


alibuild (approved):

PR #2923, opened 2023-06-29
Add metadata handling for TPCResponse CCDB objects

jezwilkinson:
Adding support for TPC parameter handler + response tasks to save/request specific pass number for multiple objects with same name.

New features:
- Task will now correctly print the used parameters after loading the object from CCDB (previously: would only report the parameters from the default object)
- Able to load specific object with different pass name for same path + timestamp (default behaviour unchanged: will load the latest uploaded object for the timestamp if "passname" configurable unspecified)

Plan to change default behaviour in future to automatically determine pass name, when pulling from AO2D metadata is available


njacazio:
Hi Jeremy, we are ready to drop the handler for TOF, thanks for the heads up!

jezwilkinson:
@ddobrigk Could you please take a quick look and approve if you think it's OK? The plan is to re-upload the CCDB objects using the new metadata scheme once this is merged so that the new functionality can be used soon.

ddobrigk (approved):
Thanks a lot @jezwilkinson , and sorry for not looking earlier! 
Looks good to me - but of course we'll have to check this once the information from the metadata is fully available as it should.

PR #2930, opened 2023-06-30
TrkSel: Autoset the track selection

njacazio:
@mfaggin @arossi81

### DIFF
--- a/Common/Core/TableHelper.h
+++ b/Common/Core/TableHelper.h
@@ -45,13 +45,15 @@ bool isTableRequiredInWorkflow(o2::framework::InitContext& initContext, const st
 /// @param flag int value of flag to set, only if initially set to -1. Initial values of 0 or 1 will be kept disregarding the table usage in the workflow.
 void enableFlagIfTableRequired(o2::framework::InitContext& initContext, const std::string& table, int& flag)
 {
+  if (flag > 0) {
+    flag = 1;
+    LOG(info) << "Table enabled: " + table;
+    return;
+  }
   if (isTableRequiredInWorkflow(initContext, table)) {
     if (flag < 0) {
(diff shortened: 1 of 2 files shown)


njacazio:
With this one does not have to enable or disable the FB table anymore as it gets filled if one of the tasks requires it in the workflow

mfaggin (dismissed):
Hi @njacazio that's for sure a very nice optimization.

I must admit that I do not get at all the difference between the case `-1` and `1` for the booleans. Looking at them here https://github.com/AliceO2Group/O2Physics/blob/85877abdc39a67f5d8f2b1a512ede2994c6d70f5/Common/Core/TableHelper.h#L52 they look to me exactly identical in their behaviour.

However, this implementation is for sure very useful, so I approve it.

njacazio:
@mfaggin -1 is for automatic setup, 1 is to force to have it produced by hand and 0 to force it not producing it

mfaggin:
> @mfaggin -1 is for automatic setup, 1 is to force to have it produced by hand and 0 to force it not producing it

I understand that this is want is meant, but in practice the 2 cases look identical to me.
This code here
https://github.com/AliceO2Group/O2Physics/blob/85877abdc39a67f5d8f2b1a512ede2994c6d70f5/Common/Core/TableHelper.h#L49C5-L55C6
seems to me that it simply does `flag = 1` in both cases, either if its original value is `-1` or `1`. Therefore, I do not get the practical difference between then

njacazio:
In case it is 1 it is enabled even if not used

njacazio:
@mfaggin you're right, they're the same, I must have had something else in mind, now should be as intended!

ddobrigk (approved):

PR #2934, opened 2023-06-30
PWGHF: Group O2 and O2Physics includes separately

vkucera:
This makes it easier to distinguish what is included from O2 and what from O2Physics.


alibuild (approved):

PR #2942, opened 2023-07-01
EventFiltering/PWGHF: Add condition for global tracks w/o DCA

fgrosa:


alibuild (approved):

PR #2951, opened 2023-07-02
Generator level for efficient not stored derived data

victor-gonzalez:
Also getting rid of the dependency on track-selection which caused crashes on the nightly tests


alibuild (approved):

PR #2953, opened 2023-07-03
PWGHF: change eta configurables and add PID method toXiPi tasks

ZFederica:
- modify eta configurables so that it's possible to cut differently on LF particle daughters and charm baryon bachelor
- add a different PID method as alternative to TrackSelectorPID

### DIFF
--- a/PWGHF/TableProducer/candidateSelectorToXiPi.cxx
+++ b/PWGHF/TableProducer/candidateSelectorToXiPi.cxx
@@ -10,7 +10,7 @@
 // or submit itself to any jurisdiction.
 
 /// \file candidateSelectorOmegac.cxx
-/// \brief Omegac → Xi Pi selection task
+/// \brief Xic and Omegac → Xi Pi selection task
 /// \author Federica Zanone <federica.zanone@cern.ch>, Heidelberg University & GSI
 
 #include "Framework/AnalysisTask.h"
@@ -50,7 +50,8 @@ struct HfCandidateSelectorToXiPi {
   Configurable<double> invMassOmegacMax{"invMassOmegacMax", 3.1, "Upper limit invariant mass spectrum charm baryon"};
 
   // kinematic selections
-  Configurable<double> etaTrackMax{"etaTrackMax", 0.8, "Max absolute value of eta"};
+  Configurable<double> etaTrackCharmBachMax{"etaTrackCharmBachMax", 0.8, "Max absolute value of eta for charm baryon bachelor"};
+  Configurable<double> etaTrackLFDauMax{"etaTrackLFDauMax", 1.0, "Max absolute value of eta for V0 and cascade daughters"};
   Configurable<double> ptPiFromCascMin{"ptPiFromCascMin", 0.15, "Min pT pi <- casc"};
   Configurable<double> ptPiFromOmeMin{"ptPiFromOmeMin", 0.2, "Min pT pi <- omegac"};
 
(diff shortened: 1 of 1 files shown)


vkucera:
Hi @ZFederica , can you please describe how is your PID method different from the one in `TrackSelectorPID` and why? Maybe it could be added in the class.

ZFederica:
Hi @vkucera, the reason why I implemented this is that, some time ago, Fabio suggested in an email to loosen the selections when using the PIDSelector, as the class seems to have some bugs. What I do is to check whether the PID information for a given track is stored (TPC and/or TOF nSigma), and, if so, compare the nSigma value for a given PID hypothesis with a configurable and consequently accept or reject the candidate. This looks more transparent to me compared to the PID class.

vkucera:
Which Fabio?
If there are bugs in the class, they should be reported and fixed instead of avoiding the use of the class by implementing custom selections.

ZFederica:
I was referring to Fabio Catalano (D2H convener). 
However, I use the same selections also in the trigger code: as I have a given CPU time budget for my analysis on hyperloop, it would be useful to me to do some trigger related tests with my own code, without running the full HF triggers, and this can be possible only if have this commits included (my trigger runs in the same process functions with the others, it's not possible to run it separately)
Could you please merge? Anyway, this tasks is not "centrally used" and this should not affect anyone else

fcatalan92:
Hi @ZFederica, what I suggested in the email were some configurations to mimic the PID selections used in Run 2, namely to use `getStatusTrackPIDTpcAndTof ` and to set the max pT ranges for the TPC and TOF PID to large values, not to avoid known bugs.
I agree with Vit that you should use the `TrackSelectorPID`, checking that it is working correctly for your use case and, eventually, reporting the bug/missing feature or fixing/extending it yourself. Finally, this task is central and meant for further reuse, e.g. for Xic0 studies. So, I would not diverge from the usual workflow.

ZFederica:
Hi @fcatalan92, sorry it was my misunderstanding. Since you suggested loosening the selections for the trackSelectorPID and mentioned that the framework is not yet fully debugged and validated, I assumed that there were problems with this specific class. And yes, I am also doing Xic0 studies myself. 
However, it would be useful to me to also this second PID check (as I mentioned before, for instance, it follows the same idea as the corresponding trigger selections). Since I did not replace the trackSelectorPID, but I have implemented this basic nSigma check as an alternative, would it be possible to include it? In general I need to run on large datasets, and this is only popssible on hyperloop.

fcatalan92:
Hi @ZFederica, I would prefer to stick to one PID strategy. 
Using this task to test your workflow for the HF trigger is not optimal, since you can still have many subtle differences between the two. Moreover, you can check if the PID is working accordingly to your expectations running on a small data sample locally.

ZFederica:
Hi @fcatalan92 @vkucera, I removed the alternative PID method, now it's only based on the trackSelectorPID

fcatalan92 (approved):

fcatalan92:
@ZFederica Thanks a lot for having followed our suggestions!
@vkucera  I think this PR is fine now.

vkucera:
> @ZFederica Thanks a lot for having followed our suggestions! @vkucera I think this PR is fine now.

Thanks @ZFederica for implementing the suggestions.
Thanks @fcatalan92 for help and checking the update.

vkucera (approved):

PR #2956, opened 2023-07-03
Mult QA: add process function for event studies at low mult

ddobrigk:
* adds a number of QA plots and better configurability to the multiplicity QA task
* allows for checking number of reconstructed PVs vs multiplicity in MC 
* allows for switching the check from equalized to non-equalized quantities


ddobrigk:
@ChiaraDeMartin95 @nepeivodaRS @njacazio this should allow us to do the main tests for the event loss / splitting corrections at low multiplicity also centrally, with the multiplicity calibration task

alibuild (approved):

PR #2978, opened 2023-07-05
First version of the kink analysis in AO2D

pganoti:
This include a part of code that does the actual kink finding and that will be later moved to a dedicated table creator


njacazio (dismissed):
Thank you Evi!

njacazio (approved):
I think we might miss the change to the cmake file but let's merge this first

PR #2985, opened 2023-07-06
[EMCAL-903] Updates of the full jet trigger task for hardware trigger…

mfasDa:
…ed mode

- Accept all EMCAL hardware trigger + readout trigger
- Add low threshold trigger class for jet trigger
- Restructure functions in order to switch between jet types
  + Extract gamma / jet trigger logics to dedicated functions
  + Template jet trigger function to be able to run on full/neutral jet collection
  + Extract process code to templated run function
  + Create process functions (and PROCESS_SWITCH) for full / neutral jet input
- Users must specify the jet type for the jet trigger via PROCESS_SWITCH, the jet type configurable is removed as the information is redundant

### DIFF
--- a/EventFiltering/PWGJE/fullJetFilter.cxx
+++ b/EventFiltering/PWGJE/fullJetFilter.cxx
@@ -12,6 +12,7 @@
 // Full Jet Filter
 // Author: Gijs van Weelden
 
+#include <array>
 #include <cmath>
 #include <string>
 #include <TMath.h>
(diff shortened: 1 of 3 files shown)


mfasDa:
@aalkin I would like to cross check whether the following works, which is used in the QA function: I have defined two process functions like:
``` c++
void processFullJets(soa::Join<aod::Collisions, aod::EvSels, aod::FullJetFilters>::iterator const& collision,
                       fullJetInfos const& jets,
                       aod::Tracks const& tracks,
                       selectedClusters const& clusters) {
}
```
Here I expect object matching. I want to forward the arguments to another function which is templated in the second argument (jet type)
``` c++
  template <typename JetCollection>
  void runQA(soa::Join<aod::Collisions, aod::EvSels, aod::FullJetFilters>::iterator const& collision,
             JetCollection const& jets,
             aod::Tracks const& tracks,
             selectedClusters const& clusters) {
   }
```
where I want to only make use of the object matched before. Those I want to pass even further to a templated function to process the actual jets, but I want to add non-framework arguments like
```c++
  template <typename JetCollection>
  std::tuple<std::vector<typename JetCollection::iterator>, std::vector<typename JetCollection::iterator>> fillJetQA(const JetCollection& jets, aod::Tracks const& tracks, selectedClusters const& clusters, const std::bitset<TriggerType_t::kNTriggers>& triggerstatus)
  {
  }
```
Does this work? It is needed because the input of the jet collection can come from two different tables which have the same structure, and I want the code to be able to process both, on user request. What I am wondering is if I pass framework arguments to other functions whether there happens a matching as well, which could eventually spoil the logics.

aalkin:
@mfasDa I do not see any immediate obstacle here, the arguments are fully formed when the `process` is invoked and they are constant references so they will stay unmodified, you can pass them to any other function calls within the process functions. And those functions can have arbitrary signatures, only the entry point process functions are parsed. 

Note that in the 3rd function you probably do not need a tuple - you can use `std::pair` or even an array, since you have two identical vector types there. Tuple is useful, but comes with a heavy compilation cost.

mpuccio (approved):

nzardosh (approved):

PR #2994, opened 2023-07-07
PWGEM/DQ: Include pdgInHistory into previous checkSignal, enable the …

feisenhu:
…use of both, adapt MCSignals

Change order of initialisation for MCProng. 
Move checkGenerationsInTime in front of new defined vectors:
"checkIfPDGInHistory" (vector of pdgCoded that should be checked)
"excludePDGInHistory" (vector of booleans checking if provided pdgCode should be excluded)

Include pdgInHistory into the existing version of checkSignal, so both can be used in addition.
Enables to check if Prong has a pdg as direct mother & has (not) a certain pdgCode somewhere else in the decay chain.

### DIFF
--- a/PWGDQ/Core/MCSignal.h
+++ b/PWGDQ/Core/MCSignal.h
@@ -130,80 +130,6 @@ bool MCSignal::CheckProng(int i, bool checkSources, const T& track)
   using P = typename T::parent_t;
   auto currentMCParticle = track;
 
-  // check if mother pdg is in history
-  if (fProngs[i].fPDGInHistory) {
-    std::vector<bool> pdgInHistory;
-    // check the PDG code of currentMCParticle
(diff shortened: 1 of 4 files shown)


PWGDQ/Core/MCSignal.h
```diff
@@ -320,6 +246,39 @@ bool MCSignal::CheckProng(int i, bool checkSources, const T& track)
     }
   }
 
+  if (fProngs[i].fPDGInHistory.size() == 0)
+    return true;
+  else { // check if mother pdg is in history
+    std::vector<int> pdgInHistory;
+    // cout << __LINE__  << " COUT LINE :: enter pdgInHistory " << endl;
```
iarsene, line 253: remove this commented line

PR #2999, opened 2023-07-07
[EventFiltering] Add possibility to apply event selection and turn off single femto channels

fgrosa:


alibuild (approved):

PR #3001, opened 2023-07-07
PWGJE: fixing bug with collision index stored with jets

nzardosh:
Fixing bug where collision table index was not properly stored in the jet table
Adding option to manually set the jet eta limits


alibuild (approved):

PR #3012, opened 2023-07-09
PWGEM/PhotonMeson: update on V0PhotonCut

dsekihat:
PWGEM/PhotonMeson: update on V0PhotonCut

### DIFF
--- a/PWGEM/PhotonMeson/Core/HistogramsLibrary.cxx
+++ b/PWGEM/PhotonMeson/Core/HistogramsLibrary.cxx
@@ -60,7 +60,7 @@ void o2::aod::emphotonhistograms::DefineHistograms(THashList* list, const char* 
     list->Add(new TH1F("hNclsITS", "number of ITS clusters", 8, -0.5, 7.5));
     list->Add(new TH1F("hChi2ITS", "chi2/number of ITS clusters", 360, 0, 36));
     list->Add(new TH2F("hXY", "X vs. Y;X;Y", 100, 0, 100, 100, -50, 50));
-    list->Add(new TH2F("hZX", "Z vs. X;Z;X", 200, -100, 100, 100, 0, 100));
+    list->Add(new TH2F("hZX", "Z vs. X;Z;X", 1000, -100, 100, 100, 0, 100));
     list->Add(new TH2F("hZY", "Z vs. Y;Z;Y", 200, -100, 100, 100, -50, 50));
     list->Add(new TH2F("hDCAxyEta", "DCAxy vs. #eta;#eta;DCA_{xy} (cm)", 400, -2, +2, 100, -50, 50));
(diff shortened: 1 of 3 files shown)


PWGEM/PhotonMeson/Core/HistogramsLibrary.cxx
```diff
@@ -60,7 +60,7 @@ void o2::aod::emphotonhistograms::DefineHistograms(THashList* list, const char*
     list->Add(new TH1F("hNclsITS", "number of ITS clusters", 8, -0.5, 7.5));
     list->Add(new TH1F("hChi2ITS", "chi2/number of ITS clusters", 360, 0, 36));
     list->Add(new TH2F("hXY", "X vs. Y;X;Y", 100, 0, 100, 100, -50, 50));
-    list->Add(new TH2F("hZX", "Z vs. X;Z;X", 200, -100, 100, 100, 0, 100));
+    list->Add(new TH2F("hZX", "Z vs. X;Z;X", 1000, -100, 100, 100, 0, 100));
     list->Add(new TH2F("hZY", "Z vs. Y;Z;Y", 200, -100, 100, 100, -50, 50));
```
mhemmer-cern, line 64: should the brining in X, Z and Y respectively not be the same over these three histograms?

PWGEM/PhotonMeson/Core/V0PhotonCut.h
```diff
@@ -138,26 +138,22 @@ class V0PhotonCut : public TNamed
...
         if (!CheckITSCuts(track)) {
           return false;
         }
-      } else if (isTPConly) {
+      } else if (track.hasTPC()) {
         if (!CheckTPCCuts(track)) {
           return false;
         }
+      } else { // remove ITS-TRD, ITS-TOF, ITS-TRD-TOF that are unrealistic tracks.
+        return false;
```
mhemmer-cern, line 156: The check down below 
```c++
else if (track.hasTPC()) {
         if (!CheckTPCCuts(track)) {
           return false;
         }
```
now can also include tracks with TRD and/or TOF, right?

dsekihat: TRD and TOF are not required at all. But, they are not "rejected".

mhemmer-cern: Yeah but previously you had an option for actual TPC only, so where TRD and TOF were rejected. Is the removal of this what we actually want?

dsekihat: TRDonly and TOFonly tracks are always rejected. In my previous commit, I noticed cuts were not applied to TPC-TRD-TOF, TPC-TOF, TPC-TRD. It is correct now.

mhemmer-cern (approved):

PR #3015, opened 2023-07-10
PWGCF: FemtoUniverse -- Adding structures for configurables

zchochul:


alibuild (approved):

PR #3018, opened 2023-07-10
PIDTPC: Add possibility to skip TPC only tracks

njacazio:


njacazio:
After discussion with @ddobrigk it does not make much sense for TOF. However for the TPC it might still play a role @jezwilkinson what do you think of this?

ddobrigk:
Thanks a lot, @njacazio ! If @jezwilkinson agrees, and if offline tests look ok, I would suggest we merge and create a core service wagon that skips tpc-only (and without NN evaluation of PID, which I guess is anyway the default?). It could save quite some CPU, I suppose...

njacazio:
That was my idea, I'm testing it

ddobrigk (approved):

njacazio:
Tested and ready to merge in agreement with what discussed with @jezwilkinson offline. We should just make sure that the wagon is correctly labelled

PR #3019, opened 2023-07-10
fix has_collision() protection on TPC response

jezwilkinson:
A "has_collision()" check was moved in https://github.com/AliceO2Group/O2Physics/pull/2923/ and stopped protecting the bc() call for tracks without associated collision. This commit should fix the issue

### DIFF
--- a/Common/TableProducer/PID/pidTPC.cxx
+++ b/Common/TableProducer/PID/pidTPC.cxx
@@ -291,16 +291,18 @@ struct tpcPid {
 
     for (auto const& trk : tracks) {
       // Loop on Tracks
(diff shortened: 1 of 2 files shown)


ddobrigk:
thanks for the quick action, @jezwilkinson - I stay on standby to approve quickly once you move it out of review, but please test to make sure the issue is gone. Thanks again!

jezwilkinson:
Looks like we should be good here, I was able to reproduce the original bug with Giovanni's AO2D and it appears to run smoothly after the patch
Apologies for the mess, the AOD chunks I used for testing apparently didn't have any tracks that fell into this bug..

ddobrigk (approved):
Thanks a lot, @jezwilkinson !

PR #3023, opened 2023-07-11
[EventFiltering] Improve selections for HF triggers that involve V0s and cascades

fgrosa:


alibuild (approved):

PR #3028, opened 2023-07-11
Fix data types table colls big.

mfaggin:


alibuild (approved):

PR #3036, opened 2023-07-11
PWGHF: Add tree creator for Xic->pkpi

phymanshu:
- Add tree creator `PWGHF/TableProducer/treeCreatorXicToPKPi.cxx`
- Modify `PWGHF/TableProducer/CMakeLists.txt` accordingly


vkucera (changes requested):
Hi @phymanshu , I understand this is a copy-paste of the Lc tree creator with the replacement `Lc` -> `Xic`. This of course produces big code duplication which should be avoided if possible. If you need this to be merged soon for your analysis, I don't have an issue with it but we should keep in mind that these two workflows should be merged into one rather soon, possibly with a configurable `Lc/Xic` switch.
For now, please update your code with the recent changes made in the Lc tree creator.

phymanshu:
The code is updated accordingly. 
Explanations:
- In the future, I have plans to update this code, which might be irrelevant to the `Lc` analysis. Therefore, I preferred to use the separate tree creator task for `Xic`. 
- `Lc` tree creator fills all the observables of my interest, therefore, I copied it from `Lc` tree creator.

PR #3050, opened 2023-07-12
Fix globBC index in run.

mfaggin:


alibuild (approved):

PR #3057, opened 2023-07-12
Add actual DCA setup

njacazio:
@mfaggin

### DIFF
--- a/Common/Core/TrackSelectionDefaults.h
+++ b/Common/Core/TrackSelectionDefaults.h
@@ -26,7 +26,7 @@ TrackSelection getGlobalTrackSelection();
 
 // Default track selection requiring a particular Run 3 ITS matching
 TrackSelection getGlobalTrackSelectionRun3ITSMatch(int matching,
-                                                   TrackSelection::GlobalTrackRun3DCAxyCut passFlag = TrackSelection::GlobalTrackRun3DCAxyCut::Default);
+                                                   int passFlag = TrackSelection::GlobalTrackRun3DCAxyCut::Default);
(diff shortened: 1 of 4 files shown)


njacazio:
@mfaggin I realized it had no effect before, what do you think?

Common/Core/TrackSelectionDefaults.h
```diff
@@ -26,7 +26,7 @@ TrackSelection getGlobalTrackSelection();
 
 // Default track selection requiring a particular Run 3 ITS matching
 TrackSelection getGlobalTrackSelectionRun3ITSMatch(int matching,
-                                                   TrackSelection::GlobalTrackRun3DCAxyCut passFlag = TrackSelection::GlobalTrackRun3DCAxyCut::Default);
+                                                   int passFlag = TrackSelection::GlobalTrackRun3DCAxyCut::Default);
```
mfaggin, line 29: did you cross check that this does not give any subtle error? In principle if it compiles it should be fine, but have you tried to run it?

mfaggin (commented):
Hi @njacazio , thanks for fixing it. Apparently the `dcaSetup.value` was missing when the selections were initialized. Just a minor comment, after the reply on it I will approve it.

Common/Core/TrackSelectionDefaults.h
njacazio: Hi, yes I'm running with this since some weeks now

mfaggin (approved):

ddobrigk (approved):
thanks!

PR #3060, opened 2023-07-13
PWGHF: add PID information to HfCandCasc table

DanielSamitz:
add TPC and TOF nSigmas into HfCandCascBase table, to be able to run candidateSelectorLcToK0sP and taskLcToK0sP without subscribing to track table

### DIFF
--- a/PWGHF/D2H/Tasks/taskLcToK0sP.cxx
+++ b/PWGHF/D2H/Tasks/taskLcToK0sP.cxx
@@ -240,7 +240,7 @@ struct HfTaskLcToK0sP {
   }
 
   void
-    process(soa::Filtered<soa::Join<aod::HfCandCascExt, aod::HfSelLcToK0sP>> const& candidates, aod::BigTracksPID const&)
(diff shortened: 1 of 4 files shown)


vkucera:
Hi @DanielSamitz , the candidate creators and selectors are currently producing tables that are supposed to be consumed on-the-fly, which means that the track table is always available. These tables are quite big already and take a significant fraction of the train memory consumption. Adding 12 more columns in that table to avoid subscription to tracks does not seem like an optimal solution.
Can you describe your motivation in more details so that we can try to find a better solution?

vkucera:
Hi @DanielSamitz , thanks for the details. The solution for the problem you mentioned with the `TrackSelectorPID` is going to be merged soon. I will make an related announcement when that is done.
I remember having a discussion when Mattia presented his studies with a similar setup but now I cannot remember the disadvantages of using the tree creator. Can you please remind me?

fcatalan92:
After last week's discussion at D2H, I think this PR can be closed. @DanielSamitz Do you agree?

PR #3061, opened 2023-07-13
Add has_mcParticle check to LUT maker

ddobrigk:
@njacazio this is definitely a minimal change that we need to add for code safety...  but on top of that, maybe for Run 3 LUT stuff, we should do a minimum crossed rows selection and so on?


njacazio (approved):
Thanks David!

PR #3072, opened 2023-07-14
LF: HMPID Split IU and non IU tracks

njacazio:
@alcaliva before approving this I let you confirm that it's ok.
I split IU and non IU tracks since you want to propagate them anyways.
For curiosity, @ddobrigk and @aalkin do you know why the compilation was doing fine and what was the behavior when asking for a track's pT?


PWGLF/Tasks/AntimatterAbsorptionHMPID.cxx
```diff
@@ -390,22 +390,50 @@ struct AntimatterAbsorptionHMPID {
...
+  using TrackCandidatesIU = soa::Join<aod::TracksIU, aod::TracksCovIU, aod::TracksExtra, aod::TracksDCA,
+                                      PidInfoTPC, PidInfoTOF,
+                                      aod::TrackSelection, aod::TrackSelectionExtension>;
+
+  // Propagated to PV tracks
+  using TrackCandidates = soa::Join<aod::TracksIU, aod::TracksCovIU, aod::TracksExtra, aod::TracksDCA,
+                                    PidInfoTPC, PidInfoTOF,
+                                    aod::TrackSelection, aod::TrackSelectionExtension>;
+
+  // using TrackCandidates = soa::Join<aod::Tracks, aod::TracksCov, aod::TracksIU, aod::TracksCovIU, aod::TracksExtra, aod::TracksDCA, aod::pidTPCLfFullPi, aod::pidTOFFullPi, aod::pidTPCLfFullKa, aod::pidTOFFullKa, aod::pidTPCLfFullPr, aod::pidTOFFullPr, aod::pidTPCLfFullDe, aod::pidTOFFullDe, aod::pidTPCLfFullTr, aod::pidTOFFullTr, aod::pidTPCLfFullHe, aod::pidTOFFullHe, aod::pidTPCLfFullAl, aod::pidTOFFullAl, aod::TrackSelection, aod::TrackSelectionExtension, aod::TOFSignal, aod::pidTOFmass, aod::pidTOFbeta>;
```
njacazio, line 416: @ddobrigk @aalkin here we subscribed to both IU and non IU tracks

alibuild (approved):

njacazio:
> @njacazio interesting, so the tables are essentially identical and of course joinable because they have the same size. The join would simply throw away duplicated column types. Then the column access happens by string name of the `arrow::Column` and it looks for the first one, because it is assumed to be unique within the table. In plain words, only the `pT` method from the first table in the join would survive, and it would return the value from the first column with name `fPt`. I'll think about how we can protect against such cases in the future.

Hi @aalkin thanks a lot, all clear!

PR #3073, opened 2023-07-14
[PWGLF] Adding check for kaons as primary vertex contributors

jloemker:
Hi @njacazio,

here is the small change that you suggested, let me know if there is anything else I should include for it.

Cheerio,
Johanna


njacazio (approved):
Thanks a lot @jloemker!! @skundu692 this is what we need!

PR #3088, opened 2023-07-17
PWGHF: separating general debug flag from PVrefit flag

fcolamar:
Introduced a specific flag for PV refit, since the related debug lines add lots of test to the stdout, preventing to run on trains (where it's often useful to have the histograms produced from the rest of the debug calls).
@mfaggin I hope I didn't forget any debug line.


mfaggin:
hi @fcolamar nothing against it! Green light for me!

PR #3089, opened 2023-07-17
O2Physics tasks use GRP, so pull in the library

davidrohr:
This was working so far only accidentally, since most tasks are pulling in the library via other dependencies, while other tasks were ok with only the header present, which is present in any case from the O2 installation.

### DIFF
--- a/Common/Core/CMakeLists.txt
+++ b/Common/Core/CMakeLists.txt
@@ -15,7 +15,7 @@ o2physics_add_library(AnalysisCore
                        PID/ParamBase.cxx
                        CollisionAssociation.cxx
                        TrackSelectionDefaults.cxx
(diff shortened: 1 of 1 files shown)


jgrosseo:
Thanks! This should of course be merged. I just wonder how large the library and if it pulls in other libraries and if in the long-term the formats used by analysis could be in a smaller one. Adding @ddobrigk

davidrohr:
`DataFormatsParameters` is super small, it basically contains only GRP data formats. And it was linked to all analysis library but one already via dependency chains.
Actually I wonder if there are not other similar data format libraries missing in O2Physics, where at the moment the headers are available, but compilation will fail once a non-inline function is used. Anyway, can be fixed step by step if a problem appears again.

PR #3102, opened 2023-07-18
PWGCF: FemtoUniverse - adding phi to the producer and creating a track-phi task

zchochul:
* Adding the option to store Phi mesons in the FemtoAOD.root data for the FemtoUniverse. 
* Adding the track-phi task for FemtoUniverse.

In the next PR:  
* enable storage of Phi will be in the process switch (because not everyone will need them, so we don't want to save them),
* check selection for the tracks, now only PID is working, I would like to use setSelection (like in the FemtoUniverse).


alibuild (approved):

PR #3111, opened 2023-07-19
TOF for PID configurable

alcaliva:


njacazio:
I'm wondering if we should move this to the QC part

njacazio (approved):
If, let's do this after PR is merged

alcaliva:
> 

Hi Nicolò,
indeed, this is a good idea. I suggest that we keep it in the LF folder for testing. Once we are happy with the setting and it runs full steam, we can move it to QC. 
What do you think?

njacazio:
Nono I meant, LF QC https://github.com/AliceO2Group/O2Physics/tree/master/PWGLF/Tasks/QC !

alcaliva:
OK. Yes, at the next round, I move it

njacazio:
Top!

Issue #2178, opened 2023-03-09
Core: Replace PDG features of `RecoDecay` with `O2DatabasePDG`


vkucera:
JIRA https://alice.its.cern.ch/jira/browse/O2-3782

vkucera:
PRs:
- [x] https://github.com/AliceO2Group/AliceO2/pull/11409
- [x] https://github.com/root-project/root/pull/12968
- [x] https://github.com/AliceO2Group/AliceO2/pull/11557
- [x] https://github.com/AliceO2Group/O2Physics/pull/3104
- [x] https://github.com/AliceO2Group/AliceO2/pull/11668
- [x] https://github.com/AliceO2Group/O2Physics/pull/3105
- [x] Announce deprecation of `getMassPDG` on Mattermost
- [x] https://github.com/AliceO2Group/O2Physics/pull/3199
- [ ] Migrate code:
  - [ ] PWGHF
    - [ ] https://github.com/AliceO2Group/O2Physics/pull/3115
    - [ ] Make a helper class for header functions that use `getMassPDG`
    - [ ] Replace `getMassPDG` calls with `Mass` calls (using the helper for header functions).
- [ ] Assess CPU consumption of PDG database calls. Implement a cache if needed.
- [ ] Remove `getMassPDG`.

vkucera:
Based on Hyperloop wagon tests, replacing `getMassPDG` with `constexpr` masses saves 50-60 MB per device in memory.

vkucera:
Current situation:
* Masses in `O2/CommonConstants/PhysicsConstants.h` are different from the ones provided by the `O2DatabasePDG` service.
* `O2P/PWGHF/Core/PDG.h` + the update script provides constants and PDG enums for particles used in the analyses, which have names consistent with the naming scheme in `ROOT/PDG_t` and mass values equal to those in `O2DatabasePDG`.
* We need the enums, the constants and the service and we need to keep them synced.

Steps for unifying PDG masses:
* Copy `O2P/PWGHF/Core/PDG.h` (+ script) into `O2/CommonConstants` while renaming the namespace "`o2::analysis::pdg`" -> "`o2::constants::physics`".
* Include (with "`#include`") `O2/CommonConstants/PDG.h` in `O2/CommonConstants/PhysicsConstants.h`.
[...]

PR #3115, opened 2023-07-19
PWGHF: Replace `RecoDecay::getMassPDG`

vkucera:
Removes dependency on `RecoDecay::getMassPDG` in all PWGHF code and in code that includes PWGHF data model headers.
Requires https://github.com/AliceO2Group/AliceO2/pull/11668

- Adds a new `HfHelper` class in `PWGHF/Core/HfHelper.h`.
  - All helper functions in the data model headers that used to call `RecoDecay::getMassPDG` have been replaced by equivalent member methods of the helper class.
  - A new struct member `HfHelper hfHelper` has been added in tasks where needed.
- Adds a new header `PWGHF/Core/PDG.h` with additional named PDG codes (previously in `SelectorCuts.h`) and declared particle mass constants.
  - Body of the header is generated by the script `PWGHF/Core/make_pdg_header.py` which takes care of getting the masses from `O2DatabasePDG` and correcting them.
  - Masses declared in `o2::constants::physics` are ignored as they only provide `float` precision (and even less in some cases) and are not linked to the PDG database in any way, creating a possible mismatch of values.
  - In case other PWGs would use it (better in my opinion), it could be moved to `Core`.
- Adds a new header `PWGHF/Utils/utilsAnalysis.h` for analysis utilities.
  - The `findBin` function has been moved there from `SelectorCuts.h` to allow its usage in workflows that do not need the default selector configurables.
 
Rationale for not using the `O2DatabasePDG` service with the helper class:

- The service cannot be fully contained within the helper class. The service must be a separate struct member of each task and the helper class can only hold a pointer to it. Therefore having the service linked with the helper class requires an additional header include, an additional struct member and an initialisation of the pointer in the `init` function. This makes it even impossible to use in some tasks, such as `PWGCF/FemtoWorld/TableProducer/femtoWorldProducerTask.cxx` which has so many members that it cannot take any more.
- In most cases, the PDG of the particle is known at compile time so there is no benefit getting the mass during run time.
- For cases where the PDG is not known at compile time (e.g. loops over `McParticles`), the service can be used directly.
- Declared mass constants make it possible to move mass-dependent helper functions out from the helper class into dynamic columns of candidate tables, making the class unnecessary in some tasks (which would be the next step).
- Compile-time constants should also in principle reduce any possible overhead in memory or CPU time from loading the service, initialising the PDG database and retrieving the database objects.
- The header can be generated with the script any time to update mass values if needed, which provides full control of changes of the masses we use in the analysis, as opposed to getting updates from ROOT silently and blindly.


vkucera:
- Validated output against upstream/master for D0, D+, Ds, Lc, Xic tasks.
- No significant differences in memory and CPU profiles. The memory consumption actually looks even slightly lower.

memory-master
:arrow_up: upstream/master

memory-pdg
:arrow_up: This PR

aalkin (dismissed):

aalkin (approved):

iarsene (approved):

nzardosh (approved):

saganatt (approved):

alibuild (approved):

ddobrigk:
Hi everyone, 

First of all, thank you very much to @vkucera for his extensive work on improving our use of memory! This is really much needed for us to streamline the analysis. As a next step, and once we get rid of all `getMassPDG`calls, I would like to suggest that we merge the mass definitions and common use constants to the main file that already exists in AliceO2 for that purpose - `CommonConstants/PhysicsConstants.h` - that is already used by other tasks in O2Physics exactly to avoid depending on TDatabasePDG calls. We can also make any adjustments that may be necessary to that PhysicsConstants file, such as adjusting PDG masses to higher precision, and applying automatic update scripts to that file instead of to a PWG-specific file. Unifying this into a single, non-PWG-dependent file will avoid multiple points for the same information, and putting this info in O2 instead of O2Physics is important because this information is also used extensively in simulation. 

A concern in this regard is O2 merge speed, and that's a reasonable concern. I promise I will, as much as I can, push for any updates to be approved and merged as quickly as possible, and IMHO we should still get releases whenever we require it (such as if there's an important update to these constants, though I think (hope?) maybe this is not that common). 

Thank you very much! 
Best Regards,
David 

(P.S.: @vkucera we already discussed this a bit but I thought of posting a message here just to summarize. I hope that's OK with you :-D and thanks again!)

vkucera:
> Hi everyone,
> 
> First of all, thank you very much to @vkucera for his extensive work on improving our use of memory! This is really much needed for us to streamline the analysis. As a next step, and once we get rid of all `getMassPDG`calls, I would like to suggest that we merge the mass definitions and common use constants to the main file that already exists in AliceO2 for that purpose - `CommonConstants/PhysicsConstants.h` - that is already used by other tasks in O2Physics exactly to avoid depending on TDatabasePDG calls. We can also make any adjustments that may be necessary to that PhysicsConstants file, such as adjusting PDG masses to higher precision, and applying automatic update scripts to that file instead of to a PWG-specific file. Unifying this into a single, non-PWG-dependent file will avoid multiple points for the same information, and putting this info in O2 instead of O2Physics is important because this information is also used extensively in simulation.
> 
> A concern in this regard is O2 merge speed, and that's a reasonable concern. I promise I will, as much as I can, push for any updates to be approved and merged as quickly as possible, and IMHO we should still get releases whenever we require it (such as if there's an important update to these constants, though I think (hope?) maybe this is not that common).
> 
> Thank you very much! Best Regards, David
> 
> (P.S.: @vkucera we already discussed this a bit but I thought of posting a message here just to summarize. I hope that's OK with you :-D and thanks again!)

Hi @ddobrigk , thanks for the summary, I fully agree with you. As discussed in person, this is just one PR out of many on this issue, so it might be even better if you can add the summary as a comment in the related [GitHub issue](https://github.com/AliceO2Group/O2Physics/issues/2178) where I am collecting all the PRs and tracking the progress of this effort. We can continue the discussion there.

PR #3130, opened 2023-07-20
Update LFResonanceTables.h

yashpatley:
Added Spherocity Column in ResoCollision


BongHwi:
Hello @yashpatley,
I think the old `Sphericity` was the typo, could you update the name of this variable instead of adding a new variable?

PR #3133, opened 2023-07-20
EventFiltering: Fix bug in usage of recalibrated PID for Xi+bachelor trigger

fgrosa:


alibuild (approved):

PR #3135, opened 2023-07-20
Update the default parameters

BongHwi:
And apply the DCA cuts


alibuild (approved):

PR #3141, opened 2023-07-21
PWGDQ：Add definition for HE and CS frames

zjxiongOvO:
As I currently do not know how to obtain Beam information from CCDB, I will define the Beam information as a pp collision at 13.6 TeV in the initial version.

### DIFF
--- a/PWGDQ/Core/VarManager.h
+++ b/PWGDQ/Core/VarManager.h
@@ -344,6 +344,9 @@ class VarManager : public TObject
     kVertexingProcCode,
     kVertexingChi2PCA,
     kCosThetaHE,
+    kCosThetaCS,
+    kPhiHE,
+    kPhiCS,
     kPsiPair,
     kDeltaPhiPair,
     kQuadDCAabsXY,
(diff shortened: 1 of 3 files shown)


PWGDQ/Core/VarManager.h
```diff
@@ -1253,14 +1256,44 @@ void VarManager::FillPair(T1 const& t1, T2 const& t2, float* values)
     values[kPsiPair] = (t1.sign() > 0) ? TMath::ASin((v1.Theta() - v2.Theta()) / xipair) : TMath::ASin((v2.Theta() - v1.Theta()) / xipair);
   }
 
-  // CosTheta Helicity calculation
+  // set as default values
+  // TO DO: get the correct values from CCDB
+  double CenterOfMassEnergy = 13600.;                                                                       // GeV
```
iarsene, line 1261: Please add a static data member to the VarManager which will hold the beam energy, instead of hardcoding the value here. You should also add a setter function, so that the user can configure this value. Otherwise this would not work for other colliding energies. In the future, we should use the metadata/CCDB to get this value.

iarsene (approved):

PR #3142, opened 2023-07-21
PWGCF: Update FemtoDream

ariedel-cern:
This PR contains:
 - adding support for MC processing of Lambda particles
 - address comments from previous PR https://github.com/AliceO2Group/O2Physics/pull/3118
 - typo fixes

### DIFF
--- a/PWGCF/FemtoDream/FemtoDreamParticleHisto.h
+++ b/PWGCF/FemtoDream/FemtoDreamParticleHisto.h
@@ -69,9 +69,9 @@ class FemtoDreamParticleHisto
 
     std::string folderSuffix = static_cast<std::string>(o2::aod::femtodreamMCparticle::MCTypeName[mc]).c_str();
 
-    mHistogramRegistry->add((folderName + folderSuffix + "/hMometum").c_str(), "; #it{p} (GeV/#it{c})", kTH1F, {tempFitVarMomentumAxis});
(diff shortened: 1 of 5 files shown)


PWGCF/FemtoDream/FemtoDreamParticleHisto.h
```diff
@@ -223,28 +226,27 @@ class FemtoDreamParticleHisto
...
-        LOG(warn) << "Mometum at TPC inner wall is only available if the Producer Task generates debug information.";
+      case aod::femtodreamparticle::kPtpc:
+        LOG(warn) << "Momentum at TPC inner wall is only available if the Producer Task generates debug information.";
         momentum = part.tpcInnerParam();
         break;
       default:
-        LOG(warn) << "MomentumType " << momentum << " no implemented. Use pT of the Track for debug Histograms.";
+        LOG(warn) << "MomentumType " << MomentumType << " no implemented. Use pT of the Track for debug Histograms.";
         momentum = part.pt();
     }
```
victor-gonzalez, line 247: The idea of having an `enum` type is that you instead of passing an`int` you pass a variable that has the `enum` type (I'm surprised the compiler allowed you to hide the `enum` type  using the same name for a variable) in that way you have, at design level, full control of what the allowed values.
I approve it for the time being but change it in the next iteration

As it is now it is not only misleading but also dangerous in the medium term

victor-gonzalez (approved):
Please, consider my comment for the next iteration

PR #3143, opened 2023-07-21
PWGHF: candidateCreatorCascade: fix McGen

DanielSamitz:
Bug fix for finding generated $\Lambda_c^+\to pK^0_S$ and $\Lambda_c^- \to \bar{p}K^0_S$.  Currently $\Lambda_c^-$ are missed because the flag `acceptAntiParticles` in `RecoDecay::isMatchedMCGen` does not work for the $K^0_S$ (does not have an antiparticle, so the PDG code -310 that `RecoDecay::isMatchedMCGen` is looking for in that case does not exist, such that no $\Lambda_c^-$ is ever accepted)

### DIFF
--- a/PWGHF/TableProducer/candidateCreatorCascade.cxx
+++ b/PWGHF/TableProducer/candidateCreatorCascade.cxx
@@ -266,7 +266,7 @@ struct HfCandidateCreatorCascadeMc {
     int indexRec = -1;
     std::vector<int> arrDaughLcIndex;
     std::array<int, 3> arrDaughLcPDG;
-    std::array<int, 3> arrDaughLcPDGRef = {2212, 211, -211};
+    std::array<int, 3> arrDaughLcPDGRef = {+kProton, +kPiPlus, -kPiPlus};
 
     // Match reconstructed candidates.
     rowCandidateCasc->bindExternalIndices(&tracks);
--- a/PWGHF/TableProducer/candidateCreatorCascade.cxx
+++ b/PWGHF/TableProducer/candidateCreatorCascade.cxx
@@ -295,7 +295,7 @@ struct HfCandidateCreatorCascadeMc {
       MY_DEBUG_MSG(isK0SfromLc, LOG(info) << "correct K0S in the Lc daughters: posTrack --> " << indexV0DaughPos << ", negTrack --> " << indexV0DaughNeg);
 
       // if (isLc) {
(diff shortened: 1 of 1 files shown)


PWGHF/TableProducer/candidateCreatorCascade.cxx
```diff
@@ -330,7 +334,7 @@ struct HfCandidateCreatorCascadeMc {
             auto daughI = particlesMC.rawIteratorAt(arrDaughLcIndex[iProng]);
             arrDaughLcPDG[iProng] = daughI.pdgCode();
           }
-          if (!(arrDaughLcPDG[0] == arrDaughLcPDGRef[0] && arrDaughLcPDG[1] == arrDaughLcPDGRef[1] && arrDaughLcPDG[2] == arrDaughLcPDGRef[2])) { // this should be the condition, first bach, then v0
+          if (!(arrDaughLcPDG[0] == sign * arrDaughLcPDGRef[0] && arrDaughLcPDG[1] == arrDaughLcPDGRef[1] && arrDaughLcPDG[2] == arrDaughLcPDGRef[2])) { // this should be the condition, first bach, then v0
```
vkucera, line 337: Do you not need `sign` for all daughters?

vkucera (commented):
Hi @DanielSamitz , thanks for spotting this bug. Indeed, there is no entry for `-kK0Short` in the PDG database.

vkucera:
Also, please the hardcoded PDG codes in `arrDaughLcPDGRef`.

PWGHF/TableProducer/candidateCreatorCascade.cxx
DanielSamitz: No, because the $\pi^+$ and the $\pi^-$ do not swap place. They always come from a decay of a $K^0_S\to \pi^++\pi^-$.
The decay chains are
$L_c^+\to p+K^0 \to p + K^0_S \to p + \pi^+ + \pi^-$
$L_c^-\to \bar{p} + \bar{K^0} \to \bar{p} + K^0_S \to \bar{p} +\pi^++\pi^-$

DanielSamitz:
> Also, please the hardcoded PDG codes in arrDaughLcPDGRef.

Do you mean
`{2212, 211, -211}` --> `{kProton, kPiPlus, -kPiPlus}`
?

PWGHF/TableProducer/candidateCreatorCascade.cxx
vkucera: OK, if the positive pion always comes first, then it should be fine. Thanks

vkucera (approved):

PR #3145, opened 2023-07-21
EventFiltering: fix soft pion selection for D* in Ds1+ decays

fgrosa:


alibuild (approved):

PR #3150, opened 2023-07-22
Add a function to set collision system information

zjxiongOvO:


PWGDQ/Core/VarManager.cxx
```diff
@@ -24,6 +24,8 @@ float VarManager::fgValues[VarManager::kNVars] = {0.0f};
 std::map<int, int> VarManager::fgRunMap;
 TString VarManager::fgRunStr = "";
 std::vector<int> VarManager::fgRunList = {0};
+float VarManager::CenterOfMassEnergy = 13600;         // GeV
+float VarManager::MassofCollidingParticle = 9.382720; // GeV
```
iarsene, line 28: I would like to use naming conventions for the class data members, so please change these two names such that they start with "fg", similar to the other ones. The "f" means it is a data member, and the "g" means its static. These are rather old conventions, but work very well when reading the code.

PWGDQ/Core/VarManager.cxx
```diff
@@ -168,6 +170,23 @@ float VarManager::GetRunIndex(double Runnumber)
   return index;
 }
 
+//__________________________________________________________________
+void VarManager::SetCollisionSystem(TString system, float Energy)
```
iarsene, line 174: change "Energy" to "energy"

PWGDQ/Core/VarManager.h
```diff
@@ -577,9 +580,11 @@ class VarManager : public TObject
...
   static void SetVariableDependencies(); // toggle those variables on which other used variables might depend
 
-  static std::map<int, int> fgRunMap; // map of runs to be used in histogram axes
-  static TString fgRunStr;            // semi-colon separated list of runs, to be used for histogram axis labels
-  static std::vector<int> fgRunList;  // vector of runs, to be used for histogram axis
+  static std::map<int, int> fgRunMap;   // map of runs to be used in histogram axes
+  static TString fgRunStr;              // semi-colon separated list of runs, to be used for histogram axis labels
+  static std::vector<int> fgRunList;    // vector of runs, to be used for histogram axis
+  static float CenterOfMassEnergy;      // collision energy
+  static float MassofCollidingParticle; // mass of the colliding particle
```
iarsene, line 587: change names as commented also in VarManager.cxx

iarsene (approved):

PR #3160, opened 2023-07-24
PWGCF: FemtoUniverse - adding modular structure to the producer

zchochul:
Added features:
* every type of analysis has now its own process,
* more modular code.

Planned for the future:
* development on the selection part (figuring out the most efficient and easy from the point of view of a user solution for the particle cuts).


alibuild (approved):

PR #3165, opened 2023-07-24
TPC PID: add compatibility with "skipTPCOnly" option for TPC PID NN application

jezwilkinson:
refinements to TPC and TPCFull table producers:
- Adds ability to skip TPC-only tracks in NN application
- Skip tracks with missing TPC in NN application (and fix to fill table with -999 for invalid tracks as expected)
- minor bugfix in accessing NCl normalisation for NN
- Soft fallback with warning in case specific pass requested but not available (will be used with metadata access)


jezwilkinson:
@ddobrigk @jgrosseo This rework makes it possible to use the skipTPCOnly option also with the NN post-corrections (and fixes a minor bug in how -999s were handled in general by the network), should make the NN application a fair bit more efficient when that option is set.

ddobrigk (approved):
Thanks!!

PR #3167, opened 2023-07-24
LF: Change Sp datamodel

njacazio:
@jloemker could you also have a look pls?


alibuild (approved):

PR #3168, opened 2023-07-25
PWGCF: FemtoUniverse -- Global track cuts applied in the producer

zchochul:
This PR adds:
* global tracks cuts for the tracks in the producer,
* minimized amount of lines of code.


alibuild (approved):

PR #3177, opened 2023-07-25
PWGHF: Improve tree creator for D0 mesons and fix table names of tree creators

fgrosa:
This PR includes:
- improvements in the D0 tree creator including the possibility to store a "lite" tree for ML trainings (similar to the D+ one)
- fix of tree creator table names (the previous ones should not work with the AOD merger)

### DIFF
--- a/PWGHF/TableProducer/treeCreatorD0ToKPi.cxx
+++ b/PWGHF/TableProducer/treeCreatorD0ToKPi.cxx
@@ -25,6 +25,7 @@
 
 using namespace o2;
 using namespace o2::framework;
+using namespace o2::framework::expressions;
 using namespace o2::aod::hf_cand_2prong;
 
 namespace o2::aod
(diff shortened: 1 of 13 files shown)


PWGHF/TableProducer/treeCreatorD0ToKPi.cxx
```diff
@@ -61,6 +62,7 @@ DECLARE_SOA_COLUMN(DecayLengthNormalised, decayLengthNormalised, float);
 DECLARE_SOA_COLUMN(DecayLengthXYNormalised, decayLengthXYNormalised, float);
 DECLARE_SOA_COLUMN(Cpa, cpa, float);
 DECLARE_SOA_COLUMN(CpaXY, cpaXY, float);
+DECLARE_SOA_COLUMN(MaxNormalisedDeltaIP, maxNormalisedDeltaIP, float);
```
vkucera, line 65: Do you not want to store this in the full table?

vkucera (changes requested):
Hi @fgrosa , thanks. Can you please also fix the table names in the tree creators in `PWGHF/ALICE3/TableProducer`?

PWGHF/TableProducer/treeCreatorD0ToKPi.cxx
fgrosa: I added it in the "lite" version because I would like to use it, but I was unsure whether this was the case also for those who use the "full" version. However I added it now (in the end it is only one more variable indeed)

fgrosa:
> Hi @fgrosa , thanks. Can you please also fix the table names in the tree creators in `PWGHF/ALICE3/TableProducer`?

Hi @vkucera, thanks for the comments! I did it and I also noticed that two other tree creators in `PWGHF/TableProducer` would be more suited for the `ALICE3` folder, namely the Lb one (which explicitly uses ALICE3 PID tables) and the Xicc one (which will not be accessible for Run3-4). Do you know why they were not moved? Would you like me to do it in this PR?

vkucera (approved):

PR #3180, opened 2023-07-25
Fix bug on the track loop

BongHwi:
and modify the PID selection


njacazio:
What was the culprit?

njacazio (approved):

BongHwi:
I found that the number of tracks and combinations are highly affected with the current TPC, TOF selections (probably due to the array method applied in Kaon).

alibuild (approved):

njacazio:
Weird

PR #3182, opened 2023-07-25
Hmp table converter

gvolpe79:


Common/TableProducer/hmpConverter.cxx
```diff
@@ -0,0 +1,63 @@
...
+      auto hmpidYTrack = -999.; // dummy
+      auto hmpidXMip = -999.;   // dummy
+      auto hmpidYMip = -999.;   // dummy
+      auto hmpidNPhotons = hmpData.hmpidNPhotons();
+      auto hmpidQMip = hmpData.hmpidQMip();
+      auto hmpidClusSize = -999; // dummy
+      auto hmpidMom = -999;      // dummy
+      auto hmpidPhotsCharge = phots;
+
+      HMPID_001(track,
```
ddobrigk, line 43: Thanks a lot! However, here I think you need the track *index*, not the track; you should probably be using .trackId() directly and in fact I do not think you need the tracks subscription at all for this converter. I guess as is now, with track() instead of trackId(), this converter will probably even crash... Not sure. Did you try to run it?

gvolpe79: so instead of track = hmpData.track(), you suggest to do trackid = hmpData.trackId()?
I just compiled, I have not run it.

ddobrigk: Yes exactly, and then when filling the table provide that trackId - that should be enough I think! 

And perhaps it would be good to try and run it. It won't do anything particularly useful because it will generate the table in version 001 that won't be used for now but it would be good to check that it does not crash :-D

gvolpe79: to run it I should have aod in the 000 format I guess, and then how the converter is used?

ddobrigk: Well actually I think format 000 is what we currently store everywhere so any existing AO2D should work fine (unless I am overlooking something)... So basically you can just run your converter right away - and if you remove the tracks table subscription I think you won't have any dependencies, even, so it should be straightforward to test!

gvolpe79: David the 000 format has never used. There was just an HMPID table in AnalysisDataModel.h as place holder (not implemented by me), when I started to implement the AO2D for the HMPID I modified that place holder table and then Jan raise the point about the backward compatibility, so I implemented versions 000 and 001, but at the moment AO2D for the HMPID doesn't exist!

ddobrigk (approved):

PR #3187, opened 2023-07-26
[EventFiltering] Avoid to use TPC at high pT for proton femto and hardcode masses

fgrosa:


alibuild (approved):

PR #3105, opened 2023-07-18
Add the `O2DatabasePDG::Mass` function in the PDG tutorial

vkucera:
Requires https://github.com/AliceO2Group/AliceO2/pull/11668

PR #3199, opened 2023-07-27
RecoDecay: Add message about deprecating getMassPDG

vkucera:
Requires #3105

### DIFF
--- a/Common/Core/RecoDecay.h
+++ b/Common/Core/RecoDecay.h
@@ -535,9 +535,12 @@ class RecoDecay
   /// \return particle mass
   static double getMassPDG(int pdg)
   {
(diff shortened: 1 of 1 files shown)


vkucera:
@ddobrigk What do you think?

aalkin:
There is a conventional method to do it (https://en.cppreference.com/w/cpp/language/attributes/deprecated), it produces warnings during compilation.

vkucera:
> There is a conventional method to do it (https://en.cppreference.com/w/cpp/language/attributes/deprecated), it produces warnings during compilation.

Hi @aalkin , thanks, I was not aware of that. I can add it. I would also keep the error though, because I suspect that most people do not read (or care about) the compilation warnings as long as it compiles and the runtime error makes this migration reminder more visible.

aalkin:
I have to agree :)

PR #3203, opened 2023-07-27
Remove CFCore dependency

njacazio:


saganatt:
@Eloviyo seems CF Core is not used in your task and it causes compilation problems for some. Do you agree on the removal of this dependency?

Eloviyo:
Hi, yes it is okay to remove it then.

saganatt (approved):

PR #3208, opened 2023-07-28
Otf pid

NNicassio99:
Implementation of ALICE 3 on-the-fly RICH PID, integration of tracking uncertainty in TOF PID


ALICE3/TableProducer/onTheFlyRichPid.cxx
```diff
@@ -0,0 +1,621 @@
...
+    o2::math_utils::sincos(particle.phi(), s, c);
+    o2::math_utils::rotateZInv(particle.vx(), particle.vy(), x, params[0], s, c);
+    params[1] = particle.vz();
+    params[2] = 0.; // since alpha = phi
+    auto theta = 2. * std::atan(std::exp(-particle.eta()));
+    params[3] = 1. / std::tan(theta);
+    params[4] = charge / particle.pt();
+
+    // Initialize TrackParCov in-place
+    new (&o2track)(o2::track::TrackParCov)(x, particle.phi(), params, covm);
```
aalkin, line 290: If you are creating a new TrackParCov anyway, why not just return it by value? It would be considerably faster.

ALICE3/Core/DelphesO2TrackSmearer.cxx
```diff
@@ -224,13 +224,13 @@ bool TrackSmearer::smearTrack(O2Track& o2track, int pdg, float nch)
 }
 
 /*****************************************************************/
-// relative uncertainty on pt
+// absolute uncertainty on pt
```
njacazio, line 227: Hi @NNicassio99 is this intended, it would break also the other users

ALICE3/TableProducer/onTheFlyRichPid.cxx
aalkin: The codebase is full of code that was not reviewed properly, there is always something that can be improved. Current implementation is, essentially, doing what the copy constructor would do anyway. The better implementation would be a function that simply returns TrackParCov by value to benefit from the return value optimization.

ALICE3/Core/DelphesO2TrackSmearer.cxx
```diff
@@ -229,8 +229,8 @@ double TrackSmearer::getPtRes(int pdg, float nch, float eta, float pt)
 {
   float dummy = 0.0f;
   auto lutEntry = getLUTEntry(pdg, nch, 0., eta, pt, dummy);
-  auto val = sqrt(lutEntry->covm[14]) * lutEntry->pt;
-  return val;
```
njacazio, line 233: Sorry but also this, I don't understand, why these changes?

ALICE3/Core/DelphesO2TrackSmearer.cxx
hscheid: Hi @njacazio thanks for also taking a look at the code.
When I implemented this, it was thought as a starting point for the whole pid in nsigmas.
In case Nicola needs these changes, I think it would be Ok.
I suggest checking that the uncertainties are all returned in absolute values.

njacazio: Hi @hscheid in principle the smearer should not be changed as a core. What you can do is to add pieces if you think that something is missing

hscheid: If I am not wrong, this is a custom getter that is not used in the smearer itself. The smearTrack methods use the LUT entries directly. Maybe another approach could be to implement a second method that returns the absolute uncertainty values.

njacazio: Yes, that would be good, or get the relative uncertainty and scale it by pT or so

ALICE3/Core/DelphesO2TrackSmearer.cxx
```diff
@@ -239,9 +239,9 @@ double TrackSmearer::getEtaRes(int pdg, float nch, float eta, float pt)
 {
   float dummy = 0.0f;
   auto lutEntry = getLUTEntry(pdg, nch, 0., eta, pt, dummy);
-  auto sigmatgl = sqrt(lutEntry->covm[9]);           // sigmatgl2
-  auto etaRes = 1 / (sqrt(1 + sigmatgl * sigmatgl)); // propagate tgl to eta uncertainty
```
njacazio, line 243: I'm also not sure I understand this, apologies @NNicassio99 :)

hscheid (approved):

hscheid (approved):

PR #3214, opened 2023-07-28
PWGCF: FemtoUniverse -- PID without TOF and Phi cuts

zchochul:
Done: 
* Phi meson cuts,
* removing (temporarily TOF from PID) in track-Phi.


alibuild (approved):

PR #3227, opened 2023-07-31
EventFiltering: add class for helper functions

fgrosa:
To merge only once the tag for the 2023 filter campaign will be frozen.


EventFiltering/PWGHF/HFFilterHelpers.h
```diff
@@ -35,19 +35,25 @@
...
 #include "Common/Core/RecoDecay.h"
 #include "Common/Core/trackUtilities.h"
 #include "PWGHF/DataModel/CandidateReconstructionTables.h"
 #include "PWGHF/DataModel/CandidateSelectionTables.h"
 
+using namespace o2;
+using namespace o2::framework;
+using namespace o2::framework::expressions;
+using namespace o2::constants::math;
+
```
vkucera, line 56:
```suggestion
```
Using directives should never appear in headers.

vkucera (changes requested):

EventFiltering/PWGHF/HFFilterHelpers.h
fgrosa: Done, thanks. For my education, is it for memory consumption or is there any other reason?

vkucera (approved):

PR #3230, opened 2023-07-31
PWGJE: changing gen level particle selection to physical primary

nzardosh:


alibuild (approved):

PR #3231, opened 2023-07-31
Skip collisions whose BC falls in the ITS ramp

mpuccio:


alibuild (approved):

PR #3233, opened 2023-08-01
Fix trivial mistake on LS

BongHwi:
- from `if (trk1.sign() * trk2.sign() > 0)` to  `if (trk1.sign() * trk2.sign() < 0)`


alibuild (approved):

PR #3236, opened 2023-08-01
PWGEM: New cut setting and adaptation of old ones

feisenhu:
Adapt cut setting to select e+/e- tracks separately and tracks in +/- eta range.

### DIFF
--- a/PWGDQ/Core/CutsLibrary.cxx
+++ b/PWGDQ/Core/CutsLibrary.cxx
@@ -2040,50 +2040,50 @@ AnalysisCut* o2::aod::dqcuts::GetAnalysisCut(const char* cutName)
   }
 
   if (!nameStr.compare("electronPID_TPCnsigma")) {
-    cut->AddCut(VarManager::kTPCnSigmaEl, -3., 3., false, VarManager::kPin, 0.0, 2e+10, false);
-    cut->AddCut(VarManager::kTPCnSigmaPi, -3., 3., true, VarManager::kPin, 0.0, 2e+10, false);
-    cut->AddCut(VarManager::kTPCnSigmaKa, -3., 3., true, VarManager::kPin, 0.0, 2e+10, false);
-    cut->AddCut(VarManager::kTPCnSigmaPr, -3., 3., true, VarManager::kPin, 0.0, 2e+10, false);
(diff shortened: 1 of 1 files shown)


dsekihat (approved):

PWGDQ/Core/CutsLibrary.cxx
```diff
@@ -2022,50 +2040,50 @@ AnalysisCut* o2::aod::dqcuts::GetAnalysisCut(const char* cutName)
...
+    cut->AddCut(VarManager::kTPCnSigmaPr, -3., 3., true, VarManager::kPin, 0.0, 2e+10, false);
     return cut;
   }
 
   if (!nameStr.compare("electronPID_TPCnsigma_corr")) {
-    cut->AddCut(VarManager::kTPCnSigmaEl_Corr, -3., 2., false, VarManager::kPin, 0.0, 1e+10, false);
-    cut->AddCut(VarManager::kTPCnSigmaPi_Corr, -3., 3.5, true, VarManager::kPin, 0.0, 1e+10, false);
-    cut->AddCut(VarManager::kTPCnSigmaKa, -3., 4., true, VarManager::kPin, 0.0, 1e+10, false);
-    cut->AddCut(VarManager::kTPCnSigmaPr_Corr, -3., 4., true, VarManager::kPin, 0.0, 1e+10, false);
+    cut->AddCut(VarManager::kTPCnSigmaEl_Corr, -3., 2., false, VarManager::kPin, 0.0, 2e+10, false);
```
dsekihat, line 2051: Hi Florian, just a quick question. why do you have to change 1e+10 to 2e+10? Both are extremely large number, right?

feisenhu: Hey Daiki, when we ran over the full statistic of the LHC22o data set we increased the range of a couple of histograms (mee, pt, etc.) But I noticed when looking at some QA plots that the single track pt was cut at 10 GeV/c. To make sure we use the full range of the histograms we would need to increase the acceptance in pt and p.

dsekihat: 1e+10 does not corresponds to 10 GeV/c. it is 1000000000 GeV/c.

feisenhu: Yes you are absolutely right. This was my mistake. Sorry.

dsekihat (approved):

PR #3256, opened 2023-08-02
PWG-LF: memory fix in PP omega task

nepeivodaRS:


njacazio:
Hi Roman, very good, thanks! In the future probably you can get rid of the isMC if you use `doprocessGen` to check if the MC filling was enabled or not

njacazio (approved):

PR #3286, opened 2023-08-08
PWGMM: dndeta: skeleton of pure MC task to use with mctracks-to-aod

aalkin:


alibuild (approved):

PR #3299, opened 2023-08-09
Change DQ filter task name in the filter tables

mpuccio:
@vfeuilla: when changing the filter task is essential to change it here as well, otherwise the CEFP will ignore the presence of the DQ filters in the chain


alibuild (approved):

vfeuilla:
Oh you're right! I forgot about this part, because the initial idea was not to change the task's names but to replace the old task with the new one... Anyway thank you very much for noticing and correcting it! I'll make sure not to forget (if there is a) next time

PR #3305, opened 2023-08-10
Allow for downscaling of the f1p trigger

mpuccio:


alibuild (approved):

PR #3313, opened 2023-08-11
Simplifying the singles data taking. Use configuration

victor-gonzalez:
Also fine tuning the track selection cuts


alibuild (approved):

PR #3319, opened 2023-08-14
Fix the 0 collision cases

mpuccio:


mpuccio:
@njacazio LF PID was crashing in some corner cases with DF wo any collisions. These DFs are junk, but we must avoid failure, as the neighbouring DFs might be good.

njacazio (approved):

njacazio:
Super, thanks!

PR #3320, opened 2023-08-15
PWGCF: FemtoUniverse -- Add global track filter + custom filter (with configurables)

zchochul:
* Setting global tracks (using filters) to reduce the number of processed pairs.
* Setting custom cuts (using filters and configurables).


alibuild (approved):

PR #3324, opened 2023-08-15
Remove using-directives from common headers

vkucera:
Using-directives should never appear in headers because they irreversibly pollute the global namespace wherever they are included and can lead to unexpected behaviour, such as name conflicts. See [O2 CodingGuidelines](https://rawgit.com/AliceO2Group/CodingGuidelines/master/coding_guidelines.html?showone=Using_declarations_and_directives#Using_declarations_and_directives).

In fact, the `evsel` namespace was indeed defined ambiguously, once as `o2::aod::evsel` in `EventSelection.h` and another time as just `evsel` in `EventSelectionParams.h`. Maybe @ekryshen can comment whether this was intentional or not.

@ddobrigk I think we should have a CI check that forbids `using namespace` in header files. That would require fixing remaining occurrences in the PWG directories (which I did not do in this PR).

### DIFF
--- a/Common/CCDB/EventSelectionParams.cxx
+++ b/Common/CCDB/EventSelectionParams.cxx
@@ -11,7 +11,7 @@
 
 #include "EventSelectionParams.h"
 
-namespace evsel
+namespace o2::aod::evsel
 {
 const char* selectionLabels[kNsel] = {
   "kIsBBV0A",
@@ -49,6 +49,8 @@ const char* selectionLabels[kNsel] = {
   "kIsTriggerTVX",
   "kIsINT1"};
 }
+
+using namespace o2::aod::evsel;
 
 EventSelectionParams::EventSelectionParams(int system, int run)
 {
--- a/Common/CCDB/EventSelectionParams.cxx
+++ b/Common/CCDB/EventSelectionParams.cxx
@@ -48,7 +48,7 @@ const char* selectionLabels[kNsel] = {
   "kNoPileupTPC",
   "kIsTriggerTVX",
   "kIsINT1"};
-}
+} // namespace o2::aod::evsel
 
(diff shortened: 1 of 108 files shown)


ekryshen:
@vkucera It wasn't intentional. Thanks for the fix.

nzardosh (dismissed):

ekryshen (dismissed):

ddobrigk (dismissed):
Thanks! No objection to adding a CI; perhaps @TimoWilken can help with that? Thanks a lot!

vkucera:
Sorry, there were a few fixes needed. It should be ready for merging now.

aalkin (dismissed):

njacazio (approved):

aalkin:
Regarding `evsel` and `aod::evsel`, I would prefer them separate. Namespace `aod` is reserved for columns and tables. Probably the functional namespace can be something like `evsel_impl` or `evsel_detail`, since it is supposed to be distinct from `aod::evsel` which contains column definitions.

vkucera:
> 

Hi @aalkin , ok. Are you fine with renaming in a separate PR?

saganatt (approved):

fcatalan92 (approved):

aalkin:
@vkucera yes, of course, let's not waste the approvals :)

vkucera:
@ddobrigk @TimoWilken Is it fine for you to merge this without waiting for the remaining approvals?

ekryshen (approved):

ktf:
Merged this now. I have noticed two other quite bad practices, while reviewing this:

* For no reason one should have `#include <iostream>` in headers, as it bloats every single file which includes it with statics.
* In general there should not be any reason for having statics in the header files.

PR #3339, opened 2023-08-17
Changes for MC processing

ddobrigk:


ddobrigk:
@lhusova @ChiaraDeMartin95 @kcui1 This should allow us to calculate efficiencies (to first order). The efficiency denominator is filled both for all collisions and for collisions that have at least one PV, so we should be able to do also some cross-checks of that here. For the 2pc, though, likely the "in-analysis efficiency" is the most important (i.e. the one in which you assume you have a PV reconstructed) because of the existence of the triggering + same-collision-associated particle...

njacazio (approved):

PR #3349, opened 2023-08-18
[EMCAL-903] Make check for hardware trigger class case-insensitive

mfasDa:


mfasDa:
Trigger classes in LHC23f have lower-case detector strings, therefore check for trigger class fails due to case sensitivity

mpuccio (approved):

PR #3358, opened 2023-08-22
General update on Reso2Initializer

BongHwi:
- Remove RecoDecay GetPDGMass function
- Add bypass option for ccdb reading part
- Modify string comparison to int comparison
- Add FT0 multiplicity
- Remove temporary TPC multiplicity


alibuild (approved):

PR #3369, opened 2023-08-23
Optimise the binning of histograms

BongHwi:
By introducing variable


alibuild (approved):

PR #3372, opened 2023-08-24
PWGCF: FemtoUniverse -- Removing unused configurables from MC producer + fixing names in track-Phi

zchochul:
* Removing unused configurables and other small fixes added to the MC producer.
* Fixing naming in the track-Phi task.
* Adding femto and angular containers for same and mixed events for the track-track.
* Adding PDG vs pT plot for MC Truth particles.


alibuild (approved):

PR #3351, opened 2023-08-21
PWGHF: Add ML selection to Ds candidate selector + B0 matching to Bs candidate creator

pstahlhu:
Adding QA histogram + ML selection to DS candidate selector and B0 MC matching to Bs candidate creator @fgrosa @fcatalan92 .

PR #3376, opened 2023-08-24
PWGHF: simplify logic to parss of ML models

fgrosa:
Waiting for https://github.com/AliceO2Group/O2Physics/pull/3351 to be merged to include also Ds.


fgrosa:
@AlexBigO @DanielSamitz @pstahlhu this PR slightly simplifies the configuration for the ML application in your selectors. In particular, only one configurable is now present for the ML models that works both for models from CCDB or not instead of having two separate vectors. Now the vector of model paths is always initialised to a vector with length `nptlims - 1` and they have to be initialised in the `init` function to avoid the logic of the empty vector that was there before (which was not optimal because it is not possible to have an empty vector on hyperloop).

adubla (approved):

PR #3389, opened 2023-08-27
PWGLF: Remove centrality tabel from MC events

BongHwi:


alibuild (approved):

PR #3390, opened 2023-08-28
PWGEM/PhotonMeson: Add MCV0 task

f3sch:
1. add '.cache' to .gitignore
  clangd stores the index inside the project root when using background
  indexing, other IDEs might also do this. this file is should not be shared and thus is okay to ignore in vc
2. add exploratory task for v0 photons @ddobrigk @dsekihat

### DIFF
--- a/PWGEM/PhotonMeson/Tasks/CheckMCV0.cxx
+++ b/PWGEM/PhotonMeson/Tasks/CheckMCV0.cxx
@@ -9,35 +9,18 @@
 // granted to it by virtue of its status as an Intergovernmental Organization
 // or submit itself to any jurisdiction.
 
-/// \brief write relevant information for photon conversion analysis to an AO2D.root file.
+/// \brief check the v0 phase-space
 /// dependencies: o2-analysis-lf-lambdakzeromcfinder
-/// \author daiki.sekihata@cern.ch
-
-// #include "PWGEM/PhotonMeson/DataModel/gammaTables.h"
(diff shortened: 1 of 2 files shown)


PWGEM/PhotonMeson/Tasks/CheckMCV0.cxx
```diff
@@ -9,35 +9,15 @@
 // granted to it by virtue of its status as an Intergovernmental Organization
 // or submit itself to any jurisdiction.
 
-/// \brief write relevant information for photon conversion analysis to an AO2D.root file.
-/// dependencies: o2-analysis-lf-lambdakzeromcfinder
```
mhemmer-cern, line 13: the brief description and dependencies should  stay I think. I thought clang would even complain if they are gone

PWGEM/PhotonMeson/Tasks/CheckMCV0.cxx
```diff
@@ -9,35 +9,15 @@
...
-/// \author daiki.sekihata@cern.ch
```
mhemmer-cern, line 14: same for the author, it shouldn't be removed, author ship should only be added.

mhemmer-cern (changes requested):

mhemmer-cern (approved):
Seems good from EM side

mhemmer-cern (approved):

ddobrigk (approved):

PR #3393, opened 2023-08-28
PWGHF: Add debug histograms for BDT application in D0 selector

lucamicheletti93:
Adding debug histograms for BDT application

### DIFF
--- a/PWGHF/TableProducer/candidateSelectorD0.cxx
+++ b/PWGHF/TableProducer/candidateSelectorD0.cxx
@@ -77,8 +77,20 @@ struct HfCandidateSelectorD0 {
 
   using TracksSel = soa::Join<aod::TracksWDcaExtra, aod::TracksPidPi, aod::TracksPidKa>;
 
+  // Define histograms
+  AxisSpec axisMassDmeson{200, 1.7f, 2.1f};
+  AxisSpec axisBdtScore{100, 0.f, 1.f};
+  AxisSpec axisSelStatus{2, -0.5f, 1.5f};
+
+  HistogramRegistry registry{"registry",
+                             {{"hMassDmesonSel", ";#it{M}(D) (GeV/#it{c}^{2});counts", {HistType::kTH1F, {axisMassDmeson}}}}};
+
   void init(InitContext& initContext)
   {
+    if (applyMl) {
+      registry.add("DebugBdt/hBdtScoreVsStatus", ";BDT score;status", {HistType::kTH2F, {axisBdtScore, axisSelStatus}});
+    }
+
     selectorPion.setRangePtTpc(ptPidTpcMin, ptPidTpcMax);
     selectorPion.setRangeNSigmaTpc(-nSigmaTpcMax, nSigmaTpcMax);
     selectorPion.setRangeNSigmaTpcCondTof(-nSigmaTpcCombinedMax, nSigmaTpcCombinedMax);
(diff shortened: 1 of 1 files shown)


PWGHF/TableProducer/candidateSelectorD0.cxx
```diff
@@ -77,8 +77,20 @@ struct HfCandidateSelectorD0 {
 
   using TracksSel = soa::Join<aod::TracksWDcaExtra, aod::TracksPidPi, aod::TracksPidKa>;
 
+  // Define histograms
+  AxisSpec axisMassDmeson{200, 1.7f, 2.1f};
+  AxisSpec axisBdtScore{100, 0.f, 1.f};
+  AxisSpec axisSelStatus{2, -0.5f, 1.5f};
+
+  HistogramRegistry registry{"registry",
+                             {{"hMassDmesonSel", ";#it{M}(D) (GeV/#it{c}^{2});counts", {HistType::kTH1F, {axisMassDmeson}}}}};
```
fgrosa, line 86: I would define this also only in case of debug (outside `if (applyML))

PWGHF/TableProducer/candidateSelectorD0.cxx
```diff
@@ -77,8 +77,20 @@ struct HfCandidateSelectorD0 {
...
+
   void init(InitContext& initContext)
   {
+    if (applyMl) {
+      registry.add("DebugBdt/hBdtScoreVsStatus", ";BDT score;status", {HistType::kTH2F, {axisBdtScore, axisSelStatus}});
```
fgrosa, line 91: Can you please add also those for the other 2 BDT scores?

fgrosa (changes requested):
Thanks @lucamicheletti93! I have a couple of minor comments, then it is fine for me.

vkucera (approved):

PWGHF/TableProducer/candidateSelectorD0.cxx
```diff
@@ -346,6 +359,15 @@ struct HfCandidateSelectorD0 {
...
+        registry.fill(HIST("DebugBdt/hBdtScore1VsStatus"), outputMl[0], statusD0);
+        registry.fill(HIST("DebugBdt/hBdtScore1VsStatus"), outputMl[0], statusD0bar);
+        registry.fill(HIST("DebugBdt/hBdtScore2VsStatus"), outputMl[1], statusD0);
+        registry.fill(HIST("DebugBdt/hBdtScore2VsStatus"), outputMl[1], statusD0bar);
+        registry.fill(HIST("DebugBdt/hBdtScore3VsStatus"), outputMl[2], statusD0);
+        registry.fill(HIST("DebugBdt/hBdtScore3VsStatus"), outputMl[2], statusD0bar);
+      }
+      if (statusD0 != 0 || statusD0bar != 0) {
+        registry.fill(HIST("hMassDmesonSel"), invMassD0ToPiK(candidate));
       }
```
vkucera, line 371: @lucamicheletti93 This causes a crash if `!applyMl` because `hMassDmesonSel` does not exist. I guess it should be moved inside the block just above. Can you please fix it?

vkucera: Or define the histogram outside the `applyMl` condition if meant for general debugging.
Actually, the path is wrong so the histogram does not exist in any case.

lucamicheletti93:
Ciao @vkucera! Here's the PR with the fix https://github.com/AliceO2Group/O2Physics/pull/3420

PR #3398, opened 2023-08-29
PWGHF: Add candidate tag column to D0 tree creator

Mingyu3360715:
This commit create a candidate type column to the output table, which is going to be used for future analysis. The content is based on the previous D2H report on June 9th.


vkucera:
Hi @Mingyu3360715 , as @fgrosa commented in the D2H meeting, this new column does not add any new information in the table and the relevant values can be computed from the already present columns.

Mingyu3360715:
Hi @vkucera , indeed this column doesnot add new information, but it combines the flagMcMatchRec, originMcRec and candidateSelFlag information into one single column. So if I need to derive the MC table for ML study, the lite table itself could provide all the necessary information. 
Another way of adding this column is to extend the current candidateSelFlag column, since it only stores flags for isSelD0() and isSelD0bar() information. What do you think?

vkucera:
What would you extend it with? Isn't all the information you need already stored in the columns `CandidateSelFlag`, `FlagMc` and `OriginMcRec`?

fgrosa:
Hi @Mingyu3360715 I would indeed avoid to duplicate the information, as @vkucera pointed out. One possibility though is to replace the current information stored with your single column in the lite version of the tree (in this case we do not duplicate the information, but we replace it with a more compact format).

Mingyu3360715:
Thanks a lot for the comment @fgrosa . I removed the CandidateType column in the HFCANDD0FULL table and replaced the CandidateSelFlag in the HFCANDD0LITE as you pointed.

PWGHF/TableProducer/treeCreatorD0ToKPi.cxx
```diff
@@ -390,11 +405,36 @@ struct HfTreeCreatorD0ToKPi {
...
+          } else {
+            if (candidate.flagMcMatchRec() == -(1 << DecayType::D0ToPiK)) {
+              candType |= o2::aod::full::ReflectedD0;
+            }
+          }
+          if (candidate.originMcRec() == RecoDecay::OriginType::Prompt) {
+            candType |= o2::aod::full::Prompt;
+          } else {
+            candType |= o2::aod::full::NonPrompt;
+          }
```
fgrosa, line 421: > Thanks a lot for the comment @fgrosa . I removed the CandidateType column in the HFCANDD0FULL table and replaced the CandidateSelFlag in the HFCANDD0LITE as you pointed.

Hi @Mingyu3360715 thanks a lot! If I am not mistaken, you should also remove the `FlagMc` and `OriginMcRec` ones, since they are also not needed anymore. Finally, you could move this part of the code inside the `fillTable` method in case of `fillCandidateLiteTable` to 1) avoid code duplication and 2) avoid the computation of `candType` in case of full tables.

Mingyu3360715:
Hi @fgrosa , thanks for the detailed comment. The modification has been made based on it. 
Hi @vkucera , could you please have a look at it and comment if you have any suggestion.

PWGHF/TableProducer/treeCreatorD0ToKPi.cxx
```diff
@@ -32,6 +32,17 @@ namespace o2::aod
...
+
+enum SelectionType {
+  SelD0 = BIT(0),
+  SelD0bar = BIT(1),
+  IsSignal = BIT(2),
+  ReflectedD0 = BIT(3),
+  ReflectedD0bar = BIT(4),
+  Prompt = BIT(5),
+  NonPrompt = BIT(6)
+};
```
vkucera, line 44: You don't need that many options, because reflection implies signal and prompt implies not non-prompt.
Similar kind of bitmap is implemented in https://github.com/AliceO2Group/O2Physics/blob/master/PWGHF/HFC/TableProducer/correlatorDMesonPairs.cxx#L47-L51
In your case, you should need just 5 bits: selD0, selD0bar, trueD0, trueD0bar, prompt.

PWGHF/TableProducer/treeCreatorD0ToKPi.cxx
```diff
@@ -229,10 +239,34 @@ struct HfTreeCreatorD0ToKPi {
   }
 
   template <typename T, typename U>
-  auto fillTable(const T& candidate, const U& prong0, const U& prong1, int candFlag, double invMass, double cosThetaStar,
+  auto fillTable(const T& candidate, const U& prong0, const U& prong1, int candFlag, int candType, double invMass, double cosThetaStar,
                  double ct, double y, double e, int8_t flagMc, int8_t origin)
   {
     if (fillCandidateLiteTable) {
+      if (flagMc == 0) {
```
vkucera, line 246: The line `flagMc == 0` does not seem correct.

PWGHF/TableProducer/treeCreatorD0ToKPi.cxx
```diff
@@ -229,10 +239,34 @@ struct HfTreeCreatorD0ToKPi {
...
+      }
+      if (flagMc == 1) {
+        if (flagMc == -(1 << DecayType::D0ToPiK)) {
+          candType |= o2::aod::full::IsSignal;
+        } else {
+          if (flagMc == (1 << DecayType::D0ToPiK)) {
+            candType |= o2::aod::full::ReflectedD0bar;
+          }
+        }
+      }
```
vkucera, line 263: You can simplify this to just setting trueD0 or trueD0bar.

PWGHF/TableProducer/treeCreatorD0ToKPi.cxx
```diff
@@ -229,10 +239,34 @@ struct HfTreeCreatorD0ToKPi {
...
+      if (flagMc == 0) {
+        if (flagMc == (1 << DecayType::D0ToPiK)) {
+          candType |= o2::aod::full::IsSignal;
+        } else {
+          if (flagMc == -(1 << DecayType::D0ToPiK)) {
+            candType |= o2::aod::full::ReflectedD0;
+          }
+        }
+      }
+      if (flagMc == 1) {
```
vkucera, line 255: Again, the `flagMc == 1` does not seem correct.

PWGHF/TableProducer/treeCreatorD0ToKPi.cxx
```diff
@@ -229,10 +239,34 @@ struct HfTreeCreatorD0ToKPi {
...
+            candType |= o2::aod::full::ReflectedD0bar;
+          }
+        }
+      }
+      if (origin == RecoDecay::OriginType::Prompt) {
+        candType |= o2::aod::full::Prompt;
+      }
+      if (origin == RecoDecay::OriginType::NonPrompt) {
+        candType |= o2::aod::full::NonPrompt;
+      }
```
vkucera, line 269: This can be simplified to just setting prompt.

PWGHF/TableProducer/treeCreatorD0ToKPi.cxx
```diff
@@ -32,6 +32,17 @@ namespace o2::aod
 {
 namespace full
 {
+
+enum SelectionType {
+  SelD0 = BIT(0),
```
vkucera, line 37: Define simple integers here and set the bits using the `SETBIT` macro.

PWGHF/TableProducer/treeCreatorD0ToKPi.cxx
```diff
@@ -354,11 +386,14 @@ struct HfTreeCreatorD0ToKPi {
...
+      int candType = 0;
       if (candidate.isSelD0()) {
-        fillTable(candidate, prong0, prong1, 0, invMassD0ToPiK(candidate), cosThetaStarD0(candidate), ctD, yD, eD, 0, 0);
+        candType |= o2::aod::full::SelD0;
+        fillTable(candidate, prong0, prong1, 0, candType, invMassD0ToPiK(candidate), cosThetaStarD0(candidate), ctD, yD, eD, 0, 0);
       }
       if (candidate.isSelD0bar()) {
-        fillTable(candidate, prong0, prong1, 1, invMassD0barToKPi(candidate), cosThetaStarD0bar(candidate), ctD, yD, eD, 0, 0);
+        candType |= o2::aod::full::SelD0bar;
+        fillTable(candidate, prong0, prong1, 1, candType, invMassD0barToKPi(candidate), cosThetaStarD0bar(candidate), ctD, yD, eD, 0, 0);
```
vkucera, line 396: In the `isSelD0bar` block, `candType` can already have the `isSelD0` bit set which you probably do not want. Therefore you should assign the value with `=` instead of adding the bit.
You don't even need to define the `candType` variable. You can just pass the value directly to the `fillTable` function.

vkucera (changes requested):

PR #3399, opened 2023-08-29
PWGCF: FemtoUniverse -- Changing PID configurables

zchochul:
Previously, to select a particle for analysis, two configurables had to be changed: `ConfTrackChoicePartOne` (which selects the PID) and `ConfPDGCodePartOne` (which sets the particle's mass). This could cause unnecessary errors if the user forgot to change both of these variables. I decided to get rid of one of them. Now the only thing to remember is to set the PDG code of the particle.


alibuild (dismissed):

alibuild (approved):

PR #3401, opened 2023-08-29
PWGHF: make the reduced data format for B hadrons self contained

fgrosa:
@AlexBigO @apalasciano I made few changes in the reduced data model for B hadrons:
- I made it self contained so that the access to the parent `AO2D` files should not be needed anymore
- I changed a bit the data model, fragmenting more the tables (e.g. the covariances for pion and D tracks are separated now) and replacing the topological variables for the D mesons with the position of their decay vertex (so that we can recompute the topological variables if we want).

I still have to test the code, but you can already have a look if the changes are fine for you. Thanks!

### DIFF
--- a/PWGHF/D2H/DataModel/ReducedDataModel.h
+++ b/PWGHF/D2H/DataModel/ReducedDataModel.h
@@ -45,7 +45,7 @@ DECLARE_SOA_COLUMN(Bz, bz, float); //! Magnetic field in z-direction
 DECLARE_SOA_COLUMN(OriginalCollisionCount, originalCollisionCount, int); //! Size of COLLISION table processed
 } // namespace hf_reduced_collision
 
-DECLARE_SOA_TABLE(HfReducedCollisions, "AOD", "HFREDCOLLISION", //! Table with collision for reduced workflow
+DECLARE_SOA_TABLE(HfRedCollisions, "AOD", "HFREDCOLLISION", //! Table with collision for reduced workflow
                   soa::Index<>,
                   collision::PosX,
                   collision::PosY,
@@ -58,9 +58,9 @@ DECLARE_SOA_TABLE(HfReducedCollisions, "AOD", "HFREDCOLLISION", //! Table with c
                   collision::CovZZ,
                   hf_reduced_collision::Bz);
 
-using HfReducedCollision = HfReducedCollisions::iterator;
+using HfRedCollision = HfRedCollisions::iterator;
 
-DECLARE_SOA_TABLE(HfOriginalCollisionsCounter, "AOD", "HFCOLCOUNTER", //! Table with original number of collisions
+DECLARE_SOA_TABLE(HfOrigColCounts, "AOD", "HFORIGCOLCOUNT", //! Table with original number of collisions
                   hf_reduced_collision::OriginalCollisionCount);
 
 namespace hf_track_par_cov
@@ -117,16 +117,19 @@ DECLARE_SOA_COLUMN(Pz, pz, float); //! Momentum in z-direction in GeV/c
 
 namespace hf_track_index_reduced
 {
-DECLARE_SOA_INDEX_COLUMN(HfReducedCollision, hfReducedCollision); //! ReducedCollision index
-DECLARE_SOA_INDEX_COLUMN(Track, track);                           //! Track index
-DECLARE_SOA_COLUMN(HasTPC, hasTPC, bool);                         //! Flag to check if track has a TPC match
-DECLARE_SOA_COLUMN(HasTOF, hasTOF, bool);                         //! Flag to check if track has a TOF match
+DECLARE_SOA_INDEX_COLUMN(HfRedCollision, hfRedCollision); //! ReducedCollision index
+DECLARE_SOA_COLUMN(TrackId, trackId, int);                //! Original track index
+DECLARE_SOA_COLUMN(Prong0Id, prong0Id, int);              //! Original track index
+DECLARE_SOA_COLUMN(Prong1Id, prong1Id, int);              //! Original track index
+DECLARE_SOA_COLUMN(Prong2Id, prong2Id, int);              //! Original track index
+DECLARE_SOA_COLUMN(HasTPC, hasTPC, bool);                 //! Flag to check if track has a TPC match
+DECLARE_SOA_COLUMN(HasTOF, hasTOF, bool);                 //! Flag to check if track has a TOF match
 } // namespace hf_track_index_reduced
 
-DECLARE_SOA_TABLE(HfTracksReduced, "AOD", "HFTRACKRED", //! Table with track information for reduced workflow
+DECLARE_SOA_TABLE(HfRedTracks, "AOD", "HFREDTRACK", //! Table with track information for reduced workflow
                   soa::Index<>,
                   hf_track_index_reduced::TrackId,
-                  hf_track_index_reduced::HfReducedCollisionId,
+                  hf_track_index_reduced::HfRedCollisionId,
                   HFTRACKPARCOV_COLUMNS);
 
 namespace hf_track_pid_reduced
@@ -135,9 +138,9 @@ DECLARE_SOA_COLUMN(Pt, pt, float); //! Transverse momentum of the track in GeV/c
 } // namespace hf_track_pid_reduced
 
 // table with all attributes needed to call statusTpcAndTof() in the selector task
-DECLARE_SOA_TABLE(HfTracksPidReduced, "AOD", "HFTRACKPIDRED", //! Table with PID track information for reduced workflow
+DECLARE_SOA_TABLE(HfRedPidTracks, "AOD", "HFREDTRACKPID", //! Table with PID track information for reduced workflow
                   o2::soa::Index<>,
-                  hf_track_index_reduced::HfReducedCollisionId,
+                  hf_track_index_reduced::HfRedCollisionId,
                   hf_track_pid_reduced::Pt,
                   hf_track_index_reduced::HasTPC,
                   hf_track_index_reduced::HasTOF,
@@ -160,10 +163,10 @@ DECLARE_SOA_COLUMN(InvMass, invMass, float);         //! Invariant mass of 2pron
 
 } // namespace hf_cand_2prong_reduced
 
-DECLARE_SOA_TABLE(HfCand2ProngReduced, "AOD", "HFCAND2PRONGRED", //! Table with 2prong candidate information for reduced workflow
+DECLARE_SOA_TABLE(HfRedCand2Prongs, "AOD", "HFREDCAND2PRONG", //! Table with 2prong candidate information for reduced workflow
                   o2::soa::Index<>,
-                  hf_track_index::Prong0Id, hf_track_index::Prong1Id,
-                  hf_track_index_reduced::HfReducedCollisionId,
+                  hf_track_index_reduced::Prong0Id, hf_track_index_reduced::Prong1Id,
+                  hf_track_index_reduced::HfRedCollisionId,
                   HFTRACKPARCOV_COLUMNS,
                   hf_cand_2prong_reduced::CPA,
                   hf_cand_2prong_reduced::DecayLength,
@@ -185,10 +188,10 @@ auto invMassDplusToPiKPi(const T& pVec0, const T& pVec1, const T& pVec2)
 }
 } // namespace hf_cand_3prong_reduced
 
-DECLARE_SOA_TABLE(HfCand3ProngReduced, "AOD", "HFCAND3PRONGRED", //! Table with 3prong candidate information for reduced workflow
+DECLARE_SOA_TABLE(HfRedCand3Prongs, "AOD", "HFREDCAND3PRONG", //! Table with 3prong candidate information for reduced workflow
                   o2::soa::Index<>,
-                  hf_track_index::Prong0Id, hf_track_index::Prong1Id, hf_track_index::Prong2Id,
-                  hf_track_index_reduced::HfReducedCollisionId,
+                  hf_track_index_reduced::Prong0Id, hf_track_index_reduced::Prong1Id, hf_track_index_reduced::Prong2Id,
+                  hf_track_index_reduced::HfRedCollisionId,
                   HFTRACKPARCOV_COLUMNS,
                   hf_cand_3prong_reduced::CPA,
                   hf_cand_3prong_reduced::DecayLength,
@@ -211,24 +214,21 @@ DECLARE_SOA_COLUMN(EtaProng1, etaProng1, float); //! Pseudorapidity of the track
 } // namespace hf_b0_mc
 
 // table with results of reconstruction level MC matching
-DECLARE_SOA_TABLE(HfDPiMcRecReduced, "AOD", "HFDPIMCRECRED", //! Table with reconstructed MC information on DPi(<-B0) pairs for reduced workflow
+DECLARE_SOA_TABLE(HfMcRecRedDpPis, "AOD", "HFMCRECREDDPPI", //! Table with reconstructed MC information on DPi(<-B0) pairs for reduced workflow
                   hf_cand_b0::Prong0Id,
                   hf_track_index::Prong1Id,
                   hf_cand_b0::FlagMcMatchRec,
-                  hf_cand_b0::OriginMcRec,
                   hf_cand_b0::DebugMcRec,
                   hf_b0_mc::PtMother);
 
 // Table with same size as HFCANDB0
-DECLARE_SOA_TABLE(HfB0McRecReduced, "AOD", "HFB0MCRECRED", //! Reconstruction-level MC information on B0 candidates for reduced workflow
+DECLARE_SOA_TABLE(HfMcRecRedB0s, "AOD", "HFMCRECREDB0", //! Reconstruction-level MC information on B0 candidates for reduced workflow
                   hf_cand_b0::FlagMcMatchRec,
-                  hf_cand_b0::OriginMcRec,
                   hf_cand_b0::DebugMcRec,
                   hf_b0_mc::PtMother);
 
-DECLARE_SOA_TABLE(HfB0McGenReduced, "AOD", "HFB0MCGENRED", //! Generation-level MC information on B0 candidates for reduced workflow
+DECLARE_SOA_TABLE(HfMcGenRedB0s, "AOD", "HFMCGENREDB0", //! Generation-level MC information on B0 candidates for reduced workflow
                   hf_cand_b0::FlagMcMatchGen,
-                  hf_cand_b0::OriginMcGen,
                   hf_b0_mc::PtTrack,
                   hf_b0_mc::YTrack,
                   hf_b0_mc::EtaTrack,
@@ -246,7 +246,7 @@ namespace hf_cand_b0_config
 DECLARE_SOA_COLUMN(MySelectionFlagD, mySelectionFlagD, int8_t); //! Flag to filter selected D+ mesons
 } // namespace hf_cand_b0_config
 
-DECLARE_SOA_TABLE(HfCandB0Config, "AOD", "HFCANDB0CONFIG", //! Table with configurables information for reduced workflow
+DECLARE_SOA_TABLE(HfCandB0Configs, "AOD", "HFCANDB0CONFIG", //! Table with configurables information for reduced workflow
                   hf_cand_b0_config::MySelectionFlagD);
 
 namespace hf_bplus_mc
@@ -266,22 +266,19 @@ DECLARE_SOA_COLUMN(EtaProng1, etaProng1, float); //! Pseudorapidity of the track
 } // namespace hf_bplus_mc
 
 // table with results of reconstruction level MC matching
-DECLARE_SOA_TABLE(HfD0PiMcRecReduced, "AOD", "HFD0PIMCRECRED", //! Table with reconstructed MC information on D0Pi(<-B+) pairs for reduced workflow
+DECLARE_SOA_TABLE(HfMcRecRedD0Pis, "AOD", "HFMCRECREDD0PI", //! Table with reconstructed MC information on D0Pi(<-B+) pairs for reduced workflow
                   hf_cand_bplus::Prong0Id,
                   hf_track_index::Prong1Id,
                   hf_cand_bplus::FlagMcMatchRec,
-                  hf_cand_bplus::OriginMcRec,
                   hf_bplus_mc::PtMother);
 
 // Table with same size as HFCANDBPLUS
-DECLARE_SOA_TABLE(HfBpMcRecReduced, "AOD", "HFBPMCRECRED", //! Reconstruction-level MC information on B+ candidates for reduced workflow
+DECLARE_SOA_TABLE(HfMcRecRedBps, "AOD", "HFMCRECREDBP", //! Reconstruction-level MC information on B+ candidates for reduced workflow
                   hf_cand_bplus::FlagMcMatchRec,
-                  hf_cand_bplus::OriginMcRec,
                   hf_bplus_mc::PtMother);
 
-DECLARE_SOA_TABLE(HfBpMcGenReduced, "AOD", "HFBPMCGENRED", //! Generation-level MC information on B+ candidates for reduced workflow
+DECLARE_SOA_TABLE(HfMcGenRedBps, "AOD", "HFMCGENREDBP", //! Generation-level MC information on B+ candidates for reduced workflow
                   hf_cand_bplus::FlagMcMatchGen,
-                  hf_cand_bplus::OriginMcGen,
                   hf_bplus_mc::PtTrack,
                   hf_bplus_mc::YTrack,
                   hf_bplus_mc::EtaTrack,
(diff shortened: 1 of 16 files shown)


PWGHF/D2H/DataModel/ReducedDataModel.h
```diff
@@ -183,16 +169,39 @@ auto invMassDplusToPiKPi(const T& pVec0, const T& pVec1, const T& pVec2)
...
+                  aod::track::Px<aod::track::Signed1Pt, aod::track::Snp, aod::track::Alpha>,
+                  aod::track::Py<aod::track::Signed1Pt, aod::track::Snp, aod::track::Alpha>,
+                  aod::track::Pz<aod::track::Signed1Pt, track::Tgl>);
+
+DECLARE_SOA_TABLE(HfRed2ProngsCov, "AOD", "HFRED2PRONGSCOV", //! Table with 2prong candidate covariance for reduced workflow
+                  o2::soa::Index<>,
+                  HFTRACKPARCOV_COLUMNS,
+                  o2::soa::Marker<1>);
+
+DECLARE_SOA_TABLE(HfRed3Prongs, "AOD", "HFRED3PRONG", //! Table with 3prong candidate information for reduced workflow
```
AlexBigO, line 190: Maybe add a `S` at the end of the table description to be consistent with `HFRED2PRONGS`

AlexBigO:
Ciao @fgrosa ! Thanks a lot :)

Despite the MC part that does not work currently (because of the `hf_cand_b0` and `hf_track_index` in `HfMcRecRedDpPis`), I only encountered one issue when testing the code: we need to `Spawns<aod::HfRedTracksExt>` at some point otherwise it complains as we subscribe to `HfRedTracks`. Either we do it at the end of the preliminary workflow (requiring to save the extended table which is not the optimal solution) or we do it in the reduced candidate creator. This second option requires to subscribe to `HfRedTracksBase` in the main `process` function (as only this table is saved in our derived AOD file) and then `Spawns` our extended table in the `Expressions` structure.

PWGHF/D2H/DataModel/ReducedDataModel.h
fgrosa: I think that the convention is to put the `s` in the first argument and not in the third one (see e.g. https://github.com/AliceO2Group/AliceO2/blob/90c5ac72684d34daa85cb024ba7cdb1e0f126c0a/Framework/Core/include/Framework/AnalysisDataModel.h#L314), so I would let it like this if you agree

AlexBigO: Indeed ! I also ended up doing it like this. In this spirit, we'll need to remove the `s` in `HFCAND2PRONGS`.

Actually, if we want to use `DECLARE_SOA_INDEX_COLUMN_FULL` (with an index pointing to a table), we need the table to be defined as follows: `DECLARE_SOA_TABLE(Name, "AOD", Description)` with `Name = Description + 's'`. This requires to redefine the name of `HfRedTrackBase` in `HfRedTrackBases`, otherwise the merger is looking for the name of the table (with the last letter removed): `hfredtrackbas` in our case (and it creates an error).

AlexBigO:
Hello @fgrosa @apalasciano !

Here are the modifications applied to the work already done by Fabrizio:
- fix `Name` and `Description` of table that we we want to store and who have index column pointing to them: the convention `Name` = `Description`s must be followed so that the `o2-aod-merger` works
- `process` -> `processData` in DPi creator
- only store indices of our derived AODs and not of the "parent" AO2Ds (mandatory to follow the self-contained strategy)
- split B0 candidate table into 2 tables: we extract the columns with `prong*Id` and put them in a dedicated table that we'll `Join` later on. This is especially convenient because `HfCandB0Base` is common to standard and reduced workflows while the tables containing the `prong*Id` differs from standard and reduced workflows and allows us to define index columns pointing to our derived AODs while keeping the same code structure as the standard workflow
- ensure that we apply the same preselections on DPi pairs in DPi creator and B0 creator: we store the invariant mass window cut in the config table and then impose this value in the B0 creator. We actually apply a slightly tighter cut (using a new configurable `invMassWindowDPiTolerance`) to be sure we don't form "new" DPi pairs at the B0 creator level
- apply invariant mass preselection on the square of the invariant mass (`RecoDecay::m2` instead of `RecoDecay::m`) to avoid unnecessary consumption due to `std::sqrt` call
- `Spawns` the extension of the Reduced Track table in B0 creator so we can call it later in the worklow (in the selector especially)
- follow new convention and get rid of `getMassPDG` by using `O2DatabasePDG` service. *With the test I made, I was only able to make it work by calling getters of the `pdg` instance in the `init` function, it did not work at the beginning of the `struct`* <-- tell me if this is the way to go or if I can improve it :)

Disclaimer: for the moment, these modifs are only for B0 but we'll implement them for B+ if you find them satisfactory :D

PWGHF/D2H/TableProducer/candidateCreatorB0Reduced.cxx
```diff
@@ -44,26 +46,33 @@ struct HfCandidateCreatorB0Reduced {
...
+
+  // O2DatabasePDG service
+  Service<o2::framework::O2DatabasePDG> pdg;
+
+  double massPi{0.};
+  double massD{0.};
+  double massB0{0.};
+  double invMass2DPi{0.};
+  double invMass2DPiMin{0.};
+  double invMass2DPiMax{0.};
```
fgrosa, line 61: These two variables can be declared directly in the `process` function, no need to have them as datamembers

fgrosa (commented):
Hi @AlexBigO thanks a lot for testing the code and fixing the remaining issues! I have only a very minor comment and then I think that we can extend your fixes to the B+. After that, we will fix the conflicts and we can mark the PR as ready for review in my opinion.

PWGHF/D2H/TableProducer/candidateCreatorB0Reduced.cxx
AlexBigO: Hi @fgrosa ! Thanks for the feedback. I will apply your comment and transpose the fixes to B+ (I will try to do it by tomorrow).

AlexBigO: Also, now we should be able to remove the leptons columns of the PID table. For the moment (for B0 and B+) we only need Pi but should we also store K and Pr?

fgrosa: Hi @AlexBigO that's great. I would only store the pion PID (keep in mind that already the pion one is not that powerful, since most of the tracks are pions anyway, so there is no need to have something complex for the PID of the bachelor)

PWGHF/D2H/DataModel/ReducedDataModel.h
```diff
@@ -183,16 +163,66 @@ auto invMassDplusToPiKPi(const T& pVec0, const T& pVec1, const T& pVec2)
...
+
+DECLARE_SOA_TABLE(HfRed2ProngsCov, "AOD", "HFRED2PRONGSCOV", //! Table with 2prong candidate covariance for reduced workflow
                   o2::soa::Index<>,
-                  hf_track_index::Prong0Id, hf_track_index::Prong1Id, hf_track_index::Prong2Id,
-                  hf_track_index_reduced::HfReducedCollisionId,
                   HFTRACKPARCOV_COLUMNS,
-                  hf_cand_3prong_reduced::CPA,
-                  hf_cand_3prong_reduced::DecayLength,
-                  hf_cand_3prong_reduced::InvMass);
+                  o2::soa::Marker<1>);
```
mfaggin, line 184: can you just remind me what this `Marker<1>` means?

PWGHF/D2H/TableProducer/dataCreatorDplusPiReduced.cxx
```diff
@@ -75,16 +82,20 @@ struct HfDataCreatorDplusPiReduced {
...
   o2::base::Propagator::MatCorrType matCorr = o2::base::Propagator::MatCorrType::USEMatCorrLUT;
   int runNumber;
 
-  double massPi = RecoDecay::getMassPDG(kPiPlus);
-  double massD = RecoDecay::getMassPDG(pdg::Code::kDMinus);
-  double massB0 = RecoDecay::getMassPDG(pdg::Code::kB0);
-  double massDPi{0.};
-  double invMassD{0.};
+  // O2DatabasePDG service
+  Service<o2::framework::O2DatabasePDG> pdg;
```
mfaggin, line 92: Here you define this `pdg` object because later you need it to get the particle masses from PDG. But where do you define it? You declare it here w/o definition, and then you de-reference the pointer in lines 155, 156, 157. Am I overlooking something?

PWGHF/D2H/TableProducer/dataCreatorDplusPiReduced.cxx
```diff
@@ -188,38 +207,31 @@ struct HfDataCreatorDplusPiReduced {
...
-    static int nCol = 0;
     for (const auto& collision : collisions) {
-      nCol++;
 
       // helpers for ReducedTables filling
-      int hfReducedCollisionIndex = hfReducedCollision.lastIndex() + 1;
-      std::vector<int64_t> selectedTracksPion;
-      selectedTracksPion.reserve(trackIndices.size());
+      int indexHfReducedCollision = hfReducedCollision.lastIndex() + 1;
+      std::map<int64_t, int64_t> selectedTracksPion;
```
mfaggin, line 231: For better code readability, here I'd add a comment explaining that:
 - the `key` is the `track.globalIndex()`
 - the `value` is the track index in the table of the found HF pions

PWGHF/D2H/DataModel/ReducedDataModel.h
AlexBigO: Hi @mfaggin , this `Marker<>` allows to define several tables with the same columns. As we want to save the same information in `HfRed2ProngsCov` and `HfRed3ProngsCov`, but we still want to distinguish them by name and description, we mark the 2prongs table with `Marker<1>` and the 3prongs one with `Marker<2>`.

PWGHF/D2H/TableProducer/dataCreatorDplusPiReduced.cxx
AlexBigO: Actually, this is the first time I use this new functionality. I just followed the [tutorial](https://github.com/AliceO2Group/O2Physics/blob/master/Tutorials/src/usingPDGService.cxx#L20) indicated in the error messages of the log file.

PWGHF/D2H/DataModel/ReducedDataModel.h
mfaggin: thank you!

PWGHF/D2H/TableProducer/dataCreatorDplusPiReduced.cxx
mfaggin: Thanks for providing the link to the tutorial. Yes, in principle the default constructor is called (see here: https://github.com/AliceO2Group/AliceO2/blob/dev/Framework/PhysicsSupport/include/Framework/O2DatabasePDGPlugin.h#L35C21-L35C42). Thank you.

PWGHF/TableProducer/candidateCreatorBplus.cxx
```diff
@@ -331,10 +349,9 @@ struct HfCandidateCreatorBplusExpressions {
 
   void processMc(aod::HfCand2Prong const& dzero,
                  aod::TracksWMc const& tracks,
-                 aod::McParticles const& mcParticles)
+                 aod::McParticles const& mcParticles,
+                 aod::HfCandBplusProngs const& candsBplus)
```
mfaggin, line 353: Linked to the comment above, I'm just wondering if you decided to implement this because it behaves better than `rowCandidateBPlus` (better CPU performance, code readability, etc.)

PWGHF/TableProducer/candidateCreatorBplus.cxx
```diff
@@ -42,7 +43,8 @@ using namespace o2::aod::hf_cand_bplus;
 
 /// Reconstruction of B± → D0bar(D0) π± → (K± π∓) π±
 struct HfCandidateCreatorBplus {
-  Produces<aod::HfCandBplusBase> rowCandidateBase;
+  Produces<aod::HfCandBplusBase> rowCandidateBase;     // table defined in CandidateReconstructionTables.h
+  Produces<aod::HfCandBplusProngs> rowCandidateProngs; // table defined in CandidateReconstructionTables.h
```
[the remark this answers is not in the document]
AlexBigO: The initial motivation for this "splitting" of the candidate table into 2 tables (i.e. extracting the prongs indices in a new table) was to have a common table (the Base one) for both Standard and Reduced workflows, while the table with prongs can be different for Standard and Reduced workflows (as the table the index columns are pointing to different tables).

Actually, thanks to [this kind of line](https://github.com/AliceO2Group/O2Physics/pull/3401/files#diff-bc149e5241d965436914a3a7c144b71bcceb435b5bd18cf7887925e11151355dR752), we subscribe to the joined tables in selectors and analysis tasks.

PWGHF/TableProducer/candidateCreatorBplus.cxx
AlexBigO: As, in this `processMc`, we only want to access the prongs of the B0 candidate, the information we need is in the `HfCandBplusProngs` table only and not in `HfCandBplusExt` table spawned just above, so we need to subscribe to this "prongs table".

mfaggin: ok the point is that you you need only indices, therefore no need to query the full table with topological variables. Thank you.

PWGHF/TableProducer/candidateCreatorBplus.cxx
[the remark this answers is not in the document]
mfaggin: If I get it properly, the thing is that in this `candidateCreatorBplus.cxx` the candidate D0 daughter can be queried only from the `aod::HfCand2Prong` when  `candsDPerCollision` is defined. Therefore, one should also save the `aod::HfCand2Prong` but this is not what you want (by definition, otherwise you would not have started this job). In addition, also the `Preslice` is t=done referring to the tables either of the normal collisions (here) or the reduced one (in the `candidateCreatorBplusReduced`), introducing the same storing issue als ofor the collisions.
Am I right?

PWGHF/D2H/TableProducer/candidateCreatorBplusReduced.cxx
```diff
@@ -45,19 +47,23 @@ struct HfCandidateCreatorBplusReduced {
...
+  double massD0{0.};
+  double massBplus{0.};
   double bz{0.};
 
   // Fitter for B vertex (2-prong vertex filter)
   o2::vertexing::DCAFitterN<2> df2;
 
-  Preslice<aod::HfCand2ProngReduced> candsDPerCollision = hf_track_index_reduced::hfReducedCollisionId;
-  Preslice<aod::HfTracksReduced> tracksPionPerCollision = hf_track_index_reduced::hfReducedCollisionId;
+  Preslice<soa::Join<aod::HfRed2Prongs, aod::HfRed2ProngsCov>> candsDPerCollision = hf_track_index_reduced::hfRedCollisionId;
```
mfaggin, line 65: In the `candidateCreatorBplus.cxx` there D-pi candidates are filtered accounting for the response of the `candidateSelectorD0` (see here: https://github.com/AliceO2Group/O2Physics/blob/c8f778a304ee9723d1d05a95dd47a240fded21e6/PWGHF/TableProducer/candidateCreatorBplus.cxx#L96) while in your case you are not doing it. This is because you already do the D0 selections in `dataCreatorD0PiReduced`, right?

PWGHF/D2H/TableProducer/candidateCreatorB0Reduced.cxx
```diff
@@ -44,19 +46,23 @@ struct HfCandidateCreatorB0Reduced {
...
+  double massD{0.};
+  double massB0{0.};
   double bz{0.};
 
   // Fitter for B vertex (2-prong vertex filter)
   o2::vertexing::DCAFitterN<2> df2;
 
-  Preslice<aod::HfCand3ProngReduced> candsDPerCollision = hf_track_index_reduced::hfReducedCollisionId;
-  Preslice<aod::HfTracksReduced> tracksPionPerCollision = hf_track_index_reduced::hfReducedCollisionId;
+  Preslice<soa::Join<aod::HfRed3Prongs, aod::HfRed3ProngsCov>> candsDPerCollision = hf_track_index_reduced::hfRedCollisionId;
```
mfaggin, line 64: In the `candidateCreatorB0.cxx`` there D-pi candidates are filtered accounting for the response of the candidateSelectorD0 (see here: https://github.com/AliceO2Group/O2Physics/blob/c8f778a304ee9723d1d05a95dd47a240fded21e6/PWGHF/TableProducer/candidateCreatorB0.cxx#L90)
while while in your case you are not doing it. This is because you already do the D0 selections in dataCreatorDplusPiReduced, right?

PWGHF/TableProducer/candidateCreatorB0.cxx
```diff
@@ -40,6 +41,7 @@ using namespace o2::framework::expressions;
 /// Reconstruction of B0 candidates
 struct HfCandidateCreatorB0 {
   Produces<aod::HfCandB0Base> rowCandidateBase; // table defined in CandidateReconstructionTables.h
+  Produces<aod::HfCandB0Prongs> rowCandidateProngs; // table defined in CandidateReconstructionTables.h
```
mfaggin, line 44: see the discussion done at line 47 in `candidateCreatorBplus.cxx`

mfaggin (commented):
Hi @fgrosa , @AlexBigO and @apalasciano , thanks a lot for your huge work.
I tried to go through the several files and I put here and there some comments. Most of them are for my personal understanding, hoping they can be useful for a further thoughts on the code.

Thank you.
Mattia

PWGHF/D2H/TableProducer/candidateSelectorBplusToD0PiReduced.cxx
```diff
@@ -89,10 +91,10 @@ struct HfCandidateSelectorBplusToD0PiReduced {
     }
   }
 
-  void process(HfCandBplus const& hfCandBs,
-               HfCand2ProngReduced const&,
-               HfTracksPidReduced const&,
-               HfCandBpConfig const& configs)
+  void process(HfRedCandBplus const& hfCandBs,
+               HfRed2Prongs const&,
```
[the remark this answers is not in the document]
AlexBigO: Indeed, we don't subscribe to the equivalent table for B0. We don't need it anymore.

PWGHF/D2H/TableProducer/candidateCreatorBplusReduced.cxx
AlexBigO: Exactly, this part of the Standard creator is already implemented at the DPi pair creation level of the Split workflow.

PWGHF/D2H/TableProducer/candidateCreatorB0Reduced.cxx
AlexBigO: Yes, same answer as for Bplus.

PWGHF/D2H/TableProducer/candidateSelectorBplusToD0PiReduced.cxx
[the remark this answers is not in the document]
mfaggin: thanks @AlexBigO for the confirmation. Then please remove it in your next commit.

PWGHF/TableProducer/candidateCreatorBplus.cxx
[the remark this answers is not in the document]
AlexBigO: Yes but that's not the whole story (sadly for us ^^ that's why we had to fix the current version).

There are 2 different points:
1. As you said, we can't store the whole candidate tables (B0, Bplus) so we needed to define our own with minimal number of columns.
*Doing so, everything worked (the current version on the master branch is functional but only locally, without using `o2-aod-merger`).*
2. The second point is tricky. With the current version on the master branch, our reduced AODs for B0, Bplus candidates contain index columns pointing towards Standard 2/3-prong candidates tables. As we don't store these Standard tables, the `o2-aod-merger` was looking for indices to tables that are not there. The solution we went with consists in simply extracting these *culprit* column indices in dedicated tables and create a Standard version of these tables (with columns pointing to Standard tables) and a Reduced version of these tables (with columns pointing to Reduced tables). This fix contains the whole "self-contained" aspect of this new version of the workflow.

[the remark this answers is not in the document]
mfaggin: What you say reflects (and explains better wrt my comment) what I understood. Thanks again.

AlexBigO:
Thanks @mfaggin ! We took your comments into account in the latest commit.

PWGHF/D2H/TableProducer/candidateSelectorBplusToD0PiReduced.cxx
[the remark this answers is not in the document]
AlexBigO: I removed it (see [here](https://github.com/AliceO2Group/O2Physics/blob/05c010e7e2a3ccf61ee3ed4c2ed09fb554e8e8bf/PWGHF/D2H/TableProducer/candidateSelectorBplusToD0PiReduced.cxx#L94)) but it does not appear on this review comment (I guess it is because this comment is from a previous version of the branch).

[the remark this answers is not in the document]
mfaggin: sorry I overlooked it. Fine for me then! Thanks a lot.

mfaggin (dismissed):

PWGHF/TableProducer/candidateCreatorBplus.cxx
```diff
@@ -344,10 +361,12 @@ struct HfCandidateCreatorBplusExpressions {
 
     // Match reconstructed candidates.
     // Spawned table can be used directly
-    for (const auto& candidate : *rowCandidateBPlus) {
+    for (const auto& candidate : candsBplus) {
+      // Printf("New rec. candidate");
```
vkucera, line 365:
```suggestion
```

PWGHF/TableProducer/candidateCreatorB0.cxx
```diff
@@ -343,8 +360,8 @@ struct HfCandidateCreatorB0Expressions {
     int8_t debug = 0;
 
     // Match reconstructed candidates.
-    // Spawned table can be used directly
-    for (const auto& candidate : *rowCandidateB0) {
+    for (const auto& candidate : candsB0) {
+      // Printf("New rec. candidate");
```
vkucera, line 364:
```suggestion
```

vkucera (changes requested):
Hi @fgrosa @AlexBigO , thanks for the huge work. I have no objections against the data model changes and I guess the D2H part has been heavily tested already. Thanks also to @mfaggin for the detailed review.
I have just two points:
- Please remove the `Printf` comments. If you want to keep some comments for debugging, please use the `LOG` macros with the `debug` level.
- In my draft PR that removes `RecoDecay::getMassPDG` I replace the calls with hard-coded `constexpr` masses from a header. I see that you added the PDG service and the `pdg->Mass` calls. Are you fine if I replace those and remove the service completely?

fgrosa:
> Hi @fgrosa @AlexBigO , thanks for the huge work. I have no objections against the data model changes and I guess the D2H part has been heavily tested already. Thanks also to @mfaggin for the detailed review. I have just two points:
> 
>     * Please remove the `Printf` comments. If you want to keep some comments for debugging, please use the `LOG` macros with the `debug` level.
> 
>     * In my draft PR that removes `RecoDecay::getMassPDG` I replace the calls with hard-coded `constexpr` masses from a header. I see that you added the PDG service and the `pdg->Mass` calls. Are you fine if I replace those and remove the service completely?

Hi @vkucera thanks for the feedback!
We are fine in replacing the call to the PDG service with the masses from the header. Do you have an idea of when it is supposed to be merged? Should we wait or merge this PR and implement the change of the masses in a second one?

vkucera (approved):

vkucera:
> > Hi @fgrosa @AlexBigO , thanks for the huge work. I have no objections against the data model changes and I guess the D2H part has been heavily tested already. Thanks also to @mfaggin for the detailed review. I have just two points:
> > ```
> > * Please remove the `Printf` comments. If you want to keep some comments for debugging, please use the `LOG` macros with the `debug` level.
> > 
> > * In my draft PR that removes `RecoDecay::getMassPDG` I replace the calls with hard-coded `constexpr` masses from a header. I see that you added the PDG service and the `pdg->Mass` calls. Are you fine if I replace those and remove the service completely?
> > ```
> 
> Hi @vkucera thanks for the feedback! We are fine in replacing the call to the PDG service with the masses from the header. Do you have an idea of when it is supposed to be merged? Should we wait or merge this PR and implement the change of the masses in a second one?

It would probably be easier to merge the PDG PR first (can be done today) and propagate the changes into this one, if that's fine with you.

PR #3409, opened 2023-08-30
add BC converter

ddobrigk:


alibuild (approved):

ddobrigk:
This change is connected to https://github.com/AliceO2Group/AliceO2/pull/11681 and allows for the conversion of BCs_000 to BCs_001. It is an intermediate step in the addition of CTP inputs to the data model, after which a final step in O2 is required.

PR #3412, opened 2023-08-31
[COMMON] updates on event-plane reconstruction

jikim1290:
updates on ep reco.


jikim1290:
Please approve this PR if available

ddobrigk (approved):
thanks!

ddobrigk:
Ah wait - I think I approved too quickly... @jikim1290 , are you sure about this bit: `if (am != 0)` - seems like `am` is a `float` and you are comparing floats without a tolerance (e.g. `1e-8`). _Usually_ this is ok with zero but you should _really_ still add a tolerance whenever comparing floats. Can you please still change that? Also, please check for other occurrences of float comparison without tolerances as there seem to be a few others. Thank you!

jikim1290:
@ddobrigk Thanks for comment, I have added the tolerance.

ddobrigk (approved):

PR #3420, opened 2023-09-01
PWGHF: Fix issue with the debug histogram for BDT in the D0 selector

lucamicheletti93:
Moving the BDT debug histogram


vkucera (approved):
Thanks @lucamicheletti93

PR #3429, opened 2023-09-04
Temporarily disable the ITS ramp rejection

mpuccio:


alibuild (approved):

PR #3434, opened 2023-09-05
PWGHF: Improve MC matching in Bs workflow

pstahlhu:
@fgrosa @fcatalan92 Add B0 matching for generated MC particles. Fix: Fill B0 bin of QA histogram in Bs-task.


fcatalan92 (dismissed):
LGTM

PWGHF/D2H/Tasks/taskBs.cxx
```diff
@@ -249,12 +250,12 @@ struct HfTaskBs {
...
-          if (TESTBIT(flagMcMatchRecBs, DecayTypeMc::BsToDsPiToKKPiPi)) { // Bs → Ds- π+ → (K- K+ π-) π+
-            registry.fill(HIST("hDecayTypeMc"), 1 + DecayTypeMc::BsToDsPiToKKPiPi, invMassCandBs, ptCandBs);
+          if (TESTBIT(flagMcMatchRecBs, DecayTypeMc::B0ToDsPiToKKPiPi)) { // B0(bar) → Ds± π∓ → (K- K+ π±) π∓
+            registry.fill(HIST("hDecayTypeMc"), 1 + DecayTypeMc::B0ToDsPiToKKPiPi, invMassCandBs, ptCandBs);
           } else if (TESTBIT(flagMcMatchRecBs, DecayTypeMc::PartlyRecoDecay)) { // Partly reconstructed decay channel
             registry.fill(HIST("hDecayTypeMc"), 1 + DecayTypeMc::PartlyRecoDecay, invMassCandBs, ptCandBs);
-          } else {
-            registry.fill(HIST("hDecayTypeMc"), 1 + DecayTypeMc::OtherDecay, invMassCandBs, ptCandBs);
+          } else { // Other decay
+            registry.fill(HIST("hDecayTypeMc"), 1 + DecayTypeMc::NDecayTypeMc, invMassCandBs, ptCandBs);
```
vkucera, line 258: This change breaks the semantics of the `enum` names. You fill a bin labelled "Other decays" so it makes more sense to fill it with a value of the corresponding name, i.e. `DecayTypeMc::OtherDecay`.
If you don't need `NDecayTypeMc`, remove it.

PWGHF/D2H/Tasks/taskBs.cxx
```diff
@@ -124,11 +124,12 @@ struct HfTaskBs {
     registry.add("hYGenWithProngsInAcceptance", "MC particles (generated-daughters in acceptance);B^{0}_{s} candidate #it{y}^{gen};entries", {HistType::kTH2F, {{100, -2., 2.}, axisPt}});
 
     if (checkDecayTypeMc) {
-      constexpr uint8_t kNBinsDecayTypeMc = DecayTypeMc::NDecayTypeMc;
+      constexpr uint8_t kNBinsDecayTypeMc = DecayTypeMc::NDecayTypeMc + 1;
       TString labels[kNBinsDecayTypeMc];
       labels[DecayTypeMc::BsToDsPiToKKPiPi] = "B^{0}_{s} #rightarrow (D^{-}_{s} #rightarrow K^{-} K^{+} #pi^{-}) #pi^{+}";
+      labels[DecayTypeMc::B0ToDsPiToKKPiPi] = "B^{0} #rightarrow (D^{+}_{s} #rightarrow K^{-} K^{+} #pi^{+}) #pi^{-}";
```
vkucera, line 130: `+` -> `#plus`
`-` -> `#minus`
will print the proper characters

vkucera (changes requested):

PWGHF/D2H/Tasks/taskBs.cxx
pstahlhu: Thanks for your comments @vkucera! I thought having `DecayTypeMc::OtherDecay` in the `enum` might be a bit confusing with regards to the tree creator: The flag was never set to `DecayTypeMc::OtherDecay`, so all the candidates which are neither B0 or Bs nor a partly reconstructed decay kept the flag =0 value. At first glance, it might be unclear what these candidates stored in the trees are exactly since no candidates appeared to fulfill the "Other decay" category. I considered setting the flag to `DecayTypeMc::OtherDecay` when all previous cases weren't fulfilled, but I thought that it might be unnecessary use of computing power, so I decided to remove `DecayTypeMc::OtherDecay` overall. Maybe you could comment whether it is worth setting the flag to `DecayTypeMc::OtherDecay`.

PWGHF/D2H/Tasks/taskBs.cxx
```diff
@@ -249,12 +250,12 @@ struct HfTaskBs {
         registry.fill(HIST("hChi2PCARecBg"), candidate.chi2PCA(), ptCandBs);
 
         if (checkDecayTypeMc) {
-          if (TESTBIT(flagMcMatchRecBs, DecayTypeMc::BsToDsPiToKKPiPi)) { // Bs → Ds- π+ → (K- K+ π-) π+
-            registry.fill(HIST("hDecayTypeMc"), 1 + DecayTypeMc::BsToDsPiToKKPiPi, invMassCandBs, ptCandBs);
+          if (TESTBIT(flagMcMatchRecBs, DecayTypeMc::B0ToDsPiToKKPiPi)) { // B0(bar) → Ds± π∓ → (K- K+ π±) π∓
+            registry.fill(HIST("hDecayTypeMc"), 1 + DecayTypeMc::B0ToDsPiToKKPiPi, invMassCandBs, ptCandBs);
```
vkucera, line 254: Why did you remove filling the `BsToDsPiToKKPiPi` cases?

PWGHF/D2H/Tasks/taskBs.cxx
vkucera: It's fine to keep the flag = 0 for those candidates. You don't fill the "Other decay" bin based on the flag anyway.
OK, then, if the value `OtherDecay` is used only for the histogram bins and not anywhere else, I think it's fine to remove it.

PWGHF/D2H/Tasks/taskBs.cxx
vkucera: OK, I see.

vkucera (approved):

PR #3444, opened 2023-09-10
Changes in PWGLF/TableProducer/cascadebuilder.cxx

lhusova:
Implementation of pt dep cosPA cut


ddobrigk:
Thanks @lhusova, this looks ok! I only have one question: are you happy with a cut that is of the type `B/pT` and not `A+B/pT`? I always assumed it safer to have a nonzero `A` so that we do not risk over-tightening the cut at high pT... but it depends on the outcome your studies, of course. Thanks!

mbombara (approved):

PR #3451, opened 2023-09-12
Avoid crash with no collisions

mpuccio:
@ddobrigk have a look if it makes sense to you


chiarapinto (approved):

ddobrigk:
> @ddobrigk have a look if it makes sense to you

Looks reasonable, thanks a lot! Will merge! I might also do the same to the cascade builder for completeness...

ddobrigk:
...correction to myself: I think the cascade builder actually doesn't need the fix because it is a loop that will not be entered in case `collisions.size()` is zero. OK, good!

PR #3453, opened 2023-09-12
Fix MC and Event Mixing

fmazzasc:


PWGLF/TableProducer/lithium4analysis.cxx
```diff
@@ -384,9 +437,23 @@ struct lithium4analysis {
...
 
-              if (!FillCandidateInfo(track1, track2, false)) {
+              if (!FillCandidateInfo(track1, track2, false, true)) {
                 continue;
               }
+
+              // fill TOF info outside to avoide responseBeta crash
+              auto& cand = l4Candidates.back();
+              if (track1.hasTOF()) {
+                float beta = responseBetaMC.GetBeta(track1);
```
njacazio, line 447: Ciao Francesco, thanks a lot! Is this because you want to avoid computing beta for all tracks?

fmazzasc: Ciao Nicolò, yes!

njacazio: maybe we should make it a dynamic column, thx!

njacazio (approved):

PR #3457, opened 2023-09-14
PWGJE: Task producing jet tracking qa histograms

alicecaluisi:
Hello,
I have written this code for performing track quality assurance in the jet framework. In order to do so, I also modified the CMakeLists. 
Here @jloemker and anybody who has suggestions, please let me know!

### DIFF
--- a/PWGJE/Tasks/CMakeLists.txt
+++ b/PWGJE/Tasks/CMakeLists.txt
@@ -84,4 +84,8 @@ if(FastJet_FOUND)
                         SOURCES jetTutorialSkeleton.cxx
                         PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::PWGJECore O2Physics::AnalysisCore
                         COMPONENT_NAME Analysis)
+    o2physics_add_dpl_workflow(track-jet-qa
+                        SOURCES trackJetqa.cxx
+                        PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::PWGJECore O2Physics::AnalysisCore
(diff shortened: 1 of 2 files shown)


nzardosh:
Dear Alice,

Thank you for your commit. The task you have setup seems not to use anything from the jet framework (you are just running over tracks and collisions, but not jets). In this case is this (the JE folder) the best place for this task and do other tasks similar already exist in other folders? It seems maybe the common folder where the trackqa task is in might be better?

Thanks,
Nima

nzardosh:
Hi Both, 

Ok I see. My question arose because you are currently making no jet related selections on the tracks you look at. If you will make some jet related cuts the question would be whether this task can be added as a process function to the JetQA task we already have or if its better as a separate task? We just want to avoid having many duplicate task but if you have aa strong preference we can keep this separate.
Another thing I would like to let you know is that we are making an update to the jet framework which will mean that the framework will produce its own version of the track table (we will discuss this in full next week at the O2 session of the PWGJE meeting) which will then be consumed by the jet finder. A downside of this is that some of the track information you currently are accessing will not be available in the jet framework as easily as before and if you add jet related cuts you will need to potentially rewrite the logic of your processes. If you prefer to wait till Tuesday after the discussions to send this PR through it might be good but if you want to start hyperloop productions before I can merge it now.

Thanks,
Nima

alicecaluisi:
Hello Nima,

It would be great if you could merge it now, so that I can start running on hyperloop as soon as possible!

Thank you!

nzardosh (approved):

PR #3469, opened 2023-09-17
Add checks for null pointer and increase the range of histogram

abmodak:


PWGMM/Mult/Tasks/heavy-ion-mult.cxx
```diff
@@ -66,7 +66,7 @@ AxisSpec axisEvent{4, -0.5, 3.5, "#Event"};
 AxisSpec axisVtxZ{800, -20, 20, "Vertex Z"};
 AxisSpec axisDCA = {601, -3.01, 3.01};
 AxisSpec axisPT = {1000, -0.05, 49.95};
-AxisSpec axisMult{2000, -0.5, 1999.5, "Multiplicity"};
+AxisSpec axisMult{2000, -0.5, 9999.5, "Multiplicity"};
```
aalkin, line 69: This would make bins with weird centers and widths. I recommend using `{10001, -0.5, 10000.5}` if you really want this to reach these numbers. You can also use `ConfigurableAxis` instead, so that you would not need to make PR or recompile each time you want to change the binning.

abmodak: Hi Anton, thank you for your suggestion. Now it is done.

aalkin (approved):

PR #3492, opened 2023-09-21
PWGJE: added correlation of MCP and MCD jets, ...

joonsukbae:
commented hfcandidates() histograms which is not working

### DIFF
--- a/PWGJE/Tasks/jetfinderQA.cxx
+++ b/PWGJE/Tasks/jetfinderQA.cxx
@@ -51,6 +51,7 @@ struct JetFinderQATask {
   Configurable<float> trackPtMin{"trackPtMin", 0.15, "minimum pT acceptance for tracks"};
   Configurable<float> trackPtMax{"trackPtMax", 100.0, "maximum pT acceptance for tracks"};
   Configurable<std::string> trackSelections{"trackSelections", "globalTracks", "set track selections"};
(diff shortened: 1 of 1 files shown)


joonsukbae:
Hi Nima, thank you for the comments.

I think those histogram fill same things but there's a need for rebinning due to the low statistics in higher pT especially for the general purposed MC or maybe even for Jet-Jet enhanced MC such as LHC23d4. I couldn't checked LHC23d4 because this code in HY is not running so that I have to check it.
Rebinning 1D in macro after analysis is easy, but as I know it's not possible and not recommended for 2D or more degree histograms.
Or, how about changing "h3_jet_r_jet_pt_part_jet_pt" to THnSparse? I think it'll be difficult to see the entries and result briefly, but THnSparse can be rebinned after analysis.

Thanks,
Joonsuk

nzardosh:
Hi Joonsuk,

Thanks for the explanation. But why can the TH2 or TH3 objects not be rebinned? I think they can be. There are the Rebin3D() or RebinX() etc functions which you can use or you can also project the histograms and then rebin them. I think for histograms with less than 4 dimensions a THnSparse should not be needed and the current implementation is useable.

Thanks,
Nima

joonsukbae:
I see. I'll look for the method that can be implemented in the macro. Thank you for the information.

PR #3496, opened 2023-09-21
PWGHF: Add McCollisionId to the event table of treeCreatorLcToPKPi

DelloStritto:
Adding McCollisionId to the event table to match it with the MC particle table.
@qgp @vkucera

### DIFF
--- a/PWGHF/TableProducer/treeCreatorLcToPKPi.cxx
+++ b/PWGHF/TableProducer/treeCreatorLcToPKPi.cxx
@@ -341,7 +341,7 @@ struct HfTreeCreatorLcToPKPi {
     for (const auto& collision : collisions) {
       rowCandidateFullEvents(
         collision.globalIndex(),
-        0,
+        -1,
         collision.numContrib(),
         collision.posX(),
         collision.posY(),
--- a/PWGHF/TableProducer/treeCreatorLcToPKPi.cxx
+++ b/PWGHF/TableProducer/treeCreatorLcToPKPi.cxx
@@ -205,9 +205,6 @@ struct HfTreeCreatorLcToPKPi {
     // Filling event properties
     rowCandidateFullEvents.reserve(collisions.size());
     for (const auto& collision : collisions) {
-      if (!collision.has_mcCollision()) {
-        continue;
-      }
       rowCandidateFullEvents(
         collision.globalIndex(),
         collision.mcCollisionId(),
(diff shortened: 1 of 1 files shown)


PWGHF/TableProducer/treeCreatorLcToPKPi.cxx
```diff
@@ -80,6 +80,7 @@ DECLARE_SOA_COLUMN(OriginMcGen, originMcGen, int8_t);
 DECLARE_SOA_COLUMN(IsCandidateSwapped, isCandidateSwapped, int8_t);
 DECLARE_SOA_INDEX_COLUMN_FULL(Candidate, candidate, int, HfCand3Prong, "_0");
 // Events
+DECLARE_SOA_INDEX_COLUMN(McCollision, mccollision);
```
vkucera, line 83:
```suggestion
DECLARE_SOA_INDEX_COLUMN(McCollision, mcCollision);
```

PWGHF/TableProducer/treeCreatorLcToPKPi.cxx
```diff
@@ -203,8 +205,13 @@ struct HfTreeCreatorLcToPKPi {
     // Filling event properties
     rowCandidateFullEvents.reserve(collisions.size());
     for (const auto& collision : collisions) {
+      auto mcCollisionId = 0;
+      if (collision.has_mcCollision()) {
+        mcCollisionId = collision.mcCollisionId();
+      }
```
vkucera, line 211: Do we need to save the collisions that do not have a matching MC collision?

vkucera (changes requested):

vkucera (dismissed):
Thanks @DelloStritto 
@qgp What do you think?

PWGHF/TableProducer/treeCreatorLcToPKPi.cxx
```diff
@@ -203,8 +205,12 @@ struct HfTreeCreatorLcToPKPi {
     // Filling event properties
     rowCandidateFullEvents.reserve(collisions.size());
     for (const auto& collision : collisions) {
+      if (!collision.has_mcCollision()) {
+        continue;
```
qgp, line 209: @DelloStritto Do we really want to silently drop these? Wouldn't it make sense to keep them and store -1 for the MC collision ID?

qgp:
Looks good to me. Only question about silently dropping collisions without an MC collision (see inline comment.

PWGHF/TableProducer/treeCreatorLcToPKPi.cxx
vkucera: @qgp Where would we use these collisions without a matching reconstruction level?

vkucera (approved):

PR #3500, opened 2023-09-22
Fix non-identified track histo. filling in case of no PID.

mfaggin:


alibuild (approved):

PR #3503, opened 2023-09-22
PWGHF: Expression columns to compute combined PID variables

AlexBigO:
Hi @vkucera !

In the end I defined a new `struct` before the main one in the selector, to `Spawns` the PID tables. I aim at using these tables in `prong0_as()` command, and I don't know if there is a way to both `Spawns` and consume a table in a `struct` (as I understand it, it is not possible), so spawning them *a priori* in a different `struct` is the only solution I found for the moment.

I added the columns and tables definitions in `CandidateReconstructionTables.h`, and put the possibility to compute combined PID variables also for the tiny/non-full versions of the PID tables.

### DIFF
--- a/PWGHF/DataModel/CandidateReconstructionTables.h
+++ b/PWGHF/DataModel/CandidateReconstructionTables.h
@@ -33,7 +33,6 @@
 #include "PWGHF/Core/SelectorCuts.h"
 
 using namespace o2::analysis;
-using namespace o2::framework::expressions;
 
 namespace o2::aod
 {
@@ -61,10 +60,13 @@ using TracksPidTinyPi = soa::Join<aod::pidTPCPi, aod::pidTOFPi>;
 using TracksPidTinyKa = soa::Join<aod::pidTPCKa, aod::pidTOFKa>;
 using TracksPidTinyPr = soa::Join<aod::pidTPCPr, aod::pidTOFPr>;
 
-namespace pidtpctof_utils
+namespace pid_tpc_tof_utils
 {
-// Function to combine TPC and TOF NSigma (for ML purposes)
-template <bool tiny = false, typename T1 = Node, typename T2>
+/// Function to combine TPC and TOF NSigma (for ML purposes)
+/// \param tpcNSigma is the (binned) NSigma separation in TPC (if tiny = true)
+/// \param tofNSigma is the (binned) NSigma separation in TOF (if tiny = true)
+/// \return Node containing the combined NSigma of TPC and TOF
+template <bool tiny = false, typename T1 = o2::framework::expressions::Node, typename T2>
 T1 combineNSigma(const T2& tpcNSigma, const T2& tofNSigma)
 {
   float defaultNSigmaTolerance = .1f;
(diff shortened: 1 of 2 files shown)


PWGHF/TableProducer/candidateSelectorDplusToPiKPi.cxx
```diff
@@ -214,6 +220,8 @@ struct HfCandidateSelectorDplusToPiKPi {
       auto trackNeg = candidate.prong1_as<TracksSel>();  // negative daughter (positive for the antiparticles)
       auto trackPos2 = candidate.prong2_as<TracksSel>(); // positive daughter (negative for the antiparticles)
 
+      LOG(info) << "Combined PID = " << trackPos1.tpctofNSigmaPi();
```
AlexBigO, line 223: I just noticed I forgot to remove this, I will do it when I push the next commit with the PR comments.

vkucera:
Hi @AlexBigO , thanks for your work.

Spawning and consuming a table within the same struct is possible with `bindExternalIndices`, as done with the candidate tables here: https://github.com/AliceO2Group/O2Physics/blob/master/PWGHF/TableProducer/candidateCreator2Prong.cxx#L246

PWGHF/DataModel/CandidateReconstructionTables.h
```diff
@@ -54,6 +55,92 @@ using TracksPidPi = soa::Join<aod::pidTPCFullPi, aod::pidTOFFullPi>;
 using TracksPidKa = soa::Join<aod::pidTPCFullKa, aod::pidTOFFullKa>;
 using TracksPidPr = soa::Join<aod::pidTPCFullPr, aod::pidTOFFullPr>;
 
+using TracksPidTinyEl = soa::Join<aod::pidTPCEl, aod::pidTOFEl>;
+using TracksPidTinyMu = soa::Join<aod::pidTPCMu, aod::pidTOFMu>;
+using TracksPidTinyPi = soa::Join<aod::pidTPCPi, aod::pidTOFPi>;
+using TracksPidTinyKa = soa::Join<aod::pidTPCKa, aod::pidTOFKa>;
+using TracksPidTinyPr = soa::Join<aod::pidTPCPr, aod::pidTOFPr>;
+
+namespace pidtpctof_utils
```
vkucera, line 64:
```suggestion
namespace pid_tpc_tof_utils
```

PWGHF/DataModel/CandidateReconstructionTables.h
```diff
@@ -54,6 +55,92 @@ using TracksPidPi = soa::Join<aod::pidTPCFullPi, aod::pidTOFFullPi>;
...
+{
+// Function to combine TPC and TOF NSigma (for ML purposes)
+template <bool tiny = false, typename T1 = Node, typename T2>
+T1 combineNSigma(const T2& tpcNSigma, const T2& tofNSigma)
```
vkucera, line 68: Please make it a Doxygen style documentation.

PWGHF/DataModel/CandidateReconstructionTables.h
```diff
@@ -54,6 +55,92 @@ using TracksPidPi = soa::Join<aod::pidTPCFullPi, aod::pidTOFFullPi>;
...
+                ifnode(tpcNSigma > defaultNSigma, nabs(tpcNSigma),                                                                            // only TPC
+                       ifnode(tofNSigma > defaultNSigma, nabs(tofNSigma),                                                                     // only TOF
+                              1.f * tofNSigma)));                                                                                             // no TPC nor TOF
+}
+} // namespace pidtpctof_utils
+
+namespace pidtpctof
+{
+// Combined TPC and TOF NSigma
+DECLARE_SOA_EXPRESSION_COLUMN(TPCTOFNSigmaEl, tpctofNSigmaEl, //! Combined Nsigma separation with the TPC & TOF detectors for electron
```
vkucera, line 93:
```suggestion
DECLARE_SOA_EXPRESSION_COLUMN(TpcTofNSigmaEl, tpcTofNSigmaEl, //! Combined NSigma separation with the TPC & TOF detectors for electron
```

vkucera (changes requested):

AlexBigO:
> Hi @AlexBigO , thanks for your work.
> 
> Spawning and consuming a table within the same struct is possible with `bindExternalIndices`, as done with the candidate tables here: https://github.com/AliceO2Group/O2Physics/blob/master/PWGHF/TableProducer/candidateCreator2Prong.cxx#L246

Hi @vkucera ! Thanks for the comments. Actually, I tried but could not find the right syntax for what I want to do. When using `rowTable -> bindExtrernalIndices(&tracks)` (where we have `Tracks const& tracks` in `process`), one can then loop over `Table` via `for (const auto& row : *rowTable)`. However, I don't want to do that here, I would like to access `Table` via en `INDEX_COLUMN` with a command line such as `.prong0_as<Table>`. Would you know how to do it?

PWGHF/DataModel/CandidateReconstructionTables.h
```diff
@@ -33,6 +33,7 @@
 #include "PWGHF/Core/SelectorCuts.h"
 
 using namespace o2::analysis;
+using namespace o2::framework::expressions;
```
vkucera, line 36: No using directive in headers.
(The one above is yet to be removed.)

vkucera (changes requested):

PWGHF/DataModel/CandidateReconstructionTables.h
```diff
@@ -54,6 +54,95 @@ using TracksPidPi = soa::Join<aod::pidTPCFullPi, aod::pidTOFFullPi>;
...
+using TracksPidTinyPr = soa::Join<aod::pidTPCPr, aod::pidTOFPr>;
+
+namespace pid_tpc_tof_utils
+{
+/// Function to combine TPC and TOF NSigma (for ML purposes)
+/// \param tpcNSigma is the (binned) NSigma separation in TPC (if tiny = true)
+/// \param tofNSigma is the (binned) NSigma separation in TOF (if tiny = true)
+/// \return Node containing the combined NSigma of TPC and TOF
+template <bool tiny = false, typename T1 = o2::framework::expressions::Node, typename T2>
+T1 combineNSigma(const T2& tpcNSigma, const T2& tofNSigma)
```
vkucera, line 70: What is the motivation for having the return type templated?

vkucera:
Thanks @AlexBigO , it looks good to me. I just have a curious question on the templated return type. Otherwise am happy with merging unless others have further comments.

PWGHF/DataModel/CandidateReconstructionTables.h
AlexBigO: Actually, it is a remnant of my tests during development (because I initially wrote this method to combine NSigma in a dynamic column in the reduced data model). I can remove it and directly impose the type as `o2::framework::expressions::Node` if you prefer.

vkucera: If there is currently no use case for having the template, it is indeed preferable to return an explicit type. Thanks

vkucera (approved):

PR #3508, opened 2023-09-26
Adding more axes for monitoring vs. time.

mfaggin:


alibuild (approved):

miranov25:
Hello Mattia.

maybe better would be 1/pt instead of the pt. Like that, we will group based on the same intrinsic resolution which is important

1/pt - (3,0,5)

Regards
Marian

mfaggin:
> Hello Mattia.
> 
> maybe better would be 1/pt instead of the pt. Like that, we will group based on the same intrinsic resolution which is important
> 
> 1/pt - (3,0,5)
> 
> Regards Marian

Hi @miranov25 I have implemented 2 new axis in the sparse, 1 for 1/pt and 1 for the charge (see here: https://github.com/AliceO2Group/O2Physics/pull/3513). If you can suggest a reasonable binning for 1/pt we can change it at runtime. Thank you.

miranov25:
Hello @mfaggin 

I'll prepare the presentation for tomorrow. You can rely on me, we should also have an analysis of your data. Today I will send the presentation to the tpc group to comment on it and correct wordings. If you add me to agenda I prefer to make a link in pull request to the slides

As for the new version of the time series, I have 2 preliminary slides describing what to add and with which granularity, maybe you can add them to the pull request.

I can answer possible questions on mattermost.  For other variables we can reiterate.


### Granularity (nbins,min,max)
* Time 10 ms
* Eta (6,-1.0,1.0)
* qPt(4,0.1,5)

### Variables:
* <TPC →  ITS>  matching efficiency
* <TPC number of cluster>,  chi2?
* <ITS number of clusters>,  chi2?
* DCAr,DCAz for tracks with ITS and without ITS (for |DCAr|<6)
* Other variables sensitive to distortion bias and efficiency drops

PR #3514, opened 2023-09-27
Add TPC and ITS clusters for time monitoring.

mfaggin:


miranov25:
We can reduce memory usage as we can use 80, or 40 bins

mfaggin:
> We can reduce memory usage as we can use 80, or 40 bins

done in the new commit

alibuild (approved):

PR #3526, opened 2023-09-29
Hotfix: Move spheoricty function from CollCuts

BongHwi:
It has to be placed in the main task.
Add a configurable variable to bypass this calculation.


alibuild (approved):

PR #3529, opened 2023-09-29
[EMCAL-733] Fill amplitude histogram for high an low gain separately

mfasDa:


alibuild (approved):
Auto-approving on behalf of @mfasDa.

PR #3536, opened 2023-10-02
[PWGJE] Produce derived table for tree analysis

jloemker:
Hi @alicecaluisi, @mverwe and all,

this PR contains all the files and changes required for a fast tree analysis. This is required in the Jetframework as we want to test customized track cuts for a good pT resolution at high pT. It is a very first attempt and we will for sure add some more information in the following weeks.

Here are run and config files for local checks (https://cernbox.cern.ch/s/wEp4WermJZuikAl).
The **TrackJetQa.h** defines the data model where we declare the content of our columns.
The **jettrackderived.cxx** produces the table for the derived collisions and tracks
And the **trackJetqa.cxx** is the task from @alicecaluisi that has now an additional process function for the derived table - here you need to modify a few more things:

1) Extend the processDerived to fill all track and event properties.
Try to use templates (or functions) where you perform your collision/track checks and fill all the histograms via functions that are outside of the process(). We don't want to duplicate the code from process to process, but we want to use the same building blocks for the standard and derived process. Right now I am only filling the vertex histograms and no tracks in the processDerived. As you can see, I was lazy and cp'ed your code from the standard process. Continuing like this will increase the file length without any use - you should try to fill the same histograms in the derived process as in the standard with as little amount of lines as possible. Therefore templates !

2) Add more track checks: I suggest to add a couple of manual checks for the values you select for eta, pt, phi and so on. You did add a configurable for the trackselections but you never use it in the code.

3) I noticed that you added manual some histograms in eta vs phi in certain pt ranges. To shorten your code you might want to consider THnSparse where you can fill multiple axis and this get the correlation of eta and phi for every possible pt bin.

*) Once the derived process function is complete we will put this on hyperloop with very loose cuts and then you can apply your personal track cut variations to the derived tables = optimize the uncertainty at for high pT tracks.

Cheerio,
Johanna


jloemker:
There is no meaningful log message for the build/O2Physcis/o2/macOS.
I believe this is a github issue as I already produced results with this code locally.

@nzardosh , can we merge this ?

nzardosh (approved):

PR #3544, opened 2023-10-04
[PWG-LF] Replace the binning definition to the process

BongHwi:
The current binning definition is placed out of the process, so It was not updated based on the variable axis.
This caused a failure in mixing when we didn't have a multiplicity calibration.


alibuild (dismissed):

alibuild (approved):

PR #3549, opened 2023-10-05
[Commom] Add task for calculations of event plane resolution

jikim1290:
Add task for calculations of event plane resolution

### DIFF
--- a/Common/Tasks/CMakeLists.txt
+++ b/Common/Tasks/CMakeLists.txt
@@ -73,3 +73,8 @@ o2physics_add_dpl_workflow(qvectors-correction
                     SOURCES qVectorsCorrection.cxx
                     PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::AnalysisCore
                     COMPONENT_NAME Analysis)
+
+o2physics_add_dpl_workflow(evtplanes-resolution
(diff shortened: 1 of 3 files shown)


jgrosseo:
@ddobrigk Should this really be in common?

jikim1290:
@jgrosseo This work is originated from DPG service work:
https://alice-glance.cern.ch/service-work/tasks/609

The event plane method can be used for so various PWGs that I decided locate them in common.
In AliPhysics, the same task is in AliPhysics/PWG/FLOW (let me know if wrong).

If you think the current place is not proper, we can move these codes.

jgrosseo:
> @jgrosseo This work is originated from DPG service work: https://alice-glance.cern.ch/service-work/tasks/609
> 
> The event plane method can be used for so various PWGs that I decided locate them in common. In AliPhysics, the same task is in AliPhysics/PWG/FLOW (let me know if wrong).
> 
> If you think the current place is not proper, we can move these codes.

Fine with me

ddobrigk:
Looks ok to me. I also think that common tools for centrality calculation etc are also under ./Common so the location is fine. Thanks!

ddobrigk (approved):

PR #3559, opened 2023-10-06
Include configurable to use generated pt/eta for acceptance selection

hscheid:
@ddobrigk this should be what we discussed. could you have a look?


alibuild (approved):

PR #3563, opened 2023-10-07
Bugfix: fix use of T instead of V for V0A

ddobrigk:


alibuild (approved):

PR #3576, opened 2023-10-09
PWGDQ: update cuts and histograms for dalitz selection

glegras:


PWGDQ/Core/VarManager.h
```diff
@@ -1256,9 +1257,9 @@ void VarManager::FillPair(T1 const& t1, T2 const& t2, float* values)
   values[kDeltaPtotTracks] = Ptot1 - Ptot2;
 
   if (fgUsedVars[kPsiPair]) {
-    values[kDeltaPhiPair] = (t1.sign() > 0) ? (v1.Phi() - v2.Phi()) : (v2.Phi() - v1.Phi());
+    values[kDeltaPhiPair] = (t1.sign() * values[kMagField] > 0.) ? (v1.Phi() - v2.Phi()) : (v2.Phi() - v1.Phi());
```
iarsene, line 1260: What happens if the kDeltaPhiPair and kPsiPair are used in a task other than the DalitzSelection ?
We need to have the magnetic field value initialized such that these quantities behave as before in case there is not magnetic field value from the CCDB. So I suggest that instead of using a VarManager variable, you create instead a static data member for the VarManager. And then, if that is initialized, you can use it in the calculation of these quantities, otherwise, you can initialize it with a neutral value (1). Then the user can handle this on his own (by using the knowledge of which are runs are positive or negative polarity)

glegras: Hi Ionut, 
I am not sure I understand what is the difference between my and your proposal. In both case, it is assumed that the user gives a proper value to the magnetic field. 
If the magnetic field has not been initialized, then it takes a default value, it would behave the same if it is a VarManager variable or a static data member.

iarsene (approved):

PWGDQ/Core/VarManager.h
iarsene: Hi Gauthier,
The difference is that if we leave this as a regular VarManager variable, it may be initialized or overwritten in the analysis tasks to some value such as -999 by mistake and you would not know.

PR #3591, opened 2023-10-11
PWGMM: dndeta: Fix wrong comparison in filter

aalkin:


alibuild (approved):

PR #3592, opened 2023-10-11
Common: Adding a protection in the collisionAssociator task

DelloStritto:
Adding a protection for tracks associated with a 0-dimensional BC vector.

@fgrosa @jgrosseo

### DIFF
--- a/Common/Core/CollisionAssociation.h
+++ b/Common/Core/CollisionAssociation.h
@@ -133,6 +133,10 @@ class CollisionAssociation
         for (const auto& ambTrack : ambiguousTracks) {
           if constexpr (isCentralBarrel) { // FIXME: to be removed as soon as it is possible to use getId<Table>() for joined tables
             if (ambTrack.trackId() == track.globalIndex()) {
+              if(!ambTrack.has_bc() || ambTrack.bc().size() == 0) {
+                globalBC.push_back(-1);
+                break;
+              }
               globalBC.push_back(ambTrack.bc().begin().globalBC());
               break;
             }
@@ -163,6 +167,9 @@ class CollisionAssociation
         }
 
         float trackTime = track.trackTime();
+        if (globalBC[track.filteredIndex()] < 0 ) {
+          continue;
+        }
         const int64_t bcOffsetWindow = (int64_t)globalBC[track.filteredIndex()] + trackTime / o2::constants::lhc::LHCBunchSpacingNS - (int64_t)collBC;
         if (std::abs(bcOffsetWindow) > bOffsetMax) {
           continue;
--- a/Common/Core/CollisionAssociation.h
+++ b/Common/Core/CollisionAssociation.h
@@ -133,7 +133,7 @@ class CollisionAssociation
         for (const auto& ambTrack : ambiguousTracks) {
           if constexpr (isCentralBarrel) { // FIXME: to be removed as soon as it is possible to use getId<Table>() for joined tables
             if (ambTrack.trackId() == track.globalIndex()) {
-              if(!ambTrack.has_bc() || ambTrack.bc().size() == 0) {
+              if (!ambTrack.has_bc() || ambTrack.bc().size() == 0) {
                 globalBC.push_back(-1);
                 break;
               }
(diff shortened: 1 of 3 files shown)


Common/Core/CollisionAssociation.h
```diff
@@ -133,6 +133,10 @@ class CollisionAssociation
         for (const auto& ambTrack : ambiguousTracks) {
           if constexpr (isCentralBarrel) { // FIXME: to be removed as soon as it is possible to use getId<Table>() for joined tables
             if (ambTrack.trackId() == track.globalIndex()) {
+              if (!ambTrack.has_bc() || ambTrack.bc().size() == 0) {
```
aalkin, line 136: This is redundant, for the slice index `has_` method checks if both start and end indices are non-negative. The size of the valid slice can not be less than 1.

fgrosa: Hi @aalkin thanks for the feedback! We checked with Luigi one file where the previous code was crashing, and the `ambTrack.has_bc()` condition was not sufficient, we had to add the second one.

DelloStritto: Hi Anton,
with the condition `!ambTrack.has_bc()` we did not manage to reject the track which was making the task crash while we did it with the requirement `ambTrack.bc().size() == 0`, so I am not sure they are fully equivalent.

Do you suggest to remove the condition `!ambTrack.has_bc()`?

Thanks a lot for the review,
Luigi

aalkin: The getter for the slice index looks like this
```cpp
    template <typename T>
    auto _Getter_##_as() const
    {
      if (O2_BUILTIN_UNLIKELY(mBinding == nullptr)) {
        o2::soa::notBoundTable(#_Table_);
      }
      if (O2_BUILTIN_UNLIKELY(!has_##_Getter_())) {
        return static_cast<T const*>(mBinding)->emptySlice();
      }
      auto a = *mColumnIterator;
      auto t = static_cast<T const*>(mBinding)->rawSlice(a[0], a[1]);
      static_cast<T const*>(mBinding)->copyIndexBindings(t);
      return t;
    } 
```
Size = 0 can only happen if `has_` condition is already false, or the data is somehow malformed. Probably there is an entry where start < end?

jgrosseo: a[0] cannot be equal to a[1]?

jgrosseo: We could provide the file to Anton if we can show that the framework does not correctly catch this case...

aalkin: It is assumed that `start >= end`, otherwise the index does not make sense. If the file that causing this is known, just printing out the index values (that one can get with `bcIds()` ) for the cases where `has_bc()` is true but size is 0 is enough to investigate this. By all means, merge this workaround for now, but I think it is clear that if AODs contain incorrectly set indices, it should be addressed.

aalkin: > a[0] cannot be less than a[1]?

I'm very sorry, that was typo. I've edited the original comment now.

fgrosa: @jgrosseo @aalkin it might indeed be that the AO2Ds that we are analysing have some bug with the indices since it happens only in not too recent MC samples (produced before the change in the ambiguous track table in AO2Ds by Jan for instance). We can indeed check the `bcIds()` in the pathological case (we did not remember that it was possible to get them) to help understanding the issue and possibly implement a way to catch it in the framework.
However, since these are the only available MC productions with HF it would be good to have a protection merged so that we can start producing the derived data for HF analyses. Thanks a lot!

DelloStritto: Hi @aalkin, I checked the AmbTrack `bcIds` and the result is that **bcIds()[0] is equal to bcIds()[1]** for the pathological track.
From my understanding, this is what we should expect when `ambTrack.bc().size() == 0`, or am I wrong?
If so, it seems that for some reason the `!ambTrack.has_bc()` is not rejecting AmbTrack with 0 bc size

aalkin: This is unexpected, the slice with start=end should be size 1. I'll try to reproduce this with the test.

fgrosa: @DelloStritto @aalkin thanks for the checks! @jgrosseo can we in the meantime merge this PR? Based on the last messages, I would say that we should also keep the check on the size to properly catch the pathological tracks in the existing MC productions. Thanks!

jgrosseo (approved):

Common/Core/CollisionAssociation.h
jgrosseo: I am not against merging this. @aalkin Let us know when you have news...

DelloStritto: Ok, I can share the file I am testing with you, thanks @aalkin

PR #3594, opened 2023-10-11
Additional 3d histograms (optional) and low k* cut

gmantzar:


gmantzar:
Changes:

- One additional histogram (and we plan to add more, that's why we introduced also configurable to activate them, because they'll consume much memory and are not necessary for every analysis)
- An optional cut for low kstar

victor-gonzalez (approved):
I approve your changes because they are correct but if you foresee you will have more conditionals in your pair loop I would suggest you using an `enum` for the different conditions and then template your functions accordingly

PR #3609, opened 2023-10-13
First code version for three-body task

lauraser:
The first version of the code extends the two-body femtoscopy to the three-body case. Besides the three-body task, the femtoDreamContainer for three-body case was created separately as it requires different histograms. We will consider merging the containers once the code is validated and finalised. 
This version of the code works only for three identical track selections, it will be extended as the next step to any three particles (once this code is validated on the hyperloop to provide reasonable results).

### DIFF
--- a/PWGCF/FemtoDream/CMakeLists.txt
+++ b/PWGCF/FemtoDream/CMakeLists.txt
@@ -24,6 +24,11 @@ o2physics_add_dpl_workflow(femtodream-pair-track-track
           PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::AnalysisCore
           COMPONENT_NAME Analysis)
 
+o2physics_add_dpl_workflow(femtodream-pair-track-track-track
+          SOURCES femtoDreamPairTaskTrackTrackTrack.cxx
+          PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::AnalysisCore
+          COMPONENT_NAME Analysis)
+
 o2physics_add_dpl_workflow(femtodream-debug-track
           SOURCES femtoDreamDebugTrack.cxx
           PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::AnalysisCore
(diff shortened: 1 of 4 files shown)


PWGCF/FemtoDream/CMakeLists.txt
```diff
@@ -24,6 +24,11 @@ o2physics_add_dpl_workflow(femtodream-pair-track-track
           PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::AnalysisCore
           COMPONENT_NAME Analysis)
 
+o2physics_add_dpl_workflow(femtodream-pair-track-track-track
+          SOURCES femtoDreamPairTaskTrackTrackTrack.cxx
+          PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::AnalysisCore
+          COMPONENT_NAME Analysis)
```
victor-gonzalez, line 30: For the next iteration think on giving it the `triplet-track-track-track` name

PWGCF/FemtoDream/femtoDreamPairTaskTrackTrackTrack.cxx
```diff
@@ -0,0 +1,421 @@
...
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+/// \file femtoDreamPairTaskTrackTrackTrack.cxx
```
victor-gonzalez, line 12: See my above comment about the use of `triplet` instead of `pair` in the filename

victor-gonzalez (approved):
I approve for the time being but, please, consider the suggested name change

alibuild (approved):

PR #3610, opened 2023-10-13
PWGCF: FemtoUniverse -- Fixing CPR and track cleaning

zchochul:
Fixing the logic and uncommenting CPR and track cleaning for FemtoUniverse.


alibuild (approved):

PR #3616, opened 2023-10-14
PWGHF: Implement efficiency computation macro for D2H

AlexBigO:
Hello @fcatalan92 @mfaggin !

Following our discussion via email, here is a proposition of macro for efficiency computation.


PWGHF/D2H/Macros/efficiency.py
```diff
@@ -0,0 +1,389 @@
...
+    Returns
+    ----------
+    - h_eff: histogram containing teff information
+    """
+
+    n_bins = h_eff.GetXaxis().GetNbins()
+    for i_bin in range(1, n_bins+1):
+        h_eff.SetBinContent(i_bin, teff.GetEfficiency(i_bin))
+        # TH1 can't handle asymmetric errors so we take the max
+        err_max = max(teff.GetEfficiencyErrorLow(i_bin), teff.GetEfficiencyErrorUp(i_bin))
```
mfaggin, line 161: I'm wondering in which case it can happen that the efficiency has an asymmetric error. If I am not wrong, the binomial uncertainty is always symmetric right?
Just to be sure that I'm not overlooking anything, the code looks fine.

PWGHF/D2H/Macros/efficiency.py
```diff
@@ -0,0 +1,389 @@
...
+    latex.SetTextFont(42)
+    watermark = cfg['output']['plots']['watermark']
+
+    # open output file
+    out_file = TFile(out_dir + out_label + '.root', 'recreate')
+
+    # compute efficiencies
+    efficiencies = {}
+    is_retrieved_pt_interval = False
+    for (key, h_rec), (_, h_gen) in zip(dic_rec.items(), dic_gen.items()):
```
mfaggin, line 295: What about adding a couple of protections/warning here, checking wheter the cases
`h_rec is None`
and
`h_gen is None`
are satisfied?

mfaggin (commented):
thanks a lot @AlexBigO , good that this macro can become available for everybody! For me it looks fine, but I leave the time to @fcatalan92 to have a look.

I just have 1 question for my understanding, see the dedicated comment.
Best regards.

Mattia

PWGHF/D2H/Macros/efficiency.py
AlexBigO: Hi Mattia, thanks for the reply. Indeed, I must say I could not think of a case where this would happen, but I put it as extra safety after reading about it in a ROOT forum. But, in our use case, I don't think that we truly need it, if I am not misssing something.

mfaggin: > Hi Mattia, thanks for the reply. Indeed, I must say I could not think of a case where this would happen, but I put it as extra safety after reading about it in a ROOT forum. But, in our use case, I don't think that we truly need it, if I am not misssing something.

Hi @AlexBigO , exactly! This is my point.
What about adding a warning when this happens? Something like:

`if abs(teff.GetEfficiencyErrorLow(i_bin) - teff.GetEfficiencyErrorUp(i_bin))/teff.GetEfficiencyErrorLow(i_bin) > 1e-05:
    print a warning`

Something similar to this. What do you think?

AlexBigO: Ah, this is better indeed! Thanks, I implemented it, and will push it along with future comments.

mfaggin: thank you!

AlexBigO:
Some comments on my last commit:
- I fixed a bug: I had forgotten to adapt the histogram projections along X or Y axis to the axis of rapidity set in the config file
- I applied Mattia's suggestion
- I wanted to add the possibility to rebin input histograms, it can be useful in case pT-binning is not the same for ML model training/optimisation and for the subsequent analysis. Using `.Scale(1., 'width')` after rebinning caused issues with `TEfficiency`, so I tried overcoming this issue with the trick described in the code comments.

PWGHF/D2H/Macros/efficiency.py
```diff
@@ -127,13 +129,29 @@ def compute_efficiency(
...
             h_gen = h_gen.ProjectionY()
 
+    # rebin histograms, if enabled
+    if pt_bins_limits is not None:
+        n_bins = len(pt_bins_limits) - 1
+
+        h_rec = h_rec.Rebin(n_bins, 'hRec', pt_bins_limits)
+        # using Scale causes weights-related complications in TEfficiency
+        # trick: we SetBinContent + Sumw2 (gives same result as w/o rebinning)
+        h_rec.SetBinContent(1, h_rec.GetBinContent(1) / h_rec.GetBinWidth(1))
```
mfaggin: probably I'm overlooking something, but why are you touching here only the bin 1? You have just rebinned your `h_rec` and in principle you'd like to normalize all the bins, right? The same for the `h_gen` below

mfaggin:
It looks good to me. @fcatalan92 can you have a look as well?

AlexBigO:
Sorry for the many attempts to solve the MegaLinter issues, the `typing` caused tricky errors, but it passed the tests, eventually :)

PWGHF/D2H/Macros/efficiency.py
```diff
@@ -0,0 +1,448 @@
...
+    return path
+
+
+# pylint: disable=too-many-arguments
+def configure_canvas(
+    name_canvas: str,
+    pt_min: Union[int, float],
+    min_y_axis: Union[int, float],
+    pt_max: Union[int, float],
+    max_y_axis: Union[int, float],
```
vkucera, line 104:
```suggestion
    pt_min: Union[int, float],
    y_axis_min: Union[int, float],
    pt_max: Union[int, float],
    y_axis_max: Union[int, float],
```
More consistent to have the attribute as a suffix.

PWGHF/D2H/Macros/config_efficiency.yml
```diff
@@ -0,0 +1,37 @@
...
+  axis: Y # rapidity axis in input histograms, options available: X, Y
+  cut: 0.5 # selection |y| < cut on the generated histogram, null if not cut
+
+output:
+  dir: efficiency/
+  save:
+    TEfficiency: True # save TEfficiency instances in output .root file
+    TH1: True # save TH1 histograms in output .root file
+    TCanvas: # save TCanvas instances in extra files (on top of the standard output .root file)
+      individual: True # for inididual canvas (All, Prompt, Nonprompt separately)
```
fcatalan92, line 26:
```suggestion
      individual: True # for individual canvas (All, Prompt, Nonprompt separately)
```

fcatalan92 (commented):
Hi @AlexBigO, @mfaggin, all looks good to me. I just found a small typo.

PWGHF/D2H/Macros/config_efficiency.yml
AlexBigO: Hi @fcatalan92 , thanks! :)

fcatalan92 (approved):

PR #3619, opened 2023-10-16
PWGJE: Added jet area for QA tasks

joonsukbae:
referred to Fig. 28 in https://arxiv.org/abs/2202.01548 which are pp 13 TeV results analyzed by Yongzhen Hou


nzardosh (changes requested):
could you please change A to area in the axis title?

joonsukbae:
> could you please change A to area in the axis title?

Sure, I fixed it as you said. I saw it also used as #it{A}_{jet}. Which one do you think reduce confusion much?

nzardosh:
> > could you please change A to area in the axis title?
> 
> Sure, I fixed it as you said. I saw it also used as #it{A}_{jet}. Which one do you think reduce confusion much?

yes thats a good idea :)

nzardosh (approved):

PR #3621, opened 2023-10-16
[MCHistograms] Properly skip collision without mcCollision

benedikt-voelkel:


benedikt-voelkel:
@ddobrigk This should be straightforward to merge, right?
Let me know if there is anything else to do/refine.
In particular in PbPb I see sometimes
`[21244:access-mc-truth]: [12:10:32][FATAL] Unhandled o2::framework::runtime_error reached the top of main of o2-analysistutorial-mc-histograms, device shutting down. Reason: Accessing invalid index for mcCollision`

ddobrigk (approved):

PR #3629, opened 2023-10-16
PWGHF: Lc-H correlator and task development

zz951:
Lc-Hadron correlator and task development, which Lc is for the decay channel Lc->PKPi

### DIFF
--- a/PWGHF/HFC/TableProducer/correlatorLcHadrons.cxx
+++ b/PWGHF/HFC/TableProducer/correlatorLcHadrons.cxx
@@ -1,0 +1,700 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+/// \file correlatorLcHadrons.cxx
+/// \brief Lc-Hadrons correlator task - data-like, Mc-Reco and Mc-Gen analyses
+///
+/// \author Marianna Mazzilli <marianna.mazzilli@cern.ch>
+/// \author Zhen Zhang <zhenz@cern.ch>
+
+#include "Framework/AnalysisTask.h"
+#include "Framework/HistogramRegistry.h"
+#include "Framework/runDataProcessing.h"
+
+#include "Common/Core/TrackSelection.h"
+#include "Common/DataModel/Centrality.h"
+#include "Common/DataModel/EventSelection.h"
+#include "Common/DataModel/Multiplicity.h"
+#include "Common/DataModel/TrackSelectionTables.h"
+
+#include "PWGHF/Core/HfHelper.h"
+#include "PWGHF/DataModel/CandidateReconstructionTables.h"
+#include "PWGHF/DataModel/CandidateSelectionTables.h"
+#include "PWGHF/HFC/DataModel/CorrelationTables.h"
+
+using namespace o2;
+using namespace o2::analysis;
+using namespace o2::framework;
+using namespace o2::framework::expressions;
+
+///
+/// Returns deltaPhi values in range [-pi/2., 3.*pi/2.], typically used for correlation studies
+///
+double getDeltaPhi(double phiLc, double phiHadron)
+{
+  return RecoDecay::constrainAngle(phiHadron - phiLc, -o2::constants::math::PIHalf);
+}
+
+/// definition of variables for Lc hadron pairs (in data-like, Mc-reco and Mc-kine tasks)
+const int nBinsPtMassAndEfficiency = o2::analysis::hf_cuts_lc_to_p_k_pi::nBinsPt;
+const double efficiencyLcDefault[nBinsPtMassAndEfficiency] = {};
+auto vecEfficiencyLc = std::vector<double>{efficiencyLcDefault, efficiencyLcDefault + nBinsPtMassAndEfficiency};
+
+// histogram binning definition
+const int massAxisBins = 120;
+const double massAxisMin = 1.98;
+const double massAxisMax = 2.58;
+const int phiAxisBins = 32;
+const double phiAxisMin = -o2::constants::math::PIHalf;
+const double phiAxisMax = 3. * o2::constants::math::PIHalf;
+const int yAxisBins = 100;
+const double yAxisMin = -2.;
+const double yAxisMax = 2.;
+const int ptLcAxisBins = 180;
+const double ptLcAxisMin = 0.;
+const double ptLcAxisMax = 36.;
+
+// definition of ME variables
+std::vector<double> zBins{VARIABLE_WIDTH, -10.0, -2.5, 2.5, 10.0};
+std::vector<double> multBins{VARIABLE_WIDTH, 0., 200., 500., 5000.};
+std::vector<double> multBinsMcGen{VARIABLE_WIDTH, 0., 20., 50., 500.}; // In McGen multiplicity is defined by counting primaries
+using BinningType = ColumnBinningPolicy<aod::collision::PosZ, aod::mult::MultFV0M<aod::mult::MultFV0A, aod::mult::MultFV0C>>;
+BinningType corrBinning{{zBins, multBins}, true};
+
+using SelectedCollisions = soa::Filtered<soa::Join<aod::Collisions, aod::Mults, aod::LcSelection>>;
+using SelectedTracks = soa::Filtered<aod::TracksWDca>;
+using SelectedCandidatesData = soa::Filtered<soa::Join<aod::HfCand3Prong, aod::HfSelLc>>;
+using SelectedCandidatesMcRec = soa::Filtered<soa::Join<aod::HfCand3Prong, aod::HfSelLc, aod::HfCand3ProngMcRec>>;
+using SelectedCollisionsMcGen = soa::Filtered<soa::Join<aod::McCollisions, aod::LcSelection>>;
+using SelectedTracksMcGen = soa::Filtered<aod::McParticles>;
+
+// Code to select collisions with at least one Lambda_c
+struct HfCorrelatorLcHadronsSelection {
+  Produces<aod::LcSelection> lcSel;
+
+  Configurable<int> selectionFlagLc{"selectionFlagLc", 1, "Selection Flag for Lc"};
+  Configurable<float> yCandMax{"yCandMax", 0.8, "max. cand. rapidity"};
+  Configurable<float> ptCandMin{"ptCandMin", 1., "min. cand. pT"};
+
+  HfHelper hfHelper;
+  SliceCache cache;
+
+  Partition<soa::Join<aod::HfCand3Prong, aod::HfSelLc>> selectedLcCandidates = aod::hf_sel_candidate_lc::isSelLcToPKPi >= selectionFlagLc || aod::hf_sel_candidate_lc::isSelLcToPiKP >= selectionFlagLc;
+  Partition<soa::Join<aod::HfCand3Prong, aod::HfSelLc, aod::HfCand3ProngMcRec>> selectedLcCandidatesMc = aod::hf_sel_candidate_lc::isSelLcToPKPi >= selectionFlagLc || aod::hf_sel_candidate_lc::isSelLcToPiKP >= selectionFlagLc;
+
+  void processLcSelectionData(aod::Collision const& collision,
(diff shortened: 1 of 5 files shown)


vkucera (commented):
Hi @zz951 , some general comments:
- Add a `\note` field in the documentation of files that are based on existing workflows.
- Fix the order of struct members. You now declare many of them in between functions.
- Fix naming, especially capitalisation and remove underscores.
- Remove the prefix `o2::framework::` from `InitContext`, you are importing the namespace.
- Your process functions are set for MC rec. analysis by default. Is this intended?

PWGHF/DataModel/CandidateReconstructionTables.h
```diff
@@ -1025,6 +1025,28 @@ DECLARE_SOA_TABLE(DHadronRecoInfo, "AOD", "DHADRONRECOINFO", //! D0-Hadrons pair
                   aod::hf_correlation_d0_hadron::MDbar,
                   aod::hf_correlation_d0_hadron::SignalStatus);
 
+// definition of columns and tables for Lc-Hadron correlation pairs
+namespace hf_correlation_lc_hadron
+{
```
vkucera, line 1030: Please move your data model definitions into `PWGHF/HFC/DataModel/CorrelationTables.h`

zz951: Ok, they are move into `PWGHF/HFC/DataModel/CorrelationTables.h` now, thanks a lot!

fgrosa:
@mmazzilli @deepathoms could you please have a look since you are more expert of HF-hadron correlations? Thanks!

PWGHF/HFC/Tasks/taskCorrelationLcHadrons.cxx
```diff
@@ -0,0 +1,271 @@
...
+const TString stringPtHadron = "#it{p}_{T}^{Hadron} (GeV/#it{c});";
+const TString stringPoolBin = "poolBin;";
+const TString stringDeltaEta = "#it{#eta}^{Hadron}-#it{#eta}^{#Lambda_c};";
+const TString stringDeltaPhi = "#it{#varphi}^{Hadron}-#it{#varphi}^{#Lambda_c} (rad);";
+const TString stringLcHadron = "#Lambda_c,Hadron candidates ";
+const TString stringSignal = "signal region;";
+const TString stringSideband = "sidebands;";
+const TString stringMcParticles = "Mc gen - #Lambda_c,Hadron particles;";
+const TString stringMcReco = "Mc reco - #Lambda_c,Hadron candidates ";
+
```
deepathoms, line 57: Add a comment for the objects defined below.

PWGHF/HFC/Tasks/taskCorrelationLcHadrons.cxx
```diff
@@ -0,0 +1,271 @@
...
+const int nPtBinsCorrelations = 8;
+const double pTBinsCorrelations[nPtBinsCorrelations + 1] = {0., 2., 4., 6., 8., 12., 16., 24., 99.};
+auto vecBinsPtCorrelations = std::vector<double>{pTBinsCorrelations, pTBinsCorrelations + nPtBinsCorrelations + 1};
+const double signalRegionInnerDefault[nPtBinsCorrelations] = {2.269, 2.269, 2.269, 2.269, 2.269, 2.269, 2.269, 2.269};
```
deepathoms, line 61: Is it a standard practice to use 'inner' and 'outer' instead of maybe 'low' and 'high' mass region? The reason I ask is in the signal region, the lower value 2.269 is called inner region, but in the left SB region, the lower value 2.209 is called outer. the usage is clear for the SB region, but for the signal region, it seems arbitrary. Ignore my comment if this is a standard practice.

zz951: I think it should be a standard way, for they are used in dplus-h, ds-h and d0-h in the same way.

PWGHF/HFC/Tasks/taskCorrelationLcHadrons.cxx
zz951: ok, thanks a lot!

PWGHF/HFC/TableProducer/correlatorLcHadrons.cxx
```diff
@@ -0,0 +1,700 @@
...
+        }
+        if (yCandMax >= 0. && std::abs(hfHelper.yLc(candidate)) > yCandMax) {
+          continue;
+        }
+        if (ptCandMin >= 0. && candidate.pt() < ptCandMin) {
+          continue;
+        }
+        isLcFound = 1;
+        break;
+      }
```
fgrosa, line 137: This part of the code could be a template function to avoid code duplication (you have exactly the same in `processSelectionData()`)

PWGHF/HFC/TableProducer/correlatorLcHadrons.cxx
```diff
@@ -0,0 +1,700 @@
...
+  PROCESS_SWITCH(HfCorrelatorLcHadronsSelection, processLcSelectionMcRec, "Process Lc Selection McRec", false);
+
+  void processLcSelectionMcGen(aod::McCollision const& mcCollision,
+                               aod::McParticles const& mcParticles)
+  {
+    bool isLcFound = 0;
+    for (const auto& particle : mcParticles) {
+      if (std::abs(particle.pdgCode()) != pdg::Code::kLambdaCPlus) {
+        continue;
+      }
```
fgrosa, line 150: Is it intended to check only the PDG code and not the decay channel?

PWGHF/HFC/TableProducer/correlatorLcHadrons.cxx
```diff
@@ -0,0 +1,700 @@
...
+        }
+      } // end inner loop (Tracks)
+    }   // end outer Lc loop
+    registry.fill(HIST("hZvtx"), collision.posZ());
+    registry.fill(HIST("hMultV0M"), collision.multFV0M());
+  }
+  PROCESS_SWITCH(HfCorrelatorLcHadrons, processMcRec, "Process Mc Reco mode", false);
+
+  /// Lc-Hadron correlation pair builder - for Mc gen-level analysis
+  void processMcGen(aod::McCollision const& mcCollision,
```
fgrosa, line 493: What's the reason to keep MC gen and MC reco separated in two process functions? Do you need only one of the two in some cases?

PWGHF/HFC/TableProducer/correlatorLcHadrons.cxx
```diff
@@ -0,0 +1,700 @@
...
+        }
+        isLcFound = 1;
+        break;
+      }
+    }
+    lcSel(isLcFound);
+  }
+  PROCESS_SWITCH(HfCorrelatorLcHadronsSelection, processLcSelectionMcRec, "Process Lc Selection McRec", false);
+
+  void processLcSelectionMcGen(aod::McCollision const& mcCollision,
```
fgrosa, line 143: What's the reason to keep MC gen and MC reco separated in two process functions? Do you need only one of the two in some cases?

PWGHF/HFC/TableProducer/correlatorLcHadrons.cxx
```diff
@@ -0,0 +1,700 @@
...
+                            t2.pt(),
+                            poolBin);
+          entryLcHadronRecoInfo(hfHelper.invMassLcToPiKP(t1), false);
+        }
+      }
+    }
+  }
+  PROCESS_SWITCH(HfCorrelatorLcHadrons, processMcRecMixedEvent, "Process Mixed Event McRec", false);
+
+  void processMcGenMixedEvent(SelectedCollisionsMcGen const& collisions,
```
fgrosa, line 643: What's the reason to keep MC gen and MC reco separated in two process functions? Do you need only one of the two in some cases?

PWGHF/HFC/Tasks/taskCorrelationLcHadrons.cxx
```diff
@@ -0,0 +1,271 @@
...
+      {"hCorrel2DVsPtSidebandsMcRec", stringLcHadron + stringSideband + stringDeltaPhi + stringDeltaEta + stringPtLc + stringPtHadron + stringPoolBin + "entries", {HistType::kTHnSparseD, {{64, -o2::constants::math::PIHalf, 3. * o2::constants::math::PIHalf}, {200, -4., 4.}, {10, 0., 10.}, {11, 0., 11.}, {9, 0., 9.}}}},
+      {"hDeltaEtaPtIntMcGen", stringMcParticles + stringDeltaEta + "entries", {HistType::kTH1F, {{200, -4., 4.}}}},
+      {"hDeltaPhiPtIntMcGen", stringMcParticles + stringDeltaPhi + "entries", {HistType::kTH1F, {{64, -o2::constants::math::PIHalf, 3. * o2::constants::math::PIHalf}}}},
+      {"hCorrel2DPtIntMcGen", stringMcParticles + stringDeltaPhi + stringDeltaEta + "entries", {HistType::kTH2F, {{64, -o2::constants::math::PIHalf, 3. * o2::constants::math::PIHalf}, {200, -4., 4.}}}},
+      {"hCorrel2DVsPtMcGen", stringMcParticles + stringDeltaPhi + stringDeltaEta + stringPtLc + stringPoolBin + "entries", {HistType::kTHnSparseD, {{64, -o2::constants::math::PIHalf, 3. * o2::constants::math::PIHalf}, {200, -4., 4.}, {10, 0., 10.}, {11, 0., 11.}, {9, 0., 9.}}}}, // note: axes 3 and 4 (the Pt) are updated in the init()
```
fgrosa, line 118: You repeat the same axes many times (e.g. `{200, -4., -4.}` is repeated 15 times), you could consider to define some `AxisSpec` objects to simplify possible changes in the future.

PWGHF/HFC/Tasks/taskCorrelationLcHadrons.cxx
```diff
@@ -0,0 +1,271 @@
...
+        registry.fill(HIST("hCorrel2DPtIntSidebandsMcRec"), deltaPhi, deltaEta, efficiencyWeight);
+        registry.fill(HIST("hDeltaEtaPtIntSidebandsMcRec"), deltaEta, efficiencyWeight);
+        registry.fill(HIST("hDeltaPhiPtIntSidebandsMcRec"), deltaPhi, efficiencyWeight);
+      }
+    } // end loop
+  }
+  PROCESS_SWITCH(HfTaskCorrelationLcHadrons, processMcRec, "Process Mc Reco mode", false);
+
+  /// Lc-Hadron correlation pair filling task, from pair tables - for Mc gen-level analysis (no filter/selection, only true signal)
+  void processMcGen(aod::LcHadronPair const& pairEntries)
```
fgrosa, line 242: What's the reason to keep MC gen and MC reco separated in two process functions? Do you need only one of the two in some cases?

fgrosa (changes requested):
Thanks @zz951 for the implementation! I added few more comments, once implemented it can be merged for me.

PWGHF/HFC/TableProducer/correlatorLcHadrons.cxx
fcolamar: Ciao Fabrizio, yes, that selection is just used to understand which events has to be considered for the mixed-event, and there it's enough to have Lc particle in the event, not necessary that it decays in the specific decay channel used in the analysis

PWGHF/HFC/TableProducer/correlatorLcHadrons.cxx
fcolamar: Ciao Fabrizio, that is because, as also for the HFC D-h correlation codes, the McRec and McGen processes produce and fill the same tables (in particular aod::LcHadronPair), but with different information (gen or rec), so only one of them can be executed in the workflow

PWGHF/HFC/TableProducer/correlatorLcHadrons.cxx
fgrosa: @fcolamar thanks for the reply! Wouldn't it be more efficient to have 2 tables that one can fill in one go?

fcolamar: Yes that alternate approach would also work, but the original idea was to avoid having a proliferation of tables, also because the McGen is expected to be employed only for the MC closure check, so to be seldom used

PWGHF/HFC/Tasks/taskCorrelationLcHadrons.cxx
fcolamar: Yes this is not a super-ideal notation, but I would be consistent with the nomenclature of the other frameworks, Maybe at some point it's better to switch to low-high, but with a dedicated PR involving all the affected codes.

PWGHF/HFC/Tasks/taskCorrelationLcHadrons.cxx
zz951: Thank you a lot, I gave a new commit based on the comments!

fgrosa (approved):

PR #3636, opened 2023-10-17
LF: Update spectra with ITS and TPC only requirements

njacazio:


alibuild (approved):

PR #3639, opened 2023-10-18
PWGCF: Femtouniverse -- Fixing Phi mesons in MC 

zchochul:
Phi mesons in MC data are NOT primary particles, so for now, this is a workaround to have any results for MC (added as a comment in the file).


alibuild (approved):

PR #3641, opened 2023-10-18
Remove the PID selection flag and add HasTOF

BongHwi:
The PID Selection flag has been introduced for the filter-level selection but never used.

To have a simplest shape, I removed the selection flag and add the hasTOF variable.


alibuild (approved):

PR #3652, opened 2023-10-19
[EventFiltering, PWGHF] Fix typo and move to HF masses

fgrosa:


alibuild (dismissed):

PWGHF/Core/PDG.h
```diff
@@ -70,6 +70,7 @@ constexpr double MassPhi = 1.019455;
 constexpr double MassSigmaC0 = 2.45376;
 constexpr double MassSigmaCPlusPlus = 2.45402;
 constexpr double MassX3872 = 3.87165;
+constexpr double MassXiB0 = 5.7924;
```
vkucera, line 73: Please do not edit this block by hand. It is generated by the Python script. So the script is the place to be updated and then this header should be updated with the generated output.

vkucera (changes requested):

PWGHF/Core/PDG.h
fgrosa: Thanks! Sorry indeed it wasn't that clear to me how this worked, it should be fixed now

vkucera: No problem, if you have an idea, how to make the documentation clearer, let me know. Thanks for the update.

vkucera (approved):

PR #3656, opened 2023-10-20
Tools, PWGHF: Add flexibility to ML model input features definition

AlexBigO:
Hi @fcatalan92 , @fgrosa !

I moved as much methods as I could to the mother class and put the `enum` as a template typename to give more generality. Also, this way the macros work for any type`U`, and we avoid `static_cast<int>` as we had before.

From what I found, calling `getInputFeatures` does not create a copy of the vector (feature of modern C++ compiler) so I left it this way.

### DIFF
--- a/PWGHF/Core/HfMlResponseDplusToPiKPi.h
+++ b/PWGHF/Core/HfMlResponseDplusToPiKPi.h
@@ -22,6 +22,32 @@
 
 #include "PWGHF/Core/HfMlResponse.h"
 
+// Fill the map of available input features
+// the key is the feature's name (std::string)
+// the value is the corresponding value in EnumInputFeatures
+#define FILL_MAP_DPLUS(FEATURE)                                          \
+  {                                                                      \
+#FEATURE, static_cast < uint8_t>(InputFeaturesDplusToPiKPi::FEATURE) \
+  }
+
+// Check if the index of mCachedIndices (index associated to a FEATURE)
+// matches the entry in EnumInputFeatures associated to this FEATURE
+// if so, the inputFeatures vector is filled with the FEATURE's value
+// by calling the corresponding GETTER from OBJECT
+#define CHECK_AND_FILL_VEC_DPLUS_FULL(OBJECT, FEATURE, GETTER)     \
+  case static_cast<uint8_t>(InputFeaturesDplusToPiKPi::FEATURE): { \
+    inputFeatures.emplace_back(OBJECT.GETTER());                   \
+    break;                                                         \
+  }
+
+// Specific case of CHECK_AND_FILL_VEC_DPLUS_FULL(OBJECT, FEATURE, GETTER)
+// where OBJECT is named candidate and FEATURE = GETTER
+#define CHECK_AND_FILL_VEC_DPLUS(GETTER)                          \
+  case static_cast<uint8_t>(InputFeaturesDplusToPiKPi::GETTER): { \
+    inputFeatures.emplace_back(candidate.GETTER());               \
+    break;                                                        \
+  }
+
 namespace o2::analysis
 {
 
@@ -57,8 +83,8 @@ enum class InputFeaturesDplusToPiKPi : uint8_t {
   tpcTofNSigmaKa2
 };
 
-template <typename TypeOutputScore = float, class EnumInputFeatures = InputFeaturesDplusToPiKPi>
-class HfMlResponseDplusToPiKPi : public HfMlResponse<TypeOutputScore, EnumInputFeatures>
+template <typename TypeOutputScore = float>
+class HfMlResponseDplusToPiKPi : public HfMlResponse<TypeOutputScore>
 {
  public:
   /// Default constructor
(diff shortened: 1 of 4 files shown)


PWGHF/Core/HfMlResponseDplusToPiKPi.h
```diff
@@ -0,0 +1,120 @@
...
+namespace o2::analysis
+{
+
+enum InputFeaturesDplusToPiKPi : int8_t {
+  ptProng0 = 0,
+  ptProng1,
+  ptProng2,
+  impactParameter0,
+  impactParameter1,
+  impactParameter2,
```
fgrosa, line 34: Maybe we could specify `XY` on these and add those in the `Z` direction (this would also allow us to include the case of the HF triggers, for which we could move to this code as soon as stable)

PWGHF/Core/HfMlResponseDplusToPiKPi.h
```diff
@@ -0,0 +1,120 @@
...
+  decayLengthXYNormalised,
+  cpa,
+  cpaXY,
+  maxNormalisedDeltaIP,
+  tpcTofNSigmaPi0,
+  tpcTofNSigmaPi1,
+  tpcTofNSigmaPi2,
+  tpcTofNSigmaKa0,
+  tpcTofNSigmaKa1,
+  tpcTofNSigmaKa2
```
fgrosa, line 45: Maybe we could also add the possibility to have the Nsigma variables for TPC and TOF separately?

fgrosa (changes requested):
Hi @AlexBigO thanks a lot for the development and for providing a first use case for the D+ that can be used as an example for the other channels! I only have a couple of comments regarding the features included (minor because of course we can always update the list in the future).

PWGHF/Core/HfMlResponseDplusToPiKPi.h
AlexBigO: I may have missed something, but it looks like the `Z` component is not even defined in the [`HfCand3Prong` table](https://github.com/AliceO2Group/O2Physics/blob/master/PWGHF/DataModel/CandidateReconstructionTables.h#L770). In the [3prong candidate creator](https://github.com/AliceO2Group/O2Physics/blob/master/PWGHF/TableProducer/candidateCreator3Prong.cxx#L207), we only store the `XY` component via `getY` defined [here](github.com/AliceO2Group/AliceO2/blob/dev/DataFormats/Reconstruction/include/ReconstructionDataFormats/DCA.h#L57).

So, we should add the definition in the 3prong table, store `impactParameter*.getZ()` in 3 prong creator, so we can retrieve it in the Dplus selector :)

fgrosa: Indeed you are right, I missed it! 
Then let's do it in a second step, together with the change in the creator(s).

AlexBigO: But, addig `XY` in the enum entries can already be done, it is indeed clearer this way.

fgrosa (approved):

Tools/ML/MlResponse.h
```diff
@@ -113,6 +128,20 @@ class MlResponse
...
+  /// Method to translate configurable input features strings into integers
+  /// \param cfgInputFeatures array of input features names
+  void cacheInputFeaturesIndices(std::vector<std::string> const& cfgInputFeatures)
+  {
+    setAvailableInputFeatures();
+    for (const auto& inputFeature : cfgInputFeatures) {
+      if (mAvailableInputFeatures.count(inputFeature)) {
+        mCachedIndices.emplace_back(mAvailableInputFeatures[inputFeature]);
+      } else {
+        LOG(fatal) << "Input feature not available. Please check your configurables.";
```
vkucera, line 140: Report which feature is not available to make the debugging transparent.

fcatalan92 (commented):
Hi @AlexBigO, thanks a lot for the developments! I have a few suggestions, mostly minor.
The only major one is that I would not use the template for the map index, the eventual cast can be included in the macros you provide.

PWGHF/Core/HfMlResponseDplusToPiKPi.h
```diff
@@ -0,0 +1,162 @@
...
+#include <vector>
+
+#include "PWGHF/Core/HfMlResponse.h"
+
+namespace o2::analysis
+{
+
+enum InputFeaturesDplusToPiKPi : int8_t {
+  ptProng0 = 0,
+  ptProng1,
```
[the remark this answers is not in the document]
AlexBigO: Hi @vkucera , thanks for the comments.

No, we don't need it to be the same name as the getters. However, this `enum` will be as a dictionary: it will impose the naming convention to use in JSON.

Tools/ML/MlResponse.h
```diff
@@ -193,13 +222,16 @@ class MlResponse
...
-  std::vector<int> mCutDir = {};                  // direction of the cuts on the model scores (no cut is also supported)
-  o2::framework::LabeledArray<double> mCuts = {}; // array of cut values to apply on the model scores
+  std::vector<o2::ml::OnnxModel> mModels;              // OnnxModel objects, one for each bin
+  uint8_t mNModels = 1;                                // number of bins
+  uint8_t mNClasses = 3;                               // number of model classes
+  std::vector<double> mBinsLimits = {};                // bin limits of the variable (e.g. pT) used to select which model to use
+  std::vector<std::string> mPaths = {""};              // paths to the models, one for each bin
+  std::vector<int> mCutDir = {};                       // direction of the cuts on the model scores (no cut is also supported)
+  o2::framework::LabeledArray<double> mCuts = {};      // array of cut values to apply on the model scores
+  std::map<std::string, U> mAvailableInputFeatures;    // map of available input features
```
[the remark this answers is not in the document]
AlexBigO: Indeed, I just did it to avoid using `static_cast<uint8_t>` when filling `mAvailableFeatures` or comparing an `enum` entry and a `mCachedIndices` entry (in `getInputFeatures`). But I can add those and explicitely put the type `uint8_t`.

Tools/ML/MlResponse.h
```diff
@@ -28,6 +28,21 @@
...
+#define CHECK_AND_FILL_VEC(GETTER)                  \
+  case U::GETTER:                                   \
+    inputFeatures.emplace_back(candidate.GETTER()); \
+    break
+
+#define CHECK_AND_FILL_VEC_FULL(OBJECT, FEATURE, GETTER) \
+  case U::FEATURE:                                       \
+    inputFeatures.emplace_back(OBJECT.GETTER());         \
+    break
+
```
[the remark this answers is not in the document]
AlexBigO: I did not need `;` as I put it after calling the macro. I added it now that I also added braces.

PWGHF/TableProducer/candidateSelectorDplusToPiKPi.cxx
```diff
@@ -117,6 +118,7 @@ struct HfCandidateSelectorDplusToPiKPi {
         hfMlResponse.setModelPathsLocal(onnxFileNames);
       }
       hfMlResponse.init();
+      hfMlResponse.cacheInputFeaturesIndices(namesInputFeatures);
       outputMl.assign(((std::vector<int>)cutDirMl).size(), -1.f); // dummy value for ML output
```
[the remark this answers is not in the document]
AlexBigO: Sorry, what do you mean by keeping it empty?

Tools/ML/MlResponse.h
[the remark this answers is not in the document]
AlexBigO: Actually, doing this (and also documenting the `template` of the classes as adivsed by Vit), I realized that it might be better to change from `typename U = uint8_t` to `class EnumInputFeatures = uint8_t` and go from `enum` to `enum class` for the available input features.

This way, it makes it a bit more clear that this template argument expects a whole `enum` instance and not just a type (while keeping `typename` could me misleading). I tested it and it works, so just tell me @fcatalan92 , @vkucera , which version you prefer (here in the example of D+):
- `typename EnumInputFeatures = InputFeaturesDplusToPiKPi` with `enum InputFeaturesDplusToPiKPi`
- or `class EnumInputFeatures = InputFeaturesDplusToPiKPi` with `enum class InputFeaturesDplusToPiKPi`
?

PS: knowing that the default value of this `template` can still be set to `uint8_t`.

PWGHF/TableProducer/candidateSelectorDplusToPiKPi.cxx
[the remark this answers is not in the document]
fcatalan92: Hi @AlexBigO, with no entries instead of dummy values. The column that is filled is a variable-length array so can have different sizes for each entry and the vector could also be empty.

Tools/ML/MlResponse.h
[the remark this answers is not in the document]
AlexBigO: After discussion with Fabio, I implemented the second option with `enum class`.

PWGHF/TableProducer/candidateSelectorDplusToPiKPi.cxx
```diff
@@ -112,12 +114,14 @@ struct HfCandidateSelectorDplusToPiKPi {
...
-        hfMlResponse.setModelPathsCCDB(onnxFileNames, ccdbApi, modelPathsCCDB.value, timestampCCDB);
+        hfMlResponse.setModelPathsCCDB(onnxFileNames,
+                                       ccdbApi, modelPathsCCDB.value, timestampCCDB);
       } else {
         hfMlResponse.setModelPathsLocal(onnxFileNames);
       }
+      hfMlResponse.cacheInputFeaturesIndices(namesInputFeatures);
       hfMlResponse.init();
-      outputMl.assign(((std::vector<int>)cutDirMl).size(), -1.f); // dummy value for ML output
+      // outputMl.assign(((std::vector<int>)cutDirMl).size(), -1.f); // dummy value for ML output
```
fcatalan92, line 124:
```suggestion
```

PWGHF/TableProducer/candidateSelectorDplusToPiKPi.cxx
```diff
@@ -251,27 +255,13 @@ struct HfCandidateSelectorDplusToPiKPi {
...
-                                         trackPos2.tpcTofNSigmaPi(),
-                                         trackPos2.tpcTofNSigmaKa()};
-
+        std::vector<float> inputFeatures = hfMlResponse.getInputFeatures(candidate, trackPos1, trackNeg, trackPos2);
         bool isSelectedMl = hfMlResponse.isSelectedMl(inputFeatures, ptCand, outputMl);
         hfMlDplusToPiKPiCandidate(outputMl);
 
+        LOG(info) << "pt prong0: " << candidate.ptProng0() << " " << inputFeatures[0];
+        LOG(info) << "cpa: " << candidate.cpa() << " " << inputFeatures[8];
+
```
fcatalan92, line 264:
```suggestion
```
Looks like a debug leftover

PWGHF/Core/HfMlResponseDplusToPiKPi.h
```diff
@@ -0,0 +1,167 @@
...
+      FILL_MAP(tpcTofNSigmaKa1),
+      FILL_MAP(tpcTofNSigmaKa2)};
+  }
+};
+
+} // namespace o2::analysis
+
+#undef FILL_MAP
+#undef CHECK_AND_FILL_VEC_FULL
+#undef CHECK_AND_FILL_VEC
```
fcatalan92, line 165: This does not look safe to me. What happens if we use different similar classes, e.g. for different particles, in the same file? Does it still work?

fcatalan92 (commented):
Hi @AlexBigO, thanks a lot for the updates! I added a couple of minor comments.
I have some doubts about the usage of the macros still, having checked also the [coding guidelines](https://rawgit.com/AliceO2Group/CodingGuidelines/master/coding_guidelines.html?showone=Preprocessor_Macros#Preprocessor_Macros).  

What about moving these macros to each specific class and naming them as `FILL_MAP_DPLUS` etc. ? Another possibility would be to drop them and have a few more lines of code (even if they are handy to avoid copy-paste mistakes)

PWGHF/Core/HfMlResponseDplusToPiKPi.h
AlexBigO: Indeed, this I don't know.

AlexBigO:
> What about moving these macros to each specific class and naming them as `FILL_MAP_DPLUS` etc. ? Another possibility would be to drop them and have a few more lines of code (even if they are handy to avoid copy-paste mistakes)

Yes, we could do this too (I would still keep them because it gives lighter code and avoid errors). Actually, I templated the whole class with the `enum` so these macros would work for all channels. If we define a macro for each channel, we don't need this template anymore.

PWGHF/Core/HfMlResponseDplusToPiKPi.h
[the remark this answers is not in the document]
vkucera: You mean these would be the labels of the array that goes in `hfMlResponse.configure(binsPtMl, cutsMl, cutDirMl, nClassesMl);`?

PWGHF/Core/HfMlResponseDplusToPiKPi.h
```diff
@@ -0,0 +1,193 @@
...
+#include <vector>
+
+#include "PWGHF/Core/HfMlResponse.h"
+
+// Fill the map of available input features
+// the key is the feature's name (std::string)
+// the value is the corresponding value in EnumInputFeatures
+#define FILL_MAP_DPLUS(FEATURE)                                          \
+  {                                                                      \
+#FEATURE, static_cast < uint8_t>(InputFeaturesDplusToPiKPi::FEATURE) \
```
vkucera, line 30:
```suggestion
#FEATURE, static_cast <uint8_t>(InputFeaturesDplusToPiKPi::FEATURE) \
```
Is the extra space intentional?

PWGHF/Core/HfMlResponseDplusToPiKPi.h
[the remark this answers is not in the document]
AlexBigO: It would be the labels used in the configurable vector of strings `namesInputFeatures` called here `hfMlResponse.cacheInputFeaturesIndices(namesInputFeatures);`.

PWGHF/Core/HfMlResponseDplusToPiKPi.h
AlexBigO: No :) I also noticed it, it appears after `clang-format`

PWGHF/Core/HfMlResponseDplusToPiKPi.h
vkucera: @fcatalan92 What is the issue with different particles? Why could the same macros not be used for all particles and undefined at the end of the file?

PWGHF/Core/HfMlResponseDplusToPiKPi.h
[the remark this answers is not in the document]
vkucera: OK, got it. Thanks!

PWGHF/Core/HfMlResponseDplusToPiKPi.h
fcatalan92: @vkucera The point is that these macros are defined and undefined in different headers. So, if we imagine a task using the ML inference for two different particles we get:
```C++
#include "PWGHF/Core/HfMlResponse.h"  // not needed but for the sake of clarity
#include "PWGHF/Core/HfMlResponseDplusToPiKPi.h"
#include "PWGHF/Core/HfMlResponseOtherParticle.h"
```
This means that the macros are defined one time (due to the include guard) and undefined two times. I don't know if this would break compilation, but it looks potentially problematic.

fcatalan92 (approved):
@AlexBigO I'm fully satisfied of the code now. Thanks a lot again for the nice work!

PWGHF/Core/HfMlResponseDplusToPiKPi.h
vkucera: One has to undefine the macros after the last use, of course, so in this case it would be after `#include "PWGHF/Core/HfMlResponseOtherParticle.h"` in the `cxx` file.

vkucera: But since that would lead to a less clean code, I'm happy with the current version.

vkucera (approved):

PR #3658, opened 2023-10-20
preliminary support for Monte Carlo analysis + support for PROCESS_SW…

abilandz:
…ITCH + switch to templates in few function to trim down code bloat

### DIFF
--- a/PWGCF/MultiparticleCorrelations/Core/MuPa-MemberFunctions.h
+++ b/PWGCF/MultiparticleCorrelations/Core/MuPa-MemberFunctions.h
@@ -38,17 +38,22 @@
 // void BookResultsHistograms()
 
 // b) Called directly in process(...):
-// void Preprocess(aod::Collision const& collision, eRecSim rs);
-//   Bool_t MaxNumberOfEvents(const Int_t rs);
-//   void GetParticleWeights();
-//   void DetermineAndPropagateRunNumber(aod::Collision const& collision);
-//   void CheckCurrentRunNumber(aod::Collision const& collision);
-// template <typename T> void FillEventHistograms(aod::Collision const& collision, T const& tracks, const Int_t rs, eBeforeAfter ba)
-// template <typename T> Bool_t EventCuts(aod::Collision const& collision, T const& tracks)
-// template <typename T> void MainLoopOverParticles(T const& tracks, eRecSim rs)
-//   template <typename T> void FillParticleHistograms(T const& track, eRecSim rs, eBeforeAfter ba)
-//   template <typename T> void ParticleCuts(T const& track, eRecSim rs)
-// void ResetEventByEventQuantities();
+// void SteerRec(CollisionRec const& collision, TracksRec const& tracks)
+// void SteerRecSim(CollisionRec const& collision, TracksRecSim const& tracks)
+// void SteerSim(CollisionSim const& collision, TracksSim const& tracks)
+//   template <typename T> void Preprocess(T const& collision)
+//     Bool_t MaxNumberOfEvents(const Int_t rs);
+//     void GetParticleWeights();
+//     template <typename T> void DetermineAndPropagateRunNumber(T const& collision)
+//     template <typename T> void CheckCurrentRunNumber(T const& collision);
+//   template <eRecSim rs, typename T1, typename T2> void FillEventHistograms(T1 const& collision, T2 const& tracks, eBeforeAfter ba)
+//   template <typename T1, typename T2> Bool_t EventCuts(T1 const& collision, T2 const& tracks)
+//   void MainLoopOverParticlesRec(TracksRec const& tracks)
+//   void MainLoopOverParticlesRecSim(TracksRecSim const& tracks)
+//   void MainLoopOverParticlesSim(TracksSim const& tracks)
+//     template <eRecSim rs, typename T> void FillParticleHistograms(T const& track, eBeforeAfter ba)
+//     template <typename T> void ParticleCuts(T const& track, eRecSim rs)
+//   void ResetEventByEventQuantities();
 
 // Double_t Weight(const Double_t &value, const char *variable)
 // void CalculateEverything();
@@ -984,7 +989,8 @@ void BookResultsHistograms()
 
 //============================================================
 
-void Preprocess(aod::Collision const& collision, eRecSim rs)
+template <typename T>
+void Preprocess(T const& collision)
 {
   // Do all thingies before starting to process data (e.g. count number of events, fetch the run number, etc.).
 
@@ -993,7 +999,7 @@ void Preprocess(aod::Collision const& collision, eRecSim rs)
   }
 
   // *) If I reached max number of events, ignore the remaining collisions:
-  if (MaxNumberOfEvents(rs)) {
+  if (MaxNumberOfEvents()) {
     fProcessRemainingEvents = kFALSE;
   }
 
@@ -1014,21 +1020,12 @@ void Preprocess(aod::Collision const& collision, eRecSim rs)
     }
   }
 
-  // *) Check if gProcessRec, gProcessRecSim and gProcessSim global flags are in sync with eRecSim rs in this PROCESS_SWITCH:
-  // TBI 20231020 I need to perform this check only once + implement something also for gProcessRecSim
-  if (gProcessRec && rs != eRec) {
-    LOGF(error, "\033[1;33m%s gProcessRec && rs != eRec \033[0m", __PRETTY_FUNCTION__);
-    LOGF(fatal, "gProcessRec = %d, rs = %d", gProcessRec, eRec);
-  } else if (gProcessSim && rs != eSim) {
-    LOGF(error, "\033[1;33m%s gProcessSim && rs != eSim \033[0m", __PRETTY_FUNCTION__);
-    LOGF(fatal, "gProcessSim = %d, rs = %d", gProcessSim, eSim);
-  }
-
-} // void Preprocess(aod::Collision const& collision, eRecSim rs)
+} // template <typename T> void Preprocess(T const& collision)
 
 //============================================================
 
-void DetermineAndPropagateRunNumber(aod::Collision const& collision)
+template <typename T>
+void DetermineAndPropagateRunNumber(T const& collision)
 {
   // Determine and propagate run number info to already booked objects, wherever it's relevant.
   // Make sure in process(...) that this function is called only once.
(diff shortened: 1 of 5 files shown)


PWGCF/MultiparticleCorrelations/Core/MuPa-MemberFunctions.h
```diff
@@ -2015,4 +2253,210 @@ Double_t Weight(const Double_t& value,
...
+    }
+
+    // *) Fill particle histograms for reconstructed data after particle cuts:
+    FillParticleHistograms(track, rs, eAfter);
+
+    // *) Fill particle histograms for corresponding simulated particle after particle cuts:
+    // TBI 20231020 there is a duplication of code here, see above. I should get 'mcParticle' only once for this 'track' in this loop
+    if (gProcessRecSim) {
+      if (track.has_mcParticle()) {
+        auto mcParticle = track.mcParticle();
```
victor-gonzalez, line 2359: See previous comment

PWGCF/MultiparticleCorrelations/Core/MuPa-MemberFunctions.h
```diff
@@ -2015,4 +2253,210 @@ Double_t Weight(const Double_t& value,
...
+    if (gProcessRec || gProcessRecSim) {
+      //  Take kinematics from reconstructed particles:
+      dPhi = track.phi();
+      dPt = track.pt();
+      dEta = track.eta();
+    } else if (gProcessSim) {
+      //  Take kinematics from simulated particles:
+      dPhi = track.mcParticle().phi();
+      dPt = track.mcParticle().pt();
+      dEta = track.mcParticle().eta();
```
victor-gonzalez, line 2376: This will not compile either. See above comment

PWGCF/MultiparticleCorrelations/Tasks/multiparticle-correlations-ab.cxx
```diff
@@ -112,129 +118,167 @@ struct MultiparticleCorrelationsAB // this name is used in lower-case format to
...
+    }
+
+    // *) Calculate everything for selected events and particles:
+    CalculateEverything();
+
+    // *) Reset event-by-event quantities:
     ResetEventByEventQuantities();
 
-  } // void process(...)
+  } // void processSim(...)
```
victor-gonzalez, line 280: In here you are replicating three times the same procedure. That's the role of the templating mechanism.
Use just one one templated function to do the processing and call it from the three PROCESS_SWITCH'ed functions which should only contain the call to that function. Clean and no code replicated. 
To use templates is not a fashion trend :-)

victor-gonzalez (changes requested):
Please, consider the suggested changes
You have been able to compile your code because you did not call the track processing functionality (basically it was excluded from compilation). If you do it you will discover that you will not be able to compile the code
You need to modify it and the best way of doing is to fully use the template mechanism. The alternative is to replicate all your particle processing code three times as you (not appropriately) did it with the process functions. Templates are there to leave that replication work to the compiler.

PWGCF/MultiparticleCorrelations/Core/MuPa-MemberFunctions.h
```diff
@@ -972,11 +1121,67 @@ void ResetEventByEventQuantities()
...
+
+//============================================================
+
+template <typename T>
+Bool_t EventCuts(aod::Collision const& collision, T const& tracks)
+{
+
+  // Event cuts on reconstructed and simulated data.
+
+  // TBI 20231019 Do I need an additional argument "eRecSim rs"
```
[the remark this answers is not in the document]
abilandz: I forgot to delete this one. The templated version is already available and implemented in ```
template <typename T>
Bool_t EventCuts(aod::Collision const& collision, T const& tracks)
```

File: PWGCF/MultiparticleCorrelations/Core/MuPa-MemberFunctions.h [hunk 5]
```diff
@@ -1024,43 +1229,65 @@ Bool_t EventCuts(aod::Collision const& collision, aod::Tracks const& tracks)
...
+      /*
+      ceh_a.fEventHistograms[eVertex_x][eSim][ba]->Fill(collision.posX());
+      ceh_a.fEventHistograms[eVertex_y][eSim][ba]->Fill(collision.posY());
+      ceh_a.fEventHistograms[eVertex_z][eSim][ba]->Fill(collision.posZ());
+      ceh_a.fEventHistograms[eNContributors][eSim][ba]->Fill(collision.numContrib());
+      */
+      break;
+  } // switch(rs) {
+
+} // template <typename T> void FillEventHistograms(...)
```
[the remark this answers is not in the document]
abilandz: But I need here eRec/eSim always to be enum, and of fixed type eRecSim. Which other type do you have in mind here for this particular function argument, when I could benefit from templates?

In addition, implementation with switch + fixed number of enums is safe, in the sense that this argument has to be either eRec or eSim, which then safely I can use as a corresponding array index (0 or 1).

I doubt there is a big performance penalty for simple operations like this, but in any case I will perform code profiling and optimization at some point later, when all other functionalitied are added to my task, and when I see where most of execution time is going. 

On the last comment: yes. I need to template also on the collision table, so that I can process pure MC.

PWGCF/MultiparticleCorrelations/Core/MuPa-MemberFunctions.h
```diff
@@ -2015,4 +2253,210 @@ Double_t Weight(const Double_t& value,
...
+  fSelectedTracks = 0;           // reset number of selected tracks
+  for (auto& track : tracks) {
+
+    // *) Fill particle histograms for reconstructed particle before particle cuts:
+    FillParticleHistograms(track, rs, eBefore);
+
+    // *) Fill particle histograms for corresponding simulated particle before particle cuts:
+    if (gProcessRecSim) {
+      if (track.has_mcParticle()) {
+        auto mcParticle = track.mcParticle();
```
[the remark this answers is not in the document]
abilandz: Yes, I am aware of that. The whole point of this PR was to test processRecSim(...) in hyperloop, in which both reconstructed and simulated info is available, and where I can call ```track.has_mcParticle()```

The other 2, namely processRec(...) and processSim(...) are still under development, and therefore some things are commented out in the code, until I make them working

PWGCF/MultiparticleCorrelations/Core/MuPa-MemberFunctions.h
abilandz: Same reply as above.

PWGCF/MultiparticleCorrelations/Core/MuPa-MemberFunctions.h
abilandz: Yes, see the above reply.

PWGCF/MultiparticleCorrelations/Tasks/multiparticle-correlations-ab.cxx
abilandz: That was in any case my intention, I left the comment, cite:

```TBI 20231017 There is a (minor) code bloat here obviously. Fix this when it becomes possible to post the last argument to PROCESS_SWITCH via variable (ideally Configurable)```

But since it's not possible to post the last argument to PROCESS_SWITCH via variable, I have introduced now templated Steer(...) function, which is the only one called in processRec(...), processRecSim(...) and processSim(...).

File: PWGCF/MultiparticleCorrelations/Core/MuPa-MemberFunctions.h (hunk 5 above)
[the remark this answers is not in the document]
Comment (victor-gonzalez) [reply PWGCF/MultiparticleCorrelations/Core/MuPa-MemberFunctions.h:1283]: What about 
```
template <eRecSim rs, typename T1, typename T2>
void FillEventHistograms(T1 const& collision, T2 const& tracks, eBeforeAfter ba)
```
Then inside you function you use the construction
```
if constexpr (rs == eRec) {
  ..................
}

if constexpr (rs == eGen) {
  ..................
}
```
and then the compiler decides what to do and the conditional jumps are not in the generated code.

For calling the function then you use 
```
FillEventHistograms<eRec>(collision,tracks, eBefore);
```
for instance.

If you do it that way, you don't need any variable to store the process you are going to do.
Just template in the same way the whole chain. The initial call in the corresponding process function is done with the corresponding `enum` value and then that value is propagated along the whole chain and used when needed.
The compiler takes care of replicate when is needed and no additional jumps on `if` conditions are introduced.

I told you, templates are not for fashion :)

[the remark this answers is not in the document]
Comment (abilandz) [reply PWGCF/MultiparticleCorrelations/Core/MuPa-MemberFunctions.h]: OK, implemented now this way, both for FillEventHistograms and FillParticleHistograms

File: PWGCF/MultiparticleCorrelations/Core/MuPa-MemberFunctions.h [hunk 7]
```diff
@@ -1093,17 +1316,34 @@ void FillParticleHistograms(
...
+    // b) Only simulated:
+    case eSim:
+      /* TBI 20231019 not ready yet
+      cph_a.fParticleHistograms[ePhi][rs][ba]->Fill(track.phi());
+      cph_a.fParticleHistograms[ePt][rs][ba]->Fill(track.pt());
+      cph_a.fParticleHistograms[eEta][rs][ba]->Fill(track.eta());
+*/
+      break;
+  } // switch(rs) {
+
```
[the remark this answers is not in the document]
abilandz: Used now ```if constexpr (condition)```

PWGCF/MultiparticleCorrelations/Core/MuPa-MemberFunctions.h
```diff
@@ -2015,4 +2186,417 @@ Double_t Weight(const Double_t& value,
...
+  if (fVerbose) {
+    LOGF(info, "\033[1;32m%s\033[0m", __PRETTY_FUNCTION__);
+  }
+
+  // TBI 20231021 tbc...
+
+} // void MainLoopOverParticlesSim(TracksSim const& tracks)
+
+//============================================================
```
victor-gonzalez, line 2474: Please, don't replicate this code.
I will allow you to progress but I don't want to see it in the next iteration. Grrrrrh!

PWGCF/MultiparticleCorrelations/Core/MuPa-MemberFunctions.h
```diff
@@ -2015,4 +2186,417 @@ Double_t Weight(const Double_t& value,
...
+
+void SteerSim(CollisionSim const& collision, TracksSim const& tracks)
+{
+
+  // TBI 20231021 tbc...
+
+} // void SteerSim(CollisionSim const& collision, TracksSim const& tracks)
+
```
victor-gonzalez, line 2601: Same here!!!

victor-gonzalez (approved):
I will approve it for the time being
But I don't want to see this code replication at all
Imagine someone is looking at your code and replicates what you are doing
I will have to follow up this guy again explaining the same things I did to you
For easing your job mine gets complicated
Please, have it in mind

PR #3662, opened 2023-10-20
PWGEM: update MCtemplates task

feisenhu:
Update the MCtemplates task.


rbailhac:
Hi Florian, it is the same as in the table reader. I do not think that you need the FillPairVertexing in first place. I would at least put a flag to give the possibility to not enter there and avoid unnecessary CPU time.

rbailhac (dismissed):

alibuild (approved):

PR #3664, opened 2023-10-20
EventFiltering: Add max pt cuts

fgrosa:


alibuild (approved):

PR #3674, opened 2023-10-23
Replace RecoDecay::getMass with constants

qgp:
@skundu692 replacing calls to RecoDecay functions as per @vkucera's request

### DIFF
--- a/EventFiltering/PWGLF/filterf1proton.cxx
+++ b/EventFiltering/PWGLF/filterf1proton.cxx
@@ -36,6 +36,7 @@
 #include "Common/DataModel/Multiplicity.h"
 #include "Common/DataModel/PIDResponse.h"
 #include "PWGLF/DataModel/LFStrangenessTables.h"
+#include "PWGHF/Core/PDG.h"
 #include "DataFormatsTPC/BetheBlochAleph.h"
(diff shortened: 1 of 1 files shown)


vkucera (approved):
Thanks @qgp

EventFiltering/PWGLF/filterf1proton.cxx
```diff
@@ -36,6 +36,7 @@
 #include "Common/DataModel/Multiplicity.h"
 #include "Common/DataModel/PIDResponse.h"
 #include "PWGLF/DataModel/LFStrangenessTables.h"
+#include "PWGHF/Core/PDG.h"
```
mpuccio, line 39: Do we need the dependence from a PWGHF header for the PDG masses? I guess this should be moved to a more central location @vkucera

vkucera: Hi @mpuccio , yes, the plan is to move the masses to O2 eventually (see the timeline in the linked issue). I would avoid moving the header around for now at least until we get rid of `getMassPDG` completely and we can replace the header in one go in bulk.

mpuccio (approved):

PR #3675, opened 2023-10-23
Adding new tasks and moving old ones for 'Femto3D' package

glromane:
--Implementing small changes in "singletrackselector" task and moving them to "Femto3D" folder in PWGCF branch;

--adding a QA task that uses tables from "singletrackselector";

--adding a femto task that does mixing, pair creation for obtaining a femtoscopic correlation function;

### DIFF
--- a/PWGDQ/Core/HistogramsLibrary.cxx
+++ b/PWGDQ/Core/HistogramsLibrary.cxx
@@ -477,6 +477,12 @@ void o2::aod::dqhistograms::DefineHistograms(HistogramManager* hm, const char* h
       hm->AddHistogram(histClass, "pdca_vs_Rabs_vs_p", "pDCA vs R_{abs} vs p", false, 2000, 0.0, 20.0, VarManager::kP, 100, 0., 200., VarManager::kMuonRAtAbsorberEnd, 200, 0.0, 1000., VarManager::kMuonPDca);
       hm->AddHistogram(histClass, "pdca_vs_Rabs_vs_pt", "pDCA vs R_{abs} vs pt", false, 2000, 0.0, 20.0, VarManager::kPt, 100, 0., 200., VarManager::kMuonRAtAbsorberEnd, 200, 0.0, 1000., VarManager::kMuonPDca);
     }
+    if (subGroupStr.Contains("mft-pid")) {
+      hm->AddHistogram(histClass, "hMftTrackEtaVsPt", "", false, 100, -5.f, -2.f, VarManager::kEta, 100, 0.f, 20.f, VarManager::kPt);
+      hm->AddHistogram(histClass, "hMftNClusters", "", false, 16, -0.5f, 15.5f, VarManager::kMftNClusters);
+      hm->AddHistogram(histClass, "hMftClusterSize", "", false, 101, -0.5f, 100.5f, VarManager::kMftClusterSize);
+      hm->AddHistogram(histClass, "hMftMeanClusterSize", "", false, 200, 0.f, 20.f, VarManager::kMftMeanClusterSize);
+    }
     if (subGroupStr.Contains("mc")) {
       hm->AddHistogram(histClass, "Pt_vs_PtMC", "pT vs MC pT", false, 200, 0.0, 20.0, VarManager::kPt, 200, 0.0, 20.0, VarManager::kMCPt);
       hm->AddHistogram(histClass, "Eta_vs_EtaMC", "#eta vs MC #eta", false, 50, -1.0, 1.0, VarManager::kEta, 50, -1.0, 1.0, VarManager::kMCEta);
--- a/PWGDQ/Core/VarManager.cxx
+++ b/PWGDQ/Core/VarManager.cxx
@@ -438,6 +438,12 @@ void VarManager::SetDefaultVarNames()
   fgVariableUnits[kIsLegFromOmega] = "";
   fgVariableNames[kMuonNClusters] = "muon n-clusters";
   fgVariableUnits[kMuonNClusters] = "";
+  fgVariableNames[kMftNClusters] = "MFT n-clusters";
+  fgVariableUnits[kMftNClusters] = "";
+  fgVariableNames[kMftClusterSize] = "MFT cluster size";
+  fgVariableUnits[kMftClusterSize] = "";
+  fgVariableNames[kMftMeanClusterSize] = "<MFT cluster size>";
+  fgVariableUnits[kMftMeanClusterSize] = "";
   fgVariableNames[kMuonRAtAbsorberEnd] = "R at the end of the absorber";
   fgVariableUnits[kMuonRAtAbsorberEnd] = "cm";
   fgVariableNames[kMuonPDca] = "p x dca";
--- a/PWGDQ/Core/VarManager.h
+++ b/PWGDQ/Core/VarManager.h
@@ -345,6 +345,9 @@ class VarManager : public TObject
     kMuonDCAy,
     kMuonTime,
     kMuonTimeRes,
+    kMftNClusters,
+    kMftClusterSize,
+    kMftMeanClusterSize,
 
     // MC particle variables
     kMCPdgCode,
@@ -1080,6 +1083,25 @@ void VarManager::FillTrack(T const& track, float* values)
 {
   if (!values) {
     values = fgValues;
+  }
+
+  if constexpr ((fillMap & TrackMFT) > 0) {
+    values[kPt] = track.pt();
+    values[kEta] = track.eta();
+    values[kPhi] = track.phi();
+    values[kMftNClusters] = track.nClusters();
+
+    uint64_t mftClsAndFlgs = track.mftClusterSizesAndTrackFlags();
+    double meanClusterSize = 0;
+    for (int i = 0; i < 10; ++i) {
+      double size = (mftClsAndFlgs >> (i * 6)) & 0x3fULL;
+      values[kMftClusterSize + i] = (mftClsAndFlgs >> (i * 6)) & 0x3fULL;
+      if (size > 0) {
+        meanClusterSize += size;
+      }
+    }
+    meanClusterSize /= track.nClusters();
+    values[kMftMeanClusterSize] = meanClusterSize;
   }
 
   // Quantities based on the basic table (contains just kine information and filter bits)
--- a/PWGDQ/DataModel/ReducedInfoTables.h
+++ b/PWGDQ/DataModel/ReducedInfoTables.h
@@ -270,12 +270,18 @@ DECLARE_SOA_COLUMN(FilteringFlags, filteringFlags, uint8_t); //!
 DECLARE_SOA_COLUMN(Pt, pt, float);   //!
 DECLARE_SOA_COLUMN(Eta, eta, float); //!
 DECLARE_SOA_COLUMN(Phi, phi, float); //!
+DECLARE_SOA_COLUMN(Sign, sign, int); //!
+DECLARE_SOA_COLUMN(MftClusterSizesAndTrackFlags, mftClusterSizesAndTrackFlags, uint64_t); //!
 } // namespace reducedmft
 
 // MFT track kinematics
 DECLARE_SOA_TABLE(ReducedMFTTracks, "AOD", "RMFTTR", //!
                   o2::soa::Index<>, reducedmft::ReducedEventId, reducedmft::FilteringFlags,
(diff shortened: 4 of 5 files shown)


victor-gonzalez (changes requested):
Please, use the same sub-directories structure for `Femto3D` as the other functional areas, as for example `FemtoUniverse`

Have in mind that the mandate is that there should be only one derived data datamodel for `Femto` analysis. There has been extensive work in this area and such a datamodel already exists. We will further iterate on that (see my e-mail)

zchochul:
Please try to reduce the amount of commented code. (like here: https://github.com/AliceO2Group/O2Physics/blob/08eb0e5793909714d354b60b31281572316af107/PWGCF/Femto3D/femto3dPairTask.cxx#L45)

PWGCF/Femto3D/TableProducer/singleTrackSelector.cxx
```diff
@@ -0,0 +1,180 @@
...
+                 collision.posZ(),
+                 d_bz);
+
+    for (auto& track : tracks) {
+
+      for (auto i : particlesToReject) {
+        // if satisfied, want to continue in the upper loop (over tracks) -- skip the current track
+        // cannot use simple 'continue' since it will be applied to the current loop, so have to use 'goto'
+        if (TOFselection(track, std::make_pair(i, rejectWithinNsigmaTOF))) {
+          goto skip;
```
victor-gonzalez, line 141: Please, no `goto`
Use meaningful name variables to control the execution flow

PWGCF/Femto3D/TableProducer/singleTrackSelector.cxx
```diff
@@ -0,0 +1,180 @@
...
+                   track.phi(),
+                   singletrackselector::packInTable<singletrackselector::nsigma::binning>(track.tofNSigmaPr()),
+                   singletrackselector::packInTable<singletrackselector::nsigma::binning>(track.tpcNSigmaPr()),
+                   singletrackselector::packInTable<singletrackselector::nsigma::binning>(track.tofNSigmaDe()),
+                   singletrackselector::packInTable<singletrackselector::nsigma::binning>(track.tpcNSigmaDe()));
+
+          break; // break the loop with particlesToKeep after the 'if' condition is satisfied -- don't want double entries
+        }
+
+    skip:; // here we send with 'goto'
```
victor-gonzalez, line 172: Please, no labels
Use meaningful name variables to control the execution flow

PWGCF/Femto3D/Tasks/femto3dPairTask.cxx
```diff
@@ -0,0 +1,305 @@
...
+
+          registry.fill(HIST("p_first"), track.p());
+          if (_particlePDG_1 == 2212) {
+            registry.fill(HIST("nsigmaTOF_first"), track.p(), track.tofNSigmaPr());
+            registry.fill(HIST("nsigmaTPC_first"), track.p(), track.tpcNSigmaPr());
+          }
+          if (_particlePDG_1 == 1000010020) {
+            registry.fill(HIST("nsigmaTOF_first"), track.p(), track.tofNSigmaDe());
+            registry.fill(HIST("nsigmaTPC_first"), track.p(), track.tpcNSigmaDe());
+          }
```
victor-gonzalez, line 167: You are replicating in memory everything that you get from the tables with a much more inefficient way of accessing to the stored information. 
This will have a huge penalty in execution time and in memory requirements
This goes against the O2 execution model (SoA vs AoS)
It seems that one of the reasons for storing the information is for extracting the mixed event results
You have to use the mixed events framework that is standard in O2Physics
It works without any kind of data replication

PWGCF/Femto3D/Tasks/femto3dPairTask.cxx
```diff
@@ -0,0 +1,305 @@
...
+
+          registry.fill(HIST("p_second"), track.p());
+          if (_particlePDG_2 == 2212) {
+            registry.fill(HIST("nsigmaTOF_second"), track.p(), track.tofNSigmaPr());
+            registry.fill(HIST("nsigmaTPC_second"), track.p(), track.tpcNSigmaPr());
+          }
+          if (_particlePDG_2 == 1000010020) {
+            registry.fill(HIST("nsigmaTOF_second"), track.p(), track.tofNSigmaDe());
+            registry.fill(HIST("nsigmaTPC_second"), track.p(), track.tpcNSigmaDe());
+          }
```
victor-gonzalez, line 186: Same comment as above

PWGCF/Femto3D/Tasks/femto3dPairTask.cxx
```diff
@@ -0,0 +1,305 @@
...
+            for (auto iii : selectedtracks_2[colId_2]) {
+              FemtoPair* Pair = new FemtoPair(ii, iii, IsIdentical);
+              if (!Pair->IsClosePair(_deta, _dphi, _radiusTPC))
+                registry.fill(HIST("ME"), Pair->GetKstar());
+            }
+          }
+        }
+      } // end of mixed event non-idetical
+
+    } //====================================== end of mixing non-identical ======================================
```
victor-gonzalez, line 273: Same comment as above

victor-gonzalez (changes requested):
Thanks for reallocating the source files in the suggested structure

For the actual code, please, consider the suggested comments, specially the one encouraging the use of the standard O2Physics mixed events framework and not to replicate in memory what is received from tables

glromane:
> Thanks for reallocating the source files in the suggested structure
> 
> For the actual code, please, consider the suggested comments, specially the one encouraging the use of the standard O2Physics mixed events framework and not to replicate in memory what is received from tables

Dear Victor, thanks for your comments, here is my overall reply (I also commented on your separate comments).

I took into account your suggestions and applied them in the new commit.
Small things as namespace, goto etc. I fixed. For the "FemtoParticle", "FemtoPair" objects and mixing procedure I'm putting a more detailed comment:

First, for the tables I cannot use filtering for the PID selection which means that even in the filtered with the track cuts table I will have lots of unwanted entries. And if I pass the table with the "fake" tracks to the mixer (as in the FemtoDream) I'll have lots of "fake" pairs to process the PID selection for. The number of pairs scales as n*(n-1) (x2 because it's a pair) and I think it's very inefficient and instead I do the PID before the mixing and pairs I build doesn't need to be checked anymore. So, instead of 2*n*(n-1) PID checks, I do it only n-times. Yes, for that I need to store the info and allocate memory but I gain a lot in CPU efficiency. And for the use of O2 mixing algorithm after the PID I'd need to make another table and duplicate the objects anyway, so it's the same only use of the "FemtoParticle" class is much more convenient for me.

There are also some other for wrt to the previous commit made for the sake of efficiency (for example I got rid of the function that does the permutations and replaced it with just nested loops).

We will continue working on improvements and of course, if we don't meet the hyper loop requirements we will deal with it, but before we need to have our tasks there to test and see how to proceed.

If there is something unclear, please, tell me, I'll give you more details. Thank you.

Sincerely yours, Gleb.

PWGCF/Femto3D/Core/femto3dPairTask.h
```diff
@@ -0,0 +1,224 @@
...
+  } else {
+    TLorentzVector fourmomentasum = *first4momentum + *second4momentum;
+
+    first4momentum->Boost((-1) * fourmomentasum.BoostVector());
+    second4momentum->Boost((-1) * fourmomentasum.BoostVector());
+
+    TVector3 qinv = first4momentum->Vect() - second4momentum->Vect();
+    return 0.5 * abs(qinv.Mag());
+  }
+}
```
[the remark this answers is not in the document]
glromane: Fixed, put them in the "o2::aod::singletrackselector::" namespace.

PWGCF/Femto3D/DataModel/singletrackselector.h
```diff
@@ -0,0 +1,201 @@
...
+                           [](nsigma::binning::binned_t nsigma_binned) -> float { return singletrackselector::unPack<nsigma::binning>(nsigma_binned); });
+DECLARE_SOA_DYNAMIC_COLUMN(TOFNSigmaDe, tofNSigmaDe,
+                           [](nsigma::binning::binned_t nsigma_binned) -> float { return singletrackselector::unPack<nsigma::binning>(nsigma_binned); });
+DECLARE_SOA_DYNAMIC_COLUMN(TPCNSigmaDe, tpcNSigmaDe,
+                           [](nsigma::binning::binned_t nsigma_binned) -> float { return singletrackselector::unPack<nsigma::binning>(nsigma_binned); });
+
+} // namespace singletrackselector
+
+DECLARE_SOA_TABLE_FULL(SingleTrackSel, "SelTracks", "AOD", "STSEL", // Table of the variables for single track selection.
+                       o2::soa::Index<>,
```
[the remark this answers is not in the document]
glromane: Corrected.

PWGCF/Femto3D/DataModel/singletrackselector.h
```diff
@@ -0,0 +1,201 @@
...
+                       singletrackselector::TPCNSigmaPr<singletrackselector::StoredTPCNSigmaPr>,
+                       singletrackselector::TOFNSigmaDe<singletrackselector::StoredTOFNSigmaDe>,
+                       singletrackselector::TPCNSigmaDe<singletrackselector::StoredTPCNSigmaDe>);
+
+} // namespace o2::aod
+#endif // PWGCF_DATAMODEL_SINGLETRACKSELECTOR_H_
+
+template <typename TrackType>
+inline bool TPCselection(TrackType const& track, std::pair<int, std::vector<float>> const& PIDcuts)
+{
```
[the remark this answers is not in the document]
glromane: This structure contains a PDG/PIDcuts pair and if I'm not mistaken the pair is created in only ones at the beginning of a task (QA of PairTask) and this function only reads it since it's by a "const" reference.

PWGCF/Femto3D/TableProducer/singleTrackSelector.cxx
glromane: I put a flag instead. No 'goto' anymore.

PWGCF/Femto3D/TableProducer/singleTrackSelector.cxx
glromane: Removed.

PWGCF/Femto3D/Tasks/femto3dPairTask.cxx
glromane: Maybe it doesn't seem right but I'll try to explain why I think it's more convenient to use it like this. The main reason is the mixing (I put a more detailed comment about the mixing in the main reply). Since we need to store some info per particle for the mixing later I decided to use this "FemtoParticle" object that contains the bare minimum: pT, Eta, Phi, magField and signedPDG (updated in the last commit wrt to the previous one -- now consumes less memory).

Actually I was trying to store iterators/pointers to the tracks but the ones that are created after slicing or filtering are transient (temporary) ones and cannot be used to access an object (for the mixing) after the selection. So, I don't see any other way to do it apart from creating another additional table which would lead to duplication anyway. But using the class "FemtoPatricle" is more convenient because I can put there needed functions, pass the object to the "FemtoPair" etc.

PWGCF/Femto3D/Tasks/femto3dPairTask.cxx
glromane: Here you're right. I fixed this and now I use only one "FemtoPair" object and just change the stored objects inside it. Thanks for pointing it out to me.

victor-gonzalez:
We cannot approve the code as it is
One of the rules we have is to approve code that could be use as reference for other people 
Clearly this code is against the O2 rules

Consider this procedure. Once you have your filtered collisions/tracks, your derived data, consider using an additional column you can join to your filtered tracks table which will tag your filtered tracks according to your selection criteria; then use the standard O2 mixed events after filtering the filtered tracks over the additional column

If you plan to elaborate more on your code exclude everything which is not the filter
I will approve the filter.  Then, running the filter on hyperloop will allow you to get enough derived data for your local tests in case you need them

jgrosseo (changes requested):
See: https://github.com/AliceO2Group/O2Physics/pull/3675#issuecomment-1815138879

jgrosseo:
Just to add: if you think something is missing in the O2 event mixing framework, please let us know. We can organize a short slot in the WP4/14 meeting and it can be discussed if it cannot be resolved by simply replying here...

PR #3678, opened 2023-10-23
PWGHF: D0 selector: Improve cut on normalised daughter imp. par. and decay length

cterrevo:
Normalized impPard0 hardcoded cut removed and Normalized impPard0 selection added to D0 selection matrix.


PWGHF/TableProducer/candidateSelectorD0.cxx
```diff
@@ -174,9 +171,7 @@ struct HfCandidateSelectorD0 {
     if (candidate.decayLengthXY() > cuts->get(pTBin, "decay length XY")) {
       return false;
     }
-    if (candidate.decayLengthNormalised() * candidate.decayLengthNormalised() < 1.0) {
-      // return false; // add back when getter fixed
-    }
```
mfaggin, line 179: I'm just wondering if this kind of cut (of course not with an hardcoded value) might be useful for non-prompt D analyses. @fgrosa @fcatalan92 what do you think?
Probably it's a minor point, since in line 165 there is the `minimum decay length` cut that does the same on the decay length, rather then the normalized one.

mfaggin (commented):
Ciao @cterrevo thank you very much for the work. I just have a minor comment, let's see also the response from the others. The rest looks fine to me.

fgrosa:
> Normalized impPard0 hardcoded cut removed and Normalized impPard0 selection added to D0 selection matrix.

@vkucera @nzardosh since you were the original authors of the D0 selector could you have a look whether this is fine with you? (@fcolamar and @mmazzilli since you also use the D0 often you could also cross check this). Thanks!

PWGHF/TableProducer/candidateSelectorD0.cxx
fgrosa: @mfaggin these lines do not remove any cut, since it was commented (and not pT-differential). Apart from that, I would say that normalised decay length and normalised decay length XY should be similar (but also I would use them in a BDT for the non-prompt, I would not apply a linear cut at the level of the selector).

mfaggin: thanks @fgrosa , it sounds good to me (and yes I saw that this was useless since the `return` was commented out, I was just thinking for a future fix and usage. But you already replied, thanks).

All good for me, let's wait the response of the others.

vkucera:
> I admit I don't remember anymore the rationale for adding each cut, but the current menu looks a bit confusing to me.
> 
>     * There are currently two cuts on the decay length:
>       
>       * `minimum decay length` for the minimum, which for some reason squares the values, which I don't understand,
>       * `decay length` for the maximum, which is not obvious from the name.
>       
>       It would be better if they had both clear and consistent names, e.g. `min decay length`, `max decay length`.
> 
>     * What does the `d` in `impPard0 norm` stand for?
> 
>     * Does removing the, albeit inactive, cut on `decayLengthNormalised` mean that this quantity (and the related stored error) is not needed anymore?

I just noticed that @fgrosa mentioned he would use `decayLengthNormalised` for the BDT, so I guess we keep it. :-)

cterrevo:
Hello Vit,  
some replies here: 
> There are currently two cuts on the decay length: minimum decay length for the minimum, which for some reason squares the values, which I don't understand,  decay length for the maximum, which is not obvious from the name.
It would be better if they had both clear and consistent names, e.g. min decay length, max decay length.
--> indeed I think that some of these cuts are copied from the method IsSelectedSpecialCuts of AliRDHFCutsD0toKpi, but Nima can confirm. 
I did not want to modify the work done by you and Nima, so I just implemented three main changes removing the hardcoded cuts  (we need to check MC productions with all the cut released) and I also profit to remove one not used cut.  

     * What does the `d` in `impPard0 norm` stand for?
impact parameter "d0"  

cheers, 
Cristina

vkucera:
Hi @cterrevo ,

- Your PR already modifies the cut calculation, so if we agree that there is no reason to square the values, it can be fixed in the same PR, as well as the names.
- If `impPar` and `d0` each mean "impact parameter", then `impPard0` is a duplication.

fgrosa:
Hi @vkucera @cterrevo thanks for the comments!
Indeed I would proceed as suggested:
- rename `impPard0` which is redundant (I would personally also add the information that is the daughter impact parameter, and that it is in XY, e.g. `dauImpParXy` or something similar)
- remove the square in the minimum decay length cut (it is not needed, surely not as implemented)
- add `Max` to the maximum decay length cuts.
- regarding the `decayLengthNormalised` either we just remove the ineffective lines, or we add it to the cut matrix. Given that we have the `decayLengthNormalisedXy` cut already, I personally find not necessary to add also the 3D one. For the BDT one just need to have it in the candidate table, not in the selector.

fcatalan92 (approved):
LGTM

PWGHF/TableProducer/candidateSelectorD0.cxx
```diff
@@ -158,25 +158,19 @@ struct HfCandidateSelectorD0 {
...
-    double decayLengthCut = std::min((candidate.p() * 0.0066) + 0.01, cuts->get(pTBin, "minimum decay length"));
-    if (candidate.decayLength() * candidate.decayLength() < decayLengthCut * decayLengthCut) {
+    if (candidate.decayLength() < cuts->get(pTBin, "minimum decay length")) {
       return false;
     }
-    if (candidate.decayLength() > cuts->get(pTBin, "decay length")) {
+    if (candidate.decayLength() > cuts->get(pTBin, "max decay length")) {
       return false;
     }
     if (candidate.decayLengthXY() > cuts->get(pTBin, "decay length XY")) {
```
fgrosa, line 170:
```suggestion
    if (candidate.decayLengthXY() > cuts->get(pTBin, "max decay length XY")) {
```

PWGHF/Core/SelectorCuts.h
```diff
@@ -300,7 +300,7 @@ static const std::vector<std::string> labelsPt = {
   "pT bin 24"};
 
 // column labels
-static const std::vector<std::string> labelsCutVar = {"m", "DCA", "cos theta*", "pT K", "pT Pi", "d0K", "d0pi", "d0d0", "cos pointing angle", "cos pointing angle xy", "normalized decay length XY", "decay length", "decay length XY", "minimum decay length"};
+static const std::vector<std::string> labelsCutVar = {"m", "DCA", "cos theta*", "pT K", "pT Pi", "d0K", "d0pi", "d0d0", "cos pointing angle", "cos pointing angle xy", "normalized decay length XY", "max decay length", "decay length XY", "minimum decay length", "norm dauImpPar XY"};
```
fgrosa, line 303:
```suggestion
static const std::vector<std::string> labelsCutVar = {"m", "DCA", "cos theta*", "pT K", "pT Pi", "d0K", "d0pi", "d0d0", "cos pointing angle", "cos pointing angle xy", "normalized decay length XY", "max decay length", "max decay length XY", "minimum decay length", "norm dauImpPar XY"};
```

fgrosa (changes requested):
Thanks @cterrevo, just two minor modifications, then we can merge it.

vkucera (changes requested):
Sorry, my comments did not get submitted.

PWGHF/Core/SelectorCuts.h
```diff
@@ -300,7 +300,7 @@ static const std::vector<std::string> labelsPt = {
   "pT bin 24"};
 
 // column labels
-static const std::vector<std::string> labelsCutVar = {"m", "DCA", "cos theta*", "pT K", "pT Pi", "d0K", "d0pi", "d0d0", "cos pointing angle", "cos pointing angle xy", "normalized decay length XY", "decay length", "decay length XY", "minimum decay length"};
+static const std::vector<std::string> labelsCutVar = {"m", "DCA", "cos theta*", "pT K", "pT Pi", "d0K", "d0pi", "d0d0", "cos pointing angle", "cos pointing angle xy", "normalized decay length XY", "max decay length", "decay length XY", "minimum decay length","norm dauImpParX"};
```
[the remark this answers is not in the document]
vkucera: But I see that you already implemented some.

PWGHF/TableProducer/candidateSelectorD0.cxx
```diff
@@ -158,25 +158,19 @@ struct HfCandidateSelectorD0 {
...
-
-    // decay exponentail law, with tau = beta*gamma*ctau
-    // decay length > ctau retains (1-1/e)
-    if (std::abs(candidate.impactParameterNormalised0()) < 0.5 || std::abs(candidate.impactParameterNormalised1()) < 0.5) {
+    if (std::abs(candidate.impactParameterNormalised0()) < cuts->get(pTBin, "norm dauImpPar XY") || std::abs(candidate.impactParameterNormalised1()) < cuts->get(pTBin, "norm dauImpPar XY")) {
       return false;
     }
-    double decayLengthCut = std::min((candidate.p() * 0.0066) + 0.01, cuts->get(pTBin, "minimum decay length"));
-    if (candidate.decayLength() * candidate.decayLength() < decayLengthCut * decayLengthCut) {
+    if (candidate.decayLength() < cuts->get(pTBin, "minimum decay length")) {
```
vkucera, line 164:
```suggestion
    if (candidate.decayLength() < cuts->get(pTBin, "min decay length")) {
```

PWGHF/Core/SelectorCuts.h
```diff
@@ -300,7 +300,7 @@ static const std::vector<std::string> labelsPt = {
   "pT bin 24"};
 
 // column labels
-static const std::vector<std::string> labelsCutVar = {"m", "DCA", "cos theta*", "pT K", "pT Pi", "d0K", "d0pi", "d0d0", "cos pointing angle", "cos pointing angle xy", "normalized decay length XY", "decay length", "decay length XY", "minimum decay length"};
+static const std::vector<std::string> labelsCutVar = {"m", "DCA", "cos theta*", "pT K", "pT Pi", "d0K", "d0pi", "d0d0", "cos pointing angle", "cos pointing angle xy", "normalized decay length XY", "max decay length", "max decay length XY", "minimum decay length", "norm dauImpPar XY"};
```
vkucera, line 303:
```suggestion
static const std::vector<std::string> labelsCutVar = {"m", "DCA", "cos theta*", "pT K", "pT Pi", "d0K", "d0pi", "d0d0", "cos pointing angle", "cos pointing angle xy", "normalized decay length XY", "max decay length", "max decay length XY", "min decay length", "norm dauImpPar XY"};
```

vkucera (changes requested):
We also have a cut on the minimum normalised  decay length in XY, which is now called `normalized decay length XY`. It would be better to rename that one as well for clarity as `min norm decay length XY`.

fgrosa (dismissed):

PWGHF/TableProducer/candidateSelectorD0.cxx
vkucera: This has not been addressed

vkucera (changes requested):

fcatalan92:
@fgrosa @vkucera Do we really want to keep this cut on the maximum decay length? Why not just removing it?
To me it is not very useful and complicates a bit the code, as we are seeing from the review.

fgrosa:
> @fgrosa @vkucera Do we really want to keep this cut on the maximum decay length? Why not just removing it? To me it is not very useful and complicates a the code, as we are seeing from the review.

Personally I don't find it very useful, however if it is clear from the description, I don't see any problem in keeping it. Maybe we could keep only one of the two (either 3D or XY) and instead add the minimum cut also for the decay length XY.

vkucera (approved):
Thanks @cterrevo for implementing the changes.

PR #3681, opened 2023-10-24
PWGHF: Adding multiplicity info in the Lc event tree

DelloStritto:


mfaggin:
all fine for me, thanks @DelloStritto

mfaggin (approved):

PR #3689, opened 2023-10-25
Tutorials/PWGHF: Add O2AT-3 skeleton

vkucera:
@DelloStritto


fcolamar:
Thanks a lot Vit, the new, split miniskim/minitask looks very nice, and evertyhing seems in order. Thanks also for the very nice readme.
As soon as an input file is available, I would also advertise this to Antonio, Alexandre and Federica, so that they can start having a look already in advance.
Fabio

fcolamar (approved):

PR #3692, opened 2023-10-25
PWGHF: PDG: Add Xi0

vkucera:
Needed in LF


alibuild (approved):

PR #3693, opened 2023-10-25
DPG: Remove RecoDecay::getMassPDG

vkucera:


njacazio (dismissed):

DPG/Tasks/AOTTrack/perfK0sResolution.cxx
```diff
@@ -73,7 +73,7 @@ struct perfK0sResolution {
       return kFALSE;
     if (v0.v0radius() < v0setting_radius)
       return kFALSE;
-    if (v0.distovertotmom(collision.posX(), collision.posY(), collision.posZ()) * RecoDecay::getMassPDG(kK0Short) > 2.684 * v0lifetime)
```
njacazio, line 76: Actually, now that I read better, could you move it to the ones defined in `O2/Common/Constants/include/CommonConstants/PhysicsConstants.h` ?

njacazio: Apologies, I meant https://github.com/AliceO2Group/AliceO2/blob/dev/Common/Constants/include/CommonConstants/PhysicsConstants.h

vkucera: Hi @njacazio , the change in this PR guarantees getting the same value as from `RecoDecay::getMassPDG`. The value in `PhysicsConstants.h` is different. If you prefer to use that one, please replace it in a separate PR.

ddobrigk: Hi @vkucera , as you know, I am with @njacazio on this one. If there is something off with the master constants file, we should fix that ASAP and switch to that instead of having masses declared in two different places. Personally, I think we are doing unnecessary work in switching to the HF header instead of immediately switching to the main O2 constants file file, but since you have agreed you will later switch all the code to O2 constants and migrate away from the HF header as soon as all RecoDecay replacements are done, I do not explictly object to this intermediate step. Thanks!

njacazio: I think we can merge https://github.com/vkucera/O2Physics/pull/31/ first, I checked the different values and it should be ok. Still we should agree on some, these might also be different wrt what the PDG reports

njacazio: Then we could fix the O2 to align with the PDG, so that the info is propagated urbi et orbi @vkucera what do you think?

vkucera: @njacazio I don't understand the motivation for https://github.com/vkucera/O2Physics/pull/31. I doesn't seem to make the unification simpler. It is instead fragmenting more the locations from where we pick the masses. The ultimate goal is to include `PDG.h` into `CommonConstants/PhysicsConstants.h` and pick all masses from there. Let's not diverge from that.

vkucera: Now that I see what is in [`PID.h`](https://github.com/AliceO2Group/AliceO2/blob/dev/DataFormats/Reconstruction/include/ReconstructionDataFormats/PID.h), it actually takes the masses from `PhysicsConstants.h`, right? So the result would be the same as in your first proposal.

vkucera: In, the end, I think that's fine.
@ddobrigk , what do you think?

vkucera: @njacazio , for my information, what is the benefit of storing the masses in a `GPUconstexpr()` array?

njacazio (approved):

njacazio:
Thanks Vit!

PR #3695, opened 2023-10-25
PWGJE: Remove RecoDecay::getMassPDG

vkucera:
Please note that this is an intermediate step of the ongoing effort to optimise memory consumption and to unify the location from where we pick the PDG masses. The ultimate goal is to get the masses from [`CommonConstants/PhysicsConstants.h`](https://github.com/AliceO2Group/AliceO2/blob/dev/Common/Constants/include/CommonConstants/PhysicsConstants.h) or from the `O2DatabasePDG` service.


vkucera:
@nzardosh please check the changes and fix the remaining use of `RecoDecay` (marked with `FIXME`). I don't know how you prefer that to be implemented.

nzardosh (approved):

PR #3700, opened 2023-10-26
PWGHF: SigmaC candidate creation w/ or w/o time-reassociated tracks 

mfaggin:
Modification on the candidate creator done to enable the case w/o time-reassociated tracks, given that the `time-to-collision-associator` is quite CPU consuming and its importance for soft pions might be not crucial (to be tested in analysis).
Major changes:
- new `process` functions to handle this
- new `template` functions to handle the different data-type for the soft pion track

### DIFF
--- a/PWGHF/TableProducer/candidateCreatorSigmac0plusplus.cxx
+++ b/PWGHF/TableProducer/candidateCreatorSigmac0plusplus.cxx
@@ -51,6 +51,7 @@ struct HfCandidateCreatorSigmac0plusplus {
 
   /// Selections on candidate soft π-,+
   Configurable<float> softPiEtaMax{"softPiEtaMax", 0.9f, "Soft pion max value for pseudorapidity (abs vale)"};
+  Configurable<float> softPiChi2Max{"softPiChi2Max", 36.f, "Soft pion max value for chi2 ITS"};
   Configurable<int> softPiItsHitMap{"softPiItsHitMap", 127, "Soft pion ITS hitmap"};
   Configurable<int> softPiItsHitsMin{"softPiItsHitsMin", 1, "Minimum number of ITS layers crossed by the soft pion among those in \"softPiItsHitMap\""};
   Configurable<float> softPiDcaXYMax{"softPiDcaXYMax", 0.065, "Soft pion max dcaXY (cm)"};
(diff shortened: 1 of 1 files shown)


PWGHF/TableProducer/candidateCreatorSigmac0plusplus.cxx
```diff
@@ -129,131 +143,201 @@ struct HfCandidateCreatorSigmac0plusplus {
...
+    }
+
+  } /// end createSigmaC
+
+  /// @brief process function for Σc0,++ → Λc+(→pK-π+) π- candidate reconstruction considering also time-reassigned tracks for soft pions
+  /// @param collisions are o2::aod::Collisions
+  /// @param trackIndices are the indices of tracks reassigned to collisions, as obtained from the track-to-collision-associator
+  /// @param tracks are the tracks (with dcaXY, dcaZ information) → soft-pion candidate tracks
+  /// @param candidates are 3-prong candidates satisfying the analysis selections for Λc+ → pK-π+ (and charge conj.)
+  void processDataTimeAssoc(aod::Collisions const& collisions,
```
fgrosa, line 286: I would change the name of this process function, since with the `trackToCollisionAssociation` one can also have the standard association and not the time based one. For instance you could use `processDataTrackToCollAssoc`

PWGHF/TableProducer/candidateCreatorSigmac0plusplus.cxx
```diff
@@ -129,131 +143,201 @@ struct HfCandidateCreatorSigmac0plusplus {
...
+        createSigmaC<true>(collision, trackSoftPi, tracks, candidates, bcWithTimeStamps);
+
+      } /// end loop over tracks for soft pion
 
-        /// loop over tracks
-        auto trackIdsThisCollision = trackIndices.sliceBy(trackIndicesPerCollision, thisCollId);
-        for (const auto& trackId : trackIdsThisCollision) {
+    } // end loop over collisions
+  }
+  PROCESS_SWITCH(HfCandidateCreatorSigmac0plusplus, processDataTimeAssoc, "Process data using also time-reassociated tracks", true);
```
fgrosa, line 314: Same comment about `trackToCollisionAssociation` nomenclature here

PWGHF/TableProducer/candidateCreatorSigmac0plusplus.cxx
```diff
@@ -129,131 +143,201 @@ struct HfCandidateCreatorSigmac0plusplus {
...
 
-          auto trackSoftPi = trackId.track_as<aod::TracksWDcaExtra>();
-          // auto trackSoftPi = tracks.rawIteratorAt(trackId.trackId());
-          histos.fill(HIST("hCounter"), 4);
+  /// @brief process function for Σc0,++ → Λc+(→pK-π+) π- candidate reconstruction without considering time-reassigned tracks for soft pions
+  /// @param collision is a o2::aod::Collision
+  /// @param tracks are the tracks (with dcaXY, dcaZ information) → soft-pion candidate tracks
+  /// @param candidates are 3-prong candidates satisfying the analysis selections for Λc+ → pK-π+ (and charge conj.)
+  void processDataNoTimeAssoc(aod::Collision const& collision,
```
fgrosa, line 320: Same comment about `trackToCollisionAssociation` nomenclature here

PWGHF/TableProducer/candidateCreatorSigmac0plusplus.cxx
```diff
@@ -129,131 +143,201 @@ struct HfCandidateCreatorSigmac0plusplus {
...
-
-    } /// end loop over collisions
-
-  } /// end processWithAmbTracks
+      /// create SigmaC candidate with the current soft pion and Lc candidates
+      createSigmaC<false>(collision, trackSoftPi, tracks, candidates, bcWithTimeStamps);
+
+    } /// end loop over tracks for soft pion
+  }
+  PROCESS_SWITCH(HfCandidateCreatorSigmac0plusplus, processDataNoTimeAssoc, "Process data using time reassociated tracks", false);
```
fgrosa, line 340: Same comment about `trackToCollisionAssociation` nomenclature here

PWGHF/TableProducer/candidateCreatorSigmac0plusplus.cxx
```diff
@@ -129,131 +143,201 @@ struct HfCandidateCreatorSigmac0plusplus {
...
+      /// therefore we need to calculate the dcaXY, dcaZ with respect to this new primary vertex
+      auto bc = collision.bc_as<o2::aod::BCsWithTimestamps>();
+      initCCDB(bc, runNumber, ccdb, isRun2Ccdb ? ccdbPathGrp : ccdbPathGrpMag, lut, isRun2Ccdb);
+      auto trackParSoftPi = getTrackPar(trackSoftPi);
+      o2::gpu::gpustd::array<float, 2> dcaInfo{-999., -999.};
+      o2::base::Propagator::Instance()->propagateToDCABxByBz({collision.posX(), collision.posY(), collision.posZ()}, trackParSoftPi, 2.f, noMatCorr, &dcaInfo);
+      if (std::abs(dcaInfo[0]) > softPiDcaXYMax || std::abs(dcaInfo[1]) > softPiDcaZMax) {
+        return;
+      }
+    }
```
fgrosa, line 266: In principle this is only needed in case you use the reassociated tracks, you could add a 
```
if constexpr (withTimeAssoc) {
} else {
}
```
To skip it in case you don't use reassociated tracks

fgrosa (changes requested):
Hi @mfaggin the PR is essentially good for me, except for a couple of details. I just wanted to point you out that you can also have the "standard association" using the `trackToCollisionAssociation` tables and hence without changing the code.

mfaggin:
Thanks @fgrosa for the suggestions. Two things:

1. with the latest commit I should have properly addressed your requests;
2. about the possibility to use the standard association: you are right, indeed this is a possibility. I tried locally to run the full chain with it and it correctly worked. A couple of points in favour of keeping what I did for this PR are:
   3.1 using derived data should be able to completely remove the track-to-collision-associator dependency at all if I do not want the time re-association for my soft pion
   3.2 in this way one can still keep the time re-association for Lc daughters but avoid doing it for soft pion candidates (maybe a NLO effect, but still a difference)

fgrosa:
> Thanks @fgrosa for the suggestions. Two things:

Hi @mfaggin thanks!
Just a clarification: my comment was not aiming at avoiding this development, but just for you to consider also the other option. 

>     1. with the latest commit I should have properly addressed your requests;
> 
>     2. about the possibility to use the standard association: you are right, indeed this is a possibility. I tried locally to run the full chain with it and it correctly worked. A couple of points in favour of keeping what I did for this PR are:
>        3.1 using derived data should be able to completely remove the track-to-collision-associator dependency at all if I do not want the time re-association for my soft pion
Indeed, however the usage of the the track-to-collision-associator with standard association might have the advantage that you reassociate only the tracks you are interested to and it reduces the combinatorial when building the Sc candidates (and this might give you a better CPU time performance).

>        3.2 in this way one can still keep the time re-association for Lc daughters but avoid doing it for soft pion candidates (maybe a NLO effect, but still a difference)
Indeed, even if this only applies when you don't run over derived data.

fgrosa (approved):

PR #3706, opened 2023-10-27
PWGCF: update to femto tutorial

ariedel-cern:
Update the tutorial to focus more on femto frameworks themselves and not so much on mixing techniques.


victor-gonzalez (approved):
Thanks!!!

PR #3724, opened 2023-10-30
Adding selection of strange-decay particles.

arossi81:
Adding selection of strange-decay particles and fixing displacement selection at generator level.

### DIFF
--- a/DPG/Tasks/AOTTrack/MonitorFilterBit.cxx
+++ b/DPG/Tasks/AOTTrack/MonitorFilterBit.cxx
@@ -114,45 +114,44 @@ struct CheckFilterBit {
   }
   PROCESS_SWITCH(CheckFilterBit, processData, "process data", true);
 
-    template <typename T> int isFromStrangeDecay(const T& particlesMC,const typename T::iterator& particle){
+  template <typename T>
+  int isFromStrangeDecay(const T& particlesMC, const typename T::iterator& particle)
+  {
 
     std::vector<std::vector<int64_t>> arrayIds{};
     std::vector<int64_t> initVec{particle.globalIndex()};
     arrayIds.push_back(initVec);
-    
-    int stage=0;
-    int strangeness=0;
-    while(stage<4 && arrayIds[stage].size()>0 && strangeness==0){
+
+    int stage = 0;
+    int strangeness = 0;
+    while (stage < 4 && arrayIds[stage].size() > 0 && strangeness == 0) {
       std::vector<int64_t> arrayIdsStage{};
-      for(auto &iPart : arrayIds[stage]){
-	auto particleMother=particlesMC.rawIteratorAt(iPart-particlesMC.offset());
-	if(particleMother.has_mothers()){
-	  for(auto iMother=particleMother.mothersIds().front();iMother<=particleMother.mothersIds().back();++iMother){
-	    if(std::find(arrayIdsStage.begin(),arrayIdsStage.end(),iMother)!=arrayIdsStage.end()){
-	      continue;
-	    }
-	    auto mother=particlesMC.rawIteratorAt(iMother-particlesMC.offset());
-	    auto motherPDG=std::abs(mother.pdgCode());
-	    if(motherPDG/1000==3){//strange baryon
-	      strangeness=1;
-	      break;	      
-	    }
-	    else if(motherPDG/1000==0 && motherPDG/100==3){
-	      strangeness=1;
-	      break;
-	    }
-	    arrayIdsStage.push_back(iMother);
-	  }
-	}
-	arrayIds.push_back(arrayIdsStage);
-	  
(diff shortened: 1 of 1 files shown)


DPG/Tasks/AOTTrack/MonitorFilterBit.cxx
```diff
@@ -114,45 +114,44 @@ struct CheckFilterBit {
...
-	    arrayIdsStage.push_back(iMother);
-	  }
-	}
-	arrayIds.push_back(arrayIdsStage);
-	  
+      for (auto& iPart : arrayIds[stage]) {
+        auto particleMother = particlesMC.rawIteratorAt(iPart - particlesMC.offset());
+        if (particleMother.has_mothers()) {
+          for (auto iMother = particleMother.mothersIds().front(); iMother <= particleMother.mothersIds().back(); ++iMother) {
+            if (std::find(arrayIdsStage.begin(), arrayIdsStage.end(), iMother) != arrayIdsStage.end()) {
```
mfaggin, line 133: can you remind me in which cases:

- the number of mothers can be more than 1 (is it when a particle is created in a scattering? I do not see this with a decay)
- among the list of mothers, two of them can be the same, i.e. have the same index?

DPG/Tasks/AOTTrack/MonitorFilterBit.cxx
```diff
@@ -114,45 +114,44 @@ struct CheckFilterBit {
...
     std::vector<int64_t> initVec{particle.globalIndex()};
     arrayIds.push_back(initVec);
-    
-    int stage=0;
-    int strangeness=0;
-    while(stage<4 && arrayIds[stage].size()>0 && strangeness==0){
+
+    int stage = 0;
+    int strangeness = 0;
+    while (stage < 4 && arrayIds[stage].size() > 0 && strangeness == 0) {
```
mfaggin, line 127: I do not understand why you are adding the condition `stage < 4` here. I guess it is a custom value (i.e. you check if a particle comes from a strange decay up to 4 stages above). Also, I see `arrayIds[stage].size() > 0` dangerous, because it might happen that you never call `arrayIds.push_back(arrayIdsStage);` because e.g. you `break;` before, and you might have unpredictable behaviours for `arrayIds[stage]`.
What about writing `while (stage < arrayIds.size() && strangeness == 0)` ?

Let me know if I'm fully misunderstanding something.

DPG/Tasks/AOTTrack/MonitorFilterBit.cxx
```diff
@@ -114,45 +114,44 @@ struct CheckFilterBit {
...
+              continue;
+            }
+            auto mother = particlesMC.rawIteratorAt(iMother - particlesMC.offset());
+            auto motherPDG = std::abs(mother.pdgCode());
+            if (motherPDG / 1000 == 3) { // strange baryon
+              strangeness = 1;
+              break;
+            } else if (motherPDG / 1000 == 0 && motherPDG / 100 == 3) {
+              strangeness = 1;
+              break;
```
mfaggin, line 143: same comment as above: wouldn't it be better to do `return 1`?

DPG/Tasks/AOTTrack/MonitorFilterBit.cxx
```diff
@@ -114,45 +114,44 @@ struct CheckFilterBit {
   }
   PROCESS_SWITCH(CheckFilterBit, processData, "process data", true);
 
-    template <typename T> int isFromStrangeDecay(const T& particlesMC,const typename T::iterator& particle){
+  template <typename T>
+  int isFromStrangeDecay(const T& particlesMC, const typename T::iterator& particle)
```
mfaggin, line 118: minor point: can it be a `bool`? Or are you foreseeing new additional develoments for you you might need more values than `0` or `1`?

mfaggin (commented):
Ciao @arossi81 thanks for your work. I have some questions, I think I'm overlooking/misunderstanding something in your implementation. Can you please clarify the points that I raised in my comments? We can also discuss them in person.

DPG/Tasks/AOTTrack/MonitorFilterBit.cxx
arossi81: ciao Mattia,
1-e.g. if a particle is produced in the material. 
2- this can happen at partonic level (e.g. shower process) but I guess (though I never checked it) also at hadronic level if  e.g. a given mother interacts multiple times with the material, which implies a change of momentum (e.g. suppose that a Xi+ passes through two ITS layers before decaying, its daughter might have multiple Xi+ in the decay tree, which however would point then to the same initial Xi). However, I admit I am not sure about this second point. Anyhow, these protections are quite conservative.

mfaggin:
Let's keep in mind these points forthe next developments

mfaggin (approved):

PR #3725, opened 2023-10-30
Add new PID function in TrackSelectorPID

ZFederica:
Add new function for PID selection to implement he following flow (more conservative than the existing ones):
- reject PID hypothesis if both TPC and TOF information are not stored
- reject PID hypothesis if both TPC and TOF reject it
- accept in every other case

### DIFF
--- a/Common/Core/TrackSelectorPID.h
+++ b/Common/Core/TrackSelectorPID.h
@@ -13,6 +13,7 @@
 /// \brief PID track selector class
 ///
 /// \author Vít Kučera <vit.kucera@cern.ch>, CERN
+/// \author Federica Zanone <federica.zanone@cern.ch>, Heidelberg University
 
 #ifndef COMMON_CORE_TRACKSELECTORPID_H_
 #define COMMON_CORE_TRACKSELECTORPID_H_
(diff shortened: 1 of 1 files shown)


vkucera:
Hi @ZFederica , can you please elaborate more on why this strategy is needed and why you don't consider the track pT?

fcatalan92 (changes requested):
@ZFederica In general the approach LGTM. I made a few minor comments. 
Please report in the D2H your studies using this PID strategy (also after this PR is merged)

@vkucera I think it is fine to have this additional PID approach in the code. Which strategy to use in analysis should be discussed in the relevant PAG meetings.

ZFederica:
Hello @fcatalan92, thank for your suggestions, I have implemented all of them. I will report in the D2H next week on Tuesday if it's okay, so that I will also present some more results on MC besides the PID methods check.

ZFederica:
Hello @vkucera, I simply tested this approach on data and it turned out to give good results

ZFederica:
The build is failing because of errors that are in parts of the code I have not touched (and I merged the latest commits in my branch). Does anyone know how to fix them? Yesterday, at the build check, they were not showing up

vkucera:
> The build is failing because of errors that are in parts of the code I have not touched (and I merged the latest commits in my branch). Does anyone know how to fix them? Yesterday, at the build check, they were not showing up

It must be some changes you added after opening the PR, probably an extra/missing brace.
Did you compile the code locally before updating your PR?

vkucera:
> @ZFederica In general the approach LGTM. I made a few minor comments. Please report in the D2H your studies using this PID strategy (also after this PR is merged)
> 
> @vkucera I think it is fine to have this additional PID approach in the code. Which strategy to use in analysis should be discussed in the relevant PAG meetings.

@fcatalan92 Shouldn't this PID strategy be **first** discussed and approved in the PAG and **then** implemented in the central code?

ZFederica:
> > The build is failing because of errors that are in parts of the code I have not touched (and I merged the latest commits in my branch). Does anyone know how to fix them? Yesterday, at the build check, they were not showing up
> 
> It must be some changes you added after opening the PR, probably an extra/missing brace. Did you compile the code locally before updating your PR?

Thanks @vkucera! I actually missed a bracket

fgrosa:
Hi @vkucera this PID strategy is also adopted in the HF triggers, so I don't see any stopper in adding it here (we might even then consider to use this class in the `HFFilter` task to have a uniform code). In general it is a very conservative PID selection which might also be used as prefilter of ML-based selections. The only doubt that I still have is that the same selection can be obtained with this method https://github.com/AliceO2Group/O2Physics/blob/a0a7bbbcd893759d2194ae897eeed1e348ff5018/Common/Core/TrackSelectorPID.h#L394 with an appropriate configuration of the `TrackSelectorPID::Conditional` status. Don't you think it's the case @vkucera and @ZFederica?

ZFederica:
Hi @fgrosa, I preferred to implement a separate function because to me it looks more transparent: it is completely independent on pt selections and on the TrackSelectorPID::Conditional cases. 
Moreover, if I did not miss anything, I think that using the function statusTpcOrTof does not allow for isolating the cases in which tracks do not have both TPC and TOF information.

fgrosa:
> Hi @fgrosa, I preferred to implement a separate function because to me it looks more transparent: it is completely independent on pt selections and on the TrackSelectorPID::Conditional cases. Moreover, if I did not miss anything, I think that using the function statusTpcOrTof does not allow for isolating the cases in which tracks do not have both TPC and TOF information.

Hi @ZFederica, while I agree on having the strategy you want to use, I think it would be better not to reimplement something that already exists. Moreover, if we do not have TPC and TOF information, the `statusTpcOrTof()` should return PID not applicable as in your function and you can accept this case in your task (btw this should be an extremely small number of cases, even cases without TPC should be a very small fraction). The conditional PID can be turned off by setting this to `0` if I am not wrong https://github.com/AliceO2Group/O2Physics/blob/1ba10d797de204202bc48100d8f5f498a6030caf/Common/Core/TrackSelectorPID.h#L86. For the pt selections you can simply set `0` to a very large number. Can you please test this configuration? Thanks a lot!

PR #3726, opened 2023-10-30
PWGHF: O2AT, adding tracks-extra and zdc converter workflows.

DelloStritto:
- The tracks-extra-converter workflow is needed after the merge of this PR:
https://github.com/AliceO2Group/AliceO2/pull/12087

- The zdc-converter workflow is needed if we decide to run the exercise on the the MC charm enriched production LHC22b1b, that would allow to see a peak also running on very small files.

@vkucera @fcolamar let me know if you agree on doing the exercise on a MC production, otherwise, I can remove the zdc-converter, but I guess we will need anyway the tracks-extra converter

### DIFF
--- a/Tutorials/PWGHF/dpl-config_skim.json
+++ b/Tutorials/PWGHF/dpl-config_skim.json
@@ -18,6 +18,8 @@
         "ccdb-url": "http://alice-ccdb.cern.ch",
         "isRun2MC": "false"
     },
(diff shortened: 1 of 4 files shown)


vkucera:
Merging to make the official skeleton work out of the box.
We will provide the latest configuration in the tutorial for participants with older O2Physics versions,

PR #3734, opened 2023-10-31
templating the rest

abilandz:


PWGCF/MultiparticleCorrelations/Core/MuPa-MemberFunctions.h
```diff
@@ -1119,65 +1057,79 @@ void ResetEventByEventQuantities()
...
+  if (gProcessSim) {
+    CollisionSim const* collisionSim = reinterpret_cast<CollisionSim const*>(collision);
+    // CollisionSim const& collision = *collisionSim;
+    //   *) Impact parameter:
+    if ((collisionSim->impactParameter() < ceh_a.fEventCuts[eImpactParameter][eMin]) ||
+        (collisionSim->impactParameter() > ceh_a.fEventCuts[eImpactParameter][eMax])) {
+      return kFALSE;
+    }
+    // ...
+  } // if (gProcessSim) {
```
victor-gonzalez, line 1123: Have you checked that this actually executes correctly for an MC datafile at reco and gen levels?
To me you are creating a shortcut for the template mechanism which is based on execution code which, in my opinion, is quite dangerous. You are doing at execution time what the compiler can do at compiling time in a much more secure way

PWGCF/MultiparticleCorrelations/Core/MuPa-MemberFunctions.h
```diff
@@ -1119,65 +1057,79 @@ void ResetEventByEventQuantities()
...
   return kTRUE;
 
-} // template <typename T> Bool_t EventCuts(aod::Collision const& collision, T const& tracks)
+} // template <typename T1, typename T2> Bool_t EventCuts(T1 const* collision, T2 const* tracks)
 
 //============================================================
 
 template <eRecSim rs, typename T1, typename T2>
-void FillEventHistograms(T1 const& collision, T2 const& tracks, eBeforeAfter ba)
+void FillEventHistograms(T1 const* collision, T2 const* tracks, eBeforeAfter ba)
```
victor-gonzalez, line 1132: That's the great point in using templates!
You can keep the reference

PWGCF/MultiparticleCorrelations/Core/MuPa-MemberFunctions.h
```diff
@@ -1189,36 +1141,37 @@ void FillEventHistograms(T1 const& collision, T2 const& tracks, eBeforeAfter ba)
...
-    ceh_a.fEventHistograms[eVertex_x][eRec][ba]->Fill(collision.posX());
-    ceh_a.fEventHistograms[eVertex_y][eRec][ba]->Fill(collision.posY());
-    ceh_a.fEventHistograms[eVertex_z][eRec][ba]->Fill(collision.posZ());
-    ceh_a.fEventHistograms[eNContributors][eRec][ba]->Fill(collision.numContrib());
   } // if constexpr (rs == eRec) {
 
   // c) Only simulated:
   if constexpr (rs == eSim) {
+    // c0) collision:
+    CollisionSim const* collisionSim = reinterpret_cast<CollisionSim const*>(collision);
```
victor-gonzalez, line 1168: Same comment

PWGCF/MultiparticleCorrelations/Core/MuPa-MemberFunctions.h
```diff
@@ -2499,7 +2239,8 @@ void CalculateEverything()
 
 //============================================================
 
-void SteerRec(CollisionRec const& collision, TracksRec const& tracks)
+template <typename T1, typename T2>
+void Steer(T1 const* collision, T2 const* tracks)
```
victor-gonzalez, line 2243: Consider templating also over the reco sim `enum`  that will simplify a lot your code and you just pass it to your function calls

PWGCF/MultiparticleCorrelations/Core/MuPa-MemberFunctions.h
```diff
@@ -2510,24 +2251,45 @@ void SteerRec(CollisionRec const& collision, TracksRec const& tracks)
...
+
+  // *) Fill event histograms before event cuts:
+  if (gProcessRec) {
+    FillEventHistograms<eRec>(collision, tracks, eBefore);
+  } else if (gProcessRecSim) {
+    FillEventHistograms<eRec>(collision, tracks, eBefore);
+    FillEventHistograms<eSim>(collision, tracks, eBefore);
+  } else if (gProcessSim) {
+    FillEventHistograms<eSim>(collision, tracks, eBefore);
+  }
```
victor-gonzalez, line 2264: If you adapt your `FillEventHistograms` function for the `eRecSim` value you will need only a single function call in here. As was indicated before in the previous PR, with the full template mechanism you don't need the global variables which need to be set via configuration

PWGCF/MultiparticleCorrelations/Core/MuPa-MemberFunctions.h
```diff
@@ -2510,24 +2251,45 @@ void SteerRec(CollisionRec const& collision, TracksRec const& tracks)
...
-  FillEventHistograms<eRec>(collision, tracks, eAfter);
+  // *) Fill event histograms after event cuts:
+  if (gProcessRec) {
+    FillEventHistograms<eRec>(collision, tracks, eAfter);
+  } else if (gProcessRecSim) {
+    FillEventHistograms<eRec>(collision, tracks, eAfter);
+    FillEventHistograms<eSim>(collision, tracks, eAfter);
+  } else if (gProcessSim) {
+    FillEventHistograms<eSim>(collision, tracks, eAfter);
+  }
```
victor-gonzalez, line 2279: Same comment as above

victor-gonzalez (changes requested):
The PR cannot be approved as it is
It is creating a way of coding which is dangerous and should not be propagated to the newcomers
Please, take the actions suggested and if you believe your way of coding is the one to follow consider presenting your motivation at the WP4+WP14 meeting on Mondays

PWGCF/MultiparticleCorrelations/Core/MuPa-MemberFunctions.h
abilandz: Yes, i tested it successfully with MC datafile at reco and gen levels, but only locally. Still this has to be checked in Hyperloop.

I wonder why you are arguing that this is dangerous - do you have a concrete example at hand when this approach would fail?

I recalled now that I already used in previous PRs casting in the code, i.e. at the line:

`pw_a.fWeightsHist[ppe] = reinterpret_cast<TH1D*>(hist->Clone());`

The context is a bit different, though. But here for instance casting is mandatory, because Clone() returns a pointer to TObject (please see this [link](https://root.cern.ch/doc/master/classTH1.html#ae6c9d17415e3fa37a342f8b0c68602b5)). Would you argue that here also casting is dangerous?

Finally, MegaLinter doesn't complain to using reinterpret_cast.

PWGCF/MultiparticleCorrelations/Core/MuPa-MemberFunctions.h
abilandz: I switched to pointers, because of casting. 

In general, I am not against templates. But also, I do not want to pack my analysis code ONLY with templates, C++ offers many more functionalities that we should cherish as well...

PWGCF/MultiparticleCorrelations/Core/MuPa-MemberFunctions.h
```diff
@@ -1189,36 +1141,37 @@ void FillEventHistograms(T1 const& collision, T2 const& tracks, eBeforeAfter ba)
...
+  ceh_a.fEventHistograms[eVertex_x][rs][ba]->Fill(collision->posX());
+  ceh_a.fEventHistograms[eVertex_y][rs][ba]->Fill(collision->posY());
+  ceh_a.fEventHistograms[eVertex_z][rs][ba]->Fill(collision->posZ());
+  // a1) tracks:
 
   // b) Only reconstructed:
   if constexpr (rs == eRec) {
-    ceh_a.fEventHistograms[eTotalMultiplicity][eRec][ba]->Fill(tracks.size());
+    // b0) collision:
+    CollisionRec const* collisionRec = reinterpret_cast<CollisionRec const*>(collision);
```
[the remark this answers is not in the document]
abilandz: That was my first attempt, but I didn't manage to compile the code that way.

For instance, if in the function argument I have a template `T1 const* collision` and then in the function body hardcoded `collision->impactParameter()` , that didn't compile when 'collision' is aod::Collision, because that one doesn't have impactParameter() member function. 

The cast mechanism which I have implemented overcome this problem.

PWGCF/MultiparticleCorrelations/Core/MuPa-MemberFunctions.h
abilandz: See my reply above - I am not in general against templates, but I do not want literally to have only templates in my code.

PWGCF/MultiparticleCorrelations/Core/MuPa-MemberFunctions.h
abilandz: I need these global variables also for other reasons at the moment in the code. 

I can remove them eventually, but that will trigger a lot of other modifications in the code.

abilandz:
> The PR cannot be approved as it is It is creating a way of coding which is dangerous and should not be propagated to the newcomers Please, take the actions suggested and if you believe your way of coding is the one to follow consider presenting your motivation at the WP4+WP14 meeting on Mondays

I do not quite get this comment. Probability is literally 0 that newcomers will be learning from my code how to develop code in O2, given the fact that there are really great O2 tutorials and documentation written for the newcomers (and here "great" really means great!).
 
In case you are arguing that casting mechanism should not be used at all in O2, then you need to elaborate why you consider it to be dangerous, with a few clear-cut examples.

Now we have situation that code compiles without any warnings, it runs efficiently both over rec and sim, MegaLinter doesn't complain anything, and PR is nevertheless rejected because code is potentially dangerous, without clarifying what is the danger...

victor-gonzalez:
>> I do not quite get this comment. Probability is literally 0 that newcomers will be learning from my code how to develop code in O2, given the fact that there are really great O2 tutorials and documentation written for the newcomers (and here "great" really means great!).

We don't want wrong pattern of use over common resources. People interested in multi-particle correlations will definitely read this code

>> In case you are arguing that casting mechanism should not be used at all in O2, then you need to elaborate why you consider it to be dangerous, with a few clear-cut examples.

Casting is not forbidden at all and much less in O2. The way it used for addressing common resources is what is dangerous. No example is needed. As any SW book that covers its evolution will say, the used constructions are really old and caused a lot of nightmares. Templates were developed to leave that burden to the compilers.

>> That was my first attempt, but I didn't manage to compile the code that way.

>> For instance, if in the function argument I have a template T1 const* collision and then in the function body hardcoded collision->impactParameter() , that didn't compile when 'collision' is aod::Collision, because that one doesn't have impactParameter() member function.

>> The cast mechanism which I have implemented overcome this problem.

Please, use the suggested below construction instead. The casting is not needed at all.

In your `EventCuts` function
```
template <eRecSim rs, typename T1, typename T2>
Bool_t EventCuts(T1 const& collision, T2 const& tracks)
{
  ..............................

  // b) Event cuts on info available only in reconstructed data:
  if constexpr (rs == eRec || rs == eRecAndSim) {
    //   *) NContributors:
    if ((collision.numContrib() < ceh_a.fEventCuts[eNContributors][eMin]) ||
        (collision.numContrib() > ceh_a.fEventCuts[eNContributors][eMax])) {
      return kFALSE;
    }
    // ...
  } // if (gProcessRec || gProcessRecSim) {

  // c) Event cuts on info available only in simulated data:
  if constexpr (rs == eSim) {
    if ((collision->impactParameter() < ceh_a.fEventCuts[eImpactParameter][eMin]) ||
        (collision->impactParameter() > ceh_a.fEventCuts[eImpactParameter][eMax])) {
      return kFALSE;
    }
    // ...
  } // if (gProcessSim) {

  return kTRUE;

} // template <eRecSim rs, typename T1, typename T2> Bool_t EventCuts(T1 const* collision, T2 const* tracks)
```

then in your `Steer` function use this construction
```
template <eRecSim rs, typename T1, typename T2>
void Steer(T1 const& collision, T2 const& tracks)
{
  ....................
  // *) Event cuts:
  if (!EventCuts<rs>(collision, tracks)) {
    return;
  }
  .................
}
```
Of course if you slightly modify your `FillEventHistograms` function to 
```
  .................
  if constexpr (rs == eRec || rs == eRecAndSim) {
  .................
  if constexpr (rs == eSim || rs == eRecAndSim) {
```
then your `Steer` function might stay as
```
template <eRecSim rs, typename T1, typename T2>
void Steer(T1 const& collision, T2 const& tracks)
{
  ....................
  // *) Fill event histograms before event cuts:
  FillEventHistograms<rs>(collision, tracks, eBefore);

  // *) Event cuts:
  if (!EventCuts<rs>(collision, tracks)) {
    return;
  }

  // *) Fill event histograms after event cuts:
  FillEventHistograms<rs>(collision, tracks, eAfter);
  .................
}
```
Please apply the same approach to your `MainLoopOverParticles` function
```
template <eRecSim rs, typename T>
void MainLoopOverParticles(T const& tracks)
{
  .....................
}
```
and then removing the casting over the tracks tables using the `if constexpr ()` construction 

In this last case I would suggest you to not replicate the code.
Your processing of `eRec` and `eRecAndSim` is basically identical  with the exception of additional code when is `eRecAndSim`.
Consider this construction
```
  if constexpr (rs == eRec || rs == eRecAndSim) {
    /* NO CASTING NEEDED */
    for (auto& track : tracks) {
      /* THE FIRST PART OF YOUR RECONSTRUCTED TRACKS PROCESSING */
      ......
      if constexpr (rs == eRecAndSim) {
        /* NO CASTING NEEDED */
        /* THE FIRST PART OF YOUR MC RECONSTRUCTED TRACKS PROCESSING */
        .......
      }
      /* THE SECOND PART OF YOUR RECONSTRUCTED TRACKS PROCESSING */
      ......
      if constexpr (rs == eRecAndSim) {
        /* NO CASTING NEEDED */
        /* THE SECOND PART OF YOUR MC RECONSTRUCTED TRACKS PROCESSING */
        .....
      }
      /* THE FINAL PART OF YOUR RECONSTRUCTED TRACKS PROCESSING */
    }
  }
```
Have in mind that the `if`s have not execution penalties because they are not actually there in the executable.
The compiler takes care of replicating the code for the different cases of `template` implementation and no `if`s are executed.

victor-gonzalez (approved):
Thanks for addressing my comments and suggestions!!!
Let's hope that for the next iteration the code replication in `MainLoopOverParticles` will also be addressed

PR #3735, opened 2023-10-31
PWG-LF: Remove table dependency

BongHwi:
Remove unused table subscriptions for stand-alone processing.


alibuild (approved):

PR #3743, opened 2023-10-31
Change QA options

ddobrigk:
This PR adds a QA histogram to the cascade pre-selector such that we can now cross-check how many cascades where actually in a given AO2D versus how many were selected for actually being built. With this histogram, it will be simpler to diagnose cascade building from the base `Cascades` table all the way to the more analysis-geared `CascDatas` table.


njacazio (approved):

njacazio:
This needs https://github.com/ddobrigk/O2Physics/pull/181 merged first

njacazio (approved):

PR #3753, opened 2023-11-01
RecoDecay: Remove getMassPDG

vkucera:
This is the last step of the `RecoDecay` part of the ongoing effort to optimise memory consumption and to unify the location from where we pick the PDG masses. The ultimate goal is to get the masses from [`CommonConstants/PhysicsConstants.h`](https://github.com/AliceO2Group/AliceO2/blob/dev/Common/Constants/include/CommonConstants/PhysicsConstants.h) (when the PDG code is known at compile time) or from the `O2DatabasePDG` service (otherwise).

See details in the dedicated issue: https://github.com/AliceO2Group/O2Physics/issues/2178

**Notes for code owners and developers:**

- Several files across O2Physics have been affected by the removal of the dependency of `RecoDecay` on `TDatabasePDG.h` and `TPDGCode.h` because they implicitly relied on it. This can be avoided if each file **explicitly** includes the headers for the things it **explicitly** needs (in this case `TDatabasePDG::Instance()` and the named PDG codes). It's a reminder for me that we should all try to pay attention to these details when writing and reviewing the code.
- I have added `FIXME` comments on the lines which should be modified in line with the aforementioned unification goal. Calling `TDatabasePDG::Instance()` each time a mass is requested causes a sizeable and unnecessary overhead in CPU and memory which can be avoided.

### DIFF
--- a/ALICE3/TableProducer/OTF/onTheFlyRICHPID.cxx
+++ b/ALICE3/TableProducer/OTF/onTheFlyRICHPID.cxx
@@ -15,6 +15,9 @@
 
 #include <utility>
 #include <cmath>
+
+#include <TPDGCode.h>
+
 #include "Framework/AnalysisDataModel.h"
 #include "Framework/AnalysisTask.h"
 #include "Framework/runDataProcessing.h"
--- a/ALICE3/TableProducer/OTF/onTheFlyTOFPID.cxx
+++ b/ALICE3/TableProducer/OTF/onTheFlyTOFPID.cxx
@@ -14,6 +14,9 @@
 //
 
 #include <utility>
+
+#include <TPDGCode.h>
+
 #include "Framework/AnalysisDataModel.h"
 #include "Framework/AnalysisTask.h"
 #include "Framework/runDataProcessing.h"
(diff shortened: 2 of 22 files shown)


ALICE3/TableProducer/OTF/onTheFlyRICHPID.cxx
```diff
@@ -15,6 +15,9 @@
 
 #include <utility>
 #include <cmath>
+
+#include <TPDGCode.h>
```
ddobrigk, line 19: Sorry if I miss something - just for my own understanding, was the addition of this header needed? It seems it is the only change in this and the other `onTheFly` task?

ddobrigk: Ah - never mind. It is needed to define variables like `kElectron, kMuonMinus, kPiPlus, kKPlus, kProton` - they existed before but not anymore. Thanks!!!

ddobrigk (approved):
Thanks a lot!

njacazio (approved):

saganatt (approved):

saganatt:
Approved for CF. @zchochul @lauraser @ariedel-cern please be aware in your teams of the FIXMES and the Vit's description of recommended coding practice for obtaining the PDG data.

fcatalan92 (approved):

PR #3757, opened 2023-11-02
PWGLF: Update the first resonance tutorial base

BongHwi:
Basic task to read the resonance table and run track loop


njacazio (approved):
Thanks a lot!!

PR #3771, opened 2023-11-03
[PWGHF] Fixed Ds selection flag

fchinu:
This PR contains:

- Added topological selection flag check before setting PID selection flag bit.

### DIFF
--- a/PWGHF/TableProducer/candidateSelectorDsToKKPi.cxx
+++ b/PWGHF/TableProducer/candidateSelectorDsToKKPi.cxx
@@ -280,10 +280,10 @@ struct HfCandidateSelectorDsToKKPi {
         }
         continue;
       }
-      if (pidDsToKKPi) {
+      if (TESTBIT(statusDsToKKPi, aod::SelectionStep::RecoTopol) && pidDsToKKPi) {
         SETBIT(statusDsToKKPi, aod::SelectionStep::RecoPID);
       }
-      if (pidDsToPiKK) {
+      if (TESTBIT(statusDsToPiKK, aod::SelectionStep::RecoTopol) && pidDsToPiKK) {
         SETBIT(statusDsToPiKK, aod::SelectionStep::RecoPID);
       }
       if (activateQA) {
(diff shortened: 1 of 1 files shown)


PWGHF/TableProducer/candidateSelectorDsToKKPi.cxx
```diff
@@ -280,10 +280,10 @@ struct HfCandidateSelectorDsToKKPi {
         }
         continue;
       }
-      if (pidDsToKKPi) {
+      if (TESTBIT(statusDsToKKPi, aod::SelectionStep::RecoTopol) && pidDsToKKPi) {
```
fcatalan92, line 283:
```suggestion
      if (topolDsToKKPi && pidDsToKKPi) {
```
No need to check the bitmap here since we have directly the `bool` from the topological selection.

PWGHF/TableProducer/candidateSelectorDsToKKPi.cxx
```diff
@@ -280,10 +280,10 @@ struct HfCandidateSelectorDsToKKPi {
...
         SETBIT(statusDsToKKPi, aod::SelectionStep::RecoPID);
       }
-      if (pidDsToPiKK) {
+      if (TESTBIT(statusDsToPiKK, aod::SelectionStep::RecoTopol) && pidDsToPiKK) {
```
fcatalan92, line 286:
```suggestion
      if (topolDsToPiKK && pidDsToPiKK) {
```

fcatalan92 (changes requested):
@fchinu Thanks for the fix, I got that with the current code you can have a candidate not passing the topological selections for a mass hypothesis for which you set the PID flag afterward. Right?

Could you do the same also for the ML part (L312-317)?

PWGHF/TableProducer/candidateSelectorDsToKKPi.cxx
fchinu: Thank you @fcatalan92 for your suggestions! Indeed, we observed candidates which were passing the PID selections without passing the topological ones for a certain mass hypothesis. I have now implemented your suggestions

fcatalan92 (approved):

PR #3772, opened 2023-11-03
PWGMM: dndeta: comply to megalinter: avoid c-style cast

aalkin:


alibuild (approved):

PR #3567, opened 2023-10-08
PWGHF: Dstar Framework Development

deependra170598:
Hi @fgrosa , @prchakra , @ssadhu , @sjaelani ,@fcatalan 
We tried to prepare creator and selector tasks for Dstar analysis. The creator task is almost done except for experts' suggestions and improvements. While in the Selector task, PID selection is to be added and It is under development. Within the next few weeks, I will try to complete it.
Here, I invite you all to suggest improvements in the code. Since It is under progress I am putting this PR in draft.

PR #3781, opened 2023-11-05
PWGHF: Fix D* finding

fgrosa:
Investigating a crash reported by @deependra170598 in the D* creator under development (see PR https://github.com/AliceO2Group/O2Physics/pull/3567), I fixed 3 bugs in the finding of D* mesons:
1) The selection flag of D0 mesons not selected because of track selections was not reset to 0, causing the creation of D* candidates associated to no D0 mesons or the previous last one (@deependra170598 this was the bug causing the crash in your code).
2) Only half of the D* candidates were found because in the 3-prong loops were triangular, however `(trackPos1 trackNeg1)_2prong trackPos2` is not equivalent to `(trackPos2 trackNeg1)_2prong trackPos1` for D* mesons because the pion from D0 and the soft pion are not exchangeable (while they are for 3-prong candidates). 
3) The D* table was filled with last 2-prong table index, however if J/psi mesons are enabled, it could not be the correct one.

I tested the new code on pp MC sample and I verified that
1) the modifications in this PR do not influence the 2-prong and 3-prong creation (while for D* mesons more signal and less background candidates are found)

image

2) he impact on the CPU time is not dramatic

- before fix

image

- after fix

image

Finally, I also modified the PVrefit for D* meson candidates in order not to remove the soft pion (as done in Run2).

### DIFF
--- a/PWGHF/TableProducer/trackIndexSkimCreator.cxx
+++ b/PWGHF/TableProducer/trackIndexSkimCreator.cxx
@@ -10,7 +10,7 @@
 // or submit itself to any jurisdiction.
 
 /// \file trackIndexSkimCreator.cxx
-/// \brief Pre-selection of 2-prong and 3-prong secondary vertices of heavy-flavour decay candidates
+/// \brief Pre-selection of 2-prong, 3-prong, D*, and cascade secondary vertices of heavy-flavour decay candidates
 ///
 /// \author Gian Michele Innocenti <gian.michele.innocenti@cern.ch>, CERN
 /// \author Vít Kučera <vit.kucera@cern.ch>, CERN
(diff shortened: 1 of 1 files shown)


mfaggin (changes requested):
Ciao @fgrosa , I went through your changes and, after digesting them, everything is clear to me. Thank you for these important fixes!

I just suggest you just two very very minor fixes in a couple of comments. Then, green light from my side!

fgrosa:
> Ciao @fgrosa , I went through your changes and, after digesting them, everything is clear to me. Thank you for these important fixes!
> 
> I suggest you just two very very minor fixes in a couple of comments. Then, green light from my side!

Thanks for having checked @mfaggin!

mfaggin (approved):

PR #3786, opened 2023-11-06
PWGHF: change PID method in toXiPi candidateSelector

ZFederica:
- change PID method: statusTpcAndTof --> statusTpcOrTof
- change default values of configurables


fgrosa (approved):
Thanks @ZFederica!

PR #3789, opened 2023-11-06
PWGHF: addition of D0 ML response class

atavirag:
Following @AlexBigO's ML response class for Dplus, I made an equivalent class for D0 mesons. I added the variables currently available in treeCreatorD0ToKPi.cxx.


AlexBigO:
Hello @atavirag , I don't have any comments on the header but I have a few regarding the selector, otherwise it looks good to me :)
1. You should define a configurable array containing the names of the wanted input features: as done [here](https://github.com/AliceO2Group/O2Physics/blob/master/PWGHF/TableProducer/candidateSelectorDplusToPiKPi.cxx#L68)
2. You need to cache the input indices to translate your configurables input features names into integers (via the `enum` defined in the header): you can use [this line](https://github.com/AliceO2Group/O2Physics/blob/master/PWGHF/TableProducer/candidateSelectorDplusToPiKPi.cxx#L122)
3. You should change the way `aod::HfMlD0` is filled when a candidate does not pass preselections. This is a bug spotted by Fabio that we corrected in a recent PR for Dplus. The idea is to define an empty vector ([see here](https://github.com/AliceO2Group/O2Physics/blob/master/PWGHF/TableProducer/candidateSelectorDplusToPiKPi.cxx#L77)) and then feed this vector to `hfMlD0Candidate` when preselections (Skim, Topol, Pid) are not fulfilled by the candidate (as done [here](https://github.com/AliceO2Group/O2Physics/blob/master/PWGHF/TableProducer/candidateSelectorDplusToPiKPi.cxx#L209) for instance).

atavirag:
Hi @AlexBigO, I have just added your comments. Thank you!

PWGHF/TableProducer/candidateSelectorD0.cxx
```diff
@@ -114,6 +116,7 @@ struct HfCandidateSelectorD0 {
       } else {
         hfMlResponse.setModelPathsLocal(onnxFileNames);
       }
+      hfMlResponse.cacheInputFeaturesIndices(namesInputFeatures);
       hfMlResponse.init();
       outputMl.assign(((std::vector<int>)cutDirMl).size(), -1.f); // dummy value for ML output
```
AlexBigO, line 121: Everything is okay on my side. Just a minor comment, now you don't need this line anymore as the `outputMl` vector (defined above `init`) is going to be passed by reference in `isSelectedMl`.

PWGHF/Core/HfMlResponseD0ToKPi.h
```diff
@@ -0,0 +1,264 @@
...
+enum class InputFeaturesD0ToKPi : uint8_t {
+  collisionId = 0,
+  posX,
+  posY,
+  posZ,
+  xSecondaryVertex,
+  ySecondaryVertex,
+  zSecondaryVertex,
+  errorDecayLength,
+  errorDecayLengthXY,
```
fgrosa, line 72: These variables are not needed given that they should not be useful for BDTs

PWGHF/Core/HfMlResponseD0ToKPi.h
```diff
@@ -0,0 +1,264 @@
...
+  chi2PCA,
+  rSecondaryVertex,
```
fgrosa, line 74: I would remove this one too

PWGHF/Core/HfMlResponseD0ToKPi.h
```diff
@@ -0,0 +1,264 @@
...
+  decayLength,
+  decayLengthXY,
+  decayLengthNormalised,
+  decayLengthXYNormalised,
+  impactParameterNormalised0,
+  ptProng0,
+  pt2Prong0,
```
fgrosa, line 81: no need to put the squared momenta (the pt is sufficient)

PWGHF/Core/HfMlResponseD0ToKPi.h
```diff
@@ -0,0 +1,264 @@
...
+  ptProng1,
+  pt2Prong1,
+  pxProng0,
+  pyProng0,
+  pzProng0,
+  pxProng1,
+  pyProng1,
+  pzProng1,
```
fgrosa, line 89: I would say that the single components of the daughter momenta are not needed

PWGHF/Core/HfMlResponseD0ToKPi.h
```diff
@@ -0,0 +1,264 @@
...
+  impactParameter0,
+  impactParameter1,
+  errorImpactParameter0,
+  errorImpactParameter1,
```
fgrosa, line 93: this is also not needed

PWGHF/Core/HfMlResponseD0ToKPi.h
```diff
@@ -0,0 +1,264 @@
...
+  nSigTpcPi0,
+  nSigTpcKa0,
+  nSigTofPi0,
+  nSigTofKa0,
+  nSigTpcPi1,
+  nSigTpcKa1,
+  nSigTofPi1,
+  nSigTofKa1,
+  m,
```
fgrosa, line 102: The mass should never be used as feature, so I would remove it

PWGHF/Core/HfMlResponseD0ToKPi.h
```diff
@@ -0,0 +1,264 @@
...
+  maxNormalisedDeltaIP,
+  impactParameterProduct,
+  cosThetaStar,
+  pt,
+  p,
```
fgrosa, line 107: The momenta of the candidate also should not be used

PWGHF/Core/HfMlResponseD0ToKPi.h
```diff
@@ -0,0 +1,264 @@
...
+  cpa,
+  cpaXY,
+  ct,
+  eta,
+  phi,
+  y,
+  e
```
fgrosa, line 114: Same for these variables

PWGHF/Core/HfMlResponseD0ToKPi.h
```diff
@@ -0,0 +1,264 @@
...
+  errorImpactParameter0,
+  errorImpactParameter1,
+  nSigTpcPi0,
+  nSigTpcKa0,
+  nSigTofPi0,
+  nSigTofKa0,
+  nSigTpcPi1,
+  nSigTpcKa1,
+  nSigTofPi1,
+  nSigTofKa1,
```
fgrosa, line 101: In addition to the single Nsigma TPC and TOF variables, I would also add the combined one (see e.g. https://github.com/AliceO2Group/O2Physics/blob/1204d8f12614b5601381c70138bcdee4224ccf5b/PWGHF/Core/HfMlResponseDplusToPiKPi.h#L78-L83)

PWGHF/Core/HfMlResponseD0ToKPi.h
```diff
@@ -0,0 +1,264 @@
...
+ public:
+  /// Default constructor
+  HfMlResponseD0ToKPi() = default;
+  /// Default destructor
+  virtual ~HfMlResponseD0ToKPi() = default;
+
+  HfHelper hfHelper;
+
+  /// Method to get the input features vector needed for ML inference
+  /// \param candidate is the Dplus candidate
```
fgrosa, line 129:
```suggestion
  /// \param candidate is the D0 candidate
```

PWGHF/Core/HfMlResponseD0ToKPi.h
```diff
@@ -0,0 +1,264 @@
...
+  /// \param prong0 is the candidate's prong0
+  /// \param prong1 is the candidate's prong1
+  /// \return inputFeatures vector
+  template <typename T1, typename T2>
+  std::vector<float> getInputFeatures(T1 const& candidate,
+                                      T2 const& prong0, T2 const& prong1)
+  {
```
fgrosa, line 136: Differently from the Dplus, in the case of the D0 we should handle the two possible mass hypotheses (KPi or PiK). This means that we should have the flexibility to fill the vector of variables with one hypothesis or the other for all the variables that depend on this mass hypothesis, such as the cost*, see https://github.com/AliceO2Group/O2Physics/blob/1204d8f12614b5601381c70138bcdee4224ccf5b/PWGHF/Core/HfHelper.h#L73-L83
 My proposal would be to add an argument here that specifies whether the vector should be filled for the hypothesis `D0toPiK` (first daughter as a pion, second one as a kaon) or the hypothesis `D0bartoKPi` (first daughter as a kaon, second one as a pion)

PWGHF/TableProducer/candidateSelectorD0.cxx
```diff
@@ -357,15 +360,7 @@ struct HfCandidateSelectorD0 {
...
-
-        std::vector<float> inputFeatures{candidate.cpa(),
-                                         candidate.cpaXY(),
-                                         candidate.decayLength(),
-                                         candidate.decayLengthXY(),
-                                         candidate.impactParameter0(),
-                                         candidate.impactParameter1(),
-                                         candidate.impactParameterProduct()};
-
+        std::vector<float> inputFeatures = hfMlResponse.getInputFeatures(candidate, trackPos, trackNeg);
```
fgrosa, line 363: In the hard-coded version we did not have variables that depend on the mass hypothesis. However, now that we have the flexibility to modify the variables via the configurable, we should differentiate the two. In particular, if the candidate is preselected as a `D0toPiK` (`statusD0`) we should get the BDT output scores corresponding to that mass hypothesis, while if selected as `D0bartoKPi` (`statusD0bar`) the other ones. This means that sometimes the `isSelectedMl` method will be called for one of the two, sometimes for both. We should then decide a convention how to store them in the `hfMlD0Candidate` table (@fcatalan92 any idea?)

fgrosa (changes requested):
Hi @atavirag thanks a lot for the implementation! I added few comments regarding variables to remove or add and then the other missing thing to address (not present for the D+) is the way to handle the two possible mass hypotheses for the D0/D0bar candidates.

PWGHF/TableProducer/candidateSelectorD0.cxx
fgrosa: Yes this might indeed be a possibility. In this case we should decide whether to fill the columns with empty vectors in case there are no BDT output scores for one of the hypothesis (based on the `statusD0` or `statusD0bar` before the BDT application) or to always test the BDT for both hypotheses and fill it.

atavirag:
Hi @fgrosa, I just made a commit changing the variables. I also added the column `hf_sel_candidate_d0::MlProbD0bar` and, at least for now, made it so that it always tests the BDT output for both mass hypotheses. Please let me know if it's okay or if you'd prefer the other option to fill the table. Thanks!

PWGHF/TableProducer/candidateSelectorD0.cxx
```diff
@@ -61,21 +62,25 @@ struct HfCandidateSelectorD0 {
...
   Configurable<int64_t> timestampCCDB{"timestampCCDB", -1, "timestamp of the ONNX file for ML model used to query in CCDB"};
   Configurable<bool> loadModelsFromCCDB{"loadModelsFromCCDB", false, "Flag to enable or disable the loading of models from CCDB"};
 
-  o2::analysis::HfMlResponse<float> hfMlResponse;
-  std::vector<float> outputMl = {};
+  o2::analysis::HfMlResponseD0ToKPi<float> hfMlResponse;
+  std::vector<float> outputMlD0 = {};
+  std::vector<float> outputMlD0bar = {};
+  std::vector<float> outputMlNotPreselectedD0 = {};
+  std::vector<float> outputMlNotPreselectedD0bar = {};
```
fgrosa, line 77: There is no need to have these variables duplicated
```suggestion
  std::vector<float> outputMlD0 = {};
  std::vector<float> outputMlD0bar = {};
```

PWGHF/TableProducer/candidateSelectorD0.cxx
```diff
@@ -277,7 +282,7 @@ struct HfCandidateSelectorD0 {
       if (!selectionTopol<reconstructionType>(candidate)) {
         hfSelD0Candidate(statusD0, statusD0bar, statusHFFlag, statusTopol, statusCand, statusPID);
         if (applyMl) {
-          hfMlD0Candidate(outputMl);
+          hfMlD0Candidate(outputMlNotPreselectedD0, outputMlNotPreselectedD0bar);
```
fgrosa, line 285:
```suggestion
          hfMlD0Candidate(outputMlD0, outputMlD0bar);
```

PWGHF/TableProducer/candidateSelectorD0.cxx
```diff
@@ -294,7 +299,7 @@ struct HfCandidateSelectorD0 {
       if (!topolD0 && !topolD0bar) {
         hfSelD0Candidate(statusD0, statusD0bar, statusHFFlag, statusTopol, statusCand, statusPID);
         if (applyMl) {
-          hfMlD0Candidate(outputMl);
+          hfMlD0Candidate(outputMlNotPreselectedD0, outputMlNotPreselectedD0bar);
```
fgrosa, line 302:
```suggestion
          hfMlD0Candidate(outputMlD0, outputMlD0bar);
```

PWGHF/TableProducer/candidateSelectorD0.cxx
```diff
@@ -342,7 +347,7 @@ struct HfCandidateSelectorD0 {
       if (pidD0 == 0 && pidD0bar == 0) {
         hfSelD0Candidate(statusD0, statusD0bar, statusHFFlag, statusTopol, statusCand, statusPID);
         if (applyMl) {
-          hfMlD0Candidate(outputMl);
+          hfMlD0Candidate(outputMlNotPreselectedD0, outputMlNotPreselectedD0bar);
```
fgrosa, line 350:
```suggestion
          hfMlD0Candidate(outputMlD0, outputMlD0bar);
```

PWGHF/TableProducer/candidateSelectorD0.cxx
```diff
@@ -357,29 +362,23 @@ struct HfCandidateSelectorD0 {
...
-                                         candidate.impactParameterProduct()};
-
-        bool isSelectedMl = hfMlResponse.isSelectedMl(inputFeatures, ptCand, outputMl);
-        hfMlD0Candidate(outputMl);
-
-        if (!isSelectedMl) {
+        if (!isSelectedMlD0 && !isSelectedMlD0bar) {
           statusD0 = 0;
           statusD0bar = 0;
         }
```
fgrosa, line 374:
```suggestion
        if (!isSelectedMlD0) {
          statusD0 = 0;
        }
        if (!isSelectedMlD0bar) {
          statusD0bar = 0;
        }
```

PWGHF/TableProducer/candidateSelectorD0.cxx
```diff
@@ -263,7 +268,7 @@ struct HfCandidateSelectorD0 {
       if (!(candidate.hfflag() & 1 << aod::hf_cand_2prong::DecayType::D0ToPiK)) {
         hfSelD0Candidate(statusD0, statusD0bar, statusHFFlag, statusTopol, statusCand, statusPID);
         if (applyMl) {
-          hfMlD0Candidate(outputMl);
+          hfMlD0Candidate(outputMlNotPreselectedD0, outputMlNotPreselectedD0bar);
```
fgrosa, line 271:
```suggestion
          hfMlD0Candidate(outputMlD0, outputMlD0bar);
```

fgrosa, line 271: Before this, e.g. in line `267` you should add
```
outputMlD0.clear();
outputMlD0bar.clear();
```
I cannot put it as a suggestion because that line is not modifed in your PR.

fgrosa (changes requested):
Hi @atavirag thanks! I still have a couple of suggestions, then I would wait for @fcatalan92's feedback on the table with the two vectors and the decision whether to apply anyway always the ML model to both mass hypotheses (as in your implementation) or instead whether to do it only for the hypothesis which is preselected (maybe better).

fcatalan92:
Hi @atavirag and @fgrosa, I would apply the model only to candidates that pass the pre-selections, so that we do not perform the model computation if not needed. I would also leave the vector empty in case the model is not applied, this prevents to use a wrong information later on, since if the vector is accessed we should get a crash (the correct way should be to check the flag of the ML selection before accessing the ML output).

PWGHF/TableProducer/candidateSelectorD0.cxx
```diff
@@ -357,33 +363,40 @@ struct HfCandidateSelectorD0 {
...
+        if (!isSelectedMlD0) {
           statusD0 = 0;
+        }
+        if (!isSelectedMlD0bar) {
           statusD0bar = 0;
         }
+
+        if (isSelectedMlD0 || isSelectedMlD0bar) {
+          hfMlD0Candidate(outputMlD0, outputMlD0bar);
+        }
```
fcatalan92, line 387: You should always fill the table, otherwise it will be not joinable with those of the candidates and selection flag.

fcatalan92 (changes requested):
Hi @atavirag, thanks for the updates. I added a comment on the code.
Moreover, I have another remark: at the end of the part related to the ML application, you should clear the vectors, otherwise the ML output will be wrongly filled, e.g., also for candidates that do not pass the pre-selections.

Before exiting the `if`  you can do

```C++
    outputMlD0.clear();
    outputMlD0bar.clear();
```

atavirag:
> Hi @atavirag, thanks for the updates. I added a comment on the code. Moreover, I have another remark: at the end of the part related to the ML application, you should clear the vectors, otherwise the ML output will be wrongly filled, e.g., also for candidates that do not pass the pre-selections.
> 
> Before exiting the `if` you can do
> 
> ```c++
>     outputMlD0.clear();
>     outputMlD0bar.clear();
> ```

The vectors are currently being cleared at the beginning of the for loop, in lines 266 and 267. That is, before they get the chance to get filled again. Would it be okay to leave it like that?

fcatalan92 (approved):

vkucera:
This PR introduced a bug which prevents the workflow from running.
Fixed in https://github.com/AliceO2Group/O2Physics/pull/3872
@atavirag Please always test your changes locally before making a PR.

atavirag:
> This PR introduced a bug which prevents the workflow from running. Fixed in #3872 @atavirag Please always test your changes locally before making a PR. Also this new subscription should be mentioned in the PR description as it is not directly related to the ML response.

I am sorry, thank you for the fix. I will add the subscription to the description.

PR #3794, opened 2023-11-07
Next multiplicity

njacazio:


njacazio:
This splits the filling of the tables following the requests from other tasks

ddobrigk:
Thanks a lot @njacazio ! This looks good! :-D

...please don't kill me for this comment, but maybe what we could also do - maybe as a next development after this PR - is consider splitting the barrel counters in a modular way, too. This because these slicers actually are the CPU bottleneck of this task, and quite likely, not all 4 of them are always needed...

ddobrigk (approved):

njacazio:
Hi @ddobrigk sure I agree

PR #3796, opened 2023-11-07
removed code replication in MainLoop*

abilandz:


victor-gonzalez (approved):
Thank you!
I'm very pleased to observe that you now master the template approach!!

PR #3806, opened 2023-11-08
PWGLF: Resonance Tutorial Update

BongHwi:
Apply the common function shape for all tutorial tasks while keeping the same structure.


alibuild (dismissed):

alibuild (approved):

PR #3807, opened 2023-11-08
PWGLF: Tutorial update - Simplify the TOF selection part

BongHwi:
TOF and TPC selection is super easy to read


alibuild (approved):

PR #3814, opened 2023-11-09
Update HistogramsLibrary.cxx

ahkhanm:
add multiplicity for pp collision

### DIFF
--- a/PWGDQ/Core/HistogramsLibrary.cxx
+++ b/PWGDQ/Core/HistogramsLibrary.cxx
@@ -493,7 +493,7 @@ void o2::aod::dqhistograms::DefineHistograms(HistogramManager* hm, const char* h
         hm->AddHistogram(histClass, "Mass_Pt_CentFT0C", "", false, 500, 0.0, 5.0, VarManager::kMass, 400, 0.0, 40.0, VarManager::kPt, 20, 0.0, 100.0, VarManager::kCentFT0C);
       }
       if (subGroupStr.Contains("mult")) {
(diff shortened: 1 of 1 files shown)


ahkhanm:
@iarsene hi can you please have a look into my modification . I added histogram for multiplicity in pp collisions

PWGDQ/Core/HistogramsLibrary.cxx
```diff
@@ -492,6 +492,9 @@ void o2::aod::dqhistograms::DefineHistograms(HistogramManager* hm, const char* h
         hm->AddHistogram(histClass, "Pt_CentFT0C", "", false, 500, 0.0, 1.5, VarManager::kPt, 20, 0.0, 100.0, VarManager::kCentFT0C);
         hm->AddHistogram(histClass, "Mass_Pt_CentFT0C", "", false, 500, 0.0, 5.0, VarManager::kMass, 400, 0.0, 40.0, VarManager::kPt, 20, 0.0, 100.0, VarManager::kCentFT0C);
       }
+      if (subGroupStr.Contains("mult")) {
+        hm->AddHistogram(histClass, "Mass_Pt_MultFV0A", "", false, 500, 0.0, 5.0, VarManager::kMass, 400, 0.0, 40.0, VarManager::kPt, 100, 0.0, 25000.0, VarManager::kMultFV0A);
```
iarsene, line 496: I suggest to decrease the number of bins on mass and pt. This is making your histogram very large which would lead to a large amount of memory usage. The typical binning in mass that we use is with bins of 40 MeV in width. Also, for the pt, you use bins of 0.1 GeV/c, which is also too much. I suggest to decrease to at most 1 GeV bins.

ahkhanm: @iarsene Thanks for the suggestion. I just modified the bins according to your suggestion.

iarsene (approved):

PR #3819, opened 2023-11-09
PWGHF: add possibility to apply centrality selection for Pb-Pb skimming

fgrosa:


PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -195,54 +215,74 @@ struct HfTrackIndexSkimCreatorTagSelCollisions {
...
       }
     }
 
     // vertex selection
     selectVertex(collision, statusCollision);
 
-    // TODO: add more event selection criteria
-
     // selected events
     if (fillHistograms && statusCollision == 0) {
```
mfaggin, line 248:
```suggestion
    if (fillHistograms) {
```

Probably you need to remove ` && statusCollision == 0` inside the `if` statement, since your `if-else` statement below is based exactly on checking whether ` statusCollision == 0` (and at this stage, if ` && statusCollision != 0`, it is because of the centrality)

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -111,6 +117,11 @@ struct HfTrackIndexSkimCreatorTagSelCollisions {
 
   void init(InitContext const&)
   {
+    std::array<int, 3> doProcess = {doprocessTrigAndCentSel, doprocessTrigSel, doprocessNoTrigSel};
+    if (std::accumulate(doProcess.begin(), doProcess.end(), 0) != 1) {
+      LOGP(fatal, "One and only one process function for collision selection can be enabled at a time!");
+    }
+
```
mfaggin, line 124:
```suggestion

if(useCentralityFT0M && useCentralityFV0A) {
   LOGP(fatal, "More than 1 centrality estimator switched-on. Choose only one! (useCentralityFT0M, useCentralityFV0A)");
}

```

Your lines https://github.com/AliceO2Group/O2Physics/pull/3819/files#diff-0da5206b9f79483df1d09f610f2645680da7417bb8f63ad933cf26999036265cR229-R235 are always run with FT0M estimator if both the `bool`s are `true`, w/o any warning or similar. This protection, in my opinion, makes the running of the code less error-prone.

mfaggin (changes requested):
Ciao @fgrosa , thank you! I just have a couple of suggestions, then the rest looks good to me.

mfaggin (approved):

PR #3820, opened 2023-11-09
PWGHF: Add possibility to select MC event based on generator information

fcatalan92:
This PR add the possibility of using the subGeneratorId information recently added to the HF MC generator to select specific events from a gap-triggered production.

### DIFF
--- a/PWGHF/Tasks/taskMcValidation.cxx
+++ b/PWGHF/Tasks/taskMcValidation.cxx
@@ -73,8 +73,7 @@ struct HfTaskMcValidationGen {
   Configurable<double> yVertexMax{"yVertexMax", 100., "max. y of generated primary vertex [cm]"};
   Configurable<double> zVertexMin{"zVertexMin", -100., "min. z of generated primary vertex [cm]"};
   Configurable<double> zVertexMax{"zVertexMax", 100., "max. z of generated primary vertex [cm]"};
-  Configurable<bool> checkEventGeneratorInfo{"checkEventGeneratorInfo", false, "Enable selection of events using the subGeneratorId information"};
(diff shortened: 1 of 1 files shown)


mfaggin (dismissed):
thanks @fcatalan92 it looks ok to me. I approve it by myself, but let's see if @fgrosa @fcolamar have further comments

PWGHF/Tasks/taskMcValidation.cxx
```diff
@@ -73,6 +73,8 @@ struct HfTaskMcValidationGen {
   Configurable<double> yVertexMax{"yVertexMax", 100., "max. y of generated primary vertex [cm]"};
   Configurable<double> zVertexMin{"zVertexMin", -100., "min. z of generated primary vertex [cm]"};
   Configurable<double> zVertexMax{"zVertexMax", 100., "max. z of generated primary vertex [cm]"};
+  Configurable<bool> checkEventGeneratorInfo{"checkEventGeneratorInfo", false, "Enable selection of events using the subGeneratorId information"};
+  Configurable<int> eventGeneratorType{"eventGeneratorType", 4, "Which kind of events to keep (0 = MB, 4 = charm injected, 5 = beauty injected)"};
```
fgrosa, line 77: Very minor, I would not use `injected` as name, rather `triggered` since we do not inject signals but we trigger on them

vkucera, line 77: IMHO `checkEventGeneratorInfo` is not needed if `eventGeneratorType = -1` by default and `if (checkEventGeneratorInfo)` is replace with `if (checkEventGeneratorInfo >= 0)`.

fcatalan92:
Hi @fgrosa, @vkucera I implemented your suggestions.

alibuild (approved):

PR #3826, opened 2023-11-10
PWGHF: add pt-dependent Lc-window cut and fix ConfigAxes.

mfaggin:


PWGHF/Core/SelectorCuts.h
```diff
@@ -927,6 +927,55 @@ static const std::vector<std::string> labelsPt = {
...
+  5.,
+  6.,
+  8.,
+  12.,
+  24.,
+  36.};
+auto vecBinsPt = std::vector<double>{binsPt, binsPt + nBinsPt + 1};
+
+// default values for the cuts
+constexpr double cuts[nBinsPt][nCutVars] = {{0.03, 0.03},  /* 0  < pT < 1  */
```
fgrosa, line 951: I know that the cut values defined for all the other channels are `double`, however we don't need double precision for them. It is ok for the moment to use `double` for consistency, but I would keep it in mind and modify all of them to `float` in the future.

fgrosa (approved):

PR #3829, opened 2023-11-11
Move to expression columns for most used columns

ddobrigk:
This PR moves the most commonly used variables to expression columns in the strangeness model. As next few steps, the CosPAs will be changed into regular columns, so they can be filtered on, and indices and other variables will be further separated into sub-tables that are joinable.

Combined with the already modular strangeness PID tables and some further adjustments, one of the goals here is that, should the need arise, the intermediate format from the builders could already serve as derived data.


njacazio (approved):

njacazio:
Ah the ci is failing

njacazio (approved):

PR #3832, opened 2023-11-12
Update heavy-ion mult task (Add histograms for different centrality intervals and for correlations between FT0/FV0 vs global tracks)

abmodak:
Add histograms for different centrality intervals and for correlations between FT0/FV0 vs global tracks


PWGMM/Mult/Tasks/heavy-ion-mult.cxx
```diff
@@ -68,13 +71,18 @@ static constexpr TrackSelectionFlags::flagtype trackSelectionDCA =
 static constexpr TrackSelectionFlags::flagtype trackSelectionDCAXYonly =
   TrackSelectionFlags::kDCAxy;
 
+static constexpr int CentHistArray{15};
+std::shared_ptr<TH1> h1MultHistCentFT0C[CentHistArray];
```
aalkin, line 75: It is recommended to use `std::array` rather than C-arrays. In this case `std::array<std::shared_ptr<TH1>, CentHistArray> h1MultHistCentFT0C;`

PWGMM/Mult/Tasks/heavy-ion-mult.cxx
```diff
@@ -244,6 +280,64 @@ struct HeavyIonMultiplicity {
     }
   }
   PROCESS_SWITCH(HeavyIonMultiplicity, processMC, "process MC", false);
+
+  void processDataCentFT0C(CollisionDataTableCentFT0C::iterator const& collision, FilTrackDataTable const& tracks)
+  {
+    bool hasCentrality = false;
```
aalkin, line 286: Rather then making it a runtime variable, you can make it `constexpr`, then you can use it in `if constexpr ()` conditions. So `constexpr auto hasCentrality = CollisionDataTableCentFT0C::template contains<aod::CentFT0Cs>();`

aalkin (approved):

PR #3835, opened 2023-11-13
PWGCF: Update trigger

ariedel-cern:
- Add cut and histograms to study the correlation between the momentum of a track at the TPC inner wall and the reconstructed momentum.
- Add histograms to plot the particle pT vs Q3/kstar of the pair


lauraser (approved):
Please consider adding the percentage plot (difference/momentum)

PR #3853, opened 2023-11-14
PWGEM/PHOTONMESON: Add RPhi dependence plots

f3sch:
Add RPhi histogram to monitor bias of found&reco svertex against phi.


mhemmer-cern (approved):
looks good

PR #3861, opened 2023-11-14
PWGHF: Charm hadron flow task

stefanopolitano:
This PR contains a first draft of the task to compute the scalar product of charm hadrons:
- D+ and Ds meson implemented;
- Dedicated process function to compute scalar-product and event-plane resolution;


fgrosa:
> This PR contains a first draft of the task to compute the scalar product of charm hadrons:
> 
>     * D+ and Ds meson implemented;
> 
>     * TODO: update task with Qvector tables for each detector and event-plane resolution when implemented ;

Ciao @stefanopolitano, thanks for this initial draft! Please just move it in the `D2H/Task` folder, where it better belongs.

stefanopolitano:
Hi @fgrosa, thanks for the comment! I moved the task to the suggested folder

vkucera:
@fgrosa Why is this not supposed to go into HFC, where the flow analyses have been developed so far (see [`taskFlow.cxx`](https://github.com/AliceO2Group/O2Physics/blob/master/PWGHF/HFC/Tasks/taskFlow.cxx))?

fgrosa:
> @fgrosa Why is this not supposed to go into HFC, where the flow analyses have been developed so far (see [`taskFlow.cxx`](https://github.com/AliceO2Group/O2Physics/blob/master/PWGHF/HFC/Tasks/taskFlow.cxx))?

Hi @vkucera, the measurement of flow harmonics in Pb-Pb done with the EP or SP methods are done in the D2H, that's why I suggested to move it there. The one in HFC used particle correlations, so it is indeed in the right place. We have to find better names for the tasks to avoid confusion.

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,339 @@
...
+                    DsToPiKK };
+
+struct taskFlowCharmHadrons {
+  Configurable<int> harmonic{"harmonic", 2, "harmonic number"};
+  Configurable<int> qvecDetector{"qvecDetector", 0, "Detector for Q vector estimation (FT0C: 0, FT0A: 1, FT0M: 2, FV0A: 3, TPC Pos: 4, TPC Neg: 5)"};
+  Configurable<int> centDetector{"centDetector", 0, "Detector for centrality estimation (V0A: 0, T0M: 1, T0A: 2, T0C: 3"};
+  Configurable<int> selectionFlag{"selectionFlag", 1, "Selection Flag for D"};
+  Configurable<bool> storeMl{"storeMl", false, "Flag to store ML score"};
+  Configurable<bool> saveEpResoHisto{"saveEpResoHisto", false, "Flag to save event plane resolution histogram"};
+  Configurable<int> classMl{"classMl", 0, "Index of the ML class to be stored"};
```
mfaggin, line 44: Probably minor, but can you mention e.g. in the description which values this variable can have and to what it corresponds to?

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,339 @@
...
+    float pYtrack2 = cand.pyProng2();
+    float pTtrack2 = cand.ptProng2();
+    float phiTrack2 = TMath::ATan2(pYtrack2, pXtrack2);
+
+    tracksQx.push_back(TMath::Cos(harmonic * phiTrack0) * pTtrack0 / ampl);
+    tracksQy.push_back(TMath::Sin(harmonic * phiTrack0) * pTtrack0 / ampl);
+    tracksQx.push_back(TMath::Cos(harmonic * phiTrack1) * pTtrack1 / ampl);
+    tracksQy.push_back(TMath::Sin(harmonic * phiTrack1) * pTtrack1 / ampl);
+    tracksQx.push_back(TMath::Cos(harmonic * phiTrack2) * pTtrack2 / ampl);
+    tracksQy.push_back(TMath::Sin(harmonic * phiTrack2) * pTtrack2 / ampl);
```
vkucera, line 126: Use standard C++ library whenever possible.

mfaggin (commented):
Thank you very much @stefanopolitano for this important development. I just have a few comments/requests.

Best. Mattia

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,339 @@
...
+  {
+    float deltaPsi = psi1 - psi2;
+    if (deltaPsi > TMath::Pi() / harmonic) {
+      if (deltaPsi > 0.)
+        deltaPsi -= 2. * TMath::Pi() / harmonic;
+      else
+        deltaPsi += 2. * TMath::Pi() / harmonic;
+    }
+    return deltaPsi;
+  }
```
vkucera, line 143: Use `TwoPI` from O2.
This does not look like it can guarantee the returned values to be in the desired range, since it only does one pass of adjustment.
Can this not be replaced with `RecoDecay::constrainAngle` and then divided by `harmonic`?

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,339 @@
...
+  void fillThn(float mass,
+               float pt,
+               float cent,
+               float cosNPhi,
+               float cosDeltaPhi,
+               float sp,
+               float outputMl)
+  {
+    registry.fill(HIST("hSparseFlowCharm"), mass, pt, cent, cosNPhi, cosDeltaPhi, sp, outputMl);
+  }
```
vkucera, line 163: This function is used only once and it is a single line. Is it needed?

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,339 @@
...
+      registry.add("epReso/hEpResoFV0mTPCneg", "hEpResoFV0mTPCneg; centrality; #Delta#Psi_{sub}", {HistType::kTH2F, {thnAxisCent, thnAxisCosNPhi}});
+    }
+  }; // end init
+
+  /// Compute the Q vector for the candidate's tracks
+  /// \param cand is the candidate
+  /// \param tracksQx is the X component of the Q vector for the tracks
+  /// \param tracksQy is the Y component of the Q vector for the tracks
+  template <typename T1>
+  void GetQvecDtracks(const T1& cand,
```
vkucera, line 102:
```suggestion
  void getQVecDTracks(const T1& cand,
```
Respect naming conventions.

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,339 @@
...
+  /// Get the centrality
+  /// \param collision is the collision with the centrality information
+  /// \param centDetector is the detector used for the centrality estimation
+  float GetCent(CollsWithQvecs::iterator const& collision, int centDetector)
+  {
+    float cent = -999.;
+    switch (centDetector) {
+      case 0: // V0M
+        cent = collision.centFV0A();
+        break;
```
vkucera, line 174: It would be clearer to define `enum` for these values. If there are changes, it will be easier to update the `centDetector` documentation by checking the `enum` definition instead of looking inside the code which uses it.

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,339 @@
...
+      case 2: // T0A
+        cent = collision.centFT0A();
+        break;
+      case 3: // T0C
+        cent = collision.centFT0C();
+        break;
+      default:
+        break;
+    }
```
vkucera, line 186: You should add protection against unrecognised values here.

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,339 @@
...
+  /// \param collision is the collision with the Q vector information and event plane
+  /// \param candidates are the selected candidates
+  template <int decayChannel, typename T1>
+  void runFlowAnalysis(CollsWithQvecs::iterator const& collision,
+                       T1 const& candidates)
+  {
+    float QvecX = -999.;
+    float QvecY = -999.;
+    float amplQvec = -999.;
+    float evtPl = -999.;
```
vkucera, line 201:
```suggestion
    float xQVec = -999.;
    float yQVec = -999.;
    float amplQVec = -999.;
    float evtPl = -999.;
```

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,339 @@
...
+    switch (qvecDetector.value) {
+      case 0: // FT0c
+        QvecX = collision.qvecFT0CRe();
```
vkucera, line 204: Again, better to define `enum`

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,339 @@
...
+        amplQvec = collision.nTrkBPos();
+        break;
+      case 5: // TPC Negative
+        QvecX = collision.qvecBNegRe();
+        QvecY = collision.qvecBNegIm();
+        amplQvec = collision.nTrkBNeg();
+        break;
+      default:
+        break;
+    }
```
vkucera, line 231: Add protection

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,339 @@
...
+    evtPl = epHelper.GetEventPlane(QvecX, QvecY);
+
+    float cent = GetCent(collision, centDetector.value);
+
+    for (auto const& cand : candidates) {
```
vkucera, line 236:
```suggestion
    for (const auto& candidate : candidates) {
```

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,339 @@
...
+
+#include "Common/DataModel/Qvectors.h"
+#include "Common/Core/EventPlaneHelper.h"
+
+using namespace o2;
+using namespace o2::aod;
+using namespace o2::framework;
+using namespace o2::framework::expressions;
+
+enum decayChannel { DplusToPiKPi = 0,
```
vkucera, line 33:
```suggestion
enum DecayChannel { DplusToPiKPi = 0,
```

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,339 @@
...
+#include "Framework/AnalysisTask.h"
+#include "Framework/HistogramRegistry.h"
+#include "Framework/runDataProcessing.h"
+
+#include "PWGHF/Core/HfHelper.h"
+#include "PWGHF/DataModel/CandidateReconstructionTables.h"
+#include "PWGHF/DataModel/CandidateSelectionTables.h"
+
+#include "Common/DataModel/Qvectors.h"
+#include "Common/Core/EventPlaneHelper.h"
```
vkucera, line 26: Fix order

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,339 @@
...
+using namespace o2;
+using namespace o2::aod;
+using namespace o2::framework;
+using namespace o2::framework::expressions;
+
+enum decayChannel { DplusToPiKPi = 0,
+                    DsToKKPi,
+                    DsToPiKK };
+
+struct taskFlowCharmHadrons {
```
vkucera, line 37:
```suggestion
struct HfTaskFlowCharmHadrons {
```

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,339 @@
...
+               float sp,
+               float outputMl)
+  {
+    registry.fill(HIST("hSparseFlowCharm"), mass, pt, cent, cosNPhi, cosDeltaPhi, sp, outputMl);
+  }
+
+  /// Get the centrality
+  /// \param collision is the collision with the centrality information
+  /// \param centDetector is the detector used for the centrality estimation
+  float GetCent(CollsWithQvecs::iterator const& collision, int centDetector)
```
vkucera, line 168:
```suggestion
  float getCentrality(CollsWithQvecs::iterator const& collision, int centDetector)
```

vkucera: `centDetector` is a member of the struct so it does not need to be passed as argument.

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,339 @@
...
+  }
+  PROCESS_SWITCH(taskFlowCharmHadrons, processDplus, "Process Dplus candidates", true);
+
+  // Event plane
+  void processEventPlaneReso(CollsWithQvecs::iterator const& collision)
+  {
+    if (!saveEpResoHisto) {
+      LOG(fatal) << "Event plane resolution histogram not saved. Please set saveEpResoHisto to true.";
+      return;
+    }
```
vkucera, line 307: This can be caught already in `init`.

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,339 @@
...
+
+    float evCent = GetCent(collision, centDetector.value);
```
vkucera, line 309:
```suggestion
    float centrality = GetCent(collision, centDetector.value);
```

vkucera (dismissed):

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
stefanopolitano: Hi @vkucera ! Thanks a lot for the superquick review and the suggestions. I will implement them. I just have a few doubts. Regarding this comment, if I understood correctly, your idea would be to use the `RecoDecay::constrainAngle`  to subtract/add 2Pi from the DeltaPsi and scale the result by the harmonic. But this is not the same as subtracting/adding a factor of 2Pi/harmonic to the DeltaPsi. Am I right or perhaps I am missing something?

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
stefanopolitano: I implemented it in this way with the idea that we may want to add other histograms to this registry for QA purposes at some point. Anyhow, with the current state of the code you are right I could simply fill the Thn without this function

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
stefanopolitano: Ok, I will do something like it is done at skimming level then [https://github.com/AliceO2Group/O2Physics/blob/2fdafcea42ac73d894464919dc4b4643fa3e0070/PWGHF/TableProducer/trackIndexSkimCreator.cxx#L83]

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
vkucera: I might be misunderstanding of what the function should do but I am suspecting that there are several problems in the current version.

- I assume that `harmonic` is supposed to be a positive integer.
- In that case, if `deltaPsi > TMath::Pi() / harmonic` then also `deltaPsi > 0.` so the `else` clause is never executed and there is no range constrain for `deltaPsi < 0.`.
- I assume that you want to handle all possible values of `deltaPsi`. For the trivial case of `harmonic = 1`, I assume that you want to constrain `deltaPsi` within [0, 2 pi] or [-pi, pi]. Therefore I would expected that for higher harmonics, the range would scale as [0, 2 pi/harmonic] or [-pi/harmonic, pi/harmonic]. So trying to constrain `deltaPsi` to [0, pi/harmonic] does not seem correct to me.

vkucera: You are right that scaling the constrained number would not be correct. The `RecoDecay::constrainAngle` could actually be extended to scale the increments by the harmonic number.

stefanopolitano: I see your point, thanks for the clarification. Indeed, what is missing in line 136 is an absolute value I would say.
( `if (TMath::Abs(deltaPsi) > TMath::Pi() / harmonic) {`). In this way, both conditions are accessed depending on the value of `deltaPsi`.
Regarding the range, I checked the code from [Run2](https://github.com/alisw/AliPhysics/blob/1e829395dcfaf57dde72ae586f068fa631ea1c37/PWGHF/vertexingHF/vHFML/AliAnalysisTaskSECharmHadronvn.cxx#L1088) and indeed the spanned interval is [0, pi/harmonic]. In my understanding, this choice is consistent since the deltaPsi is multiplied by the harmonic (e.g. line 317) before filling the TH1, thus spanning a range of [0, pi].

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
mfaggin: no clear suggestions from my side. Indeed I was wondering how much this might depend on the `MLResponse` organization, but from your reply I understand it's not easy to comment it here properly. Let's keep it in mind for the future.

stefanopolitano: Ok, thanks! Maybe I could add the multiclass example for the D+ in the configurable definition, which I assume will be the most common case. But indeed, I don't think it will cover all the possible scenarios

mfaggin: yes this can be already helpful, thanks!

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
vkucera: OK, that helps a bit, but still, I would say that for `delta = pi/harmonic + epsilon` the Run 2 function returns `-pi/harmonic + epsilon < 0` which is outside the range you claim to want.

vkucera: So I still suspect that you actually want to constrain the angle to the range [-pi/harmonic, pi/harmonic]. Otherwise, I don't see how you can properly handle negative values.

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,339 @@
...
+          outputMl = cand.mlProbDplusToPiKPi()[classMl.value];
+      }
+      float ptCand = cand.pt();
+      float phiCand = cand.phi();
+
+      // TPC only
+      if (qvecDetector.value == 4 || qvecDetector.value == 5) {
+        float ampl = amplQvec - 3.;
+        std::vector<float> tracksQx = {-999., -999., -999.};
+        std::vector<float> tracksQy = {-999., -999., -999.};
```
fcatalan92, line 267:
```suggestion
        std::vector<float> tracksQx = {};
        std::vector<float> tracksQy = {};
```
Since in `GetQvecDtracks` you push back the values you do not overwrite your initialisation, thus in the following for loop you are always getting `-999.`. 
It would be better to use `std:arrays` since you know the size you need.

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,339 @@
...
+        }
+      } else if constexpr (std::is_same<T1, CandDplusData>::value) {
+        massCand = hfHelper.invMassDplusToPiKPi(cand);
+        if (storeMl)
+          outputMl = cand.mlProbDplusToPiKPi()[classMl.value];
+      }
+      float ptCand = cand.pt();
+      float phiCand = cand.phi();
+
+      // TPC only
```
fcatalan92, line 263: Please add a brief comment on why you are doing the following subtraction

fcatalan92 (changes requested):

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
fgrosa: Indeed it is correct to divide by `n` and not by `2` (fortunately for the `v3` in Run 2 we used the SP method, that's also why we did not spot the typo probably).
Regarding the interval of `psi` and `deltaPsi` @fcatalan92 is right, so the implementation of @stefanopolitano is correct.

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,339 @@
...
+      return;
+    }
+
+    float evCent = GetCent(collision, centDetector.value);
+    float epFT0a = epHelper.GetEventPlane(collision.qvecFT0ARe(), collision.qvecFT0AIm());
+    float epFT0c = epHelper.GetEventPlane(collision.qvecFT0CRe(), collision.qvecFT0CIm());
+    float epFT0m = epHelper.GetEventPlane(collision.qvecFT0MRe(), collision.qvecFT0MIm());
+    float epFV0a = epHelper.GetEventPlane(collision.qvecFV0ARe(), collision.qvecFV0AIm());
+    float epBPoss = epHelper.GetEventPlane(collision.qvecBPosRe(), collision.qvecBPosIm());
+    float epBNegs = epHelper.GetEventPlane(collision.qvecBNegRe(), collision.qvecBNegIm());
```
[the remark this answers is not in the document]
fgrosa: Indeed here you should add the third argument to specify the harmonic

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
vkucera: I am still not convinced.
Assuming that |psi| <= pi/2 and the target range is |deltaPsi| <= pi/n (as @fcatalan92 says):
if |psi| <= pi/2, then |psi1 - psi2| <= pi
if (deltaPsi = pi > pi/n), then the function would do: deltaPsi = deltaPsi - 2pi/n = pi(1-2/n)
This deltaPsi will be in the target range if: 1-2/n <= 1/n, which is true only if n <= 3.
So in my opinion, the function is still incorrect, unless the `if` conditions are replaced with `while` loops.

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
stefanopolitano: Hi @fcatalan92 , indeed we should find a way to organise the signal (prompt+ nonprompt) / background models in the vector storing the BDT information. For the moment, I will add a second BDT-related axis to store the bkg score, as you suggested. Thanks!

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
vkucera: Ah, but I guess your point was that the problem is the missing `harmonic` in the calls of `GetEventPlane`, right?

vkucera: Then, indeed `|psi1 - psi2| <= 2pi/n` which works fine.

stefanopolitano:
Dear @fcatalan92, @mfaggin , and @vkucera,
Thanks a lot for the comments and the discussion. I implemented the suggestion as agreed. Please, let me know if there is still something to be changed ; )

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,367 @@
...
+                     FT0C,
+                     TPCPos,
+                     TPCNeg };
+
+struct HfTaskFlowCharmHadrons {
+  Configurable<int> harmonic{"harmonic", 2, "harmonic number"};
+  Configurable<int> qvecDetector{"qvecDetector", 0, "Detector for Q vector estimation (FV0A: 0, FT0M: 1, FT0A: 2, FT0C: 3, TPC Pos: 4, TPC Neg: 5)"};
+  Configurable<int> centDetector{"centDetector", 0, "Detector for centrality estimation (V0A: 0, T0M: 1, T0A: 2, T0C: 3"};
+  Configurable<int> selectionFlag{"selectionFlag", 1, "Selection Flag for hadron (e.g. 1 for skimming, 3 for topo. and kine., 7 for PID)"};
+  Configurable<bool> storeMl{"storeMl", false, "Flag to store ML scores (e.g. 0 for Bkg, 1 for Prompt, 2 for Feed-down)"};
```
fcatalan92, line 54: The comment of the configurable is wrong since it is a `bool`

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,367 @@
...
+  Configurable<bool> saveEpResoHisto{"saveEpResoHisto", false, "Flag to save event plane resolution histogram"};
+  Configurable<std::vector<int>> classMl{"classMl", {0, 2}, "Indexes of BDT scores to be stored (e.g. 0 for Bkg, 2 for Feed-down). Two indexes max."};
```
fcatalan92, line 56: The example you make is not correct in general, the order depends on how the model is trained.

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,367 @@
...
+  }
+
+  /// Compute the delta psi in the range [0, pi/harmonic]
+  /// \param psi1 is the first angle
+  /// \param psi2 is the second angle
+  /// \note Ported from AliAnalysisTaskSECharmHadronvn::GetDeltaPsiSubInRange
+  float getDeltaPsiInRange(float psi1, float psi2)
+  {
+    float deltaPsi = psi1 - psi2;
+    if (TMath::Abs(deltaPsi) > TMath::Pi() / harmonic) {
```
fcatalan92, line 150:
```suggestion
    if (TMath::Abs(deltaPsi) > constants::math::PI / harmonic) {
```

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,367 @@
...
+        massCand = hfHelper.invMassDplusToPiKPi(candidate);
+        if (storeMl) {
+          for (unsigned int iclass = 0; iclass < classMl->size(); iclass++)
+            outputMl[iclass] = candidate.mlProbDplusToPiKPi()[classMl->at(iclass)];
+        }
+      }
+      float ptCand = candidate.pt();
+      float phiCand = candidate.phi();
+
+      // TPC only
```
fcatalan92, line 294:
```suggestion
```

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,367 @@
...
+      // If TPC is used for the SP estimation, the tracks of the hadron candidate must be removed from the TPC Q vector
+      // to avoid double counting.
```
fcatalan92, line 296:
```suggestion
      // If TPC is used for the SP estimation, the tracks of the hadron candidate must be removed from the TPC Q vector to avoid double counting.
```

fcatalan92 (changes requested):
Hi @stefanopolitano, thanks for having updated the PR. I have made just a few additional minor comments. I think you should also add the histograms for the scalar-product resolution before merging.

stefanopolitano:
Hi @fcatalan92 ! Thanks for the review, I implemented your comments and added the computation of the resolution for the scalar product method. Let me know if there is still something to be updated!

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
```diff
@@ -331,31 +343,61 @@ struct HfTaskFlowCharmHadrons {
   }
   PROCESS_SWITCH(HfTaskFlowCharmHadrons, processDplus, "Process Dplus candidates", true);
 
-  // Event plane
+  // Resolution
   void processEventPlaneReso(CollsWithQvecs::iterator const& collision)
```
fcatalan92, line 347: Change the name here, e.g. `processResolution`

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
```diff
@@ -331,31 +343,61 @@ struct HfTaskFlowCharmHadrons {
...
+      registry.fill(HIST("epReso/hEpResoFT0aTPCpos"), centrality, TMath::Cos(harmonic * getDeltaPsiInRange(epFT0a, epBPoss)));
+      registry.fill(HIST("epReso/hEpResoFT0aTPCneg"), centrality, TMath::Cos(harmonic * getDeltaPsiInRange(epFT0a, epBNegs)));
+      registry.fill(HIST("epReso/hEpResoFT0mFV0m"), centrality, TMath::Cos(harmonic * getDeltaPsiInRange(epFT0m, epFV0a)));
+      registry.fill(HIST("epReso/hEpResoFT0mTPCpos"), centrality, TMath::Cos(harmonic * getDeltaPsiInRange(epFT0m, epBPoss)));
+      registry.fill(HIST("epReso/hEpResoFT0mTPCneg"), centrality, TMath::Cos(harmonic * getDeltaPsiInRange(epFT0m, epBNegs)));
+      registry.fill(HIST("epReso/hEpResoFV0mTPCpos"), centrality, TMath::Cos(harmonic * getDeltaPsiInRange(epFV0a, epBPoss)));
+      registry.fill(HIST("epReso/hEpResoFV0mTPCneg"), centrality, TMath::Cos(harmonic * getDeltaPsiInRange(epFV0a, epBNegs)));
+    }
   }
   PROCESS_SWITCH(HfTaskFlowCharmHadrons, processEventPlaneReso, "Process event plane resolution", false);
```
fcatalan92, line 402: Update also the naming here to be consistent with the new changes

fcatalan92 (approved):

fcatalan92 (dismissed):

fcatalan92 (approved):

PR #3872, opened 2023-11-15
PWGHF: Fix missing Spawn in D0 selector

vkucera:
Fix for bug introduced in https://github.com/AliceO2Group/O2Physics/pull/3789

The tables `TracksPidPiExt` and `TracksPidKaExt` are consumed but not produced.


fgrosa (approved):
Thanks @vkucera!

vkucera:
This fix has a caveat that the D0 and D+ selector cannot run together because they produce the same tables. So we should have a dedicated workflow for the production of the PID tables.

alibuild (approved):

PR #3873, opened 2023-11-15
PWGJE: removing protection for orphan tracks from derived data

nzardosh:


alibuild (dismissed):

alibuild (approved):

raymondEhlers:
As a quick space optimization, `sign` could be a bool (if interpreted as -1 == false and +1 == true), or at least a int8_t

PR #3880, opened 2023-11-16
PWGHF: Add more centrality estimators

fgrosa:
In this PR I added more centrality estimators and I split the process functions for each centrality estimator to avoid the process of centrality tables which are not used.

### DIFF
--- a/PWGHF/TableProducer/trackIndexSkimCreator.cxx
+++ b/PWGHF/TableProducer/trackIndexSkimCreator.cxx
@@ -212,7 +212,7 @@ struct HfTrackIndexSkimCreatorTagSelCollisions {
 
   /// Collision selection
   /// \param collision  collision table with
-  template <bool applyTrigSel, bool applyCentSel, int centEstimator = CentralityEstimators::None, typename Col>
+  template <bool applyTrigSel = false, int centEstimator = CentralityEstimators::None, typename Col>
   void selectCollision(const Col& collision)
   {
     int statusCollision = 0;
@@ -233,7 +233,7 @@ struct HfTrackIndexSkimCreatorTagSelCollisions {
     }
 
     float centrality = -1.;
-    if constexpr (applyCentSel) {
+    if constexpr (centEstimator != CentralityEstimators::None) {
       if constexpr (centEstimator == CentralityEstimators::FT0A) {
         centrality = collision.centFT0A();
       } else if constexpr (centEstimator == CentralityEstimators::FT0C) {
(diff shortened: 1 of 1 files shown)


mfaggin (approved):
ciao @fgrosa , this looks well in line with what we were discussing in the lats days. Good for me!

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -208,7 +212,7 @@ struct HfTrackIndexSkimCreatorTagSelCollisions {
 
   /// Collision selection
   /// \param collision  collision table with
-  template <bool applyTrigSel, bool applyCentSel, typename Col>
+  template <bool applyTrigSel, bool applyCentSel, int centEstimator = CentralityEstimators::None, typename Col>
```
vkucera, line 215: In general, it increases readability when all non-type template arguments are provided explicitly.
Moreover here, the `None` case does not seem needed.

fgrosa: Good point, it is indeed non necessary / redundant. Indeed I thought that it was better to remove the `applyCentSel` argument and rely on the `centEstimartor` only (in case of `None` now no centrality selection is applied).

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -208,7 +212,7 @@ struct HfTrackIndexSkimCreatorTagSelCollisions {
 
   /// Collision selection
   /// \param collision  collision table with
-  template <bool applyTrigSel, bool applyCentSel, typename Col>
+  template <bool applyTrigSel = false, int centEstimator = CentralityEstimators::None, typename Col>
```
vkucera, line 215:
```suggestion
  template <bool applyTrigSel, int centEstimator, typename Col>
```
Following up on my previous comment, I would suggest to always require to explicitly provide all arguments for each instance.

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -76,6 +76,16 @@ enum EventRejection {
   NEventRejection
 };
 
+// event rejection types
+enum CentralityEstimators {
```
vkucera, line 80:
```suggestion
enum CentralityEstimator {
```

PWGHF/TableProducer/trackIndexSkimCreator.cxx
fgrosa: Sorry, I understood the opposite :)

vkucera (approved):
Thanks @fgrosa

PR #3883, opened 2023-11-16
Rework of the multiplicity and PVcontribs axis ranges settings

rolavick:


mfaggin:
thank you @rolavick

mfaggin (approved):

mfaggin:
@Mingyu3360715 when running again trains on Hyperloop for the trackQC, beware that the settings are correct. In particular:
 - check that the axis settings fro PV contributors and multiplicity are fine when you run in pp (everything should be fne if you pick-up the default values)
 - if you run on PbPb be sure to enable the option `overwriteAxisRangeForPbPb`

PR #3834, opened 2023-11-12
[PWGLF] Change V0 Cos PA to regular column for filterability

ddobrigk:
This (probably long overdue) change in the strangeness data model moves the V0 Cos PA to a regular column so that declarative filtering (much faster) will work on this variable. Together with the previous changes to the strangeness data model (#3829), even parametric CosPA cuts will be applicable as p and pT are now expression columns instead of dynamic columns. Note that the change is, among other reasons, also justified because the collision received as argument is never changed with respect to the collision that the V0 is associated to, as far as I know. If there is any use case for that, please tell me. 

Note that a lot of files are touched by this commit. I tried to be as careful as possible, but if anyone spots anything that looks suspicious, please let me know and I promise to react as quickly as I manage. Thank you!!

PR #3886, opened 2023-11-16
PWGHF: fix and update cascade-bachelor index skimming

ZFederica:
- update and improve cascade-bachelor index skimming in trackIndexSkimCreator
- implement distinction between xi and omega
- add collisionId to tables


ZFederica:
Hi @fgrosa and @fcolamar what do you think of this implementation?
It should be already compatible with the changes in the LF PR https://github.com/AliceO2Group/O2Physics/pull/3834

vkucera:
Hi @ZFederica , thanks for the development. There are many invalid C++ expressions in your code. Please make sure your code compiles locally before pushing further commits.

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -3222,18 +3227,85 @@ struct HfTrackIndexSkimCreatorLfCascades {
...
           // check not to take the same particle twice in the decay chain
-          if (trackPion1.globalIndex() == trackXiDauCharged.globalIndex() || trackPion1.globalIndex() == trackV0DauPos.globalIndex() || trackPion1.globalIndex() == trackV0DauNeg.globalIndex() || trackPion1.globalIndex() == casc.globalIndex()) {
+          if (trackPion1.globalIndex() == trackXiDauCharged.globalIndex() || trackPion1.globalIndex() == trackV0PosDau.globalIndex() || trackPion1.globalIndex() == trackV0NegDau.globalIndex()) {
             continue;
           }
 
           // primary pion track to be processed with DCAFitter
           auto trackParVarPion1 = getTrackParCov(trackPion1);
 
+          // find charm baryon decay using xi PID hypothesis
```
fgrosa, line 3237: Shouldn't we first check if the cascade mass is compatible with the Xi mass?

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -3222,18 +3227,85 @@ struct HfTrackIndexSkimCreatorLfCascades {
...
+            auto secondaryVertex2XiHyp = df2.getPCACandidate();
+
+            cascType = 0;
+            SETBIT(cascType, kXi);
+
+            // fill table row
+            rowTrackIndexCasc2Prong(thisCollId,
+                                    casc.globalIndex(),
+                                    trackPion1.globalIndex(),
+                                    cascType);
```
fgrosa, line 3257: I would use `HFflag` here. Also, you should not fill it until it is not checked also the other mass hypothesis. Moreover, it must be filled only once at the end and `HFflag` will contain the information whether it is compatible with Xi, Omega, or both.

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -3222,18 +3227,85 @@ struct HfTrackIndexSkimCreatorLfCascades {
...
+            auto secondaryVertex2OmegaHyp = df2.getPCACandidate();
+
+            cascType = 0;
+            SETBIT(cascType, kOmega);
+
+            // fill table row
+            rowTrackIndexCasc2Prong(thisCollId,
+                                    casc.globalIndex(),
+                                    trackPion1.globalIndex(),
+                                    cascType);
```
fgrosa, line 3290: Fill only once, see above.

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -3222,18 +3227,85 @@ struct HfTrackIndexSkimCreatorLfCascades {
...
+
+            std::array<float, 3> pVecOmega = {0.};
+            std::array<float, 3> pVecPion1OmegaHyp = {0.};
+            df2.getTrack(0).getPxPyPzGlo(pVecOmega);
+            df2.getTrack(1).getPxPyPzGlo(pVecPion1OmegaHyp);
+
+            auto secondaryVertex2OmegaHyp = df2.getPCACandidate();
+
+            cascType = 0;
+            SETBIT(cascType, kOmega);
```
fgrosa, line 3284: There is no point of using a bit, if it is always reset before filling. This should contain also the information of the previous hypothesis tested.

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -3222,18 +3227,85 @@ struct HfTrackIndexSkimCreatorLfCascades {
...
+
+            df2.propagateTracksToVertex();
+
+            std::array<float, 3> pVecXi = {0.};
+            std::array<float, 3> pVecPion1XiHyp = {0.};
+            df2.getTrack(0).getPxPyPzGlo(pVecXi);
+            df2.getTrack(1).getPxPyPzGlo(pVecPion1iHyp);
+
+            auto secondaryVertex2XiHyp = df2.getPCACandidate();
+
```
fgrosa, line 3249: Here we are not using the information of the propagation. We should check if the mass is compatible with the interval set for Xipi.

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -3222,18 +3227,85 @@ struct HfTrackIndexSkimCreatorLfCascades {
...
+              auto mass2ProngXiHyp = RecoDecay::m(arrMomToXi, arrMass2Prong[hf_cand_casc_lf_2prong::DecayType::XiczeroOmegaczeroToXiPi]);
+              registry.fill(HIST("hMassXicZeroOmegacZeroToXiPi"), mass2ProngXiHyp);
+            }
+          }
+
+          // find charm baryon decay using omega PID hypothesis
+          int nVtxFrom2ProngFitterOmegaHyp = df2.process(trackCascOmega, trackParVarPion1);
+          if (nVtxFrom2ProngFitterOmegaHyp > 0) {
+
+            df2.propagateTracksToVertex();
```
fgrosa, line 3274: Again here we are not using the information of the propagation. We should check if the mass is compatible with the interval set for Omegapi.

fgrosa (changes requested):
Hi @ZFederica thanks for taking care of updating the finding of Xic and Omegac candidates! However, I have few comments on the logic that should be fixed, see below.

ZFederica:
Hello @fgrosa, thanks for your suggestions, now they should be all implemented.

PWGHF/DataModel/CandidateReconstructionTables.h
```diff
@@ -824,70 +828,17 @@ DECLARE_SOA_TABLE(HfCand3ProngMcGen, "AOD", "HFCAND3PMCGEN", //!
...
-DECLARE_SOA_DYNAMIC_COLUMN(ImpactParameterProngSqSum, impactParameterProngSqSum, //!
-                           [](float impParProng0, float impParProng1) -> float { return RecoDecay::sumOfSquares(impParProng0, impParProng1); });
-DECLARE_SOA_DYNAMIC_COLUMN(MaxNormalisedDeltaIP, maxNormalisedDeltaIP, //!
-                           [](float xVtxP, float yVtxP, float xVtxS, float yVtxS, float errDlxy, float pxM, float pyM, float ip0, float errIp0, float ip1, float errIp1, float px0, float py0, float px1, float py1) -> float { return RecoDecay::maxNormalisedDeltaIP(std::array{xVtxP, yVtxP}, std::array{xVtxS, yVtxS}, errDlxy, std::array{pxM, pyM}, std::array{ip0, ip1}, std::array{errIp0, errIp1}, std::array{std::array{px0, py0}, std::array{px1, py1}}); });
-// MC matching result:
-DECLARE_SOA_COLUMN(FlagMcMatchRec, flagMcMatchRec, int8_t); //! reconstruction level
-DECLARE_SOA_COLUMN(FlagMcMatchGen, flagMcMatchGen, int8_t); //! generator level
-DECLARE_SOA_COLUMN(OriginMcRec, originMcRec, int8_t);       //! particle origin, reconstruction level
-DECLARE_SOA_COLUMN(OriginMcGen, originMcGen, int8_t);       //! particle origin, generator level
```
fgrosa, line 884: Why did you delete those columns? (maybe I missed something, if this is the case sorry!)

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -3253,17 +3272,99 @@ struct HfTrackIndexSkimCreatorLfCascades {
...
+
+            std::array<float, 3> pVecXi = {0.};
+            std::array<float, 3> pVecPion1XiHyp = {0.};
+            df2.getTrack(0).getPxPyPzGlo(pVecXi);
+            df2.getTrack(1).getPxPyPzGlo(pVecPion1XiHyp);
+
+            auto secondaryVertex2XiHyp = df2.getPCACandidate();
+
+            if (std::abs(casc.mXi() - massXi) < cascadeMassWindow) {
+              SETBIT(hfFlag, aod::hf_cand_casc_lf::DecayType2Prong::XiczeroOmegaczeroToXiPi);
```
fgrosa, line 3298: This should be only set after you test `(mass2ProngXiHyp < massXiPiMin || mass2ProngXiHyp > massXiPiMax)` (line `3304` of current draft)

PWGHF/DataModel/CandidateReconstructionTables.h
ZFederica: I only realized today that they were not used in any lf_cascade table, if I did not overlook anything

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -3253,17 +3272,99 @@ struct HfTrackIndexSkimCreatorLfCascades {
...
+
+            std::array<float, 3> pVecOmega = {0.};
+            std::array<float, 3> pVecPion1OmegaHyp = {0.};
+            df2.getTrack(0).getPxPyPzGlo(pVecOmega);
+            df2.getTrack(1).getPxPyPzGlo(pVecPion1OmegaHyp);
+
+            auto secondaryVertex2OmegaHyp = df2.getPCACandidate();
+
+            if (std::abs(casc.mOmega() - massOmega) < cascadeMassWindow) {
+              SETBIT(hfFlag, aod::hf_cand_casc_lf::DecayType2Prong::OmegaczeroToOmegaPi);
```
fgrosa, line 3333: Same here, this should be only set after you test (mass2ProngOmegaHyp < massOmegaPiMin || mass2ProngOmegaHyp > massOmegaPiMax) (line 3304 of current draft)

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -3253,17 +3272,99 @@ struct HfTrackIndexSkimCreatorLfCascades {
...
+            if (fillHistograms && statusMassOmegaPiChannel) {
+              registry.fill(HIST("hVtx2ProngXOmegaHyp"), secondaryVertex2OmegaHyp[0]);
+              registry.fill(HIST("hVtx2ProngYOmegaHyp"), secondaryVertex2OmegaHyp[1]);
+              registry.fill(HIST("hVtx2ProngZOmegaHyp"), secondaryVertex2OmegaHyp[2]);
+              registry.fill(HIST("hMassOmegacZeroToOmegaPi"), mass2ProngOmegaHyp);
+            }
+          }
+
+          // fill table row only if a vertex was found
+          if ((nVtxFrom2ProngFitterXiHyp != 0 && statusMassXiPiChannel) || (nVtxFrom2ProngFitterOmegaHyp != 0 && statusMassOmegaPiChannel)) {
```
fgrosa, line 3353: No need to test `nVtxFrom2ProngFitterXiHyp != 0` because the flag is also `0` if there is no vertex found. Moreover, `statusMassXiPiChannel` is somehow redundant (you can simply check if `hfFlag != 0`

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -3253,17 +3272,99 @@ struct HfTrackIndexSkimCreatorLfCascades {
...
+
+            if (mass2ProngXiHyp < massXiPiMin || mass2ProngXiHyp > massXiPiMax) {
+              statusMassXiPiChannel = false;
+            }
+
+            // fill histograms
+            if (fillHistograms && statusMassXiPiChannel) {
+              registry.fill(HIST("hVtx2ProngXXiHyp"), secondaryVertex2XiHyp[0]);
+              registry.fill(HIST("hVtx2ProngYXiHyp"), secondaryVertex2XiHyp[1]);
+              registry.fill(HIST("hVtx2ProngZXiHyp"), secondaryVertex2XiHyp[2]);
```
fgrosa, line 3312: Do we need them given that we do not use the secondary vertex in the selection?

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -3276,94 +3377,69 @@ struct HfTrackIndexSkimCreatorLfCascades {
...
+                if (mass3Prong < massXiPiPiMin || mass3Prong > massXiPiPiMax) {
+                  statusMassXiPiPiChannel = false;
+                }
 
-          const auto& secondaryVertex2 = df2.getPCACandidate();
+                // fill histograms
+                if (fillHistograms && statusMassXiPiPiChannel) {
+                  registry.fill(HIST("hVtx3ProngX"), secondaryVertex3[0]);
+                  registry.fill(HIST("hVtx3ProngY"), secondaryVertex3[1]);
+                  registry.fill(HIST("hVtx3ProngZ"), secondaryVertex3[2]);
```
fgrosa, line 3420: Do we need them given that we do not use the secondary vertex in the selection?

PWGHF/TableProducer/trackIndexSkimCreator.cxx
ZFederica: Sure I can move it, but I think it would not make any difference because I fill the histograms and the table only if the charm baryon mass is within the specified range, so the bit would not appear anywhere if this condition is not satisfied

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -3276,94 +3377,69 @@ struct HfTrackIndexSkimCreatorLfCascades {
...
+                  registry.fill(HIST("hMassXicPlusToXiPiPi"), mass3Prong);
+                }
+              }
 
-          // fill table row
-          rowTrackIndexCasc2Prong(casc.globalIndex(),
-                                  trackPion1.globalIndex());
+              // fill table row only if a vertex was found
+              if (nVtxFrom3ProngFitterXiHyp != 0 && statusMassXiPiPiChannel) {
```
fgrosa, line 3426: Same as above, no need to check `nVtxFrom3ProngFitterXiHyp != 0` and `statusMassXiPiPiChannel` is somehow redundant (`hfFlag` is sufficient)

PWGHF/TableProducer/trackIndexSkimCreator.cxx
ZFederica: No, they were already in place before so I did not delete them, but I could remove them

PWGHF/TableProducer/trackIndexSkimCreator.cxx
fgrosa: Indeed, but it is misleading for those who read the code (also see my comment below, I suggest to remove the status that is redundant ;))

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -3124,123 +3135,131 @@ struct HfTrackIndexSkimCreatorLfCascades {
...
+        }
+
+        //-------------------------- V0 info---------------------------
+        // pion & p <- V0 tracks
+        auto trackParCovV0PosDau0 = getTrackParCov(trackV0PosDau);
+        auto trackParCovV0NegDau1 = getTrackParCov(trackV0NegDau);
 
-        auto trackParVarXiDauCharged = getTrackParCov(trackXiDauCharged);
+        //-------------------------- cascade info----------------------
+        auto trackParCovXiDauCharged = getTrackParCov(trackXiDauCharged); // pion <- casc track to be processed with DCAfitter
```
fgrosa, line 3231: These seems not to be used (and make the compilation fail)

PWGHF/TableProducer/trackIndexSkimCreator.cxx
fgrosa: I would remove them indeed if they are not necessary

PWGHF/DataModel/CandidateReconstructionTables.h
fgrosa: At least the MC matching flags should be used in the creators, but ok they can be readded later on when needed. Thanks for the clarification!

fgrosa:
> Hello @fgrosa, thanks for your suggestions, now they should be all implemented.

Hi @ZFederica thanks a lot! I have few remaining comments and then the PR is good for me. It would be good to test if it works properly before merging it.

PWGHF/TableProducer/trackIndexSkimCreator.cxx
ZFederica: No, they were already in place before so I did not delete them, but I could remove them

ZFederica:
Hi @fgrosa, I just got a doubt: do you think it is fine to use the same track object for the xi to reconstruct both the 2prong and the 3prong decay? I was wondering this because I do the xi propagation when reconstructing the 2prong decay

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -3253,17 +3248,77 @@ struct HfTrackIndexSkimCreatorLfCascades {
...
+
+            std::array<std::array<float, 3>, 2> arrMomToXi = {pVecXi, pVecPion1XiHyp};
+            auto mass2ProngXiHyp = RecoDecay::m(arrMomToXi, arrMass2Prong[hf_cand_casc_lf::DecayType2Prong::XiczeroOmegaczeroToXiPi]);
+
+            if ((std::abs(casc.mXi() - massXi) < cascadeMassWindow) && (mass2ProngXiHyp >= massXiPiMin) && (mass2ProngXiHyp <= massXiPiMax)) {
+              SETBIT(hfFlag, aod::hf_cand_casc_lf::DecayType2Prong::XiczeroOmegaczeroToXiPi);
+            }
+
+            // fill histograms
+            if (fillHistograms && (hfFlag != 0)) {
```
fgrosa, line 3277:
```suggestion
            if (fillHistograms && (TESTBIT(hfFlag, aod::hf_cand_casc_lf::DecayType2Prong::XiczeroOmegaczeroToXiPi))) {
```

fgrosa:
> Hi @fgrosa, thanks for checking! I just got a doubt: do you think it is fine to use the same track object for the xi to reconstruct both the 2prong and the 3prong decay? I was wondering this because I do the xi propagation when reconstructing the 2prong decay.

Thanks a lot! I just have two minor comments and then it's ok for me.
Regarding your question, I would say that it should be fine (also because they should already be very close to the PV and hence close to the new 3prong vertex), but it might be better to crosscheck with some strangeness expert offline.

fgrosa (dismissed):

fgrosa (approved):

PR #3893, opened 2023-11-17
remove the post-analysis

glromane:
Splitting the original pull-request (https://github.com/AliceO2Group/O2Physics/pull/3675/) into two.


glromane:
Dear Victor,

To simplify the review process we splitted the original commit into two: now we commit only the data model (as it was agreed) and later we will commit  the post-processing task after benchmarking them. Thank you.

Sincerely yours, Gleb.

victor-gonzalez (approved):
Thanks for considering my suggestion!
If you need help with the suggested way of moving forward don't hesitate to contact us  back
We are here to help!

victor-gonzalez:
Please, check and fix the linter errors by following the `Details` link in the `MegaLinter` check

victor-gonzalez:
I took the freedom to rename the PR
If you have a better name, please, move forward

njacazio:
@victor-gonzalez many thanks for the review! We fixed the linter errors, it should be all green now!

victor-gonzalez (approved):
The use of `make_pair` in your tracks loop will penalize your execution
As your pair is ephemeral you don't really need this structure
I would suggest you not to use it

PR #3895, opened 2023-11-17
EventFiltering/PWGHF: remove obsolete task and add new task to check charm signals with trigger selections

fgrosa:
@zhangbiao-phy this is the implementation that we discussed offline to better study our trigger selections. Could you please take a look to cross check it? Thanks!


zhangbiao-phy:
> @zhangbiao-phy this is the implementation that we discussed offline to better study our trigger selections. Could you please take a look to cross check it? Thanks!

zhangbiao-phy:
> @zhangbiao-phy this is the implementation that we discussed offline to better study our trigger selections. Could you please take a look to cross check it? Thanks!

Hi @fgrosa, Thank you so much for this development! The new task is the simplified version of the HFFilter.cxx but with the specific charm species in the candidate level for particle the anti-particle, which is good for me.

mpuccio (approved):

PR #3900, opened 2023-11-18
mcgen etaHistogram and eventcounter added

hhesouno:


PWGMM/Tasks/multiplicityPbPb.cxx
```diff
@@ -115,6 +118,19 @@ struct multiplicityPbPb {
 
     histos.fill(HIST("Anton/NtrkZvtxEvents"), trackCounter, collision.posZ());
   }
+  PROCESS_SWITCH(multiplicityPbPb, process, "process real and MC data", true);
```
aalkin, line 121: The function with the name `process()` is executed unconditionally, there is no need to add a switch. If you want to make the function switchable, it should be named differently.

aalkin (changes requested):

PR #3907, opened 2023-11-20
[Common/event plane determination] making sub tables and saving track id

jikim1290:
making sub tables and saving track id


ddobrigk (approved):
Hi @jikim1290 this looks ok to me! A quick comment: I see you have arrays elements as individual regular columns, while in principle (for readability) you could also use array columns without issues. But OK, this we can leave for a future development. Thanks!

PR #3912, opened 2023-11-20
[BUG] Incorrect Preslice

aalkin:
@fjonasALICE 
I have added a check in O2 to detect invalid Preslice (https://github.com/AliceO2Group/AliceO2/pull/12267) and this task triggers it. Here is the suggested change. Could you please verify that this is a correct fix and merge or push your own? The change in O2 cannot be merged otherwise, as it will break O2Physics compilation for everyone. Thank you.


mfasDa (approved):
Hi @aalkin, thanks a lot for fixing!

PR #3916, opened 2023-11-20
MegaLinter: Make cpplint and clang-format check all C++ files. Disable clang-format

vkucera:
This configuration provides cpplint and clang-format with an exhaustive list of extensions of files to check.
The list is a union of the cpplint default list and the extensions of C++ files found in the O2 and O2Physics repositories, most importantly it adds `.C` ROOT macros.


vkucera:
Disabling clang-format to avoid conflicts with the formatting CI check due to different versions of clang-format.

vkucera:
@ddobrigk Does it look good to you?

ddobrigk (approved):

PR #3924, opened 2023-11-21
EventFiltering: fix missing magnetic field for track propagation

fgrosa:


alibuild (approved):

Issue #3761, opened 2023-11-02
PWGHF: Add PID selection in the track skimming


vkucera:
One could reduce the combinatorial background in the track index skimming by applying a PID cut on proton candidates.

Things to consider:

- TPC n_sigma cut at low pT.
- Extra dependency on PID wagons + related memory overhead.
-  Tracks should be flagged at the track selection level to process each track only once and to make use of it both in the 3-prong (p K pi) and the cascade (p K0S) vertexing devices of the skimming workflow.
- Two hypotheses for Lc 3-prong candidates (p K pi, pi K p).

PR #3939, opened 2023-11-22
PWGHF: Implement the possibility to apply proton PID for baryons in skimming

fgrosa:
This PR addresses the issue https://github.com/AliceO2Group/O2Physics/issues/3761 implementing the possibility to apply a PID selection for the proton in the Lc and Xic decays. Main features:
- not to force the dependency on the PID tasks, the process function is templatised (allowing also for the depency on TPC or TOF separately if only one of the two is needed)
- the tracks are tagged as protons or not only once in the struct for the track tagging
- a different selection for each channel can be set

In addition, in this PR I removed the `MY_DEBUG` option.

I tested that it works properly in all the cases (so it is ready for review), but before merging it I would like to compare in detail the outputs before and after this implementation and profile the performances (I will post here the results in the next days).


vkucera:
Thanks @fgrosa , I would remove all the occurrences of `MY_DEBUG` (also in `candidateCreatorCascade.cxx`) together with `utilsDebugLcToK0sP.h`.

mfaggin:
Hi @fgrosa , I looked at the code and it looks ok from my side. Feel free to approve it by yourself after yor tests.

fgrosa:
> Hi @fgrosa , I looked at the code and it looks ok from my side. Feel free to approve it by yourself after yor tests.

Hi @mfaggin, after few fixes, I checked that the same result is obtained in case of no PID selection enabled. If the PID is enabled instead, only for the baryons (Lc->pKpi in my test) is obtained a difference as expected.

image

The same holds both in case of no PVrefit and PVrefit.

I then checked the CPU and memory performance with the previous version:

image
image
 
And the new version with no PID applied:

image
image

And no overhead is observed. 
In case of PID the only overhead is due to the dependencies on the TPC and/or TOF PID tasks.

mfaggin:
Thanks Fabrizio, it looks good to me, thank you

mfaggin (approved):

PR #3941, opened 2023-11-22
Fix the mult range parameter

plariono:
This PR is to rename the multEtaRange parameter of the OTF tracker config to multEtaHalfRange as it is effectively so. In L303 it is multiplied by 2. 
Tagging @ddobrigk to review.

### DIFF
--- a/ALICE3/TableProducer/OTF/onTheFlyTracker.cxx
+++ b/ALICE3/TableProducer/OTF/onTheFlyTracker.cxx
@@ -68,7 +68,7 @@ struct OnTheFlyTracker {
 
   Configurable<float> magneticField{"magneticField", 20.0f, "magnetic field in kG"};
   Configurable<float> maxEta{"maxEta", 1.5, "maximum eta to consider viable"};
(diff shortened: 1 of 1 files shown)


njacazio (approved):

ddobrigk:
@plariono if you find that the current name is unclear, maybe it would be best to rename the parameter `multEtaRange` to `multMaxEta` instead... because what is actually happening is that this parameter is the maximum allowed eta, and the value `2.0f`  in line 303 is simply because, when requiring `-a<eta+a`, the width of the window is 2a. Could you please change the name into `multMaxEta`? Thank you!

plariono:
@ddobrigk I renamed it to `multMaxEta`.
Yes I understand the logic but the current name was unclear. Without looking at the actual code I was setting the range to 8 for ALICE 3 which made the dNdEta calculation wrong (and some cuts as well).

ddobrigk:
@plariono , yes, I see your point. I think the new name now is indeed better. Thanks!!

PR #3942, opened 2023-11-22
Changing propagation of muon in fwdtrackextension task

mcoquet642:
This PR changes the propagation of muons in this task from linear to helix.


mcoquet642:
Hello, can this be merged if there isn't any problem with it? Thanks.

PR #3952, opened 2023-11-23
ML, PWGHF: add possibility to pass CCDB paths

fgrosa:
This PR is needed to fix the download of the ML model files from CCDB with [ccdbApi.retrieveBlob](https://github.com/AliceO2Group/AliceO2/blob/a13ea078621d5e4a3a31180d341bfe16eba0b10a/CCDB/include/CCDB/CcdbApi.h#L283). In fact without this fix, if there are multiple files in the same CCDB directory they are all downloaded and copied locally to the same path, meaning that if one has multiple models in CCDB the last one that is downloaded it is copied in each local file. 
There was no issue in the past, so probably it is due to a change of behaviour in the `ccdbApi`.
@fmazzasc @fcatalan92 can you please approve from the ML side? Thanks!


fmazzasc (approved):
Looks good to me, thanks!

alibuild (approved):

PR #3956, opened 2023-11-23
PWGHF: Improve tree creator for Omega_c with ST

qgp:
- implement improvements from ST filter to tree creator
- add missing variables to tree

### DIFF
--- a/PWGHF/TableProducer/treeCreatorOmegacSt.cxx
+++ b/PWGHF/TableProducer/treeCreatorOmegacSt.cxx
@@ -146,7 +146,6 @@ DECLARE_SOA_TABLE(HfOmegaStGen, "AOD", "HFOMEGACSTGEN",
 } // namespace o2::aod
 
 struct HfTreeCreatorOmegacSt {
-  Configurable<double> bz{"bz", -5., "magnetic field"};
   Configurable<int> materialCorrectionType{"materialCorrectionType", static_cast<int>(o2::base::Propagator::MatCorrType::USEMatCorrLUT), "Type of material correction"};
   Configurable<std::string> ccdbUrl{"ccdbUrl", "http://alice-ccdb.cern.ch", "url of the ccdb repository"};
   Configurable<std::string> grpMagPath{"grpMagPath", "GLO/Config/GRPMagField", "CCDB path of the GRPMagField object"};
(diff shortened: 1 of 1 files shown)


PWGHF/TableProducer/treeCreatorOmegacSt.cxx
```diff
@@ -277,15 +290,47 @@ struct HfTreeCreatorOmegacSt {
...
+          decayLengthOmega = RecoDecay::distance(secondaryVertex, primaryVertexPos);
+          decayLengthOmegaXY = RecoDecay::distanceXY(secondaryVertex, primaryVertexPos);
+          trackParV0 = df2.getTrackParamAtPCA(0);
+          trackParBachelor = df2.getTrackParamAtPCA(1);
+          trackParV0.getPxPyPzGlo(momentaOmegaDaughters[0]);
+          trackParBachelor.getPxPyPzGlo(momentaOmegaDaughters[1]);
+          std::array<float, 3> pVec;
+          df2.createParentTrackParCov().getPxPyPzGlo(pVec);
+          std::array<float, 3> pvPos = {primaryVertex.getX(), primaryVertex.getY(), primaryVertex.getZ()};
+          cpaOmega = RecoDecay::cpa(pvPos, df2.getPCACandidate(), pVec);
```
vkucera, line 324: Can you not reuse `primaryVertexPos` here instead of creating `pvPos`? Or is there a difference because of `float` vs `double`?

PWGHF/TableProducer/treeCreatorOmegacSt.cxx
```diff
@@ -277,15 +290,47 @@ struct HfTreeCreatorOmegacSt {
...
+        } else {
+          continue;
+        }
+      } else {
+        continue;
+      }
```
vkucera, line 330: These are not needed if you negate the `if` conditions and `continue` immediately (unless you anticipate replacing `continue` with something else later).

PWGHF/TableProducer/treeCreatorOmegacSt.cxx
qgp: Indeed, I can use `primaryVertexPos`.

PWGHF/TableProducer/treeCreatorOmegacSt.cxx
qgp: Indeed, that's nicer.

qgp:
Hi @vkucera, thanks for your suggestions, both incorporated now.

vkucera (approved):

PR #3962, opened 2023-11-24
EventFiltering/PWGHF: make axes configurable to tune memory

fgrosa:


alibuild (approved):

PR #3971, opened 2023-11-24
Next centrality

njacazio:


ddobrigk (approved):
thanks @njacazio !

jgrosseo:
@njacazio @ddobrigk 
And "next centrality" broke Run 2 analysis again... Could this be tested more thoroughly, please?
https://alimonitor.cern.ch/hyperloop/wagon-test/137602

njacazio:
Hi @jgrosseo indeed you're right, I believed that as for the multiplicity updating the HL configuration would be enough, but this time I had introduced a bug for the Run 2 case, see https://github.com/AliceO2Group/O2Physics/pull/3985 for the fix. I'll make sure to validate it on run2 as well.

PR #3972, opened 2023-11-24
Allow running slightly more parallel analysis compile jobs

TimoWilken:
On build machines with little memory, such as the ARM VMs we have been allocated, compiling with the previous formula takes much too long (5 hours just for O2Physics).

On those machines, reserving 4 GiB for the system is overkill and cuts down the number of build slots unnecessarily. Reducing this to 2 GiB allows us to squeeze in one more build slot, bringing the total to 3 instead of 2.

The peak memory use per analysis build process looks accurate at around 4 GiB.

In theory, this change should reduce the time taken for analysis compilation on those machines by 1/3.


ddobrigk (approved):
Thanks @TimoWilken, this looks reasonable!

PR #3977, opened 2023-11-24
[PWGDQ] Fixing the index format for MFT in ReducedMuonsExtra

lucamicheletti93:
Fixing the index format for MFT in ReducedMuonsExtra and declaring it as for MCH tracks. 
This modification fixes the problem at AOD merging in hyperloop.


alibuild (approved):

PR #3986, opened 2023-11-25
Clean LF centrality/multiplicity table sub

njacazio:
Table required split to avoid non needed CPU usage


alibuild (approved):

PR #4005, opened 2023-11-27
PWGHF: Fix tree creator for Omega_c using ST

qgp:
- add charge of Omega and bachelor
- fix track association
- streamlining
- update histograms

### DIFF
--- a/PWGHF/TableProducer/treeCreatorOmegacSt.cxx
+++ b/PWGHF/TableProducer/treeCreatorOmegacSt.cxx
@@ -59,11 +59,11 @@ DECLARE_SOA_COLUMN(NSigmaTofPi, nSigmaTofPi, float);
 DECLARE_SOA_COLUMN(PxOmega, pxOmega, float);
 DECLARE_SOA_COLUMN(PyOmega, pyOmega, float);
 DECLARE_SOA_COLUMN(PzOmega, pzOmega, float);
-DECLARE_SOA_COLUMN(ChargeOmega, chargeOmega, bool);
+DECLARE_SOA_COLUMN(IsPositiveOmega, isPositiveOmega, bool);
 DECLARE_SOA_COLUMN(PxPion, pxPion, float);
 DECLARE_SOA_COLUMN(PyPion, pyPion, float);
 DECLARE_SOA_COLUMN(PzPion, pzPion, float);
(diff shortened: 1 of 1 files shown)


PWGHF/TableProducer/treeCreatorOmegacSt.cxx
```diff
@@ -405,9 +407,11 @@ struct HfTreeCreatorOmegacSt {
                               momenta[0][0], // omega momentum
                               momenta[0][1],
                               momenta[0][2],
+                              trackCasc.sign(),
```
mfaggin, line 410: The method `sign()` should return a `short` (https://github.com/AliceO2Group/AliceO2/blob/f63f7a183882b899c34e11eddf84e59e7a00836d/Framework/Core/include/Framework/AnalysisDataModel.h#L132C53-L132C53). Did you verify that this fill works well, given that the corresponding column expectes a `bool`? Otherwise you should probably cast this variable into a `bool`.

PWGHF/TableProducer/treeCreatorOmegacSt.cxx
```diff
@@ -405,9 +407,11 @@ struct HfTreeCreatorOmegacSt {
...
                               momenta[1][0], // pion momentum
                               momenta[1][1],
                               momenta[1][2],
+                              track.sign(),
```
mfaggin, line 414: The method `sign()` should return a `short` (https://github.com/AliceO2Group/AliceO2/blob/f63f7a183882b899c34e11eddf84e59e7a00836d/Framework/Core/include/Framework/AnalysisDataModel.h#L132C53-L132C53). Did you verify that this fill works well, given that the corresponding column expectes a `bool`? Otherwise you should probably cast this variable into a `bool`.

mfaggin (commented):
Hi @qgp , it looks good to me. I just have a small doubt, see the detailed comments.

Best. Mattia

PWGHF/TableProducer/treeCreatorOmegacSt.cxx
qgp: Hi @mfaggin, good catch, I will make this an explicit conversion. Thanks!

mfaggin (approved):
thanks a lot @qgp , it looks ok to me

PR #4013, opened 2023-11-28
PWGLF: Update the common column name

BongHwi:
Updated from `MultV0M` to `Cent`

The variable `MultV0M` has contained the multiplicity/Centrality of the event which the type (V0M.FT0) can be set during the filter process.


alibuild (approved):

PR #4014, opened 2023-11-28
Fixes for C++20

ktf:
Fixes for C++20

* {...} needs to be escaped when used in a format string.
* Use the correct constructor for GlobalFwdTrack.


ktf:
This requires a bump in O2 to actually work.

PR #4026, opened 2023-11-30
Add option to use nsigma-based mass window for omega selection

qgp:


EventFiltering/PWGLF/strangenessFilter.cxx
```diff
@@ -129,6 +131,10 @@ struct strangenessFilter {
   Configurable<float> maxCpaTrackedXi{"maxCpaTrackedXi", 1., "Maximum CPA for tracked cascades"};
   Configurable<float> minDcaTrackedOmega{"minDcaTrackedOmega", 0., "Minimum DCA for tracked cascades"};
   Configurable<float> maxCpaTrackedOmega{"maxCpaTrackedOmega", 1., "Maximum CPA for tracked cascades"};
+  Configurable<float> parNsigmaOmega0{"parNsigmaOmega0", 0.006815, "Parameter 0 for nSigma parametrisation of Omega"};
```
mpuccio, line 134: Maybe we could use a Configurable<Array2D<float>> to support parameters for both Omega and Xis in one go. Alternatively a labeledArray

qgp: Good suggestion, implemented.

PR #4031, opened 2023-11-30
multiplicityPbPb adding Z_vtx dependent histograms for MCGEN

hhesouno:


aalkin (changes requested):
In addition, there are now PAG-specific directories. Could you please move your task from `PWGMM/Tasks` to `PWGMM/Mult/Tasks`?

aalkin (approved):

PR #4036, opened 2023-11-30
Update onTheFlyTOFPID.cxx

ddobrigk:
@hscheid please do not kill me but I think I found a bug! Can you please cross-check? Sorry!


ddobrigk:
Ah, I know why I never saw this bug leading to a crash before. It is because the starting point for many tracks in OTF simulations in the transverse plane is exactly (0, 0) - but I just realised that for tracks that do not start at (0, 0) this fix is important...

PR #4043, opened 2023-12-01
PWGHF: Add extra slots for ML models inference

AlexBigO:
Hi @fgrosa , adding some slots for a finer pT binning during ML inference.


fcatalan92 (changes requested):
Hi @AlexBigO, you should update also L107 and L110 accordingly

AlexBigO:
Yes of course, thanks!

fcatalan92:
> I must say I am not sure how to solve the alibuild error because it looks like the compiler still takes `6` (the former number of bins) while I set it to `12` in the latest commit.

I think this is just the test triggered by your first commit, which failed due to the problem you already corrected. Now the test has restarted and it should be green.

PR #4052, opened 2023-12-02
Next TOF Ev Time maker

njacazio:


ddobrigk:
Thanks @njacazio! It seems the compilation error is spurious.... but in the meantime, if I may ask: did you test this in both Run 2 and Run 3 paths?

PR #4072, opened 2023-12-05
EventFiltering: add tag as generic signal from BDT (no prompt or nonprompt)

fgrosa:


alibuild (approved):

PR #4076, opened 2023-12-05
PWGHF: implement track selections for bachelor in lfCascade+bachelor decays

ZFederica:
Modify the struct HfTrackIndexSkimCreatorTagSelTracks to implement a set of selection dedicated to the bachelor track in LF cascade+bachelor decays. Until now the selections dedicated to bachelor track of V0+bachelor decays were used also for the above-mentioned channel.


ZFederica:
Hi @fgrosa, what do you think of this implementation? This would allow to use ITSQuality tracks at the skimming step for bachelor tracks in cascade+bachelor decays without affecting the selections used for V0+bachelor decays

fgrosa (commented):
Ciao @ZFederica, the implementation looks good to me, thanks!
Before we merge it however, please cross check that you get the same output for the other channels, to exclude that I overlooked something.

ZFederica:
Hi @fgrosa, I compared the output of the trackIndexSkimCreator before and after these changes, I did not see any difference as for the other chennels. However, I noticed that I forgot to extend an histogram axis, now it is fixed.

fgrosa (approved):

PR #4079, opened 2023-12-05
PWGHF: check on number of process function enabled toXiPi task

ZFederica:
Implement a check in the init function of the candidateCreatorToXiPi to make sure that only one process function is enabled (either do the indexes combinatorics or run on derived data from trackIndexSkimCreator)

### DIFF
--- a/PWGHF/TableProducer/candidateCreatorToXiPi.cxx
+++ b/PWGHF/TableProducer/candidateCreatorToXiPi.cxx
@@ -110,6 +110,11 @@ struct HfCandidateCreatorToXiPi {
     ccdb->setLocalObjectValidityChecking();
     lut = o2::base::MatLayerCylSet::rectifyPtrFromFile(ccdb->get<o2::base::MatLayerCylSet>(ccdbPathLut));
     runNumber = 0;
+
+    if (doprocessIdxCombinatorics == true && doprocessDerivedData == true) {
+      LOGF(fatal, "Cannot enable processIdxCombinatorics and processDerivedData at the same time. Please choose one.");
+    }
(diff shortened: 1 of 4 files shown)


PWGHF/TableProducer/candidateCreatorToXiPi.cxx
```diff
@@ -110,6 +110,10 @@ struct HfCandidateCreatorToXiPi {
     ccdb->setLocalObjectValidityChecking();
     lut = o2::base::MatLayerCylSet::rectifyPtrFromFile(ccdb->get<o2::base::MatLayerCylSet>(ccdbPathLut));
     runNumber = 0;
+
+    if (doprocessIdxCombinatorics == true && doprocessDerivedData == true) {
```
fgrosa, line 114:
```suggestion
    if (doprocessIdxCombinatorics && doprocessDerivedData) {
```

fgrosa (changes requested):
Ciao @ZFederica I just have a small comment. In the future, you can consider to remove completely the process function that computes the combinatorics, since you can in principle always use the `trackIndexSkimCreator` task to do it and then you avoid to have two equivalent tasks to maintain

PWGHF/TableProducer/candidateCreatorToXiPi.cxx
vkucera: One should also check that they are not both `false`.

ZFederica:
@fgrosa yes I will remove the first process, I would just keep it for a little bit longer for validation

fgrosa (approved):

PR #4100, opened 2023-12-07
PWGMM: dndeta: update task structure

aalkin:
* Reduce code duplication
* Make process function names more clear


alibuild (approved):

PR #4101, opened 2023-12-07
Add typecast

ddobrigk:
...this should not be necessary, but apparently it is (not understood yet).


ddobrigk:
@creetz16 thanks!! :-D

njacazio (approved):

PR #4112, opened 2023-12-08
PWGJE: Broader area range for larger R

joonsukbae:
According to the previous analysis in pp collisions 13 TeV by Yongzhen Hou, jet area range should be larger in case of R = 0.6 or larger.


nzardosh:
Hi @joonsukbae, can you also please add this to the HF and Full jetfinderQA tasks as well please? Also please increase the bins to 300 as well

joonsukbae:
Hi @nzardosh, I think it should be merged before another edits:)

nzardosh (approved):

PR #4118, opened 2023-12-08
PWGHF: Add tree creator in B0 reduced task + fix bug in Dmeson ML selection

AlexBigO:
Hi @fgrosa , I am creating this draft PR so you can have a look at the code to implement a tree creator in our B0 reduced task.


fgrosa:
Hi @AlexBigO thanks for the implementation! 
However, I would not put the tree in the selector, but rather in the task. In fact, the tree should be an output object alternative to the histograms / sparses. It would be better to move the definition of the table used for the tree in the task itself (as done in the tree creators) and simply fill it or not (depending on the `fillTree` configurable) in the `fillCand` methods.

AlexBigO:
Okay, I had misunderstood :)

Here is the new implementation. In the arguments of the `process` functions, I had to add two spare subscriptions:
- `TracksPion` as we access the PID information on Pion (prong1) to fill the tree
- B0 selection table (that was only in `processData`)

I also added a safety for background filling.

Tell me if this is okay.

PWGHF/D2H/Tasks/taskB0Reduced.cxx
```diff
@@ -351,6 +437,76 @@ struct HfTaskB0Reduced {
         }
       }
     }
+    if (fillTree) {
```
mfaggin, line 440: General comment (take it more as a suggestion for a future PR): you can consider to unify the `fillCand` and `fillCandMcReco` functions in a unique `template` function, and manage with a `constexpr bool IS_MC` variable the decision whether to fill data or MC objects (and whether to do MC-specific operations).
This might help reducing code duplications and be helpful in case you develop new features (for example, as it is now if you add a new variable in the tree you must remember to add the column filling twice).

mfaggin (commented):
thanks a lot @AlexBigO , I just want to give you a few suggestions, see the detailed comments. Best. Mattia

PWGHF/D2H/Tasks/taskB0Reduced.cxx
AlexBigO: Hi @mfaggin ! Thanks a lot for the comments, I am implementing them.

Yes, I fully agree, I also thought about this template improvement, and we'll indeed keep it for a future PR :)

mfaggin (approved):
Thanks a lot @AlexBigO

PR #4123, opened 2023-12-09
Added analysis of TPC cluster number

rosarioturrisi:


DPG/Tasks/AOTTrack/qaMatchEff.cxx
```diff
@@ -734,6 +874,36 @@ struct qaMatchEff {
     histos.add("MC/PID/zDCA_tpcits_pr", "DCA along z - protons TPC+ITS tag;dca [cm]", kTH1D, {{200, -20.0, 20.0}}, true);
     histos.add("MC/PID/xyDCA_tpcits_pr", "DCA in x-y plane - protons TPC+ITS tag;dca [cm]", kTH1D, {{200, -20.0, 20.0}}, true);
 
+    //
+    // pions only
+    // all
+
+    //
+    histos.add("MC/TPCclust/tpcNClsFound_pi", "Number of TPC found clusters - pions", kTH1F, {{161, -0.5, 160.5}}, true);
```
mfaggin, line 882: what about the cases `tpcNClsFindable_pi`, `tpcCrossedRows_pi` and `tpcsFindableMinusCrossedRows_pi`?

DPG/Tasks/AOTTrack/qaMatchEff.cxx
```diff
@@ -788,6 +958,32 @@ struct qaMatchEff {
     //
     // protons only
     // all
+    //
+    //
+    // TPC found clusters distribution
+    histos.add("MC/TPCclust/tpcNClsFound_pr", "Number of TPC found clusters - protons", kTH1F, {{161, -0.5, 160.5}}, true);
```
mfaggin, line 964: same as for `_pi`

DPG/Tasks/AOTTrack/qaMatchEff.cxx
```diff
@@ -814,6 +1010,32 @@ struct qaMatchEff {
     //
     // kaons only
     // all
+
+    // TPC found clusters distribution
+    histos.add("MC/TPCclust/tpcNClsFound_ka", "Number of TPC found clusters - kaons", kTH1F, {{161, -0.5, 160.5}}, true);
```
mfaggin, line 1015: same as for `_pi`

DPG/Tasks/AOTTrack/qaMatchEff.cxx
```diff
@@ -1128,8 +1358,86 @@ struct qaMatchEff {
...
+      }
+
+      //  TPC clusters all pions
+      //
+      if (isPion) {
+        if constexpr (IS_MC) { ////////////////////////   MC
+          histos.get<TH1>(HIST("MC/TPCclust/tpcNClsFound_pi"))->Fill(clustpc);
+          histos.get<TH1>(HIST("MC/TPCclust/tpcNClsFindable_pi"))->Fill(findcltpc);
+          histos.get<TH1>(HIST("MC/TPCclust/tpcCrossedRows_pi"))->Fill(crowstpc);
+          histos.get<TH1>(HIST("MC/TPCclust/tpcsFindableMinusCrossedRows_pi"))->Fill(crowstpc);
```
mfaggin, line 1383: in my opinion these lines must be put inside the scope here: https://github.com/AliceO2Group/O2Physics/pull/4123/files#diff-a7ac2a00ff9015a809fe16e13dad8125e5e29054d61b13c605d1ef7dee975d96R1364-R1368

In the way you are writing now, everything is protected by `isPion`, which implies the application of a PID selection and this is not what you usually show

DPG/Tasks/AOTTrack/qaMatchEff.cxx
```diff
@@ -1128,8 +1358,86 @@ struct qaMatchEff {
...
+        }
+      }
+      //  TPC clusters all protons
+      //
+      if (isProton) {
+        if constexpr (IS_MC) { ////////////////////////   MC
+          histos.get<TH1>(HIST("MC/TPCclust/tpcNClsFound_pr"))->Fill(clustpc);
+          histos.get<TH1>(HIST("MC/TPCclust/tpcNClsFindable_pr"))->Fill(findcltpc);
+          histos.get<TH1>(HIST("MC/TPCclust/tpcCrossedRows_pr"))->Fill(crowstpc);
+          histos.get<TH1>(HIST("MC/TPCclust/tpcsFindableMinusCrossedRows_pr"))->Fill(crowstpc);
```
mfaggin, line 1413: same comment as above for `isPion`

DPG/Tasks/AOTTrack/qaMatchEff.cxx
rosarioturrisi: this is for "all" pions, without any selection but the pid, to keep the task from diverging in size tried to put this one as a reference. It is true that ~10 histos more will not make a real difference... Real point is that should review this task :)

DPG/Tasks/AOTTrack/qaMatchEff.cxx
rosarioturrisi: Right, it could be useful to make a "pears to pears" comparison with MC, but it is also true that can keep the old way to compare...12 more hosts, again not a big deal :(

mfaggin: Sorry, probably I'm just confused due to the too large amount of filled histograms (we really must rearrange the task in the near future).

 1. Here you are proposing a few histograms to check what happens in MC when looking at tracks reconstructed as in data, w/o specific track selections
 2. In this place instead https://github.com/AliceO2Group/O2Physics/pull/4123/files#diff-a7ac2a00ff9015a809fe16e13dad8125e5e29054d61b13c605d1ef7dee975d96R2287-R2290 you look only at true pions reconstructed at least with TPC
 
 Am I right? Then I would change the name of the histos in case 1. e.g. by adding a suffix (e.g. `_PID` or something that allows us to understand that it is as in data).

mfaggin (approved):
thanks @rosarioturrisi . For next time, we should think to reshape a bit the task (i.e. merging histograms into TH2, TH3, ...) and make it less crowded

PR #4135, opened 2023-12-11
PWGCF - adding DO mesons to FemtoUniverse

kgwizdzi:
Adding the code for D0/D0bar mesons to the FemtoUniverse. I've modified .h files in the Core folder. I've added missed part for TempFitVarName in the FemtoDerived.h file. In the producer I've added the function for filling table with D mesons and process for them. I've modified CMakeLIsts.txt, because I've also added a first version of a task to run D0 analysis (check the simple distributions, like invMass, pT).

### DIFF
--- a/PWGCF/FemtoUniverse/TableProducer/femtoUniverseProducerTask.cxx
+++ b/PWGCF/FemtoUniverse/TableProducer/femtoUniverseProducerTask.cxx
@@ -229,7 +229,7 @@ struct femtoUniverseProducerTask {
     Configurable<int> ConfPDGCodePartTwo{"ConfPDGCodePartTwo", 321, "Particle 2 - PDG code"};
   } ConfPhiChildTwo;
 
-  //D0/D0bar mesons
+  // D0/D0bar mesons
   struct : o2::framework::ConfigurableGroup {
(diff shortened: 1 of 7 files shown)


PWGCF/FemtoUniverse/TableProducer/femtoUniverseProducerTask.cxx
```diff
@@ -440,6 +453,13 @@ struct femtoUniverseProducerTask {
                        -999., -999.,
                        -999., -999., -999.,
                        -999.); // QA for phi
+    } else if constexpr (isD0) {
```
zchochul, line 456: Maybe the better way to go would be to name the bool isPhi -> isPhiOrD0? It would minimize the amount of code ;)

zchochul: If you change that, I'm happy to approve the changes

kgwizdzi: Done. I've changed it.

zchochul (approved):

PR #4141, opened 2023-12-12
Generic Framework changes

EmilGorm:
- Changes to GFW folder structure
- Added GFWConfig core class
- Changes to flowGenericFramework task
   - Now uses GFWConfig
   - process functions for MC gen and reco
   - Updated QA plots

Change in PWGDQ/dqFlow.cxx
   - Changed includes to reflect the change in GFW folder structure


PWGCF/GenericFramework/Tasks/flowGenericFramework.cxx
```diff
@@ -0,0 +1,445 @@
...
+          QAregistry.fill(HIST("phi_reco"), particle.phi());
+          QAregistry.fill(HIST("eta_reco"), particle.eta());
+          QAregistry.fill(HIST("pt_reco"), particle.pt());
+        }
+      }
+    }
+    for (uint l_ind = 0; l_ind < corrconfigs.size(); l_ind++) {
+      FillFC(fFC_reco, corrconfigs.at(l_ind), centrality, l_Random);
+    }
+  }
```
victor-gonzalez, line 436: In principle your MC reco processing should not differ from your data processing
The only difference should be the signature of your process function and the fact that in the case of the MC reco you check for potentially associated `mcparticle` 

For your next iteration consider the following construction for your `processData` function 
```
template <typename TrackObject>
inline int8_t ProcessTrack(TrackObject const& track)
{
  /* your processing for track either data or MC reco */

  if constexpr (framework::has_type_v<aod::mctracklabel::McParticleId, typename TrackObject::all_columns>) {
    if (track.mcParticleId() < 0) {
      /* no label!! process accordingly */
    } else {
      /* do your specific MC reco processing */
    }
  }

  /* your additional processing for track either data or MC reco */
```

What is inside the if `constexpr () {...}` is handled by the compiler in the appropriate way. In this way you don't need to replicate your code for data and for MC reco. Replicating code is always error prone! 

You can obtain a similar effect with your `datatype` template parameter

victor-gonzalez (approved):
I approve it for the time being but consider my comment on no replicating code for the next iteration

iarsene (approved):

PR #4150, opened 2023-12-13
M ctemplates

feisenhu:
Addition to #4144.
I forgot to include the pair cut as an `AnalysisCompositeCut`
Introduce second pair cut with mass interval


alibuild (approved):

PR #4151, opened 2023-12-13
Change default V0Type to standard V0

ddobrigk:
@shahor02

### DIFF
--- a/Common/TableProducer/v0converter.cxx
+++ b/Common/TableProducer/v0converter.cxx
@@ -23,7 +23,7 @@ struct V0Converter {
   void process(aod::V0s_001 const& v0s)
   {
     for (auto& v0 : v0s) {
-      uint8_t bitMask = static_cast<uint8_t>(1);
+      uint8_t bitMask = static_cast<uint8_t>(2); // first bit on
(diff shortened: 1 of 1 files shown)


ddobrigk:
So just so I don't get this wrong: bit zero should off, bit 1 should be on -> this means `uint8_t = 2`. Please shout in case I am wrong!

Common/TableProducer/v0converter.cxx
```diff
@@ -23,7 +23,7 @@ struct V0Converter {
   void process(aod::V0s_001 const& v0s)
   {
     for (auto& v0 : v0s) {
-      uint8_t bitMask = static_cast<uint8_t>(255);
+      uint8_t bitMask = static_cast<uint8_t>(2); // second bit on
```
shahor02, line 26: I think the default  should be `uint8_t bitMask = static_cast<uint8_t>(1); // 1st bit on`
since this means standard v0. 2nd bit is set additionally to flag photon-only check.

ddobrigk: Hi @shahor02 - sorry, yes, fixed now! Thanks!

alibuild (approved):

PR #4152, opened 2023-12-13
PWGHF: Fix delta mass (phi,KK) selection in track skimming

fchinu:
This PR fixes a bug in the selection on the invariant mass difference of the reconstructed kaons pair and the phi meson


fgrosa (approved):
Many thanks @fchinu!

fchinu:
> Many thanks @fchinu!

Thanks a lot to you @fgrosa!

PR #4156, opened 2023-12-13
Switch table producers that produce V0s to produce V0s_001 temporarily

ddobrigk:
@shahor02 this is required so that the change in O2 can go or, indeed, the analysis CI might complain as the new version of O2 will not compile against the older version of O2Physics. Once O2Physics switches to being built against the new O2, I can switch these tasks to produce the generic `V0s`. 

Technically, once this PR is merged, we could ignore the fullCI because then for sure the compilation will work...


njacazio (approved):
Thanks for following this David

PR #4173, opened 2023-12-15
PWGJE : Adding the ability to cut on pT hat in the QA tasks

nzardosh:
Based on work by @dajones2


alibuild (approved):

PR #4196, opened 2023-12-18
PWGHF:Add missing repropagation of reassociated tracks

fgrosa:
In this PR I added the re-propagation of re-associated pion tracks (not done before). 
In addition, it removes the usage of the LUT for the DCAFitter since HF decays should always be within the beampipe, there is no need of material correction (this should be fixed everywhere in HF code, since it allows us to spare memory)


alibuild (approved):

PR #4201, opened 2023-12-18
Safety changes for V0 type

ddobrigk:
This PR takes into account the recent implementation of the `V0Type` column in the V0 data model to ensure safety of the analysis of V0s. It means to compartmentalize the standard V0s and the V0s built solely for the purpose of cascade finding, since these two populations have very different selection criteria at the secondary vertexer level and therefore cannot be mixed. 

The main change is that standard V0s (including photons, if those are meant to be built) are saved in `V0Datas`, while V0s built *solely* for the sake of cascade analysis are built in the `V0fCDatas` ('for Cascades') table. This means that no duplication of `DCAFitter` calls takes place at building/analysis time, while making absolutely sure that the V0s dedicated solely for cascade building are not analysed by mistake by an inadvertent analyser who missed the need to check the `V0Type` (which is also duplicated further in the analysis tables for completeness). 

The `V0DataLink` table now has two index columns: a column to `V0Datas` if the V0 passes standard V0 cuts and a column to `V0fCDatas` if the V0 does *not* pass V0 cuts but passes cascade cuts. Of course, this means that logically only one of these can ever be non-negative at a single time. 

An important side effect of this is that de-referencing a `V0Datas` table from a cascade analysis is strongly discouraged. For now, I have disabled that functionality altogether, but may add it as a separate table soon. However, it is only to be used with great care. On the positive side, this means that, while previously a cumbersome de-referencing step was necessary to de-reference track indices when dealing with cascades, this is now trivial as the positive and negative V0 track indices are referenced directly (instead of the original V0 index).

### DIFF
--- a/PWGLF/DataModel/LFStrangenessTables.h
+++ b/PWGLF/DataModel/LFStrangenessTables.h
@@ -121,7 +121,7 @@ DECLARE_SOA_INDEX_COLUMN(V0, v0);                                       //!
 DECLARE_SOA_INDEX_COLUMN_FULL(PosTrackExtra, posTrackExtra, int, DauTrackExtras, "_PosExtra"); //!
 DECLARE_SOA_INDEX_COLUMN_FULL(NegTrackExtra, negTrackExtra, int, DauTrackExtras, "_NegExtra"); //!
 DECLARE_SOA_INDEX_COLUMN(StraCollision, straCollision);                                        //!
-DECLARE_SOA_INDEX_COLUMN(MotherMCPart, motherMCPart);                                  //!
+DECLARE_SOA_INDEX_COLUMN(MotherMCPart, motherMCPart);                                          //!
 
 //______________________________________________________
 // REGULAR COLUMNS FOR V0CORES
@@ -555,7 +555,7 @@ DECLARE_SOA_INDEX_COLUMN_FULL(NegTrackExtra, negTrackExtra, int, DauTrackExtras,
 DECLARE_SOA_INDEX_COLUMN_FULL(BachTrackExtra, bachTrackExtra, int, DauTrackExtras, "_BachExtra");          //!
 DECLARE_SOA_INDEX_COLUMN_FULL(StrangeTrackExtra, strangeTrackExtra, int, DauTrackExtras, "_StrangeExtra"); //!
 DECLARE_SOA_INDEX_COLUMN(StraCollision, straCollision);                                                    //!
-DECLARE_SOA_INDEX_COLUMN(MotherMCPart, motherMCPart);                                                          //!
+DECLARE_SOA_INDEX_COLUMN(MotherMCPart, motherMCPart);                                                      //!
 
 //______________________________________________________
 // REGULAR COLUMNS FOR CASCCORES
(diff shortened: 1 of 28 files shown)


dsekihat (dismissed):

ddobrigk:
Hi @dsekihat - I'm sorry for dismissing your approval. There was still one last bug introduced in the end that I had to solve (subscription didn't exactly match typecast, which can't be done without an ugly runtime crash)...

PWGLF/DataModel/LFStrangenessTables.h
```diff
@@ -145,6 +145,9 @@ DECLARE_SOA_COLUMN(DCANegToPV, dcanegtopv, float);         //! DCA negative pron
 DECLARE_SOA_COLUMN(V0CosPA, v0cosPA, float);               //! V0 CosPA
 DECLARE_SOA_COLUMN(DCAV0ToPV, dcav0topv, float);           //! DCA V0 to PV
 
+// Type of V0 from the svertexer (photon, regular, from cascade)
+DECLARE_SOA_COLUMN(V0Type, v0Type, uint8_t); //! DCA V0 to PV
```
f3sch, line 149: The description is not correct.

PWGLF/DataModel/LFStrangenessTables.h
```diff
@@ -325,10 +328,15 @@ DECLARE_SOA_DYNAMIC_COLUMN(PositiveEta, positiveeta, //! positive daughter eta
                            [](float PxPos, float PyPos, float PzPos) -> float { return RecoDecay::eta(std::array{PxPos, PyPos, PzPos}); });
 DECLARE_SOA_DYNAMIC_COLUMN(PositivePhi, positivephi, //! positive daughter phi
                            [](float PxPos, float PyPos) -> float { return RecoDecay::phi(PxPos, PyPos); });
+
+DECLARE_SOA_DYNAMIC_COLUMN(IsStandardV0, isStandardV0, //! is standard V0
+                           [](uint8_t V0Type) -> bool { return V0Type & (1 << 0); });
+DECLARE_SOA_DYNAMIC_COLUMN(IsPhotonV0, isPhotonV0, //! is photon V0
```
f3sch, line 334: I have been thinking about the naming here. This is extremely misleading actually because this flag is only set to true if the photo mass hypothesis _and_ if a TPConly track is one of the prongs. This could potentially lead analyzer to believe that using this flag that they get photons only but in fac they are _also only_ then getting tpconly prongs… a better name might be ‘isPhotonTPConly’ (or something else, I am lacking creativity here).

f3sch (commented):
Thanks for your works David! I have one comment and one suggestion.

PWGLF/DataModel/LFStrangenessTables.h
f3sch: I am not even sure if this will then propagate to the documentation.

ddobrigk: So the autogenerated doc problem is currently being investigated by @pbuehler (super thanks!) - there seem to be some sophisticated new preprocessor-defined tables in the framework that complicate the auto-generation. Once that's sorted out I think these comments do in fact go into the docs... I'll keep an eye :-D

ChiaraDeMartin95 (approved):

fgrosa (approved):

dsekihat (approved):

PR #4206, opened 2023-12-19
track to collision associator optimization

jgrosseo:


jgrosseo:
This improves the speed by a factor 6 for central barrel collisions. It may be more as now the slowest task is something else. The main optimization is still switched off for MFT. However, a factor 2 should also apply to MFT (without any loss of association).

Needs to be carefully checked for central barrel (@fgrosa) and MFT (@sarahherrmann). No need to merge before the holiday break. Happy holidays!

fgrosa:
Hi @jgrosseo I have checked the effect on the HF finding, and I got exactly the same result as the one obtained with the old version of the code running on a full `AO2D.root` (pp@500 kHz)

image

So from the central barrel point of view I would say that it is ok. Thanks a lot for the development!

jgrosseo:
Thanks @fgrosa for checking!

I have rebased to master and merged with the uint -> int change. I move it to review now. 
@sarahherrmann Do you have feedback?

sarahherrmann:
> Thanks @fgrosa for checking!
> 
> I have rebased to master and merged with the uint -> int change. I move it to review now. @sarahherrmann Do you have feedback?

Hello Jan Fiete, 
Thanks for this code development,
I did not have time to test it out so far, but I will try to run some tests tomorrow.

Cheers, 
Sarah

alibuild (approved):

PR #4207, opened 2023-12-19
[PWGDQ] Adding histograms and reduced table for MFT pid studies

lucamicheletti93:
This code processes and skims the updated AO2Ds containing the MFT cluster size, expliting the existing process function and adding an additional one for the option without event filtering.


alibuild (dismissed):

alibuild (approved):

PR #4214, opened 2023-12-20
PWGHF: Derived data table for the event mixing

Simple-Shyam:
The pull request includes the definition of the table of D+ and associated track properties. The identification of a single event can be done using globalBCID and timestamp. The correlator task will fill these tables and later these tables can be used for the event-mixing. In this case, we can mix D+ with the tracks with different events (in each pool bin already added). This will allow to have very large statistics even for rare heavy-flavor decays.


fgrosa:
Hi @Simple-Shyam, thanks a lot for starting this development. However, if I am not missing anything, your logic would require the access to the parent AO2D, since the BC index is defined within a DF of the parent AO2D. Moreover, it would be good to have uniform implementations for the different D mesons, so it would be good to follow the example of @scattaru for the Ds presented in the last HFC meeting (see https://indico.cern.ch/event/1303185/contributions/5719930/attachments/2774322/4834534/Ds-h_correlations_Derived_Data_Samuele_Cattaruzzi.pdf)

Simple-Shyam:
> Hi @Simple-Shyam, thanks a lot for starting this development. However, if I am not missing anything, your logic would require the access to the parent AO2D, since the BC index is defined within a DF of the parent AO2D. Moreover, it would be good to have uniform implementations for the different D mesons, so it would be good to follow the example of @scattaru for the Ds presented in the last HFC meeting (see https://indico.cern.ch/event/1303185/contributions/5719930/attachments/2774322/4834534/Ds-h_correlations_Derived_Data_Samuele_Cattaruzzi.pdf)

Hi @fgrosa, Thanks so I can't access this bcID and timestamp in the case of derived data produced for D-meson?
I will define the new header file for D+ derived table but the size will be larger than Ds in this case, I can reduce the variables to be filled in the tree. In this case, I can close it and create a new header file as done for Ds.

fcolamar:
Ciao @fgrosa, @Simple-Shyam,+in this case the underlying idea is a bit different w.r.t. Samuele is doing, i.e. is to run the same-event analysis fully online, starting from the general HF derived data, while for the mixed-event analysis store at the same time a fraction (to be optimised based on the sample) of D mesons and tracks in dedicated trees, with only the information needed for correlations (kinematics+bcID/timestamp), and then perform the mixing in a subsequent stage on the stored (and DF-merged, to increase the ME plot stat) tables. When building these tables there's still access to the original AODs, since the HF-derived data are linked, so the bcID/timestamp should still be available, right?

In general, a fear that I have is that for particles which have more selected candidates per event as the D0 or probably the D+, and especially in large collision systems with several candidates per event as PbPb (this probably also for the Ds), The Ds approach of storing tables for all tracks and candidates which contain a large list of variables could lead to too large derived data. For sure this would work for e.g. systematic studies, since there one can select a random subset of collisions to store, but for the full same-event analysis, that should exploit the full sample, it could be problematic, that's why this alternate approach was being developed (it can be of course that the bottleneck in this case is the single-event cpu running time, but we have to see).

What do you think?
Fabio

fgrosa:
Ciao @fcolamar and @Simple-Shyam,
> Ciao @fgrosa, @Simple-Shyam,+in this case the underlying idea is a bit different w.r.t. Samuele is doing, i.e. is to run the same-event analysis fully online, starting from the general HF derived data, while for the mixed-event analysis store at the same time a fraction (to be optimised based on the sample) of D mesons and tracks in dedicated trees, with only the information needed for correlations (kinematics+bcID/timestamp), and then perform the mixing in a subsequent stage on the stored (and DF-merged, to increase the ME plot stat) tables. When building these tables there's still access to the original AODs, since the HF-derived data are linked, so the bcID/timestamp should still be available, right?

The issue is not in the task that produces it, but when you analyse your derived AO2Ds. Since they are self contained (and not linked), the merger did not know how to rearrange the indices of the BCs and you can have two identical BC indices that come from different DFs of the parent AO2D (processed to produce your derived data), but they would infact correspond to 2 different BCs (but you have no way to know it later on). Everytime you store an index (in this case the BC index), you should also store the corresponding table, otherwise it cannot work.

> 
> In general, a fear that I have is that for particles which have more selected candidates per event as the D0 or probably the D+, and especially in large collision systems with several candidates per event as PbPb (this probably also for the Ds), The Ds approach of storing tables for all tracks and candidates which contain a large list of variables could lead to too large derived data. For sure this would work for e.g. systematic studies, since there one can select a random subset of collisions to store, but for the full same-event analysis, that should exploit the full sample, it could be problematic, that's why this alternate approach was being developed (it can be of course that the bottleneck in this case is the single-event cpu running time, but we have to see).
> 
This might indeed be the case in Pb-Pb (to be studied though), but for pp collisions I don't think it's the case (moreover, the Ds if not selected with reasonable tight selections has typically more candidates than D0 and D+, so I am not sure that the size is more of a concern for D+ and D0 rather than the Ds).
In my opinion, if producing these derived data with selections that are the loosest you would use in the analysis (which are anyway rather tight) you can reduce significantly the size (anyway these derived data can be redone multiple times). Moreover, it allows you to be the only one accessing the input data (because they are for your analysis) and you don't risk to have the analysis stuck because under conferences everybody needs to access the original AO2Ds (the current estimates show that it will not be practically possible to access all the original AO2Ds, to be of course evaluated).

> What do you think? Fabio

Simple-Shyam:
> Ciao @fcolamar and @Simple-Shyam,
> 
> > Ciao @fgrosa, @Simple-Shyam,+in this case the underlying idea is a bit different w.r.t. Samuele is doing, i.e. is to run the same-event analysis fully online, starting from the general HF derived data, while for the mixed-event analysis store at the same time a fraction (to be optimised based on the sample) of D mesons and tracks in dedicated trees, with only the information needed for correlations (kinematics+bcID/timestamp), and then perform the mixing in a subsequent stage on the stored (and DF-merged, to increase the ME plot stat) tables. When building these tables there's still access to the original AODs, since the HF-derived data are linked, so the bcID/timestamp should still be available, right?
> 
> The issue is not in the task that produces it, but when you analyse your derived AO2Ds. Since they are self contained (and not linked), the merger did not know how to rearrange the indices of the BCs and you can have two identical BC indices that come from different DFs of the parent AO2D (processed to produce your derived data), but they would infact correspond to 2 different BCs (but you have no way to know it later on). Everytime you store an index (in this case the BC index), you should also store the corresponding table, otherwise it cannot work.
> 
> > In general, a fear that I have is that for particles which have more selected candidates per event as the D0 or probably the D+, and especially in large collision systems with several candidates per event as PbPb (this probably also for the Ds), The Ds approach of storing tables for all tracks and candidates which contain a large list of variables could lead to too large derived data. For sure this would work for e.g. systematic studies, since there one can select a random subset of collisions to store, but for the full same-event analysis, that should exploit the full sample, it could be problematic, that's why this alternate approach was being developed (it can be of course that the bottleneck in this case is the single-event cpu running time, but we have to see).
> 
> This might indeed be the case in Pb-Pb (to be studied though), but for pp collisions I don't think it's the case (moreover, the Ds if not selected with reasonable tight selections has typically more candidates than D0 and D+, so I am not sure that the size is more of a concern for D+ and D0 rather than the Ds). In my opinion, if producing these derived data with selections that are the loosest you would use in the analysis (which are anyway rather tight) you can reduce significantly the size (anyway these derived data can be redone multiple times). Moreover, it allows you to be the only one accessing the input data (because they are for your analysis) and you don't risk to have the analysis stuck because under conferences everybody needs to access the original AO2Ds (the current estimates show that it will not be practically possible to access all the original AO2Ds, to be of course evaluated).
> 
> > What do you think? Fabio

Ciao @fcolamar and @fgrosa,

Thank you for clearing things up. Now should I close this pull request? Since I can't see the commit from Ds at the moment so I will wait for that then I can follow the same approach once it's on git.

fcolamar:
Ciao @fgrosa,
about the BC identification, this is the reason for having the timestamp information as well, which identifies univoquely the bunch crossing with a time information which is not reset at each DF (it's not an index). If I'm not missing anything, I think that the combination of the two numbers, i.e. globalindex of the collision within the BC (now really to be considered as a number)+ timestamp of the BC, should allow for univoquely identify a collision also on a self-contained data (similarly to the approach of BC+orbit+period used in run2 for some analyses).

About the second point, your comment on the access to the original AOD is absolutely reasonable. I would anyway probably try to keep both approaches (online, and full offline on self.-contained derived data) open for the moment, until we are fully sure of the feasibility of the offline on all systems). Also because apart from the Ds, all the other channels are currently based on the online approach and it will be not immediate to translate to the full offline.
This being said, for what concerns the ME (which is the specific point of this PR), we can discuss and wee if there's a way of defining the tables similarly to Samuele's case in order to have a similar struture, and then decide if to do everything offline or the ME only (on a subset of the data) based on furhter evaluation of the resources consumptions.

mmazzilli:
Hi all, 
sorry to jump in the discussion. This is just to confirm that indeed in an internal discussion it was agreed to proceed with the two approaches (full offline and online) for Ds and for D+/D0 for the reasons that @fcolamar reported. 
An accurate estimate of the resource consumption and of the table size can help in understanding whether it is possible to use the same approach for all D mesons..
However, for now I am in favour of continuing with these approaches (unless of a major issue with the BC identification as mentioned in the discussion).
Cheers, 
Marianna

fgrosa:
Ciao @fcolamar, 

> Ciao @fgrosa, about the BC identification, this is the reason for having the timestamp information as well, which identifies univoquely the bunch crossing with a time information which is not reset at each DF (it's not an index). If I'm not missing anything, I think that the combination of the two numbers, i.e. globalindex of the collision within the BC (now really to be considered as a number)+ timestamp of the BC, should allow for univoquely identify a collision also on a self-contained data (similarly to the approach of BC+orbit+period used in run2 for some analyses).
> 

Ok, of this I am not 100% sure, but it might well be that you are right. Thanks for the clarification! So the idea here is to group D and associated tracks in pool bins, combine them and just check they do not come from the same BC. 

> About the second point, your comment on the access to the original AOD is absolutely reasonable. I would anyway probably try to keep both approaches (online, and full offline on self.-contained derived data) open for the moment, until we are fully sure of the feasibility of the offline on all systems). Also because apart from the Ds, all the other channels are currently based on the online approach and it will be not immediate to translate to the full offline. This being said, for what concerns the ME (which is the specific point of this PR), we can discuss and wee if there's a way of defining the tables similarly to Samuele's case in order to have a similar struture, and then decide if to do everything offline or the ME only (on a subset of the data) based on furhter evaluation of the resources consumptions.

Yes, I think that having the two possibilities could be definitely be useful. My suggestion is not to abandon a priori the approach of Samuele, because it can have a lot of advantages (very small memory and CPU used, trains that run with autosubmission and take very short time to finish, unlike the trains that run on original AO2Ds).

fcolamar:
Ciao @fgrosa,

> Ciao @fcolamar,
> 
> > Ciao @fgrosa, about the BC identification, this is the reason for having the timestamp information as well, which identifies univoquely the bunch crossing with a time information which is not reset at each DF (it's not an index). If I'm not missing anything, I think that the combination of the two numbers, i.e. globalindex of the collision within the BC (now really to be considered as a number)+ timestamp of the BC, should allow for univoquely identify a collision also on a self-contained data (similarly to the approach of BC+orbit+period used in run2 for some analyses).
> 
> Ok, of this I am not 100% sure, but it might well be that you are right. Thanks for the clarification! So the idea here is to group D and associated tracks in pool bins, combine them and just check they do not come from the same BC.
Yes exactly as you say, though to be precise Shyam checks that D and track don't come from the same collision (via collision id within th BC + time identifier of the BC).
@Simple-Shyam checked this for sure locally on an input file and that approach was working fine, I don't actually remember if he also checked after merging the DF, I tend to remember so but not 100% sure, maybe he can confirm.

> > About the second point, your comment on the access to the original AOD is absolutely reasonable. I would anyway probably try to keep both approaches (online, and full offline on self.-contained derived data) open for the moment, until we are fully sure of the feasibility of the offline on all systems). Also because apart from the Ds, all the other channels are currently based on the online approach and it will be not immediate to translate to the full offline. This being said, for what concerns the ME (which is the specific point of this PR), we can discuss and wee if there's a way of defining the tables similarly to Samuele's case in order to have a similar struture, and then decide if to do everything offline or the ME only (on a subset of the data) based on furhter evaluation of the resources consumptions.
> 
> Yes, I think that having the two possibilities could definitely be useful. My suggestion is not to abandon a priori the approach of Samuele, because it can have a lot of advantages (very small memory and CPU used, trains that run with autosubmission and take very short time to finish, unlike the trains that run on original AO2Ds).
I fully agree with you on this point. It's surely more profitable to use his approach if the size of the produced files is under control, for all the reasons you mention, and I was also supporting this idea during the discussion we had last month, but since this is not 100% sure yet I'd suggest to proceed with the online approach + offline ME to have an alternate and backup solution.

Simple-Shyam:
> Ciao @fgrosa,
> 
> > Ciao @fcolamar,
> > > Ciao @fgrosa, about the BC identification, this is the reason for having the timestamp information as well, which identifies univoquely the bunch crossing with a time information which is not reset at each DF (it's not an index). If I'm not missing anything, I think that the combination of the two numbers, i.e. globalindex of the collision within the BC (now really to be considered as a number)+ timestamp of the BC, should allow for univoquely identify a collision also on a self-contained data (similarly to the approach of BC+orbit+period used in run2 for some analyses).
> > 
> > 
> > Ok, of this I am not 100% sure, but it might well be that you are right. Thanks for the clarification! So the idea here is to group D and associated tracks in pool bins, combine them and just check they do not come from the same BC.
> 
> Yes exactly as you say, though to be precise Shyam checks that D and track don't come not only from the same BC but also from the same collision in the BC (via collision id within th BC + time identifier of the BC). @Simple-Shyam checked this for sure locally on an input file and that approach was working fine, I don't actually remember if he also checked after merging the DF, I tend to remember so but not 100% sure, maybe he can confirm.
> 
> > > About the second point, your comment on the access to the original AOD is absolutely reasonable. I would anyway probably try to keep both approaches (online, and full offline on self.-contained derived data) open for the moment, until we are fully sure of the feasibility of the offline on all systems). Also because apart from the Ds, all the other channels are currently based on the online approach and it will be not immediate to translate to the full offline. This being said, for what concerns the ME (which is the specific point of this PR), we can discuss and wee if there's a way of defining the tables similarly to Samuele's case in order to have a similar struture, and then decide if to do everything offline or the ME only (on a subset of the data) based on furhter evaluation of the resources consumptions.
> > 
> > 
> > Yes, I think that having the two possibilities could definitely be useful. My suggestion is not to abandon a priori the approach of Samuele, because it can have a lot of advantages (very small memory and CPU used, trains that run with autosubmission and take very short time to finish, unlike the trains that run on original AO2Ds).
> 
> I fully agree with you on this point. It's surely more profitable to use his approach if the size of the produced files is under control, for all the reasons you mention, and I was also supporting this idea during the discussion we had last month, but since this is not 100% sure yet I'd suggest to proceed with the online approach + offline ME to have an alternate and backup solution.

Thanks Fabio, Fabrizio, and Marinna for the suggestion. I will check after merging the data frames as suggested by Fabio and will confirm the approach. We checked before running the AOD merger (with a few files each having many data frames).

fgrosa:
> > Ciao @fgrosa,
> > > Ciao @fcolamar,
> > > > Ciao @fgrosa, about the BC identification, this is the reason for having the timestamp information as well, which identifies univoquely the bunch crossing with a time information which is not reset at each DF (it's not an index). If I'm not missing anything, I think that the combination of the two numbers, i.e. globalindex of the collision within the BC (now really to be considered as a number)+ timestamp of the BC, should allow for univoquely identify a collision also on a self-contained data (similarly to the approach of BC+orbit+period used in run2 for some analyses).
> > > 
> > > 
> > > Ok, of this I am not 100% sure, but it might well be that you are right. Thanks for the clarification! So the idea here is to group D and associated tracks in pool bins, combine them and just check they do not come from the same BC.
> > 
> > 
> > Yes exactly as you say, though to be precise Shyam checks that D and track don't come not only from the same BC but also from the same collision in the BC (via collision id within th BC + time identifier of the BC). @Simple-Shyam checked this for sure locally on an input file and that approach was working fine, I don't actually remember if he also checked after merging the DF, I tend to remember so but not 100% sure, maybe he can confirm.
> > > > About the second point, your comment on the access to the original AOD is absolutely reasonable. I would anyway probably try to keep both approaches (online, and full offline on self.-contained derived data) open for the moment, until we are fully sure of the feasibility of the offline on all systems). Also because apart from the Ds, all the other channels are currently based on the online approach and it will be not immediate to translate to the full offline. This being said, for what concerns the ME (which is the specific point of this PR), we can discuss and wee if there's a way of defining the tables similarly to Samuele's case in order to have a similar struture, and then decide if to do everything offline or the ME only (on a subset of the data) based on furhter evaluation of the resources consumptions.
> > > 
> > > 
> > > Yes, I think that having the two possibilities could definitely be useful. My suggestion is not to abandon a priori the approach of Samuele, because it can have a lot of advantages (very small memory and CPU used, trains that run with autosubmission and take very short time to finish, unlike the trains that run on original AO2Ds).
> > 
> > 
> > I fully agree with you on this point. It's surely more profitable to use his approach if the size of the produced files is under control, for all the reasons you mention, and I was also supporting this idea during the discussion we had last month, but since this is not 100% sure yet I'd suggest to proceed with the online approach + offline ME to have an alternate and backup solution.
> 
> Thanks Fabio, Fabrizio, and Marinna for the suggestion. I will check after merging the data frames as suggested by Fabio and will confirm the approach. We checked before running the AOD merger (with a few files each having many data frames).

Thanks for the replies! Just one clarification: the merger was not expected to raise an error (actually is a warning) since the column in your table is not an index column, but not to be able to rearrange the BC indices (but ok now that comment does not hold anymore, because you clarified me what was the logic behind it).

Simple-Shyam:
> > > Ciao @fgrosa,
> > > > Ciao @fcolamar,
> > > > > Ciao @fgrosa, about the BC identification, this is the reason for having the timestamp information as well, which identifies univoquely the bunch crossing with a time information which is not reset at each DF (it's not an index). If I'm not missing anything, I think that the combination of the two numbers, i.e. globalindex of the collision within the BC (now really to be considered as a number)+ timestamp of the BC, should allow for univoquely identify a collision also on a self-contained data (similarly to the approach of BC+orbit+period used in run2 for some analyses).
> > > > 
> > > > 
> > > > Ok, of this I am not 100% sure, but it might well be that you are right. Thanks for the clarification! So the idea here is to group D and associated tracks in pool bins, combine them and just check they do not come from the same BC.
> > > 
> > > 
> > > Yes exactly as you say, though to be precise Shyam checks that D and track don't come not only from the same BC but also from the same collision in the BC (via collision id within th BC + time identifier of the BC). @Simple-Shyam checked this for sure locally on an input file and that approach was working fine, I don't actually remember if he also checked after merging the DF, I tend to remember so but not 100% sure, maybe he can confirm.
> > > > > About the second point, your comment on the access to the original AOD is absolutely reasonable. I would anyway probably try to keep both approaches (online, and full offline on self.-contained derived data) open for the moment, until we are fully sure of the feasibility of the offline on all systems). Also because apart from the Ds, all the other channels are currently based on the online approach and it will be not immediate to translate to the full offline. This being said, for what concerns the ME (which is the specific point of this PR), we can discuss and wee if there's a way of defining the tables similarly to Samuele's case in order to have a similar struture, and then decide if to do everything offline or the ME only (on a subset of the data) based on furhter evaluation of the resources consumptions.
> > > > 
> > > > 
> > > > Yes, I think that having the two possibilities could definitely be useful. My suggestion is not to abandon a priori the approach of Samuele, because it can have a lot of advantages (very small memory and CPU used, trains that run with autosubmission and take very short time to finish, unlike the trains that run on original AO2Ds).
> > > 
> > > 
> > > I fully agree with you on this point. It's surely more profitable to use his approach if the size of the produced files is under control, for all the reasons you mention, and I was also supporting this idea during the discussion we had last month, but since this is not 100% sure yet I'd suggest to proceed with the online approach + offline ME to have an alternate and backup solution.
> > 
> > 
> > Thanks Fabio, Fabrizio, and Marinna for the suggestion. I will check after merging the data frames as suggested by Fabio and will confirm the approach. We checked before running the AOD merger (with a few files each having many data frames).
> 
> Thanks for the replies! Just one clarification: the merger was not expected to raise an error (actually is a warning) since the column in your table is not an index column, but not to be able to rearrange the BC indices (but ok now that comment does not hold anymore, because you clarified me what was the logic behind it).

Ciao Fabio, Fabrizio, and Marianna,

Thank you. I stored in the table GlobalCollision Index (instead of BCId) as an int and the timestamp as int64_t and I further checked after merging the data frames the branches have the same numbers for the global collision ID before and after merging (AOD merger) as expected because it's stored as a separate branch. I committed the change on this PR.

fcolamar:
Hi Shyam,
thanks for the update (that was mandatory, since as discussed internally it's the combination GlobalCollision Index+timestamp that allows to univoquely identify the collision) and the related checks.
I'm approving this with the idea, previously discussed, of having for now the two possibilites open, offline-ME and full-offline approach.

fcolamar (approved):

PR #4215, opened 2023-12-20
PWGCF:  R2P2 correlations for Identified particles

R-Rohith:


PWGCF/TwoParticleCorrelations/Tasks/r2p2-4-id.cxx
```diff
@@ -0,0 +1,1175 @@
...
+    histos.add("h2d_npt_eta1Phi1Eta2Phi2PP", "np_T for +ve+ve particles", kTH2D, {etaphi1, etaphi2});
+    histos.add("h2d_npt_eta1Phi1Eta2Phi2MM", "np_T for -ve-ve particles", kTH2D, {etaphi1, etaphi2});
+    histos.add("h1d_n1_pt", "p_T", kTH1D, {{100, 0, 5, "p_T"}});
+    histos.add("h2d_n1_tpc1", "TPC", kTH2D, {{100, 0, 5, "p"}, {200, 0, 200, "#frac{dE}{dx}"}});
+    histos.add("h2d_n1_tpc2", "TPC", kTH2D, {{100, 0, 5, "p"}, {200, 0, 200, "#frac{dE}{dx}"}});
+    histos.add("h2d_n1_tof1", "TOF", kTH2D, {{100, 0, 5, "p"}, {50, 0.2, 1.2, "#beta"}});
+    histos.add("h2d_n1_tof2", "TOF", kTH2D, {{100, 0, 5, "p"}, {50, 0.2, 1.2, "#beta"}});
+    //------------------------------------------------------------------------
+  }
+  void process(soa::Filtered<soa::Join<aod::Collisions, aod::EvSels>>::iterator const& filteredCollisions, soa::Filtered<soa::Join<aod::Tracks, aod::TrackSelection, aod::pidTPCPi, aod::pidTOFPi, aod::pidTPCEl, aod::pidTOFbeta, aod::TracksExtra>> const& tracks)
```
victor-gonzalez, line 64: As you are subscribing to a single collision use a singular name for it, `filteredCollisions` is misleading

PWGCF/TwoParticleCorrelations/Tasks/r2p2-4-id.cxx
```diff
@@ -0,0 +1,1175 @@
...
+            histos.fill(HIST("h2d_npt_eta1Phi1Eta2Phi2PP"), 36 * etabin1 + phibin1 + 0.5, 36 * etabin2 + phibin2 + 0.5, track2.pt());
+            histos.fill(HIST("h2d_ptn_eta1Phi1Eta2Phi2PP"), 36 * etabin1 + phibin1 + 0.5, 36 * etabin2 + phibin2 + 0.5, track1.pt());
+            histos.fill(HIST("h2d_ptpt_eta1Phi1Eta2Phi2PP"), 36 * etabin1 + phibin1 + 0.5, 36 * etabin2 + phibin2 + 0.5, track1.pt() * track2.pt());
+          } else { //-ve particle pair
+            histos.fill(HIST("h2d_n2_eta1Phi1Eta2Phi2MM"), 36 * etabin1 + phibin1 + 0.5, 36 * etabin2 + phibin2 + 0.5);
+            histos.fill(HIST("h2d_npt_eta1Phi1Eta2Phi2MM"), 36 * etabin1 + phibin1 + 0.5, 36 * etabin2 + phibin2 + 0.5, track2.pt());
+            histos.fill(HIST("h2d_ptn_eta1Phi1Eta2Phi2MM"), 36 * etabin1 + phibin1 + 0.5, 36 * etabin2 + phibin2 + 0.5, track1.pt());
+            histos.fill(HIST("h2d_ptpt_eta1Phi1Eta2Phi2MM"), 36 * etabin1 + phibin1 + 0.5, 36 * etabin2 + phibin2 + 0.5, track1.pt() * track2.pt());
+          }
+        }
```
victor-gonzalez, line 117: This is extremely inefficient!! The task will be hardly penalized regarding CPU usage!
The checks on the bin acceptability should not be done within the double inner tracks loop
The cascaded `if`s also impose a huge execution penalty
I would suggest caching the histograms pointers in indexed variables according to the kind of track. In this routine there is only `P` and `M` then you can have a 2D 2x2 matrix where the (0,0), (0,1), (1,0), and (1,1) correspond to the MM, MP, PM, and PP combinations. In a previous loop over the tracks you check the proper binning and assign 0 to the track if it is negative an 1 if it is positive, -1 if it is out of proper binning. Then in the loop just use the assigned indexes for addressing the proper histogram checking only that the indexes are not negative

victor-gonzalez (changes requested):
Thanks for moving forward with this nice analysis

Please consider my comments regarding execution performance

Additionally, replicating code is extremely error prone
I would suggest to define generic templated `init` and `process` functions which are able to consider any kind of pair combination
Then for each task you will call the corresponding templated `init` and `process` according to the pair combination the task is addressing. In this way the code for `init` and `process` will only be written once

The performance issue I commented for the case of charged particles correlation functions affects of course to all the replicas (that is the problem of having code replicated; each time something is fixed, it has to be fixed in all replicas which is extremely error prone)

PWGCF/TwoParticleCorrelations/Tasks/r2p2-4-id.cxx
R-Rohith: So, a name like 'selectedCollision' ?

PWGCF/TwoParticleCorrelations/Tasks/r2p2-4-id.cxx
R-Rohith: Sir, I don't know a way to assign the index to a track, other than using a vector to store the indexes. Is there any other way to do it?

PWGCF/TwoParticleCorrelations/Tasks/r2p2-4-id.cxx
```diff
@@ -0,0 +1,437 @@
...
+      //-----------------------------------------------------------------------
+      etabin2 = (track2.eta() + 0.8) * 15;
+      phibin2 = 36 * track2.phi() / (2 * constants::math::PI);
+      if ((etabin2 < 0) || (etabin2 >= 24) || (phibin2 < 0) || (phibin2 >= 36))
+        continue;
+      sign = (track1.sign() + track2.sign() + 2) / 2;
+      h2d_2p[sign][0]->Fill(36 * etabin1 + phibin1 + 0.5, 36 * etabin2 + phibin2 + 0.5);
+      h2d_2p[sign][1]->Fill(36 * etabin1 + phibin1 + 0.5, 36 * etabin2 + phibin2 + 0.5, track2.pt());
+      h2d_2p[sign][2]->Fill(36 * etabin1 + phibin1 + 0.5, 36 * etabin2 + phibin2 + 0.5, track1.pt());
+      h2d_2p[sign][3]->Fill(36 * etabin1 + phibin1 + 0.5, 36 * etabin2 + phibin2 + 0.5, track1.pt() * track2.pt());
```
victor-gonzalez, line 215: I would suggest to have the array at the histogram level and not at the full histograms collection
With the current approach it is not easy to know which histogram is being filled which in case of issues is really difficult to debug

I meant something like
```
std::vector<std::vector<TH2F*>> h2d_n2_eta1Phi1Eta2Phi2;
```
which will be addressed as
```
h2d_n2_eta1Phi1Eta2Phi2[tag_track1][tag_track2]->Fill(...);
```

Please, take this as a suggestion which will make your code more readable and easy to debug in case of need
But it is not mandatory

PWGCF/TwoParticleCorrelations/Tasks/r2p2-4-id.cxx
```diff
@@ -0,0 +1,437 @@
...
+      sign2 = (track2.sign() + 1) / 2;
+      h2d_2p[sign1][sign2][0]->Fill(36 * etabin1 + phibin1 + 0.5, 36 * etabin2 + phibin2 + 0.5);
+      h2d_2p[sign1][sign2][1]->Fill(36 * etabin1 + phibin1 + 0.5, 36 * etabin2 + phibin2 + 0.5, track2.pt());
+      h2d_2p[sign1][sign2][2]->Fill(36 * etabin1 + phibin1 + 0.5, 36 * etabin2 + phibin2 + 0.5, track1.pt());
+      h2d_2p[sign1][sign2][3]->Fill(36 * etabin1 + phibin1 + 0.5, 36 * etabin2 + phibin2 + 0.5, track1.pt() * track2.pt());
+      //-----------------------------------------------------------------------
+    }
+  }
+  histos.fill(HIST("h1i_n1_multPM"), mult);
+}
```
victor-gonzalez, line 284: Please, consider that if the processing of the singles is considered in an independent function, most probably the processing of the pairs can be done without the need of replicating the like and unlike codes

PWGCF/TwoParticleCorrelations/Tasks/r2p2-4-id.cxx
```diff
@@ -0,0 +1,437 @@
...
+  return WorkflowSpec{
+    adaptAnalysisTask<r2p24ch>(cfgc),
+    adaptAnalysisTask<r2p24pi>(cfgc),
+    adaptAnalysisTask<r2p24ka>(cfgc),
+    adaptAnalysisTask<r2p24pr>(cfgc),
+    adaptAnalysisTask<r2p24pik>(cfgc),
+    adaptAnalysisTask<r2p24pip>(cfgc),
+    adaptAnalysisTask<r2p24pk>(cfgc),
+  };
+}
```
victor-gonzalez, line 437: Please, consider if with my previous comments this structure can be simplified 

With the current structure the processing per collision (basically going trough the double inner tracks loop) is repeated for each of the tasks that, in principle as they are defined, will be always executed in parallel. For the CPU consuming point of view this will definitely be penalized 

Consider the tagging of the tracks as described, which will need the tagging task, and then a single task which should incorporate the double inner tracks loop which, with the proper array of histograms, should fill all the needed histograms requiring only one double inner tracks loop

victor-gonzalez (changes requested):
Thanks for implementing my previous suggestions regarding not replicating code

I've added few additional suggestions which will improve the readability of the code, which definitely will make it less error prone and easier to debug it, and also which should improve the CPU usage impact

victor-gonzalez (changes requested):
Thanks for implementing my suggestions

Please consider the `namespace` renaming and the identification of tracks in the second inner loop

The other comments are suggestions which will help in the further development of the analysis

PWGCF/TwoParticleCorrelations/Tasks/r2p2-4-id.cxx
```diff
@@ -0,0 +1,444 @@
...
+//-----------------------------------------------------------------------------
+template <typename T>
+bool (*pidarray[4])(T) = {NO_PID, PID_PION, PID_KAON, PID_PROTON}; // Array of PID functions
+
+namespace o2::aod
+{
+namespace idr2p2columns
+{
+DECLARE_SOA_COLUMN(BinFlag, binflag, bool); // To flag tracks without proper binning
+DECLARE_SOA_COLUMN(PID, pid, int);          // Flag for type of particle
```
victor-gonzalez, line 94: Please, one single column. Use a negative value when not accepted
Also probably int is too large. See my below comments

PWGCF/TwoParticleCorrelations/Tasks/r2p2-4-id.cxx
```diff
@@ -0,0 +1,444 @@
...
+      phibin = 36 * track.phi() / (2 * constants::math::PI);
+      if ((etabin < 0) || (etabin >= 24) || (phibin < 0) || (phibin >= 36))
+        binflag = false;
+      else
+        binflag = true;
+      pid = 0;
+      for (int i = 0; i < 4; i++) {
+        if (pidarray<decltype(track)>[i](track))
+          pid = pid * 10 + i;
+      }
```
victor-gonzalez, line 115: What about a track that passes both the pion and the kaon cuts?

PWGCF/TwoParticleCorrelations/Tasks/r2p2-4-id.cxx
```diff
@@ -0,0 +1,444 @@
...
+
+            hist.h2d_2p[i][sign1][sign2][0]->Fill(36 * etabin1 + phibin1 + 0.5, 36 * etabin2 + phibin2 + 0.5);                            // h2d_n2_eta1Phi1Eta2Phi2*
+            hist.h2d_2p[i][sign1][sign2][1]->Fill(36 * etabin1 + phibin1 + 0.5, 36 * etabin2 + phibin2 + 0.5, track2.pt());               // h2d_npt_eta1Phi1Eta2Phi2*
+            hist.h2d_2p[i][sign1][sign2][2]->Fill(36 * etabin1 + phibin1 + 0.5, 36 * etabin2 + phibin2 + 0.5, track1.pt());               // h2d_ptn_eta1Phi1Eta2Phi2*
+            hist.h2d_2p[i][sign1][sign2][3]->Fill(36 * etabin1 + phibin1 + 0.5, 36 * etabin2 + phibin2 + 0.5, track1.pt() * track2.pt()); // h2d_ptpt_eta1Phi1Eta2Phi2*
+            if (f < 1)
+              break;
+          }
+          if (e < 1)
+            break;
```
victor-gonzalez, line 429: This continues being too much processing for being in the double track inner loop

Consider the following tagging/indexing
```
  kCharged, /* reco track NOT identified as pi, ka, nor pr */
  kPion,    /* reco identified as pi */
  kKaon,    /* reco identified as ka */
  kProton,  /* reco identified as pr */
```
when BOTH tracks ARE `kCharged` only the ch-ch associated histograms are filled
when BOTH trakcs ARE NOT `kCharged`, after filling the corresponding id1-id2 histograms, the ch-ch associated histograms are additionally filled

Does it make sense?

victor-gonzalez (changes requested):
Thanks for addressing my comments
The situation has improved quite a lot but in O2 we are forced to reduce to what is really needed the task memory consumption and the task CPU consumption
My comments try to address both so please have a look at them

PWGCF/TwoParticleCorrelations/Tasks/r2p2-4-id.cxx
R-Rohith: Then the PID column will store a value '12', indicating that the track has been identified as a both a Pion (1) and a Kaon (2).

PWGCF/TwoParticleCorrelations/Tasks/r2p2-4-id.cxx
R-Rohith: Yes I understand what you're saying. The reason why I had to use the double loop for filling the histograms is that, as I mentioned above, a track can be identified as more than one kind of particle with the current PID technique. Thus in one iteration of the inner track loop, I may have to fill several histograms of different combinations of particles. Suppose, as the worst case,  particle 1 and particle 2 are both identified to be pion, kaon, and proton, then in the inner track loop the histograms need to be filled for the 7 different combinations of particles. 
I had planned that after improving the PID, where a track is identified as only one kind of particle, I would drop the double loop and use an algorithm similar to what you suggested now.

PWGCF/TwoParticleCorrelations/Tasks/r2p2-4-id.cxx
R-Rohith: Ok, I'll change the datatype to int8_t

PWGCF/TwoParticleCorrelations/Tasks/r2p2-4-id.cxx
victor-gonzalez: But that is basically wrong 
One track can only be of one species. It is wrong to identifying it as two or more
The multiplicity cannot be artificially increased considering one track as multiple tracks

Of course, a pion is a charged track and a pion. But for this case have a look at my previous comment in the track loop

R-Rohith: Yes, I understand. I've done the necessary changes now.

victor-gonzalez (approved):
Thanks for addressing my previous comments

I approve it but I would  suggest trying to simplify things according to my comment

PR #4217, opened 2023-12-20
PWGCF: FemtoUniverse -- Deleting the workaround for Phi mesons in the MC Truth producer

zchochul:


alibuild (approved):

PR #4222, opened 2023-12-21
PWGCF: update FemtoDream

ariedel-cern:
- Add functionality to use masked mixing with MC for track-track and track-v0 task
- fix bug in pair cleaner
- Replace multiplicity signal with multiplicity percentile in producer task
- remove V0-only producer. This is done in anticipation of removing the reduced producer as well and make the standard producer more modular


victor-gonzalez (approved):
It would be good if at some point you can create a specific PR just for organizing the directory structure in the conventional `Core`, `DataModel`, `TableProducer`, `Tasks`, and perhaps `Utils` or `Tools` in your case

PR #4229, opened 2023-12-24
CF:identified particle

sweta29singh:
mean pT fluctuations in Identified particles at 5.36TeV


victor-gonzalez:
Dear Sweta, the tutorials are not open for modifications
Please, decide together with your PAG coordinators which functional area your analysis fits into and then we can decide where to put your analysis source files

PR #4230, opened 2023-12-27
Fixed crash when merging

EmilGorm:
- Fixed undefined behaviour when trying to merge unitialised FlowContainer
- Added FillQA and FillGFW methods to reduce code duplication
- Replaced uninitialised global variables in ProcessTrack scope


PWGCF/GenericFramework/Tasks/flowGenericFramework.cxx
```diff
@@ -390,58 +381,64 @@ struct GenericFramework {
...
-      fGFW->Fill(track.eta(), fPtAxis->FindBin(pt) - 1, phi, wacc * weff, 1);
+      fGFW->Fill(eta, fPtAxis->FindBin(pt) - 1, phi, w, 1);
     if (WithinPtPOI)
-      fGFW->Fill(track.eta(), fPtAxis->FindBin(pt) - 1, phi, wacc * weff, 2);
+      fGFW->Fill(eta, fPtAxis->FindBin(pt) - 1, phi, w, 2);
     if (WithinPtPOI && WithinPtRef)
-      fGFW->Fill(track.eta(), fPtAxis->FindBin(pt) - 1, phi, wacc * weff, 4);
+      fGFW->Fill(eta, fPtAxis->FindBin(pt) - 1, phi, w, 4);
+    return;
+  }
```
victor-gonzalez, line 428: For the next iteration consider templating on the the track/particle object instead
Both have the same getters and your code will be simpler and more efficient

PWGCF/GenericFramework/Tasks/flowGenericFramework.cxx
```diff
@@ -390,58 +381,64 @@ struct GenericFramework {
...
+  {
+    if (dt == kGen) {
+      QAregistry.fill(HIST("phi_gen"), phi);
+      QAregistry.fill(HIST("eta_gen"), eta);
+      QAregistry.fill(HIST("pt_gen"), pt);
+    } else {
+      QAregistry.fill(HIST("phi"), phi);
+      QAregistry.fill(HIST("eta"), eta);
+      QAregistry.fill(HIST("pt_dcaXY_dcaZ"), pt, dcaxy, dcaz);
+    }
```
victor-gonzalez, line 441: For the next iteration consider templating, additionally to the `datatype`) on two objects instead: `maintrack` and `assoctrack` of different `typename`s 
From `maintrack` you always extract `phi`, `eta`, and `pT` and from `assoctrack` you extract the `dcaxy` and `dcaz`
When you call from reconstructed data you call it being both the same
When you call from MC reconstructed you call it with `mcparticle` and with `track`
When you call from MC generated you call it with both the same `mcparticle` it should not complain because you will not use the `assoctrack` 

Something similar can be applied to `setCurrentParticleWeights` then you will avoid introducing the tables row information in the functions calls

victor-gonzalez (approved):

PR #4239, opened 2024-01-04
Improved TOF intercept calculation

ddobrigk:
@njacazio This PR exchanges the old radial-intercept-with-TOF calculation in the strangeness framework with a much improved calculation in which the 18-sector segmentation of the TOF is considered properly. The intercept is still calculated with an effective TOF radius, but this calculation should now be significantly more accurate than the previous one and uses an effective TOF radius of 377.64cm that has been extracted from the full geometrical model of the TOF. Further developments will still come with more extensive testing.


ddobrigk:
So I have to admit I have absolutely no idea how a PWGCF change came into this PR - it seems it was megalinter somehow, which is really very confusing. I will close this and re-open a clean PR with the same changes...

PR #4241, opened 2024-01-05
[PWGLF] Strangeness TOF PID updates

ddobrigk:
* improves TOF intercept calculation with an analytical segmented detector intercept (no energy losses considered but better geometry) 
* segments true raw info separately from actual pid for ease of exploring with derived data


ddobrigk:
Supersedes the ill-fated #4239

njacazio (approved):

PR #4247, opened 2024-01-08
DPG/AOTTrack: add task for efficiency evaluation with tag-and-probe with D mesons

fgrosa:
This PR adds a task to study the tracking efficiency with tag-and-probe using 3-prong decays of D mesons.
In particular, it works in two steps:
1) The first step finds 2-track vertices of pipi from D+, KK from phi originating from Ds or D+ decays and Kpi from D0 originating in D* decays. These 2-prong vertices are the "tag" and are supposed to have high purity (topological and PID cuts can be tuned)
In this step a table of "tags" containing 2-track indices and a collision index is filled.
2) The second step attaches a third track (probe) to the 2-prong vertices to complete the decay (D+ -> Kpipi, Ds/D+ -> phipi, D*+ -> D0pi). In this step, only track cuts are applied (currently global tracks, global tracks without ITS requirements and global tracks without TPC requirements are defined, but more can be added later if needed). Then, 2D histograms are filled with the mass of the D meson vs the pT of the third track. By evaluating the D-meson signals vs the probe track pT one can evaluate the efficiency of a given track-selection requirement (in this initial version the ITS-TPC matching is the focus).

It works properly, but to evaluate the actual performance a larger statistics than the one analysable locally is needed.

@mfaggin  @iouribelikov can you take a look? Thanks!
Tagging also @stefanopolitano @fcatalan92 @fchinu  @fprino @arossi81 who are involved in the development.


mfaggin (approved):
thank you @fgrosa !

PR #4255, opened 2024-01-10
PWG-CF Mean pt fluc update

TGahlaut1:


victor-gonzalez (changes requested):
Thanks for moving forward this analysis!!!

There are two main things that need to be done before we proceed with a bit more detailed review. They can be done in any order but both should be done as a first step.

- All format errors must be fixed. There are additional information on them following the `Details` link of each check and also on the source files under the `Files changed` tab
- The analysis source code has to be moved to the appropriate directory under `PWGCF`. For that, I would suggest to create a directory named `EbyEFluctuations` (you can check the appropriateness of the name with your PAG coordinators) under `PWGCF` and then replicate the `O2Physics` conventional directory structure, i.e.
```
PWGCF/EbyEFluctuations/Core
PWGCF/EbyEFluctuations/DataModel
PWGCF/EbyEFluctuations/TableProducer
PWGCF/EbyEFluctuations/Tasks
``` 
A model for the directory structure and the contents of the corresponding `CMakeList.txt` needed files can be found under `PWGCF/MultiParticleCorrelations`
The analysis source files have to be placed in this directory structure, as they are today under the `Tasks` subdirectory, and the overall `PWGCF/CMakeLists.txt` file needs to be modified to incorporate the new subdirectory

PR #4262, opened 2024-01-10
Added vn-pt and other changes

EmilGorm:
- Added vn-pt container
- Added class to handle pt-correlation algorithm
- Added TProfile derived class to handle bootstrapping

Generic framework analysis task
- The filter is apparently created before the configurables with some form of placeholder for single-valued configurables only - moved the pt min/max configurable to single-valued otherwise default value is not updated to the .json file input
- Use templated trackobject throughout most of the code
- Renamed datatype: now kReco and kGen only, where kReco is reconstructed data or mc events
-


victor-gonzalez (changes requested):
Please, consider my comments

victor-gonzalez (approved):

PR #4268, opened 2024-01-10
PWGEM: adjust TOF PID cut, momentum dependent

feisenhu:
Adjust PID selection for triggers, so it is momentum dependent in the pion rejection


alibuild (approved):

dsekihat (approved):

PR #4273, opened 2024-01-11
PWG-UD Enable different types of vetoes with FIT detectors:

pbuehler:
veto trigger available: TVX, TSC. TCE. TOR


alibuild (approved):

PR #4275, opened 2024-01-11
make multiplicityTable CCDB settings configurable

ddobrigk:
@strogolo here is the change you asked for. Sorry it took me a while. I had many local changes that I had to organise!


alibuild (approved):

PR #4278, opened 2024-01-11
PWGLF: Move V0 analysis to derived data format, add etas for cascs

ddobrigk:


ddobrigk:
@lhusova this adds the eta dynamics also for the cascades :-D + the V0 example task

ddobrigk:
With this PR I am adapting the old `lambdakzeroanalysis` example to actually use the new derived data format, as I surmise that particular task is not being used by anyone since the last meaningful update has been O(years) ago. In case there are any objections, please shout!

ddobrigk:
@lhusova now I added also the sel8 criteria to this PR in the end - I thought this was better than not saving the events as we may need that info for cross-checks. Maybe I should also rename the `lambdakzero` task to `derivedlambdakzeroanalysis` to match the syntax of the one for cascades ... but this I can decide later! Have a nice weekend, everyone

njacazio (approved):

PR #4285, opened 2024-01-11
Update tableMaker.cxx

bhsaho:


iarsene (dismissed):

iarsene:
@bhsaho, please fix the crashed tests.
Also, did you check whether you can use some of the other process functions? For example, the process function you propose will store all events, even those which do not have muons. What kind of skimmed set you want to implement with this ?

bhsaho:
> @bhsaho, please fix the crashed tests. Also, did you check whether you can use some of the other process functions? For example, the process function you propose will store all events, even those which do not have muons. What kind of skimmed set you want to implement with this ?

> @iarsene, Thank you for your reply. I am trying to fix the crashed test.  With the proposed process function, I want to acess the multiplicity table without any eventfilter

PR #4317, opened 2024-01-15
enable momentum-dependent Tsallis downscaling for V0 tree

jezwilkinson:
Patch to TPC PID track skimmer task to properly enable momentum-dependent downscaling via Tsallis function (method was already implemented and applied for TPC+TOF tree, but not properly enabled for V0 candidates)


jezwilkinson:
Hi @chiarazampolli @noferini , this is a small fix to the functionality on the PID skimmer task - could you please take a look + mark it for approval? Thanks

noferini:
Hi @jezwilkinson , it is fine with me.
Did you run a quick local test to validate it?

jezwilkinson:
> Hi @jezwilkinson , it is fine with me. Did you run a quick local test to validate it?

Thanks - Yes, I tested it locally on a few AO2Ds and it works as expected (left is before patch, right is after with same input/config) 
image

noferini (approved):
Fine with me

PR #4323, opened 2024-01-15
PWGCF: Create directory structure in FemtoDream

ariedel-cern:
- Create directory structure in FemtoDream
- Add configurable in producer to overwrite usage of multiplicity percentile


victor-gonzalez (approved):
Thanks for moving to the standard layout!!!

PR #4330, opened 2024-01-16
PWGLF: add table extenders for stra derived ana

ddobrigk:


ddobrigk:
@lhusova so, this adds two 'spawner' tasks whose purpose is to generate the extended table for analysis of derived data. This saves quite some space with respect to the previous approach since expression columns are then generated on-the-spot and can be filtered upon too. As a side remark, maybe we could keep simple utils like this in a separate directory to avoid having too many tasks (and especially such semi-automated ones) under `TableProducer`? Up to @njacazio @skundu692 - just let me know. Thanks!

ChiaraDeMartin95 (approved):

PR #4335, opened 2024-01-16
Introduce forward track propagation to UD

nburmaso:
Adding forward track propagation task. Tracks are propagated to primary vertex if available, to {0,0,0} otherwise. The output is track table `UDFwdTracksProp` and table of covariances `UDFwdTracksCovProp`, content is the same as for the standard tables, so one can easily switch to the UD tables containing propagated tracks.


pbuehler (approved):
Looks good to me.

PR #4336, opened 2024-01-16
PWGHF: Change in number of pT Bins and selection criteria for d0SoftPiNormalised

deependra170598:
1. Increased number of pT bins for selected Dstar and D0 candidates
2. Inequality condition for ```d0SoftPiNormalised``` is corrected.

### DIFF
--- a/PWGHF/TableProducer/candidateSelectorDstarToD0Pi.cxx
+++ b/PWGHF/TableProducer/candidateSelectorDstarToD0Pi.cxx
@@ -187,7 +187,7 @@ struct HfCandidateSelectorDstarToD0Pi {
     if (std::abs(candidate.impParamSoftPi()) > cutsDstar->get(binPt, "d0SoftPi")) {
       return false;
     }
-    if (std::abs(candidate.normalisedImpParamSoftPi()) > cutsDstar->get(binPt, "d0SoftPiNormalised")) {
+    if (std::abs(candidate.normalisedImpParamSoftPi()) < cutsDstar->get(binPt, "d0SoftPiNormalised")) {
       return false;
     }
     // selection on pT of soft Pi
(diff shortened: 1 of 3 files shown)


fgrosa (changes requested):
Hi @deependra170598 the changes are all fine with me, except for the one regarding the cut on the normalised impact parameter of the soft pion. Please check if the comment below makes sense to you.

PWGHF/TableProducer/candidateSelectorDstarToD0Pi.cxx
```diff
@@ -187,7 +187,7 @@ struct HfCandidateSelectorDstarToD0Pi {
     if (std::abs(candidate.impParamSoftPi()) > cutsDstar->get(binPt, "d0SoftPi")) {
       return false;
     }
-    if (std::abs(candidate.normalisedImpParamSoftPi()) > cutsDstar->get(binPt, "d0SoftPiNormalised")) {
+    if (std::abs(candidate.normalisedImpParamSoftPi()) < cutsDstar->get(binPt, "d0SoftPiNormalised")) {
```
fgrosa, line 190: This was correct as it was, since the soft pion is not displaced (it comes from the primary vertex, at least for prompt production). Hence, if the value is larger than a given value it should be rejected (i.e. it should return `false`)

deependra170598: OK. So what should be that threshold value?  In train run on hyperloop I used value 0.5.

fgrosa: You can put a very large value for the moment (i.e. 1000) so that it does not cut any candidate, and it can be then studied in the future

fgrosa (approved):

PR #4362, opened 2024-01-18
CF 2-prong candidate filter and tables

jaelpark:


PWGCF/DataModel/CorrelationsDerived.h
```diff
@@ -79,6 +79,39 @@ using CFTrack = CFTracks::iterator;
...
 using CFTrackWithLabel = CFTracksWithLabel::iterator;
+
+//------transient CF-filter to CF-2prong-filter
+DECLARE_SOA_TABLE(CFCollRefs, "AOD", "CFCOLLREF", o2::soa::Index<>, track::CollisionId); //! Transient cf collision index table
+
+using CFCollRef = CFCollRefs::iterator;
+
+namespace cftrackref
+{
+DECLARE_SOA_INDEX_COLUMN(Track, track);
```
jgrosseo, line 90: This one we cannot reuse from the data model itself?

PWGCF/DataModel/CorrelationsDerived.h
```diff
@@ -79,6 +79,39 @@ using CFTrack = CFTracks::iterator;
...
+DECLARE_SOA_TABLE(CFTrackRefs, "AOD", "CFTRACKREF", o2::soa::Index<>, track::CollisionId, cftrackref::TrackId); //! Transient cf track index table
+
+using CFTrackRef = CFTrackRefs::iterator;
+//------
+
+namespace cf2prongtrack
+{
+DECLARE_SOA_INDEX_COLUMN_FULL(CFTrackProng0, cfTrackProng0, int, CFTracks, "_0"); //! Index to prong 1 CFTrack
+DECLARE_SOA_INDEX_COLUMN_FULL(CFTrackProng1, cfTrackProng1, int, CFTracks, "_1"); //! Index to prong 2 CFTrack
+// type
```
jgrosseo, line 101: Can we remove this comment?

PWGCF/TableProducer/filter2Prong.cxx
```diff
@@ -0,0 +1,74 @@
...
+        }
+      }
+      for (auto& cftrack : cftracks) {
+        if (c.prong1Id() == cftrack.trackId()) {
+          prongCFId[1] = cftrack.globalIndex();
+          break;
+        }
+      }
+      // look-up the collision id
+      auto collisionId = cfcollisions.begin().globalIndex();
```
jgrosseo, line 59: Could cfcollisions have 0 length?

PWGCF/TableProducer/filter2Prong.cxx
```diff
@@ -0,0 +1,74 @@
...
+      uint8_t m = 0u;
+      if ((c.hfflag() & (1 << aod::hf_cand_2prong::DecayType::D0ToPiK)) != 0)
+        m |= aod::cf2prongtrack::kD0ToPiK;
+      output2ProngTracks(collisionId,
```
jgrosseo, line 63: maybe skip if m == 0 or do we want to write everything else?

jaelpark: I thought why not all since there are not too many of them per collision (only 1 to 3 or so)?

PWGCF/TableProducer/filter2Prong.cxx
jaelpark: That should always be 1. On line 40 I check already if tracks is 0 but I could check both there?

PWGCF/DataModel/CorrelationsDerived.h
jaelpark: There's `track::CollisionId` but what could be the corresponding `TrackId`? (trying to find one in "AO2D tables")

jgrosseo: You could use for example `o2::ambigous::TrackId` I think from AnalysisDataModel.h

jgrosseo: But if you find it more clean to redefine it, that is also OK for me

PWGCF/TableProducer/filter2Prong.cxx
jgrosseo: This could skip, couldn't it: https://github.com/AliceO2Group/O2Physics/blob/ac5086c65511ddc02ef6726bb1cd5d9fd824bb5a/PWGCF/TableProducer/filterCorrelations.cxx#L100

PWGCF/TableProducer/filter2Prong.cxx
jgrosseo: I think either define also masks for other 2 prongs, or do not write them.

PWGCF/DataModel/CorrelationsDerived.h
jaelpark: I saw the `ambiguous` but the description for it in the documentation implied something else so didn't think it was the right choice.

PWGCF/TableProducer/filter2Prong.cxx
jaelpark: Yes, although then it should also skip all the tracks. But I'm adding the collision size check too just in case.

PWGCF/TableProducer/filter2Prong.cxx
jaelpark: In the end I'm skipping m == 0 now. I've also changed the mask column into a "decay" column since there are many two prongs such as the strange particles and others that the code is generic enough to also handle in the future if needed. The selection in the correlation code is still done by masking though.

jgrosseo (approved):

PR #4369, opened 2024-01-18
add tree creator for ML lambda studies (data only)

maciacco:


njacazio:
Ciao Mario, can't you do this already with the current data model?

maciacco:
Ciao Nicolò, thanks for your suggestion. I understand the data model I proposed is similar to the one for v0 QA. I would like to add a slim data model to optimise machine learning selections for the antideuteron-lambda analysis, avoiding overlapping changes with other analyses.

njacazio:
No it's not the V0 QA that I meant, it's really the strangeness data model. As discussed in the PAG https://indico.cern.ch/event/1353511/contributions/5699259/attachments/2765705/4817424/DDChinellato-Strangeness-10.pdf  we're already producing derived data with this scheme therefore my question.

maciacco:
Thanks for the reference. Indeed we use the strangeness model in the analysis. However, to train and test the BDT we need to produce slim derived datasets with only the columns that are used as inputs of the BDT. The slim derived dataset will be extracted only from a small fraction of the statistics.

PR #4374, opened 2024-01-19
PWGCF: FemtoUniverse -- Change in MC Truth Producer (Phi doesn't need to be Physical Primary) and changes in the Producer (extended distription for one config and removing not neccesary if statement)

zchochul:


alibuild (approved):

PR #4375, opened 2024-01-19
PWGMM: dndeta: Consolidate response histograms to reduce output size

aalkin:
* fix wrong process naming


alibuild (approved):

PR #4385, opened 2024-01-19
 Added workflow of charged flattenicity

gbencedi:


aalkin:
@gbencedi sorry for the issues! These tests should be fixed now, but you need to update the PR to restart them. Any minor change would work.

gbencedi:
@aalkin thanks! It seems I also had to make some fixes. All should be fine now.

aalkin (approved):

PR #4410, opened 2024-01-22
PWGHF: adding a lite version of the LcpKpi tree

DelloStritto:


vkucera (approved):

mfaggin:
if I can comment @DelloStritto @vkucera for the time being maybe we can avoid completely to insert uncertainty variables, which are not trustable in data and MC and fill memory w/o being so useful for the moment. It would be enough just to comment-out columns and fillings related to them, to be restored asap when the track-uncertaintiesi in reconstruction will be better.

what do you think?

DelloStritto:
> if I can comment @DelloStritto @vkucera for the time being maybe we can avoid completely to insert uncertainty variables, which are not trustable in data and MC and fill memory w/o being so useful for the moment. It would be enough just to comment-out columns and fillings related to them, to be restored asap when the track-uncertaintiesi in reconstruction will be better.
> 
> What do you think?

Hi @mfaggin,
I would tend to agree with you. All these quantities at the moment are not useful at all, and removing them for the moment would make the output way lighter and easier to handle.

vkucera:
> if I can comment @DelloStritto @vkucera for the time being maybe we can avoid completely to insert uncertainty variables, which are not trustable in data and MC and fill memory w/o being so useful for the moment. It would be enough just to comment-out columns and fillings related to them, to be restored asap when the track-uncertaintiesi in reconstruction will be better.
> 
> What do you think?

Fine with me

vkucera (approved):

PR #4420, opened 2024-01-22
PWGLF: Fix autodetect in strangeness builders

rspijkers:
Look for Casc/V0 Indices tables instead of Datas tables

### DIFF
--- a/PWGLF/TableProducer/cascadebuilder.cxx
+++ b/PWGLF/TableProducer/cascadebuilder.cxx
@@ -513,11 +513,11 @@ struct cascadeBuilder {
         for (auto const& input : device.inputs) {
           if (device.name.compare("cascade-initializer") == 0)
             continue; // don't listen to the initializer, it's just to extend stuff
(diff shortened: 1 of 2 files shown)


ddobrigk:
Thanks a lot @rspijkers for catching this! However, I think the subscription that the autodetect should look for is the `V0Cores` and `[]CascCores` tables. It's entirely possible to analyse without subscribing to the `Indices` but not possible to analyse without subscribing to the `Cores`... Can you please adjust? Many many thanks!! :-D

rspijkers:
Hi @ddobrigk , done! Would it make sense to add some extra logic to the V0-builder autodetect in case one is looking at cascades? Instead of trying to match settings with the cascade-builder (which doesn't make much sense), one can look at the process that subscribes to the cascades and scan that for any V0 settings.

ddobrigk:
@rspijkers thank you very very much! :-D

njacazio (approved):
Thanks!

PR #4433, opened 2024-01-23
PWGEM/PhotonMeson: update pcb to reduce bkg

dsekihat:
update pcb to reduce bkg


alibuild (approved):

PR #4441, opened 2024-01-23
PWGHF: handle DCAFitter exception in LfCascades struct trackIndexSkimCreator

ZFederica:
Add a try-catch expression to handle DCAFitter exceptions in LfCascades struct trackIndexSkimCreator, add histogram to keep track of number of exceptions vs successful DCAFitter calls

### DIFF
--- a/PWGHF/TableProducer/trackIndexSkimCreator.cxx
+++ b/PWGHF/TableProducer/trackIndexSkimCreator.cxx
@@ -3550,12 +3550,12 @@ struct HfTrackIndexSkimCreatorLfCascades {
             nVtxFrom2ProngFitterXiHyp = df2.process(trackCascXi2Prong, trackParVarPion1);
           } catch (...) {
             LOGF(info, "Exception caught: failed to find charm baryon vertex (2prong - xi)");
-            if (fillHistograms){
+            if (fillHistograms) {
               registry.fill(HIST("hFitterStatusXi2Prong"), 1);
             }
             continue;
           }
-          if (fillHistograms){
+          if (fillHistograms) {
             registry.fill(HIST("hFitterStatusXi2Prong"), 0);
           }
 
@@ -3591,12 +3591,12 @@ struct HfTrackIndexSkimCreatorLfCascades {
             nVtxFrom2ProngFitterOmegaHyp = df2.process(trackCascOmega, trackParVarPion1);
           } catch (...) {
             LOGF(info, "Exception caught: failed to find charm baryon vertex (2prong - omega)");
-            if (fillHistograms){
+            if (fillHistograms) {
               registry.fill(HIST("hFitterStatusOmega2Prong"), 1);
             }
             continue;
           }
-          if (fillHistograms){
+          if (fillHistograms) {
             registry.fill(HIST("hFitterStatusOmega2Prong"), 0);
           }
 
@@ -3672,12 +3672,12 @@ struct HfTrackIndexSkimCreatorLfCascades {
                 nVtxFrom3ProngFitterXiHyp = df3.process(trackCascXi3Prong, trackParVarPion1, trackParVarPion2);
               } catch (...) {
                 LOGF(info, "Exception caught: failed to find charm baryon vertex (3prong - xi)");
-                if (fillHistograms){
+                if (fillHistograms) {
                   registry.fill(HIST("hFitterStatusXi3Prong"), 1);
                 }
                 continue;
               }
-              if (fillHistograms){
+              if (fillHistograms) {
                 registry.fill(HIST("hFitterStatusXi3Prong"), 0);
               }
 
--- a/PWGHF/TableProducer/trackIndexSkimCreator.cxx
+++ b/PWGHF/TableProducer/trackIndexSkimCreator.cxx
@@ -2189,7 +2189,15 @@ struct HfTrackIndexSkimCreator {
             is2ProngPreselected(pVecTrackPos1, pVecTrackNeg1, dcaInfoPos1[0], dcaInfoNeg1[0], cutStatus2Prong, whichHypo2Prong, isSelected2ProngCand);
 
             // secondary vertex reconstruction and further 2-prong selections
-            if (isSelected2ProngCand > 0 && df2.process(trackParVarPos1, trackParVarNeg1) > 0) { // should it be this or > 0 or are they equivalent
+            int nVtxFrom2ProngFitter = 0;
+            try {
+              nVtxFrom2ProngFitter = df2.process(trackParVarPos1, trackParVarNeg1);
+            } catch (...) {
+              LOGF(info, "Exception caught: failed to find vertex (2prong)");
+              continue;
+            }
+
+            if (isSelected2ProngCand > 0 && nVtxFrom2ProngFitter > 0) { // should it be this or > 0 or are they equivalent
               // get secondary vertex
               const auto& secondaryVertex2 = df2.getPCACandidate();
               // get track momenta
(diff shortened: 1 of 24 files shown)


PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -3509,7 +3514,16 @@ struct HfTrackIndexSkimCreatorLfCascades {
           auto trackParVarPion1 = getTrackParCov(trackPion1);
 
           // find charm baryon decay using xi PID hypothesis
-          int nVtxFrom2ProngFitterXiHyp = df2.process(trackCascXi2Prong, trackParVarPion1);
+          int nVtxFrom2ProngFitterXiHyp = 0;
+          try {
+            nVtxFrom2ProngFitterXiHyp = df2.process(trackCascXi2Prong, trackParVarPion1);
+          } catch (...) {
+            LOGF(info, "Exception caught: failed to find charm baryon vertex (2prong - xi)");
+            registry.fill(HIST("hFitterStatusXi2Prong"), 1);
```
fgrosa, line 3522: Shouldn't you check `if (fillHistograms)` before? Otherwise the histogram is not added to the registry and it should crash.

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -3509,7 +3514,16 @@ struct HfTrackIndexSkimCreatorLfCascades {
...
+            continue;
+          }
+          registry.fill(HIST("hFitterStatusXi2Prong"), 0);
```
fgrosa, line 3525: Same here

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -3537,7 +3551,16 @@ struct HfTrackIndexSkimCreatorLfCascades {
           }
 
           // find charm baryon decay using omega PID hypothesis
-          int nVtxFrom2ProngFitterOmegaHyp = df2.process(trackCascOmega, trackParVarPion1);
+          int nVtxFrom2ProngFitterOmegaHyp = 0;
+          try {
+            nVtxFrom2ProngFitterOmegaHyp = df2.process(trackCascOmega, trackParVarPion1);
+          } catch (...) {
+            LOGF(info, "Exception caught: failed to find charm baryon vertex (2prong - omega)");
+            registry.fill(HIST("hFitterStatusOmega2Prong"), 1);
```
fgrosa, line 3559: Same here

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -3537,7 +3551,16 @@ struct HfTrackIndexSkimCreatorLfCascades {
...
+            continue;
+          }
+          registry.fill(HIST("hFitterStatusOmega2Prong"), 0);
```
fgrosa, line 3562: Same here

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -3605,7 +3628,16 @@ struct HfTrackIndexSkimCreatorLfCascades {
               auto trackParVarPion2 = getTrackParCov(trackPion2);
 
               // reconstruct Xic with DCAFitter
-              int nVtxFrom3ProngFitterXiHyp = df3.process(trackCascXi3Prong, trackParVarPion1, trackParVarPion2);
+              int nVtxFrom3ProngFitterXiHyp = 0;
+              try {
+                nVtxFrom3ProngFitterXiHyp = df3.process(trackCascXi3Prong, trackParVarPion1, trackParVarPion2);
+              } catch (...) {
+                LOGF(info, "Exception caught: failed to find charm baryon vertex (3prong - xi)");
+                registry.fill(HIST("hFitterStatusXi3Prong"), 1);
```
fgrosa, line 3636: Same here

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -3605,7 +3628,16 @@ struct HfTrackIndexSkimCreatorLfCascades {
...
+                continue;
+              }
+              registry.fill(HIST("hFitterStatusXi3Prong"), 0);
```
fgrosa, line 3639: Same here

fgrosa (changes requested):
Hi @ZFederica thanks! Indeed this would be good also for the other channels without cascades. Would you like to take care about it or do you prefer me to do it?

ZFederica:
Hi @fgrosa, sure I can take care. Would you like to have the hisograms for the other channels too?

PWGHF/TableProducer/trackIndexSkimCreator.cxx
ZFederica: Yes, thanks, I overlooked it :)

fgrosa:
> Hi @fgrosa, sure I can take care. Would you like to have the hisograms for the other channels too?

I think we can skip them. Thanks!

vkucera:
Hi @ZFederica @fgrosa , before adding these exception catches into other places, I would us like agree on what we know about these exceptions and what we want to do about them. AFAIK, the LF cascades seem to be the only scenario where the DCA throws them. I am not familiar enough with the internals of the DCA fitter but my naive expectation is that if the fitting just fails, it should return 0. If it throws exceptions, to me it seems to indicate a problem which should be understood and fixed.
At https://github.com/AliceO2Group/O2Physics/blob/master/PWGHF/TableProducer/candidateCreatorToXiPi.cxx#L201
the exceptions are reported with errors messages whereas this PR reports them at info level. So my questions is: is this PR adding more debugging to fix the problem? Otherwise, I don't see the point of adding more info messages in the log which will not lead to any action.

ZFederica:
Hi @vkucera, sorry I pushed the changes before seeing your comment. I can of course change/remove the log info messages if needed. 
As for the exceptions, I think they come from a problem with the covariance matrix (see https://alimonitor.cern.ch/jobs/stackAnalysis.jsp?hy=282883,282884,282885,282886,282887,282888,282890,282891,282892,282893,282894,282895,282896). I am not aware of similar errors in the 2 and 3prong fitter because I never worked with those structs. A similar implementation is also used in the LF builders when building V0s (so when using "normal" tracks and not objects built with combinatorics), see https://github.com/AliceO2Group/O2Physics/blob/f2a1f35e5d6206b81a580aa06a9c5bb4da2f4d08/PWGLF/TableProducer/lambdakzerobuilder.cxx#L620-L629

vkucera:
@ZFederica , I'm not asking you to remove the messages, rather I'm trying to understand what you plan to do with them. Are you going to use them to debug the problem or report it in an O2 meeting?

ZFederica:
@vkucera it's just for completeness (for instance if one wishes to check in which file(s) this "problem" occurs it's possible to see it in the log file), as for now I do not personally have any specific plan to use this messages

fgrosa:
> @vkucera it's just for completeness (for instance if one wishes to check in which file(s) this "problem" occurs it's possible to see it in the log file), as for now I do not personally have any specific plan to use this messages

Hi @ZFederica @vkucera indeed this is a good point, I would rather use `LOGF(debug)` than `LOGF(info)` to avoid possible very large log files, if for some reason it fails frequently

fgrosa:
> > @vkucera it's just for completeness (for instance if one wishes to check in which file(s) this "problem" occurs it's possible to see it in the log file), as for now I do not personally have any specific plan to use this messages
> 
> Hi @ZFederica @vkucera indeed this is a good point, I would rather use `LOGF(debug)` than `LOGF(info)` to avoid possible very large log files, if for some reason it fails frequently (or to remove them completely, as suggested by @vkucera )

ZFederica:
@fgrosa @vkucera in the end I just removed the log messages

fgrosa (approved):

PR #4452, opened 2024-01-24
PWGHF: update saved variables in toXiPi workflow

ZFederica:
Update saved variables in toXiPi workflow:
- remove chi2AtPCACandidate in candidateCreator and treeCreator
- add PV chi2 in treeCreator


PWGHF/TableProducer/treeCreatorToXiPi.cxx
```diff
@@ -260,6 +263,9 @@ struct HfTreeCreatorToXiPi {
       candidate.decLenV0(),
       candidate.errorDecayLengthCharmBaryon(),
       candidate.errorDecayLengthXYCharmBaryon(),
+      candidate.impactParCascXY() / candidate.errImpactParCascXY(),
+      candidate.impactParPiFromCharmBaryonXY() / candidate.errImpactParPiFromCharmBaryonXY(),
+      candidate.decLenCharmBaryon() / candidate.errorDecayLengthCharmBaryon(),
```
fcatalan92, line 268: There are no dynamic columns providing directly the normalised values?

fcatalan92 (approved):
Hi @ZFederica, I have only a small question. In any case, the PR is fine.

PWGHF/TableProducer/treeCreatorToXiPi.cxx
ZFederica: Hi @fcatalan92, in the tree creators I never saw dynamic column being used. I could use such columns for the candidate creator table, but there I do not store these variables yet (I do not use them neither in the candidateCreator nor in the candidateSelector). This was my idea, is it fine?

fcatalan92: Hi Federica, I'm referring to the part where you fill the table. The point is that you can define the operations you are doing here as dynamic columns of the `HfCandToXiPi` table, and then call `candidate.normDecLenXXX()`.
BTW, looking at the table produced by the candidate creator, https://github.com/AliceO2Group/O2Physics/blob/5e5d04beba8ea82e35ba673b9fdf039acfa9870b/PWGHF/DataModel/CandidateReconstructionTables.h#L1087 I see you do not use dynamic or expression column at all. Maybe you could revise it in the future.

PR #4454, opened 2024-01-24
PWGHF: flags to fill only with signal or background in Lc tree creator (lite).

mfaggin:


PWGHF/TableProducer/treeCreatorLcToPKPi.cxx
```diff
@@ -305,14 +307,15 @@ struct HfTreeCreatorLcToPKPi {
...
+      bool isMcCandidateSignal = std::abs(candidate.flagMcMatchRec()) == (1 << o2::aod::hf_cand_3prong::DecayType::LcToPKPi);
       auto fillTable = [&](int CandFlag,
                            int FunctionSelection,
                            float FunctionInvMass,
                            float FunctionCt,
                            float FunctionY,
                            float FunctionE) {
         double pseudoRndm = trackPos1.pt() * 1000. - (int64_t)(trackPos1.pt() * 1000);
-        if (FunctionSelection >= 1 && pseudoRndm < downSampleBkgFactor) {
+        if (FunctionSelection >= 1 && (/*keep all*/ (!keepOnlySignalMc && !keepOnlyBkg) || /*keep only signal*/ (keepOnlySignalMc && isMcCandidateSignal) || /*keep only background and downsample it*/ (keepOnlyBkg && !isMcCandidateSignal && pseudoRndm < downSampleBkgFactor))) {
```
fgrosa, line 318: For the other channels there is also the possibility to set a maximum pT for the downsampling (in order to keep all the candidates at high pT, where the statistics is always a bit hungry). See e.g. https://github.com/AliceO2Group/O2Physics/blob/9358ec7f4700cd7f753bdb7d5f09cdcb7e8b756c/PWGHF/TableProducer/treeCreatorDplusToPiKPi.cxx#L214. Maybe you could include it as well (same for the data below, for which is even more important maybe)

fgrosa (changes requested):

PWGHF/TableProducer/treeCreatorLcToPKPi.cxx
mfaggin: thanks @fgrosa , done in the last commit

fgrosa (approved):

PR #4471, opened 2024-01-25
PWGHF: allow the application of PID to the bachelor pion regardless the D daughters

fgrosa:
@AlexBigO can you have a look? Thanks!

### DIFF
--- a/PWGHF/D2H/TableProducer/candidateSelectorB0ToDPiReduced.cxx
+++ b/PWGHF/D2H/TableProducer/candidateSelectorB0ToDPiReduced.cxx
@@ -37,7 +37,7 @@ struct HfCandidateSelectorB0ToDPiReduced {
   Configurable<float> ptCandMin{"ptCandMin", 0., "Lower bound of candidate pT"};
   Configurable<float> ptCandMax{"ptCandMax", 50., "Upper bound of candidate pT"};
   // Enable PID
-  Configurable<bool> usePid{"usePid", true, "Switch for PID selection at track level"};
+  Configurable<int> usePionPid{"usePionPid", 1, "Switch for PID selection for the bachelor pion (0: none, 1: TPC or TOF, 2: TPC and TOF)"};
   Configurable<bool> acceptPIDNotApplicable{"acceptPIDNotApplicable", true, "Switch to accept Status::NotApplicable [(NotApplicable for one detector) and (NotApplicable or Conditional for the other)] in PID selection"};
   // TPC PID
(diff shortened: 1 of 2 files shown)


PWGHF/D2H/TableProducer/candidateSelectorB0ToDPiReduced.cxx
```diff
@@ -163,15 +156,15 @@ struct HfCandidateSelectorB0ToDPiReduced {
...
-      if (!selectionFlagDAndUsePidInSync) {
-        hfSelB0ToDPiCandidate(statusB0ToDPi);
-        continue;
-      }
       // track-level PID selection
-      if (usePid) {
+      if (usePionPid) {
         auto trackPi = hfCandB0.template prong1_as<TracksPion>();
-        int pidTrackPi = selectorPion.statusTpcAndTof(trackPi);
+        int pidTrackPi{TrackSelectorPID::Status::NotApplicable};
```
AlexBigO, line 162: Hi @fgrosa ! Everything looks okay, thanks :)
Maybe we could avoid an `if` condition by doing this
```
int pidTrackPi = selectorPion.statusTpcAndTof(trackPi);
if (usePionPid == 2) {
  pidTrackPi = selectorPion.statusTpcAndTof(trackPi);
}
```
as, if the condition `usePionPid` is fulfilled, it means that it can only be `1` or `2` (the case `=1` would be our default condition).

fgrosa: Hi @AlexBigO thanks for the feedback! I still have a preference to keep it as it is (if you prefer I can put `if else` instead of `if else if`) because with your suggestion, if `usePionPid == 2` then we compute the status twice instead of once (it is not a simple return, see https://github.com/AliceO2Group/O2Physics/blob/52f1faf340047b3abbf92f59adc23af04d530298/Common/Core/TrackSelectorPID.h#L390-L442)

AlexBigO: Indeed, my option is counter-productive :) Yes, then going for `if then` is fine, thanks!

fgrosa: Done in https://github.com/AliceO2Group/O2Physics/pull/4471/commits/723376e81657e64c6935dda73e2b6e9c76dce9c0

alibuild (approved):

PR #4473, opened 2024-01-25
PWGMM: dndeta: fix event efficiency histogram

aalkin:
For the case where gen-level centrality is not available, we fall back to reco-level, so the efficiency histogram should only be filled after `c_gen` is updated.


alibuild (approved):

PR #4479, opened 2024-01-26
Luminosity stability analysis.

JosueMtzGar:
Luminosity stability analysis made by PWGMM/Luminosity group.


PWGMM/Lumi/Tasks/CMakeLists.txt
```diff
@@ -34,3 +34,12 @@ o2physics_add_dpl_workflow(fitvdm
         O2::DetectorsCommonDataFormats
         O2::DetectorsVertexing
         COMPONENT_NAME Analysis)
+
+o2physics_add_dpl_workflow(lumistab
+        SOURCES lumiStability.cxx
+        PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::AnalysisCore
+        O2::ReconstructionDataFormats
```
aalkin, line 41: There is no need to link to libraries that are not used, please remove the extras. If or when you add code to this task that requires these libraries, you can add them.

PWGMM/Lumi/Tasks/lumiStability.cxx
```diff
@@ -0,0 +1,105 @@
...
+
+      std::bitset<8> fT0Triggers = ft0.triggerMask();
+      bool vertex = fT0Triggers[o2::fdd::Triggers::bitVertex];
+
+      if(vertex){
+        histos.fill(HIST("BCFT0"), localBC);
+        NTrigVertexFT0++;
+      } // vertex true
+    } // loop over FT0 events
+    rowtrigvertex(NTrigVertexFDD,NTrigVertexFT0);
```
aalkin, line 95: The `process()` is called by framework once for a dataframe, that usually contains many timeframes. So the table would have a single entry for a dataframe, is this your intent?

PWGMM/Lumi/Tasks/CMakeLists.txt
JosueMtzGar: Alright, I will do the changes

PWGMM/Lumi/Tasks/lumiStability.cxx
JosueMtzGar: I declared this table and I saved the data to just have a comparison with what I get in the histograms. The intention of this is that it have to store the same information as histograms. Actually it isn't necessary to the analysis. I could remove it.

aalkin: How you want your output is up to you, if you prefer it as a ROOT TTree then table is the way to go. Note, however, that for running on hyperloop you will have to declare it as a (slim) derived data and there are restrictions on the output size.

JosueMtzGar: So I prefer the histograms for now. I will do the changes

JosueMtzGar:
Dear reviewers, I have done the corresponding changes.

PWGMM/Lumi/Tasks/lumiStability.cxx
```diff
@@ -27,79 +27,58 @@ using BCsWithTimestamps = soa::Join<aod::BCs, aod::Timestamps>;
...
   void process(aod::FT0s const& ft0s, aod::FDDs const& fdds, aod::BCsWithTimestamps const&)
   {
-    int NTrigVertexFDD = 0;
-    int NTrigVertexFT0 = 0;
-
-    for( auto const& fdd: fdds){
+    for (auto const& fdd : fdds) {
       auto bc = fdd.bc_as<BCsWithTimestamps>();
-      if(!bc.timestamp()) continue;
+      if (!bc.timestamp())
```
aalkin, line 48: Two minor comments, first is that the coding guidelines require curly braces even around one-line branches, so here and in line 65 the code should be
```cpp
if (bc.timestamp() == 0)
{
   continue;
}
```
and second is that it is always better to make the comparison explicit, instead on relying on implicit conversion to bool.  The latter could lead to incorrect code logic.

aalkin (approved):
I'm approving this to start the tests, however, please consider implementing my comments.

PR #4481, opened 2024-01-26
PWGHF: add sign of soft pion in D* table to avoid to access the track table

fgrosa:
@deependra170598 I made a small change in the D* table adding the soft pion sign, to avoid to access the track table later on to retrieve this information.


alibuild (approved):

PR #4492, opened 2024-01-27
PWGLF: extra changes for data volume reduction

ddobrigk:


ddobrigk:
@lhusova this + the previous commit in which I added an optional daughter eta selection at the derived data generation stage should reduce our data volume (and hence improve also our speed) further for analysis

njacazio (approved):

PR #4425, opened 2024-01-22
First version of TrackTuner for dca (and pt) smearing in MC

mfaggin:
Service to smear track parameters based on the dcaXY, dcaZ comparison in data and MC (in Run 2 called "improver-task", mostly used in HF).

The smearing is done on the `y` track parameter at the MC particle production point. The smearing is based on resolution, mean and pull ratio of dcaXY, dcaZ w.r.t. primary vertex measured in data and MC. These parameters are estimated vs. global-track pt and stored into an input `.root` file, which can be read from CCDB.

The `TrackTuner` can be configured via `Configurable<std::string> trackTunerParams`  and it is enabled via `Configurable<bool> useTrackTuner`, both in the `TrackPropagation` workflow.

The `TrackTuner` allows also to smear the q/pt. at the moment, this is done just via a fixed scaling (strategy to be decided afterwards).

Example of `TrackTuner` configuration via the `TrackPropagation` workflow:

```
"useTrackTuner": "true",
"trackTunerParams": "debugInfo=0|updateTrackCovMat=1|updateCurvature=0|updatePulls=1|isInputFileFromCCDB=1|pathInputFile=Users/m/mfaggin/test/inputsTrackTuner/PbPb2022|nameInputFile=trackTuner_DataLHC22sPass5_McLHC22l1b2_run529397.root|usePvRefitCorrections=0|oneOverPtCurrent=0|oneOverPtUpgrded=0"
```
 Credits: H. Sharma (@phymanshu  main developer), A. Rossi (@arossi81), M. Faggin (@mfaggin)
 Tagging also PWGHF convener @fgrosa @fcolamar and PWGHF PAG coordinators @fcatalan92 @mmazzilli @vkucera @nzardosh @deepathoms @NicoleBastid

PR #4496, opened 2024-01-29
Restore processCovariance name.

mfaggin:
Restoring the original `processCovariance` name, unproperly changed into `processCovarianceData` in the PR #4425 .

- @benedikt-voelkel this should solve the issues you observed in the ALIBI-NIGHTLIES
- @jgrosseo this would imply to revert back the configurations of Hyperloop service wagons on tomorrow

Sorry to both for the inconvenience, we overlooked this detail

### DIFF
--- a/Common/TableProducer/trackPropagation.cxx
+++ b/Common/TableProducer/trackPropagation.cxx
@@ -77,8 +77,8 @@ struct TrackPropagation {
       nEnabledProcesses++;
     }
 
-    if (doprocessCovarianceData) {
-      LOG(info) << "Enabling processCovarianceData";
+    if (doprocessCovariance) {
+      LOG(info) << "Enabling processCovariance";
       nEnabledProcesses++;
     }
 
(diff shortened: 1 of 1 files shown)


jgrosseo (changes requested):
Not sure this is the best now. We adjusted Hyperloop already. I think alibi should also be adjusted.

mfaggin:
> Not sure this is the best now. We adjusted Hyperloop already. I think alibi should also be adjusted.

actually `processCovariance` should keep working in both data and MC, and the name `processCovarianceData` might be misleading. I tend to prefer restoring the previous name, and afaik @benedikt-voelkel agrees as well

benedikt-voelkel:
I don't have a very strong opinion and was not aware that Hyperloop was adjusted already.

Just for clarity: Yes, analyses failed on alibi. But since it relies on O2DPG, it is in fact O2DPG that would need the updates.
So yes, we could change O2DPG.

jgrosseo:
This is part of O2DPG for the analysis QC?

jgrosseo:
OK, then let's revert this and adjust Hyperloop again tomorrow morning... :-(

jgrosseo (approved):

PR #4499, opened 2024-01-29
PWGLF: optional proper lifetime at builder time, def off

ddobrigk:


ddobrigk:
not merging megalinter since that again unfortunately touches files beyond this PR (which is probably a bug in the autogenerated megalinter PR)

njacazio (approved):

PR #4502, opened 2024-01-29
Common: protect against underflow

ddobrigk:
@sigurdnese @strogolo


alibuild (approved):

PR #4505, opened 2024-01-29
PWGDQ: Changes fro DQ correlation

shreyasiacharya:


iarsene (changes requested):
it has to be clarified why a new workflow copying most of table-reader is needed

PWGDQ/Core/VarManager.h
```diff
@@ -2510,6 +2510,15 @@ void VarManager::FillDileptonHadron(T1 const& dilepton, T2 const& hadron, float*
...
   }
+  if (fgUsedVars[kMass]) {
+    values[kMass] = dilepton.mass();
+  }
+  if (fgUsedVars[kPt1]) {
+    values[kPt1] = dilepton.pt();
+  }
+  if (fgUsedVars[kPt2]) {
+    values[kPt2] = hadron.pt();
+  }
```
iarsene, line 2521: These variables are filled already in this function. They have different names, but exist just in the same function. Also, the kMass, kPt1 and kPt2 are used for very different purposes and there is a potential for confusion when filling histogram classes.

iarsene (approved):

PR #4510, opened 2024-01-29
PWGHF: Passing pool bin definitions from the json file 

Simple-Shyam:
The pool bin definitions for the event mixing are hardcoded. Therefore, I added the option for passing the pool bin definition from the JSON file. We can define the multiplicity binning by looking at the multiplicity histogram saved in the correlator task.


PWGHF/HFC/TableProducer/correlatorDplusHadrons.cxx
```diff
@@ -170,6 +166,9 @@ struct HfCorrelatorDplusHadrons {
   Configurable<float> multMax{"multMax", 10000., "maximum multiplicity accepted"};
   Configurable<std::vector<double>> binsPt{"binsPt", std::vector<double>{o2::analysis::hf_cuts_dplus_to_pi_k_pi::vecBinsPt}, "pT bin limits for candidate mass plots and efficiency"};
   Configurable<std::vector<double>> efficiencyD{"efficiencyD", std::vector<double>{efficiencyDmeson}, "Efficiency values for Dplus meson"};
+  ConfigurableAxis confMultBins{"confMultBins", {VARIABLE_WIDTH, 0.0f, 2000.0f, 6000.0f, 100000.0f}, "Mixing bins - multiplicity"};
+  ConfigurableAxis confVtxBins{"confVtxBins", {VARIABLE_WIDTH, -10.0f, -2.5f, 2.5f, 10.0f}, "Mixing bins - z-vertex"};
+  ConfigurableAxis confMultBinsMcGen{"confMultBinsMcGen", {VARIABLE_WIDTH, 0.0f, 20.0f, 50.0f, 500.0f}, "Mixing bins - multiplicity"}; // In MCGen multiplicity is defined by counting tracks
```
vkucera, line 171:
```suggestion
  ConfigurableAxis binsMultiplicity{"binsMultiplicity", {VARIABLE_WIDTH, 0.0f, 2000.0f, 6000.0f, 100000.0f}, "Mixing bins - multiplicity"};
  ConfigurableAxis binsZVtx{"binsZVtx", {VARIABLE_WIDTH, -10.0f, -2.5f, 2.5f, 10.0f}, "Mixing bins - z-vertex"};
  ConfigurableAxis binsMultiplicityMc{"binsMultiplicityMc", {VARIABLE_WIDTH, 0.0f, 20.0f, 50.0f, 500.0f}, "Mixing bins - MC multiplicity"}; // In MCGen multiplicity is defined by counting tracks
```

vkucera (changes requested):
Please rename the helper task `HfDplusSelection` to `HfCorrelatorDplusHadronsDplusSelection` according to the HF naming scheme.

PWGHF/HFC/TableProducer/correlatorDplusHadrons.cxx
Simple-Shyam: Yes changed according to the suggestion.

PWGHF/HFC/TableProducer/correlatorDplusHadrons.cxx
```diff
@@ -242,6 +241,7 @@ struct HfCorrelatorDplusHadrons {
                    aod::TracksWDca const& tracks,
                    soa::Join<aod::HfCand3Prong, aod::HfSelDplusToPiKPi> const& candidates, aod::BCsWithTimestamps const&)
   {
+    BinningType corrBinning{{binsZVtx, binsMultiplicity}, true};
```
vkucera, line 244: Can you make this a member of the struct and initialise it only once in `init`?

Simple-Shyam: Hi @vkucera I was thinking this option before and tried also. When I initilaize there then it fails compilation also tried now. Do you have any example for this ? 
[compile.log](https://github.com/AliceO2Group/O2Physics/files/14097320/compile.log)

Simple-Shyam:
```
BinningType corrBinning; // declared in stuct
corrBinning{{binsZVtx, binsMultiplicity}, true}; // initialized in the init function
```
Again getting the compilation error. May be definition in Init needs to be changed?

Simple-Shyam: Thanks, I need to understand looking at the example, I tried a few ways but not working at the moment.

vkucera: Thanks for the update.
Did you test that it works with different binning provided from JSON?
Can you do the same with `corrBinningMcGen` too?

Simple-Shyam: No in the previous case I tested it was working. After changing it I just checked for compilation (I will check for json also).  For MCGen, it's different we need getTracksSize which is a local function.
```
    using BinningTypeMCGen = FlexibleBinningPolicy<std::tuple<decltype(getTracksSize)>, aod::mccollision::PosZ, decltype(getTracksSize)>;
    BinningTypeMCGen corrBinningMcGen{{getTracksSize}, {binsZVtx, binsMultiplicityMc}, true};
    ```

Simple-Shyam: I checked the definition from json file and it works fine. For the Mc, I already explained the reason.

vkucera: Thanks

vkucera (approved):

PR #4517, opened 2024-01-30
Common: Modified time threshold for detectors where track time res is…

sarahherrmann:
… a range

For fwd tracks the track time res is gaussian

I used the contains<o2::aod::MFTTracks>() to check if the templated track table contains (or is) the MFt track table

### DIFF
--- a/Common/Core/CollisionAssociation.h
+++ b/Common/Core/CollisionAssociation.h
@@ -242,13 +242,29 @@ class CollisionAssociation
           if constexpr (isCentralBarrel) {
             if (mUsePvAssociation && track.isPVContributor()) {
               thresholdTime = trackTimeRes;
-            } else if (TESTBIT(track.flags(), o2::aod::track::TrackTimeResIsRange)) {
-              thresholdTime = std::sqrt(sigmaTimeRes2) + mTimeMargin;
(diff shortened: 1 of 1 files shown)


Common/Core/CollisionAssociation.h
```diff
@@ -242,13 +242,29 @@ class CollisionAssociation
...
+          else
+          {
+            //the track is not a central track
+            if constexpr (TTracks::template contains<o2::aod::MFTTracks>())
+            {
+              //then the track is an MFT track, or an MFT track with additionnal joined info
+              //in this case TrackTimeResIsRange
+              thresholdTime = trackTimeRes + mNumSigmaForTimeCompat * std::sqrt(collTimeRes2) + mTimeMargin;
+            }
+            else
```
aalkin, line 262: It would probably be better to explicitly check `FwdTracks` here, same as with `MFTTracks`.

jgrosseo:
I am OK with this. @fgrosa Do you have comments?

fgrosa:
> I am OK with this. @fgrosa Do you have comments?

Sorry for the late reply. It looks good to me as well, thanks @sarahherrmann!

jgrosseo (approved):

PR #4531, opened 2024-01-31
EventSelectionQA, Fix fetched BC type for correct dereference

nburmaso:


benedikt-voelkel:
@ekryshen  @strogolo can we merge this to fix the AnalysisQC?
Thanks a lot :)

ekryshen (approved):
Thanks for the fix!

PR #4537, opened 2024-01-31
PWGEM/DQ: EM filters, Boolean list for like-sign pair selection

feisenhu:
This is a draft to include the dielectron triggers to the filtering schemes.

Using the filterPPwithAssociation.cxx task
Adding two filters: 
kLMeeIMR     (triggering for ee pairs with mass 1.3 to 3.5 GeV/c2)
kLMeeHMR   (triggering for ee pairs with mass above 3.5 GeV/c2)

Need to select LS pairs only for EM dielectron triggers.
So I implemented the selection of LS pairs by a comma separated list of booleans for each filter selected.

I would like to ask @vfeuilla to have a look at this PR. 
Local errors have prevented me to test it, yet.


vfeuilla:
Hi @feisenhu, It looks good on concept to me! I think you have an MegaLinter error to fix (use `reinterpret_cast<TString*>(objArrayLS->At(icut))` instead of `(TString*)objArrayMuonLS->At(icut)`), but the way to do the decision to keep the LS is good to me

feisenhu:
I have updated the task with fixing the MegaLinter Error and also tested the implementation.

dsekihat (approved):

PR #4539, opened 2024-01-31
PWGHF: further optimise HF combinatorial

fgrosa:
This PR adds two optimisations for the HF combinatorial:
- For Lc->pK0S, the bachelor track is filtered considering the proton PID (no other hypotheses are used)
- For 3-prong channels, it adds the possibility to apply a cut on the decay length computed with the first two tracks to avoid the third loop


mfaggin (approved):
thanks @fgrosa I approve it

PR #4557, opened 2024-02-02
PWGLF: h-strange corr. analysis subscription cleanup

kcui1:
Here we change V0datas to V0dataswithoutrackX


ddobrigk:
> Here we change V0datas to V0dataswithoutrackX

Thanks @kcui1 , I just changed the name of the PR to avoid confusion :-)

njacazio (approved):

PR #4591, opened 2024-02-05
PWGHF: protect access to cascData in toXiPi creator

ZFederica:
Protect access to cascData in candidateCreatorToXiPi (prevents the jobs from failing if the selections used for v0/cascades changed between derived data production and the start of the analysis on derived data)


ZFederica:
Hello @fgrosa, this should fix the problem with the invalid index access

fgrosa (approved):

fgrosa:
> Hello @fgrosa, this should fix the problem with the invalid index access

Thanks for finding and fixing it!

PR #4596, opened 2024-02-05
DPG: Lambda daughters momentum resolution study

creetz16:
Short task to study the momentum resolution of (Anti-)Lambda daughter tracks in MC.
@njacazio 

I extended the `V0Covs` table to include the daughter track covariance matrices. @ddobrigk could you have a look if this is fine? Or would it be better to access the covariances of the daughters at the vertex in a different way?


njacazio:
@creetz16 could you also merge https://github.com/creetz16/O2Physics/pull/8 first?

njacazio (approved):

PR #4617, opened 2024-02-06
Creating full jet pp spectra analysis task 

Archita-Dash:
This PR creates my analysis task (in PWGJE/Tasks) for studying the fulljet pp spectra in Run3. For now, this simple task selects on the EMCAL clusters (with JE track and event selections) and thereby fills some of the basic fulljet QA histograms as a first look. I have tested this locally using apass4 Run3 pp data and it works!

Once merged, I will be gradually developing the task for my analysis in Run3. 
Note: The task is kind of hard-coded for jet radius R=0.4 , for now. But I will change this soon.

### DIFF
--- a/PWGJE/Tasks/CMakeLists.txt
+++ b/PWGJE/Tasks/CMakeLists.txt
@@ -116,4 +116,8 @@ if(FastJet_FOUND)
                         SOURCES jetLundReclustering.cxx
                         PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::PWGJECore O2Physics::AnalysisCore FastJet::FastJet FastJet::Contrib
                         COMPONENT_NAME Analysis)
+    o2physics_add_dpl_workflow(full-jet-spectra-pp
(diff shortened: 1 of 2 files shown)


Archita-Dash:
Hi Nima,

Thank you for your comments. Well, my intention (and preference) was to create this separate fulljet analysis specific task but yes, I agree for now it only serves the purpose of QA. However, as I am locally developing this task constantly for my analysis, I thought it would be feasible if I have it on the official repo in order to make the changes easily.

Nevertheless, I also agree with what you just said above. I can make a new commit once I have developed this task in to a more analysis specific one. :) In that case, should I close this commit myself?

Cheers,
Archita

nzardosh:
Hi Archita,

Thanks for your reply and for agreeing to the change. I think indeed it would be best if we develop the QA tasks for any QA that is needed (you should feel free to add any required plots to the QA task) and you put your task up when it is more analysis specific :)
Yes please go ahead and close the PR.

Thanks,
Nima

PR #4621, opened 2024-02-06
LF/DPG: Add separate table for dau Covs

njacazio:


njacazio:
follow up on https://github.com/AliceO2Group/O2Physics/pull/4596

njacazio:
@ddobrigk

ddobrigk:
Thanks @njacazio - also now I also realised the soa marker is unnecessary in these tables cause originally I meant to do some things I did not do. That can be removed ... if you want to do it right away :-) otherwise I can also do that. I will also inspect the cascadebuilder, I think currently it always subscribes to `V0Covs` but strictly this is only necessary if people disable the option of the dcafitter to use pure geometric distances (and nobody disables that). I will check that shortly. Thanks again :-D

alibuild (approved):

njacazio:
@creetz16

PR #4630, opened 2024-02-07
EMCALCorrection, Fix fetched McParticles type for correct dereference

benedikt-voelkel:


benedikt-voelkel:
@mfasDa @ddobrigk could this be merged?
It will finally fix our AnalysisQC which is failing for quite some time now

mfasDa (approved):

PR #4644, opened 2024-02-08
DPL Analysis: remove uses of selected packs

aalkin:
This removes uses of selected packs from O2Physics, so that we can proceed with removing their recursive template definitions from O2. This will eventually considerably improve compilation times.


mpuccio (approved):

mpuccio:
I can't merge directly as this PR touches another directory as well, but OK for me.

jgrosseo (approved):

PR #4646, opened 2024-02-08
fixed issue with accessing BC globalIndex from event-selection

iarsene:


alibuild (approved):

PR #4648, opened 2024-02-09
Split centrality and multiplicity tables, produce only what is needed

njacazio:


ddobrigk (approved):
Cross-checked with @njacazio to be fine, approving, thanks!

PR #4661, opened 2024-02-10
PWGHF: Implement B0 ML and templatize task

AlexBigO:
Hi @fgrosa , I implemented the ML inference for B0 in the reduced workflow.

Modifs:
- new ML-dedicated header for B0
- reduced B0 selector adapted so one can use ML selection
- reduced B0 task templatized to process data/MC and also to process either with ML information on D meson or on B0 (if we want D meson ML scores, it is to train BDT for B0; and if we want B0 ML scores, we already used the D meson ML scores for the training so we don't need to access them again --> no need to access D meson and B0 ML scores simultaneously).
- a bug was fixed: the B0 candidate tables consumed by the MC process functions did not use to be filtered, they are now

@apalasciano I tag you in case you're interested in using ML for B+ :)

### DIFF
--- a/PWGHF/D2H/TableProducer/candidateSelectorB0ToDPiReduced.cxx
+++ b/PWGHF/D2H/TableProducer/candidateSelectorB0ToDPiReduced.cxx
@@ -68,15 +68,15 @@ struct HfCandidateSelectorB0ToDPiReduced {
   Configurable<std::vector<std::string>> namesInputFeatures{"namesInputFeatures", std::vector<std::string>{"feature1", "feature2"}, "Names of ML model input features"};
   // CCDB configuration
   Configurable<std::string> ccdbUrl{"ccdbUrl", "http://alice-ccdb.cern.ch", "url of the ccdb repository"};
-  Configurable<std::vector<std::string>> modelPathsCCDB{"modelPathsCCDB", std::vector<std::string>{"EventFiltering/PWGHF/BDTDPlus"}, "Paths of models on CCDB"};
-  Configurable<std::vector<std::string>> onnxFileNames{"onnxFileNames", std::vector<std::string>{"ModelHandler_onnx_DPlusToKPiPi.onnx"}, "ONNX file names for each pT bin (if not from CCDB full path)"};
+  Configurable<std::vector<std::string>> modelPathsCCDB{"modelPathsCCDB", std::vector<std::string>{"path_ccdb/BDT_B0/"}, "Paths of models on CCDB"};
+  Configurable<std::vector<std::string>> onnxFileNames{"onnxFileNames", std::vector<std::string>{"ModelHandler_onnx_B0ToDPi.onnx"}, "ONNX file names for each pT bin (if not from CCDB full path)"};
   Configurable<int64_t> timestampCCDB{"timestampCCDB", -1, "timestamp of the ONNX file for ML model used to query in CCDB"};
   Configurable<bool> loadModelsFromCCDB{"loadModelsFromCCDB", false, "Flag to enable or disable the loading of models from CCDB"};
   // variable that will store the value of selectionFlagD (defined in dataCreatorDplusPiReduced.cxx)
   int mySelectionFlagD = -1;
 
   o2::analysis::HfMlResponseB0ToDPi<float> hfMlResponse;
-  std::vector<float> outputMlNotPreselected = {};
+  float outputMlNotPreselected = -1.;
   std::vector<float> outputMl = {};
   o2::ccdb::CcdbApi ccdbApi;
 
@@ -223,7 +223,7 @@ struct HfCandidateSelectorB0ToDPiReduced {
         // B0 ML selections
         std::vector<float> inputFeatures = hfMlResponse.getInputFeatures<withDmesMl>(hfCandB0, trackPi);
         bool isSelectedMl = hfMlResponse.isSelectedMl(inputFeatures, ptCandB0, outputMl);
-        hfMlB0ToDPiCandidate(outputMl);
+        hfMlB0ToDPiCandidate(outputMl[1]); // storing ML score for signal class
 
         if (!isSelectedMl) {
           hfSelB0ToDPiCandidate(statusB0ToDPi);
--- a/PWGHF/D2H/TableProducer/candidateSelectorB0ToDPiReduced.cxx
+++ b/PWGHF/D2H/TableProducer/candidateSelectorB0ToDPiReduced.cxx
@@ -21,6 +21,7 @@
 #include "Common/Core/TrackSelectorPID.h"
 
 #include "PWGHF/Core/HfHelper.h"
+#include "PWGHF/Core/HfMlResponseB0ToDPi.h"
 #include "PWGHF/Core/SelectorCuts.h"
 #include "PWGHF/DataModel/CandidateReconstructionTables.h"
 #include "PWGHF/DataModel/CandidateSelectionTables.h"
@@ -33,6 +34,7 @@ using namespace o2::analysis;
 
 struct HfCandidateSelectorB0ToDPiReduced {
   Produces<aod::HfSelB0ToDPi> hfSelB0ToDPiCandidate; // table defined in CandidateSelectionTables.h
+  Produces<aod::HfMlB0ToDPi> hfMlB0ToDPiCandidate;   // table defined in CandidateSelectionTables.h
 
   Configurable<float> ptCandMin{"ptCandMin", 0., "Lower bound of candidate pT"};
   Configurable<float> ptCandMax{"ptCandMax", 50., "Upper bound of candidate pT"};
@@ -57,16 +59,33 @@ struct HfCandidateSelectorB0ToDPiReduced {
   Configurable<LabeledArray<double>> cutsDmesMl{"cutsDmesMl", {hf_cuts_ml::cuts[0], hf_cuts_ml::nBinsPt, hf_cuts_ml::nCutScores, hf_cuts_ml::labelsPt, hf_cuts_ml::labelsDmesCutScore}, "D-meson ML cuts per pT bin"};
   // QA switch
   Configurable<bool> activateQA{"activateQA", false, "Flag to enable QA histogram"};
-
+  // B0 ML inference
+  Configurable<bool> applyB0Ml{"applyB0Ml", false, "Flag to apply ML selections"};
+  Configurable<std::vector<double>> binsPtB0Ml{"binsPtB0Ml", std::vector<double>{hf_cuts_ml::vecBinsPt}, "pT bin limits for ML application"};
+  Configurable<std::vector<int>> cutDirB0Ml{"cutDirB0Ml", std::vector<int>{hf_cuts_ml::vecCutDir}, "Whether to reject score values greater or smaller than the threshold"};
+  Configurable<LabeledArray<double>> cutsB0Ml{"cutsB0Ml", {hf_cuts_ml::cuts[0], hf_cuts_ml::nBinsPt, hf_cuts_ml::nCutScores, hf_cuts_ml::labelsPt, hf_cuts_ml::labelsCutScore}, "ML selections per pT bin"};
+  Configurable<int8_t> nClassesB0Ml{"nClassesB0Ml", (int8_t)hf_cuts_ml::nCutScores, "Number of classes in ML model"};
+  Configurable<std::vector<std::string>> namesInputFeatures{"namesInputFeatures", std::vector<std::string>{"feature1", "feature2"}, "Names of ML model input features"};
+  // CCDB configuration
+  Configurable<std::string> ccdbUrl{"ccdbUrl", "http://alice-ccdb.cern.ch", "url of the ccdb repository"};
+  Configurable<std::vector<std::string>> modelPathsCCDB{"modelPathsCCDB", std::vector<std::string>{"EventFiltering/PWGHF/BDTDPlus"}, "Paths of models on CCDB"};
+  Configurable<std::vector<std::string>> onnxFileNames{"onnxFileNames", std::vector<std::string>{"ModelHandler_onnx_DPlusToKPiPi.onnx"}, "ONNX file names for each pT bin (if not from CCDB full path)"};
+  Configurable<int64_t> timestampCCDB{"timestampCCDB", -1, "timestamp of the ONNX file for ML model used to query in CCDB"};
+  Configurable<bool> loadModelsFromCCDB{"loadModelsFromCCDB", false, "Flag to enable or disable the loading of models from CCDB"};
   // variable that will store the value of selectionFlagD (defined in dataCreatorDplusPiReduced.cxx)
   int mySelectionFlagD = -1;
 
-  HfHelper hfHelper;
-  TrackSelectorPi selectorPion;
+  o2::analysis::HfMlResponseB0ToDPi<float> hfMlResponse;
+  std::vector<float> outputMlNotPreselected = {};
+  std::vector<float> outputMl = {};
+  o2::ccdb::CcdbApi ccdbApi;
 
-  HistogramRegistry registry{"registry"};
+  TrackSelectorPi selectorPion;
+  HfHelper hfHelper;
 
   using TracksPion = soa::Join<HfRedTracks, HfRedTracksPid>;
+
+  HistogramRegistry registry{"registry"};
 
   void init(InitContext const& initContext)
   {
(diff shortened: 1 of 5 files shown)


AlexBigO:
Hi, it looks like I can't pass both Clang-format and MegaLinter simultaneously (because of a macro defined in `HfMlResponseB0ToDPi.h`). Does someone know how to overcome this, please?

PWGHF/D2H/TableProducer/candidateSelectorB0ToDPiReduced.cxx
```diff
@@ -57,17 +59,34 @@ struct HfCandidateSelectorB0ToDPiReduced {
...
+  Configurable<std::vector<int>> cutDirB0Ml{"cutDirB0Ml", std::vector<int>{hf_cuts_ml::vecCutDir}, "Whether to reject score values greater or smaller than the threshold"};
+  Configurable<LabeledArray<double>> cutsB0Ml{"cutsB0Ml", {hf_cuts_ml::cuts[0], hf_cuts_ml::nBinsPt, hf_cuts_ml::nCutScores, hf_cuts_ml::labelsPt, hf_cuts_ml::labelsCutScore}, "ML selections per pT bin"};
+  Configurable<int8_t> nClassesB0Ml{"nClassesB0Ml", (int8_t)hf_cuts_ml::nCutScores, "Number of classes in ML model"};
+  Configurable<std::vector<std::string>> namesInputFeatures{"namesInputFeatures", std::vector<std::string>{"feature1", "feature2"}, "Names of ML model input features"};
+  // CCDB configuration
+  Configurable<std::string> ccdbUrl{"ccdbUrl", "http://alice-ccdb.cern.ch", "url of the ccdb repository"};
+  Configurable<std::vector<std::string>> modelPathsCCDB{"modelPathsCCDB", std::vector<std::string>{"EventFiltering/PWGHF/BDTDPlus"}, "Paths of models on CCDB"};
+  Configurable<std::vector<std::string>> onnxFileNames{"onnxFileNames", std::vector<std::string>{"ModelHandler_onnx_DPlusToKPiPi.onnx"}, "ONNX file names for each pT bin (if not from CCDB full path)"};
```
fgrosa, line 72: Maybe here we can put a "dummy" example instead of a real one for the D+? (just to be sure that people don't run it without a proper model and get a random output)

PWGHF/D2H/TableProducer/candidateSelectorB0ToDPiReduced.cxx
```diff
@@ -57,17 +59,34 @@ struct HfCandidateSelectorB0ToDPiReduced {
   Configurable<LabeledArray<double>> cutsDmesMl{"cutsDmesMl", {hf_cuts_ml::cuts[0], hf_cuts_ml::nBinsPt, hf_cuts_ml::nCutScores, hf_cuts_ml::labelsPt, hf_cuts_ml::labelsDmesCutScore}, "D-meson ML cuts per pT bin"};
   // QA switch
   Configurable<bool> activateQA{"activateQA", false, "Flag to enable QA histogram"};
-
+  // B0 ML inference
+  Configurable<bool> applyB0Ml{"applyB0Ml", false, "Flag to apply ML selections"};
+  Configurable<std::vector<double>> binsPtB0Ml{"binsPtB0Ml", std::vector<double>{hf_cuts_ml::vecBinsPt}, "pT bin limits for ML application"};
+  Configurable<std::vector<int>> cutDirB0Ml{"cutDirB0Ml", std::vector<int>{hf_cuts_ml::vecCutDir}, "Whether to reject score values greater or smaller than the threshold"};
+  Configurable<LabeledArray<double>> cutsB0Ml{"cutsB0Ml", {hf_cuts_ml::cuts[0], hf_cuts_ml::nBinsPt, hf_cuts_ml::nCutScores, hf_cuts_ml::labelsPt, hf_cuts_ml::labelsCutScore}, "ML selections per pT bin"};
+  Configurable<int8_t> nClassesB0Ml{"nClassesB0Ml", (int8_t)hf_cuts_ml::nCutScores, "Number of classes in ML model"};
```
fgrosa, line 67: Unless I miss something, for B mesons the ML classifier should always be binary (signal and background). Hence I was wondering whether it is necessary to define a `std::vector` for it, or simply use a value. 
Maybe we could start to support a single value and in the future if there will be a usecase for multi-class classification for B mesons we can extend it.

PWGHF/D2H/TableProducer/candidateSelectorB0ToDPiReduced.cxx
```diff
@@ -57,17 +59,34 @@ struct HfCandidateSelectorB0ToDPiReduced {
...
+  Configurable<int64_t> timestampCCDB{"timestampCCDB", -1, "timestamp of the ONNX file for ML model used to query in CCDB"};
+  Configurable<bool> loadModelsFromCCDB{"loadModelsFromCCDB", false, "Flag to enable or disable the loading of models from CCDB"};
   // variable that will store the value of selectionFlagD (defined in dataCreatorDplusPiReduced.cxx)
   int mySelectionFlagD = -1;
 
-  HfHelper hfHelper;
-  TrackSelectorPi selectorPion;
+  o2::analysis::HfMlResponseB0ToDPi<float> hfMlResponse;
+  std::vector<float> outputMlNotPreselected = {};
+  std::vector<float> outputMl = {};
```
fgrosa, line 80: See above the comment about the number of classes.

PWGHF/DataModel/CandidateSelectionTables.h
```diff
@@ -237,12 +237,16 @@ DECLARE_SOA_TABLE(HfSelLcToK0sP, "AOD", "HFSELLCK0SP", //!
 
 namespace hf_sel_candidate_b0
 {
-DECLARE_SOA_COLUMN(IsSelB0ToDPi, isSelB0ToDPi, int); //!
+DECLARE_SOA_COLUMN(IsSelB0ToDPi, isSelB0ToDPi, int);                  //!
+DECLARE_SOA_COLUMN(MlProbB0ToDPi, mlProbB0ToDPi, std::vector<float>); //!
```
fgrosa, line 241: See above the comment about the number of classes.

PWGHF/D2H/Tasks/taskB0Reduced.cxx
```diff
@@ -76,6 +78,8 @@ DECLARE_SOA_TABLE(HfRedCandB0Lites, "AOD", "HFREDCANDB0LITE", //! Table with som
                   hf_cand_b0_reduced::Prong0MlScoreBkg,
                   hf_cand_b0_reduced::Prong0MlScorePrompt,
                   hf_cand_b0_reduced::Prong0MlScoreNonprompt,
+                  hf_cand_b0_lite::MlScoreBkg,
+                  hf_cand_b0_lite::MlScorePrompt,
```
fgrosa, line 82: There is no need to have both, one is the complement of the other one.

PWGHF/D2H/Tasks/taskB0Reduced.cxx
```diff
@@ -57,6 +57,8 @@ DECLARE_SOA_COLUMN(ImpactParameterProduct, impactParameterProduct, float);   //!
 DECLARE_SOA_COLUMN(Cpa, cpa, float);                                         //! Cosine pointing angle of candidate
 DECLARE_SOA_COLUMN(CpaXY, cpaXY, float);                                     //! Cosine pointing angle of candidate in transverse plane
 DECLARE_SOA_COLUMN(MaxNormalisedDeltaIP, maxNormalisedDeltaIP, float);       //! Maximum normalized difference between measured and expected impact parameter of candidate prongs
+DECLARE_SOA_COLUMN(MlScoreBkg, mlScoreBkg, float);                           //! ML score for background class
+DECLARE_SOA_COLUMN(MlScorePrompt, mlScorePrompt, float);                     //! ML score for prompt class
```
fgrosa, line 61: There is no need to have both, one is the complement of the other one.

PWGHF/D2H/Tasks/taskB0Reduced.cxx
```diff
@@ -262,14 +272,23 @@ struct HfTaskB0Reduced {
...
           registry.add("hMlScoreNonPromptDRecBg", "B^{0} candidates (unmatched);#it{p}_{T}(D^{#minus}) (GeV/#it{c});prong0, D^{#minus} ML nonprompt score;entries", {HistType::kTH2F, {axisPtDminus, axisMlScore}});
         }
+        // ML scores of B0 candidate
+        if (doprocessMcWithB0Ml) {
+          // signal
+          registry.add("hMlScoreBkgB0RecSig", "B^{0} candidates (matched);#it{p}_{T}(B^{0}) (GeV/#it{c});prong0, B^{0} ML background score;entries", {HistType::kTH2F, {axisPtB0, axisMlScore}});
+          registry.add("hMlScorePromptB0RecSig", "B^{0} candidates (matched);#it{p}_{T}(B^{0}) (GeV/#it{c});prong0, B^{0} ML prompt score;entries", {HistType::kTH2F, {axisPtB0, axisMlScore}});
+          // background
+          registry.add("hMlScoreBkgB0RecBg", "B^{0} candidates (unmatched);#it{p}_{T}(B^{0}) (GeV/#it{c});prong0, B^{0} ML background score;entries", {HistType::kTH2F, {axisPtB0, axisMlScore}});
+          registry.add("hMlScorePromptB0RecBg", "B^{0} candidates (unmatched);#it{p}_{T}(B^{0}) (GeV/#it{c});prong0, B^{0} ML prompt score;entries", {HistType::kTH2F, {axisPtB0, axisMlScore}});
```
fgrosa, line 282: There is no need to have both scores, one is the complement of the other one.

PWGHF/D2H/Tasks/taskB0Reduced.cxx
```diff
@@ -315,199 +336,161 @@ struct HfTaskB0Reduced {
...
+            registry.fill(HIST("hDecayTypeMc"), 1 + hf_cand_b0::DecayTypeMc::B0ToDplusPiToPiKPiPi, invMassB0, ptCandB0);
+          }
+          if constexpr (withDmesMl) {
+            registry.fill(HIST("hMlScoreBkgDRecSig"), ptD, candidate.prong0MlScoreBkg());
+            registry.fill(HIST("hMlScorePromptDRecSig"), ptD, candidate.prong0MlScorePrompt());
+            registry.fill(HIST("hMlScoreNonPromptDRecSig"), ptD, candidate.prong0MlScoreNonprompt());
+          }
+          if constexpr (withB0Ml) {
+            registry.fill(HIST("hMlScoreBkgB0RecSig"), ptCandB0, candidate.mlProbB0ToDPi()[0]);
+            registry.fill(HIST("hMlScorePromptB0RecSig"), ptCandB0, candidate.mlProbB0ToDPi()[1]);
```
fgrosa, line 377: There is no need to have both, one is the complement of the other one.

PWGHF/D2H/Tasks/taskB0Reduced.cxx
```diff
@@ -315,199 +336,161 @@ struct HfTaskB0Reduced {
...
+          registry.fill(HIST("hCospDRecBg"), ptD, cospD);
+          registry.fill(HIST("hCospXyDRecBg"), ptD, cospXyD);
+          if constexpr (withDmesMl) {
+            registry.fill(HIST("hMlScoreBkgDRecBg"), ptD, candidate.prong0MlScoreBkg());
+            registry.fill(HIST("hMlScorePromptDRecBg"), ptD, candidate.prong0MlScorePrompt());
+            registry.fill(HIST("hMlScoreNonPromptDRecBg"), ptD, candidate.prong0MlScoreNonprompt());
+          }
+          if constexpr (withB0Ml) {
+            registry.fill(HIST("hMlScoreBkgB0RecBg"), ptCandB0, candidate.mlProbB0ToDPi()[0]);
+            registry.fill(HIST("hMlScorePromptB0RecBg"), ptCandB0, candidate.mlProbB0ToDPi()[1]);
```
fgrosa, line 405: There is no need to have both, one is the complement of the other one.

PWGHF/D2H/Tasks/taskB0Reduced.cxx
```diff
@@ -315,199 +336,161 @@ struct HfTaskB0Reduced {
...
-        registry.fill(HIST("hMlScorePromptD"), ptD, candidate.prong0MlScorePrompt());
-        registry.fill(HIST("hMlScoreNonPromptD"), ptD, candidate.prong0MlScoreNonprompt());
+        if constexpr (withDmesMl) {
+          registry.fill(HIST("hMlScoreBkgD"), ptD, candidate.prong0MlScoreBkg());
+          registry.fill(HIST("hMlScorePromptD"), ptD, candidate.prong0MlScorePrompt());
+          registry.fill(HIST("hMlScoreNonPromptD"), ptD, candidate.prong0MlScoreNonprompt());
+        }
+        if constexpr (withB0Ml) {
+          registry.fill(HIST("hMlScoreBkgB0"), ptCandB0, candidate.mlProbB0ToDPi()[0]);
+          registry.fill(HIST("hMlScorePromptB0"), ptCandB0, candidate.mlProbB0ToDPi()[1]);
```
fgrosa, line 443: There is no need to have both, one is the complement of the other one.

PWGHF/D2H/Tasks/taskB0Reduced.cxx
```diff
@@ -315,199 +336,161 @@ struct HfTaskB0Reduced {
...
       float pseudoRndm = ptD * 1000. - (int64_t)(ptD * 1000);
-      if (ptCandB0 >= ptMaxForDownSample || pseudoRndm < downSampleBkgFactor) {
-        int8_t flag{0};
-        int8_t origin{0};
+      if (flagMcMatchRec != 0 || (((doMc && fillBackground) || !doMc) && (ptCandB0 >= ptMaxForDownSample || pseudoRndm < downSampleBkgFactor))) {
         float prong0MlScoreBkg = -1.;
         float prong0MlScorePrompt = -1.;
         float prong0MlScoreNonprompt = -1.;
+        float candidateMlScoreBkg = -1;
+        float candidateMlScorePrompt = -1;
```
fgrosa, line 477: There is no need to have both, one is the complement of the other one.

PWGHF/D2H/Tasks/taskB0Reduced.cxx
```diff
@@ -315,199 +336,161 @@ struct HfTaskB0Reduced {
...
         if constexpr (withDmesMl) {
           prong0MlScoreBkg = candidate.prong0MlScoreBkg();
           prong0MlScorePrompt = candidate.prong0MlScorePrompt();
           prong0MlScoreNonprompt = candidate.prong0MlScoreNonprompt();
         }
+        if constexpr (withB0Ml) {
+          candidateMlScoreBkg = candidate.mlProbB0ToDPi()[0];
+          candidateMlScorePrompt = candidate.mlProbB0ToDPi()[1];
+        }
```
fgrosa, line 486: There is no need to have both, one is the complement of the other one.

fgrosa (changes requested):
Thanks a lot @AlexBigO!
I essentially just have a comment about the dimensionality of the output score (since in binary classification signal score = 1 - background score I would only store and use one value)

Regarding the clang format vs Megalinter issue, we had exactly the same for the Lc->pKpi channel with @mfaggin and in the end we merged it with Megalinter red, but maybe @vkucera has a solution for that.

PWGHF/D2H/TableProducer/candidateSelectorB0ToDPiReduced.cxx
AlexBigO: Currently, the way the [`MlResponse`](https://github.com/AliceO2Group/O2Physics/blob/master/Tools/ML/MlResponse.h#L60) class is written, we need vectors as input. In the config, one just needs to deactivate one of the two scores (with `cutDir = 2`) to use only one score.

PWGHF/D2H/TableProducer/candidateSelectorB0ToDPiReduced.cxx
AlexBigO: I have to keep the `outputMl` as a vector as [this function](https://github.com/AliceO2Group/O2Physics/blob/master/Tools/ML/MlResponse.h#L183) requires a vector. However, I only store `outputMl[1]` in the AOD table in the end.

vkucera:
> Hi, it looks like I can't pass both Clang-format and MegaLinter simultaneously (because of a macro defined in `HfMlResponseB0ToDPi.h`). Does someone know how to overcome this, please?

This seems to be coming from a conflict between the different clang-format versions used in the two CI checks. For now, I would suggest to fix the formatting according to the clang-format check (which is required for the merging) and merge the PR despite the failing MegaLinter check.

AlexBigO:
> > Hi, it looks like I can't pass both Clang-format and MegaLinter simultaneously (because of a macro defined in `HfMlResponseB0ToDPi.h`). Does someone know how to overcome this, please?
> 
> This seems to be coming from a conflict between the different clang-format versions used in the two CI checks. For now, I would suggest to fix the formatting according to the clang-format check (which is required for the merging) and merge the PR despite the failing MegaLinter check.

Hi Vit, I will do this then. Thanks!

PR #4668, opened 2024-02-11
Second step towards identified analysis

victor-gonzalez:
Also collisions within ITS read out frame border are rejected


alibuild (approved):

PR #4671, opened 2024-02-12
PWGEM/PhotonMeson: revert mean ITS cluster size cut

dsekihat:
revert mean ITS cluster size cut


alibuild (approved):

PR #4674, opened 2024-02-12
Fix bc() access

fmazzasc:
@mdiotti


alibuild (approved):

PR #4681, opened 2024-02-12
Fix data type to avoid conflicts with AnalysisDataModel.h

mfaggin:
Fix to avoid conflicts with `TracksIU` defined here https://github.com/AliceO2Group/AliceO2/blob/66de30958153cd7badf522150e8554f9fcf975ff/Framework/Core/include/Framework/AnalysisDataModel.h#L392
This was creating unexpected issues with HF derived data.

Thanks @fgrosa for helping me in the debug.

@jgrosseo @ddobrigk to be merged asap.

### DIFF
--- a/Common/TableProducer/trackPropagation.cxx
+++ b/Common/TableProducer/trackPropagation.cxx
@@ -63,7 +63,6 @@ struct TrackPropagation {
   OutputObj<TH1D> trackTunedTracks{TH1D("trackTunedTracks", "", 1, 0.5, 1.5), OutputObjHandlingPolicy::AnalysisObject};
 
   using TracksIUWithMc = soa::Join<aod::StoredTracksIU, aod::McTrackLabels, aod::TracksCovIU>;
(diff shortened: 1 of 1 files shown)


jgrosseo:
Good catch! Shouldn't we just rename the `using TracksUI` to something else?

mfaggin:
> Good catch! Shouldn't we just rename the `using TracksUI` to something else?

Hi @jgrosseo not a strong opinion actually, in the end I decided to put the `Join`ed type explicitly given that it is the only place where it is used. If you have a preference I can change, no problem, just let me know.

jgrosseo (approved):

jgrosseo:
> > Good catch! Shouldn't we just rename the `using TracksUI` to something else?
> 
> Hi @jgrosseo not a strong opinion actually, in the end I decided to put the `Join`ed type explicitly given that it is the only place where it is used. If you have a preference I can change, no problem, just let me know.

I guess both is fine...

PR #4699, opened 2024-02-13
Use aggregate initialisation for JetFinder

ktf:


ktf:
@qgp @nzardosh this should actually work...

qgp:
Duplicate of #4697

qgp:
> fine with me, although this version handles the ghost* as before. i do not know if that is actually desirable, though.

@ktf What's wrong with default initialisation (instead of zero initialisation) for these variables?

ktf:
Before your fix, the `ghost<something>` variables were not 0-initialised, but initialised starting from the respecive `<something>`. For example, `ghostEtaMin` was using `eta_Min` as initial value. I guess this was the reason / motivation behind the explicit template. You can still retain that behavior, but then you need to fix your aggregate initialisation using:

```C++
struct JetFinder {
   float etaMin = 0.9;
   ...
   float ghostEtaMin = etaMin;
   ...
};
```

This is what would have the same effect as the old constructor.

ktf:
PR updated to show what I mean...

PWGJE/Core/JetFinder.h
```diff
@@ -79,6 +79,7 @@ class JetFinder
   fastjet::RecombinationScheme recombScheme = fastjet::E_scheme;
   fastjet::Strategy strategy = fastjet::Best;
   fastjet::AreaType areaType = fastjet::active_area;
+
```
ktf, line 82:
```suggestion
```

qgp:
Sorry, then I missed your point before. Indeed, I discussed this with Nima and we deliberately chose to initialise the `ghost<sth>` with fixed values and not derive them from the constituent limits. I should have mentioned this in my PR, sorry for the omission.

ktf:
Ok, fair enough. Thank you for understanding the original issue!

PR #4703, opened 2024-02-13
PWGJE : Fixing wrong dereferenced tables

nzardosh:
- jet finders moved to new folder and separated into different tasks to avoid additional memory consumption of inactive tasks

- added possibility to apply area cut to jets

- fixed dereferencing errors

- added histograms to QA tasks and applied bug fixes

- jet matching now only matches jets with the same resolution parameter

- added ability to write derived data based on event-wise constituent subtracted jets


alibuild (approved):

PR #4711, opened 2024-02-14
[PWGUD] Exclusive 2pi, 2K, 2particles, 4pi, 6pi, rho polarisation

siragoni:
First task for analysis in the central barrel of:
- 2pi states
- 2K states 
- 2K states without PID, using only ITS
- 4pi states 
- 6pi states (3+3 tau topology)
- Rho polarisation


siragoni:
Hi @pbuehler this is the task Anisa and I are working on together with different preliminary ideas, which we would like to run on hyperloop to see the effects with the full statistics. Thanks!

pbuehler (approved):

PR #4725, opened 2024-02-15
K0Perf: Add check on the ITS IB

njacazio:
- add way to compute inv mass from daughters

### DIFF
--- a/DPG/Tasks/AOTTrack/V0Cascades/perfK0sResolution.cxx
+++ b/DPG/Tasks/AOTTrack/V0Cascades/perfK0sResolution.cxx
@@ -103,12 +103,14 @@ struct perfK0sResolution {
 
   Configurable<float> v0lifetime{"v0lifetime", 3., "n ctau"};
   Configurable<float> nSigTPC{"nSigTPC", 10., "nSigTPC"};
+  Configurable<int> itsIbSelectionPos{"itsIbSelectionPos", 0, "Flag for the ITS IB selection on positive daughters: -1 no ITS IB, 0 no selection, 1 ITS IB"};
+  Configurable<int> itsIbSelectionNeg{"itsIbSelectionNeg", 0, "Flag for the ITS IB IB selection on negative daughters: -1 no ITS IB, 0 no selection, 1 ITS IB"};
   Configurable<int> trdSelectionPos{"trdSelectionPos", 0, "Flag for the TRD selection on positive daughters: -1 no TRD, 0 no selection, 1 TRD"};
   Configurable<int> trdSelectionNeg{"trdSelectionNeg", 0, "Flag for the TRD selection on negative daughters: -1 no TRD, 0 no selection, 1 TRD"};
   Configurable<int> tofSelectionPos{"tofSelectionPos", 0, "Flag for the TOF selection on positive daughters: -1 no TOF, 0 no selection, 1 TOF"};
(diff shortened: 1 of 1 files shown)


njacazio:
@mfaggin can we use the tuner on this setup?

mfaggin:
> @mfaggin can we use the tuner on this setup?

ciao @njacazio , the trackTuner smears track parameters only for tracks propagated to the primary vertex. As far as I understood, the lambdaK0Builder works starting from TracksIU only.

Now, from what I understand in this PR you are getting the V0s, reconstructed from TracksIU, and for each daughter you retrieve the corresponding Track propagated to the PV. I would say that this should work.

Anyway, @phymanshu and @arossi81 are working to interface the track tuner also with the lambdaKzeroBuilder to smear the pt (still WIP)

njacazio:
Ok, but we're also asking for a hit in the IB, so it should be close enough. Thanks!

mfaggin:
> Ok, but we're also asking for a hit in the IB, so it should be close enough. Thanks!

Hi @njacazio I would say that it is not the key point for the trackTuner to work. The key point is that you wotk with tracks picked-up from the `Tracks` tabel, and not the `TracksIU`. But indeed it looks to me that you are doing well it here. We'll see.

njacazio:
it is indeed the key point in view of the development on the lambdakzerobuilder

alibuild (approved):

arossi81:
Hi @mfaggin , @njacazio ,
the track tuner per-se does not require tracks to be propagated at the PV, it actually starts from trackIU and propagate them at the production point (taken from the MC particle). Now, for the smearing done for the position, the issues related to secondary particles from strange decays is whether one has a data-driven input for the smearing: what we get from the DCA fits of all propagated tracks is an input that is well suited for primary particles. It can be used also for secondaries, but it is not necessarily realistic. 

Concerning the momentum smearing for K0s, something that we verified while doing the study with @phymanshu is that you cannot calculate the invariant mass with the momenta at the primary vertex: you need to take them at approximately the decay point, otherwise you get a worse mass distribution (due to the curvature effect). This was not a minor effect. THat's why what @phymanshu did is to change the TrackIU parameter in the V0builder before all operations, so that everything is done properly.

mfaggin:
> Hi @mfaggin , @njacazio , the track tuner per-se does not require tracks to be propagated at the PV, it actually starts from trackIU and propagate them at the production point (taken from the MC particle). Now, for the smearing done for the position, the issues related to secondary particles from strange decays is whether one has a data-driven input for the smearing: what we get from the DCA fits of all propagated tracks is an input that is well suited for primary particles. It can be used also for secondaries, but it is not necessarily realistic.
> 
> Concerning the momentum smearing for K0s, something that we verified while doing the study with @phymanshu is that you cannot calculate the invariant mass with the momenta at the primary vertex: you need to take them at approximately the decay point, otherwise you get a worse mass distribution (due to the curvature effect). This was not a minor effect. THat's why what @phymanshu did is to change the TrackIU parameter in the V0builder before all operations, so that everything is done properly.

To avoid misunderstanding: what @arossi81 mentions is 100% correct, but one cas smear track parameters only for those propagated to the PV  _**with the current implementation **_ of the trackTuner . This was the core of my previous comments. As I said, Himanshu is working on it to extend it ot the V0builder as you mention.

arossi81:
Hi @mfaggin , now I see what you mean: that all modified tracks are later propagated to the PV, yous was not a comment on the input but on the output.

njacazio:
I think we understood each other, thanks to both! Is there some parametrization I can start from?

PR #4730, opened 2024-02-15
PWGHF: Fix MC check to get charm mother information

AlexBigO:
Hello again @fgrosa 

Thanks for reviewing and merging the PR #4723 so quickly :) and sorry, but I just realized that:
- the D mesons can decay into phi mesons then decaying into pions and kaons ([config file](https://github.com/AliceO2Group/O2DPG/blob/master/MC/config/PWGHF/pythia8/generator/pythia8_charmtriggers_with_Bforced_decay.cfg)) so the `maxDepth` set in the `getMother` function should be `2` for all the D meson hypothesis (and not only D*)
- it might be better to fill the MC check table without this selection on `flagMcMatchRec` so that one could access each component (pure signal, B0->Ds Pi, partly reco decay, ...) in the invariant mass distribution.

### DIFF
--- a/PWGHF/D2H/TableProducer/dataCreatorDplusPiReduced.cxx
+++ b/PWGHF/D2H/TableProducer/dataCreatorDplusPiReduced.cxx
@@ -433,13 +433,9 @@ struct HfDataCreatorDplusPiReduced {
                   pdgCodeProng3 = particleProng1.pdgCode();
                   // look for common c-hadron mother among prongs 0, 1 and 2
                   for (const auto& cHadronMotherHypo : cHadronMotherHypos) {
-                    int8_t depthMax = 1;
-                    if (cHadronMotherHypo == Pdg::kDStar) {
-                      depthMax = 2;
-                    }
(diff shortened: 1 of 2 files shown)


fgrosa:
> Hello again @fgrosa
> 
> This is a fix of the PR #4723
> 
> I just realized that:
> 
>     * the D mesons can decay into phi mesons then decaying into pions and kaons ([config file](https://github.com/AliceO2Group/O2DPG/blob/master/MC/config/PWGHF/pythia8/generator/pythia8_charmtriggers_with_Bforced_decay.cfg)) so the `maxDepth` set in the `getMother` function should be `2` for all the D meson hypothesis (and not only D*)
> 
>     * it might be better to fill the MC check table without this selection on `flagMcMatchRec` so that one could access each component (pure signal, B0->Ds Pi, partly reco decay, ...) in the invariant mass distribution.

Hi @AlexBigO, thanks, indeed I also missed it!
It should be `1` only for 2-prong decays that cannot be resonant (for instance the D0 -> Kpi, that we do not have here however).

fgrosa (approved):

PR #4746, opened 2024-02-16
PWGEM: update MCtemplate task, 

feisenhu:
An update to test if current version  runs on hyper loop. 

still have to update track selection for selected events 
(There is no event selection for when filling track histograms)


alibuild (approved):

PR #4754, opened 2024-02-17
PWGLF: add binned info table for test

ddobrigk:


ddobrigk:
@romainschotter @lhusova this experimental change to the MC handling in the derived data packs binned data now into the derived MC output so that the generated are stored there as well. It saves a `std::vector` containing the histogram contents of the generated particles, one vector per DF; the analysis task will then have to aggregate these vectors and re-fill the histograms. This is a stopgap solution: our derived data handling should be able to add counters that are much better than that, most notably also merging multiple sets of this information over several DFs when generating derived data _(N.B.: saving this vector is a precursor to the better approach of saving an actual histogram that inherently contains the same binned information in a vector inside of it)._

ddobrigk:
enabled auto-merge: this grants the `Common` approval that is also needed for this PR to go through, which is a small modification to the generic MC Collision Extra table (of course, the PWG-LF approval will still be needed on top of that)

njacazio (approved):

alibuild (approved):

PR #4765, opened 2024-02-19
[PWGDQ] collision association QA and additional process function in table-maker with associations

iarsene:
1) table-maker-with-assoc
   Added more process functions, for the pp (filtered) and Pb-Pb like cases.
   Fixed statistics histograms
2) table-reader
   Fixed the bug with not working Configurables in Filter expressions
3) table-reader-with-assoc
   Added QA for collisoin track associations
4) VarManager
   Added function for two event correlations
5) HistogramsLibrary
   Added histogram family for two event correlations


alibuild (approved):

PR #4769, opened 2024-02-19
PWGHF: New features in LcpKpi task and tree creator

DelloStritto:
In this PR:
- The forward percentile multiplicity estimators has been added to the LcpKpi tree creator
- The ML cut is now applied in the Lc selector


mfaggin (approved):
thanks @DelloStritto I approve it. Let's keep in mind that asap we should move to a  flagging as done e.g. in the D+ and Ds selectors, were a bitmap is used to store the information of which selections the candidate survives (preselections, PID, ML, ...)

PR #4782, opened 2024-02-20
PWGHF: Adding some input features to the HfMlResponseDsToKKPi

scattaru:
Dear all,
I've just added some input features (combined PID and impactParameterXY) in the HfMlResponseDsToKKPi to be used for the ML selection.

Best regards,
Samuele


fcatalan92 (approved):
Hi @scattaru, LGTM!
There is something strange happening with the MegaLinter checks, basically it is now suggesting to undo the fix proposed before. @vkucera could you have a look?

fcatalan92 (approved):
Hi @scattaru, I think the check is broken so even if you apply its suggestions it will not go green. It is non blocking for the merge although so we can go ahead.

scattaru:
Hi @fcatalan92, ok perfect. Thanks a lot!

vkucera:
> Hi @scattaru, I think the check is broken so even if you apply its suggestions it will not go green. It is non blocking for the merge although so we can go ahead.

The check is not broken. It is just using a more recent version of clang-format. PR to fix it has been pending.

PR #4783, opened 2024-02-20
PWGHF: Adding reduced data model for Ds resonances

Luca610:
Added tables for charm resonances reduced analysis and data creator for DS1 candidates


PWGHF/D2H/DataModel/ReducedDataModel.h
```diff
@@ -370,6 +370,35 @@ DECLARE_SOA_TABLE(HfCandBpConfigs, "AOD", "HFCANDBPCONFIG", //! Table with confi
...
+                  o2::soa::Index<>,
+                  hf_track_index_reduced::Prong0Id, hf_track_index_reduced::Prong1Id,
+                  hf_track_index_reduced::HfRedCollisionId,
+                  hf_cand::XSecondaryVertex, hf_cand::YSecondaryVertex, hf_cand::ZSecondaryVertex,
+                  hf_reso_cand_reduced::InvMass,
+                  hf_reso_cand_reduced::Px,
+                  hf_reso_cand_reduced::Py,
+                  hf_reso_cand_reduced::Pz);
+
+DECLARE_SOA_TABLE(HfRedD, "AOD", "HFREDD", //! Table with D+/D* candidate information for resonances reduced workflow
```
fgrosa, line 393: I would suggest a different name (also considering that, if we add the D0, it will be likely a separate table)
```suggestion
DECLARE_SOA_TABLE(HfRed3PrNoTrks, "AOD", "HFRED3PRNOTRK", //! Table with D+/D* candidate information for resonances reduced workflow
```

PWGHF/D2H/DataModel/ReducedDataModel.h
```diff
@@ -370,6 +370,35 @@ DECLARE_SOA_TABLE(HfCandBpConfigs, "AOD", "HFCANDBPCONFIG", //! Table with confi
...
+DECLARE_SOA_COLUMN(Py, py, float);
+DECLARE_SOA_COLUMN(Pz, pz, float);
+} // namespace hf_reso_cand_reduced
+
+DECLARE_SOA_TABLE(HfRedVzeros, "AOD", "HFREDVZERO", //! Table with V0 candidate information for resonances reduced workflow
+                  o2::soa::Index<>,
+                  hf_track_index_reduced::Prong0Id, hf_track_index_reduced::Prong1Id,
+                  hf_track_index_reduced::HfRedCollisionId,
+                  hf_cand::XSecondaryVertex, hf_cand::YSecondaryVertex, hf_cand::ZSecondaryVertex,
+                  hf_reso_cand_reduced::InvMass,
```
fgrosa, line 388: Depending on the cuts applied, a V0 can be compatible with more than one hypothesis (K0S, Lambda, AntiLambda). Even if currently you implemented only one channel, I would suggest to add the three columns here, as well as a bitmap (that can be `uint8_t`) that has the information of the mass hypothesis (K0S, Lambda, or AntiLambda), so that it will be easier to add the others later on.

PWGHF/D2H/TableProducer/dataCreatorDV0Reduced.cxx
```diff
@@ -0,0 +1,184 @@
...
+  Processed = 0,
+  NoDV0Selected,
+  DV0Selected,
+  kNEvent
+};
+
+/// Creation of D-V0 pairs
+struct HfDataCreatorDV0Reduced {
+  // Produces AOD tables to store track information
+  Produces<aod::StraCollisions> lfReducedCollision;  // Defined in PWGLF/DataModel/LFStrangenessTables.h
```
fgrosa, line 56: Why are you using here the LF table instead of the `HfRedCollisions`?

PWGHF/D2H/TableProducer/dataCreatorDV0Reduced.cxx
```diff
@@ -0,0 +1,184 @@
...
+
+    for (const auto& candD : candsD) {
+      bool fillHfCandD = false;
+      float invMassD = hfHelper.invMassDplusToPiKPi(candD);
+      std::array<float, 3> pVecD = candD.pVector();
+
+      for (const auto& v0 : V0s) {
+        std::array<float, 3> pVecV0 = {v0.pxpos() + v0.pxneg(), v0.pypos() + v0.pyneg(), v0.pzpos() + v0.pzneg()};
+        float ptV0 = sqrt(pVecV0[0] * pVecV0[0] + pVecV0[1] * pVecV0[1]);
+        auto invMass2DV0 = RecoDecay::m2(std::array{pVecD, pVecV0}, std::array{MassDPlus, MassK0Short});
```
fgrosa, line 124: In order to compute the DV0 invariant mass we should propagate the V0 to the PV. You can see an example here:
https://github.com/AliceO2Group/O2Physics/blob/84fcfb8279ce172a7bd47e1d2e55e183fc7adb40/EventFiltering/PWGHF/HFFilter.cxx#L599-L603

PWGHF/D2H/TableProducer/dataCreatorDV0Reduced.cxx
```diff
@@ -0,0 +1,184 @@
...
+    std::map<int64_t, int64_t> selectedV0s;
+    bool fillHfReducedCollision = false;
+
+    for (const auto& candD : candsD) {
+      bool fillHfCandD = false;
+      float invMassD = hfHelper.invMassDplusToPiKPi(candD);
+      std::array<float, 3> pVecD = candD.pVector();
+
+      for (const auto& v0 : V0s) {
+        std::array<float, 3> pVecV0 = {v0.pxpos() + v0.pxneg(), v0.pypos() + v0.pyneg(), v0.pzpos() + v0.pzneg()};
```
fgrosa, line 122:
```suggestion
        std::array<float, 3> pVecV0 = {v0.px(), v0.py(), v0.pz()};
```

PWGHF/D2H/TableProducer/dataCreatorDV0Reduced.cxx
```diff
@@ -0,0 +1,184 @@
...
+    static const AxisSpec axisEvents = {kNBinsEvents, 0.5, kNBinsEvents + 0.5, ""};
+    registry.add("hEvents", "Events;;entries", HistType::kTH1F, {axisEvents});
+    for (int iBin = 0; iBin < kNBinsEvents; iBin++) {
+      registry.get<TH1>(HIST("hEvents"))->GetXaxis()->SetBinLabel(iBin + 1, labels[iBin].data());
+    }
+    registry.add("hMassDToPiKPi", "D^{#minus} candidates;inv. mass (p^{#minus} K^{#plus} #pi^{#minus}) (GeV/#it{c}^{2});entries", {HistType::kTH1F, {{100, 1.7, 2}}});
+    registry.add("hMassV0", "K0^{s} candidates;inv. mass (#pi^{#plus}#pi^{#minus}) (GeV/#it{c}^{2});entries", {HistType::kTH1F, {{100, 0.35, 0.65}}});
+    registry.add("hPtD", "D^{#minus} candidates;D^{#minus} candidate #it{p}_{T} (GeV/#it{c});entries", {HistType::kTH1F, {{100, 0., 10.}}});
+    registry.add("hPtV0", "V0 candidates;V0 candidate #it{p}_{T} (GeV/#it{c});entries", {HistType::kTH1F, {{100, 0., 10.}}});
+    registry.add("hMassDs1", "Ds1 candidates;Ds1 candidate #inv. mass (GeV/#it{c}^{2});entries", {HistType::kTH1F, {{100, 2., 3.}}});
```
fgrosa, line 102:
```suggestion
    registry.add("hMassDs1ToDstarK0S", "Ds1 candidates;Ds1 candidate #it{M}(D^{*+}K0S) (GeV/#it{c}^{2});entries", {HistType::kTH1F, {{100, 2., 3.}}});
```
We could also consider to put the delta mass M(D K0S) - M(D) (this is what we typically do for resonances)

PWGHF/D2H/TableProducer/dataCreatorDV0Reduced.cxx
```diff
@@ -0,0 +1,184 @@
...
+
+  // O2DatabasePDG service
+  Service<o2::framework::O2DatabasePDG> pdg;
+
+  bool isHfCandResoConfigFilled = false;
+
+  using TracksPidAll = soa::Join<aod::pidTPCFullEl, aod::pidTPCFullMu, aod::pidTPCFullPi, aod::pidTPCFullKa, aod::pidTPCFullPr,
+                                 aod::pidTOFFullEl, aod::pidTOFFullMu, aod::pidTOFFullPi, aod::pidTOFFullKa, aod::pidTOFFullPr>;
+  using TracksPIDWithSel = soa::Join<aod::TracksWCovDcaExtra, TracksPidAll, aod::TrackSelection>;
+  using TracksPIDWithSelAndMc = soa::Join<TracksPIDWithSel, aod::McTrackLabels>;
```
fgrosa, line 74: It looks to me that the PID information is not used in this task, so I would avoid to subscribe to the tables

PWGHF/D2H/TableProducer/dataCreatorDV0Reduced.cxx
```diff
@@ -0,0 +1,184 @@
...
+  using CandsDFiltered = soa::Filtered<soa::Join<aod::HfCand3Prong, aod::HfSelDplusToPiKPi>>;
```
fgrosa, line 75: The Ds1 decays into D*, while the Ds2* decays into D+. Depending which channel you want to implement here I would be consistent (either you rename the histogram Ds2Star, or you use the D* instead of the D+ if you aim at implementing the Ds1, as I would suggest indeed)

PWGHF/D2H/TableProducer/dataCreatorDV0Reduced.cxx
```diff
@@ -0,0 +1,184 @@
...
+      }
+    } // candsD loop
+    registry.fill(HIST("hEvents"), 1 + Event::Processed);
+    if (!fillHfReducedCollision) {
+      registry.fill(HIST("hEvents"), 1 + Event::NoDV0Selected);
+      return;
+    }
+    registry.fill(HIST("hEvents"), 1 + Event::DV0Selected);
+    // fill collision table if it contains a DPi pair a minima
+    lfReducedCollision(collision.posX(), collision.posY(), collision.posZ());
```
fgrosa, line 160: Instead of using the one of LF that might change in time and create conflicts with our code, I would use the `HfRedCollisions` here. I will then in a second PR separate the cov matrix elements in a joinable table used only for the B mesons.

fgrosa (changes requested):
Hi @Luca610 thank you very much for the implementation of this task! 
The main comment I have is that you are pairing D+ and K0S to build Ds1, but the Ds1 decays into D* and K0S. So, depending whether you want to switch to the Ds2* with this first implementation or to stick with the Ds1 (it's the first channel I would look at, since it's narrower and hence easier) you can consider to modify the code. If you want to move to Ds1 though, I would not through away the implementation of the D+K0S channel, but I would simply add the D*+ to have both Ds1 and Ds2* already.

Luca610:
Thanks for the comments, Fabrizio. Indeed, I confused the decay channels of the 2 resonances I was planning to study.

PWGHF/D2H/TableProducer/dataCreatorDV0Reduced.cxx
Luca610: I started using it because I didn't need the covariance and magnetic field and I was encountering issues of table duplication when I tried to define a table with the same columns in ReducedDataModel.h, But I can use HfRedCollison for sure

PWGHF/D2H/TableProducer/dataCreatorDV0Reduced.cxx
Luca610: Thank you!

PWGHF/D2H/TableProducer/dataCreatorDV0Reduced.cxx
Luca610: Ok I will use the mass difference

PWGHF/D2H/TableProducer/dataCreatorDV0Reduced.cxx
Luca610: You're right I'm removing it

PWGHF/D2H/DataModel/ReducedDataModel.h
Luca610: Ok!

PWGHF/D2H/TableProducer/dataCreatorDV0Reduced.cxx
Luca610: Thanks, I'm reverting to HfRedCollisions for the time being

PWGHF/D2H/DataModel/ReducedDataModel.h
```diff
@@ -370,6 +370,42 @@ DECLARE_SOA_TABLE(HfCandBpConfigs, "AOD", "HFCANDBPCONFIG", //! Table with confi
...
+DECLARE_SOA_TABLE(HfRedVzeros, "AOD", "HFREDVZERO", //! Table with V0 candidate information for resonances reduced workflow
+                  o2::soa::Index<>,
+                  hf_track_index_reduced::Prong0Id, hf_track_index_reduced::Prong1Id,
+                  hf_track_index_reduced::HfRedCollisionId,
+                  hf_cand::XSecondaryVertex, hf_cand::YSecondaryVertex, hf_cand::ZSecondaryVertex,
+                  hf_reso_cand_reduced::InvMassK0s, hf_reso_cand_reduced::InvMassLambda, hf_reso_cand_reduced::InvMassAntiLambda,
+                  hf_reso_cand_reduced::Px,
+                  hf_reso_cand_reduced::Py,
+                  hf_reso_cand_reduced::Pz,
+                  hf_reso_cand_reduced::V0Type);
```
fgrosa, line 397: Maybe we can add a couple of topological variables to allow a bit of cut variation later (such as cosPA and radius)

PWGHF/D2H/TableProducer/dataCreatorDV0Reduced.cxx
```diff
@@ -0,0 +1,393 @@
...
+  NoDV0Selected,
+  DV0Selected,
+  kNEvent
+};
+
+enum DecayChannel : uint8_t {
+  Ds1ToDstarK0s = 0,
+  DsStar2ToDplusK0s,
+  Xc3055ToDplusLambda
+};
```
fgrosa, line 58: As discussed offline we could be less granular here and do together D+K0S and D+Lambda (since the loop on the V0 is anyway the same)

PWGHF/D2H/TableProducer/dataCreatorDV0Reduced.cxx
```diff
@@ -0,0 +1,393 @@
...
+
+enum V0_type : int8_t {
```
fgrosa, line 60: Not sure if this is needed anymore given the comment above, but in case you can modify it to 
```suggestion
enum TypeV0 : uint8_t {
```

PWGHF/D2H/TableProducer/dataCreatorDV0Reduced.cxx
```diff
@@ -0,0 +1,393 @@
...
+  K0s = 0,
+  Lambda
+};
+
+enum D_type : int8_t {
+  Dplus = 1,
+  Dstar
+};
```
fgrosa, line 68:
```suggestion
enum TypeD : uint8_t {
  Dplus = 1,
  Dstar
};
```

PWGHF/D2H/TableProducer/dataCreatorDV0Reduced.cxx
```diff
@@ -0,0 +1,393 @@
...
+
+    ccdb->setURL(url.value);
+    ccdb->setCaching(true);
+    ccdb->setLocalObjectValidityChecking();
+    ccdb->setCreatedNotAfter(std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now().time_since_epoch()).count());
+    ccdbApi.init(url);
+    lut = o2::base::MatLayerCylSet::rectifyPtrFromFile(ccdb->get<o2::base::MatLayerCylSet>("GLO/Param/MatLUT"));
+  }
+
+  template <uint8_t DecayChannel, typename C>
```
fgrosa, line 145:
```suggestion
  template <uint8_t DecayChannel, typename CCands>
```

PWGHF/D2H/TableProducer/dataCreatorDV0Reduced.cxx
```diff
@@ -0,0 +1,393 @@
...
+  void runDataCreation(aod::Collision const& collision,
+                       C const& candsD,
```
fgrosa, line 147:
```suggestion
                       CCands const& candsD,
```

PWGHF/D2H/TableProducer/dataCreatorDV0Reduced.cxx
```diff
@@ -0,0 +1,393 @@
...
+      // initialize variables depending on decay channel
+      bool fillHfCandD = false;
+      float invMassD;
+      float massD;
+      float massV0;
+      std::array<float, 3> pVecD;
+      std::array<float, 3> secondaryVertexD;
+      std::array<int, 3> prongIdsD;
+      int8_t v0_type;
+      int8_t d_type;
```
fgrosa, line 180:
```suggestion
      uint8_t typeV0;
      uint8_t typeD;
```
Also below

PWGHF/D2H/TableProducer/dataCreatorDV0Reduced.cxx
```diff
@@ -0,0 +1,393 @@
...
+            secondaryVertexD[0] = candD.xSecondaryVertex();
+            secondaryVertexD[1] = candD.ySecondaryVertex();
+            secondaryVertexD[2] = candD.zSecondaryVertex();
+            prongIdsD[0] = candD.prong0Id();
+            prongIdsD[1] = candD.prong1Id();
+            prongIdsD[2] = candD.prong2Id();
+            v0_type = candD.sign() * V0_type::Lambda;
+            d_type = candD.sign() * D_type::Dplus;
+            massV0 = MassLambda0;
+            break;
```
fgrosa, line 228: These two should go together based on the comment above; you can check if the V0 is compatible with a Lambda then you compute the D+Lambda mass, if it is compatible with the K0S you compute the D+K0S mass (depending on the cuts, it might be both)

For this we also need to distinguish K0 and Lambda (see below)

PWGHF/D2H/TableProducer/dataCreatorDV0Reduced.cxx
```diff
@@ -0,0 +1,393 @@
...
+        // LOG(info) << "V0 p after propagation: " << pVecV0[0] << "," << pVecV0[1] << "," << pVecV0[2];
+
+        // fill histos
+        registry.fill(HIST("hPtV0"), ptV0);
+        registry.fill(HIST("hV0_type"), v0_type);
+
+        switch (DecayChannel) {
+          case DecayChannel::Ds1ToDstarK0s:
+            registry.fill(HIST("hMassK0s"), v0.mK0Short());
+            registry.fill(HIST("hMassDs1"), sqrt(invMass2DV0) - invMassD);
```
fgrosa, line 263: The `deltaMass` does not work like this, because the mass resolution of the D does not cancel. You should either use the invariant mass of the D in the DV0 invariant mass, or here you simply rescale by the PDG mass of the D

PWGHF/D2H/DataModel/ReducedDataModel.h
```diff
@@ -370,6 +370,42 @@ DECLARE_SOA_TABLE(HfCandBpConfigs, "AOD", "HFCANDBPCONFIG", //! Table with confi
...
+namespace hf_reso_cand_reduced
+{
+DECLARE_SOA_COLUMN(InvMass, invMass, float);                     //! Invariant mass of 3 prong candidate in GeV/c2
+DECLARE_SOA_COLUMN(InvMassK0s, invMassK0s, float);               //! Invariant mass of V0 candidate in GeV/c2, under K0s mass assumption
+DECLARE_SOA_COLUMN(InvMassLambda, invMassLambda, float);         //! Invariant mass of V0 candidate in GeV/c2, under Lambda mass assumption
+DECLARE_SOA_COLUMN(InvMassAntiLambda, invMassAntiLambda, float); //! Invariant mass of V0 candidate in GeV/c2, under AntiLambda mass assumption
+DECLARE_SOA_COLUMN(Px, px, float);                               //! Momentum of V0/3 prong candidate in GeV/c
+DECLARE_SOA_COLUMN(Py, py, float);
+DECLARE_SOA_COLUMN(Pz, pz, float);
+DECLARE_SOA_COLUMN(V0Type, v0Type, int); //! Integer with V0 mass hypotesis used: 0 = K0s, +1 = Lambda, -1 = AntiLambda
```
fgrosa, line 384: I would not set the exclusively, since (depending on your cuts), you might have a K0S which is also compatible with a (Anti)Lambda (if we set proper selections is rare, but not impossible)
```suggestion
DECLARE_SOA_COLUMN(V0Type, v0Type, uint8_t); //! Bitmap with mass hypothesis of the V0
```

PWGHF/D2H/TableProducer/dataCreatorDV0Reduced.cxx
```diff
@@ -0,0 +1,393 @@
...
+          std::array<float, 3> pVecV0Orig = {v0.px(), v0.py(), v0.pz()};
+          std::array<float, 3> posVecV0 = {v0.x(), v0.y(), v0.z()};
+          gpu::gpustd::array<float, 2> dcaInfo;
+          auto trackParK0 = o2::track::TrackPar(posVecV0, pVecV0Orig, 0, true);
+          trackParK0.setPID(o2::track::PID::K0);
+          trackParK0.setAbsCharge(0);
+          o2::base::Propagator::Instance()->propagateToDCABxByBz({collision.posX(), collision.posY(), collision.posZ()}, trackParK0, 2.f, matCorr, &dcaInfo);
+          getPxPyPz(trackParK0, pVecV0);
+        }
+
```
fgrosa, line 250: For each V0 candidate we need to test whether it is compatible with a K0S or with a Lambda/AntiLambda. 
To do that, I would add a configurable (or two) with the mass window of compatibility with lambda or K0S. We can also apply other basic cuts in a second iteration (such as cuts on the V0 topology and PID of the daughters)

Then, depending on the compatibility with one or more species, you should set the bit in the `typeV0` (you can use `SETBIT(typeV0, type)`, where `type` can be and integer (`0`, `1`, `2`) or better the element of an `enum` (if you for example extend the existing one adding `AntiLambda`)

PWGHF/D2H/TableProducer/dataCreatorDV0Reduced.cxx
```diff
@@ -0,0 +1,393 @@
...
+    hfCollisionCounter(collisions.tableSize());
+
+    for (const auto& collision : collisions) {
+      auto thisCollId = collision.globalIndex();
+      auto candsDThisColl = candsDplus.sliceBy(candsDplusPerCollision, thisCollId);
+      auto V0sThisColl = V0s.sliceBy(candsV0PerCollision, thisCollId);
+      runDataCreation<DecayChannel::Xc3055ToDplusLambda, CandsDplusFiltered>(collision, candsDThisColl, V0sThisColl, bcs);
+    }
+  }
+  PROCESS_SWITCH(HfDataCreatorDV0Reduced, processXc, "Process Xc to Dplus Lambda without MC info and without ML info", false);
```
fgrosa, line 387: This one should be merged with the DsStar2 (and maybe you can rename it `DplusV0`)

PWGHF/D2H/TableProducer/dataCreatorDV0Reduced.cxx
```diff
@@ -0,0 +1,393 @@
...
+    for (const auto& collision : collisions) {
+      auto thisCollId = collision.globalIndex();
+      auto candsDThisColl = candsDplus.sliceBy(candsDplusPerCollision, thisCollId);
+      auto V0sThisColl = V0s.sliceBy(candsV0PerCollision, thisCollId);
+      runDataCreation<DecayChannel::DsStar2ToDplusK0s, CandsDplusFiltered>(collision, candsDThisColl, V0sThisColl, bcs);
+    }
+  }
+  PROCESS_SWITCH(HfDataCreatorDV0Reduced, processDsStar2, "Process DsStar2 to Dplus K0s without MC info and without ML info", true);
+
+  void processDs1(aod::Collisions const& collisions,
```
fgrosa, line 353: This you can maybe rename `DstarV0`

PWGHF/D2H/DataModel/ReducedDataModel.h
```diff
@@ -370,6 +370,42 @@ DECLARE_SOA_TABLE(HfCandBpConfigs, "AOD", "HFCANDBPCONFIG", //! Table with confi
...
+DECLARE_SOA_COLUMN(DType, dType, int);   //! Integer with selected D candidate type: 1 = Dplus, -1 = Dminus, 2 = DstarPlus, -2 = DstarMinus
```
fgrosa, line 385: No need to be an `int`, it can be a `int8_t`
```suggestion
DECLARE_SOA_COLUMN(DType, dType, int8_t);   //! Integer with selected D candidate type: 1 = Dplus, -1 = Dminus, 2 = DstarPlus, -2 = DstarMinus
```

fgrosa (changes requested):
Hi @Luca610 thanks a lot for considering my comments! As discussed offline I think that it would be good to put together all the V0s to have the possibility to have multiple channels included. Moreover, I believe that it would be good to add some selection on the V0, especially the one on the mass that would allow us to decide whether a V0 is accepted as a K0, Lambda, or AntiLambda. 
Below you can see some detailed comment

PWGHF/TableProducer/candidateCreator3Prong.cxx
```diff
@@ -142,6 +142,9 @@ struct HfCandidateCreator3Prong {
       auto trackParVar1 = getTrackParCov(track1);
       auto trackParVar2 = getTrackParCov(track2);
 
+      // get sign of candidate
+      int8_t sign = track0.sign();
+
```
vkucera, line 147: Is this really needed?

fgrosa: In general I think it would be good to have the sign stored, to avoid the need to access the track table when not needed. This was the case in the previous version by @Luca610, now that he added the subscription to the track table for other reasons, I think that we can revert it and take it from the track of `prong0`. We can then discuss whether to add it in a separated PR. @Luca610 could you please change? Thanks!

Luca610: Sure

fgrosa (approved):

PWGHF/D2H/DataModel/ReducedDataModel.h
```diff
@@ -370,6 +370,48 @@ DECLARE_SOA_TABLE(HfCandBpConfigs, "AOD", "HFCANDBPCONFIG", //! Table with confi
...
+{
+DECLARE_SOA_COLUMN(InvMass, invMass, float);                     //! Invariant mass of 3 prong candidate in GeV/c2
+DECLARE_SOA_COLUMN(InvMassK0s, invMassK0s, float);               //! Invariant mass of V0 candidate in GeV/c2, under K0s mass assumption
+DECLARE_SOA_COLUMN(InvMassLambda, invMassLambda, float);         //! Invariant mass of V0 candidate in GeV/c2, under Lambda mass assumption
+DECLARE_SOA_COLUMN(InvMassAntiLambda, invMassAntiLambda, float); //! Invariant mass of V0 candidate in GeV/c2, under AntiLambda mass assumption
+DECLARE_SOA_COLUMN(Px, px, float);                               //! Momentum of V0/3 prong candidate in GeV/c
+DECLARE_SOA_COLUMN(Py, py, float);
+DECLARE_SOA_COLUMN(Pz, pz, float);
+DECLARE_SOA_COLUMN(CosP, cosp, float);       //! Cos PA of V0 candidate
+DECLARE_SOA_COLUMN(DCA, dca, float);         //! DCA of V0 candidate
```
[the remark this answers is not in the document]
vkucera: Btw, `cosPA` or `cpa` would be better, since `cosP` is missing the `A`.

PWGHF/D2H/TableProducer/CMakeLists.txt
```diff
@@ -45,3 +45,8 @@ o2physics_add_dpl_workflow(data-creator-dplus-pi-reduced
                     PUBLIC_LINK_LIBRARIES O2Physics::AnalysisCore O2::DCAFitter
                     COMPONENT_NAME Analysis)
 
+o2physics_add_dpl_workflow(data-creator-d-v0-reduced
+                    SOURCES dataCreatorDV0Reduced.cxx
+                    PUBLIC_LINK_LIBRARIES O2Physics::AnalysisCore O2::DCAFitter
```
vkucera, line 50: Where do you use `DCAFitter`?

PWGHF/D2H/TableProducer/dataCreatorDV0Reduced.cxx
```diff
@@ -0,0 +1,424 @@
...
+  // Produces AOD tables to store track information
+  Produces<aod::HfRedCollision> hfReducedCollision;  // Defined in PWGLF/DataModel/LFStrangenessTables.h
+  Produces<aod::HfOrigColCounts> hfCollisionCounter; // Defined in PWGHF/D2H/DataModel/ReducedDataModel.h
+
+  Produces<aod::HfRedVzeros> hfCandV0;   // Defined in PWGHF/D2H/DataModel/ReducedDataModel.h
+  Produces<aod::HfRed3PrNoTrks> hfCandD; // Defined in PWGHF/D2H/DataModel/ReducedDataModel.h
+
+  // CCDB configuration
+  o2::ccdb::CcdbApi ccdbApi;
+  Service<o2::ccdb::BasicCCDBManager> ccdb;
```
vkucera, line 81: Please respect the conventions on the order of members.

PWGHF/D2H/TableProducer/dataCreatorDV0Reduced.cxx
```diff
@@ -0,0 +1,424 @@
...
+    ccdbApi.init(url);
+    lut = o2::base::MatLayerCylSet::rectifyPtrFromFile(ccdb->get<o2::base::MatLayerCylSet>("GLO/Param/MatLUT"));
+  }
+
+  /// Basic selection of V0 candidates
+  /// \param v0 is the v0 candidate
+  /// \param collision is the current collision
+  /// \return a bitmap with mass hypotesis if passes all cuts
+  template <typename V0, typename Coll, typename Tr>
+  inline uint8_t isSelectedV0(const V0& v0, const Coll& collision, const std::array<Tr, 2>& dauTracks, const std::array<int, 3>& dDaughtersIDs)
```
vkucera, line 161: `isSelectedV0` implies that the return type is `bool`. Better to name it something like `getSelectionMapV0`.

PWGHF/D2H/TableProducer/dataCreatorDV0Reduced.cxx
```diff
@@ -0,0 +1,424 @@
...
+
+      // Loop on V0 candidates
+      for (const auto& v0 : V0s) {
+        auto posTrack = v0.posTrack_as<BigTracksPID>();
+        auto negTrack = v0.negTrack_as<BigTracksPID>();
+        // Apply selsection
+        v0_type = isSelectedV0(v0, collision, std::array{posTrack, negTrack}, prongIdsD);
+        if (v0_type == 0) {
+          continue;
+        }
```
vkucera, line 281:
```suggestion
        v0Type = isSelectedV0(v0, collision, std::array{posTrack, negTrack}, prongIdsD);
        if (v0Type == 0) {
          continue;
        }
```
Underscores are reserved for namespaces.

PWGHF/D2H/TableProducer/dataCreatorDV0Reduced.cxx
```diff
@@ -0,0 +1,424 @@
...
+          std::array<float, 3> pVecV0Orig = {v0.px(), v0.py(), v0.pz()};
+          std::array<float, 3> posVecV0 = {v0.x(), v0.y(), v0.z()};
+          gpu::gpustd::array<float, 2> dcaInfo;
+          auto trackParK0 = o2::track::TrackPar(posVecV0, pVecV0Orig, 0, true);
+          trackParK0.setPID(o2::track::PID::K0);
+          trackParK0.setAbsCharge(0);
+          o2::base::Propagator::Instance()->propagateToDCABxByBz({collision.posX(), collision.posY(), collision.posZ()}, trackParK0, 2.f, matCorr, &dcaInfo);
+          getPxPyPz(trackParK0, pVecV0);
+        }
+        float ptV0 = sqrt(pVecV0[0] * pVecV0[0] + pVecV0[1] * pVecV0[1]); // fill histos
```
vkucera, line 294:
```suggestion
        float ptV0 = RecoDecay::pt(pVecV0); // fill histos
```

vkucera: Btw, state the namespace you use: `std::sqrt`.

PWGHF/D2H/TableProducer/dataCreatorDV0Reduced.cxx
```diff
@@ -0,0 +1,424 @@
...
+    registry.add("hMassLambda", "Lambda candidates;inv. mass (p #pi^{#minus}) (GeV/#it{c}^{2});entries", {HistType::kTH1F, {{100, 1.05, 1.35}}});
+    registry.add("hPtDplus", "D^{#minus} candidates;D^{#minus} candidate #it{p}_{T} (GeV/#it{c});entries", {HistType::kTH1F, {{100, 0., 10.}}});
+    registry.add("hPtDstar", "D^{*} candidates;D^{*} candidate #it{p}_{T} (GeV/#it{c});entries", {HistType::kTH1F, {{100, 0., 10.}}});
+    registry.add("hPtV0", "V0 candidates;V0 candidate #it{p}_{T} (GeV/#it{c});entries", {HistType::kTH1F, {{100, 0., 10.}}});
+    registry.add("hMassDs1", "Ds1 candidates;m_{Ds1} - m_{D^{*}} (GeV/#it{c}^{2});entries", {HistType::kTH1F, {{100, 0.45, 0.7}}});
+    registry.add("hMassDsStar2", "Ds^{*}2 candidates; Ds^{*}2 - m_{D^{#plus}} (GeV/#it{c}^{2});entries", {HistType::kTH1F, {{100, 0.4, 1}}});
+    registry.add("hMassXcRes", "XcRes candidates; XcRes - m_{D^{#plus}} (GeV/#it{c}^{2});entries", {HistType::kTH1F, {{100, 1., 1.4}}});
+    registry.add("hV0_type", "V0 selection flag", {HistType::kTH1F, {{8, -0.5, 7.5}}});
+    registry.add("hD_type", "D selection flag", {HistType::kTH1F, {{5, -2.5, 2.5}}});
```
vkucera, line 146:
```suggestion
    registry.add("hV0Type", "V0 selection flag", {HistType::kTH1F, {{8, -0.5, 7.5}}});
    registry.add("hDType", "D selection flag", {HistType::kTH1F, {{5, -2.5, 2.5}}});
```

Luca610:
@vkucera,  Thank you for your comments.  Being my first analysis I am still unfamiliar with the conventions, sorry. 
 I am implementing all comments an I will include them in my next PR! Thank you again!

PR #4784, opened 2024-02-20
PWGHF: Adding TPCposTPCneg EP and SP resolutions

stefanopolitano:
This PR contains:

- The histograms for the resolution of TPCpos-TPCneg couple for event-plane and scalar-product
- The FT0a/FT0c--FT0m combinations were removed since the subsystems  are correlated


fgrosa (commented):
Hi @stefanopolitano thanks this looks good to me! I have however another question: I just noticed that you have the `V0M` in the subdetectors, however it is not clear to me what it is. In Run3 we only have `V0A`, `T0A`, and `T0C`. Does it mean that `V0M` coincides with `V0A`? Moreover, if it is meant for Run2, we also do not have the possibility to correlate it with `T0` detectors, since the `T0` was not there.

stefanopolitano:
Hi @fgrosa, thanks for having a look. I think it is just a typo in the task: it should be V0A. I'll fix it. 
Also, it is meant to fairly compare the resolution computed in the qvector task with the one we obtain here

fgrosa (approved):

PR #4786, opened 2024-02-20
EventFiltering/PWGHF: Add min cut on Xi for Xi+bachelor triggers

fgrosa:
@ZFederica could you please cross check that this is ok? Thanks!


ZFederica:
@fgrosa it looks good to me, thank you very much!

alibuild (approved):

PR #4789, opened 2024-02-20
PWGHF: Add multiplicity and BDT output scores to Ds task

fchinu:
This PR adds multiplicity and machine learning information to the output of the Ds task into a THnSparse.

This is done by adding templates to the original functions in the task, which allow to use/don't use ML and multiplicity tables. It is possible to choose which one to use by turning on the relative process function. It is also possible to choose which ML classes to save and which detector to use to evaluate the multiplicity percentiles via configurables.

In order to loop on the selected candidates which may or may not have ML information, several partitions have been defined (as `Partition<CandDsMcReco>` does not have ML information, but `Partition<CandDsMcRecoWithMl>` has it, for example)

Also, refactoring of part of the code was made, making it easier to modify and maintain.

### DIFF
--- a/PWGHF/D2H/Tasks/taskDs.cxx
+++ b/PWGHF/D2H/Tasks/taskDs.cxx
@@ -30,31 +30,62 @@ using namespace o2::analysis;
 using namespace o2::framework;
 using namespace o2::framework::expressions;
 
+enum finalState { KKPi = 0,
+                  PiKK };
+
+enum centEstimator { FT0A = 0,
+                     FT0C,
+                     FT0M,
+                     FV0A,
+                     FDDM,
+                     NTracksPV,
+                     kNMaxCentEstimators};
+
 /// Ds± analysis task
 struct HfTaskDs {
   Configurable<int> decayChannel{"decayChannel", 1, "Switch between decay channels: 1 for Ds/Dplus->PhiPi->KKpi, 2 for Ds/Dplus->K0*K->KKPi"};
   Configurable<bool> fillDplusMc{"fillDplusMc", true, "Switch to fill Dplus MC information"};
+  Configurable<int> centDetector{"centDetector", 2, "Detector for multiplicity estimation (FT0A: 0, FT0C: 1, FT0M: 2, FV0A: 3, FDDM:4, NTracksPV: 5)"};
   Configurable<int> selectionFlagDs{"selectionFlagDs", 7, "Selection Flag for Ds"};
+  Configurable<std::vector<int>> classMl{"classMl", {0, 2, 3}, "Indexes of BDT scores to be stored. Three indexes max."};
   Configurable<double> yCandGenMax{"yCandGenMax", 0.5, "max. gen particle rapidity"};
   Configurable<double> yCandRecoMax{"yCandRecoMax", 0.8, "max. cand. rapidity"};
   Configurable<std::vector<double>> binsPt{"binsPt", std::vector<double>{hf_cuts_ds_to_k_k_pi::vecBinsPt}, "pT bin limits"};
 
   HfHelper hfHelper;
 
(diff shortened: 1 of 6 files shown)


PWGHF/D2H/Tasks/taskDs.cxx
```diff
@@ -73,9 +104,34 @@ struct HfTaskDs {
...
+
+    if (centDetector < 0 || centDetector >= centEstimator::kNMaxCentEstimators)
+      LOG(warning) << "Centrality estimator not valid. Please choose between FT0A, FT0C, FT0M, FV0A, FDDM, NTracksPV. Fallback to FT0M";
+
     auto vbins = (std::vector<double>)binsPt;
+    AxisSpec ptbins = {vbins, "#it{p}_{T} (GeV/#it{c})"};
     AxisSpec ybins = {100, -5., 5, "#it{y}"};
-    registry.add("hMass", "3-prong candidates;inv. mass (KK#pi) (GeV/#it{c}^{2});entries", {HistType::kTH2F, {{600, 1.67, 2.27}, {vbins, "#it{p}_{T} (GeV/#it{c})"}}});
+    AxisSpec massbins = {600, 1.67, 2.27, "inv. mass (KK#pi) (GeV/#it{c}^{2})"};
+    AxisSpec BDTbins = {100, 0., 1., "BDT output"};
```
fcatalan92, line 123:
```suggestion
    AxisSpec MLbins = {100, 0., 1., "ML output"};
```

PWGHF/D2H/Tasks/taskDs.cxx
```diff
@@ -170,11 +250,29 @@ struct HfTaskDs {
...
   /// \param candidate is candidate
-  template <typename T1>
+  template <bool useMl, bool fillCent, typename T1>
   void fillHistoKKPi(const T1& candidate)
   {
     auto pt = candidate.pt();
-    registry.fill(HIST("hMass"), hfHelper.invMassDsToKKPi(candidate), pt);
+    if constexpr (useMl) {
+      std::vector<float> outputMl = {-999., -999., -999.};
+      if (candidate.mlProbDsToKKPi().size() > 0) {
```
fcatalan92, line 259: This should be always true if the right selection is applied in the filter using `doprocessMcWithMl`. So in principle there is no need to check

PWGHF/D2H/Tasks/taskDs.cxx
```diff
@@ -184,11 +282,29 @@ struct HfTaskDs {
...
   /// \param candidate is candidate
-  template <typename T1>
+  template <bool useMl, bool fillCent, typename T1>
   void fillHistoPiKK(const T1& candidate)
   {
     auto pt = candidate.pt();
-    registry.fill(HIST("hMass"), hfHelper.invMassDsToPiKK(candidate), pt);
+    if constexpr (useMl) {
+      std::vector<float> outputMl = {-999., -999., -999.};
+      if (candidate.mlProbDsToKKPi().size() > 0) {
```
fcatalan92, line 291: Same consideration as above

PWGHF/D2H/Tasks/taskDs.cxx
```diff
@@ -286,83 +402,101 @@ struct HfTaskDs {
     return;
   }
 
-  void process(CandDsData const& candidates)
+  /// Fill MC histograms
+  /// \param candidate is candidate
+  /// \param mcParticles are particles with MC information
+  /// \param isDplus is true if the candidate is a Dplus
```
fcatalan92, line 408:
```suggestion
  /// \param isDplus true to fill Dplus->KKpi candidates
```

fcatalan92 (changes requested):
Hi @fchinu, thanks a lot for the developments and refactoring! I have a few comments. The relevant one is that on the process function (and option) names. IMHO if their pattern is different it is easier to not make mistakes when configuring the task.

fcatalan92 (approved):

PR #4793, opened 2024-02-21
PWGHF: add centrality selection in candidate creators.

mfaggin:
@fcatalan92 @fgrosa here the implementation for the 3 prong. If you agree with it, I can go ahead also with the rest.

### DIFF
--- a/PWGHF/TableProducer/candidateCreator2Prong.cxx
+++ b/PWGHF/TableProducer/candidateCreator2Prong.cxx
@@ -43,18 +43,9 @@
 using namespace o2;
 using namespace o2::analysis;
 using namespace o2::aod::hf_cand_2prong;
+using namespace o2::aod::hf_collision_centrality;
 using namespace o2::constants::physics;
 using namespace o2::framework;
-
-// centrality selection estimators
-enum CentralityEstimator {
-  None = 0,
-  FT0A,
-  FT0C,
-  FT0M,
-  FV0A,
-  NCentralityEstimators
-};
 
 /// Reconstruction of heavy-flavour 2-prong decay candidates
 struct HfCandidateCreator2Prong {
@@ -108,18 +99,14 @@ struct HfCandidateCreator2Prong {
 
   void init(InitContext const&)
   {
-    std::array<bool, 10> doprocessDF{doprocessPvRefitWithDCAFitterN, doprocessNoPvRefitWithDCAFitterN,
-                                     doprocessPvRefitWithDCAFitterNCentFT0A, doprocessNoPvRefitWithDCAFitterNCentFT0A,
-                                     doprocessPvRefitWithDCAFitterNCentFT0C, doprocessNoPvRefitWithDCAFitterNCentFT0C,
-                                     doprocessPvRefitWithDCAFitterNCentFT0M, doprocessNoPvRefitWithDCAFitterNCentFT0M,
-                                     doprocessPvRefitWithDCAFitterNCentFV0A, doprocessNoPvRefitWithDCAFitterNCentFV0A};
-    std::array<bool, 10> doprocessKF{doprocessPvRefitWithKFParticle, doprocessNoPvRefitWithKFParticle,
-                                     doprocessPvRefitWithKFParticleCentFT0A, doprocessNoPvRefitWithKFParticleCentFT0A,
-                                     doprocessPvRefitWithKFParticleCentFT0C, doprocessNoPvRefitWithKFParticleCentFT0C,
-                                     doprocessPvRefitWithKFParticleCentFT0M, doprocessNoPvRefitWithKFParticleCentFT0M,
-                                     doprocessPvRefitWithKFParticleCentFV0A, doprocessNoPvRefitWithKFParticleCentFV0A};
+    std::array<bool, 6> doprocessDF{doprocessPvRefitWithDCAFitterN, doprocessNoPvRefitWithDCAFitterN,
+                                    doprocessPvRefitWithDCAFitterNCentFT0C, doprocessNoPvRefitWithDCAFitterNCentFT0C,
+                                    doprocessPvRefitWithDCAFitterNCentFT0M, doprocessNoPvRefitWithDCAFitterNCentFT0M};
+    std::array<bool, 6> doprocessKF{doprocessPvRefitWithKFParticle, doprocessNoPvRefitWithKFParticle,
+                                    doprocessPvRefitWithKFParticleCentFT0C, doprocessNoPvRefitWithKFParticleCentFT0C,
+                                    doprocessPvRefitWithKFParticleCentFT0M, doprocessNoPvRefitWithKFParticleCentFT0M};
     if ((std::accumulate(doprocessDF.begin(), doprocessDF.end(), 0) + std::accumulate(doprocessKF.begin(), doprocessKF.end(), 0)) != 1) {
-      LOGP(fatal, "One and only process function can be enabled at a time.");
+      LOGP(fatal, "One and only one process function must be enabled at a time.");
     }
     if (std::accumulate(doprocessDF.begin(), doprocessDF.end(), 0) == 1) {
       hVertexerType->Fill(aod::hf_cand::VertexerType::DCAFitter);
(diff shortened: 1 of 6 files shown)


PWGHF/TableProducer/candidateCreator3Prong.cxx
```diff
@@ -34,10 +34,23 @@ using namespace o2::aod::hf_cand_3prong;
...
 
+// centrality selection estimators
+enum CentralityEstimator {
+  None = 0,
+  FT0A,
+  FT0C,
+  FT0M,
+  FV0A,
+  NCentralityEstimators
+};
```
fcatalan92, line 45: In a few other implementations in the HF code also other estimators are used, see https://github.com/AliceO2Group/O2Physics/blob/master/PWGHF/D2H/Tasks/taskDs.cxx#L36-L42 and https://github.com/AliceO2Group/O2Physics/blob/master/PWGHF/TableProducer/treeCreatorLcToPKPi.cxx#L232-L237 . Maybe is it worth to add more estimators here too?

PWGHF/TableProducer/candidateCreator3Prong.cxx
```diff
@@ -79,8 +92,13 @@ struct HfCandidateCreator3Prong {
...
   {
-    if (doprocessPvRefit && doprocessNoPvRefit) {
-      LOGP(fatal, "Only one process function between processPvRefit and processNoPvRefit can be enabled at a time.");
+    std::array<bool, 10> processes = {doprocessPvRefit, doprocessNoPvRefit,
+                                      doprocessPvRefitCentFT0A, doprocessNoPvRefitCentFT0A,
+                                      doprocessPvRefitCentFT0C, doprocessNoPvRefitCentFT0C,
+                                      doprocessPvRefitCentFT0M, doprocessNoPvRefitCentFT0M,
+                                      doprocessPvRefitCentFV0A, doprocessNoPvRefitCentFV0A};
+    if (std::accumulate(processes.begin(), processes.end(), 0) != 1) {
+      LOGP(fatal, "One and only process function can be enabled at a time.");
```
fcatalan92, line 101:
```suggestion
      LOGP(fatal, "One and only one process function must be enabled at a time.");
```

PWGHF/TableProducer/candidateCreator3Prong.cxx
```diff
@@ -235,25 +274,135 @@ struct HfCandidateCreator3Prong {
...
+
+  /// @brief process function w/ PV refit and w/ centrality selection on FT0A
+  void processPvRefitCentFT0A(soa::Join<aod::Collisions, aod::CentFT0As> const& collisions,
+                              soa::Join<aod::Hf3Prongs, aod::HfPvRefit3Prong> const& rowsTrackIndexProng3,
+                              aod::TracksWCovExtra const& tracks,
+                              aod::BCsWithTimestamps const& bcWithTimeStamps)
+  {
+    runCreator3Prong</*doPvRefit*/ true, CentralityEstimator::FT0A>(collisions, rowsTrackIndexProng3, tracks, bcWithTimeStamps);
   }
+  PROCESS_SWITCH(HfCandidateCreator3Prong, processPvRefitCentFT0A, "Run candidate creator with PV refit and w/ centrality selection on FT0A", false);
```
fcatalan92, line 317: I would not define different process functions for the different centrality estimators. To keep the entropy low, an approach similar to the v2 task can be adopted https://github.com/AliceO2Group/O2Physics/blob/d2f45f354bdaf98d9ff0f8c35da0a7d37dca95d3/PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx#L72
where we keep all the possible centrality estimators together. Maybe we could keep the options with and w/o centrality, so that the centrality task can be removed from the workflow if not needed.
What we lose is the possibility to disentangle the centrality estimator at compile time as it is done at the moment. But the performance impact should not be large.

fcatalan92 (commented):
Hi @mfaggin, thanks a lot for the developments! I had a look only at the 3-prong creator for the moment. I would change the approach to handle the different centrality estimators, with the aim of simplifying the code and the configuration of the task. See the detailed comment below.

PWGHF/TableProducer/candidateCreator3Prong.cxx
fgrosa, line 45: Same as above

fgrosa (changes requested):
Hi @mfaggin thanks a lot for this important development!
I just have one comment to avoid the re-definition of the same `enum`, for the rest it looks good to me.

PWGHF/TableProducer/candidateCreator3Prong.cxx
fgrosa: I would prefer to have it split as implemented by @mfaggin, to avoid the increase of the memory when not needed. In fact, the centrality tables are produced only if consumed by later workflows. One intermediate solution could be to reduce the number of estimators here, and keep only `FT0C` and `FT0M` for the moment.

PWGHF/TableProducer/candidateCreator3Prong.cxx
mfaggin: At the moment I would stay with the few ones that we agreed for the candidate creators

PWGHF/TableProducer/candidateCreator3Prong.cxx
fcatalan92: Fine for me

PWGHF/TableProducer/candidateCreator2Prong.cxx
```diff
@@ -46,11 +46,24 @@ using namespace o2::aod::hf_cand_2prong;
...
 
+// centrality selection estimators
+enum CentralityEstimator {
+  None = 0,
+  FT0A,
+  FT0C,
+  FT0M,
+  FV0A,
+  NCentralityEstimators
+};
```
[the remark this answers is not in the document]
mfaggin: fixed also in the `trackIndexSkimCreator`

PWGHF/TableProducer/candidateCreator2Prong.cxx
```diff
@@ -248,17 +272,37 @@ struct HfCandidateCreator2Prong {
...
+          centrality = collision.centFT0A();
+        } else if constexpr (centEstimator == CentralityEstimator::FT0C) {
+          centrality = collision.centFT0C();
+        } else if constexpr (centEstimator == CentralityEstimator::FT0M) {
+          centrality = collision.centFT0M();
+        } else if constexpr (centEstimator == CentralityEstimator::FV0A) {
+          centrality = collision.centFV0A();
+        } else {
+          LOGP(fatal, "Centrality estimator not set!");
+        }
```
fcatalan92, line 298: `FT0A` and `FV0A` can be removed

fcatalan92 (changes requested):

PWGHF/TableProducer/candidateCreator2Prong.cxx
mfaggin: yes, it's done in the last commit

mfaggin: uh, I forgot it for the KF particle case, I'll solve it thanks

mfaggin:
informing also @fcolamar about this development

mfaggin:
@fgrosa @fcatalan92 @fcolamar I have just opened the PR.
For the moment, I left out the `candidateCreatorToXiPi.cxx`, since @ZFederica is still doing some tests for her derived data (and this is not yet super urgent for Pb-Pb).
The other candidateCreators are here.

fgrosa:
> @fgrosa @fcatalan92 @fcolamar I have just opened the PR. For the moment, I left out the `candidateCreatorToXiPi.cxx`, since @ZFederica is still doing some tests for her derived data (and this is not yet super urgent for Pb-Pb). The other candidateCreators are here.

Hi @mfaggin it looks good to me, thanks!

mfaggin:
@fcatalan92 I guess you have to approve the PR in order it to be merged

fcatalan92 (approved):

PR #4795, opened 2024-02-21
PWGHF: fix bug in cascade derived data

ZFederica:
In the struct that produces cascade-bachelor derived data, remove the continue statements and replace them with a boolean checking on the status of the cascade-bachelor pair

### DIFF
--- a/PWGHF/TableProducer/trackIndexSkimCreator.cxx
+++ b/PWGHF/TableProducer/trackIndexSkimCreator.cxx
@@ -3431,6 +3431,9 @@ struct HfTrackIndexSkimCreatorLfCascades {
     df3.setWeightedFinalPCA(useWeightedFinalPCA);
 
     uint8_t hfFlag = 0;
+    bool isGoogForXi2Prong = true;
+    bool isGoogForOmega2Prong = true;
+    bool isGoogForXi3Prong = true;
 
     for (const auto& collision : collisions) {
 
@@ -3510,6 +3513,8 @@ struct HfTrackIndexSkimCreatorLfCascades {
         for (auto trackIdPion1 = groupedBachTrackIndices.begin(); trackIdPion1 != groupedBachTrackIndices.end(); ++trackIdPion1) {
 
           hfFlag = 0;
+          isGoogForXi2Prong = true;
+          isGoogForOmega2Prong = true;
 
           auto trackPion1 = trackIdPion1.track_as<aod::TracksWCovDca>();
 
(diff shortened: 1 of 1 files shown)


PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -3431,6 +3431,9 @@ struct HfTrackIndexSkimCreatorLfCascades {
     df3.setWeightedFinalPCA(useWeightedFinalPCA);
 
     uint8_t hfFlag = 0;
+    bool isGoogForXi2Prong = true;
+    bool isGoogForOmega2Prong = true;
+    bool isGoogForXi3Prong = true;
```
fgrosa, line 3436: There is no need of the 3-prong one; whenever a 3-prong candidate is not selected you can continue, because in any case the triplet will not be used anymore.

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -3538,13 +3543,13 @@ struct HfTrackIndexSkimCreatorLfCascades {
...
-            continue;
+            isGoogForXi2Prong = false;
           }
-          if (fillHistograms) {
+          if (isGoogForXi2Prong && fillHistograms) {
             registry.fill(HIST("hFitterStatusXi2Prong"), 0);
           }
 
-          if (nVtxFrom2ProngFitterXiHyp > 0) {
+          if (isGoogForXi2Prong && nVtxFrom2ProngFitterXiHyp > 0) {
```
fgrosa, line 3552: `nVtxFrom2ProngFitterXiHyp` should be `0` if the `DCAFitter` throws the exception, so the `isGoodForXi2Prong` is redundant (it is sufficient to remove the `continue` above)

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -3578,13 +3583,13 @@ struct HfTrackIndexSkimCreatorLfCascades {
...
-            continue;
+            isGoogForOmega2Prong = false;
           }
-          if (fillHistograms) {
+          if (isGoogForOmega2Prong && fillHistograms) {
             registry.fill(HIST("hFitterStatusOmega2Prong"), 0);
           }
 
-          if (nVtxFrom2ProngFitterOmegaHyp > 0) {
+          if (isGoogForOmega2Prong && nVtxFrom2ProngFitterOmegaHyp > 0) {
```
fgrosa, line 3592: same here, `isGoogForOmega2Prong` is not needed

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -3658,13 +3664,13 @@ struct HfTrackIndexSkimCreatorLfCascades {
                 if (fillHistograms) {
                   registry.fill(HIST("hFitterStatusXi3Prong"), 1);
                 }
-                continue;
+                isGoogForXi3Prong = false;
```
fgrosa, line 3667: Here you can keep the continue

fgrosa (changes requested):
Hi @ZFederica thanks! I have a few comments, see below

fgrosa (approved):

PR #4797, opened 2024-02-21
PWGHF: Add efficiency task for Xi Pi decays

ZFederica:
Add dedicated task to compute efficiency for xi pi decay channel. The task has been implemented following the flow of the existing task taskMcEfficiency.cxx


ZFederica:
Hello @fgrosa, here is the task we discussed few days ago

PWGHF/Tasks/taskMcEfficiencyToXiPi.cxx
```diff
@@ -0,0 +1,435 @@
...
+    bool checkNClsIts = false;
+    if (track.itsNCls() > nClustersItsMin) {
+      checkNClsIts = true;
+    }
+    if (checkItsTrk && checkNClsIts) {
+      return true;
+    } else {
+      return false;
+    }
+  }
```
vkucera, line 122: This is equivalent to
```c++
track.isQualityTrackITS() && track.itsNCls() > nClustersItsMin
```

PWGHF/Tasks/taskMcEfficiencyToXiPi.cxx
```diff
@@ -0,0 +1,435 @@
...
+    }
+    if (track.itsNCls() > nClustersItsMin) {
+      checkNClsIts = true;
+    }
+    if (checkGlbTrk && checkNClsTpc && checkNClsIts) {
+      return true;
+    } else {
+      return false;
+    }
+  }
```
vkucera, line 107: This is equivalent to
```c++
track.isGlobalTrackWoDCA() && track.tpcNClsFound() > nClustersTpcMin && track.itsNCls() > nClustersItsMin
```

PWGHF/Tasks/taskMcEfficiencyToXiPi.cxx
```diff
@@ -0,0 +1,435 @@
...
+                kHFStepTracked,
+                kHFStepTrackedCuts,
+                kHFStepTrackedSelected,
+                kHFNSteps };
+
+  enum TrackableStep { kTrackableAll = 0,
+                       kTrackableITS,
+                       kTrackableTPC,
+                       kTrackableITSTPC,
+                       kNTrackableSteps };
```
vkucera, line 69: Remove the `k` prefixes. Add documentation

PWGHF/Tasks/taskMcEfficiencyToXiPi.cxx
```diff
@@ -0,0 +1,435 @@
...
+    for (const auto& mcParticle : genParticles) {
+
+      // check if I am treating the desired charm baryon
+      if (mcParticle.pdgCode() != pdgCode) {
+        continue;
+      }
+      /// check if the charm baryon devays to the desired channel
+      if (mcParticle.flagMcMatchGen() != decayFlagGen) {
+        continue;
+      }
```
vkucera, line 235: Are you not missing `std::abs`?

PWGHF/Tasks/taskMcEfficiencyToXiPi.cxx
```diff
@@ -0,0 +1,435 @@
...
+    auto mass = pdg->Mass(pdgCode);
+
+    int decayFlagGen = 0;
+    if (pdgCode == Pdg::kXiC0) {
+      decayFlagGen = 4;
+    } else if (pdgCode == Pdg::kOmegaC0) {
+      decayFlagGen = 2;
+    } else {
+      LOGP(fatal, "Not implemented for PDG code: ", pdgCode);
+    }
```
vkucera, line 224: Use `enum` values.

PWGHF/Tasks/taskMcEfficiencyToXiPi.cxx
```diff
@@ -0,0 +1,435 @@
...
+          selectedItsTpc[track.mcParticleId()] = true;
+        }
+        if (checkTrackItsTrk(track)) {
+          selectedIts[track.mcParticleId()] = true;
+        }
+      }
+    }
+
+    auto mass = pdg->Mass(pdgCode);
+
```
vkucera, line 216: You don't need the service. You know the PDG code at compile time so you can use the mass constants.

PWGHF/Tasks/taskMcEfficiencyToXiPi.cxx
```diff
@@ -0,0 +1,435 @@
...
+  enum TrackableStep { kTrackableAll = 0,
+                       kTrackableITS,
+                       kTrackableTPC,
+                       kTrackableITSTPC,
+                       kNTrackableSteps };
+
+  using TracksWithSelectionMC = soa::Join<aod::Tracks, aod::TracksExtra, aod::McTrackLabels, aod::TrackSelection>;
+  using candidateInfo = soa::Join<aod::HfCandToXiPi, aod::HfToXiPiMCRec, aod::HfSelToXiPi>;
+  using genLevelInfo = soa::Join<aod::McParticles, aod::HfToXiPiMCGen>;
```
vkucera, line 73:
```suggestion
  using TracksWithSelectionMC = soa::Join<aod::Tracks, aod::TracksExtra, aod::McTrackLabels, aod::TrackSelection>;
  using CandidateInfo = soa::Join<aod::HfCandToXiPi, aod::HfToXiPiMCRec, aod::HfSelToXiPi>;
  using ParticleInfo = soa::Join<aod::McParticles, aod::HfToXiPiMCGen>;
```

PWGHF/Tasks/taskMcEfficiencyToXiPi.cxx
```diff
@@ -0,0 +1,435 @@
...
+
+  Configurable<int> nClustersTpcMin{"nClustersTpcMin", 70, "Minimum number of TPC clusters requirement for pion <-- charm baryon"};
+  Configurable<int> nClustersItsMin{"nClustersItsMin", 3, "Minimum number of ITS clusters requirement for pion <- charm baryon"};
+
+  ConfigurableAxis axisPt{"axisPt", {200, 0, 20}, "pT axis"};
+  ConfigurableAxis axisMass{"axisMass", {900, 2.1, 3}, "m_inv axis"};
+
+  Service<o2::framework::O2DatabasePDG> pdg;
+  HfHelper hfHelper;
+
```
vkucera, line 52:
```suggestion
```

PWGHF/Tasks/taskMcEfficiencyToXiPi.cxx
```diff
@@ -0,0 +1,435 @@
...
+using namespace o2;
+using namespace o2::analysis;
+using namespace o2::constants::physics;
+using namespace o2::framework;
+using namespace o2::framework::expressions;
+
+struct HfTaskMcEfficiencyToXiPi {
+
+  Configurable<double> rapidityCharmBaryonMax{"rapidityCharmBaryonMax", 0.5, "Max absolute value of rapidity for charm baryon"};
```
vkucera, line 38:
```suggestion
struct HfTaskMcEfficiencyToXiPi {
  Configurable<double> rapidityCharmBaryonMax{"rapidityCharmBaryonMax", 0.5, "Max absolute value of rapidity for charm baryon"};
```

PWGHF/Tasks/taskMcEfficiencyToXiPi.cxx
```diff
@@ -0,0 +1,435 @@
...
+#include "CommonConstants/PhysicsConstants.h"
+#include "Framework/AnalysisTask.h"
+#include "Framework/HistogramRegistry.h"
+#include "Framework/O2DatabasePDGPlugin.h"
+#include "Framework/runDataProcessing.h"
+
+#include "Common/Core/RecoDecay.h"
+#include "Common/DataModel/TrackSelectionTables.h"
+
+#include "PWGHF/Core/HfHelper.h"
```
vkucera, line 26:
```suggestion
```

PWGHF/Tasks/taskMcEfficiencyToXiPi.cxx
```diff
@@ -0,0 +1,435 @@
...
+
+/// \file taskMcEfficiencyToXiPi.cxx
+/// \brief Task to analyse the MC efficiency for toXiPi charm baryon decay ("found / trackable")
+///
+/// \author Federica Zanone, Heidelberg University
+
+#include "CommonConstants/PhysicsConstants.h"
+#include "Framework/AnalysisTask.h"
+#include "Framework/HistogramRegistry.h"
+#include "Framework/O2DatabasePDGPlugin.h"
```
vkucera, line 20:
```suggestion
```

PWGHF/Tasks/taskMcEfficiencyToXiPi.cxx
```diff
@@ -0,0 +1,435 @@
...
+using namespace o2::framework;
+using namespace o2::framework::expressions;
+
+struct HfTaskMcEfficiencyToXiPi {
+
+  Configurable<double> rapidityCharmBaryonMax{"rapidityCharmBaryonMax", 0.5, "Max absolute value of rapidity for charm baryon"};
+  Configurable<double> mcLfAcceptanceEta{"mcLfAcceptanceEta", 1.0, "Max absolute value of eta for LF daughters"};
+  Configurable<double> mcPionFromCascadeAcceptancePt{"mcPionFromCascadeAcceptancePt", 0.2, "Min value of pt for pion <-- cascade"};
+  Configurable<double> mcPionFromCharmBaryonAcceptanceEta{"mcPionFromCharmBaryonAcceptanceEta", 0.8, "Max absolute value of eta for pion <-- charm baryon"};
+  Configurable<double> mcPionFromCharmBaryonAcceptancePt{"mcPionFromCharmBaryonAcceptancePt", 0.5, "Min value of pt for pion <-- charm baryon"};
```
vkucera, line 42: Use `float`

vkucera, line 42: Start the name with what quantity the object represents. I don't think the `mc` prefixes are needed.

PWGHF/Tasks/taskMcEfficiencyToXiPi.cxx
```diff
@@ -0,0 +1,435 @@
...
+    hCandidates->GetAxis(2)->SetTitle("Collision matched");
+    hCandidates->GetAxis(3)->SetTitle("Charm hadron origin");
+
+    auto hTrackablePtEta = registry.add<StepTHn>("hTrackablePtEta", "Prongs kinematics at different steps", {HistType::kStepTHnF, {{200, 0, 20}, {400, -2, 2}}, kNTrackableSteps});
+    hTrackablePtEta->GetAxis(0)->SetTitle("#it{p}_{T} (GeV/#it{c})");
+    hTrackablePtEta->GetAxis(1)->SetTitle("#eta");
+  }
+
+  template <typename T>
+  inline bool checkTrackGlbTrk(T track)
```
vkucera, line 91: You are copying the track.

PWGHF/Tasks/taskMcEfficiencyToXiPi.cxx
```diff
@@ -0,0 +1,435 @@
...
+    }
+    if (checkGlbTrk && checkNClsTpc && checkNClsIts) {
+      return true;
+    } else {
+      return false;
+    }
+  }
+
+  template <typename T>
+  inline bool checkTrackItsTrk(T track)
```
vkucera, line 110: You are copying the track.

PWGHF/Tasks/taskMcEfficiencyToXiPi.cxx
```diff
@@ -0,0 +1,435 @@
...
+      if (!selectedPid) {
+        continue;
+      }
+    }
+  }
+
+  // candidates -> join candidateCreator, candidateCreator McRec and candidateSelector tables
+  // genParticles -> join aod::McParticles and candidateCreator McGen tables
+  template <typename T1, typename T2>
+  void candidateFullLoop(T1& candidates, T2& genParticles, TracksWithSelectionMC const& tracks, aod::McCollisionLabels const& colls, int pdgCode)
```
vkucera, line 183:
```suggestion
  void candidateFullLoop(T1 const& candidates, T2 const& genParticles, TracksWithSelectionMC const& tracks, aod::McCollisionLabels const& colls, int pdgCode)
```

PWGHF/Tasks/taskMcEfficiencyToXiPi.cxx
```diff
@@ -0,0 +1,435 @@
...
+      }
+      auto cascade = genParticles.rawIteratorAt(mcParticle.daughtersIds().front());
+      auto pion = genParticles.rawIteratorAt(mcParticle.daughtersIds().back());
+      if (std::abs(cascade.pdgCode()) != kXiMinus && std::abs(cascade.pdgCode()) == kPiPlus) {
+        auto save = cascade;
+        cascade = pion;
+        pion = save;
+        auto saveIdxCharmBaryon = cascId;
+        cascId = pionId;
+        pionId = saveIdxCharmBaryon;
```
vkucera, line 262: You can use `std::swap` instead.

PWGHF/Tasks/taskMcEfficiencyToXiPi.cxx
```diff
@@ -0,0 +1,435 @@
...
+      }
+      auto lambda = genParticles.rawIteratorAt(cascade.daughtersIds().front());
+      auto pionFromCascade = genParticles.rawIteratorAt(cascade.daughtersIds().back());
+      if (std::abs(lambda.pdgCode()) != kLambda0 && std::abs(lambda.pdgCode()) == kPiPlus) {
+        auto saveCasc = lambda;
+        lambda = pionFromCascade;
+        pionFromCascade = saveCasc;
+        auto saveIdxCasc = lambdaId;
+        lambdaId = pionFromCascadeId;
+        pionFromCascadeId = saveIdxCasc;
```
vkucera, line 293: Same

PWGHF/Tasks/taskMcEfficiencyToXiPi.cxx
```diff
@@ -0,0 +1,435 @@
...
+      }
+      auto proton = genParticles.rawIteratorAt(lambda.daughtersIds().front());
+      auto pionFromLambda = genParticles.rawIteratorAt(lambda.daughtersIds().back());
+      if (std::abs(proton.pdgCode()) != kProton && std::abs(proton.pdgCode()) == kPiPlus) {
+        auto saveLambda = proton;
+        proton = pionFromLambda;
+        pionFromLambda = saveLambda;
+        auto saveIdxLambda = protonId;
+        protonId = pionFromLambdaId;
+        pionFromLambdaId = saveIdxLambda;
```
vkucera, line 316: Same

PWGHF/Tasks/taskMcEfficiencyToXiPi.cxx
```diff
@@ -0,0 +1,435 @@
...
+
+  void processOmegac0(candidateInfo const& candidates,
+                      genLevelInfo const& genParticles,
+                      TracksWithSelectionMC const& tracks,
+                      aod::McCollisionLabels const& colls)
+  {
+    int pdgCode = Pdg::kOmegaC0;
+    candidateFullLoop(candidates, genParticles, tracks, colls, pdgCode);
+  }
+  PROCESS_SWITCH(HfTaskMcEfficiencyToXiPi, processOmegac0, "Process MC for Omegac0 signal", false);
```
vkucera, line 428: 1. Why can you not have a single process function?
2. Do you want to allow running both `processXic0` and `processOmegac0` together?

vkucera (changes requested):
I don't see how the changes in other files are related to the new task. If they are not, please remove them from the PR.

ZFederica:
> I don't see how the changes in other files are related to the new task. If they are not, please remove them from the PR.

They are needed: the information about whether the collision was matched or not was not yet stored in the MC table produced by the candidate creator, but I was trying to retrieve this information in the new task therefore the build failed.

PWGHF/Tasks/taskMcEfficiencyToXiPi.cxx
ZFederica: I would leave the k to be consistent with the stye of the other task for efficiency calculation, see https://github.com/AliceO2Group/O2Physics/blob/b2decb815c1cd6d434b428b74582297fba0192a7/PWGHF/Tasks/taskMcEfficiency.cxx#L51-L67

I added documentation.

PWGHF/Tasks/taskMcEfficiencyToXiPi.cxx
ZFederica: Yes, thanks!

PWGHF/Tasks/taskMcEfficiencyToXiPi.cxx
ZFederica: Could you please clarify how you would implement this?
This function is called only with two possible pdgCodes (depending on the process function, see processXic0 and processOmegac0) and there are no other alternatives, moreover there will be no extension of the possible "cases"

PWGHF/Tasks/taskMcEfficiencyToXiPi.cxx
ZFederica: I do not want to run processXic0 and processOmegac0 together, I added a check in the init function

PWGHF/Tasks/taskMcEfficiencyToXiPi.cxx
vkucera: I meant that the entire function is equivalent to that line. You can directly return the value of the condition instead returning `true` or `false` after evaluating the condition. Plus the function is used only once which makes it unnecessary. You can just use that one-line condition directly instead of calling the function.

PWGHF/Tasks/taskMcEfficiencyToXiPi.cxx
vkucera: OK, then it was probably overlooked in the other task before. The conventions forbid the prefixes in general.

PWGHF/Tasks/taskMcEfficiencyToXiPi.cxx
vkucera: You compare the value of `decayFlagGen` with the value of the column `hf_cand_toxipi::FlagMcMatchGen`, which is filled with `BIT(OmegaczeroToXiPi)` or `BIT(XiczeroToXiPi)`. So you should use those values for `decayFlagGen` instead of explicit `4` and `2` which might stop working if the `enum` change.

PWGHF/Tasks/taskMcEfficiencyToXiPi.cxx
vkucera: The only difference between the two process functions is the value of `pdgCode` which you can control with a configurable. I don't see any benefit of having two process functions and the `candidateFullLoop` function.

ZFederica: 1) I have implemented different process functions because I followed "the style" of the other existing task for MC efficiency calculation (a process function for each set of particles to take into account)

2) I would like to have two separate loops because the first one is based only the reconstructed information, the second on the first one + on generated information. To better highlight this difference I changed the name of the first funtion

PWGHF/Tasks/taskMcEfficiencyToXiPi.cxx
```diff
@@ -0,0 +1,399 @@
...
+        LOGP(fatal, "Invalid numbers of daughters for cascade {}: {}", cascade.globalIndex(), cascade.daughtersIds().size());
+      }
+      auto lambdaId = cascade.daughtersIds()[0];
+      auto pionFromCascadeId = cascade.daughtersIds()[1];
+      if (lambdaId < 0 || pionFromCascadeId < 0) {
+        continue;
+      }
+      auto lambda = genParticles.rawIteratorAt(cascade.daughtersIds().front());
+      auto pionFromCascade = genParticles.rawIteratorAt(cascade.daughtersIds().back());
+      if (std::abs(lambda.pdgCode()) != kLambda0 && std::abs(lambda.pdgCode()) == kPiPlus) {
```
vkucera, line 261: `lambda.pdgCode()) != kLambda0` is redundant

PWGHF/Tasks/taskMcEfficiencyToXiPi.cxx
```diff
@@ -0,0 +1,399 @@
...
+        LOGP(fatal, "Invalid numbers of daughters for lambda {}: {}", lambda.globalIndex(), lambda.daughtersIds().size());
+      }
+      auto protonId = lambda.daughtersIds()[0];
+      auto pionFromLambdaId = lambda.daughtersIds()[1];
+      if (protonId < 0 || pionFromLambdaId < 0) {
+        continue;
+      }
+      auto proton = genParticles.rawIteratorAt(lambda.daughtersIds().front());
+      auto pionFromLambda = genParticles.rawIteratorAt(lambda.daughtersIds().back());
+      if (std::abs(proton.pdgCode()) != kProton && std::abs(proton.pdgCode()) == kPiPlus) {
```
vkucera, line 280: `std::abs(proton.pdgCode()) != kProton` is redundant

PWGHF/Tasks/taskMcEfficiencyToXiPi.cxx
```diff
@@ -0,0 +1,399 @@
...
+
+    double mass = 0.;
+    int decayFlagGen = 0;
+
+    if (pdgCode == Pdg::kXiC0) {
+      decayFlagGen = 4;
+      mass = MassXiC0;
+    } else if (pdgCode == Pdg::kOmegaC0) {
+      decayFlagGen = 2;
+      mass = MassOmegaC0;
```
vkucera, line 199: Use `enum`

PWGHF/Tasks/taskMcEfficiencyToXiPi.cxx
vkucera: 1. In the other efficiency task, the process functions are needed because of different combinations of tables they consume. Your task consumes the same tables in both cases, so splitting it into two process functions is just a code duplication.
2. I am not discouraging your from having `candidateRecLoop`. I was talking about `candidateFullLoop` which is called once in each `process` function with the same parameters, making it pointless.

PWGHF/Tasks/taskMcEfficiencyToXiPi.cxx
```diff
@@ -0,0 +1,399 @@
...
+    hCandidates->GetAxis(2)->SetTitle("Collision matched");
+    hCandidates->GetAxis(3)->SetTitle("Charm hadron origin");
+
+    auto hTrackablePtEta = registry.add<StepTHn>("hTrackablePtEta", "Prongs kinematics at different steps", {HistType::kStepTHnF, {{200, 0, 20}, {400, -2, 2}}, kNTrackableSteps});
+    hTrackablePtEta->GetAxis(0)->SetTitle("#it{p}_{T} (GeV/#it{c})");
+    hTrackablePtEta->GetAxis(1)->SetTitle("#eta");
+
+    if (doprocessOmegac0 == true && doprocessXic0 == true) {
+      LOGF(fatal, "Can't enable processOmegac0 and processXic0 at the same time. Please choose one.");
+    }
```
vkucera, line 85: - This should be the first check to do.
- ` == true` is redundant.
- You are not covering the case of no enabled function.

vkucera (changes requested):

PWGHF/Tasks/taskMcEfficiencyToXiPi.cxx
ZFederica: Okay, I switched to 1 process function only

PWGHF/Tasks/taskMcEfficiencyToXiPi.cxx
ZFederica: In the end I removed this check, I switched to 1 process function only where the configurables allow to select which particle to match (and I moved a check on the configurable values inside the process function itself)

vkucera: Perfect! Thanks! Now it is much simpler and clearer.

PWGHF/Tasks/taskMcEfficiencyToXiPi.cxx
vkucera: Perfect! Thanks!

vkucera (approved):
Thanks @ZFederica for implementing my suggestions. It looks good to me now.

PR #4812, opened 2024-02-21
PWGDQ/EM add recalculation of DCA param of TrackToCollAssoc tracks

feisenhu:
This PR Draft has not been tested yet!!!

Adapt `track selection` to perform a recalculation of the DCA parameter.

Following the example of the `TableReader_withAssoc` using
`VarManager::FillTrackCollision<TTrackFillMap>(track, collision);`
and the `void runMuonSelection()` function.

So far no flag has been introduced to select this recalculation.

@rbailhac @dsekihat @hscheid  I would appreciate if you could have a first look at this.
I'm not sure if this implementation is the way to go, but I tried to follow other examples.

### DIFF
--- a/PWGDQ/Core/VarManager.h
+++ b/PWGDQ/Core/VarManager.h
@@ -689,7 +689,7 @@ class VarManager : public TObject
   static void FillTrack(T const& track, float* values = nullptr);
   template <uint32_t fillMap, typename T, typename C>
   static void FillTrackCollision(T const& track, C const& collision, float* values = nullptr);
-  template <uint32_t fillMap, typename T, typename C,typename M>
+  template <uint32_t fillMap, typename T, typename C, typename M>
   static void FillTrackCollisionMatCorr(T const& track, C const& collision, M const& materialCorr, float* values = nullptr);
   template <typename U, typename T>
   static void FillTrackMC(const U& mcStack, T const& track, float* values = nullptr);
@@ -1667,8 +1667,6 @@ void VarManager::FillTrackCollisionMatCorr(T const& track, C const& collision, M
     // trackPar.propagateParamToDCA({collision.posX(), collision.posY(), collision.posZ()}, fgMagField, &dca);
     o2::base::Propagator::Instance()->propagateToDCABxByBz({collision.posX(), collision.posY(), collision.posZ()}, trackPar, 2.f, materialCorr, &dca);
     getPxPyPz(trackPar, pVec);
-    cout << __LINE__ << " COUT LINE::VarManager track.pt() = " << track.pt() << endl;
-    cout << __LINE__ << " COUT LINE::VarManager trackPar.getPt() = " << trackPar.getPt() << endl << endl;
 
     values[kTrackDCAxy] = dca[0];
     values[kTrackDCAz] = dca[1];
--- a/PWGDQ/Core/VarManager.h
+++ b/PWGDQ/Core/VarManager.h
@@ -689,8 +689,8 @@ class VarManager : public TObject
   static void FillTrack(T const& track, float* values = nullptr);
   template <uint32_t fillMap, typename T, typename C>
   static void FillTrackCollision(T const& track, C const& collision, float* values = nullptr);
-  template <uint32_t fillMap, typename T, typename C, typename M>
-  static void FillTrackCollisionMatCorr(T const& track, C const& collision, M const& materialCorr, float* values = nullptr);
+  template <uint32_t fillMap, typename T, typename C, typename M, typename P>
+  static void FillTrackCollisionMatCorr(T const& track, C const& collision, M const& materialCorr, P const& propagator, float* values = nullptr);
   template <typename U, typename T>
   static void FillTrackMC(const U& mcStack, T const& track, float* values = nullptr);
   template <uint32_t fillMap, typename T1, typename T2, typename C>
@@ -1654,8 +1654,8 @@ void VarManager::FillTrackCollision(T const& track, C const& collision, float* v
   }
 }
 
-template <uint32_t fillMap, typename T, typename C, typename M>
-void VarManager::FillTrackCollisionMatCorr(T const& track, C const& collision, M const& materialCorr, float* values)
+template <uint32_t fillMap, typename T, typename C, typename M, typename P>
+void VarManager::FillTrackCollisionMatCorr(T const& track, C const& collision, M const& materialCorr, P const& propagator, float* values)
 {
   if (!values) {
     values = fgValues;
--- a/PWGDQ/Core/VarManager.h
+++ b/PWGDQ/Core/VarManager.h
@@ -1665,7 +1665,7 @@ void VarManager::FillTrackCollisionMatCorr(T const& track, C const& collision, M
     std::array<float, 2> dca{1e10f, 1e10f};
     std::array<float, 3> pVec = {track.px(), track.py(), track.pz()};
     // trackPar.propagateParamToDCA({collision.posX(), collision.posY(), collision.posZ()}, fgMagField, &dca);
-    o2::base::Propagator::Instance()->propagateToDCABxByBz({collision.posX(), collision.posY(), collision.posZ()}, trackPar, 2.f, materialCorr, &dca);
+    propagator->propagateToDCABxByBz({collision.posX(), collision.posY(), collision.posZ()}, trackPar, 2.f, materialCorr, &dca);
     getPxPyPz(trackPar, pVec);
 
     values[kTrackDCAxy] = dca[0];
--- a/PWGDQ/Core/VarManager.h
+++ b/PWGDQ/Core/VarManager.h
@@ -689,6 +689,8 @@ class VarManager : public TObject
   static void FillTrack(T const& track, float* values = nullptr);
   template <uint32_t fillMap, typename T, typename C>
   static void FillTrackCollision(T const& track, C const& collision, float* values = nullptr);
+  template <uint32_t fillMap, typename T, typename C,typename M>
+  static void FillTrackCollisionMatCorr(T const& track, C const& collision, M const& materialCorr, float* values = nullptr);
   template <typename U, typename T>
   static void FillTrackMC(const U& mcStack, T const& track, float* values = nullptr);
   template <uint32_t fillMap, typename T1, typename T2, typename C>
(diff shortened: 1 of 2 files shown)


dsekihat:
Hi Florian, Good works. You should also modify FillTrackCollision, because dca and momentum are re-calculated there.

dsekihat:
I am sorry, 2 more comments to check.
1. please make sure that magnetic field is properly loaded.
2. please make sure variables after calling FillTrackCollision are not overwritten somewhere else.

rbailhac:
Hi Florian, 
I have a stupid question, did you try to run it ? I am not sure it will work actually due for example to ambiguous tracks. I would be not surprised that the join tables have not the same size here:
using MyBarrelTracksSelected = soa::Join<aod::Tracks, aod::TracksExtra, aod::TracksDCA,
                                         aod::pidTPCFullEl, aod::pidTPCFullPi,
                                         aod::pidTPCFullKa, aod::pidTPCFullPr,
                                         aod::pidTOFFullEl, aod::pidTOFFullPi,
                                         aod::pidTOFFullKa, aod::pidTOFFullPr,
                                         aod::DQBarrelTrackCuts>;

I would have do like the muons, join a table with the trackassociation and the DQBarrelTrackCuts first.

I think similar is done also in the tablereader_withAssoc.

Otherwise the points of Daiki are also relevant.

Best regards,

Raphaelle

rbailhac:
Just for book keeping: it may compile but I think it is still not ok:

You should remove the DQBarrelTrackCuts from  using MyBarrelTracksSelected = soa::Join<aod::Tracks, aod::TracksExtra, aod::TracksDCA,
                                         aod::pidTPCFullEl, aod::pidTPCFullPi,
                                         aod::pidTPCFullKa, aod::pidTPCFullPr,
                                         aod::pidTOFFullEl, aod::pidTOFFullPi,
                                         aod::pidTOFFullKa, aod::pidTOFFullPr,
                                         aod::DQBarrelTrackCuts>;

You should also redefined this:  void processFilterPP(MyEventsSelected const& collisions,
                       aod::BCsWithTimestamps const& bcs,
                       MyBarrelTracksSelected const& tracks,
                       MyMuons const& muons,
                       TrackAssoc const& trackAssocs, MyMuonsAssocSelected const& muonAssocs)

to  void processFilterPP(MyEventsSelected const& collisions,
                       aod::BCsWithTimestamps const& bcs,
                       MyBarrelTracksSelected const& tracks,
                       MyMuons const& muons,
                       MyBarrelTracksAssocSelected const& trackAssocs, MyMuonsAssocSelected const& muonAssocs)

And then the part below still needs modification

 for (auto trackAssoc : barrelAssocs) {
      auto track = trackAssoc.template track_as<TTracks>();
      for (int i = 0; i < fNBarrelCuts; ++i) {
        if (track.isDQBarrelSelected() & (uint32_t(1) << i)) {
          objCountersBarrel[i] += 1;
        }
      }
    }

Look at the muon case some line below...l738 currently

It is important to have a test running before approving to my opinion.

PWGDQ/Core/VarManager.h
```diff
@@ -1593,6 +1595,34 @@ void VarManager::FillTrackCollision(T const& track, C const& collision, float* v
   }
 }
 
+template <uint32_t fillMap, typename T, typename C, typename M>
```
dsekihat, line 1598: How do you set magnetic field to "Propagator::Instance()" here? Could you pass o2::base::Propagator in filter PP task to this function as an argument, please?

feisenhu:
Dear @rbailhac and @dsekihat 
Thank you for having a look.

I tried to address all your comments.

In order to use 
```cpp
o2::base::Propagator::Instance()->propagateToDCABxByBz({collision.posX(), collision.posY(), collision.posZ()}, trackPar, 2.f, materialCorr, &dca);
``` 
instead of 
```cpp
trackPar.propagateParamToDCA({collision.posX(), collision.posY(), collision.posZ()}, fgMagField, &dca);
```

I have introduced a new function in the VarManager, being called `FillTrackCollisionMatCorr`.
In this way I don't change the version of the `FillTrackCollision` which is being used from other tasks like the `tableREader-withAssoc`.

Can you tell me if the definition of 
```cpp
    // material correction for track propagation
    // o2::base::Propagator::MatCorrType matCorr = o2::base::Propagator::MatCorrType::USEMatCorrLUT;
    o2::base::Propagator::MatCorrType noMatCorr = o2::base::Propagator::MatCorrType::USEMatCorrNONE;
```
can be placed within the `FillTrackCollisionMatCorr` or it needs to stay in the filterPPtask?


Further  do I need to modify the if case within the `FillTrackCollision` function?
```cpp
  if constexpr ((fillMap & ReducedTrackBarrel) > 0 || (fillMap & TrackDCA) > 0) {
  ```

**Comment by dsekihat** [comment]
o2::base::Propagator::MatCorrType noMatCorr = o2::base::Propagator::MatCorrType::USEMatCorrNONE; should stay in filterPP.

**Comment by dsekihat** [comment]
Further do I need to modify the if case within the FillTrackCollisionMatCorr  function?
  if constexpr ((fillMap & ReducedTrackBarrel) > 0 || (fillMap & TrackDCA) > 0) {
  --> No. But, please test it first.

**Comment by feisenhu** [comment]
I now have performed a couple of tests and it seems like the recalculation is applied.

As a short overview I have printed several observables within the `FillTrackCollisionMatCorr` function of the `VarManager.h` and checked if they change after the propagation.
```
track.x() = 0.0383853,    trackPar.getX() = 0.0322487, collision.posX() = -0.0333993
track.y() = 0.399825,     trackPar.getY() = 0.399825,  collision.posY() = -0.0146474
track.z() = 0.0714165,    trackPar.getZ() = 0.0696321, collision.posZ() = 0.117118
track.pt() = 0.305525,    trackPar.getPt() = 0.305525
track.eta() = 0.287109,   trackPar.getEta() = 0.287109
track.phi() = 3.06901,    trackPar.getPhi() = 3.06898
track.px() = -0.304721,   pVec[0] = -0.30472
track.py() = 0.0221568,   pVec[1] = 0.0221659
track.pz() = 0.088929,    pVec[2] = 0.088929
track.dcaXY() = 0.384078, dca[0] = 0.382793
track.dcaZ()  = -3.83156, dca[1] = -0.0474855
``` 

As the lines above show, the DCA as well as other parameters are recalculated. The effect on the momentum as well as on `pt, eta, phi` are very minor and are also expected to do not change.


In order to go ahead and include these changes I would like to ask some people from DQ (@vfeuilla @iarsene) to have a look at this PR as well, since this is effecting the whole `DQBarrelTrackSelection`.

If there are some further tests required, please let me know.

**Comment by rbailhac** [comment]
Hi Florian, 

maybe it would be good to separate cases in your printf 

Case 1: where the collision is actually the default collision attached to the track:
 track.collisionId() == collision.globalIndex() and we would not need a priori a recalculation

Case 2: the collision is not the default collision attached to the track (with !). 

Do you see in the first case that the variables of teh track stay the same ?

ps: HF is even asking explicitely for a different collision before recalculating.

Best regards,

Raphaelle

**Comment by dsekihat** [comment]
Hi Florian, How do you propagate the updated px,py,pz to reconstruct invariant mass in filter PP task? You use pVec, but you fill neither fgValues nor values with the updated momentum. Is it fine for you?

**Comment by dsekihat** [comment]
Hi again, do you load magnetic field for the o2::base::Propagator? If it not loaded properly, I am afraid that straight-line propagation is performed, which means px,py,pz does not change, even if collision index changes.

File: PWGDQ/Core/VarManager.h [hunk 2]
```diff
@@ -1593,6 +1595,34 @@ void VarManager::FillTrackCollision(T const& track, C const& collision, float* v
...
+{
+  if (!values) {
+    values = fgValues;
+  }
+  if constexpr ((fillMap & ReducedTrackBarrel) > 0 || (fillMap & TrackDCA) > 0) {
+    auto trackPar = getTrackPar(track);
+    std::array<float, 2> dca{1e10f, 1e10f};
+    std::array<float, 3> pVec = {track.px(), track.py(), track.pz()};
+    // trackPar.propagateParamToDCA({collision.posX(), collision.posY(), collision.posZ()}, fgMagField, &dca);
+    o2::base::Propagator::Instance()->propagateToDCABxByBz({collision.posX(), collision.posY(), collision.posZ()}, trackPar, 2.f, materialCorr, &dca);
```
dsekihat, line 1609: you should not use "o2::base::Propagator::Instance()" here, as you have an argument "propagator". please call "propagator->PropagateToDCABxByBz". otherwise, I am afraid that you propagate tracks without magnetic field.

rbailhac: For the first comment on px,py,pz, the recalculated values are not used indeed. The invariant mass and ptee are also not recalculated (we would need a similar function as VarManager::FillPairPropagateMuon and use it in the pairing). Maybe I would check before your other points.

hscheid: Maybe a comment/question from my side. Do we expect a significant change in the momentum? This should have very little sensitivity to the most inner point and should be mostly determined by the TPC and ITS. Especially with the requirement on the 1st spd hit the lever arm doesn't really change much.
However, we should of course implement this right.

rbailhac: Yes, this was my point....I am not sure for the barrel that we expect anyway a big difference. Therefore I would first check how big is the difference and then decide maybe.

feisenhu:
Dear @rbailhac 
I have double checked the output for correctly assigned tracks and associated tracks.
Your statement ist true. The values for the correct assigned tracks do not change.

@dsekihat  Thank you for pointing out. I updated it to use the propagator.

PWGDQ/Core/VarManager.h
feisenhu: With a small test using 
`o2::base::Propagator::Instance()->propagateToDCABxByBz({collision.posX(), collision.posY(), collision.posZ()}, trackPar, 2.f, materialCorr, &dca);`
and
`propagator->propagateToDCABxByBz({collision.posX(), collision.posY(), collision.posZ()}, trackPar, 2.f, materialCorr, &dca);`
 assuming the magnetic field is propagated like this correctly, I don't see a difference.

iarsene:
Hi Florian, all,
I agree with this update, if we want to remove associations we need the DCA to be recomputed. The implementation seems right to me overall, but please also test how the rejection factor changes. I suggest to have a few AO2D files locally and run the workflow on them before and after. The point is to keep the same physics, but have less events selected right?

On the modification of the momentum, I agree with Raphaelle. Since for our tracks the primary vertex point is not used in constraining the momentum, the momentum of the track should not change much. The very minor changes are due probably to the different position of the PCA. Also, since all collisions are inside the beam pipe, and there is zero material there, I expect also that the very minor changes in the momentum vector components are just due to rotation in the field, so the momentum magnitude should remain exactly the same. Maybe this can be checked as well.

If you want to compute invariant masses with the updated vectors, you can use the same method as for the muon, where the helper function PropagateMuon was made, and its called for both single track and also in the equivalent pair fill function.

Also, if we all agree with this function, we can just replace the older version without the magnetic field, since these are just two similar functions. I would avoid making new alternative process functions for this. However, I don't know how big of an impact in terms of CPU this makes. For the filterPPwithAssoc and for tableMakerWithAssoc this may be critical because it runs over the whole AO2D and this directly impacts our future CPU budgets...
Cheers,
Ionut

feisenhu:
I now have done some tests on a subsample of the `LHC23zs` data. (63 GB, 20 files)

Overall `1.2056e+07` events were inspected,
and when using the same configuration as we used to, the trigger (`LMeeIMR` and `LMeeHMR`) select `422` and `30` events, without a DCA cut. This corresponds to a rejection factor of about `28568` and `401866`

Compared to that, I ran the same test, using a DCA cut (`|DCA|<3cm`):
There the number of events selected by the trigger are `413` and `30` with rejection factors of `29191` and `401866`.

The difference in the rejection factor is only ~ 2% as far as I can tell with this statistics. 
For larger test I recommend to  perform them on the grid.

Cheers Florian

hscheid:
Hi Florian,
Thanks a lot for these checks. Can the change you see be due to a fluctuation in the tests? Just to be sure.
Then on the DCA cut you apply. Is this a DCA_{xyz}?
I agree that a test on a larger sample would be very useful here. However, I am still a bit surprised by the small change, overall the events you have are not that small.
But then looking at the rejection factors themselves: These seem better than what you presented in the last trigger meeting. There you had about 20k without the TtCA and 10k with TtCA. Can you briefly comment on what changed?

Cheers,
Sebastian

feisenhu:
A Priori I don't see a reason why the rejection factor should have changed. 
The latest numbers I have presented in meetings were performed on the `LHC23zs_pass2` data. For the tests here I changed to the newer available `LHC23zs_pass3`.
The DCA cut I use is a rather loose one, selecting tracks with |DCA_{e,xy}| and |DCA_{e,z} < 3 cm.

Total number of events: `1.2056e+07`

Number of triggered events:
**Without TtCA, with DCA cut:** 
LMeeIMR: `250` events triggered, **rej. factor 48224**
LMeeHMR: `17`events triggered, **rej. factor 709176**

**With TtCA, without DCA cut:** 
LMeeIMR: `446`events triggered, **rej. factor 27031**
LMeeHMR: `31`events triggered, **rej. factor 388903**

**With TtCA, with DCA cut:** 
LMeeIMR: `436`events triggered, **rej. factor 27651**
LMeeHMR: `31`events triggered, **rej. factor 388903**

I noticed that in previous tests the ` "timeMargin"` of the  `"track-to-collision-association"` had been set 0, which I corrected for the tests above to 500.

For running without TtCA I used the `processStandardAssoc` instead of `processAssocWithTime`.
The detailed cuts can be found here: [lmee_skimmingtesta_TOF_pionrej](https://github.com/AliceO2Group/O2Physics/blob/00708ffcb8fdc16cadc08c431f70fc2db6c7467b/PWGDQ/Core/CutsLibrary.cxx#L1738) , [lmee_skimmingtesta_TOF_pionrej_noDCA](https://github.com/AliceO2Group/O2Physics/blob/00708ffcb8fdc16cadc08c431f70fc2db6c7467b/PWGDQ/Core/CutsLibrary.cxx#L1746C32-L1746C68)

In some additional tests I have checked the DCA distribution of the reassociated tracks assigned to a new collision and its values before and after the recalculation as well as the distribution after applying the track and DCA cuts. It shows that the recalculation is performed and mainly the DCAz values are corrected.

Side note: If I only test the rejection factor on one file with `717427` events the LMeeIMR selects: 18 events and LMeeHMR: 3. So here I obtain rejection factors of **39857** and **239142**. So the difference to previous tests I've done might me still due to statistics. 

Since there has been also an internal discussion about using a trigger of $1.3~GeV/c^2 < m_{ee}$ I did an additional test for its rejection factor:
LMeeIMR (1.3 < mee GeV/c2): `457` events triggered, **rej. factor 26380**.

Cheers Florian

hscheid:
Hi Florain. 
It looks good from my side.

alibuild (approved):

PR #4813, opened 2024-02-21
DPL Analysis: prevent data model being exposed to rootcling

aalkin:
The include of `AnalysisDataModel.h` in a header, that is, in turn, included in ROOT dictionary, exposes some things to rootcling that it cannot chew. This removes the include and makes the template function more generic. Please consider reviewing this soon, as it blocks https://github.com/AliceO2Group/AliceO2/pull/12724.


mhemmer-cern (dismissed):
Thank you Anton!

mhemmer-cern (approved):

PR #4818, opened 2024-02-22
multiplicityPbPb.cxx (eta and zvtx cuts)

hhesouno:


PWGMM/Mult/Tasks/CMakeLists.txt
```diff
@@ -60,6 +60,8 @@ o2physics_add_dpl_workflow(heavy-ion-mult
                     PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::AnalysisCore
                     COMPONENT_NAME Analysis)
 
+
```
aalkin, line 63: Please avoid adding extra lines in unrelated files, check with `git diff` before committing.

PWGMM/Mult/Tasks/multiplicityPbPb.cxx
```diff
@@ -123,9 +130,14 @@ struct multiplicityPbPb {
 
       histos.fill(HIST("EtaZvtxTracks"), track.eta(), collision.posZ());
       histos.fill(HIST("PhiEtaTracks"), track.phi(), track.eta());
+
+      if(track.eta()<0.1) ++trackCounter_01;
```
aalkin, line 134: I think it is supposed to be `std::abs(track.eta()) < 0.1`?

aalkin (changes requested):

PWGMM/Mult/Tasks/CMakeLists.txt
hhesouno: I am sorry, I do not know where is that coming from. I did not modify that file intentionally.

PWGMM/Mult/Tasks/multiplicityPbPb.cxx
hhesouno: Yes, thank you so much!

PR #4822, opened 2024-02-22
added boolean configurable for the QA plots

sawankumawat:
Since the program was consuming too much memory, I added boolean configurable for the QA plots, so that I do not fill their histograms when I don't want.


sawankumawat:
@BongHwi please approve

BongHwi (approved):

PR #4830, opened 2024-02-22
Common: create a MC Collision tool for MC context

ddobrigk:
This is a first test of a tool that tags collisions according to the properties of their MC neighbours. It is meant such that we can properly calculate MC efficiencies given a certain specific-case gap configuration, taking into account also that the gap itself could contain a particle of interest (e.g. consider the Xi case). 

The basic rationale is to decompose the issues we see in the gap triggering "microscopically" isolating collisions for which the neighbours have or don't have a particle of interest, and to be able to calculate efficiencies in each of the various possibilites. Related to JIRA ticket [3702](https://its.cern.ch/jira/browse/O2-3702); tagging also @ercolessi @nepeivodaRS @ChiaraDeMartin95


alibuild (approved):

PR #4836, opened 2024-02-22
[PWGDQ] Added filter bits for muon and barrel tracks to keep track whether muon propagation and TPC postcalib have been applied at skimming time

iarsene:
Additional changes:
1) Adding enum to define the bits from the filteringTagsfor the muon and barrel tracks
2) fixing the problem with initialized mag field in table-reader-with-assoc
3) In tableMaker_withAssoc, the skimmed fwd tracks are selected according to propagated track params, but the skimmed tracks written to disk will be unpropagated so that the propagation can be applied at analysis time.


alibuild (approved):

PR #4838, opened 2024-02-22
PWGHF: fix Lc polarization

mfaggin:
- inv. mass used for 4-vector calculation was always at 0 by mistake
- add protection due to double mass hypotheses


alibuild (approved):

PR #4839, opened 2024-02-22
PWGLF: Compatibility to Run2 dataset for QA

BongHwi:
Add processes that can be run with Run2 datasets.
But the structure is very bad, we need to find a simple solution for this.


alibuild (dismissed):

alibuild (approved):

PR #4840, opened 2024-02-22
PWGCF: Identified Balance Function Implementing PID

bghanley1995:
PID should now be fully implemented and the code should now make pairs based on particle species and not just on charge


victor-gonzalez (changes requested):
Please, don't leave code commented

bghanley1995:
> Please, don't leave code commented. Just remove the code that is not going to run

I believe all commented code has now been removed.

victor-gonzalez (approved):
Thanks for addressing my comments!

PR #4845, opened 2024-02-23
event cuts, etc.

abilandz:


PWGCF/MultiparticleCorrelations/Tasks/multiparticle-correlations-ab.cxx
```diff
@@ -74,6 +76,12 @@ bool gProcessSim_Run2 = false;
 bool gProcessRec_Run1 = false;
 bool gProcessRecSim_Run1 = false;
 bool gProcessSim_Run1 = false;
+bool gProcessTest = false; // minimum subscription to the tables, for testing purposes
```
victor-gonzalez, line 79: I forgot to mention this before but per each `processWhatEverName` process function within a `PROCESS_SWITCH` clause there is an implicit variable that is defined as `bool doprocessWhatEverName` that you can use in case you so consider

victor-gonzalez (approved):
Please, don't be relaxed leaving commented code

Have in mind that there is a common Q vector framework which provides corrected Q vectors (four correction steps) for selected detectors and selected data periods

PWGCF/MultiparticleCorrelations/Tasks/multiparticle-correlations-ab.cxx
abilandz: Thanks, noted! I will integrate this feature into my code alongside the next PR.

PR #4853, opened 2024-02-23
Task update PWD/AQC

akhatun:
added checks for ITSonly and global tracks effect on UPC candidates.


akhatun:
Hello, could you please approve this pull request?

PWGUD/AQC/udQCmidRap.cxx
```diff
@@ -348,19 +362,14 @@ struct UDQCmid {
       }
     }
 
-    // 4. Check for global tracks which are no vtx tracks
+    // 4. Check for global tracks
     bool globalAndVtx = isDGcandidate;
-    bool vtxAndGlobal = isDGcandidate;
     for (auto const& track : tracks) {
-      if (track.isGlobalTrack() && !track.isPVContributor()) {
+      if (track.isGlobalTrack() && track.isPVContributor()) {
```
pbuehler, line 368: Is this really what you want? I would thin that in case that

track.isGlobalTrack() && track.isPVContributor()

is true, you want to set vtxAndGlobal to true.

akhatun: Yes, the First condition is to look for PVcontributor then if the tracks are global tracks or ITS only tracks

PR #4862, opened 2024-02-24
Efficiency and QA task for identified analyses

victor-gonzalez:
Also reorganized the PID configurable and the PID selector 
Also standard use of the self configuration


alibuild (approved):

PR #4863, opened 2024-02-24
PWGLF: Fix magnetic field retrieval in lambdakzero-/cascade-builder

vkucera:
- Remove useless switch `skipGRPOquery` which was equivalent to switching between Run 2 and 3.
- Pick the right method automatically based on the enabled process function.
- Fix the wrong fatal message.
- Avoid misleading CCDB error messages caused by trying the Run 3 method with Run 2 input.
- Use the same field for DCAFitterN and KFParticle (instead of `lround` calculation for DCAFitterN and Run 3).
- Fix C++ issues reported by cpplint.

### DIFF
--- a/PWGLF/TableProducer/cascadebuilder.cxx
+++ b/PWGLF/TableProducer/cascadebuilder.cxx
@@ -660,24 +660,21 @@ struct cascadeBuilder {
         LOG(fatal) << "Got nullptr from CCDB for path " << grpPath << " of object GRPObject for timestamp " << timestamp;
       }
       o2::base::Propagator::initFieldFromGRP(grpo);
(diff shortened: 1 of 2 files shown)


ddobrigk:
Thanks @vkucera - a lot of this dates back to the very early days and was not touched and updated

njacazio (approved):

njacazio:
@vkucera this does not change the table produced, right?

vkucera:
> @vkucera this does not change the values in the table produced, right?

The magnetic field from `getNominalBz` for Run 3 can be slightly different from 5 kG because of the multiplication factor (reported in the O2 log), whereas the `lround` formula always returns exactly 5 kG.

PR #4866, opened 2024-02-25
PWGHF: removing use of material LUT

deependra170598:
Hi @fgrosa ,
In this I tried to remove use of Material LUT as instructed in [ D2H Meeting 
20/02/2024](https://indico.cern.ch/event/1363522/contributions/5818623/attachments/2804512/4893507/2024Feb20_D2H_CoordinatorsReport.pdf).

Please see if I am not making any mistake though I tested it by building and running locally.


fgrosa:
Hi @deependra170598  thanks a lot, this looks good to me. We should probably add a version of the `initCCDB` method that allows us not to get the `LUT`, but this is something we can do in a second step (do you agree @mfaggin?)

mfaggin:
hi @fgrosa yes of course. Let's do it in another PR.

fgrosa (approved):

PR #4890, opened 2024-02-27
Add TPC NSigma QA plots

romainschotter:
Add QA plots on the TPC Nsigma (as function of centrality and pT) of the positive and negative decay daughters of K0Short, Lambda and Anti-Lambda


ddobrigk:
Good afternoon! Sorry to ping so explicitly, but could a good samaritan please help us out and review and approve? Thanks! :-D

njacazio (approved):

PR #4896, opened 2024-02-27
PWGLF: implementation of Ivanov downscaling (optional)

ddobrigk:
Related to https://its.cern.ch/jira/browse/O2-4684

### DIFF
--- a/PWGLF/DataModel/LFStrangenessTables.h
+++ b/PWGLF/DataModel/LFStrangenessTables.h
@@ -166,8 +166,8 @@ DECLARE_SOA_COLUMN(XNegAtIU, xNegAtIU, float);   //! decay position X
 DECLARE_SOA_COLUMN(YNegAtIU, yNegAtIU, float);   //! decay position Y
 DECLARE_SOA_COLUMN(ZNegAtIU, zNegAtIU, float);   //! decay position Z
 
-// ivanov scaling 
-DECLARE_SOA_COLUMN(IvanovMap, ivanovMap, int);   //! coded downscale bits
+// ivanov scaling
+DECLARE_SOA_COLUMN(IvanovMap, ivanovMap, int); //! coded downscale bits
 
 // Saved from finding: DCAs
 DECLARE_SOA_COLUMN(DCAV0Daughters, dcaV0daughters, float); //! DCA between V0 daughters
--- a/PWGLF/DataModel/LFStrangenessTables.h
+++ b/PWGLF/DataModel/LFStrangenessTables.h
@@ -159,12 +159,12 @@ DECLARE_SOA_COLUMN(ZPosAtDCA, zPosAtDCA, float); //! decay position Z
 DECLARE_SOA_COLUMN(XNegAtDCA, xNegAtDCA, float); //! decay position X
 DECLARE_SOA_COLUMN(YNegAtDCA, yNegAtDCA, float); //! decay position Y
 DECLARE_SOA_COLUMN(ZNegAtDCA, zNegAtDCA, float); //! decay position Z
-DECLARE_SOA_COLUMN(XPosAtIU, xPosAtIU, float); //! decay position X
-DECLARE_SOA_COLUMN(YPosAtIU, yPosAtIU, float); //! decay position Y
-DECLARE_SOA_COLUMN(ZPosAtIU, zPosAtIU, float); //! decay position Z
-DECLARE_SOA_COLUMN(XNegAtIU, xNegAtIU, float); //! decay position X
-DECLARE_SOA_COLUMN(YNegAtIU, yNegAtIU, float); //! decay position Y
-DECLARE_SOA_COLUMN(ZNegAtIU, zNegAtIU, float); //! decay position Z
+DECLARE_SOA_COLUMN(XPosAtIU, xPosAtIU, float);   //! decay position X
+DECLARE_SOA_COLUMN(YPosAtIU, yPosAtIU, float);   //! decay position Y
+DECLARE_SOA_COLUMN(ZPosAtIU, zPosAtIU, float);   //! decay position Z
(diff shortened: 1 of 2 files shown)


miranov25:
Thanks for implementing changes.
Before reading the code- how did you check what it is doing?

miranov25:
I reviewed the code changes and found the initial part satisfactory. However, some components appear to be missing or were not located by me:

* Implementation of the isMB trigger to include a Minimum Bias sample.
* Storage of the trigger mask.
* Storage of trigger weights.

Maybe I just missed them in the implementation. If not these can be addressed later, but I deem them crucial for subsequent performance parameterization.

ddobrigk:
> I reviewed the code changes and found the initial part satisfactory. However, some components appear to be missing or were not located by me:
> 
> * Implementation of the isMB trigger to include a Minimum Bias sample.
> * Storage of the trigger mask.
> * Storage of trigger weights.
> 
> Maybe I just missed them in the implementation. If not these can be addressed later, but I deem them crucial for subsequent performance parameterization.

Hi Marian, 

Here some answers: 

* what do you mean by a "MB" sample? Would that not just be the regular operation of the task + a uniform downsampling which we already have? 
* Well, from the single snippet you gave me, I extrapolated that you wanted to have the mask be the selection criterion and not to store the mask itself, but sure, I can also have the mask be stored if desired. To be clear: the implementation that you select minimally on that mask is what you wanted, right? 
* What is a trigger weight in your logic? Why do I need to store it in the output - can't it be recalculated from the pT of each candidate? 

Thanks a lot! 
Cheers,
David

miranov25:
Hello @ddobrigk 

Life has taught me that I prefer to rely on self-consistent data rather than external data sources. If it can be avoided, I prefer to store all the necessary data together so that there are no doubts in later analyses.

**Question:** What do you mean by a "MB" sample? Would that not just be the regular operation of the task + a uniform downsampling which we already have?

**Answer:** MB is just selecting completely randomly, in the same data set. If we get a predefined data set as MB, we cannot guarantee the same conditions there. I want to ensure the MB sample has exactly the same conditions, such that if I apply some cut on the triggered sample, the same cut working conditions will be applied in the MB sample.

**Question:** Well, from the single snippet you gave me, I extrapolated that you wanted to have the mask be the selection criterion and not to store the mask itself, but sure, I can also have the mask be stored if desired. To be clear: the implementation that you select minimally on that mask is what you wanted, right?

**Answer:** Yes. I wanted to trigger on the mask and save the mask. Later, we can improve it further once you merge; we can make further changes for trigger sampling as I discussed in previous emails (different kinds of purity triggers) so that we will be able to use that data also for higher-level validation and systematic effect studies also for particle yield systematic studies (different conditions, MC/data)

**Question:** What is a trigger weight in your logic? Why do I need to store it in the output - can't it be recalculated from the pT of each candidate?

**Answer:**

Yes, if you know all the parameters which were used in a given production, the value can be recalculated without the necessity to store it. But based on what I wrote above, I try to avoid depending on an external source of information, to reduce the possibility of making a mistake and to simplify further analysis. The trigger weight array will be well compressed as usually only one trigger will be fired.

I have to say I am also not happy with that, and for the ATO-575, I want to store functions in the files, trees such that we will not need to store values, but something similar to aliases in old trees. We should store functions in a way they can be reused later directly from files. In the past, I was using GetUserInfo but I have not decided yet.

Marian

miranov25:
Hello @ddobrigk,

I recommend that we merge the changes in their current form, assuming they are functional, and tackle the other steps at a later date. We can include weights, triggers and more sophisticated approaches as I described above in a later pull request.

**As for production validation, my priority is to have a working solution as soon as possible.**

Best regards,
Marian

ddobrigk:
> Hello @ddobrigk,
> 
> I recommend that we merge the changes in their current form, assuming they are functional, and tackle the other steps at a later date. We can include weights, triggers and more sophisticated approaches as I described above in a later pull request.
> 
> **As for production validation, my priority is to have a working solution as soon as possible.**
> 
> Best regards, Marian

Hi Marian, I've opened this up for review and taken a few further steps towards what you wanted - please take a look!

miranov25:
Hello @ddobrigk,

Thank you for your work on this feature. Upon reviewing the implementation and specifically the table declaration, I was unable to identify the portion concerning track parameters. I might have missed it, so I've directly added a comment in the code regarding this.

Additionally, I'm interested in incorporating the latest content from the trackExtra table into the track information. Would it be feasible to include this? I understand it involves a significant amount of data (factor O(20-40 %) increase), but considering our extensive downsampling O(10^-3), it should be manageable.

Regards
Marian

ddobrigk:
> Hello @ddobrigk,
> 
> Thank you for your work on this feature. Upon reviewing the implementation and specifically the table declaration, I was unable to identify the portion concerning track parameters. I might have missed it, so I've directly added a comment in the code regarding this.
> 
> Additionally, I'm interested in incorporating the latest content from the trackExtra table into the track information. Would it be feasible to include this? I understand it involves a significant amount of data (factor O(20-40 %) increase), but considering our extensive downsampling O(10^-3), it should be manageable.
> 
> Regards Marian

Hi Marian, 

You have all positions (both at IU and at V0) and the covariance matrices at these two positions, so all track parameters are automatically included. About the trackExtra table, at the moment this is encoded in the already existing [DauTrackExtras](https://github.com/AliceO2Group/O2Physics/blob/d454803472ecdd5f9d427440ad8152ef56f7ea5c/PWGLF/DataModel/LFStrangenessTables.h#L128) table: you'll find some information there. 

Cheers,
David

miranov25:
We discussed with @ddobrigk and we agreed that we should proceed with merging and a small test on the hyperloop.
If needed next pull request will follow. 
Updating further in main JIRA:
https://its.cern.ch/jira/browse/O2-4684

ercolessi (approved):

PR #4919, opened 2024-02-28
Add ZDC monitoring as a function of IR

mpuccio:


fmazzasc (approved):

mpuccio:
@ddobrigk this fixes a trivial bug in the tool to get the rate from the scalers. There are still issues that we are debugging with @lietava but it would be great if we could merge this already.

ddobrigk (approved):
yep this does look like a trivial bug :-D thanks !

mpuccio:
Thanks a lot!

PR #4932, opened 2024-02-29
PWGDQ: ML inference for tableReader

DanielSamitz:
Dear @iarsene @dsekihat @feisenhu @lucamicheletti93

this is a suggestion for implementing machine learning inference via ONNX format into the table reader.

The ML selection can be applied in `AnalysisTrackSelection`, `AnalysisMuonSelection`, `AnalysisSameEventPairing` and `AnalysisPrefilterSelection`, and acts (in the sense of filling filterMaps, histogram output, ...) like a normal track cut / pair cut, that goes by the name `"MlSelection"` and has to be added to the corresponding cut configurable like any other track cut / pair cut. Additionally `cfgApplyMlSelection` has to be set to `true` to fill the VarManager with the ML output.

The user has to give a list of ONNX model files and specify the input features and cut values for the model. Any variable that is in the VarManager can in principle be used as an input feature for the model. One can define different bins in a variable that can also be freely specified via a configurable in which different models can be applied. ONNX files can be taken from CCDB or locally. 

Additionally to the normal histogram output, another folder named `ML_Output` is created, that contains QA histograms with the model scores for each class and selection status in different bins.

Also a table that is joinable with `MyBarrelTracks` / `MyMuonTracks` / `Dielectrons` / `Dimuons` tables, respectively, is created, that holds the model scores for each class as a vector. This is at the moment not used anywhere, but could be implemented in other O2 tasks (because it's joinable wiht the other tables) or be used for further studies outside O2. The filling of the tables has to be activated by setting `cfgFillScoreTable` to `true `. To save memory the default is `false` and the tables are not filled.

Since it acts like an additional track cut / pair cut, the ML inference should work with all kind of process functions (with/without covariance, decaytypes, prefiltered,...). 

I tried to keep the modifications to the existing functions as small as possible.

Let me know your opinion about it.

### DIFF
--- a/PWGDQ/Core/CutsLibrary.cxx
+++ b/PWGDQ/Core/CutsLibrary.cxx
@@ -2398,6 +2398,12 @@ AnalysisCompositeCut* o2::aod::dqcuts::GetCompositeCut(const char* cutName)
     return cut;
   }
 
+  // ML cut
+  if (!nameStr.compare("MlSelection")) {
+    cut->AddCut(GetAnalysisCut("mlSelection"));
+    return cut;
+  }
+
   // -------------------------------------------------------------------------------------------------
   //
   // Below are a list of single electron single muon and in order or optimize the trigger
@@ -4447,6 +4453,12 @@ AnalysisCut* o2::aod::dqcuts::GetAnalysisCut(const char* cutName)
 
   if (!nameStr.compare("excludePairLowMass")) {
     cut->AddCut(VarManager::kMass, 0., 0.1, true);
+    return cut;
+  }
+
+  // ML selection
+  if (!nameStr.compare("mlSelection")) {
+    cut->AddCut(VarManager::kMlIsSelected, 0.5, 1.5);
     return cut;
   }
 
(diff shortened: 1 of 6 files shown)


iarsene:
Hi @DanielSamitz,
Could you maybe make this a draft PR, so it does not yet be merged accidentally?
I just saw this now and at the moment its the ALICE week so time is a bit limited.
since this code affects many analyses, would be great if these changes can be discussed in a meeting, preferably the usual O2 DQ meeting.
Cheers, Ionut

DanielSamitz:
Hi @iarsene ,
I converted to draft to avoid accidental merging.
Sure, one could discuss this in an O2 DQ meeting.
I also had a different version of implementing this via additional tables that hold the ML selection status that are produced by completely separate tasks, but in the end I found this to be too messy and then implemented it directly in the tableReader the way you see here. In the meeting one could e.g. also discuss the other option with pros&cons to see what analyzers think what might be best suited for their purposes.
Anyway, the most important thing is always to not interfere with any analysis that is not using this feature.
Cheers, Daniel

PR #4937, opened 2024-02-29
PWGLF: Add TF Boarder cut in the common collision cut

BongHwi:
TF Border cut can be activated via the configuration: `ConfEvtTFBorderCut`


alibuild (approved):

PR #4939, opened 2024-02-29
Making more flexible the efficiency and QA task

victor-gonzalez:
Now it is feasible to order it with PID info collection or without it as well as within multiplicity classes or just MB

Additionally
- removing the self-configuration logging
- introducing the timestamp checking task


alibuild (approved):

PR #4942, opened 2024-03-01
PWGCF: Update FemtoDream

ariedel-cern:
This PR addresses two issues:
- Remove TDataBasePDG to access particle masses in the particle pair tasks. TDataBasePDG has a large memory footprint and does not have the deuteron mass. Replace it with a switch statement and the masses defined in the PhysicsConstants.h header
- Add the time frame border cut to the collision selection of the producer.

### DIFF
--- a/PWGCF/FemtoDream/Core/femtoDreamCollisionSelection.h
+++ b/PWGCF/FemtoDream/Core/femtoDreamCollisionSelection.h
@@ -19,6 +19,7 @@
 #include <string>
 #include <iostream>
 #include "Common/CCDB/TriggerAliases.h"
+#include "Common/DataModel/EventSelection.h"
 #include "Framework/HistogramRegistry.h"
 #include "Framework/Logger.h"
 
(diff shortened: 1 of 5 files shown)


PWGCF/FemtoDream/Core/femtoDreamCollisionSelection.h
```diff
@@ -83,6 +84,12 @@ class FemtoDreamCollisionSelection
   template <typename C>
   bool isSelectedCollision(C const& col)
   {
+
+    // remove events at the border of the time frame
+    if (!col.selection_bit(aod::evsel::kNoTimeFrameBorder)) {
```
victor-gonzalez, line 89: I would suggest to also remove the out of ITS read out frame collisions

victor-gonzalez (approved):

victor-gonzalez:
I cannot restart the checks
You can do it by pushing an empty commit (or implement my suggestion if you consider it :))

PR #4946, opened 2024-03-01
Common: print an explicit warning of MC scaling use

ddobrigk:
Testing for MC centrality calibration for the Pb-Pb 2023 period, almost ready...


alibuild (approved):

PR #4951, opened 2024-03-01
EventFiltering: Fix units for mass resolution parametrisation

fgrosa:


alibuild (approved):

PR #4954, opened 2024-03-01
PWGHF: Add methods to compute raw and corrected fraction

fgrosa:


vkucera:
The import errors can be disabled on the relevant lines but what pyright found looks like genuine bugs.

fgrosa:
> The import errors can be disabled on the relevant lines but what pyright found looks like genuine bugs.

Indeed there was a genuine error in the previously existing code (I don't know why it didn't trigger before) and then it was complaining about the initialization to `None` (again already there before). It is now everything fixed anyway.

alibuild (approved):

PR #4956, opened 2024-03-01
Protect the exclusive run of the timestamp check task

victor-gonzalez:
Also proper histogram run number labeling and get rid of self-initialization printouts


alibuild (approved):

PR #4961, opened 2024-03-01
PWGHF: Add chi2PCA cut to Ds selector

fchinu:
This PR adds the possibility of applying selections on the chi2PCA of Ds candidates


fgrosa (approved):

fgrosa:
> This PR adds the possibility of applying selections on the chi2PCA of Ds candidates

@fchinu be careful when you synchronise the Ds selector wagon, all the values will be reset to the default ones because you changed the dimension of the labelledarray (informing also @scattaru who also is doing an analysis with Ds)

PR #4969, opened 2024-03-03
multiplicity-Pb-Pb adding filtering for Mc generated by charges

hhesouno:


PWGMM/Mult/Tasks/CMakeLists.txt
```diff
@@ -60,6 +60,8 @@ o2physics_add_dpl_workflow(heavy-ion-mult
                     PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::AnalysisCore
                     COMPONENT_NAME Analysis)
 
+
+
```
aalkin, line 64: Please remove extra lines.

PWGMM/Mult/Tasks/multiplicityPbPb.cxx
```diff
@@ -155,6 +155,11 @@ struct multiplicityPbPb {
     for (auto& mcParticle : mcParticles) {
       ++MCparticleCounter;
       if (mcParticle.isPhysicalPrimary()) {
+        auto pdgparticle = pdg->GetParticle(mcParticle.pdgCode());
+        if (pdgparticle != nullptr)
```
aalkin, line 159: The coding guidelines require braces even around single-line branches.

PWGMM/Mult/Tasks/multiplicityPbPb.cxx
```diff
@@ -155,6 +155,11 @@ struct multiplicityPbPb {
...
+          if (pdgparticle->Charge() == 0)
```
aalkin, line 160: This value is stored as a floating point number, comparison of a floating point number to 0 is not guaranteed to be always correct. You can do something like `std::abs(pdgparticle->Charge()) > 2.` (note that the charge value is in units of |e|/3 and simulated particle stack may contain quarks).

aalkin (changes requested):

PWGMM/Mult/Tasks/multiplicityPbPb.cxx
hhesouno: Hello, 
thank you, I need to distinguish charged and neutral particles. Will it work if I use std::abs(pdgparticle->Charge()) > 1?

aalkin: The value of `1` corresponds to 1/3 of the electron charge, so using condition `>1` would let particles with the charge of 2/3 of the electron charge into your sample. Asking for a particles with |charge| more than 2, i.e. `> 2/3|e|`, or, alternatively, skipping particles with |charge| less than 3. `< 1|e|`, is what you want.

PR #4986, opened 2024-03-04
PWGCF: FemtoUniverse -- Fixing DetaDphi* cut

zchochul:
Before: Common delta eta delta phi* plots for same and mixed events (sum).

After: Separate plots for same and mixed, allowing you to divide the plots later and get a useful QA plot!


PWGCF/FemtoUniverse/Core/FemtoUniverseDetaDphiStar.h
```diff
@@ -197,13 +201,87 @@ class FemtoUniverseDetaDphiStar
...
+        } else {
+          histdetadpimixed[i][1]->Fill(deta, dphiAvg);
+        }
+      }
+      return pass;
+    } else {
+      LOG(fatal) << "FemtoUniversePairCleaner: Combination of objects not defined - quitting!";
+      return false;
+    }
+  }
```
victor-gonzalez, line 272: Basically you are repeating the full code
I would suggest either template on the kind of pair origin, same or mixed, or pass an additional parameter with such origin.
Then you rely on that origin to peek the proper histograms

victor-gonzalez (changes requested):
I believe it is worth to consider my suggestion

PWGCF/FemtoUniverse/Core/FemtoUniverseDetaDphiStar.h
zchochul: Thank you! I will do that ;)

victor-gonzalez (commented):
I would not use a `string` for coding same/mixed
Have in mind that you usually are in the pair loop, two inner loops, and you need to compare each character of the string
I would suggest using an `enum`, `kSame`, `kMixed`, (you probably already have something like that in the source/include hierarchy) with which you just compare one single value that is by far much more efficient

PR #4989, opened 2024-03-04
[PWGDQ] Using template process function for muons and electron channels

lucamicheletti93:
Using template process function for muons and electron channels for J/psi-D0 table maker


alibuild (approved):

PR #4992, opened 2024-03-04
PWGHF: split THnSparses of HF polarization task, one per cosThStar axis.

mfaggin:
@fgrosa here the PR we discussed about. Let me know if you like this version where the `THnSparse`s can be switched on/off. This of course introduces a bit of `if` statements. If you prefer I can keep everything active w/o `if`s, let me know what you prefer.


fgrosa (approved):
Hi @mfaggin, I think that having the `if` statements is good, also considering that in the future we will add the cost* wrt to the event plane for Pb-Pb (and it is good to have the possibility to switch off some of them). Thanks!

PR #5001, opened 2024-03-05
PWGHF: Add Ds sign to the output of the tree creator

fchinu:
This PR adds the Ds sign to the tree creator output.
This is needed to study possible differences in the mass distribution between positive and negative candidates.
To avoid enlarging the lite tree, and because this information is not needed for the physics analysis, the variable was added to the full tree

### DIFF
--- a/PWGHF/TableProducer/treeCreatorDsToKKPi.cxx
+++ b/PWGHF/TableProducer/treeCreatorDsToKKPi.cxx
@@ -80,7 +80,7 @@ DECLARE_SOA_COLUMN(AbsCos3PiK, absCos3PiK, float);                           //!
 // Events
 DECLARE_SOA_COLUMN(IsEventReject, isEventReject, int); //! Event rejection flag
 DECLARE_SOA_COLUMN(RunNumber, runNumber, int);         //! Run number
-DECLARE_SOA_COLUMN(Sign, sign, int);                   //! Sign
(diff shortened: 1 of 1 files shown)


fcatalan92 (approved):

PWGHF/TableProducer/treeCreatorDsToKKPi.cxx
```diff
@@ -80,6 +80,7 @@ DECLARE_SOA_COLUMN(AbsCos3PiK, absCos3PiK, float);                           //!
 // Events
 DECLARE_SOA_COLUMN(IsEventReject, isEventReject, int); //! Event rejection flag
 DECLARE_SOA_COLUMN(RunNumber, runNumber, int);         //! Run number
+DECLARE_SOA_COLUMN(Sign, sign, int);                   //! Sign
```
mfaggin, line 83:
```suggestion
DECLARE_SOA_COLUMN(Sign, sign, int8_t);                   //! Sign
```
can you test if this works?

mfaggin (changes requested):

PWGHF/TableProducer/treeCreatorDsToKKPi.cxx
fchinu: Thanks for your suggestion! It works, I have implemented it

mfaggin (approved):
thank you!

PR #5004, opened 2024-03-05
PWGHF: Add ML inference class for Dstar->D0Pi and related modifications to the Dstar selector

Mingze129:
HfMlResponseDstarToD0Pi created

Updates to candidateSelectorDstarToD0Pi to use ML interface

@fgrosa

### DIFF
--- a/PWGHF/Core/HfMlResponseDstarToD0Pi.h
+++ b/PWGHF/Core/HfMlResponseDstarToD0Pi.h
@@ -142,7 +142,7 @@ class HfMlResponseDstarToD0Pi : public HfMlResponse<TypeOutputScore>
     for (const auto& idx : MlResponse<TypeOutputScore>::mCachedIndices) {
       switch (idx) {
           CHECK_AND_FILL_VEC_DSTAR(chi2PCAD0);
-          CHECK_AND_FILL_VEC_DSTAR(decayLengthD0); 
+          CHECK_AND_FILL_VEC_DSTAR(decayLengthD0);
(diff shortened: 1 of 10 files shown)


PWGHF/Core/HfMlResponseDstarToD0Pi.h
```diff
@@ -0,0 +1,230 @@
...
+  ptSoftPi,
+  impactParameter0,
+  impactParameter1,
+  impParamSoftPi,
+  impactParameterNormalised0,
+  impactParameterNormalised1,
+  normalisedImpParamSoftPi,
+  cosThetaStarD0,
+  massD0,
+  massDStar,
```
fgrosa, line 87: The D* mass should never been used as a ML feature, so it would be better to remove it.

PWGHF/Core/HfMlResponseDstarToD0Pi.h
```diff
@@ -0,0 +1,230 @@
...
+  ptProng1,
+  ptSoftPi,
+  impactParameter0,
+  impactParameter1,
+  impParamSoftPi,
+  impactParameterNormalised0,
+  impactParameterNormalised1,
+  normalisedImpParamSoftPi,
+  cosThetaStarD0,
+  massD0,
```
fgrosa, line 86: We could maybe have both the `massD0` and the `deltaMassD0` to allow the user to chose what to use

fgrosa (changes requested):
Hi @MZL-Diaboli thanks a lot for this development! I just have two minor comments, for the rest it looks good to me.

PR #5015, opened 2024-03-06
PWG-UD: introduce configurables to save only PV tracks and reject collisions at TF boundaries

pbuehler:


alibuild (approved):

PR #5016, opened 2024-03-06
PWGJE - jetfinder(hf)QA - random cone fix; adding centrality to track…

aimeric-landou:
… histograms

### DIFF
--- a/PWGJE/Tasks/jetfinderhfQA.cxx
+++ b/PWGJE/Tasks/jetfinderhfQA.cxx
@@ -356,11 +356,11 @@ struct JetFinderHFQATask {
 
     if (doprocessTracks || doprocessTracksWeighted) {
       registry.add("h_collisions", "event status;event status;entries", {HistType::kTH1F, {{4, 0.0, 4.0}}});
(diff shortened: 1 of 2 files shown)


PWGJE/Tasks/jetfinderhfQA.cxx
```diff
@@ -1277,25 +1277,28 @@ struct JetFinderHFQATask {
...
-            float dPhi = RecoDecay::constrainAngle(track.phi() - randomConePhi, static_cast<float>(-M_PI));
-            float dEta = track.eta() - randomConeEta;
-            if (TMath::Sqrt(dEta * dEta + dPhi * dPhi) < randomConeR) {
-              randomConePt += track.pt();
+      // removing the leading jet from the random cone
+      if (jets.size() > 0) { // if there are no jets in the acceptance (from the jetfinder cuts) then there can be no leading jet
+        float dPhiLeadingJet = RecoDecay::constrainAngle(jets.iteratorAt(0).phi() - randomConePhi, static_cast<float>(-M_PI));
+        float dEtaLeadingJet = jets.iteratorAt(0).eta() - randomConeEta;
+
+        bool jetWasInCone = false;
```
nzardosh, line 1285: Hi Aimeric, I think you'd snippet of code is wrong. This bool should be declared outside the jets.size()> 0 if loop and that if loop should and it should close before L1293. You can look at the example on matter most

nzardosh (approved):

PR #5017, opened 2024-03-06
EventFiltering/PWGHF: first implementation of SigmaC++K-, SigmaC0K0 triggers

mfaggin:


fgrosa (changes requested):
Hi @mfaggin thanks a lot for this development! I would still wait the tag for the 2023 CEFP before merging, but I have few comments already (see below). In addition to those, I would add the possibility to add min pT cuts to the resonances to possibly tune them to fulfil the rejection factor requirements (but this can also be added later).

fgrosa:
To be rebased on https://github.com/AliceO2Group/O2Physics/pull/5144

fgrosa (approved):
Thanks a lot @mfaggin!

PR #5019, opened 2024-03-06
[PWGDQ] Adding process function using central q-vectors

lucamicheletti93:
Adding process function using central q-vectors. The new function is independent with respect the original one and produces the same tables.


alibuild (approved):

PR #5022, opened 2024-03-06
PWGHF: Add workflow for producing combined TPC+TOF NSigma

vkucera:
Workflow to produce tables with combined TPC+TOF sigmas in static columns instead of extended tables with expression columns.
Rationale: Extended tables have to be spawned. Spawning cannot be conditional (yet) so it is not possible to produce extended tables for different particle species only when needed. Spawning needed tables inside different selector workflows makes it impossible to run these workflows together because of the conflicting table production.

Implementation in this PR:

- Produce normal tables with static columns for all particle species and table types (full, tiny).
- Run one process function per table to isolate subscriptions to parent PID tables (which automatically enable their filling by the PID task).
- Automatically disable process functions enabled for not needed tables. (This prevent the process function from running but it does not remove the input table subscription from the context! The input table is still considered consumed and will be produced by the PID task anyway. It is therefore still important to set the process switches correctly to avoid wasting resources.)
- Throw fatal for process functions disabled for needed tables. (Cannot be enabled.)


fgrosa (approved):
It looks good to me, thanks!

PR #5023, opened 2024-03-06
PWGHF: Remove obsolete table with HF signal info

fcatalan92:
In this way the task `o2-analysis-hf-task-mc-validation` can have sub-wagons.


fgrosa (approved):
Thanks @fcatalan92!

PR #5026, opened 2024-03-06
Hadronic rate and ZDC centroids monitoring

mpuccio:
- **Fix ctpRateFetcher runNumber checks**
- **Add ZDC centroid monitoring and additional info in the table**


mpuccio:
@ddobrigk this fixes a bug in the usage of the ctpRateFetcher that was causing a slow down in (the only) task that uses it

ddobrigk (approved):

PR #5030, opened 2024-03-07
PWGHF: add rotational background for Lc->pKpi.

mfaggin:
@fgrosa this is the PR that we discussed. Actually I did not manage to test it locally due to the issue that I've mentioned [here](https://mattermost.web.cern.ch/alice/pl/4usqpqj1piyj8xto8xiinj3aqe), waiting for its resolution.

### DIFF
--- a/PWGHF/D2H/Tasks/taskCharmPolarisation.cxx
+++ b/PWGHF/D2H/Tasks/taskCharmPolarisation.cxx
@@ -282,7 +282,7 @@ struct TaskPolarisationCharmHadrons {
           pzDau = candidate.pzProng2();
           if (bkgRotationId) {
             /// rotational background - pt of the kaon track rotated
-            invMassCharmHad = RecoDecay::m(std::array{threeVecLcProng0, threeVecLcRotatedProng1, threeVecLcProng2}, std::array{massProton, massKaon, massPi});
+            invMassCharmHad = RecoDecay::m(std::array{threeVecLcProng0, threeVecLcRotatedProng1, threeVecLcProng2}, std::array{massPi, massKaon, massProton});
             invMassCharmHadForSparse = invMassCharmHad;
           } else {
             /// original candidate (kaon track not rotated)
--- a/PWGHF/D2H/Tasks/taskCharmPolarisation.cxx
+++ b/PWGHF/D2H/Tasks/taskCharmPolarisation.cxx
@@ -60,8 +60,10 @@ struct TaskPolarisationCharmHadrons {
 
   float massPi{0.f};
   float massProton{0.f};
+  float massKaon{0.f};
   float massDstar{0.f};
   float massLc{0.f};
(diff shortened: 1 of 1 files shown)


PWGHF/D2H/Tasks/taskCharmPolarisation.cxx
```diff
@@ -138,33 +146,34 @@ struct TaskPolarisationCharmHadrons {
     const AxisSpec thnAxisCosThetaStarBeam{configThnAxisCosThetaStarBeam, "cos(#vartheta_{beam})"};
     const AxisSpec thnAxisMlBkg{configThnAxisMlBkg, "ML bkg"};
     const AxisSpec thnAxisMlNonPrompt{configThnAxisMlNonPrompt, "ML non-prompt"};
+    const AxisSpec thnAxisIsRotatedCandidate{configThnAxisIsRotatedCandidate, "0: standard candidate, 1: rotated candidate"};
 
     if (doprocessDstarWithMl || doprocessLcToPKPiWithMl) {
       if (activateTHnSparseCosThStarHelicity) {
-        registry.add("hSparseCharmPolarisationHelicity", "THn for polarisation studies with cosThStar w.r.t. helicity axis and BDT scores", HistType::kTHnSparseF, {thnAxisInvMass, thnAxisPt, thnAxisPz, thnAxisY, thnAxisCosThetaStarHelicity, thnAxisMlBkg, thnAxisMlNonPrompt});
+        registry.add("hSparseCharmPolarisationHelicity", "THn for polarisation studies with cosThStar w.r.t. helicity axis and BDT scores", HistType::kTHnSparseF, {thnAxisInvMass, thnAxisPt, thnAxisPz, thnAxisY, thnAxisCosThetaStarHelicity, thnAxisMlBkg, thnAxisMlNonPrompt, thnAxisIsRotatedCandidate});
```
fgrosa, line 153: Unless we want to implement the track rotation also for D* (which might be a good idea), I would add the additional axis only for Lc->pKpi

fgrosa (changes requested):
Hi @mfaggin, it looks good to me, I just have a small comment about the ThnSparse (currently we are adding an axis also for D*, which is not needed)

PWGHF/D2H/Tasks/taskCharmPolarisation.cxx
mfaggin: I followed your suggestion, and I removed this further axis for the D*+. It required a simpe but a bit tedious re-coding of the several `if` statements. I re-tested for Lc locally and it works.

mfaggin: then if needed we can add rot. bkg for D*+ as well, but since it does not seem strictly necessary at the moment I would avoid it

fgrosa (approved):

PR #5037, opened 2024-03-07
PWGHF: add extra cuts on correlatorDMesonPairs.cxx

atavirag:
This PR aims to update correlatorDMesonPairs.cxx by adding the cuts implemented and discussed on correlatorDMesonParisTesting.cxx. In particular:
- Adding a configurable only to select the events which have an invariant mass that is close to the PDG one
- Adding a cut to avoid having candidates who share a daughter track
- Updating the construction of the for loop for the second candidates


fgrosa (approved):
Hi @atavirag it looks good to me, thanks! As discussed offline, once you verify that you get the same output, it would be good to merge this version and the test one (I have no preference on the one that you want to keep)

PR #5038, opened 2024-03-07
PWGHF: implement new version of treeCreatorToXiPi table

ZFederica:
- Implement new version of treeCreatorToXiPi table: the new table contains less information and is filled only for candidates passing a set of selection. This will allow for having a smaller size output.
- add event table


PWGHF/TableProducer/treeCreatorToXiPi.cxx
```diff
@@ -31,6 +31,7 @@ namespace o2::aod
 namespace full
 {
 // from creator
+DECLARE_SOA_COLUMN(CollisionCounter, collisionCounter, int);
```
mfaggin, line 34: a warning from my side: in case you will build your small trees on a relatively large dataset, this counter might be not ideal id defined as `int`. Indeed, beware that the max. integer is 2147483647 (just a bit more than 2B)

PWGHF/TableProducer/treeCreatorToXiPi.cxx
```diff
@@ -294,27 +326,149 @@ struct HfTreeCreatorToXiPi {
...
+        originMc,
+        collisionMatched);
+    }
+  }
+
+  void processDataFull(aod::Collisions const& collisions, MyTrackTable const&,
+                       soa::Join<aod::HfCandToXiPi, aod::HfSelToXiPi> const& candidates)
   {
+    // Filling total number of collisions processed
+    int collCounter = 0;
```
mfaggin, line 405: same warning as above

PWGHF/TableProducer/treeCreatorToXiPi.cxx
ZFederica: Thanks @mfaggin, I switched to long int, that should be enough, right?

mfaggin: actually the limit I mention above is exactly for a `long int`, maybe it's better to go for `long long int`, if you really need it

PWGHF/TableProducer/treeCreatorToXiPi.cxx
```diff
@@ -294,27 +326,149 @@ struct HfTreeCreatorToXiPi {
...
+
+  void processDataFull(aod::Collisions const& collisions, MyTrackTable const&,
+                       soa::Join<aod::HfCandToXiPi, aod::HfSelToXiPi> const& candidates)
   {
+    // Filling total number of collisions processed
+    int64_t collCounter = 0;
+    for (const auto& collision : collisions) {
+      ++collCounter;
+    }
+    rowEv(collCounter);
```
vkucera, line 409: I suspect that you need to apply the same event selection as for your candidates.
Plus you can get the size of a table (or partition) with `.size()`.

PWGHF/TableProducer/treeCreatorToXiPi.cxx
ZFederica: @vkucera  I did not think about it, but may it depend on the value specified for `Max DF size` in the derived data tab on the hyperloop wagon?

mfaggin: guys you are right indeed, sorry, my fault

ZFederica:
Hi @vkucera, after talking with Fabrizio we agreed on including more information in the event table (needed for offline analysis). Now the PR is ready for review. Thanks!

PWGHF/TableProducer/treeCreatorToXiPi.cxx
```diff
@@ -164,26 +172,67 @@ DECLARE_SOA_TABLE(HfToXiPiFulls, "AOD", "HFTOXIPIFULL",
...
                   full::DecLenCharmBaryon, full::DecLenCascade, full::DecLenV0, full::ErrorDecayLengthCharmBaryon, full::ErrorDecayLengthXYCharmBaryon,
-                  full::NormImpParCascade, full::NormImpParPiFromCharmBar, full::NormDecayLenCharmBar, full::IsPionGlbTrkWoDca,
+                  full::NormImpParCascade, full::NormImpParPiFromCharmBar, full::NormDecayLenCharmBar, full::IsPionGlbTrkWoDca, full::PionItsNCls,
                   full::StatusPidLambda, full::StatusPidCascade, full::StatusPidCharmBaryon,
                   full::StatusInvMassLambda, full::StatusInvMassCascade, full::StatusInvMassCharmBaryon, full::ResultSelections, full::PidTpcInfoStored, full::PidTofInfoStored,
                   full::TpcNSigmaPiFromCharmBaryon, full::TpcNSigmaPiFromCasc, full::TpcNSigmaPiFromLambda, full::TpcNSigmaPrFromLambda,
                   full::TofNSigmaPiFromCharmBaryon, full::TofNSigmaPiFromCasc, full::TofNSigmaPiFromLambda, full::TofNSigmaPrFromLambda,
                   full::FlagMcMatchRec, full::DebugMcRec, full::OriginRec, full::CollisionMatched);
 
+DECLARE_SOA_TABLE(HfToXiPiLite, "AOD", "HFTOXIPILITE",
```
vkucera, line 182:
```suggestion
DECLARE_SOA_TABLE(HfToXiPiLites, "AOD", "HFTOXIPILITE",
```

vkucera: I think this is needed for merging data frames.

PWGHF/TableProducer/treeCreatorToXiPi.cxx
```diff
@@ -164,26 +172,67 @@ DECLARE_SOA_TABLE(HfToXiPiFulls, "AOD", "HFTOXIPIFULL",
...
+                  full::TofNSigmaPiFromCharmBaryon, full::TofNSigmaPiFromCasc, full::TofNSigmaPiFromLambda, full::TofNSigmaPrFromLambda,
+                  full::FlagMcMatchRec, full::OriginRec, full::CollisionMatched);
+
 } // namespace o2::aod
 
 /// Writes the full information in an output TTree
 struct HfTreeCreatorToXiPi {
 
   Produces<o2::aod::HfToXiPiFulls> rowCandidateFull;
+  Produces<o2::aod::HfToXiPiLite> rowCandidateLite;
```
vkucera, line 212:
```suggestion
  Produces<o2::aod::HfToXiPiLites> rowCandidateLite;
```

PWGHF/TableProducer/treeCreatorToXiPi.cxx
```diff
@@ -164,26 +172,67 @@ DECLARE_SOA_TABLE(HfToXiPiFulls, "AOD", "HFTOXIPIFULL",
...
+
+  using MyTrackTable = soa::Join<aod::Tracks, aod::TrackSelection, aod::TracksExtra>;
+  using MyEventTable = soa::Join<aod::Collisions, aod::EvSels>;
 
   void init(InitContext const&)
   {
   }
 
+  template <class TMyEvents, typename T>
+  void fillEvent(const T& collision, float cutZPv, const bool& newEvent, bool& isEvSel8, bool& isEventZSel)
```
vkucera, line 225:
```suggestion
  void fillEvent(const T& collision, float cutZPv, bool newEvent, bool& isEvSel8, bool& isEventZSel)
```

vkucera: Why do you pass `isEvSel8` and `isEventZSel` by reference? You reset them to `false` for each collision and I don't see the changed values used anywhere outside the function.

vkucera: Can you not just directly fill:
```c++
rowEv(newEvent, collision.sel8(), std::abs(collision.posZ()) < cutZPv);
```
?

PWGHF/TableProducer/treeCreatorToXiPi.cxx
```diff
@@ -294,27 +344,167 @@ struct HfTreeCreatorToXiPi {
...
+        originMc,
+        collisionMatched);
+    }
+  }
+
+  void processDataFull(MyEventTable const& collisions, MyTrackTable const&,
+                       soa::Join<aod::HfCandToXiPi, aod::HfSelToXiPi> const& candidates)
+  {
+    // Filling event properties
+    bool newEvent = true;
```
vkucera, line 425: When is `newEvent` set to `false`?

vkucera (changes requested):

PWGHF/TableProducer/treeCreatorToXiPi.cxx
ZFederica: This should never be set to false, it is filled in the table when a new collision is processed, so it will always be a new event --> this variable should be always true

PWGHF/TableProducer/treeCreatorToXiPi.cxx
ZFederica: Yes, sure, I modified it according to this suggestion

PWGHF/TableProducer/treeCreatorToXiPi.cxx
```diff
@@ -164,26 +171,61 @@ DECLARE_SOA_TABLE(HfToXiPiFulls, "AOD", "HFTOXIPIFULL",
...
 
-  using MyTrackTable = soa::Join<aod::Tracks, aod::TrackSelection>;
+  using MyTrackTable = soa::Join<aod::Tracks, aod::TrackSelection, aod::TracksExtra>;
+  using MyEventTable = soa::Join<aod::Collisions, aod::EvSels>;
 
   void init(InitContext const&)
   {
   }
 
+  template <class TMyEvents, typename T>
```
vkucera, line 223:
```suggestion
  template <typename T>
```

PWGHF/TableProducer/treeCreatorToXiPi.cxx
```diff
@@ -294,27 +337,147 @@ struct HfTreeCreatorToXiPi {
...
+    }
+  }
+
+  void processDataFull(MyEventTable const& collisions, MyTrackTable const&,
+                       soa::Join<aod::HfCandToXiPi, aod::HfSelToXiPi> const& candidates)
   {
+    // Filling event properties
+    rowEv.reserve(collisions.size());
+    for (const auto& collision : collisions) {
+      fillEvent<MyEventTable>(collision, zPvCut);
```
vkucera, line 420:
```suggestion
      fillEvent(collision, zPvCut);
```

vkucera (approved):

PR #5040, opened 2024-03-07
PWGHF: Add ThnSparse of the ML scores vs Vars to the Lc task

zhangbiao-phy:


PWGHF/D2H/Tasks/taskLc.cxx
```diff
@@ -235,11 +267,28 @@ struct HfTaskLc {
...
+      const AxisSpec thnAxisMultiplicity{thnConfigAxisMultiplicity, "multiplicity"};
+      const AxisSpec thnAxisChi2PCA{thnConfigAxisChi2PCA, "Chi2PCA to sec. vertex (cm)"};
+      const AxisSpec thnAxisDecLength{thnConfigAxisDecLength, "decay length (cm)"};
+      const AxisSpec thnAxisDecLengthXY{thnConfigAxisDecLengthXY, "decay length xy (cm) "};
+      const AxisSpec thnAxisCPA{thnConfigAxisCPA, "cosine of pointing angle"};
+      const AxisSpec thnAxisCPAXY{thnConfigAxisCPAXY, "cosine of pointing angle xy"};
+      const AxisSpec thnAxisBdtScoreLcBkg{thnConfigAxisBdtScoreLcBkg, "BDT bkg score (Lc)"};
+      const AxisSpec thnAxisBdtScoreLcNonPrompt{thnConfigAxisBdtScoreLcNonPrompt, "BDT non-prompt score (Lc)"};
+      const AxisSpec thnAxisCanType{thnConfigAxisCanType, "candidates type"};
+      registry.add("hBdtScoreVsVars", "THn for Lambdac candidates", HistType::kTHnSparseF, {thnAxisMass, thnAxisPt, thnAxisPtProng0, thnAxisPtProng1, thnAxisPtProng2, thnAxisMultiplicity, thnAxisChi2PCA, thnAxisDecLength, thnAxisDecLengthXY, thnAxisCPA, thnAxisCPAXY, thnAxisBdtScoreLcBkg, thnAxisBdtScoreLcNonPrompt, thnAxisCanType});
```
fgrosa, line 286: In O2 there is a maximum limit of axes that you can add to a multidimensional histogram (see https://github.com/AliceO2Group/AliceO2/blob/f09accf8cf210b047fb1fa0726a451510a72d8ad/Framework/Core/src/HistogramSpec.cxx#L82-L86), so if you put more than 10 axes it crashes.

Maybe you can foresee a version of the THnSparse with ML scores, and one with topological variables. Also consider that it will be very large (especially in Pb-Pb collisions, so having a small number of axes it could be a better idea).

PWGHF/D2H/Tasks/taskLc.cxx
```diff
@@ -235,11 +267,28 @@ struct HfTaskLc {
     registry.add("MC/reconstructed/signal/hDecLenErrSig", "3-prong candidates (matched);decay length error (cm);entries", {HistType::kTH2F, {{100, 0., 1.}, {vbins, "#it{p}_{T} (GeV/#it{c})"}}});
     registry.add("MC/reconstructed/prompt/hDecLenErrSigPrompt", "3-prong candidates (matched, prompt);decay length error (cm);entries", {HistType::kTH2F, {{100, 0., 1.}, {vbins, "#it{p}_{T} (GeV/#it{c})"}}});
     registry.add("MC/reconstructed/nonprompt/hDecLenErrSigNonPrompt", "3-prong candidates (matched, non-prompt);decay length error (cm);entries", {HistType::kTH2F, {{100, 0., 1.}, {vbins, "#it{p}_{T} (GeV/#it{c})"}}});
+
+    if (applyMl) {
+      const AxisSpec thnAxisMass{thnConfigAxisMass, "inv. mass (p K #pi) (GeV/#it{c}^{2})"};
+      const AxisSpec thnAxisPt{thnConfigAxisPt, "#it{p}_{T}(#Lambda_{c}^{+}) (GeV/#it{c})"};
+      const AxisSpec thnAxisPtProng0{thnConfigAxisPtProng0, "#it{p}_{T}(prong0) (GeV/#it{c})"};
+      const AxisSpec thnAxisPtProng1{thnConfigAxisPtProng1, "#it{p}_{T}(prong0) (GeV/#it{c})"};
+      const AxisSpec thnAxisPtProng2{thnConfigAxisPtProng2, "#it{p}_{T}(prong1) (GeV/#it{c})"};
```
fgrosa, line 276:
```suggestion
      const AxisSpec thnAxisPtProng1{thnConfigAxisPtProng1, "#it{p}_{T}(prong1) (GeV/#it{c})"};
      const AxisSpec thnAxisPtProng2{thnConfigAxisPtProng2, "#it{p}_{T}(prong2) (GeV/#it{c})"};
```

PWGHF/D2H/Tasks/taskLc.cxx
```diff
@@ -37,10 +38,36 @@ struct HfTaskLc {
   Configurable<double> yCandGenMax{"yCandGenMax", 0.5, "max. gen particle rapidity"};
   Configurable<double> yCandRecoMax{"yCandRecoMax", 0.8, "max. cand. rapidity"};
   Configurable<std::vector<double>> binsPt{"binsPt", std::vector<double>{hf_cuts_lc_to_p_k_pi::vecBinsPt}, "pT bin limits"};
+  // ThnSparse for ML outputScores and Vars
+  Configurable<bool> applyMl{"applyMl", false, "Flag to apply ML selections"};
+  ConfigurableAxis thnConfigAxisPt{"thnConfigAxisPt", {360, 0, 36}, ""};
+  ConfigurableAxis thnConfigAxisMass{"thnConfigAxisMass", {600, 1.98, 2.58}, ""};
+  ConfigurableAxis thnConfigAxisPtProng0{"thnConfigAxisPtProng0", {500, 0, 50}, ""};
+  ConfigurableAxis thnConfigAxisPtProng1{"thnConfigAxisPtProng1", {500, 0, 50}, ""};
+  ConfigurableAxis thnConfigAxisPtProng2{"thnConfigAxisPtProng2", {500, 0, 50}, ""};
```
fgrosa, line 47: Coudn't this be the same configurable? Do you need different binning for the different daughter tracks?

PWGHF/D2H/Tasks/taskLc.cxx
```diff
@@ -37,10 +38,36 @@ struct HfTaskLc {
   Configurable<double> yCandGenMax{"yCandGenMax", 0.5, "max. gen particle rapidity"};
   Configurable<double> yCandRecoMax{"yCandRecoMax", 0.8, "max. cand. rapidity"};
   Configurable<std::vector<double>> binsPt{"binsPt", std::vector<double>{hf_cuts_lc_to_p_k_pi::vecBinsPt}, "pT bin limits"};
+  // ThnSparse for ML outputScores and Vars
```
fgrosa, line 41: I would add a configurable to switch on / off the THnSparse regardless the ML (see also comment below about the axes)

PWGHF/D2H/Tasks/taskLc.cxx
```diff
@@ -305,12 +363,43 @@ struct HfTaskLc {
...
+        if (candidate.isSelLcToPKPi() >= selectionFlagLc) {
+          Mass = hfHelper.invMassLcToPKPi(candidate);
+          outputBkg = candidate.mlProbLcToPKPi()[0]; /// bkg score
+          outputFD = candidate.mlProbLcToPKPi()[1];  /// non-prompt score
+          /// Fill the ML outputScores and variables of candidate
+          registry.get<THnSparse>(HIST("hBdtScoreVsVars"))->Fill(Mass, pt, ptProng0, ptProng1, ptProng2, nTracks, chi2PCA, decayLength, decayLengthXY, cpa, cpaXY, outputBkg, outputFD, 0);
+        } else if (candidate.isSelLcToPiKP() >= selectionFlagLc) {
+          Mass = hfHelper.invMassLcToPiKP(candidate);
+          outputBkg = candidate.mlProbLcToPKPi()[0]; /// bkg score
+          outputFD = candidate.mlProbLcToPKPi()[1];  /// non-prompt score
```
fgrosa, line 380: I guess that here you should use `mlProbLcToPiKP`
Also, typically we put the FD score in 3rd position (this is not mandatory, but it is what we usually do, so it would be good to have the same convention everywhere in the code)
Also same comment about possibly empty vectors (see above)

PWGHF/D2H/Tasks/taskLc.cxx
```diff
@@ -305,12 +363,43 @@ struct HfTaskLc {
...
       registry.fill(HIST("Data/hDecLenErr"), candidate.errorDecayLength(), pt);
+
+      if constexpr (applyMl) {
+        double Mass(-1);
+        double outputBkg(-1);
+        double outputFD(-1);
+        if (candidate.isSelLcToPKPi() >= selectionFlagLc) {
+          Mass = hfHelper.invMassLcToPKPi(candidate);
+          outputBkg = candidate.mlProbLcToPKPi()[0]; /// bkg score
+          outputFD = candidate.mlProbLcToPKPi()[1];  /// non-prompt score
```
fgrosa, line 374: Typically we put the FD score in 3rd position (this is not mandatory, but it is what we usually do, so it would be good to have the same convention everywhere in the code)

Moreover, be careful that depending on the value of `selectionFlagLc` the vectors with BDT scores might be empty. You should add a protection (see e.g. https://github.com/AliceO2Group/O2Physics/blob/063c3880030ab07b026d4bf711ab10bdeef5caa7/PWGHF/D2H/Tasks/taskCharmPolarisation.cxx#L227-L251)

PWGHF/D2H/Tasks/taskLc.cxx
```diff
@@ -452,8 +551,27 @@ struct HfTaskLc {
...
           registry.fill(HIST("MC/reconstructed/nonprompt/hDecLenErrSigNonPrompt"), candidate.errorDecayLength(), pt);
         }
+        if constexpr (applyMl) {
+          double Mass(-1);
+          double outputBkg(-1);
+          double outputFD(-1);
+          if ((candidate.isSelLcToPKPi() >= selectionFlagLc) && pdgCodeProng0 == kProton) {
+            Mass = hfHelper.invMassLcToPKPi(candidate);
+            outputBkg = candidate.mlProbLcToPKPi()[0]; /// bkg score
+            outputFD = candidate.mlProbLcToPKPi()[1];  /// non-prompt score
```
fgrosa, line 561: Typically we put the FD score in 3rd position (see comment above)
Also same comment about possibly empty vectors (see above)

PWGHF/D2H/Tasks/taskLc.cxx
```diff
@@ -452,8 +551,27 @@ struct HfTaskLc {
...
+            /// Fill the ML outputScores and variables of candidate  (todo: add multiplicty)
+            registry.get<THnSparse>(HIST("hBdtScoreVsVars"))->Fill(Mass, pt, ptProng0, ptProng1, ptProng2, 0, chi2PCA, decayLength, decayLengthXY, cpa, cpaXY, outputBkg, outputFD, originType);
+          } else if ((candidate.isSelLcToPiKP() >= selectionFlagLc) && pdgCodeProng0 == kPiPlus) {
+            Mass = hfHelper.invMassLcToPiKP(candidate);
+            outputBkg = candidate.mlProbLcToPKPi()[0]; /// bkg score
+            outputFD = candidate.mlProbLcToPKPi()[1];  /// non-prompt score
```
fgrosa, line 567: I guess that this should be `mlProbLcToPiKP`
Typically we put the FD score in 3rd position (see comment above)
Also same comment about possibly empty vectors (see above)

fgrosa (changes requested):
Hi @zhangbiao-phy thanks a lot! I have few comments (the most important about the maximum number of axes), see below

PWGHF/D2H/Tasks/taskLc.cxx
zhangbiao-phy: Thank you so much for reminding me about this! The reason I keep so many vars in the Thn is that can still let the analyser do the further rectangular cut locally. Then it's a good option to reduce the vars in the ML Thn, and keep more in the Std Thn.

PWGHF/D2H/Tasks/taskLc.cxx
zhangbiao-phy: Done

PWGHF/D2H/Tasks/taskLc.cxx
zhangbiao-phy: yes, you are right! It's a mistake! Thanks for spotting this!

PWGHF/D2H/Tasks/taskLc.cxx
zhangbiao-phy: yes, I use one configurable in the new commit

zhangbiao-phy:
> Hi @zhangbiao-phy thanks a lot! I have few comments (the most important about the maximum number of axes), see below

Hi @fgrosa, Thanks for the review! Very helpful! I commit the changes based on your suggestion and comments.

PWGHF/D2H/Tasks/taskLc.cxx
```diff
@@ -37,10 +38,32 @@ struct HfTaskLc {
...
+  Configurable<bool> applyMl{"applyMl", false, "Flag to apply ML selections"};
+  ConfigurableAxis thnConfigAxisPt{"thnConfigAxisPt", {360, 0, 36}, ""};
+  ConfigurableAxis thnConfigAxisMass{"thnConfigAxisMass", {600, 1.98, 2.58}, ""};
+  ConfigurableAxis thnConfigAxisPtProng{"thnConfigAxisPtProng", {500, 0, 50}, ""};
+  ConfigurableAxis thnConfigAxisMultiplicity{"thnConfigAxisMultiplicity", {1000, 0, 1000}, ""};
+  ConfigurableAxis thnConfigAxisChi2PCA{"thnConfigAxisChi2PCA", {500, 0, 50}, ""};
+  ConfigurableAxis thnConfigAxisDecLength{"thnConfigAxisDecLength", {10, 0, 0.05}, ""};
+  ConfigurableAxis thnConfigAxisCPA{"thnConfigAxisCPA", {20, 0.8, 1}, ""};
+  ConfigurableAxis thnConfigAxisBdtScoreLc{"thnConfigAxisBdtScoreLc", {1000, 0., 1.}, ""};
+  ConfigurableAxis thnConfigAxisCanType{"thnConfigAxisCanType", {5, 0., 5.}, ""};
```
fgrosa, line 52: I would suggest less bins as default (since the current default version would create a very large histogram)
```suggestion
  ConfigurableAxis thnConfigAxisPt{"thnConfigAxisPt", {72, 0, 36}, ""};
  ConfigurableAxis thnConfigAxisMass{"thnConfigAxisMass", {300, 1.98, 2.58}, ""};
  ConfigurableAxis thnConfigAxisPtProng{"thnConfigAxisPtProng", {100, 0, 20}, ""};
  ConfigurableAxis thnConfigAxisMultiplicity{"thnConfigAxisMultiplicity", {100, 0, 1000}, ""};
  ConfigurableAxis thnConfigAxisChi2PCA{"thnConfigAxisChi2PCA", {100, 0, 20}, ""};
  ConfigurableAxis thnConfigAxisDecLength{"thnConfigAxisDecLength", {10, 0, 0.05}, ""};
  ConfigurableAxis thnConfigAxisCPA{"thnConfigAxisCPA", {20, 0.8, 1}, ""};
  ConfigurableAxis thnConfigAxisBdtScoreLc{"thnConfigAxisBdtScoreLc", {1000, 0., 1.}, ""};
  ConfigurableAxis thnConfigAxisCanType{"thnConfigAxisCanType", {5, 0., 5.}, ""};
```
I would also suggest you to define two axes for the BDT scores (since for the prompt/nonprompt score typically we need many less bins, i.e. 0.01 bin width is typically more than enough)

fgrosa (changes requested):
Hi @zhangbiao-phy thanks for implementing my comments! I have a couple of additional suggestions, then it's good for me.

PWGHF/D2H/Tasks/taskLc.cxx
```diff
@@ -235,11 +264,31 @@ struct HfTaskLc {
...
+      const AxisSpec thnAxisCanType{thnConfigAxisCanType, "candidates type"};
+
+      if (applyMl) {
+        registry.add("hnLcVarsWithBdt", "THn for Lambdac candidates", HistType::kTHnSparseF, {thnAxisMass, thnAxisPt, thnAxisMultiplicity, thnAxisBdtScoreLcBkg, thnAxisBdtScoreLcNonPrompt, thnAxisCanType});
+      } else {
+        registry.add("hnLcVars", "THn for Lambdac candidates", HistType::kTHnSparseF, {thnAxisMass, thnAxisPt, thnAxisMultiplicity, thnAxisPtProng0, thnAxisPtProng1, thnAxisPtProng2, thnAxisChi2PCA, thnAxisDecLength, thnAxisCPA, thnAxisCanType});
+      }
+    }
+  }
+  template <bool applyMl, typename CandType>
```
vkucera, line 290:
```suggestion
  }
  
  template <bool applyMl, typename CandType>
```

PWGHF/D2H/Tasks/taskLc.cxx
```diff
@@ -492,7 +650,20 @@ struct HfTaskLc {
     }
   }
 
-  PROCESS_SWITCH(HfTaskLc, processMc, "Process MC", false);
+  void
+    processMcStd(LcCandidatesMc const&, soa::Join<aod::McParticles, aod::HfCand3ProngMcGen> const& mcParticles,
+                 aod::TracksWMc const& tracksWithMc)
```
vkucera, line 655:
```suggestion
  void processMcStd(LcCandidatesMc const&, soa::Join<aod::McParticles, aod::HfCand3ProngMcGen> const& mcParticles,
                 aod::TracksWMc const& tracksWithMc)
```

fgrosa (approved):

vkucera:
Tagging @DelloStritto to keep him in the loop.

zhangbiao-phy:
> @zhangbiao-phy , your changes replaced the filter with partitions which broke the candidate grouping by collision. As a result, all candidates are now processed multiple times, once for each collision. To fix the grouping, please add slicing per collision index. Also, the added configurable `applyMl` is misleading, as it only activates histograms filling and does not apply any ML selection, which is only done in the selector. Please, always test your changes before making a PR.

Hi @vkucera, Thanks for spotting this! It's a mistake. actually, I have tested it on my laptop. I will make a PR soon.

PR #5041, opened 2024-03-07
Common | PWGHF: Take flavour oscillation into account at MC reco and gen levels

AlexBigO:
Hello @vkucera ! Here is my proposition to take B0(s) oscillation into account.

---------------------------------------

**The "issue" in a nutshell**

Let's suppose we aim at reconstructing the decay channel `B0bar -> D+ Pi-`. Currently, the code (at MC reco and gen levels) perfectly allows to do so by checking this decay channel with a set of pdg code signs `(-1 -> +1 -1)`. However, if B0 flavour oscillation occured, we actually have `B0 -> B0bar -> D+ Pi-`.
Here comes the "issue": in the genealogy of the decay products, the `B0bar` mother does not appear (when oscillation occurs), only the `B0` appears. In other words, the beauty meson after oscillation is "not seen" and one can only retrieve the pdg code of the beauty meson before oscillation, which of course leads to a rejection of the candidate as we look for `B0bar -> D+ Pi-` with a sign set `(-1 -> +1 -1)` while we only "see" `B0 -> D+ Pi-` with a sign set `(+1 -> +1 -1)`.
_Note: the problem lies in the sign of the beauty mother pdg code_

---------------------------------------

**The "fix" proposition in a nutshell**

Thanks to [Pythia status code](https://pythia.org/latest-manual/ParticleProperties.html), one can know if a decay product comes from an oscillated `B0(s)`, if so its code is equal to `92`.
So, in `RecoDecay.h`, we added a way to spot possible oscillation via this status code. If oscillation is indeed spotted, we swap the sign of the beauty mother.

Example: we have `B0 -> [B0bar "not seen"] -> D+ Pi-` but look for `B0bar -> D+ Pi-`
- without the functionality included in this PR, the B0 candidate is flagged as background (even though it is actually signal with flavour oscillation) 
- with the functionality included in this PR, if oscillation is spotted (thanks to the status code of the Dmeson and/or the bachelor pion), we swap the pdg code sign of the beauty mother (`B0` -> `B0bar`) and we then have `B0bar -> D+ Pi-`, which is exactly what we are looking for -> the B0bar candidate is flagged as signal

---------------------------------------

**Some safeties**

The code added in this PR is embedded in `template booleans` and such booleans are set to `false` by default so all the other decay channels calling the functions modified here in `RecoDecay.h` don't get affected. These booleans are only activated for the `B0` decay channel for the moment.
Tests were performed on the 3prong candidate creator, that is calling `getMatchedMcRec` and `isMatchedMcGen`: the output is indeed not affected by the code added in this PR.

### DIFF
--- a/Common/Core/RecoDecay.h
+++ b/Common/Core/RecoDecay.h
@@ -47,6 +47,8 @@ class RecoDecay
                     Prompt,
                     NonPrompt };
 
+  const uint8_t PdgStatusCodeAfterFlavourOscillation = 92;
+
   // Auxiliary functions
 
   /// Sums numbers.
@@ -581,7 +583,7 @@ class RecoDecay
       stage--;
     }
     if (sign) {
-      *sign = sgn;
+      *sign = particle.getGenStatusCode() != PdgStatusCodeAfterFlavourOscillation ? sgn : -sgn; // take possible flavour oscillation of B0(s) into account
     }
 
     return indexMother;
@@ -796,6 +798,7 @@ class RecoDecay
                              std::vector<int>* listIndexDaughters = nullptr)
   {
     // Printf("MC Gen: Expected particle PDG: %d", PDGParticle);
+    bool flagFlavourOscillation = false; // true if the B0(s) mother flavour oscillated
     int8_t sgn = 0; // 1 if the expected mother is particle, -1 if antiparticle (w.r.t. PDGParticle)
     if (sign) {
       *sign = sgn;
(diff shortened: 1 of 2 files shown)


vkucera (approved):

vkucera:
@jgrosseo @ddobrigk Can you please have a look and comment whether these missing intermediate B0 mothers are something that should be fixed in the AO2D?

fgrosa:
> @jgrosseo @ddobrigk Can you please have a look and comment whether these missing intermediate B0 mothers are something that should be fixed in the AO2D?

Hi @vkucera, I am not sure that in PYTHIA is produced at all (B0 oscillations are handled differently compared to K0 oscillations, which are described as decays into K0S and K0L, probably because of the lifetime and the many decay channels of B0 mesons and because we have no "B0S" and "B0L" equivalent)

Common/Core/RecoDecay.h
```diff
@@ -581,6 +583,11 @@ class RecoDecay
       stage--;
     }
     if (sign) {
+      if constexpr (acceptFlavourOscillation) {
+        if (std::abs(particle.getGenStatusCode()) == PdgStatusCodeAfterFlavourOscillation) { // take possible flavour oscillation of B0(s) mother into account
+          sgn *= -1;                                                                         // select the sign of the mother after oscillation (and not before)
+        }
+      }
       *sign = sgn;
```
vkucera, line 591: Why should the flip depend on whether we want to store the sign?

vkucera:
@jgrosseo @ddobrigk Does it look fine to you?

ddobrigk (approved):

PR #5044, opened 2024-03-07
PWGLF: event selection using border cuts added

ddobrigk:
@lhusova @romainschotter these are the new event selections based on ITS ROF border and TF border. Note we have now `evsel::MultSelection` stored in our derived data, so the use of evsel bits is pretty much the same as for regular AO2D analysis.


ercolessi (approved):
Hi @ddobrigk not sure but maybe in l. 658 and 710 you meant to use the other boolean? rejectTFBorder

PR #5047, opened 2024-03-07
Common/LF: add some extras for QA of evsel based on PF info

ddobrigk:
@lhusova @romainschotter following also a suggestion from @lhusova, this adds enough information for us to explore a correlation cut on Number of PV contributors and global (=ITSTPC, not necessarily PV contribs) tracks, just for a cross-check.


njacazio (approved):

alibuild (approved):

PR #5049, opened 2024-03-07
PWGHF: Make LUT optional in initCCDB

vkucera:
A quick fix to allow passing `nullptr` when LUT is not needed.


mfaggin (approved):
thanks @vkucera !

PR #5051, opened 2024-03-07
Add TOF selections + fix generated candidates loop

romainschotter:
Add TOF selections based on the deltaT info for positive and negative V0 daughters
Fix the loop on the generated candidates


njacazio (dismissed):

ddobrigk:
Hi @TimoWilken , sorry, seems like we need you again here .... When you get around to it, can you please take a look? I can't seem to spot the reason for the complaining of the linux compile, but I might be missing something.

TimoWilken:
Sorry for taking a while @ddobrigk, but it looks all green now!

ChiaraDeMartin95 (approved):

PR #5056, opened 2024-03-07
PWGCF: Identified Balanc Function Error fixes for PID

bghanley1995:
Fixed several errors with PID where code would crash without explanation. Also made the minimum value of Sigma for particle ID a configurable setting.


PWGCF/TwoParticleCorrelations/Tasks/identifiedbf.cxx
```diff
@@ -1330,6 +1329,6 @@ WorkflowSpec defineDataProcessing(ConfigContext const& cfgc)
 {
   WorkflowSpec workflow{
     adaptAnalysisTask<IdentifiedBfCorrelationsTask>(cfgc, TaskName{"IdentifiedBfCorrelationsTaskRec"}, SetDefaultProcesses{{{"processRecLevel", true}, {"processRecLevelMixed", false}, {"processCleaner", false}}}),
-    adaptAnalysisTask<IdentifiedBfCorrelationsTask>(cfgc, TaskName{"IdentifiedBfCorrelationsTaskGen"}, SetDefaultProcesses{{{"processGenLevel", true}, {"processGenLevelMixed", false}, {"processCleaner", false}}})};
+    adaptAnalysisTask<IdentifiedBfCorrelationsTask>(cfgc, TaskName{"IdentifiedBfCorrelationsTaskGen"}, SetDefaultProcesses{{{"processGenLevel", false}, {"processGenLevelMixed", false}, {"processCleaner", true}}})};
```
victor-gonzalez, line 1332: You don't necessarily need to do this at the source code level
You can modify it either in your configuration file when you run it locally or in the wagon configuration when you run it on hyperloop
But anyway, it does not harm

victor-gonzalez (approved):

PWGCF/TwoParticleCorrelations/Tasks/identifiedbf.cxx
bghanley1995: Thanks for the advice Victor! I'm still getting used to the O2 framework and the wagon configuration.

PR #5061, opened 2024-03-08
PWGHF: removing index from FullPs tree

cterrevo:
fixing issue of unbound indices in XicpKpi MC trees


fcatalan92 (approved):
Hi @cterrevo, thanks a lot for having addressed my comments

PR #5064, opened 2024-03-08
PWGHF: Add DcaXY selection to Ds selector

fchinu:
Hello @fgrosa !

As we discussed, I have implemented the possibility of having a selection on the Ds prongs' DcaXY in the candidate selector.
The default configurables values are the same as the ones for the track-index-skim-creator.


fgrosa (approved):
Hi @fchinu thanks! Indeed when we discussed offline I did not think that this cut is as a function of the track pT, so it does not modify the matrix of topological cuts, and therefore there is no issue.

PR #5065, opened 2024-03-08
PWGHF: Replace extended TPC+TOF tables with tables produced by PID creator

vkucera:
I kept the extended-table data model commented out in hope of future possibility of conditional spawning.


fgrosa (approved):
It looks good to me, thanks!

alibuild (approved):

PR #5068, opened 2024-03-08
[event plane] add safety for ccdb connection

jikim1290:
add safety for ccdb connection


ddobrigk (approved):
Thanks a lot @jikim1290 ! :-D

PR #5082, opened 2024-03-10
[PWGDQ] A first version of MillePede2 record creation task for muon alignment

Elros60:
This first version should be used only locally at this stage: it can be used to generate `MillePede2` records for alignment workflow and save residuals for some QC checks. Since this current working version is not yet considered as complete, it would be better at this moment to put it under `O2Physics/PWGDQ/Tasks/`

The on-going development will allow people to apply any aligned geometry for muon tracks and then use `table-maker`+`table-maker` to do checks. Two difficulties are currently under discussion: 1) How to pass modified tracks information to `table-maker`? Use another table? 2) If one needs to check with any aligned parameters they want, there should be a proper user-defined CCDB path or only officially submitted aligned geometry are allowed for test?

### DIFF
--- a/Common/TableProducer/qVectorsTable.cxx
+++ b/Common/TableProducer/qVectorsTable.cxx
@@ -163,74 +163,80 @@ struct qVectorsTable {
     if (cfgCCDBConst == 1) {
       fullPath = cfgQvecCalibPath;
       fullPath += "/FT0C";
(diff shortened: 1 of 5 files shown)


iarsene (approved):

iarsene:
Hi @Elros60, 
Unfortunately I see only now that with your commit you reverted some changes to the CMakeLists.txt file. This means you did not rebased before you pushed your changes. Please do not do this next time. Always rebase before you commit.
Cheers,
Ionut

PR #5083, opened 2024-03-10
PWGLF: V0 builder preselector optimization

ddobrigk:
This PR re-arranges the pre-selector logic of the V0 builder such that it is geared towards faster processing, also based on observations discussed with @ZFederica and @fgrosa regarding the less than ideal speed of the V0 builer in special use cases. The new pre-selector logic uses bitmask checking for a significant reduction of `if` calls and comparisons when processing `aod::V0s` candidates (before distance minimization calls)

Since this is a relatively non-trivial change of a key component used in many analyses, some further testing will be done before merging.

### DIFF
--- a/PWGLF/DataModel/LFStrangenessTables.h
+++ b/PWGLF/DataModel/LFStrangenessTables.h
@@ -586,9 +586,9 @@ enum v0bit {
   bitdEdxNegProton,
   bitdEdxNegHelium3,
 
-  // Mixed bits 
-  bitdEdxPosPionOrITSOnly, 
-  bitdEdxNegPionOrITSOnly, 
+  // Mixed bits
+  bitdEdxPosPionOrITSOnly,
+  bitdEdxNegPionOrITSOnly,
 
   // Monte Carlo Association bits
   bitTrueGamma,
--- a/PWGLF/DataModel/LFStrangenessTables.h
+++ b/PWGLF/DataModel/LFStrangenessTables.h
@@ -565,85 +565,81 @@ using V0Linked = V0sLinked::iterator;
 // helper for building
 namespace v0tag
 {
-namespace enums{
-  enum v0bit { 
-               // Track quality 
-               bitTrackPosGoodTPC,
-               bitTrackPosGoodITSOnly,
-               bitTrackPosGoodTPCorITSOnly,
-               bitTrackNegGoodTPC,
-               bitTrackNegGoodITSOnly,
-               bitTrackNegGoodTPCorITSOnly,
+namespace enums
+{
+enum v0bit {
+  // Track quality
+  bitTrackPosGoodTPC,
+  bitTrackPosGoodITSOnly,
+  bitTrackPosGoodTPCorITSOnly,
+  bitTrackNegGoodTPC,
+  bitTrackNegGoodITSOnly,
+  bitTrackNegGoodTPCorITSOnly,
 
-               // dEdx compatibility bits
-               bitdEdxPosElectron,
-               bitdEdxPosPion,
-               bitdEdxPosProton,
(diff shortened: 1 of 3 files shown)


ddobrigk:
This PR re-arranges the pre-selector logic of the V0 builder such that it is geared towards faster processing, also based on observations discussed with @ZFederica and @fgrosa regarding the less than ideal speed of the V0 builer in special use cases. The new pre-selector logic uses bitmask checking for a significant reduction of `if` calls and comparisons when processing `aod::V0s` candidates. 

Since this is a relatively non-trivial change of a key component used in many analyses, some further testing will be done before merging.

PWGLF/DataModel/LFStrangenessTables.h
```diff
@@ -565,45 +565,103 @@ using V0Linked = V0sLinked::iterator;
...
+DECLARE_SOA_DYNAMIC_COLUMN(IsTrueLambda, isTrueLambda, //!
+                           [](uint32_t selectionMap) -> bool { return (selectionMap & (1 << enums::bitTrueLambda)) == (1 << enums::bitTrueLambda); });
+DECLARE_SOA_DYNAMIC_COLUMN(IsTrueAntiLambda, isTrueAntiLambda, //!
+                           [](uint32_t selectionMap) -> bool { return (selectionMap & (1 << enums::bitTrueAntiLambda)) == (1 << enums::bitTrueAntiLambda); });
+DECLARE_SOA_DYNAMIC_COLUMN(IsTrueHypertriton, isTrueHypertriton, //!
+                           [](uint32_t selectionMap) -> bool { return (selectionMap & (1 << enums::bitTrueHypertriton)) == (1 << enums::bitTrueHypertriton); });
+DECLARE_SOA_DYNAMIC_COLUMN(IsTrueAntiHypertriton, isTrueAntiHypertriton, //!
+                           [](uint32_t selectionMap) -> bool { return (selectionMap & (1 << enums::bitTrueAntiHypertriton)) == (1 << enums::bitTrueAntiHypertriton); });
+
+DECLARE_SOA_COLUMN(Tag, tag, uint32_t); // pack according to enums
```
aalkin, line 644: There is DECLARE_SOA_BITMAP_COLUMN, that hold an unsigned int (of 8, 16, 32 or 64 bit) and has `getter_bit(pos)` defined that returns bool. If you make the enum convertible to int, you can just do `if (tag_bit(bitTrueLambda))` for example.

ddobrigk: Yes, that's a good point, indeed, I'll move to that shortly :-D thanks!

ddobrigk: Hi @aalkin one curiosity though: this will just offer an interface for checking the bits, right? There is no inherent speed advantage in this, right?

aalkin: Dynamic columns are not very fast in general, as they contain lambdas, however in this case those are probably optimized out by the compiler so that there is no performance overhead. But you can skip adding a bunch of dynamic columns in the data model and the syntax becomes a bit simpler.

ddobrigk: Now implemented! Thanks :-D

ddobrigk:
With the latest O2 - which is, in fact, faster since last week! - the new optimizations done here, in fact, do not improve the speed terribly. My own private measurements indicate a measurable but rather small increase in speed, pushing the urgency of this PR back a little. Still, the preselector is significantly improved in this version of the code from the algorithmic point of view and is much cleaner, so I would advocate that this still be merged once all cross-checks for correctness are successfully concluded...

PR #5091, opened 2024-03-11
K892 update

Sayan-Dhani:
Added the analysis code for K*(892)_Spherocity_Analysis; also made changes in the CMake list


BongHwi:
Hello @Sayan-Dhani ,

Thanks for the contribution, could you try to solve the issue on the Headers and MegaLinter?

BongHwi (approved):

PR #5093, opened 2024-03-11
Update for UPC analysis chain

nburmaso:
* Update MC part of the UPCCandidateProducer
* Update veto efficiency analysis (amplidude-based V0A/T0A vetoes, analysis per ZDC class)

### DIFF
--- a/PWGUD/TableProducer/UPCCandidateProducer.cxx
+++ b/PWGUD/TableProducer/UPCCandidateProducer.cxx
@@ -279,7 +279,7 @@ struct UpcCandProducer {
         continue;
       int32_t mcEventID = mcPart.mcCollisionId();
       const auto& mcEvent = mcCollisions.iteratorAt(mcEventID);
-      bool isSignal = mcEvent.generatorsID() == fSignalGenID;
+      bool isSignal = mcEvent.getSourceId() == fSignalGenID;
       if (!isSignal) {
         continue;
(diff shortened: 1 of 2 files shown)


PWGUD/TableProducer/UPCCandidateProducer.cxx
```diff
@@ -291,43 +291,48 @@ struct UpcCandProducer {
...
-          continue;
-        }
-        auto itFirst = fNewPartIDs.find(firstDaughter);
-        auto itLast = fNewPartIDs.find(lastDaughter);
-        if (itFirst != fNewPartIDs.end() && itLast != fNewPartIDs.end()) {
-          newDaughterIDs[0] = fNewPartIDs.at(daughterIDs.front());
-          newDaughterIDs[1] = fNewPartIDs.at(daughterIDs.back());
+      if (mcPart.has_daughters()) {
+        const auto& daughterIDs = mcPart.daughtersIds();
+        if (daughterIDs.size() > 0) {
```
aalkin, line 319: Note that daughters index is a slice index, the ids always return `int[2]`, its size is always 2. It is enough to check with `has_daughters()` that checks for the both numbers to be non-negative.

nburmaso: Thank you, Anton!

PR #5094, opened 2024-03-11
PWGCF: Mean pt fluc mc update

TGahlaut1:


victor-gonzalez (changes requested):
Why do you need a completely separated source file for analyzing MC events?

What it is usually done in O2Physics is 
- at it should, MC reconstructed should be analyzed with the same code, task, as for data
- the fact that MC reconstructed has associated MC collisions and MC particles, is usually handled by templated methods or by checking the presence of the corresponding table columns  
- MC generator level usually only needs to be considered at the subscription level of the corresponding, specific, methods of the same overall task which can be activated or let passive at configuration level, either from on the configuration file or on the wagon configuration in hyperloop
- most of the method used for the reconstructed level, either MC or data, can also be used at MC generator level just properly tmplating them   

Otherwise you are force to have basically two very similar source files which is error prone and difficult to maintain

PR #5103, opened 2024-03-11
Strangeness filter: fill table even when rejecting events

mpuccio:


alibuild (approved):

PR #5105, opened 2024-03-12
PWGCF: Update MeanPtFlucIdentified.cxx

TGahlaut1:


victor-gonzalez (commented):
It looks as if the code for processing reco data and reco MC is fully replicated
Also the code for processing MC generator level is mostly a replication of the previous ones

Is this really needed? Replicating code is error prone and difficult to maintain in the medium term

TGahlaut1:
> It looks as if the code for processing reco data and reco MC is fully replicated Also the code for processing MC generator level is mostly a replication of the previous ones
> 
> Is this really needed? Replicating code is error prone and difficult to maintain in the medium term

No it's not exactly same. There are a few minor changes from data to recoMC.

victor-gonzalez (approved):
OK!

I would suggest to simplify it by 95% but ...

PR #5121, opened 2024-03-12
Always fill processed event histo for accounting

mpuccio:


alibuild (approved):

PR #5149, opened 2024-03-13
PWGLF: Fix decoding of generated information from derived AO2Ds

ddobrigk:
@romainschotter (and @lhusova) this fixes the decoding of the generated-level information that is stored encoded in our MC generated data, pending a proper solution in which we store some binned information together for convenience. It has been tested to work locally, though the main caveat (a **very** serious one) is that we have to use the same binning as the derived AO2D generation. 

@lhusova note that in fact our task decodes the full generated information *also* for the cascades, so this could be useful for you too if you like (or you can copy the relevant part)


ddobrigk:
In the latest commit, I added a check in which the size of the incoming generated array is compared with the number of cells of the receiving histogram, and if these don't match, a `fatal` is issued. This should capture essentially all mistakes, although it will make it impossible to change the pT binning in the generated histograms natively in the derived data analysis for now (unless we create separate, possibly hardcoded/non-configurable (sigh) axes for that).

ercolessi (approved):

PR #5162, opened 2024-03-14
[AOTTrack] added a timeframe cut on collisions

clotteau:
Added a configurable bool to apply a timeframe cut on collisions

### DIFF
--- a/DPG/Tasks/AOTTrack/qaEventTrack.cxx
+++ b/DPG/Tasks/AOTTrack/qaEventTrack.cxx
@@ -61,6 +61,9 @@ struct qaEventTrack {
   // options to select specific events
   Configurable<bool> selectGoodEvents{"selectGoodEvents", true, "select good events"};
 
+  // option to apply a timeframe cut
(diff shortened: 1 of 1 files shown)


DPG/Tasks/AOTTrack/qaEventTrack.cxx
```diff
@@ -61,6 +61,9 @@ struct qaEventTrack {
   // options to select specific events
   Configurable<bool> selectGoodEvents{"selectGoodEvents", true, "select good events"};
 
+  // option to apply a timeframe cut
+  Configurable<bool> tfCut{"tfCut", true, "applies timeframe cut"};
```
mfaggin, line 65: I would put this to `false` as default value, for backward compatibility. If we decide then to switch it on as default in our analyses, then we'll change the configuration

mfaggin (changes requested):
hi @clotteau , many thanks this is indeed what we discussed! I just have 1 request more, see the detailed comment.

mfaggin (approved):
thank you @clotteau !

PR #5172, opened 2024-03-14
PWGEM/PhotonMeson: Add event selection cutstring and calo readout configurable

nstrangm:
- Add minbias_noTBForROFCut event cutstring for MB events without ITSROF or TBF restrictions
- Query event cut settings (NoTFB and ITFROFB) in EMEventCut.h
- Add requireCaloReadout configurable to circumvent trigger requirement (wrong in certain MCs)
- Fix bin labels in emcQC task

### DIFF
--- a/PWGEM/PhotonMeson/Core/CutsLibrary.cxx
+++ b/PWGEM/PhotonMeson/Core/CutsLibrary.cxx
@@ -44,7 +44,7 @@ EMEventCut* o2::aod::pwgem::photon::eventcuts::GetCut(const char* cutName)
   if (!nameStr.compare("minbias")) {
     cut->SetRequireFT0AND(true);
     cut->SetZvtxRange(-10.f, +10.f);
-    cut->SetRequireNoTFB(true);
+    cut->SetRequireNoTFB(false);
     cut->SetRequireNoITSROFB(false);
     return cut;
   }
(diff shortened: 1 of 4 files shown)


PWGEM/PhotonMeson/Core/CutsLibrary.cxx
```diff
@@ -57,6 +57,14 @@ EMEventCut* o2::aod::pwgem::photon::eventcuts::GetCut(const char* cutName)
     return cut;
   }
 
+  if (!nameStr.compare("minbias_noTBForROFCut")) {
```
dsekihat, line 60: Let's put like this:
"minbias" as sel8 and |Zvtx| < 10 cm only.
"minbias_notfb" as sel8 and |Zvtx| < 10 cm and No TFB.
"minbias_notfb_noitsrofb" as sel8 and |Zvtx| < 10 cm and No TFB and No ITS ROFB.

PWGEM/PhotonMeson/Core/EMEventCut.h
```diff
@@ -61,17 +61,17 @@ class EMEventCut : public TNamed
   {
     switch (cut) {
       case EMEventCuts::kFT0AND:
-        return collision.sel8();
+        return (!mRequireFT0AND || collision.sel8());
```
dsekihat, line 64: can you move mRequireFT0AND to line 44?
 if (mRequireFT0 && !IsSelected(collision, EMEventCuts::kFT0AND)) {
 same for requireNoTFB and requireNoITSROFB.

PWGEM/PhotonMeson/TableProducer/createEMReducedEvent.cxx
```diff
@@ -42,6 +42,8 @@ struct CreateEMEvent {
   Produces<o2::aod::EMReducedEventsMult> event_mult;
   Produces<o2::aod::EMReducedEventsCent> event_cent;
 
+  Configurable<bool> requireCaloReadout{"requireCaloReadout", true, "Require calorimeters readout when analyzing EMCal/PHOS"};
```
dsekihat, line 45: This is not what one should do. readout flag is stored in CCDB. This is handled by Evgeny Kryshen. If this CCDB is missing, can you ask him?

PWGEM/PhotonMeson/Core/CutsLibrary.cxx
nstrangm: Yes, I agree. I implemented it in the updated version

PWGEM/PhotonMeson/Core/EMEventCut.h
nstrangm: Yes, that seems better. I implemented it in the update

PWGEM/PhotonMeson/TableProducer/createEMReducedEvent.cxx
nstrangm: You are right, this should not be necessary. However, the kTVXinEMC alias bit is unfortunately not correctly set in multiple MCs, so we would like to implement this bypass to be able to investigate the full statistics

dsekihat (approved):

PR #5179, opened 2024-03-15
PWGLF: minor strangeness builder changes

ddobrigk:
* one more attempt at fixing generated encoding 
* to be added tomorrow: Lucia's suggested Nglobal correlation variables for CF-like event rejection studies


ddobrigk:
The problem was elsewhere. Closing

PR #5141, opened 2024-03-13
PWGHF: bitmask for ev. selection and collision process in candidate creators.

mfaggin:

PR #5057, opened 2024-03-08
PWGHF: add sel8, PV posZ and TF border cut selections in candidateCreators.

mfaggin:

PR #5186, opened 2024-03-15
PWGHF: Adding event selection in the correlatorDsHadrons task

scattaru:
Dear all,
I've modified the correlator in order to select collisions with the sel8 flag

Cheers,
Samuele

### DIFF
--- a/PWGHF/HFC/TableProducer/correlatorDsHadrons.cxx
+++ b/PWGHF/HFC/TableProducer/correlatorDsHadrons.cxx
@@ -161,7 +161,7 @@ struct HfCorrelatorDsHadrons {
   ConfigurableAxis binsEta{"binsEta", {100, -2., 2.}, "#it{#eta}"};
   ConfigurableAxis binsPhi{"binsPhi", {128, -PIHalf, 3. * PIHalf}, "#it{#varphi}"};
   ConfigurableAxis binsMultiplicity{"binsMultiplicity", {200, 0., 800.}, "Multiplicity"};
-  ConfigurableAxis binsMultFT0M{"binsMultFT0M", {1000, 0., 10000.}, "Multiplicity as FT0M signal amplitude"};
+  ConfigurableAxis binsMultFT0M{"binsMultFT0M", {600, 0., 6000.}, "Multiplicity as FT0M signal amplitude"};
   ConfigurableAxis binsPosZ{"binsPosZ", {100, -10., 10.}, "primary vertex z coordinate"};
   ConfigurableAxis binsPoolBin{"binsPoolBin", {9, 0., 9.}, "PoolBin"};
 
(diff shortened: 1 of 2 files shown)


fgrosa:
> Dear all, I've modified the correlator in order to select collisions with the sel8 flag
> 
> Cheers, Samuele

Hi @scattaru, in principle the event selection can already be applied at the level of the `CandidateCreator3Prong` now (see the PRs by @mfaggin, see https://github.com/AliceO2Group/O2Physics/pull/5057 and for the counter https://github.com/AliceO2Group/O2Physics/pull/5141). Do you think you would still need to add it afterwards?

scattaru:
> > Dear all, I've modified the correlator in order to select collisions with the sel8 flag
> > Cheers, Samuele
> 
> Hi @scattaru, in principle the event selection can already be applied at the level of the `CandidateCreator3Prong` now (see the PRs by @mfaggin, see #5057 and for the counter #5141). Do you think you would still need to add it afterwards?

Hi @fgrosa, in my case I need to filter the collision table that is used for the event mixing in order to take only the collisions which satisfies the selection criteria. So the idea is to join the collision table with the one which provides the event selection information. As far as I've understood in the creator there are just histograms filled with the information of selected collisions not a table produced with that information.

fgrosa:
> > > Dear all, I've modified the correlator in order to select collisions with the sel8 flag
> > > Cheers, Samuele
> > 
> > 
> > Hi @scattaru, in principle the event selection can already be applied at the level of the `CandidateCreator3Prong` now (see the PRs by @mfaggin, see #5057 and for the counter #5141). Do you think you would still need to add it afterwards?
> 
> Hi @fgrosa, in my case I need to filter the collision table that is used for the event mixing in order to take only the collisions which satisfies the selection criteria. So the idea is to join the collision table with the one which provides the event selection information. As far as I've understood in the creator there are just histograms filled with the information of selected collisions not a table produced with that information.

Thanks for the reply! The creator also rejects candidates in events that do not fulfil the event selection, so you will not find Ds candidates in those events, but probably here you want to discard also the tracks in those events. If this is the case indeed it makes sense and we can merge it.

scattaru:
> > > > Dear all, I've modified the correlator in order to select collisions with the sel8 flag
> > > > Cheers, Samuele
> > > 
> > > 
> > > Hi @scattaru, in principle the event selection can already be applied at the level of the `CandidateCreator3Prong` now (see the PRs by @mfaggin, see #5057 and for the counter #5141). Do you think you would still need to add it afterwards?
> > 
> > 
> > Hi @fgrosa, in my case I need to filter the collision table that is used for the event mixing in order to take only the collisions which satisfies the selection criteria. So the idea is to join the collision table with the one which provides the event selection information. As far as I've understood in the creator there are just histograms filled with the information of selected collisions not a table produced with that information.
> 
> Thanks for the reply! The creator also rejects candidates in events that do not fulfil the event selection, so you will not find Ds candidates in those events, but probably here you want to discard also the tracks in those events. If this is the case indeed it makes sense and we can merge it.

Yes, exactly. In the Event Mixing the problem we faced is the limit number of Ds meson selected per DF, so there are DFs with only one Ds reconstructed so it was impossible to do the event mixing. To avoid this issue we decided to use all collisions and to mix an event with a candidate reconstructed with particles coming from other events even if they have not a candidate reconstructed. For the moment this is the fastest way, however we are also exploring other attempts to do the event mixing.

fgrosa (approved):

PR #5193, opened 2024-03-15
Common: fix forgotten line

ddobrigk:
@romainschotter @lhusova I was looking at the wrong place for my bug. It was a silly mistake in an auxiliary tool. Now the decoding of generated histograms works fine!


alibuild (approved):

PR #5204, opened 2024-03-16
V0: use collinear fit from DCAFitter

f3sch:
For photons the position is better resolved using a specialized collinear method in DCAFitter.


f3sch:
Linter fails with:
```
  --Error detail:
  PWGLF/DataModel/LFStrangenessTables.h:762:  Add #include <vector> for vector<>  [build/include_what_you_use] [4]
```

Since the changes to this file are unrelated to this, should I actually make this change?

f3sch:
Hi @ddobrigk, thanks for taking the time. I will implement your suggestion for the additional switch. For now it is TPC-only only :). BTW why does CI fail here or rather I was under the impression that always the new nightly tag would be available but from the build log it seems to be the O2 version is daily-20240312-0130-local?

f3sch:
@njacazio can this please be merged? cheers

fmazzasc (approved):

PR #5219, opened 2024-03-18
PWGEM/PhotonMeson: reduce skimmed MC data size

dsekihat:
reduce skimmed MC data size


alibuild (approved):

PR #5229, opened 2024-03-18
[PWGCF] Fix for crash in singles execution when access to CCDB

victor-gonzalez:


alibuild (approved):

PR #5230, opened 2024-03-18
[PWGDQ] Fixing var manager for flow analysis

lucamicheletti93:
Fixing varManager for flow analysis using the central framework


alibuild (approved):

PR #5233, opened 2024-03-19
PWGMM: Mult: install headers

aalkin:
* Install mult headers
* Add target O2Physics::MultCore to use the headers


alibuild (approved):

PR #5237, opened 2024-03-19
[PWGDQ] Minor fix in the EP calculation

lucamicheletti93:
Minor fix in the EP calculation


alibuild (approved):

PR #5243, opened 2024-03-19
EventFiltering/PWGHF: fix selection of soft pion from SigmaC

fgrosa:


alibuild (approved):

PR #5248, opened 2024-03-19
PWGHF: Macros for data preparation and ML training

AlexBigO:
Hello @fcatalan92 @mfaggin ! 

With @Mingyu3360715 we wrote some macros for data preparation and ML training (along with configuration files).

As many other members of the PAG are using BDTs (with `hipe4ml`), we thought it would be nice to have some "official D2H" common macros. Of course, this is not to impose any format to the members of the PAG, but rather to humbly propose a "starter pack" for newcomers or people that would like to avoid transposing Run 2 macros to Run 3 standards.

Besides, the conventions chosen in these macros (especially at the data preparation level) might not fit the conventions set for every decay channels in the `treeCreator*.cxx` tasks for instance, so this is once again to provide an example of data preparation that could be then adapted to other conventions.

Your  input is of course welcome if you think some functionalities are missing.

-----------------------
Data preparation:
- `prepare_samples.py`: a python macro based on `RDataFrame` (way faster than `pandas`) to apply local preselections and produce `.root` files with classes tags (to be used for training)
- `config_preparation.yml`: a config file for the data preparation macro (in the example of D+)

-----------------------
Training:
- `train_d2h.py`: a pyton macro performing training and testing
- `config_training.yml`: a config file for the training macro  (in the example of D+)

The way to use these macros and some extra information are provided in the `README.md`.

-----------------------
PS:
- currently, we do not provide code to locally apply BDT models as we believe there is already a great amount of code for one PR


fcatalan92:
Hi @AlexBigO, @Mingyu3360715, thanks a lot for these developments! Indeed, it would be nice to have common macros for the training of BDTs. I will have a look at them but with a slower peace given the approval period.

My first comment is that nothing prevents `PrepareSamples.C` to be a python macro using the PyROOT bindings. In python, the gymnastic needed to read json files should also be easier.

PWGHF/D2H/Macros/ML/train_d2h.py
```diff
@@ -0,0 +1,544 @@
...
+# This software is distributed under the terms of the GNU General Public
+# License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+
+# In applying this license CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+
+"""
+file: train_d2h.py
```
fcatalan92, line 16: I would use something more explicit as file name, e.g. `train_models.py`

PWGHF/D2H/Macros/ML/train_d2h.py
```diff
@@ -0,0 +1,544 @@
...
+import pandas as pd  # pylint: disable=import-error
+import xgboost as xgb  # pylint: disable=import-error
+import yaml  # pylint: disable=import-error
+
+# pylint: disable=import-error
+from hipe4ml import plot_utils
+from hipe4ml.model_handler import ModelHandler
+from hipe4ml.tree_handler import TreeHandler
+from hipe4ml_converter.h4ml_converter import H4MLConverter
+from sklearn.model_selection import train_test_split
```
fcatalan92, line 40: Since`hipe4ml` is not installed by default with O2Physics, you could start by trying to import it with something like this
```python
try:
    import hipe4ml
except ModuleNotFoundError:
    print("module 'hipe4ml' is not installed. Please install it to run this macro")
```
The other dependencies should be bundled with `hipe4ml` (to be checked for `yaml`).

PWGHF/D2H/Macros/ML/train_d2h.py
```diff
@@ -0,0 +1,544 @@
...
+            if self.indirs[cand_type] is None:
+                continue
+            for indir in self.indirs[cand_type]:
+                file = os.path.join(indir, f"{cand_type}_{self.channel}.{self.infile_extension}")
+                if os.path.isfile(file):
+                    file_lists[cand_type].append(file)
+                else:
+                    print(
+                        "\033[91mERROR: missing file, did you prepare the samples?\n"
+                        "If not, do python3 prepare_samples.py CONFIG\033[0m"
```
fcatalan92, line 177: This conflicts with the current macro for preparing the samples, which is `PrepareSamples.C`.

PWGHF/D2H/Macros/ML/train_d2h.py
```diff
@@ -0,0 +1,544 @@
...
+            x[i] = element.strip()
+
+    return x
+
+
+# pylint: disable= too-many-instance-attributes, too-few-public-methods
+class MlTrainer:
+    """
+    Class for ML training and testing
+    """
```
fcatalan92, line 75: Is a class needed? It looks to me that the code can be written as a few functions and a main

PWGHF/D2H/Macros/ML/README.md
```diff
@@ -0,0 +1,35 @@
+# Macros for ML training in D2H
+## Requirements
+In order to execute the training macro in this folder, the following python libraries are needed:
+- [hipe4ml](https://github.com/hipe4ml/hipe4ml)
+- [hipe4ml_converter](https://github.com/hipe4ml/hipe4ml_converter)
+- [alive_progress](https://github.com/rsalmei/alive-progress)
```
fcatalan92, line 6: `alive_progress` is not used in the current script

fcatalan92 (changes requested):
Hi @AlexBigO, @Mingyu3360715, I finally managed to have a look at the PR. The code looks already in good shape to me! I added some comments and suggestions line by line.

As a general comment, I would suggest you to stick to one language (python) for these macros and to use `yml` configuration files (since they support comments). In this way, it is easier to use and maintain them.

PWGHF/D2H/Macros/ML/PrepareSamples.C
```diff
@@ -0,0 +1,259 @@
...
+    chain.Add(pathInputTree);
+
+    // define dataframe from the input TTrees
+    ROOT::EnableImplicitMT(32); // tell ROOT to go parallel
+    ROOT::RDataFrame dataFrame(chain);
+
+    std::vector<std::string> colsToKeep = dataFrame.GetColumnNames();
+    if (colsToRemove.size() == 0) {
+      std::remove(colsToKeep.begin(), colsToKeep.end(), "fOriginMcRec"); // just remove cand_type if not explicitly kept
+      colsToKeep.pop_back();
```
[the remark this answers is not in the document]
AlexBigO: Indeed, this should not be there! It must be a remnant of a previous version of the code.

PWGHF/D2H/Macros/ML/PrepareSamples.C
```diff
@@ -0,0 +1,259 @@
...
+    }
```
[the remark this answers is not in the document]
AlexBigO: This has been fixed a the new version I am developing

PWGHF/D2H/Macros/ML/train_d2h.py
AlexBigO: This is solved with the new (python) version of the preparatio script

PWGHF/D2H/Macros/ML/train_d2h.py
AlexBigO: Actually, when I started writing the code I did using functions only but as I was progressing I felt that it was cleaner to access some information via class attributes rather than by having duplicated function arguments. Hence the class formalism

AlexBigO:
> Hi @AlexBigO, @Mingyu3360715, I finally managed to have a look at the PR. The code looks already in good shape to me! I added some comments and suggestions line by line.
> 
> As a general comment, I would suggest you to stick to one language (python) for these macros and to use `yml` configuration files (since they support comments). In this way, it is easier to use and maintain them.

Hi @fcatalan92 ! Thanks a lot for the comments, the `.C` macro was converted to `.py` leaving aside all unused/not needed functionalities, and going from `.json` to `.yml` for the configuration.

vkucera:
@AlexBigO , for the record, you can just merge the linked automatic MegaLinter PR to do the formatting.

AlexBigO:
> @AlexBigO , for the record, you can just merge the linked automatic MegaLinter PR to do the formatting.

Hi Vit! Indeed :D I applied the modification locally because the MegaLinter PR did not pass all the tests (but I later realised that it was because of other files than those concerned by this PR).

fcatalan92 (approved):
Hi @AlexBigO, thanks a lot for addressing my comments! All looks good to me and I imagine you run the scripts and tested them thoroughly. Now merging.

AlexBigO:
> Hi @AlexBigO, thanks a lot for addressing my comments! All looks good to me and I imagine you run the scripts and tested them thoroughly. Now merging.

Hi @fcatalan92, thanks! Yes :) I tested the "new" preparation macro with the different options in the config file. The training macro had already been tested to train BDT models for both binary and 3-class classification.

PR #5264, opened 2024-03-20
Set different onnx paths for Xis and Omegas and fix bug by switching signal and background

ChiaraDeMartin95:


alibuild (dismissed):

ercolessi (approved):

PR #5265, opened 2024-03-20
PWGLF: add TOF N-sigma for strangeness

ddobrigk:
This PR adds an N-sigma calculation path for the strangeness TOF based on the latest Pb-Pb reconstruction. Being tested for the time being, kept as draft until confirmed @njacazio


ddobrigk:
@romainschotter @lhusova This is ready to fly for V0s! We can then cut simply on TOF Nsigmas. The cascade path still needs #5290 cause I am very typo-prone but hopefully over the weekend I will also finalise that...

njacazio (approved):

PR #5275, opened 2024-03-21
PWGMM: Mult: add reduced data model and a task

aalkin:
* Reduced data model with collision, MC collision, track and particles information
* Task that produces collision and MC collision sampled trees

This is a task to create sampled trees of collision multiplicity information. Tail of multiplicity distribution is kept as is, while the bulk of the collisions are randomly discarded according to configured threshold. The rest are added with corresponding weight. The trees are intended to be small in size and used for ML studies.


alibuild (approved):

PR #5280, opened 2024-03-21
PWGHF: Add missing line to DplusPi creator

AlexBigO:
Hi @fgrosa , here is the fix :)

Note: this line was only missing at the "DplusPi creation" level, not at the "D0Pi creation" level


fgrosa:
> Hi @fgrosa , here is the fix :)
> 
> Note: this line was only missing at the "DplusPi creation" level, not at the "D0Pi creation" level

Thanks for spotting it!

fgrosa (approved):

PR #5282, opened 2024-03-21
PWGHF: explicit type-cast for whyRejectColl.

mfaggin:
@stefanopolitano is experiencing the following run-time error when using the `track-index-skim-creator` for local debugging:

```
[FATAL] Unhandled o2::framework::runtime_error reached the top of main of o2-analysis-hf-track-index-skim-creator, device shutting down. Reason: Failed to create filter: ExpressionValidationError: Function int32 castINT(uint16) not supported yet.
```

Probably, the data-type change done here https://github.com/AliceO2Group/O2Physics/commit/c8ea8cad372783aea1a9396bb4acd47d201baf73#diff-bc149e5241d965436914a3a7c144b71bcceb435b5bd18cf7887925e11151355dL204-L206 is responsible for it.


fgrosa (approved):
Thanks @mfaggin !

PR #5284, opened 2024-03-21
Incorporated the electrons and muons also to the PID QC

victor-gonzalez:
Also
- if there is TOF, always use it although the threshold were not reached (this should not make necessary to discard the garbage below a certain beta)
- the filters for the different matching now exclude the DCA cut


alibuild (approved):

PR #5294, opened 2024-03-22
PWGHF: Fill DplusPi McRec table after correlated bkg checks

AlexBigO:
Hello @fgrosa !

I found another bug :)

We need to fill the DplusPi McRec table after the decay type checks, otherwise the `flag` value is not updated.


fgrosa (approved):
Thanks @AlexBigO!

PR #5315, opened 2024-03-23
PWGMM: Mult: Fix reduced table descriptions

aalkin:


alibuild (approved):

PR #5322, opened 2024-03-23
Fix for local BC

ekryshen:


alibuild (approved):

PR #5325, opened 2024-03-23
Reduce the number of needed histograms

victor-gonzalez:
Also incorporated the nsigmas histograms


alibuild (approved):

PR #5337, opened 2024-03-25
rework on PWGMM/Mult/Tasks:heavy-ion-mult.cxx

abmodak:
1) Rework on the following process functions: "processDataCentFT0C" and "processMCCentFT0C" 
2) Add THnSparse instead of multiple histograms for different centrality bins 
3) Add also THnSparse for different particle species

### DIFF
--- a/PWGMM/Mult/Tasks/heavy-ion-mult.cxx
+++ b/PWGMM/Mult/Tasks/heavy-ion-mult.cxx
@@ -73,40 +73,12 @@ static constexpr TrackSelectionFlags::flagtype trackSelectionDCA =
 static constexpr TrackSelectionFlags::flagtype trackSelectionDCAXYonly =
   TrackSelectionFlags::kDCAxy;
 
-static constexpr int CentHistArray{15};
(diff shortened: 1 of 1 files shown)


PWGMM/Mult/Tasks/heavy-ion-mult.cxx
```diff
@@ -444,54 +378,63 @@ struct HeavyIonMultiplicity {
...
+                if (std::abs(Rectrack.eta()) > etaRange) {
+                  continue;
+                }
+                if (Rectrack.has_mcParticle()) {
+                  auto mcpart = Rectrack.mcParticle();
+                  if (std::abs(mcpart.eta()) > etaRange) {
+                    continue;
+                  }
+                  if (mcpart.isPhysicalPrimary()) {
+                    if (std::abs(mcpart.pdgCode()) == 211) {
```
aalkin, line 398: I recommend using enum instead of numerical codes, e.g. `PDG_t::kPiPlus`, etc.

PR #5339, opened 2024-03-25
PWGMM: Mult: only use HepMC in reducer if available

aalkin:


alibuild (approved):

PR #5340, opened 2024-03-25
PWGHF: Sync derived D0 collision tables with Lc

vkucera:
@nzardosh

### DIFF
--- a/PWGHF/DataModel/DerivedTables.h
+++ b/PWGHF/DataModel/DerivedTables.h
@@ -320,6 +320,8 @@ DECLARE_SOA_TABLE(StoredHfD0Bases, "AOD1", "HFD0BASE", //! Table with basic cand
 // normalised: DecayLengthNormalised, DecayLengthXYNormalised, ImpactParameterNormalised0
 DECLARE_SOA_TABLE(HfD0Pars, "AOD", "HFD0PAR", //! Table with candidate properties used for selection
                   hf_cand::Chi2PCA,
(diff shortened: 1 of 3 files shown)


fgrosa (commented):
Hi @vkucera thanks! Do you think that all these centrality estimators are needed? Otherwise we could keep only the needed ones, so avoid the creation of the unnecessary tables from the centrality task.

vkucera:
> Hi @vkucera thanks! Do you think that all these centrality estimators are needed? Otherwise we could keep only the needed ones, so avoid the creation of the unnecessary tables from the centrality task.

Hi @fgrosa , the list of estimators is based on what we already store in the Lambda_c tree creator now and, at the moment,  includes what @DelloStritto wishes to have for multiplicity studies.

DelloStritto:
> > Hi @vkucera thanks! Do you think that all these centrality estimators are needed? Otherwise we could keep only the needed ones, so avoid the creation of the unnecessary tables from the centrality task.
> 
> Hi @fgrosa , the list of estimators is based on what we already store in the Lambda_c tree creator now and, at the moment, includes what @DelloStritto wishes to have for multiplicity studies.

Hi @vkucera, @fgrosa. At some point, we can think about removing the FT0A and FT0C and keeping only the FT0M, but at the moment I still observe a non-flat trend in the FT0M. That's why I decided to store all of them

nzardosh (approved):

alibuild (approved):

PR #5345, opened 2024-03-25
PWGLF: MC finder reworked

ddobrigk:


ercolessi (approved):

ddobrigk:
With this PR, I have also cross-checked that the two algorithmic paths to get all findable V0s match perfectly with respect to one another. This means that, depending on the use cases and speed, one can choose any of them and get the same outcome. This is ready to merge :-D

ddobrigk:
> With this PR, I have also cross-checked that the two algorithmic paths to get all findable V0s match perfectly with respect to one another. This means that, depending on the use cases and speed, one can choose any of them and get the same outcome. This is ready to merge :-D

Thanks @ercolessi :-D

PR #5366, opened 2024-03-26
(PWGCF) Initial push for SPC and NUA creation Run3 tasks

MaximVirta:


PWGCF/Flow/Core/FlowJHistManager.cxx
```diff
@@ -0,0 +1,176 @@
...
+// O2 headers.
+
+// O2 Physics headers.
+#include "PWGCF/Flow/Core/FlowJHistManager.h"
+
+// Namespaces.
+using namespace o2;
+using namespace o2::framework;
+
+namespace o2::analysis::PWGCF
```
victor-gonzalez, line 23: Please use a more restrictive namespace

PWGCF/Flow/Core/FlowJHistManager.cxx
```diff
@@ -0,0 +1,176 @@
...
+{
+/// \brief Create the histograms in the QA registry.
+void FlowJHistManager::CreateHistQA()
```
victor-gonzalez, line 26: From the Jyväskylä group you already have a `PWGCF/JCorran/Core/JHistManager.cxx`
Can that be expanded to incorporate this functionality?

PWGCF/Flow/Core/FlowJSPCAnalysis.cxx
```diff
@@ -0,0 +1,420 @@
...
+#include "PWGCF/Flow/Core/FlowJSPCAnalysis.h"
+
+using namespace o2;
+using namespace o2::framework;
+using namespace std;
+
+namespace o2::analysis::PWGCF
+{
+
+void FlowJSPCAnalysis::CalculateQvectors(const vector<float>& phiAngles, const vector<float>& phiWeights)
```
victor-gonzalez, line 27: Is not this functionality already addressed in the JCORRAN package from the Jyväskylä under PWGCF?
If not, cannot be the Flow Generic Framework implementation currently under PWGCF being used instead?

PWGCF/Flow/Tasks/flowJSPCAnalysisTask.cxx
```diff
@@ -0,0 +1,207 @@
...
+    // printf("Centrality: %.2f,\tBin: %d\n", cent, cBin);
+    SPCHistograms.fill(HIST("FullCentrality"), cent);
+    vector<float> trackPhi;
+    vector<float> trackWeight;
+    int nTracks = 0;
+
+    for (auto& track : tracks) {
+      if (cfgFillQA)
+        histManager.FillTrackQA<1>(track, cBin, coll.posZ());
+      trackPhi.push_back(track.phi());
```
victor-gonzalez, line 170: In here the phi column of the tracks table is replicated in memory and this is not efficient
Please pass the actual table instead to the involved functions 'templating' them if it were needed

victor-gonzalez (changes requested):
It is feasible for this analysis to use the Flow Generic Framework implemented under PWGCF? Or the Jyväskylä group own Flow Generic Framework under PWGCF/JCorran?

PWGCF/Flow/Core/FlowJHistManager.cxx
MaximVirta: JHistManager from JCorran is based on the AliPhysics format. This class has O2 histogram implementations and hence I think we need to add functions from JCorran file to this one. I'll discuss this with Jasper.

PWGCF/Flow/Core/FlowJSPCAnalysis.cxx
MaximVirta: I couldn't find a method which would not assume the size of the Q-vectors, the n-harmonics and powers. As those change analysis to analysis, tt would be beneficial to define a general method, which doesn't make this assumption. I'll discuss this with Jasper as well.

PWGCF/Flow/Tasks/flowJSPCAnalysisTask.cxx
MaximVirta: You are correct, it's inefficient. This has been now corrected for the future task.

PWGCF/Flow/Core/FlowJHistManager.cxx
jaelpark: The JHistManager will most likely go given that it's not very compatible with the O2 output objects, and that the TNamed functionality that enables it to be written has never been properly implemented.

Also, @MaximVirta we should probably get rid of the `GetCentBin()` entirely because it's not very flexible to have the binning hardcoded. The cent index should be supplied by the top level task, like I did in my previous commits. This is obtained from a configurable axis.

PWGCF/Flow/Core/FlowJSPCAnalysis.cxx
jaelpark: We decided that we'd create a new CalculateQVectors template that can be used by all analyses under JCorran.

jaelpark: I've made a PR here https://github.com/AliceO2Group/O2Physics/pull/5632
Maxim will then make the changes around that

PR #5374, opened 2024-03-26
EventFiltering : Adding D0 and Lc tagged jet triggers

nzardosh:


mpuccio (changes requested):
Hi Nima,  you should also increase the number of filters and add name, description and FilteringTask name [here](https://github.com/AliceO2Group/O2Physics/pull/5374/files#diff-260be07f272f4bce55bc3597acec5e1eb8b6d389aaa1c237f55986867b0b232dR231), in this way you will be integrated with the CEFP logic.

mpuccio (approved):

PR #5381, opened 2024-03-26
PWGHF: Add ML scores in derived D0 and Lc tables

vkucera:


nzardosh:
Hi @vkucera , thanks for the PR. Can you provide a configurable so that the ml table is always filled even if the non ml process is selected? Otherwise all tasks that want to use the derived output later also need separate ml process functions

vkucera:
> Hi @vkucera , thanks for the PR. Can you provide a configurable so that the ml table is always filled even if the non ml process is selected? Otherwise all tasks that want to use the derived output later also need separate ml process functions

It's already there: `fillCandidateMl`

alibuild (approved):

PR #5383, opened 2024-03-26
PWGHF: fix loop over generated particles toXiPi

ZFederica:
Fix loop over generated particles in candidateCreatorToXiPi: now at each step of the decay chain all the daughters are taken into account


ZFederica:
@fcatalan92

fcatalan92 (approved):

PR #5385, opened 2024-03-27
PWGHF: implement process function using tracked cascades information toXiPi

ZFederica:
In candidateCreatorToXiPi: implement a new process function using tracked cascades information

### DIFF
--- a/PWGHF/TableProducer/candidateCreatorToXiPi.cxx
+++ b/PWGHF/TableProducer/candidateCreatorToXiPi.cxx
@@ -80,6 +80,7 @@ struct HfCandidateCreatorToXiPi {
   int runNumber;
 
   using MyCascTable = soa::Join<aod::CascDatas, aod::CascCovs>; // to use strangeness tracking, use aod::TraCascDatas instead of aod::CascDatas
+  using MyTrkCascTable = soa::Join<aod::TraCascDatas, aod::CascCovs>;
   using CascadesLinked = soa::Join<Cascades, CascDataLink>;
   using MyV0Table = soa::Join<aod::V0Datas, aod::V0Covs>;
   using MySkimIdx = soa::Filtered<aod::HfCascLf2Prongs>;
@@ -93,6 +94,13 @@ struct HfCandidateCreatorToXiPi {
 
   void init(InitContext const&)
   {
+    if (doprocessWithCascades == true && doprocessWithTrackedCascades == true) {
+      LOGF(fatal, "Cannot enable processWithCascades and processWithTrackedCascades at the same time. Please choose one.");
+    }
+    if (doprocessWithCascades == false && doprocessWithTrackedCascades == false) {
+      LOGF(fatal, "Please enable at least one process function, either processWithCascades or processWithTrackedCascades.");
+    }
+
     df.setPropagateToPCA(propagateToPCA);
     df.setMaxR(maxR);
     df.setMaxDZIni(maxDZIni);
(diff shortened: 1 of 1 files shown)


PWGHF/TableProducer/candidateCreatorToXiPi.cxx
```diff
@@ -93,6 +94,13 @@ struct HfCandidateCreatorToXiPi {
 
   void init(InitContext const&)
   {
+    if (doprocessWithCascades == true && doprocessWithTrackedCascades == true) {
+      LOGF(fatal, "Cannot enable processWithCascades and processWithTrackedCascades at the same time. Please choose one.");
+    }
+    if (doprocessWithCascades == false && doprocessWithTrackedCascades == false) {
+      LOGF(fatal, "Please enable at least one process function, either processWithCascades or processWithTrackedCascades.");
+    }
```
vkucera, line 102:
```suggestion
    if (doprocessWithCascades && doprocessWithTrackedCascades) {
      LOGF(fatal, "Cannot enable processWithCascades and processWithTrackedCascades at the same time. Please choose one.");
    }
    if (!doprocessWithCascades && !doprocessWithTrackedCascades) {
      LOGF(fatal, "Please enable at least one process function, either processWithCascades or processWithTrackedCascades.");
    }
```
- Do not use ` == true/false`. It has no effect and just makes code longer.
- Instead of two checks, you can just have one check with `std::accumulate`.

vkucera (changes requested):

PWGHF/TableProducer/candidateCreatorToXiPi.cxx
```diff
@@ -327,6 +335,226 @@ struct HfCandidateCreatorToXiPi {
...
+                                  aod::BCsWithTimestamps const&,
+                                  TracksWCovDca const&,
+                                  MyTrkCascTable const&, CascadesLinked const&,
+                                  MySkimIdx const& candidates)
+  {
+    double massPionFromPDG = MassPiPlus;    // pdg code 211
+    double massLambdaFromPDG = MassLambda0; // pdg code 3122
+    double massXiFromPDG = MassXiMinus;     // pdg code 3312
+    double massOmegacFromPDG = MassOmegaC0; // pdg code 4332
+    double massXicFromPDG = MassXiC0;       // pdg code 4132
```
fgrosa, line 350: Why is this needed? Can't you directly use the variables on the right of the ` = ` sign?

PWGHF/TableProducer/candidateCreatorToXiPi.cxx
```diff
@@ -327,6 +335,226 @@ struct HfCandidateCreatorToXiPi {
...
+                   ctOmegac, ctCascade, ctV0, ctXic,
+                   pseudorapV0Dau0, pseudorapV0Dau1, pseudorapPiFromCas, pseudorapPiFromCharmBaryon,
+                   pseudorapCharmBaryon, pseudorapCascade, pseudorapV0,
+                   dcaxyV0Dau0, dcaxyV0Dau1, dcaxyPiFromCasc,
+                   dcazV0Dau0, dcazV0Dau1, dcazPiFromCasc,
+                   dcaCascDau, dcaV0Dau, dcaCharmBaryonDau,
+                   decLenCharmBaryon, decLenCascade, decLenV0, errorDecayLengthCharmBaryon, errorDecayLengthXYCharmBaryon,
+                   hfFlag);
+
+    } // loop over LF Cascade-bachelor candidates
```
fgrosa, line 554: This is almost completely a copy paste of the other process function

fgrosa (changes requested):
Hi @ZFederica thanks a lot for the development!
My main concern is that you almost entirely copy-pasted one process function in the other. Couldn't you define a template function that is called in the two process functions passing a different table?

ZFederica:
After checking with experts, it turned out that it is not yet worth to switch to strangeness tracking

PR #5388, opened 2024-03-27
[PWGMM] Change the type of histograms in trackPropagation.cxx

sarahherrmann:
This will allow the bin entries to go up to 2^53, instead of 2.147E9


sarahherrmann:
> Do you really need that many entries? Would TH1F be enough?

I don't think it would, quoting Rene Brun on TH1F:  "If you fill with weights=1, after 1e7 entries, adding one more entry will have no effect in one channel." see https://root-forum.cern.ch/t/maximum-histogram-bin-content/5891

So to be conservative, since I can't use TH1L, I used TH1D

aalkin (approved):

PR #5395, opened 2024-03-27
[PWGDQ] Adding flow extra table from central Q-vector framework

lucamicheletti93:
The Q-vectors from the central framework are now saved in a dedicated table which can be produced separately or together the standard Q-vector table from GFW.
This development is meant for the streaming campaing for Pb-Pb 2023 data.


alibuild (approved):

PR #5407, opened 2024-03-28
Common: RecoDecay: Consider only particles coming from decay process when matching

ZFederica:
Check the production process of particles when doing the MC matching (both rec and gen) so that only particles coming from the decay are taken into account. This is optional and can be controlled using the parameter checkProcess, which is by default false (so the default behaviour is not changed wrt the current version of O2Physics).
In particular the check on the process is done using the function `getProcess()`, that returns `kPDecay` for particles produced in a decay this function. This check is implemented inside `getMatchedMcRec` and `isMatchedMcGen`. Inside these functions the checks on the number of daughters have been removed because they do not allow to properly take into account only the desiderd daughters.

### DIFF
--- a/Common/Core/RecoDecay.h
+++ b/Common/Core/RecoDecay.h
@@ -645,10 +645,8 @@ class RecoDecay
       // for (int i = 0; i < stage; i++) // Indent to make the tree look nice.
       //   printf(" ");
       // printf("Stage %d: Adding %d (PDG %d) as final daughter.\n", stage, index, PDGParticle);
-      // check if the particle comes from a decay process
-      if (particle.getProcess() == TMCProcess::kPDecay) {
-        list->push_back(particle.globalIndex());
-      }
+      // check if the particle comes from a decay process   
+      list->push_back(particle.globalIndex());
       return;
     }
     // If we are here, we have to follow the daughter tree.
@@ -745,6 +743,12 @@ class RecoDecay
         }
         // Get the list of actual final daughters.
         getDaughters(particleMother, &arrAllDaughtersIndex, arrPDGDaughters, depthMax);
+        // remove daughters not coming from the decay process
+        for(int j=0; j<static_cast<int>(arrAllDaughtersIndex.size()); j++){
+          if(particlesMC.rawIteratorAt(arrAllDaughtersIndex[j] - particlesMC.offset()).getProcess() != TMCProcess::kPDecay){
+            arrAllDaughtersIndex.erase(arrAllDaughtersIndex.begin() + j);
+          }
+        }
         // printf("MC Rec: Mother %d has %d final daughters:", indexMother, arrAllDaughtersIndex.size());
         // for (auto i : arrAllDaughtersIndex) {
         //   printf(" %d", i);
@@ -860,6 +864,7 @@ class RecoDecay
       }
       // Check that the number of direct daughters coming from the dacay is not larger than the number of expected final daughters.
       for (const auto& dau : candidate.template daughters_as<T>()) {
+        //printf("Daughter production process is %i", dau.getProcess());
         if (dau.getProcess() == TMCProcess::kPDecay) {
           dauCounter++;
         }
@@ -870,6 +875,12 @@ class RecoDecay
       }
       // Get the list of actual final daughters.
       getDaughters(candidate, &arrAllDaughtersIndex, arrPDGDaughters, depthMax);
+      // remove daughters not coming from the decay process
+      for(int j=0; j<static_cast<int>(arrAllDaughtersIndex.size()); j++){
+        if(particlesMC.rawIteratorAt(arrAllDaughtersIndex[j] - particlesMC.offset()).getProcess() != TMCProcess::kPDecay){
+          arrAllDaughtersIndex.erase(arrAllDaughtersIndex.begin() + j);
+        }
+      }
       // printf("MC Gen: Mother %ld has %ld final daughters:", candidate.globalIndex(), arrAllDaughtersIndex.size());
       // for (auto i : arrAllDaughtersIndex) {
       //   printf(" %d", i);
--- a/Common/Core/RecoDecay.h
+++ b/Common/Core/RecoDecay.h
@@ -24,7 +24,6 @@
 #include <vector>    // std::vector
 
 #include "CommonConstants/MathConstants.h"
-#include "Framework/AnalysisDataModel.h"
 #include "TMCProcess.h" // for VMC Particle Production Process
 
 /// Base class for calculating properties of reconstructed decays
--- a/Common/Core/RecoDecay.h
+++ b/Common/Core/RecoDecay.h
@@ -644,8 +644,7 @@ class RecoDecay
       // printf("getDaughters: ");
       // for (int i = 0; i < stage; i++) // Indent to make the tree look nice.
       //   printf(" ");
-      // printf("Stage %d: Adding %d (PDG %d) as final daughter.\n", stage, index, PDGParticle);
-      // check if the particle comes from a decay process   
+      // printf("Stage %d: Adding %d (PDG %d) as final daughter.\n", stage, index, PDGParticle); 
       list->push_back(particle.globalIndex());
       return;
     }
--- a/Common/Core/RecoDecay.h
+++ b/Common/Core/RecoDecay.h
@@ -24,6 +24,7 @@
 #include <vector>    // std::vector
 
 #include "CommonConstants/MathConstants.h"
+#include "Framework/AnalysisDataModel.h"
 #include "TMCProcess.h" // for VMC Particle Production Process
 
 /// Base class for calculating properties of reconstructed decays
--- a/Common/Core/RecoDecay.h
+++ b/Common/Core/RecoDecay.h
@@ -645,7 +645,12 @@ class RecoDecay
       // for (int i = 0; i < stage; i++) // Indent to make the tree look nice.
       //   printf(" ");
       // printf("Stage %d: Adding %d (PDG %d) as final daughter.\n", stage, index, PDGParticle);
-      list->push_back(particle.globalIndex());
+      printf("Daughter production process is %i\n", particle.getProcess());
+      printf("Daughter pdg code is %i\n", particle.pdgCode());
+      if(particle.getProcess() == TMCProcess::kPDecay){
+        list->push_back(particle.globalIndex());
+      }
+      
       return;
     }
     // If we are here, we have to follow the daughter tree.
@@ -684,7 +689,6 @@ class RecoDecay
     int indexMother = -1;                  // index of the mother particle
     std::vector<int> arrAllDaughtersIndex; // vector of indices of all daughters of the mother of the first provided daughter
     std::array<int, N> arrDaughtersIndex;  // array of indices of provided daughters
-    int dauCounter = 0;                    // counter particleMother's daughters coming from the decay process
     if (sign) {
       *sign = sgn;
     }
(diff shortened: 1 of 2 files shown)


ZFederica:
@fgrosa

vkucera:
Hi @ZFederica , please give more details about your suggested changes and why they are needed. This is a core header so we should be very careful what we add and change. For example, we should not need the analysis data model header.
In general, please compile your code locally first before making a PR ready for review.

Common/Core/RecoDecay.h
```diff
@@ -24,6 +24,8 @@
 #include <vector>    // std::vector
 
 #include "CommonConstants/MathConstants.h"
+#include "Framework/AnalysisDataModel.h"
+#include "TMCProcess.h" // for VMC Particle Production Process
```
vkucera, line 28: There is no file `TMCProcess.h` in O2 nor in O2Physics.

Common/Core/RecoDecay.h
```diff
@@ -644,7 +646,10 @@ class RecoDecay
       // for (int i = 0; i < stage; i++) // Indent to make the tree look nice.
       //   printf(" ");
       // printf("Stage %d: Adding %d (PDG %d) as final daughter.\n", stage, index, PDGParticle);
-      list->push_back(particle.globalIndex());
+      // check if the particle comes from a decay process
+      if (particle.getProcess() == kPDecay) {
```
vkucera, line 650: Which namespace is `kPDecay` from?

Common/Core/RecoDecay.h
```diff
@@ -683,6 +688,7 @@ class RecoDecay
     int indexMother = -1;                  // index of the mother particle
     std::vector<int> arrAllDaughtersIndex; // vector of indices of all daughters of the mother of the first provided daughter
     std::array<int, N> arrDaughtersIndex;  // array of indices of provided daughters
+    int dauCounter = 0;                    // counter particleMother's daughters coming from the decay process
```
vkucera, line 691: Why is this needed?

vkucera (changes requested):

fgrosa:
> Hi @ZFederica , please give more details about your suggested changes and why they are needed. This is a core header so we should be very careful what we add and change. For example, we should not need the analysis data model header. In general, please compile your code locally first before making a PR ready for review.

Hi @vkucera, about the reason for this development (not the technical implementation that I still did not check indeed), is that because of interactions with materials, we have in the Monte Carlo electrons and nuclei that are "daughters" of strange particles, like Xi and Omegas (this is because these particles are decayed by GEANT, that simulates correctly also the interaction with the material). If not added the check of the process that generated these particles, they are counted as decay daughters, and true signals get rejected (while they should not be).

Common/Core/RecoDecay.h
vkucera: The file is in VMC. It could be fine it does not add a dependency on any linked library.

Common/Core/RecoDecay.h
```diff
@@ -24,6 +24,8 @@
 #include <vector>    // std::vector
 
 #include "CommonConstants/MathConstants.h"
+#include "Framework/AnalysisDataModel.h"
```
vkucera, line 27: It would be better to avoid including more headers unnecessarily.

Common/Core/RecoDecay.h
vkucera, line 650:
```suggestion
      if (particle.getProcess() == TMCProcess::kPDecay) {
```

Common/Core/RecoDecay.h
```diff
@@ -718,13 +728,19 @@ class RecoDecay
...
+        dauCounter = 0; // reset daughter counter for the new particleMother
         // Check the daughter indices.
         if (!particleMother.has_daughters()) {
           // Printf("MC Rec: Rejected: bad daughter index range: %d-%d", particleMother.daughtersIds().front(), particleMother.daughtersIds().back());
           return -1;
         }
-        // Check that the number of direct daughters is not larger than the number of expected final daughters.
-        if (particleMother.daughtersIds().back() - particleMother.daughtersIds().front() + 1 > static_cast<int>(N)) {
+        // Check that the number of direct daughters coming from the dacay is not larger than the number of expected final daughters.
+        for (const auto& dau : particleMother.template daughters_as<o2::aod::McParticles>()) {
```
vkucera, line 738:
```suggestion
        for (const auto& dau : particleMother.template daughters_as<T>()) {
```
We should absolutely avoid hard-coding any table types.
Is the casting even needed?

vkucera: Replacing the simple index difference with looping over the actual particles will slow this check down significantly. Did you do any CPU time checks?

Common/Core/RecoDecay.h
ZFederica: Okay, so can I leave line 28 as it is now?

vkucera: Sorry, I missed "if" in my previous message. I meant: "It could be fine if it does not add a dependency on any linked library."

Common/Core/RecoDecay.h
ZFederica: I will remove the hard-coded table.
I need to work with elements of the McParticles table in order to be able to retrieve the information about the production process of each daughter (and the check on the production process needs to be done for all the daughters), while the function daughtersIds would return the daughter tracks slice.
I did not do CPU time checks, but I would also not know how to implement this differently.

Common/Core/RecoDecay.h
ZFederica: Having a counter is needed because the original version of the function was checking that the number of daughters coming from the decay was not larger than the number of expected final daughters, so it was simply checking on the size of the daughters array. Now I have introduced a counter because I still have to check all the daughters, but I should increase the counter only if the production process is a decay.

ZFederica:
I tested the code locally and there are some problems, please leave the PR in draft (even if it builds)

Common/Core/RecoDecay.h
```diff
@@ -644,7 +645,10 @@ class RecoDecay
       // for (int i = 0; i < stage; i++) // Indent to make the tree look nice.
       //   printf(" ");
       // printf("Stage %d: Adding %d (PDG %d) as final daughter.\n", stage, index, PDGParticle);
-      list->push_back(particle.globalIndex());
+      // check if the particle comes from a decay process
+      if (particle.getProcess() == TMCProcess::kPDecay) {
+        list->push_back(particle.globalIndex());
```
vkucera, line 650: There should be no conditions in this block. The decision to push the particle is made by the value of `isFinal`.

ZFederica: I have to add the particle to the list only if it comes from a decay, so I would simply move this condition. Where do you want me to put it?

vkucera:
Since the cases where we have to consider the process code are very limited, we agreed with @fgrosa that this check should be optional and not done by default.

So I would suggest to add a new argument `bool checkProcess = false` in `getDaughters` and the matching functions.
This will also help you to debug your implementation, because you will be able to switch it on and off in the same branch.

Second, it seems to me that we can move the check in `getDaughters` all the way up and return immediately if it fails because for the non-decay particles we are not interested in adding them or their daughters in the list.

ZFederica:
@fgrosa @vkucera I tested these changes in my local version for the xic0 matching, they work correctly and allow for recovering the expected number of candidates at the various steps of the decay chain (numbers compared with the branching ratios of cascade and V0)
Thanks @ddobrigk for your suggestions!

vkucera (changes requested):
- The correct place for checking `getProcess` is inside `getDaughters`. That is the purpose of that function. It should not be needed to check it anywhere else.
- The check should be done `if (checkProcess)` which should be `false` by default.
- You should not remove any existing checks. You can skip the check on too many direct daughters `if (checkProcess)`.

ZFederica:
Moving the check inside getDaughters did not work, the build did not fail but I also ended up with 0 candidates matched. If that function is not modified, it also does not make sense to leave the checks on the number of daughters (because I would count "daughters" not coming from the decay and there would be a mismatch in the numbers)

vkucera:
I would try this inside `getDaughter`:
```c++
    // If the particle is neither the original particle nor coming from a decay, we do nothing and exit.
    if (checkProcess && stage != 0 && particle.getProcess() != TMCProcess::kPDecay) {
      return;
    }
```

vkucera:
- Does everything work as before (in `master`) with `checkProcess = false`?
- Can you add this before the `return` statement and see for which particles it fails?
  ```c++
  LOGF(warn, "PDG code = %d, process code = %d", particle.pdgCode(), particle.getProcess());
  ```

Common/Core/RecoDecay.h
```diff
@@ -22,8 +22,10 @@
 #include <cmath>     // std::abs, std::sqrt
 #include <utility>   // std::move
 #include <vector>    // std::vector
+#include <iostream>
```
vkucera, line 25: Not needed. Unwanted.
```suggestion
```

Common/Core/RecoDecay.h
```diff
@@ -602,7 +604,7 @@ class RecoDecay
   /// \param stage  decay tree level; If different from 0, the particle itself will be added in the list in case it has no daughters.
   /// \note Final state is defined as particles from arrPDGFinal plus final daughters of any other decay branch.
   /// \note Antiparticles of particles in arrPDGFinal are accepted as well.
-  template <std::size_t N, typename T>
+  template <bool checkProcess = false, std::size_t N, typename T>
```
vkucera, line 607: Please add documentation of `checkProcess` before the `\param particle` line:
`/// \param checkProcess  switch to accept only decay daughters by checking the production process of MC particles`

Common/Core/RecoDecay.h
```diff
@@ -613,6 +615,12 @@ class RecoDecay
       // Printf("getDaughters: Error: No list!");
       return;
     }
+    if constexpr (checkProcess) {
+      if (stage != 0 && particle.getProcess() != TMCProcess::kPDecay && particle.getProcess() != TMCProcess::kPPrimary) { // decay products of HF hadrons are labeled as kPPrimary
+        return;
+      }
+    }
```
vkucera, line 622:
```suggestion
    if constexpr (checkProcess) {
      // If the particle is neither the original particle nor coming from a decay, we do nothing and exit.
      if (stage != 0 && particle.getProcess() != TMCProcess::kPDecay && particle.getProcess() != TMCProcess::kPPrimary) { // decay products of HF hadrons are labeled as kPPrimary
        return;
      }
    }
```

Common/Core/RecoDecay.h
```diff
@@ -668,7 +676,7 @@ class RecoDecay
   /// \param sign  antiparticle indicator of the found mother w.r.t. PDGMother; 1 if particle, -1 if antiparticle, 0 if mother not found
   /// \param depthMax  maximum decay tree level to check; Daughters up to this level will be considered. If -1, all levels are considered.
   /// \return index of the mother particle if the mother and daughters are correct, -1 otherwise
-  template <bool acceptFlavourOscillation = false, std::size_t N, typename T, typename U>
+  template <bool acceptFlavourOscillation = false, bool checkProcess = false, std::size_t N, typename T, typename U>
```
vkucera, line 679: Add documentation of `checkProcess`.

Common/Core/RecoDecay.h
```diff
@@ -785,15 +795,15 @@ class RecoDecay
   /// \param acceptAntiParticles  switch to accept the antiparticle
   /// \param sign  antiparticle indicator of the candidate w.r.t. PDGParticle; 1 if particle, -1 if antiparticle, 0 if not matched
   /// \return true if PDG code of the particle is correct, false otherwise
-  template <bool acceptFlavourOscillation = false, typename T, typename U>
+  template <bool acceptFlavourOscillation = false, bool checkProcess = false, typename T, typename U>
```
vkucera, line 798: Add documentation of `checkProcess`.

vkucera (changes requested):
Thanks @ZFederica , the current version looks very good.
Please see my suggestions.

Common/Core/RecoDecay.h
vkucera: Please respect the order of parameters and document them in order of appearance.

vkucera (approved):
Thanks @ZFederica , it looks good to me now.

ZFederica:
Perfect, thanks @vkucera

Common/Core/RecoDecay.h
```diff
@@ -724,12 +734,14 @@ class RecoDecay
...
         }
         // Check that the number of direct daughters is not larger than the number of expected final daughters.
-        if (particleMother.daughtersIds().back() - particleMother.daughtersIds().front() + 1 > static_cast<int>(N)) {
-          // Printf("MC Rec: Rejected: too many direct daughters: %d (expected %ld final)", particleMother.daughtersIds().back() - particleMother.daughtersIds().front() + 1, N);
-          return -1;
+        if constexpr (!checkProcess) {
+          if (particleMother.daughtersIds().back() - particleMother.daughtersIds().front() + 1 > static_cast<int>(N)) {
+            // Printf("MC Rec: Rejected: too many direct daughters: %d (expected %ld final)", particleMother.daughtersIds().back() - particleMother.daughtersIds().front() + 1, N);
+            return -1;
+          }
```
fcatalan92, line 741: @vkucera Sorry for jumping in, but isn't this check (which here we are turning off if `checkProcess` is true) redundant since re-done in https://github.com/AliceO2Group/O2Physics/blob/964ccfbc129a645503ab7edac69a873565168b30/Common/Core/RecoDecay.h#L739 ? What do you think?

vkucera: No it is not redundant.
The first check discards particles which have too many daughters before even checking how many are returned by `getDaughters`.
The second check discards the particles for which `getDaughters` did not return the expected number of daughters.

fcatalan92 (approved):

vkucera:
@ddobrigk This is ready for merging.

ddobrigk (approved):

PR #5421, opened 2024-03-30
Fine tuning the track selection

victor-gonzalez:
Also
- incorporated overall minimum momentum for analysis
- default estimator when processes with no centrality
- PID processes removal from filter collision task
- preparation for potential FullPID basis


alibuild (approved):

PR #5426, opened 2024-03-31
EventFiltering/PWGHF: spli configurables for charge states in Xic*->SigmaC

fgrosa:


alibuild (approved):

PR #5430, opened 2024-03-31
EventFiltering,PWGHF: Move HFFilter ML application one step before

fgrosa:
In this PR I moved the application of the ML models needed for the `HFFilter.cxx` in the `trackIndexSkimCreator`.
A cut on the ML scores can be applied already in the `trackIndexSkimCreator`, but the scores of the candidates passing the selections are also stored in an output table to further apply selections in the subsequent steps (needed for instance to separate prompt and non-prompt candidates).

### DIFF
--- a/PWGHF/TableProducer/trackIndexSkimCreator.cxx
+++ b/PWGHF/TableProducer/trackIndexSkimCreator.cxx
@@ -1427,7 +1427,7 @@ struct HfTrackIndexSkimCreator {
       const std::vector<std::string> mlModelPathCcdb2Prongs = {mlModelPathCCDB.value + "D0"};
       const std::array<std::vector<std::string>, kN3ProngDecays> mlModelPathCcdb3Prongs = {std::vector<std::string>{mlModelPathCCDB.value + "Dplus"}, std::vector<std::string>{mlModelPathCCDB.value + "Lc"}, std::vector<std::string>{mlModelPathCCDB.value + "Ds"}, std::vector<std::string>{mlModelPathCCDB.value + "Xic"}};
       const std::vector<double> ptBinsMl = {0., 1.e10};
-      const std::vector<int> cutDirMl = {o2::cuts_ml::CutDirection::CutSmaller, o2::cuts_ml::CutDirection::CutGreater, o2::cuts_ml::CutDirection::CutGreater};
+      const std::vector<int> cutDirMl = {o2::cuts_ml::CutDirection::CutGreater, o2::cuts_ml::CutDirection::CutSmaller, o2::cuts_ml::CutDirection::CutSmaller};
       const std::array<LabeledArray<double>, kN3ProngDecays> thresholdMlScore3Prongs = {thresholdMlScoreDplusToPiKPi, thresholdMlScoreLcToPiKP, thresholdMlScoreDsToPiKK, thresholdMlScoreXicToPiKP};
       
       // initialise 2-prong ML response
--- a/PWGHF/TableProducer/trackIndexSkimCreator.cxx
+++ b/PWGHF/TableProducer/trackIndexSkimCreator.cxx
@@ -44,6 +44,7 @@
 #include "Common/DataModel/CollisionAssociationTables.h"
 #include "Common/DataModel/EventSelection.h"
 #include "Common/DataModel/TrackSelectionTables.h"
+#include "Tools/ML/MlResponse.h"
 
 #include "PWGLF/DataModel/LFStrangenessTables.h"
 
@@ -1156,6 +1157,9 @@ struct HfTrackIndexSkimCreator {
   Produces<aod::HfDstars> rowTrackIndexDstar;
   Produces<aod::HfCutStatusDstar> rowDstarCutStatus;
   Produces<aod::HfPvRefitDstar> rowDstarPVrefit;
+  // Tables with ML scores for HF Filters
+  Produces<aod::Hf2ProngMlProbs> rowTrackIndexMlScoreProng2;
+  Produces<aod::Hf3ProngMlProbs> rowTrackIndexMlScoreProng3;
 
   Configurable<bool> isRun2{"isRun2", false, "enable Run 2 or Run 3 GRP objects for magnetic field"};
   Configurable<bool> do3Prong{"do3Prong", 0, "do 3 prong"};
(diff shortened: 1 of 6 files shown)


PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -1627,6 +1691,22 @@ struct HfTrackIndexSkimCreator {
     }
   }
 
+  /// Method to perform ML selections for 2-prong candidates after the rectangular selections
+  /// \param featuresCand is the vector with the candidate features
+  /// \param outputScores is the vector with the output scores to be filled
+  /// \param isSelected ia s bitmap with selection outcome
+  void is2ProngSelectedMlForHfFilters(std::vector<float> featuresCand, std::vector<float>& outputScores, int& isSelected)
```
vkucera, line 1698: `is...` indicates that the function returns `bool`.
```suggestion
  void applyMlSelectionForHfFilters2Prong(std::vector<float> featuresCand, std::vector<float>& outputScores, int& isSelected)
```

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -1695,6 +1775,26 @@ struct HfTrackIndexSkimCreator {
     }
   }
 
+  /// Method to perform ML selections for 2-prong candidates after the rectangular selections
+  /// \param featuresCand is the vector with the candidate features
+  /// \param outputScores is the array of vectors with the output scores to be filled
+  /// \param isSelected ia s bitmap with selection outcome
+  void is3ProngSelectedMlForHfFilters(std::vector<float> featuresCand, std::array<std::vector<float>, kN3ProngDecays>& outputScores, int& isSelected)
```
vkucera, line 1782:
```suggestion
  void applyMlSelectionForHfFilters3Prong(std::vector<float> featuresCand, std::array<std::vector<float>, kN3ProngDecays>& outputScores, int& isSelected)
```

PWGHF/TableProducer/trackIndexSkimCreator.cxx
fgrosa: Done in https://github.com/AliceO2Group/O2Physics/pull/5430/commits/87b013b121005169bdec56960b9f53ee73e735c4

PWGHF/TableProducer/trackIndexSkimCreator.cxx
fgrosa: Done in https://github.com/AliceO2Group/O2Physics/pull/5430/commits/87b013b121005169bdec56960b9f53ee73e735c4

fgrosa:
Hi @vkucera in addition to the naming conventions (I changed all the occurences, since there were many methods called `isSelected..` that were not returning a bool) I reduced the number of configurables because I discovered that there is a limit to 49 (if exceeded, a segmentation fault at runtime is thrown) and I added few histograms to monitor the BDT output scores (see example attached). I tested it locally and it provides the same result compared to the previous version, where the BDT models were applied in the `HFFilter.cxx`.
grafik

PWGHF/Core/SelectorCuts.h
```diff
@@ -94,6 +94,13 @@ static const std::vector<std::string> labelsPt{};
 
 // column labels
 static const std::vector<std::string> labelsCutBdt = {"BDTbkg", "BDTprompt", "BDTnonprompt"};
+
+// ONNX file names
+static const std::string onnxFileNameSpecies[5][1] = {{"ModelHandler_onnx_D0ToKPi.onnx"}, {"ModelHandler_onnx_DplusToPiKPi.onnx"}, {"ModelHandler_onnx_LcToPKPi.onnx"}, {"ModelHandler_onnx_DsToKKPi.onnx"}, {""}};
```
fgrosa, line 99: @vkucera Megalinter fails because it wants this to be `static const char*` instead of `static const std::string`, but it does not compile with `char` instead of `std::string`. Any idea how to fix it?

vkucera: What is the error?
`static const char* onnxFileNameSpecies[5][1] = ...` compiles just fine.

fgrosa: Did you try? It does not compile for me (full log attached):
```
/home/fgrosa/alice/sw/SOURCES/O2Physics/dev_trigger/0/PWGHF/TableProducer/trackIndexSkimCreator.cxx:1227:234: error: could not convert '{"onnxFileNames", {o2::analysis::hf_cuts_bdt_multiclass::onnxFileNameSpecies[0], 5, 1, o2::analysis::hf_cuts_bdt_multiclass::labelsSpecies, o2::analysis::hf_cuts_bdt_multiclass::labelsPt}, "ONNX file names for ML models"}' from '<brace-enclosed initializer list>' to 'o2::framework::Configurable<o2::framework::LabeledArray<std::__cxx11::basic_string<char> > >'
 1227 |   Configurable<LabeledArray<std::string>> onnxFileNames{"onnxFileNames", {hf_cuts_bdt_multiclass::onnxFileNameSpecies[0], 5, 1, hf_cuts_bdt_multiclass::labelsSpecies, hf_cuts_bdt_multiclass::labelsPt}, "ONNX file names for ML models"};
      |                                                                                                                                                                                                                                          ^
      |                                                                                                                                                                                                                                          |
      |                                                                                                                                                                                                                                          <brace-enclosed initializer list>
```
[log.txt](https://github.com/AliceO2Group/O2Physics/files/14847459/log.txt)

vkucera: I can confirm seeing the same errors as you do.
There are 2 other files in O2Physics with a `Configurable<LabeledArray<std::string>>` and, although the MegaLinter logs are no longer available for the PRs that introduced those configurables, I can see that the checks failed as well, so I don't have an example that would pass.
In this case, I admit I don't understand what cpplint considers wrong in your implementation so I would be fine with merging it regardless.
On the other hand, it would be good to understand why the conversion from `static const char*` fails. Maybe @aalkin could clarify whether this is expected.

fgrosa: Hi @vkucera many thanks for checking and confirming! Megalinter would not complain if the `string` is not global/static, so if I define it directly in the task, but I don't particularly like this solution.

alibuild (approved):

PWGHF/Core/SelectorCuts.h
aalkin: I'm a bit late for the party, but for the static strings consider using `std::string_view` instead of `std::string` that is allocatable and thus cannot really be static. Labeled array with strings is special, since string is not a simple type and the constructor therefore expects an array of string objects or an array of something that can be converted to string, but since `const char*` is an "array" in itself, its interpretation is ambiguous.

PR #5444, opened 2024-04-02
PWGCF: Fix compilation warnings: missing-braces

vkucera:


vkucera:
> Surprised that the clang-format didn't catch the comments ordering! Thanks Vit for pushing this!!!

Indeed, I also expected the formatting to fail.

PR #5445, opened 2024-04-02
Trigger compilation errors on warnings

vkucera:
Trigger compilation errors on warnings from the following categories:
- default,
- `-Wall`,
- `-Wextra`.

Warnings present in at least one of the categories above are grouped into the following lists:
- common (present for both `clang` 15 (Mac builds) and `gcc` 12 (Linux builds)),
- only `clang`,
- only `gcc`.

Lists were taken from https://github.com/pkolbus/compiler-warnings

`-Wsign-compare` will be added later when cleaned up in the current code base (hundreds).

Requires:
- https://github.com/alisw/alidist/pull/5516


vkucera:
To be finalised after SQM. (Following up on [the discussion at WP14](https://indico.cern.ch/event/1416151/#29-warnings-in-o2physics).)

dependencies/O2PhysicsCompileFlags.cmake
```diff
@@ -11,13 +11,53 @@
...
...nitialized-value;analyzer-write-to-const;analyzer-write-to-string-literal;array-bounds=1;array-compare;array-parameter=2;bool-compare;builtin-declaration-mismatch;cannot-profile;cast-function-type;catch-value=1;class-memaccess;clobbered;coverage-invalid-line-number;coverage-mismatch;dangling-pointer=2;format-contains-nul;format-diag;format-overflow=1;format-truncation=1;format=1;if-not-aligned;implicit-fallthrough=3;inherited-variadic-ctor;init-list-lifetime;init-self;interference-size;invalid-memory-model;literal-suffix;lto-type-mismatch;maybe-uninitialized;memset-elt-size;mismatched-dealloc;missing-attributes;missing-profile;missing-requires;missing-template-keyword;multistatement-macros;non-template-friend;nonnull-compare;openmp-simd;overflow;packed-bitfield-compat;packed-not-aligned;pmf-conversions;prio-ctor-dtor;restrict;return-local-addr;scalar-storage-order;sized-deallocation;strict-aliasing=3;strict-overflow=1;stringop-overflow=2;stringop-overread;stringop-truncation;subobject-linkage;switch-outside-range;switch-unreachable;sync-nand;terminate;tsan;type-limits;unused-local-typedefs;use-after-free=2;virtual-move-assign;vla-parameter;volatile-register-var;zero-length-bounds")
```
ktf, line 24: I am not sure this is a great idea, actually. Can't you simply enable larger classes of warning?

ktf (changes requested):
Can you check if this does not explode the log size?

dependencies/O2PhysicsCompileFlags.cmake
vkucera: As listed in the PR description, these should correspond to the union of default, `-Wall` and `-Wextra` for the respective compilers. If we enable larger classes, don't we risk getting new errors from new warnings added into those classes?

vkucera: Following up on our private discussion, I have simplified the flags by treating all enabled warnings as errors and defining lists for the excluded ones. See 75fb9c0.

vkucera:
> Can you check if this does not explode the log size?

It does not.

ktf:
There is some other warning -> error. Could you have a look?

vkucera:
> There is some other warning -> error. Could you have a look?

If you can confirm that this PR looks fine for you, I'm ready to open PRs fixing the existing warnings so that this one can pass the build tests.

vkucera:
> > There is some other warning -> error. Could you have a look?
> 
> If you can confirm that this PR looks fine for you, I'm ready to open PRs fixing the existing warnings so that this one can pass the build tests.

Let me rephrase: If all the build tests pass, will you be fine with approving and merging this PR as it is now?

victor-gonzalez (approved):
Thanks!! @vkucera

rolavick (dismissed):

vkucera:
> There is some other warning -> error. Could you have a look?

@ktf All green now. Can you please approve?

victor-gonzalez (approved):
Thanks!! @vkucera

PR #5451, opened 2024-04-02
Remove unused constant kHighFv0Flat

mpuccio:
This was causing a bug in the interface with the CEFP by increasing the size of the keepEvent array and shifting the bits by 1


alibuild (approved):

PR #5453, opened 2024-04-02
Fix for `-Wnonull-compare` warning

victor-gonzalez:
Also for not used variable


alibuild (approved):

PR #5456, opened 2024-04-02
Standardise scalers reporting in LF

mpuccio:
This is an example to be followed by the other PWGs to enable automatic verification of the consistency of the scalers with the CEFP reporting


mpuccio:
@skundu692 @ercolessi @ChiaraDeMartin95, I modified your code too. I will soon ask all the other PWGs to adapt their summary histogram to have the same format. @ercolessi @ChiaraDeMartin95 : in the case of the strangeness histogram I found a shift in the labels, now it should be fine, but have a closer look, thanks!

alibuild (approved):

PR #5469, opened 2024-04-04
PWGCF: FemtoUniverse -- added support for arbitrary QA histogram folder names

Eloviyo:
- added optional argument for the QA histogram init() to specify general histogram folder names. By default, the original behavior is used, in which the folder names are generated by particle name and the suffix.
- The V0 task now uses the extended functionality to specify folder names for the different V0 types, which was not possible before.

### DIFF
--- a/PWGCF/FemtoUniverse/Core/FemtoUniverseParticleHisto.h
+++ b/PWGCF/FemtoUniverse/Core/FemtoUniverseParticleHisto.h
@@ -20,6 +20,7 @@
 #define PWGCF_FEMTOUNIVERSE_CORE_FEMTOUNIVERSEPARTICLEHISTO_H_
 
 #include <string>
+#include <optional>
 #include "PWGCF/FemtoUniverse/DataModel/FemtoDerived.h"
 #include "Framework/HistogramRegistry.h"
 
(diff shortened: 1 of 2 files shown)


PWGCF/FemtoUniverse/Core/FemtoUniverseParticleHisto.h
```diff
@@ -339,18 +341,25 @@ class FemtoUniverseParticleHisto
   template <bool isMC, bool isDebug, typename T>
   void fillQA(T const& part)
   {
+    fillQA2<isMC, isDebug, T>(part, HIST(o2::aod::femtouniverseparticle::ParticleTypeName[mParticleType]) + HIST(mFolderSuffix[mFolderSuffixType]));
```
saganatt, line 344: Do you need `fillQA2` if `fillQA` does nothing apart from calling `fillQA2`?
If yes, could you give more descriptive names? E.g., `fillQA` for the outer function and `doFillQA` for the inner? This at least suggests which function wraps which one.

PWGCF/FemtoUniverse/Core/FemtoUniverseParticleHisto.h
```diff
@@ -209,62 +211,62 @@ class FemtoUniverseParticleHisto
   /// Called by init both in case of reconstructed data/ Monte Carlo, and for Monte Carlo Truth
   /// \tparam T Data type of the particle
   /// \param part Particle
-  template <o2::aod::femtouniverseMCparticle::MCType mc, typename T>
-  void fillQA_base(T const& part)
+  template <o2::aod::femtouniverseMCparticle::MCType mc, typename T, typename H>
+  void fillQA_base(T const& part, H const& h)
```
saganatt, line 215: Nice optimization!
Please use more speakable names. Soon nobody will remember what `h` stands for. For example, `histFolder`.

saganatt (changes requested):

PWGCF/FemtoUniverse/Core/FemtoUniverseParticleHisto.h
Eloviyo: So the idea is to be able to plot QA plots for each V0 particle for V0-V0 combinations without affecting the default settings. fillQA2 was added to accept a general histogram path. The original fillQA is retained because other analysis still use it. fillQA uses the fillQA2 with the template h set to the original default behavior.

PWGCF/FemtoUniverse/Core/FemtoUniverseParticleHisto.h
Eloviyo: Thanks, will do so.

saganatt (approved):

PR #5479, opened 2024-04-04
Adding global tracks without DCA as possible input for jet finding

hahassan7:
Adding globalTrackWoDCA as a possible track selection to be given for the jet finder. Need in some HF-jet measurements.


PWGJE/Core/JetDerivedDataUtilities.h
```diff
@@ -341,6 +344,9 @@ uint8_t setTrackSelectionBit(T const& track)
   if (track.isGlobalTrackWoPtEta()) {
     SETBIT(bit, JTrackSel::globalTrack);
   }
+  if (track.isGlobalTrackWoDCA()) {
```
nzardosh, line 347: Hi Hadi,

Do you know if this track selection applies the pT and eta cut already in the track selection task? We should try to avoid this so we can set them ourselves in the jet finder

PR #5484, opened 2024-04-04
PWGEM/PhotonMeson: Include true pi0 for EMC with converted photons

nstrangm:
- Add function that checks the decay chain of a cluster to see, whether it originated from a gamma from a pi0 or eta decay
- Include true pi0s in output whose photons converted in the detector material (mostly TOF)


mhemmer-cern (approved):
Looks good

PR #5493, opened 2024-04-05
Fix logic and reporting of table auto-enabling

vkucera:
Current problematic behaviour:

- `if (required && flag == 0)`:
  - info "Table disabled but required" (correct, but should be at least warning if not error or fatal)
  - `flag = 0` (redundant)
  - info "Table disabled and not required" (incorrect)
- `if (!required && flag == 0)`:
  - `flag = 0` (redundant)
  - info "Table disabled and not required" (correct)
- `if (!required && flag < 0)`:
  - `flag = 0` (correct)
  - info "Table disabled and not required" ("disabled" is misleading when referring to -1)
 
Expected behaviour:

- `if (required && flag == 0)`:
  - warn "Table disabled but required"
- `if (!required && flag == 0)`:
  - info "Table disabled and not required"
- `if (!required && flag < 0)`:
  - `flag = 0`
  - info "Auto-disabling table"

Also in the `bool flag` version, setting `flag = false` is redundant.

@njacazio please check.

### DIFF
--- a/Common/Core/TableHelper.cxx
+++ b/Common/Core/TableHelper.cxx
@@ -68,7 +68,6 @@ void enableFlagIfTableRequired(o2::framework::InitContext& initContext, const st
     LOG(info) << "Auto-enabling table: " + table;
     return;
   }
-  flag = false;
   LOG(info) << "Table disabled and not required: " + table;
 }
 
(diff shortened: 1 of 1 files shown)


njacazio:
Looks good, thanks Vit

vkucera:
> Looks good, thanks Vit

Thanks @njacazio

ddobrigk (approved):

PR #5502, opened 2024-04-05
Common: Fix bug in inel > 0 embedding

ddobrigk:
@nepeivodaRS many thanks for catching this!!


alibuild (approved):

PR #5518, opened 2024-04-08
Add detector/ITS cluster map QA plots with both pos. and neg. info

romainschotter:
Add QA plots containing det. and ITS cluster maps of both pos. and neg. daughters in order to correlate pos. and neg. configurations


romainschotter:
@ddobrigk

njacazio (approved):

PR #5525, opened 2024-04-08
[event plane] fix memory lick

jikim1290:
fix memory lick


jgrosseo (approved):

jgrosseo:
If this fixes it, we should inform Anton because then something is wrong in the framework...

ktf:
Did it actually fix it?

jikim1290:
@ktf not really.. could you please check dm in mattermost? thanks

PR #5529, opened 2024-04-08
PWGEM/PhotonMeson: add RZ cut to reconstructed-true photon in MC

dsekihat:
PWGEM/PhotonMeson: add RZ cut to reconstructed-true photon in MC


alibuild (approved):

PR #5537, opened 2024-04-09
[event plane] using shared_ptr

jikim1290:
using shared_ptr

### DIFF
--- a/Common/TableProducer/qVectorsTable.cxx
+++ b/Common/TableProducer/qVectorsTable.cxx
@@ -111,15 +111,15 @@ struct qVectorsTable {
   // Enable access to the CCDB for the offset and correction constants and save them
   // in dedicated variables.
   Service<o2::ccdb::BasicCCDBManager> ccdb;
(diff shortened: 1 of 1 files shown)


Common/TableProducer/qVectorsTable.cxx
```diff
@@ -185,10 +185,10 @@ struct qVectorsTable {
         if (cfgFT0CCorr->size() < 48) {
           LOGF(fatal, "No proper correction factor assigned for FT0C");
         } else {
-          cfgCorr.push_back(cfgFT0CCorr);
+          cfgCorr.push_back(std::make_shared<std::vector<float>>(cfgFT0CCorr));
         }
       } else {
-        cfgCorr.push_back(*(objft0c));
+        cfgCorr.push_back(std::make_shared<std::vector<float>>(*(objft0c)));
```
jgrosseo, line 191: I think there is still a leak here because the object returned by getForTimeStamp is not deleted...

PR #5562, opened 2024-04-10
PWGHF: Adding utility function for single track cut on min/max dcaxy

stefanopolitano:
This PR contains a utility function for the track's DCA selection. 

- The selection is added in the `candidateSelectorDplusToPiKPi.cxx`
- The utility function is applied in the `candidateSelectorDsToKKPi.cxx` and  `trackIndexSkimCreator.cxx`

### DIFF
--- a/PWGHF/TableProducer/candidateSelectorDplusToPiKPi.cxx
+++ b/PWGHF/TableProducer/candidateSelectorDplusToPiKPi.cxx
@@ -54,6 +54,9 @@ struct HfCandidateSelectorDplusToPiKPi {
   // topological cuts
   Configurable<std::vector<double>> binsPt{"binsPt", std::vector<double>{hf_cuts_dplus_to_pi_k_pi::vecBinsPt}, "pT bin limits"};
   Configurable<LabeledArray<double>> cuts{"cuts", {hf_cuts_dplus_to_pi_k_pi::cuts[0], hf_cuts_dplus_to_pi_k_pi::nBinsPt, hf_cuts_dplus_to_pi_k_pi::nCutVars, hf_cuts_dplus_to_pi_k_pi::labelsPt, hf_cuts_dplus_to_pi_k_pi::labelsCutVar}, "Dplus candidate selection per pT bin"};
+  // DCAxy selections
+  Configurable<LabeledArray<double>> cutsSingleTrack{"cutsSingleTrack", {hf_cuts_single_track::cutsTrack[0], hf_cuts_single_track::nBinsPtTrack, hf_cuts_single_track::nCutVarsTrack, hf_cuts_single_track::labelsPtTrack, hf_cuts_single_track::labelsCutVarTrack}, "Single-track selections"};
+  Configurable<std::vector<double>> binsPtTrack{"binsPtTrack", std::vector<double>{hf_cuts_single_track::vecBinsPtTrack}, "track pT bin limits for DCA XY pT-dependent cut"};
   // QA switch
   Configurable<bool> activateQA{"activateQA", false, "Flag to enable QA histogram"};
   // ML inference
@@ -161,7 +164,24 @@ struct HfCandidateSelectorDplusToPiKPi {
     if (std::abs(candidate.maxNormalisedDeltaIP()) > cuts->get(pTBin, "max normalized deltaIP")) {
       return false;
     }
+    if (!isSelectedCandidateDcaXY(candidate)) {
+      return false;
+    }
     return true;
+  }
+
+  /// Single-track cuts
+  /// \param candidate is the Ds candidate
+  /// \return true if all the prongs pass the selections
+  template <typename T1>
+  bool isSelectedCandidateDcaXY(const T1& candidate)
+  {
+    if (isSelectedTrackDcaXY(binsPtTrack, cutsSingleTrack, candidate.ptProng0(), candidate.impactParameter0()) &&
+        isSelectedTrackDcaXY(binsPtTrack, cutsSingleTrack, candidate.ptProng1(), candidate.impactParameter1()) &&
+        isSelectedTrackDcaXY(binsPtTrack, cutsSingleTrack, candidate.ptProng2(), candidate.impactParameter2())) {
+      return true;
+    }
+    return false;
   }
 
   /// Apply PID selection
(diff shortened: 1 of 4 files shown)


PWGHF/Utils/utilsAnalysis.h
```diff
@@ -36,6 +36,27 @@ int findBin(T1 const& binsPt, T2 value)
   }
   return std::distance(binsPt->begin(), std::upper_bound(binsPt->begin(), binsPt->end(), value)) - 1;
 }
+
+/// Single-track cut on DCAxy
+/// \param trackPt is the prong pT
+/// \param dcaXY is the prong dcaXY
+/// \return true if track passes all cuts
```
vkucera, line 43: Please complete the documentation

PWGHF/Utils/utilsAnalysis.h
```diff
@@ -36,6 +36,27 @@ int findBin(T1 const& binsPt, T2 value)
...
+template <typename T1, typename T2>
+bool isSelectedTrackDcaXY(T1 const& binsPtTrack, T2 const& cutsSingleTrack, const float& trackPt, const float& dcaXY)
```
vkucera, line 45: Do not pass simple types by reference.
```suggestion
bool isSelectedTrackDcaXY(T1 const& binsPt, T2 const& cuts, const float pt, const float dcaXY)
```

PWGHF/Utils/utilsAnalysis.h
```diff
@@ -36,6 +36,27 @@ int findBin(T1 const& binsPt, T2 value)
...
+{
+  auto pTBinTrack = findBin(binsPtTrack, trackPt);
```
vkucera, line 47:
```suggestion
  auto binPt = findBin(binsPtTrack, trackPt);
```

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -657,7 +636,7 @@ struct HfTrackIndexSkimCreatorTagSelTracks {
     iCut = 5;
     if (statusProng > 0) {
       for (int iCandType = 0; iCandType < CandidateType::NCandidateTypes; ++iCandType) {
-        if (TESTBIT(statusProng, iCandType) && !isSelectedTrackDCA(trackPt, dca, iCandType)) {
+        if (TESTBIT(statusProng, iCandType) && !isSelectedTrackDcaXY(binsPtTrack, &cutsSingleTrack[iCandType], trackPt, dca[0])) {
```
vkucera, line 639:
```suggestion
        if (TESTBIT(statusProng, iCandType) && !isSelectedTrackDcaXY(binsPtTrack, cutsSingleTrack[iCandType], trackPt, dca[0])) {
```
Shouldn't the `&` be removed?

vkucera (changes requested):

PWGHF/TableProducer/trackIndexSkimCreator.cxx
stefanopolitano: Hi @vkucera ! Thanks a lot for your review! Regarding this point, if I remove the `&` I get this error:

`PWGHF/Utils/utilsAnalysis.h:53:25: error: base operand of '->' has non-pointer type 'const o2::framework::LabeledArray<double>'` 

I guess this is because, in the utils function, I am getting the selection information assuming `cuts` to be a pointer as done in the D meson selector. I found the `&` solve the issue, but I don't know if there are other solutions. If you have any suggestions on this just let me know and I will try

vkucera: OK. In the selectors it is `Configurable<LabeledArray>` (which is a pointer) while in the skim creator it is just `<LabeledArray>` which is not a pointer.
Then maybe, you can make it explicit in the function signature: `const T2* const cutsSingleTrack`

stefanopolitano: Hi @vkucera , thanks a lot for the suggestion! Just for my understanding, you are suggesting modifying the definition of the function making `T2` a pointer explicitly without changing the rest, or am I missing something? Because if I try that I get the following error for the Dplus selector (and similarly for the Ds):

`PWGHF/TableProducer/candidateSelectorDplusToPiKPi.cxx:179:29: note:   mismatched types 'const T2*' and 'o2::framework::Configurable<o2::framework::LabeledArray<double> >'`,

which I am not sure how to fix. If you have any further suggestions I will try them out

stefanopolitano: Hi @vkucera, do you have any further suggestions? Thanks!

PWGHF/TableProducer/candidateSelectorDplusToPiKPi.cxx
```diff
@@ -161,9 +164,26 @@ struct HfCandidateSelectorDplusToPiKPi {
...
+  /// \return true if all the prongs pass the selections
+  template <typename T1>
+  bool isSelectedCandidateDcaXY(const T1& candidate)
+  {
+    if (isSelectedTrackDcaXY(binsPtTrack, cutsSingleTrack, candidate.ptProng0(), candidate.impactParameter0()) &&
+        isSelectedTrackDcaXY(binsPtTrack, cutsSingleTrack, candidate.ptProng1(), candidate.impactParameter1()) &&
+        isSelectedTrackDcaXY(binsPtTrack, cutsSingleTrack, candidate.ptProng2(), candidate.impactParameter2())) {
+      return true;
+    }
+    return false;
```
vkucera, line 184: You can return the `bool` value of the condition directly if you don't plan to add other checks. If you do, plan to add other checks, it is better to invert the logic and return `false` as soon as one check fails and return `true` at the end otherwise.
```suggestion
    return (isSelectedTrackDcaXY(binsPtTrack, cutsSingleTrack, candidate.ptProng0(), candidate.impactParameter0()) &&
        isSelectedTrackDcaXY(binsPtTrack, cutsSingleTrack, candidate.ptProng1(), candidate.impactParameter1()) &&
        isSelectedTrackDcaXY(binsPtTrack, cutsSingleTrack, candidate.ptProng2(), candidate.impactParameter2()));
```

stefanopolitano: Yes, indeed, so far I  want to check the single-track DCA, hence I can directly return the bool value of the condition. I will implement this suggestion and update the PR. Thanks a lot!

PWGHF/Utils/utilsAnalysis.h
vkucera: Not resolved

vkucera: Ah, sorry. You replaced the commits.

vkucera:
Please do not overwrite the PR history by force-pushing.

vkucera (approved):

PR #5563, opened 2024-04-10
beta-gamma cutoff, enable flags per species for NN

jezwilkinson:
Addresses several issues that popped up with NN application in trigger selection:

- impose (configurable) cutoff in expected beta-gamma to remove artifacts reported by PWG-CF in very low-momentum antiprotons where network extrapolation causes issues (below this value only BB is used, betagamma >0.45 is set as a conservative default ) - plots below are expected dE/dx and nsigma for protons, (left) without the betagamma cutoff and (right) with the cutoff
image
image
- add protections for case where the collision table is not populated (CCDB timestamp request now only references the collision table if the track is explicitly associated to a collision, and from the BC table directly otherwise. Also disables NN correction for tracks with no associated collision, as the application uses info from the collision table)
- add configurable flags to allow per-species disabling of NN application - default value is 1 for each species, and should be set to 0 for each species where it should be disabled (the global "useNetworkCorrection" still rules here and does not apply at all if set to 0, regardless of the species switches)

pinging for info: @wiechula @ChSonnabend @mpuccio @lietava


mpuccio:
Hello, just to say that for the trigger we actually used it in production and it works, it would be great to have it also merged in master.

iarsene (approved):

PR #5570, opened 2024-04-11
[PWGDQ] Generalisation of cumulant calculation to weighted Q-vectors

Elros60:
This PR will generalise the way we calculate correlations for cumulant method together with a rearrangement of variables regarding the generalised framework. Necessary information will be provided in `RTDIMUONALL` and `REREFFLOW` tables for later-on post-processing stage for `c_n` and `v_n`.


lucamicheletti93 (changes requested):
Dear @Elros60, this PR modify the data model for the q-vector table, therefore it can break the processing of the latest DQ strams for the Pb-Pb. I would suggest to convert this PR as draft and postpone the merging before the next large scale productions.

PR #5573, opened 2024-04-11
PWGMM: Mult: use concepts instead of functions

aalkin:


aalkin:
@vkucera Note that Megalinter has a false positive here - it confuses `concept` syntax with C-style cast. Is there any updated configuration that knows about c++20?

vkucera:
> @vkucera Note that Megalinter has a false positive here - it confuses `concept` syntax with C-style cast. Is there any updated configuration that knows about c++20?

Indeed, that looks worth reporting in the cpplint repo. You can add `// NOLINT` on the two lines for now.
I am not aware of any C++20 configuration.

alibuild (approved):

PR #5577, opened 2024-04-11
PWGHF: Ds-h correlations, minor adjustments in assoc particle efficiency evaluation

scattaru:
Dear all,
- I've fixed the evaluation of associated particle efficiency requiring to use only particles coming from sel8 events for the reconstructed part (numerator). 
- I've fixed the secondary particle contamination evaluation requiring that the Ds meson used to build the correlations is a real signal candidate.

Let me know if there are any comments,
Best regards,
Samuele


scattaru:
Dear all,
Following what it was reported by @fgrosa in the HF O2 Analysis channel on the usage of sel8 in the processes which use MC samples not anchored I think that inserting the selection as I did the process will not work anymore. So I can add a configurable to choose if I want to apply the sel8 or not which seems to me the fast way at the moment or I can also remove that part. For the moment I would convert this PR in a draft.

Best regards,
Samuele

scattaru:
Hi @fgrosa, 
Thanks a lot for the suggestion, in future I'll rewrite that part of the code following the way it is done in the TrackIndexSkimCreator

Thanks,
Samuele

PR #5586, opened 2024-04-12
Revert "PWGMM: Mult: use concepts instead of functions (#5573)"

aalkin:


alibuild (approved):

PR #5597, opened 2024-04-12
[PWGLF] Simplify track selection options

BongHwi:
- Simplify the track cuts for a better systematic check
- Increase the histogram size from float to double type.


alibuild (approved):

PR #5618, opened 2024-04-15
modifing randomness mechanism in a pair rel. momentum calculation

glromane:
Dear @victor-gonzalez ,

In this commit we're trying to improve the randomness mechanism for a pair assignment. In case of 3d we're sensitive to the projections of the relative momentum (in 1d we take an abs. value -- irrelevant). In principle pairs must be the same but the pair order [particle1, particle2] must be random. The pairing has a pattern and in order to eliminate it we used shuffling before, this time we just randomly choose the sign of the momenta difference p1 - p2 I.e. order. We expect it to be a more correct way of eliminating any patterns. Could you please approve this commit? Thank you.

Sincerely yours, Gleb.


victor-gonzalez (approved):
Nice!
I wonder if, as you peak the time since epoch, a pattern in the execution time would not as well bias the pair ordering

PR #5633, opened 2024-04-16
implementation of TuneOnData for MC in TPC PID

jezwilkinson:
Original pull request for TuneOnData https://github.com/AliceO2Group/O2Physics/pull/4733 had diverged due to large conflicting changes made in https://github.com/AliceO2Group/O2Physics/pull/5563. This commit resolves the conflicts and re-implements the tuneondata option based on the new NN configurables (beta-gamma cutoff and per-species "useNetwork" flags).

### DIFF
--- a/Common/Core/PID/TPCPIDResponse.h
+++ b/Common/Core/PID/TPCPIDResponse.h
@@ -77,6 +77,9 @@ class Response
   /// Gets the number of sigmas with respect the expected value
   template <typename CollisionType, typename TrackType>
   float GetNumberOfSigma(const CollisionType& collision, const TrackType& trk, const o2::track::PID::ID id) const;
+  // Number of sigmas with respect to expected for MC, defining a tune-on-data signal value
+  template <typename CollisionType, typename TrackType>
+  float GetNumberOfSigmaMCTuned(const CollisionType& collision, const TrackType& trk, const o2::track::PID::ID id, float mcTunedTPCSignal) const;
   /// Gets the deviation to the expected signal
   template <typename TrackType>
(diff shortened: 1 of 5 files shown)


jezwilkinson:
(for information: @wiechula @njacazio )

njacazio:
Hi, this looks good, shall we release it and enable a test wagon so that analysers can test it?

wiechula:
Fine from my side. But I would not enable it by default.

njacazio:
> Fine from my side. But I would not enable it by default.

I fully agree

njacazio:
Hi @jezwilkinson what do you think? Can we go ahead?

jezwilkinson:
Hi @njacazio ,
Sorry, I was away on a group event and completely offline the last couple of days.
Sounds great to me! 
> "I would not enable it by default."
Indeed the process switch is defaulted to false (not least because the TuneOnData process loop requires subscription to MC info and so wouldn't work on real data). We can announce in the mattermost how to use it

ddobrigk (approved):
thanks a lot!

njacazio:
Hi, I propose that we add a debug wagon with the tune on data enabled. I will do this tomorrow with the new tag and test it

jgrosseo:
@jezwilkinson @ddobrigk This PR introduced a problem: Now TPC full and TPC tiny produce both the same table which means on Hyperloop wagons which require full and wagons which require tiny cannot be combined into a train anymore. It should be removed from one of the two, and if that is not good, be moved to a separate executable...
image

njacazio:
Indeed https://github.com/AliceO2Group/O2Physics/pull/5946

PR #5660, opened 2024-04-17
PWGMM: Mult: remove MC-only parameter from training features

aalkin:


alibuild (approved):

PR #5692, opened 2024-04-18
PWGHF: insert try-catch for secondary vertex fitting

mfaggin:
Protection against corner-cases that can cause `invalid track covariance` run-time errors when enabling the track-tuner in new MC productions.
Some tests done on a file of `LHC24d3` show that these corner-cases are of the order of `10^-5` (see an example below with `candidateCreator3Prong`):

c


fgrosa (approved):
Thanks a lot @mfaggin!

fcatalan92 (approved):

PR #5696, opened 2024-04-18
PWGEM/PhotonMeson: change default values for configurables

dsekihat:
PWGEM/PhotonMeson: change default values for configurables


alibuild (approved):

PR #5705, opened 2024-04-19
PWGHF: reducing number of bins for ML scores in D0 output ThnSparse

Mingyu3360715:
The number of bins for Bkg and FD score are reduced from 100 to 50, which allows us to able to create smaller AnalysisResults.root that can be merged on a large dataset. @atavirag

### DIFF
--- a/PWGHF/D2H/Tasks/taskD0.cxx
+++ b/PWGHF/D2H/Tasks/taskD0.cxx
@@ -74,7 +74,7 @@ struct HfTaskD0 {
   Partition<D0CandidatesMlMc> selectedD0CandidatesMlMc = aod::hf_sel_candidate_d0::isRecoHfFlag >= selectionFlagHf;
   Partition<D0CandidatesMlMcKF> selectedD0CandidatesMlMcKF = aod::hf_sel_candidate_d0::isRecoHfFlag >= selectionFlagHf;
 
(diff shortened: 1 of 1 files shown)


mfaggin:
Hi @Mingyu3360715 , in my opinion this would not change dramatically the situation, because you have other axes that have much more bins (for example you an axis with 360 bins for the pt. Whats the reason for such fine binning? The only one that I see is pt shape systematics in MC but nothing else).

My very strong suggestion is to make all the axes configurables, i.e. use `ConfigurableAxis` so that at runtime you can decide which binning to use avoiding doing new PRs every time.

mfaggin (changes requested):
Consider using `ConfigurableAxis` (see an example here https://github.com/AliceO2Group/O2Physics/blob/master/PWGHF/D2H/Tasks/taskLc.cxx#L43)

Mingyu3360715:
> Consider using `ConfigurableAxis` (see an example here https://github.com/AliceO2Group/O2Physics/blob/master/PWGHF/D2H/Tasks/taskLc.cxx#L43)

Hi @mfaggin , thanks a lot for the instruction. I have updatedthe ThnSparse axes with configurable axes. Please have a look when you are available.

mfaggin (changes requested):
hi @Mingyu3360715 , you still miss the usage of `AxisSpec` using your `ConfigurableAxis` as input. Please, have a look at the link I provided before and, on top of recompiling locally, try to run it before committing.

Mingyu3360715:
Hi @mfaggin , I've added the missing AxisSpec and tested locally. I see Andrea also made the PR while I was away in the past hour. Our commit looks to be same essentially, so it is fine to use any one of them.

PR #5706, opened 2024-04-19
PWGMM: Mult: granular event selection

aalkin:
Granular event selection to be able to exclude TF border selection in MC.


alibuild (approved):

PR #5710, opened 2024-04-19
PWGHF: Use `pVector` instead of `std::array{px, py, pz}`

vkucera:
- Simplifies code by using `pVector` columns from `Tracks`, `McParticles` and `ReducedDataModel.h` tables instead of 
explicitly constructing arrays from `px`, `py`, `pz`.
- Simplifies calculations with momenta, using proper `RecoDecay` functions.
- Limits the scope of some local variables in `trackIndexSkimCreator.cxx`.

Requires:
- https://github.com/AliceO2Group/AliceO2/pull/12991
- https://github.com/AliceO2Group/AliceO2/pull/13043


fgrosa (approved):
Hi @vkucera it looks good to me (I hope I did not overlooked any instance), thanks a lot!

PR #5718, opened 2024-04-19
adding table with pileup flags and new analysis distributions

glromane:
Dear @victor-gonzalez ,

In this commit we add an additional table with pile-up flags to the data model. Task "singleTrackSelectorExtra.cxx" just produces a dummy table with the flags and needed only for compatibility with the derived data produced earlier, since the analysis tasks now use this additional collision table.

We also implemented a momentum cut (to the table producer) starting from which we remove tracks that are out of stored TOF Nsigma range -- this would decrease the derived data size significantly.

The MC task is now uses cent. and pair kT binning (as the main Pair task).
The Pair task also now includes dEta(dPhi*) histograms -- we were asked to add them also to the data distributions at the PAG (before it was only in MC).

Could you please approve this commit? Thank you.

Sincerely yours, Gleb.

@njacazio @ercolessi @sofiatomassini

### DIFF
--- a/PWGCF/Femto3D/TableProducer/singleTrackSelector.cxx
+++ b/PWGCF/Femto3D/TableProducer/singleTrackSelector.cxx
@@ -171,7 +171,8 @@ struct singleTrackSelector {
 
       for (auto ii : particlesToKeep)
         if (o2::aod::singletrackselector::TPCselection(track, std::make_pair(ii, keepWithinNsigmaTPC))) {
-          if(track.p() > _pRemoveTofOutOfRange && !o2::aod::singletrackselector::TOFselection(track, std::make_pair(ii, std::vector<float>{-10.0, 10.0}), 10.0)) continue;
+          if (track.p() > _pRemoveTofOutOfRange && !o2::aod::singletrackselector::TOFselection(track, std::make_pair(ii, std::vector<float>{-10.0, 10.0}), 10.0))
+            continue;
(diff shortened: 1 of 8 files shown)


PWGCF/Femto3D/TableProducer/singleTrackSelector.cxx
```diff
@@ -100,7 +102,7 @@ struct singleTrackSelector {
   std::vector<int> particlesToKeep;
   std::vector<int> particlesToReject;
 
-  void init(InitContext&)
+  void init(InitContext& context)
```
victor-gonzalez, line 105: Please, don't include `context` as you are not using it
There is a campaign to convert the compiler warning to errors to help to improve the code error detection
Unused variable is one of the warnings that are being removed before converting it to an error

victor-gonzalez (changes requested):
Please, additionally, fix the compiler warnings
There is an initiative to make the compiler warning as errors and before doing that all the warnings have to be removed

victor-gonzalez (dismissed):

victor-gonzalez:
There seems to be still a few warnings in `femto3DPairTask.cxx` and `femto3DPairTask.h`
You will be able to see them in the `Details` link of the `build/O2Physics/o2` check
Could you please fix them?
Thanks!

victor-gonzalez (approved):

PR #5724, opened 2024-04-20
fixed the low entries in the ME background 

sawankumawat:
…upper index policy I used for the ME background, but since in the event mixing tracks are coming from the different events, we do not need to exclude the (1,2) and (2,1) pairs in case of Ks Ks


sawankumawat:
@BongHwi Please approve.
Thank you.

BongHwi (approved):

PR #5727, opened 2024-04-20
[PWGDQ] Add complete resolution factors for SP and EP with 3 subevents using full TPC, FT0A, FT0C

Elros60:
* Add information for flow analysis in `DileptonsFlow` table
* Add collisionId to do proper averaging over events
* Add correlation configuration for full {3 -3} to have well-defined Q_3
* Add 3 subevents resolutions for SP and EP using full TPC, FT0A, FT0C
* Add new performance plots as function of centrality


PWGDQ/Core/HistogramsLibrary.cxx
lucamicheletti93, line 1: @Elros60 could you create a dedicated histogram sub group for the cross products?

PWGDQ/Tasks/tableReader.cxx
lucamicheletti93, line 1: @Elros60 in the current version of the the DQ streams (https://alimonitor.cern.ch/hyperloop/view-dataset/823) the collision index is not saved, therefore modifying the current process function with this extra requirement will break the running. I would suggest to create dedicated process function with the extension "WithColl". Once the new streams are produced the old one will be deleted

lucamicheletti93 (changes requested):
@Elros60  Thanks a lot for the commit! Please have a look at my minor comments

PWGDQ/Core/HistogramsLibrary.cxx
Elros60: Yes, I put them now in a subgroup called "cross" under the subgroup "qvector".

PWGDQ/Tasks/tableReader.cxx
Elros60: Thanks a lot!
I made a separated process, and brought all old process back as before.

Elros60:
I just rebased my branch with new version of `master' and cleaned up what Victor has done within https://github.com/AliceO2Group/O2Physics/pull/5741 and https://github.com/AliceO2Group/O2Physics/pull/5777 to avoid redundant calculations. I will commit with all new modifications very shortly.

PR #5730, opened 2024-04-20
PWGEM/PhotonMeson: add process EE MC only

dsekihat:
PWGEM/PhotonMeson: add process EE MC only


alibuild (approved):

PR #5757, opened 2024-04-22
PWGHF: Fix compilation warnings

vkucera:
@stefanopolitano, I suppose the missing `break` is a bug, right?


alibuild (approved):

PR #5762, opened 2024-04-22

vkucera:
`type-limits`

@abilandz I commented out lines checking condition which can never be true because of the unsigned type of the `fRandomSeed` member.


abilandz:
> `type-limits`
> 
> @abilandz I commented out lines checking condition which can never be true because of the unsigned type of the `fRandomSeed` member.

Thanks, noted!

victor-gonzalez (approved):

PR #5764, opened 2024-04-22
Common: Fix compilation warnings

vkucera:
`ignored-qualifiers`
`sign-compare`
`implicit-fallthrough`


Common/TableProducer/mcCollsExtra.cxx
```diff
@@ -135,7 +135,7 @@ struct mcCollisionExtra {
       if (iter != sortedIndices.end()) {
         int index = iter - sortedIndices.begin();
         for (int iMcColl = index + 1; iMcColl < index + 17; iMcColl++) {
```
aalkin, line 137: Declaring `iMcCol1` as `size_t` here is probably better.

vkucera: But `index + 1` is not guaranteed to be non-negative, is it?

aalkin: `index` (which should be obtained with `std::distance`, probably) cannot be negative since it should not be before the `.begin()`.

vkucera: Yes, but it has type `int`.

aalkin: This is why I recommend to use `auto index = std::distance(iter, sortedIndices.begin());`

vkucera:
@jgrosseo @iarsene @ktf @ddobrigk Can you check, please?

vkucera:
@jgrosseo @iarsene @ktf @ddobrigk Bump

njacazio (approved):

ddobrigk (approved):
Thank you very much!

PR #5767, opened 2024-04-22
DPG: Fix compilation warnings

vkucera:
`implicit-fallthrough`
@fgrosa , I suppose that you want to check only the condition for the given `channel` and not continue to other checks if the condition is not satisfied, right?


fgrosa:
> `implicit-fallthrough` @fgrosa , I suppose that you want to check only the condition for the given `channel` and not continue to other checks if the condition is not satisfied, right?

Hi @vkucera indeed the `break` were missing, thanks a lot!

PR #5779, opened 2024-04-22
PWGHF: Add ConfigurableAxis to taskD0.cxx

atavirag:
Hi @Mingyu3360715, @mfaggin,

 I added the changes requested in #5705 since it is now nighttime in China, but we are a bit pressed time-wise to get results using a bigger dataset.

Cheers!


mfaggin:
same as PR #5705 , closing it

PR #5783, opened 2024-04-22
Add prefix to configurable group

mpuccio:


mpuccio:
@ariedel-cern @lauraser: without this modification the generation of the json during the compilation was breaking in my mac because there were two configurable with the same name (kstar, then mT...) and this was throwing an exception in boost options. This should not break your code, but please have a look

ariedel-cern:
Hi @mpuccio ,
I think Laura had a similar problem a few days ago. I think we fixed it by updating alidist so we recompiled of O2 with O2Physics which triggered the default .json configuration files to be reproduced. And this fixed the name clash.
Ah, yes, I was just about to add a comment here.

PR #5786, opened 2024-04-22
Fix bit position when having more than 64 triggers

mpuccio:
Brown-paper-bag bug


alibuild (approved):

PR #5792, opened 2024-04-22
PWGCF: dEdx plots for IdentifedBF

bghanley1995:


victor-gonzalez (approved):

victor-gonzalez:
Brian, you have to locally rebase your changes to the latest version of O2 Physics

bghanley1995:
Thank you Victor, I thought I'd updated it properly. I'll make extra sure going forward

victor-gonzalez:
Well, if you look at this PR up, above these comments, you see that there are commits since two months most of them are probably not from you.
Normally if your local repo is updated (pull) with the latest version, only your latest commit should be shown
It is cleaner but also a guarantee that you have the latest locally

PR #5805, opened 2024-04-23
PWGLF - K892 update the binning config

BongHwi:
From continuous binning to split binning.


alibuild (approved):

PR #5808, opened 2024-04-23
resonance study using sgproducer drive data

sdudi123:


abylinkin:
Sorry, I can't approve a copy-paste of my code without significant modifications. 
You didn't even gave the new name to the struct, which I assume could cause build errors.

Otherwise, what is the aim of lines 200-207? Looks like they have no effect.

PR #5815, opened 2024-04-23
PWGLF: Update QA plot to 3d histogram

BongHwi:
Before: 2D histogram (pT vs nSigma)
Now: 3D histogram (multiplicity vs pT vs nSigma)


alibuild (approved):

PR #5816, opened 2024-04-23
Add particle MC ID + ITS chi2/Ncls in derived data + extra event selections in the analysis task

romainschotter:
3 modifications : 
- introduction of the particle MC ID for the V0 mother and decay daughters **in the derived data model** (to keep track of the V0 population and spot duplicated V0s)
- introduction of the ITS chi2/Ncls **in the derived data model** (to separate afterburner tracks from full-tracking tracks)
- introduction of 5 extra event selections **in the analysis task** _derivedlambdakzeroanalysis_ (Not same bunch pile-up, IsVertexITSTPC, IsVertexTOFMatched, IsVertexTRDMatched, IsGoodZvtxFT0VsPV)

@ddobrigk can you please have a look at this PR? 
Thank you very much!


PWGLF/TableProducer/converters/strarawcentsconverter.cxx
```diff
@@ -16,17 +16,50 @@
...
 
+// Converts V0 version 001 to 002
 struct strarawcentsconverter {
+  Produces<aod::DauTrackExtras_001> dauTrackExtras_001;
+  Produces<aod::V0MCCores_001> v0MCCores_001;
   Produces<aod::StraRawCents_001> straRawCents_001;
   Produces<aod::StraRawCents_003> straRawCents_003;
 
-  void processStraRawCents000to001(aod::StraRawCents_000 const& straRawCents_000)
+  void process000to001(aod::StraRawCents_000 const& straRawCents_000, aod::DauTrackExtras_000 const& dauTrackExtras_000, aod::V0MCCores_000 const& v0MCCores_000)
```
ddobrigk, line 26: You should create another converter, I believe, rather than trying to convert different types of tables in only one task. This leads to some issues, as the same task has multiple "Produces" calls and those empty out tables that are not replaced (= if their corresponding process function is disabled). Can you please adjust? Sorry for the trouble, thanks!

ddobrigk (changes requested):

PWGLF/TableProducer/converters/v0sconverter.cxx
```diff
@@ -0,0 +1,59 @@
...
+#include "Framework/runDataProcessing.h"
+#include "Framework/AnalysisTask.h"
+#include "Framework/AnalysisDataModel.h"
+#include "PWGLF/DataModel/LFStrangenessTables.h"
+
+using namespace o2;
+using namespace o2::framework;
+
+// Converts V0 version 001 to 002
+struct v0sconverter {
```
ddobrigk, line 20: This name is a bad idea, considering there is already a 'v0 converter' that is under `Common` in O2Physics. It might be better to rename to v0mccoreconverter or similar...

PWGLF/TableProducer/converters/v0sconverter.cxx
```diff
@@ -0,0 +1,59 @@
...
+  Produces<aod::DauTrackExtras_001> dauTrackExtras_001;
+  Produces<aod::V0MCCores_001> v0MCCores_001;
+
+  void process000to001(aod::DauTrackExtras_000 const& dauTrackExtras_000, aod::V0MCCores_000 const& v0MCCores_000)
```
ddobrigk, line 24: This approach will only work to update AO2Ds that are MC, since your process function cannot be used correctly in real data (as real data AO2Ds do not have `V0MCCores` stored). I am afraid you have to separate into two process functions that can be turned on and off independently...

ddobrigk:
Super thanks @romainschotter ! 
@ChiaraDeMartin95 @ercolessi I think this is ready to merge. When you get a moment, could you please do the honors? Many thanks!

ddobrigk (approved):

njacazio (approved):

PR #5826, opened 2024-04-24
PWGJE: Task for reconstructing the secondary vertices inside the jet

hahassan7:
Implementation of a task that does the 2 and 3 prongs secondary vertex reconstruction inside the jet. It produces two tables, one table with the SV indices inside the jet which is also joinable with the Jet table. Another table with the reconstructed 2-prong and 3-prong vertices.


nzardosh:
Hi hadi, could you please fix the formatting problems?

hahassan7:
The formatter is complaining about a namespace:
  PWGJE/DataModel/JetTagging.h:74:  Namespace should be terminated with "// namespace _datatype_"  [readability/namespace] [5]
 
But the thing is, this whole namespace is inside a #define block, so there is no way we can put it. I saw many codes do the same as I did but for some reasons the formatter doesn't complain about them.

vkucera:
> The formatter is complaining about a namespace: PWGJE/DataModel/JetTagging.h:74: Namespace should be terminated with "// namespace _datatype_" [readability/namespace] [5]
> 
> But the thing is, this whole namespace is inside a #define block, so there is no way we can put it. I saw many codes do the same as I did but for some reasons the formatter doesn't complain about them.

I don't see any of the column definitions using any macro parameters. Why do you not define them in a single namespace?

vkucera:
> > The formatter is complaining about a namespace: PWGJE/DataModel/JetTagging.h:74: Namespace should be terminated with "// namespace _datatype_" [readability/namespace] [5]
> > But the thing is, this whole namespace is inside a #define block, so there is no way we can put it. I saw many codes do the same as I did but for some reasons the formatter doesn't complain about them.
> 
> I don't see any of the column definitions using any macro parameters. Why do you not define them in a single namespace?

In fact, all these columns already exist in the HF data model. Why don't you use those instead of duplicating them?

hahassan7:
But if we fill this table "aod::HFCand3Prongs" will this conflict with the original table reconstructed from the HF framework (if it's on hyperloop) ?
You are right using the namespace from HF framework might save us writing the whole namespace again. I will fix this.

hahassan7:
As we discussed, I remove the duplicate functions runCreator2Prong and runCreator3Prong and merged them into one single function runCreatorNProng<N> where N is the number of prongs for the vertex to be reconstructed from. 
Unfortunately I wasn't able to merged the process functions and we had to keep the DCAFitter, because the DCAFitter must be initialized with a constant expression.

nzardosh (approved):

PR #5838, opened 2024-04-25
PWGLF: Include event correction

BongHwi:
It introduces a new table `ResoMCCollisions` to keep the MC event flags.

Resoinitializer now saves the event correction histogram `hMCEventIndices` when we run the MC process.

The analyser can get the event correction factor, such as trigger efficiency from this.

Similarly, the signal loss correction factor can be calculated.


alibuild (approved):

PR #5845, opened 2024-04-25
Protection against null pointer in CCDB access

ekryshen:


alibuild (approved):

PR #5858, opened 2024-04-26
PWG-DQ: MchEff fix on event filtering

zconesa:


zconesa:
Dear @iarsene , @dsekihat , @lucamicheletti93 
It seems the compilation test is failing on the VarManager which I did not modify. The modification I did is in a single line and should be straight forward.
Could you please check?
Thanks

iarsene (approved):

PR #5860, opened 2024-04-26
Add QC plots of event plane and event selection + add phi of cascade in output tree

ChiaraDeMartin95:


alibuild (approved):

PR #5866, opened 2024-04-26
[PWGEM/DQ] seperate calculation of phiV, add pairType as template for…

feisenhu:
… FillPairMC

Write a function to calculate phiV, in order to call it in different FillPair functions, like `FillPair`, `FillPairMC`, `FillPairME`.


alibuild (approved):

PR #5867, opened 2024-04-26
added 2nd filter bit to diffraction filter

pbuehler:
1st bit for DG events with less than 4 PV tracks
2nd bit for DG events with more than 3 PV tracks


pbuehler:
Hi @mpuccio this is the PR we discussed recently. The filter table of the diffraction filter has now 2 bits. The first bit is the selected events with up to 3 PV contributors and the 2nd bit is for the selected events with 4 and more PV contributors. For 2024 one can downscale the first bit as much as needed but keep the second bit unscaled. With this it should be possible to reach an overall reduction factor of 2x10⁴.

mpuccio (approved):

PR #5871, opened 2024-04-26
Add info about events used to compute event plane with FT0C (LF)

ChiaraDeMartin95:


alibuild (approved):

PR #5887, opened 2024-04-29
PWGJE : Adding V0 jets to the framework

nzardosh:
Changes in the JE and LF frameworks to add V0 jets
In the LF framework the changes are only cosmetic (changing table names and adding copy tables with different origins) - changes have been discussed with and approved by @ddobrigk


PWGLF/DataModel/LFStrangenessTables.h
```diff
@@ -165,6 +165,21 @@ DECLARE_SOA_TABLE(DauTrackMCIds, "AOD", "DAUTRACKMCID", // index table when usin
 using DauTrackExtras = DauTrackExtras_001;
 using DauTrackExtra = DauTrackExtras::iterator;
 
+DECLARE_SOA_TABLE(StoredDauTrackExtras, "AOD1", "DAUTRACKEXTRA", //! detector properties of decay daughters
```
njacazio, line 168: Is this used somewhere? I could not find it anywhere

PWGLF/DataModel/LFStrangenessTables.h
```diff
@@ -180,6 +195,13 @@ DECLARE_SOA_TABLE(MotherMCParts, "AOD", "MOTHERMCPART", //! mother MC informatio
 
 using MotherMCPart = MotherMCParts::iterator;
 
+DECLARE_SOA_TABLE(StoredMotherMCParts, "AOD1", "MOTHERMCPART", //! mother MC information, abbreviated name due to size limit
```
njacazio, line 198: Idem

PWGLF/DataModel/LFStrangenessTables.h
GijsvWeelden: The "stored" versions of tables are necessary for the JE derived data format.
Although `DauTrackExtras` is not necessary for my specific analysis, I will likely be of importance to a future analysis.

njacazio: You could move it to the JE part then?

GijsvWeelden: I'm not sure that would work. @nzardosh would know, I think.

nzardosh: Hi Nicolo, thanks for taking a look! At the moment it isn't used and we can remove it. But talking to David he said that this table is needed for analyses at later stages so we preemptively made it available. If you prefer though I'm happy to remove it for now. 

I would however prefer to keep them with the original table definitions since if you change the original table you need to correspondingly change these too, otherwise it won't compile (for the ones that are used). If it's in the JE framework it would make it difficult for the LF code writers to find it. 

Thanks,
Nima

PWGLF/DataModel/LFStrangenessTables.h
nzardosh: Same reply as above. These are generally needed when we want to make derived data. Currently on the first PR we haven't implemented the derived data part itself yet (this we will do next). But we wanted to make the changes needed already to the LF code so that we hopefully wouldn't need to bother you in future PRs.

nzardosh:
the tables have been removed

njacazio (approved):

PWGLF/DataModel/LFStrangenessTables.h
njacazio: Ok, I did not have the full plan in mind

PWGLF/DataModel/LFStrangenessTables.h
njacazio: idem

dsekihat (approved):

alibuild (approved):

PR #5895, opened 2024-04-29
[DPG-AOTt] Fix efficiency task, add DCA task

njacazio:


njacazio:
Tested and fixed

alibuild (approved):

DPG/Tasks/AOTTrack/qaEfficiency.cxx
```diff
@@ -1714,7 +1714,7 @@ struct QaEfficiency {
   SliceCache cache;
   Preslice<o2::aod::Tracks> perCollision = o2::aod::track::collisionId;
   Preslice<o2::aod::McParticles> perCollisionMc = o2::aod::mcparticle::mcCollisionId;
-  void processMC(o2::aod::McCollision const&,
+  void processMC(o2::aod::McCollisions const&,
```
mfaggin, line 1717: big question: in this way, I am afraid that you loop over `collisions` for every `McCollision`, because reconstructed collisions are not grouped with MC collisions, am I right?

mfaggin: @njacazio

njacazio: I think that's the intended way, I loop over all collisions in the DF

mfaggin: Still not clear to me. If you have 1000 reco collisions, and each of them correspond to a reco collision, then w/o the grouping you have 1000 /*MC collisions*/ TIMES 1000 /*reco. collisions*/, because for each MC collision you loop over the full reco. coll. table of the DF. Am I right?

mfaggin: this grouping is done thanks to the index in the `McCollisionLabel`, sorry @njacazio for the noise. Fine for me then.

mfaggin (approved):

DPG/Tasks/AOTTrack/qaEfficiency.cxx
njacazio: np

mfaggin (approved):

mfaggin (commented):
ah no sorry @njacazio another big doubt: aren't you filling twice the table of generated particles in this way? I mean: this is done once here https://github.com/AliceO2Group/O2Physics/blob/2a6a4798f32f4d0b0e35839b83b05a98c03e42b3/DPG/Tasks/AOTTrack/qaEfficiency.cxx#L1775-L1792

and then immediately later re-looping over all MC particles... but this introduces double counting, right?

njacazio:
> ah no sorry @njacazio another big doubt: aren't you filling twice the table of generated particles in this way? I mean: this is done once here
> 
> https://github.com/AliceO2Group/O2Physics/blob/2a6a4798f32f4d0b0e35839b83b05a98c03e42b3/DPG/Tasks/AOTTrack/qaEfficiency.cxx#L1775-L1792
> 
> and then immediately later re-looping over all MC particles... but this introduces double counting, right?

Hi @mfaggin thanks for checking. Indeed I loop twice on the particles, in reco events and not reco events.

I fill histograms with 
```
  template <int pdgSign, o2::track::PID::ID id, bool recoEv = false>
  void fillMCParticleHistograms(const o2::aod::McParticles::iterator& mcParticle, const bool doMakeHistograms)
```
The first call (reco event) is with e.g.
`fillMCParticleHistograms<pdgSign, o2::track::PID::Muon, true>(particle, doMu);` where `true` stands for the is `recoEv`.
The second time (all events) it is with
`fillMCParticleHistograms<pdgSign, o2::track::PID::Muon>(mcParticle, doMu);`

Selections are the same but if `recoEv` is true only the histograms of interest are concerned, e.g.:
```
    if constexpr (recoEv) {
      h->fill(HIST(hPtGeneratedRecoEv[histogramIndex]), mcParticle.pt());
      if (isPhysicalPrimary(mcParticle)) {
        h->fill(HIST(hPtGeneratedPrmRecoEv[histogramIndex]), mcParticle.pt());
      }
      return;
    }
```

alibuild (approved):

PR #5912, opened 2024-04-30
DPG: fix collision grouping with MC coll. and acceptance gen. particles.

mfaggin:
@njacazio in this PR I introduce a couple of fixes w.r.t. your last modifications for `processMC`:

- the `process` function must have a `MCCollision` (w/o `s`) if we want the reconstructed collisions to be grouped with generated ones. In this way, `mcParticles` are already grouped 
- when counting the generated particles in reconstructed collisions, you were missing the acceptance criteria, which were introducing significant differences between the two efficiencies


alibuild (approved):

njacazio (approved):
Thanks Mattia :)

PR #5918, opened 2024-04-30
PWGLF: Apply additional vertex z cut on MC case

BongHwi:
z vertex cut is applied in the resonance initializer but in the MC case, we don't apply due to find the vertex selection efficiency.


alibuild (approved):

alibuild (approved):

PR #5928, opened 2024-04-30
PWGLF: Add two more flags in resonance MC event histogram

BongHwi:
Add two more flags: `kTrig10`, `kSel10`


alibuild (approved):

PR #5930, opened 2024-05-01
UPC Resonance analysis

sdudi123:


rolavick (changes requested):
Dear Sandeep,
for better readability, please check your code it fully follows the o2 naming conventions https://rawgit.com/AliceO2Group/CodingGuidelines/master/naming_formatting.html in particular, the name of the file should be sgResonances.cxx
In addition, would be great if you mark your pesonal task with some marking, i.e. your initials, so the general names are reserved for general purposes. That means the name of the file could be sgResonancesSD.cxx. The same applies to task name in CMakeList -> inclusive-resonances-sd
Also, add to the file info on author and when the development start in the following format
// \author Sasha Bylinkin, alexander.bylinkin@gmail.com
// \since  April 2023

rolavick (approved):

rolavick (approved):

PR #5941, opened 2024-05-02
Preparation for ONNXRuntime bump. Do not merge

pzhristov:
These changes should be merged together with the bump of ONNXRuntime and flatbuffers in alidist.


ktf:
Can't we merge them independently, actually? It seems to me you properly have `#ifdef` for all the changes.

fcatalan92:
Hi @pzhristov, thanks a lot for taking care of bumping ONNXRuntime! This is indeed needed, and I and @fmazzasc received some requests in the past months too. However, I would do it after the end of the SQM approval campaign (~2 weeks from now) to avoid possible issues for the analyzers that uses ML models and are rushing for the preliminaries.
In any case, I think this PR can go ahead.

pzhristov:
> Can't we merge them independently, actually? It seems to me you properly have `#ifdef` for all the changes.

This is the idea, that the changes are exposed only after we bump ONNXRuntime and flatbuffers in https://github.com/alisw/alidist/pull/5452, however I wanted to be sure that the modification in dependencies/ worked.

ktf:
I would also prefer we merge this (given all the code is actually ifdeffed) and we simply avoid bumping ONNX. Doing so will allow us to continue with the GCC 13.0 migration in the associated PR and be ready to switch right after the SQM deadline...

fcatalan92 (approved):

fcatalan92:
Hi @pzhristov @ktf, sure, it is fine for me to merge this PR

ChSonnabend:
Just tested the installation with O2Physics with a dummy model using the workflow: `o2-analysistutorial-apply-onnx-model --modelPath ./test_torch.onnx`. It works. Since this uses the central ONNX header (like the TPC PID workflows), everything else should work the same way.

PR #5946, opened 2024-05-02
[PID] Split tables for tiny and full tune on data

njacazio:


njacazio:
@jezwilkinson we realized that now we cannot run anymore the full and tiny tables in the same workflow. This should fix it but it's not perfect. A solution would be to start merging the two workflow into one.
We can decide to rename the full table instead
What do you think?

jezwilkinson:
@njacazio This seems like a reasonable workaround to me (just for info what does the "Mark<1>" in the "tiny" table do?)
And yes, I think next goal will be to merge the executables to clean this situation up.

njacazio:
It allows to have the same table twice, I think otherwise the metadata will be the same and the compiler will complain that the table is defined twice

Common/DataModel/PIDResponse.h
```diff
@@ -611,6 +611,7 @@ DECLARE_SOA_COLUMN(DeDxTunedMc, mcTunedTPCSignal, float); //! TPC signal after T
 } // namespace mcpidtpc
 
 DECLARE_SOA_TABLE(mcTPCTuneOnData, "AOD", "MCTPCTUNEONDATA", mcpidtpc::DeDxTunedMc);
+DECLARE_SOA_TABLE(mcTPCTuneOnDataTiny, "AOD", "MCTPCTUNEONDATATINY", mcpidtpc::DeDxTunedMc, o2::soa::Mark<1>);
```
jgrosseo, line 614:
```suggestion
DECLARE_SOA_TABLE(mcTPCTuneOnDataTiny, "AOD", "MCTPCTUNEONDATATINY", mcpidtpc::DeDxTunedMc, o2::soa::Marker<1>);
```

jgrosseo (approved):

jgrosseo:
> @njacazio This seems like a reasonable workaround to me (just for info what does the "Mark<1>" in the "tiny" table do?) And yes, I think next goal will be to merge the executables to clean this situation up.

And it is called `Marker` :-D

njacazio:
Thanks @jgrosseo !

ddobrigk (approved):

PR #5948, opened 2024-05-02
[PWGDQ] New analysis task for asymmetric barrel pairs. Implement reduced ZDC table.

sigurdnese:
Implement reduced ZDC table, which is filled in table-maker-with-assoc.
New analysis task for asymmetric barrel pairs, e.g. D0->K-pi+.


PWGDQ/TableProducer/tableMaker_withAssoc.cxx
```diff
@@ -478,6 +490,29 @@ struct TableMaker {
       VarManager::fgValues[VarManager::kTimestamp] = bc.timestamp();
       VarManager::fgValues[VarManager::kRunIndex] = VarManager::GetRunIndex(bc.runNumber());
       VarManager::FillEvent<TEventFillMap>(collision); // extract event information and place it in the fValues array
+      if constexpr ((TEventFillMap & VarManager::ObjTypes::Zdc) > 0) {
```
sigurdnese, line 493: Open to cosmetic changes and suggestions regarding this block. Maybe there is somewhere better to put this?

iarsene: I think its fine like this

iarsene (approved):

PWGDQ/Tasks/tableReader_withAssoc.cxx
```diff
@@ -1438,6 +1453,322 @@ struct AnalysisSameEventPairing {
...
+                } else {
+                  fHistMan->FillHistClass(histNames[icut][2].Data(), VarManager::fgValues);
+                  if (isAmbiInBunch) {
+                    fHistMan->FillHistClass(histNames[icut][5].Data(), VarManager::fgValues);
+                  }
+                  if (isAmbiOutOfBunch) {
+                    fHistMan->FillHistClass(histNames[icut][8].Data(), VarManager::fgValues);
+                  }
+                }
+              }
```
iarsene, line 1734: The split of the histogramming for in bunch and out of bunch ambiguity could be useful, but if you didn't wanted it for now, you don't have to add it. In my case, for the same-event pairing, i had to understand a bit some of the effects on jpsi. But anyway, i think it might be useful for you as well. So maybe leave it as it is.

iarsene (approved):

iarsene:
This PR is modifying the reduced data model file, by adding a table with ZDC information. We agreed today at the O2 DQ to have discussions in the meeting before touching core aspects of the framework, such as core workflows or the data model. However, because there is a time component for this case, and because it just adds a new table (not modifying an old one), i approved and merged this PR.

PR #5960, opened 2024-05-02
polarization studies in the kstar and phi resonance

sawankumawat:


skundu692:
Avoid to copy a code. It does not make any sense. A flow code for kstar is already exist if you want to add polarization then add it there.

skundu692 (changes requested):
There is already a code exist for kstar flow. Would it be possible to add this observables there in order to avoid the repetition of large ammount of existing code. If you want we could discuss in resonance PAG

dmallick2:
yes the existing phi code can be used to calculate polarization and flow using event plane. This code for polarization studies  for different axes ( helicity frame, production plane, beam and random axes ). It is possible to make a single macro. To prepare a different macro for such polarization studies where event plane calculations are not needed.

skundu692:
@dmallick2 would it be possible to add your observable in the existing task and fill additional histogram. If you want we could also discuss this in resonance PAG.

dmallick2:
@skundu692. okay. we will discuss in PAG meeting.

PR #5968, opened 2024-05-03
PWGHF: Replacing partition with filter

Simple-Shyam:
I replaced the partition with the filter..


fcolamar (approved):
This is ok to me, after having discussed offline with Shyam about a couple of modifications he made

PR #5970, opened 2024-05-03
Avoid unintialised variables

ktf:


ktf:
This is breaking the daily build. Not sure how come it was not detected earlier, but it's obviously wrong, depending on the cut.

PR #5981, opened 2024-05-03
PWGLF: Add hasITS, hasTPC information

BongHwi:
This will be used to compare the ITSonly track and ITS-TPC track and so on...


alibuild (approved):

PR #5982, opened 2024-05-03
PWGHF: fix index for grouped table in charm polarisation task

fgrosa:


mfaggin (approved):
good catch @fgrosa , thank you

alibuild (approved):

PR #5991, opened 2024-05-04
PWGLF: add found tag in findable exercise

ddobrigk:
@lhusova @romainschotter with this PR, we can generate an extra flag in the findable derived data that is able to tell us easily if each and every "findable" candidate was actually found correctly. The ratio of found/findable is therefore viable within a single derived dataset with all findable, adding extra debug capabilities as we can more easily and reliably calculate the combined global tracking + svertexing efficiency... 

I will still test this before declaring it ready to merge. The changes necessary for this were a bit more complicated than I originally expected...

### DIFF
--- a/PWGLF/DataModel/LFStrangenessTables.h
+++ b/PWGLF/DataModel/LFStrangenessTables.h
@@ -703,12 +703,12 @@ DECLARE_SOA_TABLE(V0MCRefs, "AOD", "V0MCREF", //! index table when using AO2Ds
 using V0sLinked = soa::Join<V0s, V0DataLink>;
 using V0Linked = V0sLinked::iterator;
 
-namespace v0data 
+namespace v0data
 {
 DECLARE_SOA_COLUMN(IsFound, isFound, bool); //! is this FindableV0 actually in the V0s table?
 }
 
-// Major bypass for simultaneous found vs findable study 
+// Major bypass for simultaneous found vs findable study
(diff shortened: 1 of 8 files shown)


ddobrigk:
Tested to be OK and does not affect normal operation (with found candidates).

chiarapinto (dismissed):

chiarapinto:
Hi @ddobrigk , Megalinter is complaining, can you have a look?

ddobrigk:
@romainschotter @lhusova in the latest commit just added to this PR, I have included the ability to have "asymmetric filling" of the `V0MCCores` table. This because, at the moment, `V0Cores` and `V0MCCores` are filled essentially together in Monte Carlo ('symmetrically'), while in fact to unambiguously and conveniently deal with duplicates we'd have to improve that mechanism to account for unique (generated mcparticle pairs) entries of `V0MCCores` that are linked to non-unique (reconstructed)`V0Cores`  via a new index table. This is now done and the result is: 

image

This mechanism will work both in found and findable monte carlo data, if enabled, but it will crucially break the joinability of V0Cores and V0MCCores. We can handle that either creating a separate task to deal with duplicate studies (easier and far more convenient - in fact, my preferred solution, since it keeps things clean) or making adjustments to the existing `derivedlambdakzeroanalysis` task. We can discuss next week...

ddobrigk:
@chiarapinto this is now ready to be merged and megalinter is also happy - please take a look whenever convenient. Many thanks!!

chiarapinto (approved):

PR #6000, opened 2024-05-06
(EP Dev): Q-vector check added

MaximVirta:


fgrosa (changes requested):
Hi @MaximVirta thanks a lot for the development!
Thee part that you developed is correct, but what is still missing is the filling of the tables conditional to the usage or not of a specific detector. At the moment all the Q-vector tables are filled in any case, while they should not be filled if `useDetector` for a given detector is `false` (lines `498` to `504` in the new file). Also, all the calibration steps are always applied while again it makes no sense to apply them if the detector is not used (lines `478` to `494`).

fgrosa (approved):
It looks good to me now, thanks!

ddobrigk (approved):

PR #6003, opened 2024-05-06
Updating qaHMPID task

gvolpe79:


njacazio:
Ciao Giacomo, Megalinter is complaining, could you have a look?

gvolpe79:
Fixed!

njacazio:
Super, thx!

DPG/Tasks/AOTTrack/PID/qaHMPID.cxx
```diff
@@ -30,26 +32,71 @@ struct pidHmpidQa {
   Configurable<float> maxDistance{"maxDistance", 5.f, "Maximum HMPID distance between the track and the cluster (cm)"};
   Configurable<float> minCharge{"minCharge", 120.f, "Minimum HMPID charge collected in the cluster"};
 
+  OutputObj<TTree> pTree{"pTree", OutputObjHandlingPolicy::AnalysisObject};
```
njacazio, line 35: Are you sure that you want to save a TTree? In this case I would suggest to produce tables instead, such that we can do the post analysis within the O2. @mfaggin what do you think?

mfaggin: I totally agree, better to go with a table.

njacazio: Perfect, @gvolpe79 let us know if you need some example to follow, the sintax is similar to the one of the tree but there's a better management (including e.g. merging) on the hyperloop.

gvolpe79: I removed the TTeee, could you merge like it is now? Then I will add the table!

njacazio (approved):
Thanks Giacomo

PR #6005, opened 2024-05-06
PID: Refactor TPC/TPCFull table producers to single task

jezwilkinson:
- Refactor the TPC PID table producers to run from a single task (removing pid-tpc and keeping pid-tpc-full)
- Simplified process loops and restructured table creation for standard/MC TuneOnData modes so that all table production is done in a single shared function
- Corrected filling of MC TuneOnData signal table to only fill when requested by another process


jezwilkinson:
@njacazio @wiechula @ChSonnabend FYI
Just a draft currently (I have tested the compilation but still need to run it through and make sure it doesn't suddenly do anything unexpected with the output values - will open it for merging once those checks are done) - this is the merge for the TPC PID tasks that we discussed a few weeks back, should be a significant cleanup compared to before and allows both of the tables to be produced at the same time by a single wagon.

jezwilkinson:
Marking as ready for review, the test cases were consistent between the latest tag and this commit.
@njacazio Just to clarify the effects of this for the PID wagons on hyperloop once this is merged:
- the wagons based on "o2-analysis-pid-tpc" (the old 'tiny' producer) will be obsolete as the executable is removed
- "o2-analysis-pid-tpc-**full**" will produce both the tiny, full, and the TuneOnData table, automatically as needed by other tasks in the workflow. Some of the configurable names change (e.g. pid-el becomes pid-full-el and pid-tiny-el) but as these are all anyway defaulted to -1 it shouldn't need any change in the core wagon configurations.
- all wagons that depend on "o2-analysis-pid-tpc" will need to update their dependencies to "o2-analysis-pid-tpc-full" (will need to be announced for users to do this as I guess there is no way to do this centrally). For people already using "o2-analysis-pid-tpc-full", no special action is needed.
I leave it up to you how you prefer to manage this with the core wagon definitions

jgrosseo:
Has this been tested also on Run 2 cases?

For the core service wagons, this has to be closely synchronized. Whatever change we do it will not work for old tags anymore where people run on. Or we make new service wagons. Adding also @ddobrigk 

Will TPCBase also disappear in the dependency chain?

jgrosseo:
And of course thanks a lot for your work, @jezwilkinson ! :-)

jezwilkinson:
> Has this been tested also on Run 2 cases?

Not explicitly yet, but the TPC task doesn't have any kind of different handling for run2 vs run3 data so I can't imagine this would behave any differently. I can try a quick extra test on a couple of run2 ao2ds in case.

> For the core service wagons, this has to be closely synchronized. Whatever change we do it will not work for old tags anymore where people run on. Or we make new service wagons. Adding also @ddobrigk
Agreed, I guess there would need to at least temporarily be some "legacy" wagon for use on older tags unless there's some way of treating one as an alias of the other in the newer version.
For what it's worth, this update should pretty much halve the CPU overhead for the TPC PID on trains that currently include both pidTPC and pidTPCFull since it only needs to go through the track loop once and perform the NN application once instead of twice.

> Will TPCBase also disappear in the dependency chain?
Currently no, to be discussed with @njacazio whether this should also be merged but for now it is left separate.

jgrosseo:
I had a look and would propose the following:

- We wait for SQM approvals to finish to avoid mental stress of @ddobrigk :)
- Currently we have PIDTPC and PIDTPCFull service wagons. If we would simply keep the former executable and not the `full` one (so opposite to the current PR), then a) everyone who currently uses the PIDTPC does not have to do anything, and b) people who currently use PIDTPCFull have to change if they want to use the current tag (and keep PIDTPCFull if they use an old tag).

What do you think? Did I get it right?

jezwilkinson:
Hi @jgrosseo ,
since SQM is now over can this be considered for merging again? There are new developments being planned for the task to (among other things) update the occupancy estimator in the NNs and possibly some special handling for sector A11 in the more recent data, and it would help a lot to have this cleanup merged before those get started.
Regarding the Hyperloop service wagons : Yes, the immediate consequence of this is that user dependencies will need to be updated to use whichever of the executables is kept (whether this is done as "pid-tpc" or "pid-tpc-full" doesn't really matter too much, similar adaptation will need to be done either way and no real preference as to which name is kept).

Cheers,
--Jeremy

jgrosseo:
> Hi @jgrosseo , since SQM is now over can this be considered for merging again? There are new developments being planned for the task to (among other things) update the occupancy estimator in the NNs and possibly some special handling for sector A11 in the more recent data, and it would help a lot to have this cleanup merged before those get started. Regarding the Hyperloop service wagons : Yes, the immediate consequence of this is that user dependencies will need to be updated to use whichever of the executables is kept (whether this is done as "pid-tpc" or "pid-tpc-full" doesn't really matter too much, similar adaptation will need to be done either way and no real preference as to which name is kept).
> 
> Cheers, --Jeremy

I am happy to go ahead.

We will keep only one executable in O2Physics in the long-term, so let's then use the `pid-tpc` as it does not make sense to have  `full` without something else. Could you make this change?

> > Has this been tested also on Run 2 cases?
> 
> Not explicitly yet, but the TPC task doesn't have any kind of different handling for run2 vs run3 data so I can't imagine this would behave any differently. I can try a quick extra test on a couple of run2 ao2ds in case.

Did this test succeed? If not done, one could use the recipe and files here:
https://alimonitor.cern.ch/train-workdir/tests/0022/00224550/stdout.log

Also @ddobrigk should give his opinion.

jgrosseo:
For Hyperloop, the best may be 
* to rename the existing wagons with an `Old` suffix to still run on old tags (to be kept for a month maybe)
* create new TPCPID wagons with the new configuration once this PR has been merged
The list of PIDTPC wagons is rather long at present 
image

jezwilkinson:
> 
> We will keep only one executable in O2Physics in the long-term, so let's then use the `pid-tpc` as it does not make sense to have `full` without something else. Could you make this change?

Done (https://github.com/AliceO2Group/O2Physics/pull/6005/commits/3295ecdda160a9b925917e2796d5541e852471a1)

> 
> > > Has this been tested also on Run 2 cases?
> > 
> > 
> > Not explicitly yet, but the TPC task doesn't have any kind of different handling for run2 vs run3 data so I can't imagine this would behave any differently. I can try a quick extra test on a couple of run2 ao2ds in case.
> 
> Did this test succeed? If not done, one could use the recipe and files here: https://alimonitor.cern.ch/train-workdir/tests/0022/00224550/stdout.log

Test is done now, the behaviour is the same for Run2 as it is for run3.

jgrosseo:
Thanks a lot. From my side this is fine code-wise. We need feedback from David still. Then we need to plan when to do this as it requires quite some wagon reshuffling...

ddobrigk:
Thanks everyone, this is definitely a good idea. I think we should also start to take advantage of metadata, since that will greatly simplify the hyperloop dependencies. @njacazio I will then also mark this as approved, can you please help us with dealing with the core service wagon work? Many many thanks!

ddobrigk:
> Thanks a lot. From my side this is fine code-wise. We need feedback from David still. Then we need to plan when to do this as it requires quite some wagon reshuffling...

As to the when, I am not sure but perhaps Monday next week is actually better than now? We may run into plenty of maintenance problems and in that sense it's best done on a Monday. Any opinions?

jgrosseo:
> > Thanks a lot. From my side this is fine code-wise. We need feedback from David still. Then we need to plan when to do this as it requires quite some wagon reshuffling...
> 
> As to the when, I am not sure but perhaps Monday next week is actually better than now? We may run into plenty of maintenance problems and in that sense it's best done on a Monday. Any opinions?

Just that you @ddobrigk and me are both away on Monday at the same workshop ;-)

jgrosseo:
@jezwilkinson: @ddobrigk and me had a chat now.
We propose to merge this next Wednesday, such that we can do all the adjustments to wagons on Thursday morning where both of us are back to the offices. What do you think?

For the concrete plan:
- Rename all existing TPCPID service wagons (adding `deprecated`). Two have this tag already (apass3)
- Recreate the needed ones. Is the following list complete? PIDTPC, _MC, _Run2, _withTPConly, NN, MC_TuneOnData, NN_withTPConly
- Change dependencies in our integration test wagons and see if it all works
- Announce

Let us know your comments.

jezwilkinson:
Hi @jgrosseo ,
I think the wagon cases would be:

TPC ; TPC_NN ; TPC_MCTuneOnData ; TPC_MCTuneOnData_NN ; and then those with "_withTPCOnly" for analyses that need PID on TPC-only tracks.

Unless there's something specific in the dependency chain, I don't think it needs a separate set of Run2 wagons (both Run2/Run3 use the same process loop), and a special MC configuration is only required if TuneOnData is used (if ToD is disabled for whatever reason then, again, it only needs processStandard).

jgrosseo (approved):

jgrosseo:
Then tomorrow morning, we will adjust the service wagons. This may cause intermittent failures in the nightly trains.

jezwilkinson:
@jgrosseo Thanks!

So then do you prefer to make the changes to the core wagons before we announce, or should i send something out already today?

jgrosseo:
@jezwilkinson Sorry for my silence, I was travelling today. I propose that I do the renaming and creating the new ones (I guess copying the settings from before should be more or less clear clear). I could start around 8:30. Do you think you'll be online? You can then do the announcement when all works...

PR #6007, opened 2024-05-06
PWGHF: Add charm-hadron channels to taskMcValidation + workflow improvements

fcatalan92:
- Add generated-level check for charm baryons and Ds->K0*K->KKpi
- Split reconstructed validation part into multiple process functions to allow running with less dependencies (default is HF index-skim-creator only)

@cterrevo @ZFederica @zhangbiao-phy @DelloStritto @stefanopolitano FYI


fcatalan92 (commented):
Merging to catch the next tag

PR #6017, opened 2024-05-06
PWGHF: add D meson beauty mother ptB in THnSparse, add function for getting ptB in HfHelper

Mingyu3360715:


Mingyu3360715:
> It looks like a copy-paste of a `RecoDecay` function. Why is it needed?

Hi Vit, the pt weight  need to be considered in the D0 analysis. A method returning the beauty mother pt (ptB) could be convenient when filling the THnSparse, like what was done in Run 2. Do you have any suggestion on it, or is there an easy way to get the ptB?

vkucera:
> > It looks like a copy-paste of a `RecoDecay` function. Why is it needed?
> 
> Hi Vit, the pt weight need to be considered in the D0 analysis. A method returning the beauty mother pt (ptB) could be convenient when filling the THnSparse, like what was done in Run 2. Do you have any suggestion on it, or is there an easy way to get the ptB?

It seems that you use the same logic as `RecoDecay::getCharmHadronOrigin` to find the beauty mother. Would it make sense to extend the function to store the index of the mother particle and then get its pT?

fcatalan92 (commented):
Hi @vkucera, I agree with you, the functionality of `RecoDecay::getCharmHadronOrigin` can be extended to return the index of the found mother. However, I guess Mingyu's developments are needed for a check related to his analysis targeting SQM preliminaries. So, if he does not have time to do the changes, I would propose to merge the code as it is (we can open an issue to keep track of the proposed improvement).

fcatalan92 (approved):

Mingyu3360715:
> Hi @vkucera, I agree with you, the functionality of `RecoDecay::getCharmHadronOrigin` can be extended to return the index of the found mother. However, I guess Mingyu's developments are needed for a check related to his analysis targeting SQM preliminaries. So, if he does not have time to do the changes, I would propose to merge the code as it is (we can open an issue to keep track of the proposed improvement).

@fcatalan92 @vkucera Thanks a lot for the suggestion. And thanks for considering the need for our ongoing analysis. I will address the proposal right after the preliminary period. Also reminding @atavirag for this PR.

vkucera:
> Hi @vkucera, I agree with you, the functionality of `RecoDecay::getCharmHadronOrigin` can be extended to return the index of the found mother. However, I guess Mingyu's developments are needed for a check related to his analysis targeting SQM preliminaries. So, if he does not have time to do the changes, I would propose to merge the code as it is (we can open an issue to keep track of the proposed improvement).

@fcatalan92 Fine with me.

vkucera (approved):

PR #6026, opened 2024-05-07
Use TVX instead of sel8 (the other bits are manually checked)

mpuccio:


alibuild (approved):

PR #6031, opened 2024-05-07
PWGLF k892analysis: adding some extra plots, cuts to exclude overflows, cut on centrality

murioni:


BongHwi:
Hello @murioni , thanks for the update!

The primary goal of introducing the resonance task is to manage event selection via the `ResoInitializer` while keeping the resonance reconstruction in a separate analysis task. 
Although I have to admit that we have some experimental features of the further event selection in the k892 analysis task, 
 perhaps we could integrate additional selections directly into the initializer?

Regarding the `cMaxCentr` selection, my suggestion would be to handle these events offline.

Looking forward to discussing this further!

murioni:
> Hello @murioni , thanks for the update!
> 
> The primary goal of introducing the resonance task is to manage event selection via the `ResoInitializer` while keeping the resonance reconstruction in a separate analysis task. Although I have to admit that we have some experimental features of the further event selection in the k892 analysis task, perhaps we could integrate additional selections directly into the initializer?
> 
> Regarding the `cMaxCentr` selection, my suggestion would be to handle these events offline.
> 
> Looking forward to discussing this further!

Hello @BongHwi , Thank you! I'll remove the `cMaxCentr` selection.

BongHwi (approved):

PR #6053, opened 2024-05-08
PWGHF: Fixed bugs in Resonance Analysis

Luca610:
dataCreatorCharmResoReduced: fixed Dstar mass calculation
candidateCreatorCharmResoReduced: fixed loop on candidates belonging to the same collision

### DIFF
--- a/PWGHF/D2H/TableProducer/candidateCreatorCharmResoReduced.cxx
+++ b/PWGHF/D2H/TableProducer/candidateCreatorCharmResoReduced.cxx
@@ -79,11 +79,15 @@ struct HfCandidateCreatorCharmResoReduced {
   Partition<aod::HfRedVzeros> candidatesK0s = aod::hf_reso_cand_reduced::v0Type == (uint8_t)1 || aod::hf_reso_cand_reduced::v0Type == (uint8_t)3 || aod::hf_reso_cand_reduced::v0Type == (uint8_t)5;
   Partition<aod::HfRedVzeros> candidatesLambda = aod::hf_reso_cand_reduced::v0Type == (uint8_t)2 || aod::hf_reso_cand_reduced::v0Type == (uint8_t)4;
 
+  Preslice<aod::HfRedVzeros> candsV0PerCollision = aod::hf_track_index_reduced::hfRedCollisionId;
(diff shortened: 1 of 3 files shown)


PWGHF/D2H/TableProducer/candidateCreatorCharmResoReduced.cxx
```diff
@@ -130,7 +135,7 @@ struct HfCandidateCreatorCharmResoReduced {
     if (channel == DecayChannel::Ds2StarToDplusK0s || channel == DecayChannel::XcToDplusLambda || channel == DecayChannel::LambdaDminus) {
       massD = massDplus;
     } else if (channel == DecayChannel::Ds1ToDstarK0s) {
-      massD = massDstar;
+      massD = massDstar - massD0;
```
fgrosa, line 138: This is not correct because we compute the invariant mass of D*+ and K0S, having the deltaM in the formula would not make sense. To compute the invariant mass different as in Run2 (i.e. M(KpipiK0S) - M(Kpipi)) one should use all the daughters, but we do not store them. Since we verified that we get more or less the same resolution simply by using the PDG mass also for the D in the decay, it was correct before:
```suggestion
      massD = massDstar;
```

fgrosa (changes requested):
Hi @Luca610 thanks! I just have one comment, then it is ready to be merged.

PWGHF/D2H/TableProducer/candidateCreatorCharmResoReduced.cxx
Luca610: Hi Fabrizio, I agree with you. However in the context of the function `isDSelected`, i am initializing the mass of the D* as a difference in order to apply a cut on the invariant mass of the Dstar daughter only, and currently I am saving the mass difference for the D*. In the computation of the resonance invariant mass (line 221) i am using the PDG masses

fgrosa: Ah good! Sorry, I misunderstood it. If you are anyway computing the invariant mass using the D* mass and not the delta it's fine. Thanks for the explanation!

fgrosa (approved):

PR #6059, opened 2024-05-08
[event plane] q-vector calculation for different harmonics in one task

jikim1290:
q-vector calculation for different harmonics in one task


Common/TableProducer/qVectorsTable.cxx
```diff
@@ -444,79 +405,124 @@ struct qVectorsTable {
...
+      qvecReBNeg.push_back(qvecRe[(kBNeg + 1) * 4 * id + kBNeg * 4 + 3]);
+      qvecImBNeg.push_back(qvecIm[(kBNeg + 1) * 4 * id + kBNeg * 4 + 3]);
     }
 
-    // Fill the columns of the Qvectors table if they are found for a detector.
+    // Fill the columns of the Qvectors table.
     qVector(cent, IsCalibrated, qvecRe, qvecIm, qvecAmp);
     if (useDetector["QvectorFT0Cs"])
-      qVectorFT0C(IsCalibrated, qvecRe[kFT0C * 4 + 3], qvecIm[kFT0C * 4 + 3], sumAmplFT0C);
+      qVectorFT0C(IsCalibrated, qvecReFT0C, qvecImFT0C, qvecAmp[kFT0C]);
```
strogolo, line 515: Dear @jikim1290,
I looked carefully at your PR and this is not the change we discussed offline.
I'm taking as an example the qVectorFT0C table. Here you are replacing qvecRe[kFT0C *4 +3], which is the Q-vector at the last step of correction, with qvecReFT0C, which is the array with Q vectors at each correction step.
However, the proposal was simply to have the possibility to configure the correction step that an analyser can use. 
So our proposal was to replace  `qvecRe[kFT0C * 4 + 3], qvecIm[kFT0C * 4 + 3]` with `qvecRe[kFT0C * 4 + cfgCorrStep], qvecIm[kFT0C * 4 + cfgCorrStep]`. In this way, if an analyser set cfgCorrStep == 3 gets the fully-corrected Q vector, while if cfgCorrStep == 1 gets the Q vector with only Recentering correction applied.

Is there a reason for the current PR where you are storing everything in the detector-specific table of Q vectors?

Common/TableProducer/qVectorsTable.cxx
```diff
@@ -77,7 +77,8 @@ struct qVectorsTable {
 
   Configurable<float> cfgMinPtOnTPC{"cfgMinPtOnTPC", 0.15, "minimum transverse momentum selection for TPC tracks participating in Q-vector reconstruction"};
   Configurable<float> cfgMaxPtOnTPC{"cfgMaxPtOnTPC", 5., "maximum transverse momentum selection for TPC tracks participating in Q-vector reconstruction"};
-  Configurable<int> cfgnMod{"cfgnMod", 2, "Modulation of interest"};
+//  Configurable<int> cfgnMod{"cfgnMod", 2, "Modulation of interest"};
+  Configurable<std::vector<int>> cfgnMods{"cfgnMods", {2, 3}, "Modulation of interest"};
```
strogolo, line 81: Why are you adding here more harmonics in one task? This makes the combination of this task with the analyses task much heavy when running. We agreed at AOT of having the task running one harmonic at a time.

jikim1290: Hi, this is update for request from other analyzers who want to see v2 and v3 at the same event for measuring v2/v3.
I was updating the code before your request.
For the step problem, you can always access qvectors at each step when using the table of qvecRe and qvecIm, which need to be stored for validation purpose.
You can already get recentered FT0C qvector with qvecRe[kFT0C*4+1].
I don't think most of users will not use incompletely corrected qvectors, and storing full corrected one would be enough for "QvecFT0CRe" table in terms of memory.
The user can run only one time if cfgnMods is set {2} instead {2, 3}, which calculates psi2 only.

strogolo: Hi @jikim1290,
I see your point but this is not the right way to proceed because this is a centralized task and changes should be discussed at least with the coordinators...even better would be a presentation at the AOT meeting. This task is used by other analysers involved in flow analyses and a change from your side to satisfy someone may cause problems to other analysers (especially in the approval period.
Alternatively, the analyzer who requested this change may contact the AOT coordinators directly and propose to us the change. I hope you understand the centrality of this task and the reason why I'm not in favor to this change.

Then, as we already discussed, it would be much lighter to have a configurable to set the correction step rather than using the entire qvector table. Last, but not least, it's not completely true when you say that most of users will not use incompletely corrected qvectors, because in the past we saw that it was even enough using just first or second level of correction.

jikim1290: Hi @strogolo, Yes you are correct. I need to discuss things in AOT firstly. I was discussing this inside group..
I close this and will open this again with the proper collaborative step. thanks!

PR #6067, opened 2024-05-08
PWGHF: add table in process func mc toXiPi

ZFederica:
Add McCollisions table in process function mc arguments


fgrosa (approved):
Thanks @ZFederica!

PR #6068, opened 2024-05-08
PWGJE: Adding WTA DeltaR calculation to hJet task.

dajones2:


nzardosh (dismissed):

PWGJE/Tasks/jetHadronRecoil.cxx
```diff
@@ -166,6 +170,13 @@ struct hJetAnalysis {
       registry.fill(HIST("hJetPt"), jet.pt());
       registry.fill(HIST("hJetEta"), jet.eta());
       registry.fill(HIST("hJetPhi"), jet.phi());
+      for (auto& jetWTA : jet.template matchedJetGeo_as<std::decay_t<W>>()) {
+        double deltaPhi = jetWTA.phi() - jet.phi();
```
vkucera, line 174: You should constrain the angle to be within +-Pi. Otherwise you'll get nonsensical values of `dR`.
See
https://github.com/AliceO2Group/O2Physics/blob/4ecf524826c7279cb172e35ac2b390f9696dd84c/Common/Core/RecoDecay.h#L214-L220

PWGJE/Tasks/jetHadronRecoil.cxx
```diff
@@ -166,6 +170,13 @@ struct hJetAnalysis {
...
+        double deltaEta = jetWTA.eta() - jet.eta();
+        double dR = sqrt(pow(deltaPhi, 2) + pow(deltaEta, 2));
```
vkucera, line 176:
```suggestion
        double dR = RecoDecay::sqrtSumOfSquares(deltaPhi, deltaEta);
```
https://github.com/AliceO2Group/O2Physics/blob/4ecf524826c7279cb172e35ac2b390f9696dd84c/Common/Core/RecoDecay.h#L84-L88

vkucera (changes requested):

PWGJE/Tasks/jetHadronRecoil.cxx
dajones2: I am not sure why this would ever be a problem, as I am only comparing geometrically matched jets. I do have a detected deltaPhi function that can be used if needed.

dajones2: Should be working now

dajones2: I need the second dPhi function for the delta phi observable which I want to be constrained between 0 and 2pi. I can change to use this constrainAngle function now though

dajones2: Should now be implemented

PWGJE/Tasks/jetHadronRecoil.cxx
```diff
@@ -166,8 +161,15 @@ struct hJetAnalysis {
       registry.fill(HIST("hJetPt"), jet.pt());
       registry.fill(HIST("hJetEta"), jet.eta());
       registry.fill(HIST("hJetPhi"), jet.phi());
+      for (auto& jetWTA : jet.template matchedJetGeo_as<std::decay_t<W>>()) {
+        double deltaPhi = RecoDecay::constrainAngle(jetWTA.phi() - jet.phi(), -M_PI);
```
vkucera, line 165:
```suggestion
        double deltaPhi = RecoDecay::constrainAngle(jetWTA.phi() - jet.phi(), -o2::constants::math::PI);
```
Do not use `M_PI`. It is not a C++ standard and in O2Physics, we have it defined among other constants.

vkucera (approved):

nzardosh (approved):

PR #6095, opened 2024-05-10
[PWGEM] correct PDGCode for MC signal eeFromCharedD

feisenhu:


PWGDQ/Core/MCSignalLibrary.cxx
```diff
@@ -718,7 +718,7 @@ MCSignal* o2::aod::dqmcsignals::GetMCSignal(const char* name)
   }
   // D+/- -> e and D+/- -> e
   if (!nameStr.compare("eeFromChargedD")) {
-    MCProng prong(2, {11, 421}, {true, true}, {false, false}, {0, 0}, {0, 0}, {false, false}, false, {502}, {true});
+    MCProng prong(2, {11, 411}, {true, true}, {false, false}, {0, 0}, {0, 0}, {false, false}, false, {502}, {true});
```
vkucera, line 721:
```suggestion
    MCProng prong(2, {kElectron, kDPlus}, {true, true}, {false, false}, {0, 0}, {0, 0}, {false, false}, false, {502}, {true});
```
You will avoid typos if you use the names.

vkucera: You need to include the proper headers and use the proper names. See:
- https://github.com/AliceO2Group/AliceO2/blob/221c1f0fabeb7d92d3c894d00791e5b174f5c07d/Common/Constants/include/CommonConstants/PhysicsConstants.h#L33
- https://root.cern/doc/master/TPDGCode_8h.html

alibuild (approved):

PR #6110, opened 2024-05-11
Remove some unnecessary configurables

ChunzhengLab:
Remove some unnecessary configurables to ensure not exceeding the limits.
Locally tested, the [WARN] message about "no input" has disappeared.


DPG/Tasks/AOTTrack/qaMatchEff.cxx
```diff
@@ -1298,9 +1293,6 @@ struct qaMatchEff {
       return false;
     if (!cutObject.IsSelected(track, TrackSelection::TrackCuts::kDCAxy))
       return false;
-    // dcaZ selection to simulate the dca cut in QC ()
-    if (abs(track.dcaZ()) > dcaMaxCut->get("TrVtx", "dcaZ"))
-      return false;
```
mfaggin, line 1303: removing this you never apply the dcaZ cut. Actually, since you set this cut via the `cutObject` you can you it directly in this way:

```
if (!cutObject.IsSelected(track, TrackSelection::TrackCuts::kDCAxy))
      return false;
```

mfaggin (approved):
thank you @ChunzhengLab

PR #6111, opened 2024-05-12
PWGLF: findable <-> analysis connection

ddobrigk:
This PR changes the findable study such that it is possible to connect directly to the analysis level efficiencies for the full determination of the entire chain of losses (and possible optimization). Relevant for @lhusova @romainschotter @gianniliveraro. I expect to have a short summary of the first findings of this study very soon. Some specifics: 

@romainschotter : the bit-check machinery included in the `derivedlambdakzeroanalysis` has been modularized so that I can use that as a module from the findable task, with its different subscriptions. The main advantage is that I can do *exactly* the same selections within the findable exercise in a trivial way. However, I have not modified the `derivedlambdakzeroanalysis` yet - we can discuss how to proceed there. 

@gianniliveraro : this also makes the findable tool able to deal with other hypotheses (i.e. Lambda, conversions), following your very appropriate suggestion; however, the last "TPC PID" check has yet to be adapted for that case.


ddobrigk:
@gianniliveraro the fix for pair production process check for conversions is in place! This should give you plenty of gammas now with the findable ...

ddobrigk:
Now with all megalinter fixes included and ready to merge :-D

njacazio (approved):

PR #6122, opened 2024-05-13
Add PV plots/cuts

ilikmeta:


PWGCF/Flow/Tasks/FlowGFWPbPb.cxx
```diff
@@ -341,6 +359,33 @@ struct FlowGFWPbPb {
...
+      vtxz = collision.posZ();
+      float zRes = TMath::Sqrt(collision.covZZ());
+      if (zRes > 0.25 && collision.numContrib() < 20)
+        vtxz = -999;
+    }
+
+    auto multNTracksPV = collision.multNTracksPV();
+
+    if (centrality >= 70. || centrality < 0)
+      return 0;
```
victor-gonzalez, line 373: Why not returning `False` instead, it will be more readable

PWGCF/Flow/Tasks/FlowGFWPbPb.cxx
```diff
@@ -341,6 +359,33 @@ struct FlowGFWPbPb {
...
+    if (multTrk < fMultCutLow->Eval(centrality))
+      return 0;
+    if (multTrk > fMultCutHigh->Eval(centrality))
+      return 0;
+
+    // V0A T0A 5 sigma cut
+    if (abs(collision.multFV0A() - fT0AV0AMean->Eval(collision.multFT0A())) > 5 * fT0AV0ASigma->Eval(collision.multFT0A()))
+      return 0;
+
     return 1;
```
victor-gonzalez, line 389: and `True` in here
Readability is a much for further code understanding

victor-gonzalez (approved):

PWGCF/Flow/Tasks/FlowGFWPbPb.cxx
ilikmeta: You mean that to obtain collisions from 0 to 70 % centrality I should return 1 ?

PWGCF/Flow/Tasks/FlowGFWPbPb.cxx
ilikmeta: Here you mean to have return 0 insted of return 1 in line 389?

victor-gonzalez:
I'm talking about readability of the code
Instead of using `0` use `false` and instead of using `1` use `true`

ilikmeta:
Got it! I will so ASAP. Thanks!

You want me to change `o` to `false` anywhre else in the code or only for those you just mentioned?

PR #6123, opened 2024-05-13
PWGJE: Add MCcollision check if it has its correspond collision  while counting MCP jets

joonsukbae:
Many comments at JP and JE-O2 meeting pointed out the issue of jet reconstruction efficiency that it's lower than Run 2 result. But due to the continuous readout, collision matching efficiency, which were never been a problem in Run 2, should be separately considered for validation. I set a default as "false", so it wouldn't change any process and results for the previous users.


PWGJE/Tasks/jetfinderQA.cxx
```diff
@@ -65,6 +65,7 @@ struct JetFinderQATask {
   Configurable<float> jetAreaFractionMin{"jetAreaFractionMin", -99.0, "used to make a cut on the jet areas"};
   Configurable<float> leadingConstituentPtMin{"leadingConstituentPtMin", -99.0, "minimum pT selection on jet constituent"};
   Configurable<float> randomConeR{"randomConeR", 0.4, "size of random Cone for estimating background fluctuations"};
+  Configurable<bool> checkMCcollisionsMatched{"checkMCcollisionsMatched", false, "0: count whole MCcollisions, 1: select MCcollisions which only have their correspond collisions"};
```
nzardosh, line 68: please change the name too checkMcCollisionIsMatched

PWGJE/Tasks/jetfinderQA.cxx
```diff
@@ -623,19 +624,31 @@ struct JetFinderQATask {
   }
   PROCESS_SWITCH(JetFinderQATask, processJetsMCDWeighted, "jet finder QA mcd with weighted events", false);
 
-  void processJetsMCP(soa::Join<aod::ChargedMCParticleLevelJets, aod::ChargedMCParticleLevelJetConstituents>::iterator const& jet, JetParticles const&)
+  void processJetsMCP(soa::Join<aod::ChargedMCParticleLevelJets, aod::ChargedMCParticleLevelJetConstituents>::iterator const& jet, JetParticles const&, JetMcCollisions const&, soa::Filtered<JetCollisionsMCD> const& collisions)
```
nzardosh, line 627: put collision at the beggining of the process function. 
I think instead of looping over all collisions for each jet you can slice the collision table based on the jet.McCollisionId(). If there is a McCollision in the slice then you can consider it matched

nzardosh (approved):

PWGJE/Tasks/jetfinderQA.cxx
joonsukbae: Thanks you for the exact comment.
Possibly, slicing the McCollisions would be needed such like below:
`if (checkMcCollisionsMatched && !jet.mcCollision()**.has_collision()**) {
      return;
    }`
However, as I know, there's no way to call the collision from mcCollision at this moment, right?
I guess I need to declare a new table or something? What do you think of it?
Or if there's a good way to avoid such big task, please let me know.

joonsukbae: Or another idea I have is slicing the collisions loop such as pseudo code below (Please tell me if you have any idea how to rewrite this with the correct one):
 ` Preslice<JetCollisionsMCD> CollisionsPerMcpJets =  soa::Join<aod::ChargedMCParticleLevelJets, aod::ChargedMCParticleLevelJetConstituents, aod::ChargedMCParticleLevelJetEventWeights>::JMcCollisionId;
  ...
  collisionspermcpejt = collisions.sliceBy(CollisionsPerMcpJets, jet.mcCollision().globalIndex());
  for (auto const& collision : collisions) {
    ...
  }`

joonsukbae:
I tested it locally and the results say it works exactly as I intended, cutting McCollision.

PWGJE/Tasks/jetfinderQA.cxx
```diff
@@ -623,19 +624,33 @@ struct JetFinderQATask {
   }
   PROCESS_SWITCH(JetFinderQATask, processJetsMCDWeighted, "jet finder QA mcd with weighted events", false);
 
-  void processJetsMCP(soa::Join<aod::ChargedMCParticleLevelJets, aod::ChargedMCParticleLevelJetConstituents>::iterator const& jet, JetParticles const&)
+  PresliceUnsorted<soa::Filtered<JetCollisionsMCD>> CollisionsPerMcpJets = aod::jmccollisionlb::mcCollisionId;
```
nzardosh, line 627: The name should be changed from CollisionsPerMcpJets to CollisionsPerMCPCollision

PWGJE/Tasks/jetfinderQA.cxx
```diff
@@ -65,6 +65,7 @@ struct JetFinderQATask {
   Configurable<float> jetAreaFractionMin{"jetAreaFractionMin", -99.0, "used to make a cut on the jet areas"};
   Configurable<float> leadingConstituentPtMin{"leadingConstituentPtMin", -99.0, "minimum pT selection on jet constituent"};
   Configurable<float> randomConeR{"randomConeR", 0.4, "size of random Cone for estimating background fluctuations"};
+  Configurable<bool> checkMcCollisionsMatched{"checkMcCollisionsMatched", false, "0: count whole MCcollisions, 1: select MCcollisions which only have their correspond collisions"};
```
nzardosh, line 68: checkMcCollisionsMatched -> checkMcCollisionIsMatched

PWGJE/Tasks/jetfinderQA.cxx
```diff
@@ -623,19 +624,33 @@ struct JetFinderQATask {
...
   {
     if (!jetfindingutilities::isInEtaAcceptance(jet, jetEtaMin, jetEtaMax, trackEtaMin, trackEtaMax)) {
       return;
     }
     if (!isAcceptedJet<JetParticles>(jet)) {
       return;
     }
+    if (checkMcCollisionsMatched) {
+      bool isMatchFound = false;
+      auto collisionspermcpejt = collisions.sliceBy(CollisionsPerMcpJets, jet.mcCollision().globalIndex());
```
nzardosh, line 638: jet.mcCollision().globalIndex() -> jet.mcCollisionId()

PWGJE/Tasks/jetfinderQA.cxx
nzardosh, line 627: move this to just below the init function

PWGJE/Tasks/jetfinderQA.cxx
```diff
@@ -623,19 +624,33 @@ struct JetFinderQATask {
...
+      for (auto const& collision : collisionspermcpejt) {
+        if (collision.mcCollision().globalIndex() == jet.mcCollision().globalIndex()) {
```
nzardosh, line 640: Do we need to check for this? Isnt it always true since you slice?

PWGJE/Tasks/jetfinderQA.cxx
```diff
@@ -649,6 +664,19 @@ struct JetFinderQATask {
         registry.fill(HIST("h_jet_ptcut_part"), jet.pt(), N * 0.25, jet.eventWeight());
       }
     }
+    if (checkMcCollisionsMatched) {
```
nzardosh, line 667: same comments above apply here too

joonsukbae:
I accepted all of your comments and just simplified the process as possible as I can.

nzardosh (approved):

PR #6128, opened 2024-05-13
[PWGUD] task phi->ee, p+pbar, p+Jpsi resonances

siragoni:
p+pbar task:

- added pt distributions for all interesting invariant mass ranges
- added control plots for polarisation of events in the sidebands to assure the nature of the physical process

phi->ee:

- introduced an exclusive phi->ee channel for measurements with SG events

p+Jpsi resonances:

- added a fun task to look for pentaquarks candidates requiring a Jpsi and a proton in TOF
- to be analysed with SG events

### DIFF
--- a/PWGUD/Tasks/exclusivePentaquark.cxx
+++ b/PWGUD/Tasks/exclusivePentaquark.cxx
@@ -1,0 +1,317 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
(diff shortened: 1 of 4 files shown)


PWGUD/Tasks/exclusivePentaquark.cxx
abylinkin, line 1: Lines 137-143: Don't you want to consider the track selection as implemented in 'Core/SGTrackSelector.h'?

PWGUD/Tasks/exclusiveTwoProtons.cxx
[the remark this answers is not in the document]
siragoni: Hi Sasha, where it matters e.g. L203, the proton mass is taken from `o2::constants::physics::MassProton`. The one in the polarisation macros it can be skipped. The reason is that it the helicity frame is dependent only on the rest frame of the Jpsi and its flight direction in the LAB frame

PWGUD/Tasks/exclusivePentaquark.cxx
siragoni: I will implement it for the next time. I need to test it - I remember you mentioned it, but I didn't have time to do that yet. My proposal is to approve this one and then by the end of next week I will push the update...

PR #6133, opened 2024-05-13
PWGHF: Ds-h, Correct a bug in histogram name

scattaru:
Dear all,
Just fixed the name of an histogram.

Cheers,
Samuele

### DIFF
--- a/PWGHF/HFC/TableProducer/correlatorDsHadrons.cxx
+++ b/PWGHF/HFC/TableProducer/correlatorDsHadrons.cxx
@@ -598,9 +598,10 @@ struct HfCorrelatorDsHadrons {
         isDsPrompt = particle.originMcGen() == RecoDecay::OriginType::Prompt;
         isDecayChan = particle.flagMcDecayChanGen() == decayChannel;
         std::vector<int> listDaughters{};
+        std::array<int, 3> arrDaughDsPDG = {+kKPlus, -kKPlus, kPiPlus};
         std::array<int, 3> prongsId;
         listDaughters.clear();
-        RecoDecay::getDaughters(particle, &listDaughters, std::array{}, 2);
+        RecoDecay::getDaughters(particle, &listDaughters, arrDaughDsPDG, 2);
         if (listDaughters.size() == 3) {
           for (auto iProng = 0; iProng < listDaughters.size(); ++iProng) {
             auto daughI = mcParticles.rawIteratorAt(listDaughters[iProng]);
--- a/PWGHF/HFC/TableProducer/correlatorDsHadrons.cxx
+++ b/PWGHF/HFC/TableProducer/correlatorDsHadrons.cxx
@@ -597,21 +597,22 @@ struct HfCorrelatorDsHadrons {
         // prompt and non-prompt division
         isDsPrompt = particle.originMcGen() == RecoDecay::OriginType::Prompt;
         isDecayChan = particle.flagMcDecayChanGen() == decayChannel;
+        std::vector<int> listDaughters{};
+        std::array<int, 3> prongsId;
+        listDaughters.clear();
+        RecoDecay::getDaughters(particle, &listDaughters, std::array{0}, 1);
+        if (listDaughters.size() == 3) {
+          for (auto iProng = 0; iProng < listDaughters.size(); ++iProng) {
+            auto daughI = mcParticles.rawIteratorAt(listDaughters[iProng]);
+            prongsId[iProng] = daughI.globalIndex();
+          }
+        }
         // Ds Hadron correlation dedicated section
         for (const auto& particleAssoc : mcParticles) {
           if (std::abs(particleAssoc.eta()) > etaTrackMax || particleAssoc.pt() < ptTrackMin || particleAssoc.pt() > ptTrackMax) {
             continue;
           }
-
-          if (particle.daughtersIds().size() != 3) {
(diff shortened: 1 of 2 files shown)


fgrosa (dismissed):

scattaru:
Dear all,
Sorry I have caught another error. In the process of analysis at the kinematic level I forgot to remove the candidate daughters from the correlations. I would like to ask if is it correct the way I have done. Since the Ds decay as Ds -> Phi Pi -> KKPi, the daughters taken from the MC particle tagged as Ds are the two kaon pair and the pion or the phi meson and the pion?

Thanks a lot in advance,
Cheers,
Samuele

PWGHF/HFC/TableProducer/correlatorDsHadrons.cxx
```diff
@@ -602,6 +602,18 @@ struct HfCorrelatorDsHadrons {
           if (std::abs(particleAssoc.eta()) > etaTrackMax || particleAssoc.pt() < ptTrackMin || particleAssoc.pt() > ptTrackMax) {
             continue;
           }
+
+          if (particle.daughtersIds().size() != 3) {
```
fgrosa, line 606: `daughterIds()` has always size 2 and it contains the ids of the first and the last daughters only. You should in principle loop from the first and the last, but also in this case it would not work in your case because the daughters would be a pion and a phi meson, so you should loop iteratively up to the final state particles. I would suggest you to use the `getDaughters` method in `RecoDecay` to easily get them: https://github.com/AliceO2Group/O2Physics/blob/f347b6bf54e06ce7271e963ebf0d065f40781962/Common/Core/RecoDecay.h#L600

fgrosa (changes requested):

PWGHF/HFC/TableProducer/correlatorDsHadrons.cxx
```diff
@@ -597,11 +597,24 @@ struct HfCorrelatorDsHadrons {
         // prompt and non-prompt division
         isDsPrompt = particle.originMcGen() == RecoDecay::OriginType::Prompt;
         isDecayChan = particle.flagMcDecayChanGen() == decayChannel;
+        std::vector<int> listDaughters{};
+        std::array<int, 3> prongsId;
+        listDaughters.clear();
+        RecoDecay::getDaughters(particle, &listDaughters, std::array{0}, 1);
```
fgrosa, line 603: The depth should be 2 (otherwise it checks only the first decay step, hence -> phi pi)
```suggestion
        RecoDecay::getDaughters(particle, &listDaughters, std::array{0}, 2);
```

PWGHF/HFC/TableProducer/correlatorDsHadrons.cxx
fgrosa: Hi @scattaru it should be like that, provided that you set the `depth` to `2` and not `1` (see comment above). To be 100% sure though I would suggest you to test it locally with some printouts ;)

scattaru: Hi @fgrosa thanks a lot! Then I'll try to test locally with these changes and see if it works

fgrosa: Hi @scattaru I think that you should also specify the pdg codes of the final daughter particles, otherwise they are never considered as `final` and they are not stored in the list of indices

vkucera: https://github.com/AliceO2Group/O2Physics/blob/30983eeaa6f821437864a0fa1d88a217d464cd76/Common/Core/RecoDecay.h#L597

fgrosa: > Without specifying the PDG codes, `getDaughters` will return the list of all daughters at the given maximum depth of the decay tree.

ah thanks, I indeed overlooked that! @scattaru my last comment then should be that you should replace `std::array{0}` with an empty array `std::array{}`, since otherwise you are providing one PDG code (`0`)

scattaru: Hi @fgrosa, @vkucera, thanks a lot for your help! Now I have replaced the array with an empty one. Looking at the similar part of the code written in the candidatecreator3prong I have used the array initialized with zero. I thought that there were not problems since that if the PDG scheme is the one used in PYTHIA there should not be particles with a PDG code = 0, but I could be wrong. However to be sure it is better as you suggested Fabrizio to use the empty array.

vkucera: @scattaru `{0}` would be fine since you correctly pointed out that there is no particle with that PDG.

scattaru:
I've checked locally that the code works adding in the histogram the number of daughters found for each Ds meson (bin 1,2,3), and in the bin 0 there should be the number of times that the `listDaughters` is not equal to 3. Looking at the histogram it seems that it works, there are always found 3 daughters.
Counter_daughters

Thanks a lot!
Cheers,
Samuele

PWGHF/HFC/TableProducer/correlatorDsHadrons.cxx
fgrosa: @scattaru @vkucera sorry for the second wrong suggestion! You can keep it with the daughter pdgs explicitly, or with `{0}`, both are fine for me.

scattaru: Hi @fgrosa, for me it is fine the way it is implemented now

fgrosa (approved):

PR #6140, opened 2024-05-14
Fix multiplicity table filling

njacazio:


njacazio:
@ddobrigk this fixes a bug when the table was not filled with the proper equalized multiplicity in case the non equalized tables were not subscribed to

ddobrigk (approved):

PR #6163, opened 2024-05-15
PWGHF: Extend and modify the derived-format data model

vkucera:
- Add original MC collision global indices.
- Rename the table with indices of the derived reconstructed collisions.
- Make rapidity a static column (for offline selections).
- Use `RecoDecayPtEtaPhi` methods in definitions of dynamic columns.
- Remove primary-vertex position from the candidate tables (redundant if no PV refit).
- Unify collision getters across species and table types (standard/stored).


nzardosh (approved):

alibuild (approved):

nzardosh (dismissed):

nzardosh (approved):

PR #6164, opened 2024-05-15
Add debug histograms for cent/mult

njacazio:


ddobrigk (approved):

vkucera:
@njacazio Adding the `THashList` object has introduced this error message:
```
[401931:internal-dpl-aod-global-analysis-file-sink]: [19:47:38][ERROR] Cannot read class info from buffer.
```
Please fix it.

PR #6187, opened 2024-05-16
[PWG-DQ] adding computation of the ITS ROF cut in VarManager

iarsene:


alibuild (approved):

iarsene:
Normally we should use the ITS ROF border cut from the event selection task, however, this development is just temporary to allow applying this cut on existing older DQ skims, for SQM2024 preliminary results. The code is replicating in the VarManager the code from eventSelection.cxx.

PR #6203, opened 2024-05-20
PWGMM: Mult: use explicit INEL>0 for binned case as well

aalkin:


alibuild (approved):

PR #6209, opened 2024-05-20
PWGHF: Exclude PWGHF/ALICE3 from compilation

vkucera:
Saves about 40 minutes of integrated CPU time.


alibuild (approved):

PR #6211, opened 2024-05-20
[PWGDQ] Updating histograms for Data / MC comparison

lucamicheletti93:
- addition of **dimuon-multi-diff** histogram class for multi-differential studies
- Modification of histogram ranges to allow better Data to MC comparison


alibuild (approved):

PR #6218, opened 2024-05-21
PWGMM: Mult: add process id in binned case

aalkin:


alibuild (approved):

PR #6228, opened 2024-05-21
Investigate ITS tracks

ilikmeta:


PWGCF/Flow/Tasks/FlowGFWPbPb.cxx
```diff
@@ -442,6 +452,25 @@ struct FlowGFWPbPb {
...
+
+      if (track.hasITS() && !track.hasTPC())
+        multITSnoTPC++;
+
+      if (track.hasITS())
+        multITSonly++;
+
+      if (track.hasITS() && track.hasTPC())
+        multITSTPC++;
+    }
```
victor-gonzalez, line 467: Has this code run locally and checked that it is getting the expected results?
Usually, if global tracks are required, they all have ITS and TPC by construction

victor-gonzalez (approved):
Have a look at my comment

PWGCF/Flow/Tasks/FlowGFWPbPb.cxx
ilikmeta: Thank you for your comment! My intention here is to see how the global tracks differ from ITS only by plotting histograms. Then, I would remove the requirement of global tracks and retest. My idea is to try to figure out why the elliptic flow from Pb-Pb Run 3 ends up below Run 2. The attached plot might help.
image

PR #6229, opened 2024-05-21
PWGCF: Added flags for Full PID IdBfFilter

bghanley1995:
Fixed an error on run where code was looking for Full PID from TPC and normal from TOF

### DIFF
--- a/PWGCF/TwoParticleCorrelations/TableProducer/identifiedBfFilter.cxx
+++ b/PWGCF/TwoParticleCorrelations/TableProducer/identifiedBfFilter.cxx
@@ -1210,7 +1210,8 @@ MatchRecoGenSpecies IdentifiedBfFilterTracks::trackIdentification(TrackObject co
   if (recoIdMethod == 0) {
     sp = kIdBfCharged;
   } else if (recoIdMethod == 1) {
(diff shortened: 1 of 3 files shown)


victor-gonzalez:
> Fixed an error on run where code was looking for Full PID from TPC and normal from TOF

Just to clarify, this description is misleading
The PR introduces Full PID in both TPC and TOF. Previously there were normal PID for both TPC and TOF
But that is not an error. With Full PID you just get better precision

victor-gonzalez (changes requested):
This code will not run unless the line
```
    if constexpr (framework::has_type_v<aod::pidtpc::TPCNSigmaPi, typename TrackObject::all_columns>) {
```
is also modified accordingly

Reminder: PR code should have been tested locally before submission

bghanley1995:
Hello Victor, I am confused. I did run locally and I don't seen any problem. I obviously trust your expertise, but I did not get any errors and saw results I expected to see when running the code.

bghanley1995:
Oh I see. Is it because with non-full TPC PID the call is `aod::pidtpc_tiny::TPCNSigmaPi`? I see my oversight in testing.

victor-gonzalez:
You should change the configuration for run it with Full PID
Also you have to use the Full TOF and Full TPC tasks in your workflow
In that way you will fully test your changes

victor-gonzalez (approved):

PR #6232, opened 2024-05-22
Update qaEfficiency.cxx

upasanasharma31:
To check simply adding the PDG value of the mother.


mfaggin (dismissed):

mfaggin:
@upasanasharma31 please fix the formatting issue, otherwise we cannot merge

njacazio:
Hi @upasanasharma31 I think this might be an issue with the HL more than an issue with the task, if we merge this we will get by default a request on the mothers

PR #6153, opened 2024-05-15
PWGHF: Basic Lambda_b task setup

mvolkl:
I added a basic framework to the Lb task for initial studies. In tests there was also a problem with taking the collision from the pion rather than the Lc in the candidate creator, which is occasionally empty, so I changed that.

PR #6237, opened 2024-05-22
PWGHF: LambdaB task and updates selector and creator

mvolkl:


vkucera:
Please try to avoid duplicating PRs. The original PR https://github.com/AliceO2Group/O2Physics/pull/6153 is already under review so it would be better to track the changes there.

PR #6260, opened 2024-05-23
DPG/PWGHF: monitor single-track efficiency from HF decays

mfaggin:
@fgrosa @fcolamar @gluparel @scattaru @mmazzilli @deepathoms this should help for the analysis of Samuele.
I still want to add the same code for the case w/o grouping by `mcCollisions`, I'll add it tomorrow and then I'll open the PR


scattaru:
Hi @mfaggin, 
Thanks a lot for this update!

Cheers,
Samuele

mfaggin:
In this PR I add the possibility to calculate the efficiency for tracks from HF decays.
This implied some changes:

- in `processMC` I needed to insert manually the loop over `McCollisions` grouping by hand the `Collisions`. This allows me to have the ungrouped `McParticles` table, avoiding unpredictable crashes in `getCharmHadronOrigin` (for b tracks, it might happen that the track is assigned to the wrong reco. collision, and if this `collision` belongs to a `mcCollision` different from the correct one for the beauty track and we work with grouped `mcCollisions`, then we miss the decay chain and the method fails)
- I also added the possibility to cut on the Z coordinate of the generated vertex (up to now, only a cut on the reco. posZ was applied)
- I replied the code in `processMCWithoutCollisions`. This duplication can be useful to study how the efficiency change if we ignore the track association to collisions (in `processMC` the efficiency is computed considering gen. particles only for reco. collisions, i.e beign a pure tracking efficiency indeed)

Tagging @fgrosa @fcolamar @scattaru @gluparel @deepathoms @mmazzilli in case they have further comments.
Tagging also @njacazio just to check whether I am not doing silly mistakes (btw from some check locally it seems ok)

alibuild (approved):

PR #6282, opened 2024-05-25
Common: add downscale factor to BC table for studies

ddobrigk:


alibuild (approved):

PR #6295, opened 2024-05-27
PWGLF: Remove V0Datas tables from cascadePreselector processes in the cascadebuilder

creetz16:
Hi @ddobrigk :)
Would it be possible to add these changes to the cascadebuilder? They would allow to run the cascadebuilder with the KF without having to run the lambdakzerobuilder to produce the V0Datas tables. 
As far as I see, the aod::V0Datas subscriptions in the cascadePreselector processes are not necessary. I ran some tests locally to make sure and everything looks fine so far. Am I overlooking something or could we indeed remove those subscriptions?

### DIFF
--- a/PWGLF/TableProducer/Strangeness/cascadebuilder.cxx
+++ b/PWGLF/TableProducer/Strangeness/cascadebuilder.cxx
@@ -1628,6 +1628,7 @@ struct cascadeBuilder {
   template <class TTrackTo, typename TCascTable>
   void buildKFStrangenessTables(TCascTable const& cascades)
   {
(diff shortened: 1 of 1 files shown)


ddobrigk:
Hi @creetz16 , indeed, I think this is a great idea and it would be useful to merge this - thanks a lot! Thanks also for testing. :-D

No objections to carrying on from my side.

creetz16:
Cool! Thanks a lot @ddobrigk :)

njacazio (approved):

njacazio:
Thanks @creetz16

PR #6307, opened 2024-05-28
PID tools: modify handleParamBase.h to use STF/ETF

jezwilkinson:
- Modify PID parameter handler (for CCDB uploads) to use STF/ETF instead of SOR/EOR information from RCT object on CCDB (following https://its.cern.ch/jira/browse/O2-3868)
  - in case STF/ETF are not available for the requested run numbers (e.g. in Run 2 data), the SOR/EOR are used as fallback
- Automatically adds recommended 2-minute buffer to start and end of validity when values from CCDB are used
- Fix minor bug in printout when setting timestamp


jezwilkinson:
@jgrosseo @iarsene @ddobrigk @ktf Hi, could this please be merged? this will ensure that future CCDB uploads for the TPC PID follow the most recent DPG guidelines for the timestamps

iarsene (approved):

PR #6311, opened 2024-05-28
PWGHF: Add configurable wrapper for event selection

vkucera:
Defines a `struct` which encapsulates event selection configuration and methods into a single task member with all its inner configurables exposed in a JSON group.

- Task configurables replaced with wrapper configurables.
  - Exposed in JSON in a `hfEvSel` block, e.g.:
    ```json
    "hf-track-index-skim-creator-tag-sel-collisions": {
        "fillHistograms": "true",
        "hfEvSel": {
            "centralityMin": "0",
            ...
        },
    ```
  - Device options (used in context inspection and on Hyperloop) get a prefix in their names: `zPvPosMax` -> `hfEvSel.zPvPosMax`.
- Creating of histograms and the function `setLabelHistoEvSel` replaced with the `addHistograms` wrapper method.
- Function `monitorCollision` replaced with the `fillHistograms` wrapper method which does not need histogram pointers as arguments.
  - Histogram filling simplified into a loop over rejection reasons, which guarantees the correct order of event rejection.
- Function `getHfCollisionRejectionMask` replaced with a wrapper method which does not need configurables as arguments.

Moves declarations of the `CentralityEstimator` enumerator into a separate header to make its usage independent from the data model and more modular.

Output validated, no significant difference in CPU or memory consumption.


mfaggin (dismissed):
thanks a lot @vkucera , it looks good to me!

mfaggin (approved):

PR #6321, opened 2024-05-29
Update qaEfficiency.cxx

upasanasharma31:
IU tracks


mfaggin (changes requested):
the case number 6 is "custom track cuts via configurables" but you are applying a cut hard-coded. I would make it configurable in some way

njacazio (approved):

PR #6322, opened 2024-05-29
Check the source of the signal for skimming

rbailhac:
The two reasons are:
a. Otherwise it is not possible to skimm with primary electrons for example. The rejection factor is much too small including secondary electrons from conversion happening far away (TRD...)
b. Otherwise there is an inconsistency between the signals defined at the reconstruction level (no source checked) and for reconstructed tracks in the histograms..


rbailhac:
Hi @iarsene , @dsekihat, @feisenhu  and @lucamicheletti93 , is this pull request ok for you ? This would allow us (dielectron) to skimm MC data with a resonable rejection factor, keeping all primary electrons. Thanks.

feisenhu (approved):

PR #6348, opened 2024-05-31
Common: BC table reduction -> min threshold for saving

ddobrigk:
Adds a minimum FT0C threshold to keep data volume down when saving the BC table for cross-checks.


ddobrigk:
The error looks unrelated. @ktf can you please check? Thank you very much!

PR #6352, opened 2024-05-31
Support divided centrality analyses for ITS-TPC m.e.

ChunzhengLab:
This PR introduces the following things:
- Implemented support for performing divided centrality analyses for the ITS-TPC matching efficiency by setting new process functions
- Implemented event selection 
- Optimized the settings for configurables

### DIFF
--- a/DPG/Tasks/AOTTrack/qaMatchEff.cxx
+++ b/DPG/Tasks/AOTTrack/qaMatchEff.cxx
@@ -14,14 +14,16 @@
 ///
 /// \author Rosario Turrisi  <rosario.turrisi@pd.infn.it>, INFN-PD
 /// \author Mattia Faggin <mattia.faggin@ts.infn.it>, UniTs & INFN-TS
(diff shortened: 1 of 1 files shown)


mfaggin (changes requested):
Hi @ChunzhengLab many thanks for your work.

Actually, in my opinion you can use a much simpler approach, which avoids code duplication. As you correctly do, you have to define a new `process` function which consumes the proper `Collisions` table `Join`ed with the centrality one. However, I would not define a new `fillHistogramsCentrality` function which fills many more different histograms, but after selecting the events per centrality I'd just go directly with the pre-existing `fillHistograms` function, so that you go an fill exactly the same histograms as now, the track loop is only in one place (i.e.: if we change something there it's valid for both pp and Pb-Pb w/o need to duplicate code and risk to forget in one of the 2 places) and in this way you can simply operate in Hyperloop with multiple suwagons changing the centrality cut, but then in all cases you analyse always histograms called in the same way.

What do you think?

ChunzhengLab:
Thank you @mfaggin , you are right. This way is simpler and avoids lots of duplication. I will make the changes as you said.

PR #6359, opened 2024-06-01
added cascades in/out of jets and task to study properties of tracked cascades

alcaliva:


alcaliva:
@ddobrigk @fmazzasc , do you mind looking at the task tracked_cascade_properties.cxx? 
The idea would be to study the average cluster size of tracked cascades to potentially use this cut to select clean incoming Xi for my Xi-Silicon scattering measurement

ercolessi (approved):

PR #6367, opened 2024-06-03
[PWGUD] Fix angle selection for exclusive phi->ee

siragoni:
Exclusive phi->KK:
- adding selections on the average cluster sizes in ITS 
- misidentified electrons are already relativistic -> small cluster size

Exclusive phi->ee:
- identified bug in the selection for the angles of the pair
- added study on the aperture of the angle of the pair (names could be improved though)

Pentaquark searches:
- relaxed global pT requirement on the tracks
- tightened a lot the pT requirement on the lepton tracks
- tightened the aperture of the Jpsi mass window to look for pentaquarks

### DIFF
--- a/PWGUD/Tasks/exclusivePentaquark.cxx
+++ b/PWGUD/Tasks/exclusivePentaquark.cxx
@@ -140,7 +140,7 @@ struct ExclusivePentaquark {
       if (trk.itsNCls() < 7) {
         continue;
       }
-      if (trk.pt() < 0.3) {
+      if (trk.pt() < 0.1) {
         continue;
       }
       if (!(std::abs(trk.dcaZ()) < 2.)) {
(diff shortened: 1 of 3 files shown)


siragoni:
Hi @abylinkin @rolavick this PR is good to go as far as I am concerned

rolavick (commented):
Hi Simone,
it seems to me that for some of the changes it would be more flexible to prepare them as configurables. Are you sure you want to have it like this or don't you want to improve the code?

siragoni:
> Hi Simone, it seems to me that for some of the changes it would be more flexible to prepare them as configurables. Are you sure you want to have it like this or don't you want to improve the code?

It is true that some of these changes will in the future work better as configurables, once I see they really are improving the selections. Where they could actually work better:
- in the pentaquark task, but we do not see pentaquark yet (in fact this might very well be the last iteration of this specific task since there is also a sever lack of statistics)
- in the exclusive phi->KK selection for the average cluster sizes, for systematic checks in the future, but at the moment I am not even sure what will be the effect of this particular cut
In conclusion: I agree with you, but for the future once I see the effects of these changes (same as in the `trackselector.h` helper functions)

rolavick:
> > Hi Simone, it seems to me that for some of the changes it would be more flexible to prepare them as configurables. Are you sure you want to have it like this or don't you want to improve the code?
> 
> It is true that some of these changes will in the future work better as configurables, once I see they really are improving the selections. Where they could actually work better:
> 
> * in the pentaquark task, but we do not see pentaquark yet (in fact this might very well be the last iteration of this specific task since there is also a sever lack of statistics)
> * in the exclusive phi->KK selection for the average cluster sizes, for systematic checks in the future, but at the moment I am not even sure what will be the effect of this particular cut
>   In conclusion: I agree with you, but for the future once I see the effects of these changes (same as in the `trackselector.h` helper functions)

Hi Simone,
ok, I will approve the PR as it is then.
Roman

rolavick (approved):

PR #6376, opened 2024-06-04
PWGHF: Tables for offline event mixing

deependra170598:
Dear @fgrosa @vkucera , 
In this PR I have tried to add a separate table for DstarOnly properties so that we can try offline event mixing on the derived dataset of Dstar and Associated Hadron Tables. These tables can be enabled with the provided configurable "enableSeparateTables" and by default set to False.


PWGHF/HFC/TableProducer/correlatorDstarHadrons.cxx
```diff
@@ -204,6 +210,19 @@ struct HfCorrelatorDstarHadrons {
                               assocParticle.pt(),
                               timestamp,
                               binNumber);
+
+          if (enableSeparateTables == true) {
```
fgrosa, line 214:
```suggestion
          if (enableSeparateTables) {
```

PWGHF/HFC/TableProducer/correlatorDstarHadrons.cxx
```diff
@@ -218,8 +237,28 @@ struct HfCorrelatorDstarHadrons {
                               assocParticle.pt(),
                               timestamp,
                               binNumber);
+          if (enableSeparateTables == true) {
```
fgrosa, line 240:
```suggestion
          if (enableSeparateTables) {
```

PWGHF/HFC/TableProducer/correlatorDstarHadrons.cxx
```diff
@@ -218,8 +237,28 @@ struct HfCorrelatorDstarHadrons {
...
+                      triggerParticle.eta(),
+                      triggerParticle.pt(),
+                      triggerParticle.invMassAntiDstar(),
+                      triggerParticle.invMassD0Bar(),
+                      timestamp,
+                      binNumber);
+          }
         } // endif
 
+        if (enableSeparateTables == true) {
```
fgrosa, line 253:
```suggestion
        if (enableSeparateTables) {
```

fgrosa (changes requested):
Hi @deependra170598 I just have few minor comments, then we can merge it.

PWGHF/HFC/TableProducer/correlatorDstarHadrons.cxx
```diff
@@ -218,8 +237,28 @@ struct HfCorrelatorDstarHadrons {
...
+            rowsDstar(collision.globalIndex(),
+                      gItriggerParticle,
+                      triggerParticle.phi(),
+                      triggerParticle.eta(),
+                      triggerParticle.pt(),
+                      triggerParticle.invMassAntiDstar(),
+                      triggerParticle.invMassD0Bar(),
+                      timestamp,
+                      binNumber);
+          }
```
vkucera, line 250: You can avoid filling `rowsDstar` at two places. The two instances just differ in the two masses, which you can assign inside this `if else` block and then have a single call of `rowsDstar()` outside.

PWGHF/HFC/TableProducer/correlatorDstarHadrons.cxx
vkucera, line 1: Remove the ineffective `triggerParticle.invMassD0();` statement on line 198.

vkucera (changes requested):

PWGHF/HFC/TableProducer/correlatorDstarHadrons.cxx
vkucera: Same for `rowsDstarHadronPair`.

deependra170598:
> Why do you still fill `rowsDstarHadronPair` even if `enableSeparateTables`?

Right now, `rowsDstarHadronPair` Table is being used in `taskCorrelatorDstarHadron.cxx`.  This structure is consistent with the other particle's TableProducers and Tasks. Not filling `rowsDstarHadronPair` table will result in a crash. 
This time only separate tables have been introduced for offline event mixing.

vkucera:
> > Why do you still fill `rowsDstarHadronPair` even if `enableSeparateTables`?
> 
> Right now, `rowsDstarHadronPair` Table is being used in `taskCorrelatorDstarHadron.cxx`. This structure is consistent with the other particle's TableProducers and Tasks. Not filling `rowsDstarHadronPair` table will result in a crash. This time only separate tables have been introduced for offline event mixing.

Does it mean that you will remove `rowsDstarHadronPair` from `taskCorrelatorDstarHadron` and replace it with the two tables in the next PR? Or you want to keep all tables and use them in different tasks?

deependra170598:
> > > Why do you still fill `rowsDstarHadronPair` even if `enableSeparateTables`?
> > 
> > 
> > Right now, `rowsDstarHadronPair` Table is being used in `taskCorrelatorDstarHadron.cxx`. This structure is consistent with the other particle's TableProducers and Tasks. Not filling `rowsDstarHadronPair` table will result in a crash. This time only separate tables have been introduced for offline event mixing.
> 
> Does it mean that you will remove `rowsDstarHadronPair` from `taskCorrelatorDstarHadron` and replace it with the two tables in the next PR? Or you want to keep all tables and use them in different tasks?

For the time being, I am keeping all the tables until event mixing problem is not fixed in O2Physics or any other solution does not come. If some future solution demands only `rowsDstarHadronPair` tables then I have to remove new tables added now and if situation demands to remove `rowsDstarHadronPair` table then that I will do that.

fgrosa (approved):

fgrosa:
Hi @deependra170598 this is good for me, let's see if it is good also for @vkucera

vkucera (approved):

PR #6384, opened 2024-06-05
[PWGDQ] Adding lower flow armonics calculation

lucamicheletti93:
- ZNA and ZNC Q-vectors calculation
- Temporary solution: no energy calibration included, for the moment set to alpha, c = 1


alibuild (approved):

PR #6386, opened 2024-06-05
added pions in jet + histo with ITSclsSize*tgl

alcaliva:


alcaliva:
@njacazio , I added the histograms for pions. What is still missing are DCA distributions to extract the fraction of secondaries and histograms for the efficiency. WIll add them in the next commit

ercolessi (approved):

PR #6391, opened 2024-06-05
PWGEM/PhotonMeson: reduce histogram bins

dsekihat:
PWGEM/PhotonMeson: reduce histogram bins


alibuild (approved):

PR #6403, opened 2024-06-06
Extend centrality QA

ercolessi:
This PR adds some QA histograms to the task for debug (optional).
Note that this would add EventSelection as a dependency to the centrality table, which is however already required by the multiplicity table.

Cc: @njacazio


ddobrigk (approved):
Thanks @ercolessi ! 

This reminds me we should really get started with the use of metadata, since that will greatly simplify dependencies on hyperloop once implemented widely...

PR #6420, opened 2024-06-07
[PWGDQ] Fixing bugs of table-reader-with-association

lucamicheletti93:
- Additional protections for same-event-pairing
- Removal of fConfigQA from event selection causing breaking when not enabled


alibuild (approved):

PR #6426, opened 2024-06-07
PWGHF: fix vector access if searchUpToQuark=true

fcatalan92:
If `searchUpToQuark` is `true` in `getCharmHadronOrigin`, the vector with the indexes of B mothers could be empty


alibuild (approved):

PR #6439, opened 2024-06-09
PWGDQ:LMee: added occupancy variable to data model and var manager

jjungIKF:


feisenhu:
The pull request seems fine from my side.
@iarsene could you have a second look. I'm not sure if there was some effort from DQ side for the occupancy.

iarsene:
Hi @jjungIKF, @feisenhu,
I think we need this kind of variables. I am actually also preparing a rather big PR (related mainly to MC, but also to this) and I would like to ask to not merge this. From my side it will take a couple of days to push that. 
There is also a technical reason. If you implement it like this, you will modify the extended event table and all existing skims will become unusable with the new O2Physics versions. We should discuss in a meeting these changes before we merge them. But anyway, i suggest we don't merge this. What I will do is that I will create a completely new table where I will implement this variable, but in addition several other variables which are needed, like number of TPC, ITS, TRD, TOF tracks, and a few others.

rbailhac:
Hi @iarsene ,

I think you have a point with the backward compatibility. This would mean we will need an additional process function in the tablemaker and in the tablereader to have these variables, I guess. Hopefully the other variables you plan to pack in this additional table are useful for us as well and not too much CPUs.

You know now that we are waiting for this. Please let us know your progress and when we can start to look at the data in term of occupancy.

Thanks,

Raphaelle

iarsene:
Hi @rbailhac,
Yes, I think I should have it quickly. I mostly have the code and I am now testing locally. Regarding process functions, if I remember correctly, we are at the limit with the number of process functions for the old tableMaker workflow, but I hope I can merge some by replacing them with configurables. I have to see. I might send an email to make a poll regarding the process functions we really want. We need to converge on a smaller set of possible combinations, something on the line of what I added in the new "_withAssoc" workflows.

iarsene:
Hi, so if you all agree maybe we should close this and wait for my PR?

PR #6452, opened 2024-06-10
PWGHF: move from TH1F to TH1D for monitor histos and set the default min dca cut to 0

fgrosa:


mfaggin (dismissed):
thanks @fgrosa

alibuild (approved):

PR #6460, opened 2024-06-10
add PhotonIsolationQA.cxx task

berend01:
This task is designed to select EMCal clusters originating from prompt photons. Cluster selection is performed using Pt_Iso (density of particles around the cluster minus UE density) and cluster shape. The task can also be run over Monte Carlo data to obtain a correction factor for the purity measurement using the ABCD method.


PWGJE/Tasks/PhotonIsolationQA.cxx
```diff
@@ -0,0 +1,364 @@
...
+  Configurable<int> maxNLM{"maxNLM", 2, "Maximal amount of local maxima per cluster"};
+  Configurable<bool> ExoticContribution{"ExoticContribution", false, "Exotic cluster in the data"};
+  Configurable<float> TrackMatchingRadius{"TrackMatchingRadius", 0.05, "maximum radius between track and cluster in eta phi plane"};
+  Configurable<float> minDPhi{"minDPhi", 0.01, "Minimum dPhi between track and cluster"};
+
+  Filter PosZFilter = nabs(aod::collision::posZ) < maxPosZ;
+  Filter clusterDefinitionSelection = (o2::aod::emcalcluster::definition == mClusterDefinition) && (o2::aod::emcalcluster::time >= minTime) && (o2::aod::emcalcluster::time <= maxTime) && (o2::aod::emcalcluster::energy > minClusterEnergy) && (o2::aod::emcalcluster::nCells >= minNCells) && (o2::aod::emcalcluster::nlm <= maxNLM) && (o2::aod::emcalcluster::isExotic == ExoticContribution);
+  Filter emccellfilter = aod::calo::caloType == 1; // mc emcal cell
+
+  using selectedCollisions = soa::Filtered<collisionEvSelIt>;
```
nzardosh, line 87: what is collisionEvSelIt?

nzardosh (commented):
Hi. I think this is a task that should be reviewed by the EMCAL experts since there is no use of jets. So I will leave it up to them. Thanks!

PWGJE/Tasks/PhotonIsolationQA.cxx
berend01: it is defined in line 58: using collisionEvSelIt = o2::soa::Join<o2::aod::Collisions, o2::aod::EvSels>;

PWGJE/Tasks/PhotonIsolationQA.cxx
```diff
@@ -0,0 +1,364 @@
...
+      if (dPhi < 0) {
+        dPhi = 2. * M_PI - dPhi;
+      }
+      double distance = sqrt(pow(dEta, 2) + pow(dPhi, 2));
+      if (distance <= TrackMatchingRadius) {
+        double abs_p = abs(match.track_as<myGlobTracks>().p());
+        if ((cluster.energy() / abs_p) < 1.75) {
+          return true;
+        }
+      }
```
mfasDa, line 182: I am wondering how this method was developed. In run2 we use a pt-dependent matching with a matching function, independent in eta and phi. Having a hard coded, and rather large cut, can lead to fake matches at high-E.

mfasDa (approved):
The PR can be merged, but for the next iteration a more optimal track matching criterion should be considered.

PWGJE/Tasks/PhotonIsolationQA.cxx
fjonasALICE: @mfasDa for Run 2 prompt photons we usually used a fixed dEta and dPhi.  We choose a wide cut, because we found that missed matches may lead to unnatural shapes in the isolation energy that are self correlated with photon pt. In any case, for isolated prompt photons the track matching has strong interplay. This the track matching misses get caught by isolation cut and vice versa

PWGJE/Tasks/PhotonIsolationQA.cxx
```diff
@@ -0,0 +1,364 @@
...
+    }
+    return false;
+  }
+
+  // sums the pt of all tracks around the cluster within a radius of R = 0.4
+  double sum_Pt_tracks_in_cone(const auto& cluster, myGlobTracks const& tracks, double Cone_Radius = 0.4)
+  {
+    double sum_Pt = 0.;
+    for (const auto& track : tracks) {
+      double dphi = cluster.phi() - track.phi();
```
fjonasALICE, line 192: dphi calculation is I think not fully correct? e.g. for phi from 0 to 2pi, if cluster phi would theoretically be close to 2pi and track phi very close to 0, then you would get a very large dPhi distance even though cluster and track are close. Double check definitions throughout. Maybe I am wrong

berend01: You are correct. I have fixed that bug in the PR that is now pending (https://github.com/AliceO2Group/O2Physics/pull/6689). In the pending PR i also fixed the track matching (i think). The emcal-correction task will now fully handle the tracks matched to a cluster

PR #6462, opened 2024-06-10
PWGCF - add task for flow analysis with Generic Framework

cnkoster:


victor-gonzalez (changes requested):
Thanks for implementing this analysis
Please consider 
- move the source code of your task to `PWGCF/Flow/Tasks`
- don't include `Task` in your source code file name
- use a meaningful, short, title for the PR based on the functionality incorporated and not the source file/s
- don't leave code commented

PWGCF/Flow/Tasks/flowAnalysisGF.cxx
```diff
@@ -0,0 +1,686 @@
...
+    registry.fill(HIST("pt_phi_aft"), track.pt(), phimodn);
+    return true;
+  }
+
+  enum datatype {
+    kReco,
+    kGen
+  };
+
+  template <datatype dt>
```
victor-gonzalez, line 465: Any reason for `dt` being a template parameter and not just a function parameter?

PWGCF/Flow/Tasks/flowAnalysisGF.cxx
```diff
@@ -0,0 +1,686 @@
...
+          continue;
+        auto val = fGFW->Calculate(corrconfigs.at(l_ind), i - 1, kFALSE).real() / dnx;
+        if (TMath::Abs(val) < 1)
+          (dt == kGen) ? fFC_gen->FillProfile(Form("%s_pt_%i", corrconfigs.at(l_ind).Head.c_str(), i), centmult, val, dnx, rndm) : fFC->FillProfile(Form("%s_pt_%i", corrconfigs.at(l_ind).Head.c_str(), i), centmult, val, dnx, rndm);
+      }
+    }
+    return;
+  }
+
+  template <datatype dt, typename TCollision, typename TTracks>
```
victor-gonzalez, line 495: Same question here

PWGCF/Flow/Tasks/flowAnalysisGF.cxx
```diff
@@ -0,0 +1,686 @@
...
+        fWeights->Fill(track.phi(), track.eta(), vtxz, track.pt(), centrality, 0);
+
+      if (!setCurrentParticleWeights(weff, wacc, track.phi(), track.eta(), track.pt(), vtxz))
+        return;
+
+      registry.fill(HIST("phi_eta_vtxZ_corrected"), track.phi(), track.eta(), vtxz, wacc);
+      if (cfgFillQA)
+        FillTrackQA<kReco>(track, vtxz);
+
+      FillGFW(track, weff, wacc);
```
victor-gonzalez, line 567: I guess you don't need to replicate this code, you already have it in the reco section above
I would suggest the first `if` condition handles the MC generated, its `else` handles the reconstructed with an inner `if` for the handling of the MC reco particularity and the rest in common for data and MC reco

PWGCF/Flow/Tasks/flowAnalysisGF.cxx
```diff
@@ -0,0 +1,686 @@
...
+      return;
+    const auto centrality = collision.centFT0C();
+    auto bc = collision.bc_as<aod::BCsWithTimestamps>();
+    if (cfgUseAdditionalEventCut && !eventSelected(collision, tracks.size(), centrality))
+      return;
+    if (cfgFillQA)
+      FillEventQA(collision, tracks);
+    loadCorrections(bc.timestamp());
+    auto field = (cfgMagField == 99999) ? getMagneticField(bc.timestamp()) : cfgMagField;
+    processCollision<kReco>(collision, tracks, centrality, field);
```
victor-gonzalez, line 651: Why replicating this code?
I would recommend a single routine for processing reconstructed, be data or MC detector level, and then call it from both `process` functions

PWGCF/Flow/Tasks/flowAnalysisGF.cxx
```diff
@@ -0,0 +1,686 @@
...
+
+  void processRun2(soa::Filtered<soa::Join<aod::Collisions, aod::EvSels, aod::Mults, aod::CentRun2V0Ms>>::iterator const& collision, aod::BCsWithTimestamps const&, myTracks const& tracks)
+  {
+    if (!collision.sel7())
+      return;
+    const auto centrality = collision.centRun2V0M();
+    auto bc = collision.bc_as<aod::BCsWithTimestamps>();
+    loadCorrections(bc.timestamp());
+    auto field = (cfgMagField == 99999) ? getMagneticField(bc.timestamp()) : cfgMagField;
+    processCollision<kReco>(collision, tracks, centrality, field);
```
victor-gonzalez, line 676: Most probably this one can be also incorporated to the above mentioned unified reconstructed handling with the appropriate check of an existing column in the case of Run2 (as you check for the existence of the `McParticleId` in the case of reconstructed tracks for detecting MC detector level tracks)

PWGCF/Flow/Tasks/flowAnalysisGF.cxx
cnkoster: I copied it from the flowGenericFramework task, where a datatype is used. I think it is because dt is either kGen or kReco and when using it in this way it is always clear which one is used in the different process functions.

PWGCF/Flow/Tasks/flowAnalysisGF.cxx
cnkoster: Thank you for the suggestion, however, I'm not sure how to implement it correctly because for the data "track" is used and for the MC reconstructed "track.mcParticle()" is used. I couldn't find an easy way to treat both of them in a common statement for all reconstructed without duplicating the code, could it be I'm missing something?

victor-gonzalez (changes requested):
Thanks for addressing my comments
Please consider the additional ones

victor-gonzalez (approved):
Thanks for addressing my comments!

PR #6474, opened 2024-06-11

vkucera:
unused-parameter
missing constructor
non-enum vs enum in conditional expression
missing-braces

### DIFF
--- a/PWGCF/JCorran/Core/JFFlucAnalysis.cxx
+++ b/PWGCF/JCorran/Core/JFFlucAnalysis.cxx
@@ -127,7 +127,7 @@ TComplex JFFlucAnalysis::Four(int n1, int n2, int n3, int n4)
 #undef C
 
 //________________________________________________________________________
-void JFFlucAnalysis::UserExec(Option_t*  /*popt*/)
+void JFFlucAnalysis::UserExec(Option_t* /*popt*/)
 {
   TComplex corr[kNH][nKL];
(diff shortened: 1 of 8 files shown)


PWGCF/JCorran/Core/JFFlucAnalysis.cxx
```diff
@@ -41,7 +43,8 @@ JFFlucAnalysis::JFFlucAnalysis(const char* /*name*/) : fVertex(0),
 }
 
 //________________________________________________________________________
-JFFlucAnalysis::JFFlucAnalysis(const JFFlucAnalysis& a) : fVertex(a.fVertex),
+JFFlucAnalysis::JFFlucAnalysis(const JFFlucAnalysis& a) : TNamed(),
```
victor-gonzalez, line 46: Although it seems they are not using the object name, in principle this should be 
```
JFFlucAnalysis::JFFlucAnalysis(const JFFlucAnalysis& a) : TNamed(a),
```

victor-gonzalez (changes requested):
@vkucera Thanks for doing this!!!
Just a minor request from my side

PWGCF/JCorran/Core/JFFlucAnalysis.cxx
vkucera: Done

alibuild (approved):

PR #6495, opened 2024-06-12
[PWG-DQ] Adding MC workflows which consider track-collision associations

iarsene:
Includes a new table-maker (table-maker-mc-with-assoc) and a new analysis workflow (dq-efficiency-with-assoc).


alibuild (approved):

PR #6496, opened 2024-06-12
Track pool creator for V0s and kinks

fmazzasc:
Utility to create offline V0 and kink pools looking at their track time compatibility. Hypertriton task adapted consequently


alibuild (approved):

PR #6502, opened 2024-06-12
PWGCF: Update MeanPtFlucIdentified.cxx

TGahlaut1:


victor-gonzalez (approved):
That look much more better and you will appreciate it when you evolve your code
As an additional comment, I would suggest for the next iteration to use an `enum` to define the meaning of the `0`, `1`, ..., `7`, ... when you call the `template` functions
That will make the code even clearer
Thanks!

PR #6509, opened 2024-06-13
[PID] use metadata

njacazio:


ddobrigk:
@njacazio this is awesome :-D - did you check that all is fine also for the Run 2 path? I have to ask ;-)

njacazio:
For run 2 it's ok

ddobrigk (approved):

jgrosseo:
@njacazio @ddobrigk The meta data is disabled on Hyperloop at present!
There is a bug preventing its use.

Does this also work without metadata? Otherwise revert.

njacazio:
Hi, 
Apologies I just read the comment. The pr should be safe for disabled metadata

PR #6510, opened 2024-06-13
PWGEM: Add cut with only TPC electron PID, no tracking

feisenhu:
Adding a cut using only electron PID from TPC with no tracking cuts.


alibuild (approved):

PR #6519, opened 2024-06-14
PWGLF : add event selection QA

Yunseul-Bae:
Dear Experts,

I am Yunseul, and I have recently stated working on the f0(980) analysis.
In this PR, I have added a histogram for selection QA, which plots the number of events per event selection step. Additionally, I have plotted the z primary vertex distribution using only the sel8 selection.

A few days ago, I submitted a similar PR, but it was closed due to a Git issue. I apologize for any confusion this may have caused.

### DIFF
--- a/PWGLF/Utils/collisionCuts.h
+++ b/PWGLF/Utils/collisionCuts.h
@@ -233,5 +233,4 @@ class CollisonCuts
   float mZvtxMax = 999.f;                          ///< Maximal deviation from nominal z-vertex (cm)
 };
 } // namespace o2::analysis
-
 #endif // PWGLF_UTILS_COLLISIONCUTS_H_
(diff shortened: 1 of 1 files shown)


BongHwi:
Hello @Yunseul-Bae,

Thank you for your contribution! 
It looks like you intended to add the QA histogram for event selection and to fill the z-Vertex histogram before all selections. 
I have a couple of suggestions to improve this:

1. Please add a separate z-Vertex QA histogram before the selection process. The current z-Vertex histogram is necessary for QA, so we should maintain the existing structure without modifications.
2. Could you update the code structure to use bitwise flags for event selection? The current implementation with `sel8` won't work properly unless `mCheckOffline` is activated.

Thanks again for your efforts!

PR #6522, opened 2024-06-14
adding new 2trck cut

glromane:
Dear @victor-gonzalez ,

in this commit we add a new double track cut. Could you please approve it. Thank you.

Sincerely yours, Gleb.

### DIFF
--- a/PWGCF/Femto3D/Core/femto3dPairTask.h
+++ b/PWGCF/Femto3D/Core/femto3dPairTask.h
@@ -16,6 +16,8 @@
 #ifndef PWGCF_FEMTO3D_CORE_FEMTO3DPAIRTASK_H_
 #define PWGCF_FEMTO3D_CORE_FEMTO3DPAIRTASK_H_
 
+#define THETA(eta) 2.0 * atan(exp(-eta))
+
 // #include "Framework/ASoA.h"
 // #include "Framework/DataTypes.h"
 // #include "Framework/AnalysisDataModel.h"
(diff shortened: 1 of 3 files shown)


PWGCF/Femto3D/Core/femto3dPairTask.h
```diff
@@ -253,6 +260,24 @@ bool FemtoPair<TrackType>::IsClosePair(const float& deta, const float& dphi, con
...
+    return -100.f;
+
+  float dtheta = THETA(_first->eta()) - THETA(_second->eta());
+  float res = 0.0;
+
+  for (const auto& radius : TPCradii) {
+    res += sqrt(pow(2.0 * radius * sin(0.5 * GetPhiStarDiff(radius)), 2) + pow(2.0 * radius * sin(0.5 * dtheta), 2));
+  }
+
+  return 100.0 * res / 9;
```
victor-gonzalez, line 278: I would recommend using the length of the `TPCradii` instead of `9`

victor-gonzalez (dismissed):
Have a look at my comment
I approve it but I will wait for merging it in case you want to modify it

PWGCF/Femto3D/Core/femto3dPairTask.h
glromane: Dear @victor-gonzalez , thank you your comment. The size if fixed so I used just 9, now I changed it to std::array and calling the size() method. Is it OK like this?

victor-gonzalez: Thanks! It is OK
My point was that if someone in the future changes the size of the array by including more radii they only need to care about the array and nothing else

victor-gonzalez (approved):

PWGCF/Femto3D/Core/femto3dPairTask.h
glromane: ah, I see. Thanks!

PR #6525, opened 2024-06-14
PWGHF: LambdaB task, add normalisation to likelihood

mvolkl:


mvolkl:
Sorry. I'll try to remove the earlier commits that are somehow still in.

vkucera:
No problem. But please avoid closing a PR and creating a new one from the same branch. It is completely useless. If you need to fix something in your branch history, you can just mark the PR as a draft and push to the branch when your done fixing it.

PR #6533, opened 2024-06-14
PWGHF: use ZORRO utility to selected triggered events in HF code

fgrosa:
This PR adds the possibility to select events that were triggered by one (or multiple) specific software triggers using the `ZORRO` utility in the HF code.

### DIFF
--- a/PWGHF/D2H/Utils/utilsRedDataFormat.h
+++ b/PWGHF/D2H/Utils/utilsRedDataFormat.h
@@ -27,7 +27,7 @@ namespace o2::hf_evsel
 /// \tparam centEstimator centrality estimator
 /// \param collision collision to test against the selection criteria
 template <bool useEvSel, o2::hf_centrality::CentralityEstimator centEstimator, typename Coll, typename BC>
-void checkEvSel(Coll const& collision, BC const& bc, o2::hf_evsel::HfEventSelection& hfEvSel, int& zvtxColl, int& sel8Coll, int& zvtxAndSel8Coll, int& allSelColl, o2::framework::Service<o2::ccdb::BasicCCDBManager> const& ccdb)
+void checkEvSel(Coll const& collision, BC const& bc, o2::hf_evsel::HfEventSelection& hfEvSel, int& zvtxColl, int& sel8Coll, int& zvtxAndSel8Coll, int& zvtxAndSel8CollAndSoftTrig, int& allSelColl, o2::framework::Service<o2::ccdb::BasicCCDBManager> const& ccdb)
 {
   float centrality{-1.f};
   const auto rejectionMask = hfEvSel.getHfCollisionRejectionMask<useEvSel, o2::hf_centrality::CentralityEstimator::None>(collision, centrality, bc, ccdb);
@@ -39,6 +39,9 @@ void checkEvSel(Coll const& collision, BC const& bc, o2::hf_evsel::HfEventSelect
   }
   if (!TESTBIT(rejectionMask, o2::hf_evsel::EventRejection::PositionZ) && !TESTBIT(rejectionMask, o2::hf_evsel::EventRejection::Trigger)) {
     zvtxAndSel8Coll++;
+  }
+  if (!TESTBIT(rejectionMask, o2::hf_evsel::EventRejection::PositionZ) && !TESTBIT(rejectionMask, o2::hf_evsel::EventRejection::Trigger) && !TESTBIT(rejectionMask, o2::hf_evsel::EventRejection::SoftwareTrigger) ) {
+    zvtxAndSel8CollAndSoftTrig++;
   }
   if (rejectionMask == 0) {
     allSelColl++;
--- a/PWGHF/D2H/Utils/utilsRedDataFormat.h
+++ b/PWGHF/D2H/Utils/utilsRedDataFormat.h
@@ -26,11 +26,11 @@ namespace o2::hf_evsel
 /// \tparam useEvSel use information from the EvSel table
 /// \tparam centEstimator centrality estimator
 /// \param collision collision to test against the selection criteria
-template <bool useEvSel, o2::hf_centrality::CentralityEstimator centEstimator, typename Coll, typename BC>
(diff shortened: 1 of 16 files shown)


PWGHF/Utils/utilsEvSelHf.h
```diff
@@ -117,8 +125,8 @@ struct HfEventSelection : o2::framework::ConfigurableGroup {
   /// \param collision collision to test against the selection criteria
   /// \param centrality collision centrality variable to be set in this function
   /// \return bitmask with the event selection criteria not satisfied by the collision
-  template <bool useEvSel, o2::hf_centrality::CentralityEstimator centEstimator, typename Coll>
-  uint16_t getHfCollisionRejectionMask(const Coll& collision, float& centrality)
+  template <bool useEvSel, o2::hf_centrality::CentralityEstimator centEstimator, typename Coll, typename BC>
+  uint16_t getHfCollisionRejectionMask(const Coll& collision, float& centrality, const BC& bc, o2::framework::Service<o2::ccdb::BasicCCDBManager> const& ccdb)
```
vkucera, line 129:
```suggestion
  uint16_t getHfCollisionRejectionMask(const Coll& collision, float& centrality, const BC& bc, o2::framework::Service<o2::ccdb::BasicCCDBManager> const& ccdb)
```
I don't think it is needed to pass `bc` as argument. It can be retrieved from `collision`.

vkucera: Hmm, I get your point, but I can also see that you had to add almost 30 new lines `auto bc = ` just to pass `bc` as the argument.

fgrosa: Done in [f5be35a](https://github.com/AliceO2Group/O2Physics/pull/6533/commits/f5be35a8d879a866908f2e13c4eb38408c6b05ae). I am still not convinced that accessing twice per candidate the same table is worth sparing 3 lines of code per task, but given that is minor if you prefer so we can proceed like this. Thanks!

PWGHF/Utils/utilsEvSelHf.h
```diff
@@ -117,8 +125,8 @@ struct HfEventSelection : o2::framework::ConfigurableGroup {
   /// \param collision collision to test against the selection criteria
   /// \param centrality collision centrality variable to be set in this function
   /// \return bitmask with the event selection criteria not satisfied by the collision
-  template <bool useEvSel, o2::hf_centrality::CentralityEstimator centEstimator, typename Coll>
-  uint16_t getHfCollisionRejectionMask(const Coll& collision, float& centrality)
+  template <bool useEvSel, o2::hf_centrality::CentralityEstimator centEstimator, typename Coll, typename BCs>
+  uint16_t getHfCollisionRejectionMask(const Coll& collision, float& centrality, const BCs& /*bcs*/, o2::framework::Service<o2::ccdb::BasicCCDBManager> const& ccdb)
```
vkucera, line 129:
```suggestion
  template <bool useEvSel, o2::hf_centrality::CentralityEstimator centEstimator, typename BCs, typename Coll>
  uint16_t getHfCollisionRejectionMask(const Coll& collision, float& centrality, o2::framework::Service<o2::ccdb::BasicCCDBManager> const& ccdb)
```

PWGHF/D2H/Utils/utilsRedDataFormat.h
```diff
@@ -25,19 +26,26 @@ namespace o2::hf_evsel
...
 /// \param collision collision to test against the selection criteria
-template <bool useEvSel, o2::hf_centrality::CentralityEstimator centEstimator, typename Coll>
-void checkEvSel(Coll const& collision, o2::hf_evsel::HfEventSelection& hfEvSel, int& zvtxColl, int& sel8Coll, int& zvtxAndSel8Coll, int& allSelColl)
+template <bool useEvSel, o2::hf_centrality::CentralityEstimator centEstimator, typename Coll, typename BCs>
+void checkEvSel(Coll const& collision, BCs const& bcs, o2::hf_evsel::HfEventSelection& hfEvSel, int& zvtxColl, int& sel8Coll, int& zvtxAndSel8Coll, int& zvtxAndSel8CollAndSoftTrig, int& allSelColl, o2::framework::Service<o2::ccdb::BasicCCDBManager> const& ccdb)
 {
   float centrality{-1.f};
-  const auto rejectionMask = hfEvSel.getHfCollisionRejectionMask<useEvSel, o2::hf_centrality::CentralityEstimator::None>(collision, centrality);
-  if (!TESTBIT(rejectionMask, o2::hf_evsel::EventRejection::Trigger))
+  const auto rejectionMask = hfEvSel.getHfCollisionRejectionMask<useEvSel, o2::hf_centrality::CentralityEstimator::None>(collision, centrality, bcs, ccdb);
```
vkucera, line 33:
```suggestion
  const auto rejectionMask = hfEvSel.getHfCollisionRejectionMask<useEvSel, o2::hf_centrality::CentralityEstimator::None, BCs>(collision, centrality, ccdb);
```

vkucera (commented):
For reference, this is what I had in mind.

PWGHF/Utils/utilsEvSelHf.h
fgrosa: If any, I think it's a negligilbe overhead I would say, so we can proceed with the new one 👍

PWGHF/Utils/utilsEvSelHf.h
fgrosa: Ah, I did not know that was enough like this. It is indeed less verbose, I will change it in the next commit

vkucera: Thanks

vkucera, line 129: Also please add `/// \param ccdb ...` in the documentation

fgrosa: It should all be done here [d7a1886](https://github.com/AliceO2Group/O2Physics/pull/6533/commits/d7a1886852a4b68448f3d4af61b5e37af57b8fab). Could you please take a look if this is what you had in mind? Thanks!

vkucera: Thanks

vkucera:
Thanks @fgrosa , it looks good to me.
There is just the compilation error to fix.

fgrosa:
> Thanks @fgrosa , it looks good to me. There is just the compilation error to fix.

Strange that it compiled locally, now fixed anyway. Thanks!

alibuild (approved):

PR #6536, opened 2024-06-14
Common: add metadata helper

njacazio:


ddobrigk (approved):
Thanks @njacazio ! I'll attempt to incorporate this already in centrality and then shortly afterwards also in the strangeness services :-D

PR #6548, opened 2024-06-16
update of occupancy-based event selection

altsybee:
Occupancy variable in evSel now is weighted wrt time separation b/n given collision and other events. 
Also, several new occupancy-based selection bits are added.


ekryshen (approved):
Thanks, Igor! Approving on my side. I guess another approval is required from David for changes in eventSelection.cxx

ddobrigk (approved):

PR #6555, opened 2024-06-17
[PWGDQ ]Adding extra table to compute Q-vectors

lucamicheletti93:
Adding extra table to compute Q-vectors from ZDC tables. The table is joinable and filled together with the Q-vectors from central framework


alibuild (approved):

PR #6559, opened 2024-06-17
Avoid double counting in the CEFP summary plot

mpuccio:
The double counting has been introduced with the extension of the trigger mask to 128 bits


alibuild (approved):

PR #6562, opened 2024-06-17
[PWGDQ] Adding event selection for triggered events

lucamicheletti93:
Adding event selection for triggered events by CEFP using the Zorro tool


lucamicheletti93:
Updates done

PR #6566, opened 2024-06-17
Add macro to check the bcRanges.root files

mpuccio:


alibuild (approved):

PR #6573, opened 2024-06-18
PWGHF: factorize centrality table in Lc tree creator.

mfaggin:
Tagging @DelloStritto @gluparel @strogolo @saganatt @zhangbiao-phy @1994ra for their info.
With this, the `centrality` workflow can be removed from the dependencies, if centrality is not needed.
Beware that there are no more `processData` and `processMc`, but `processDataNoCentrality`, `processDataWithCentrality`, `processMcNoCentrality`, `processMcWithCentrality` are now present.


fgrosa (approved):
Hi @mfaggin, thanks! This looks good to me, it's good not to depend on centrality if not needed. Let's see if the other Lc users have concerns, otherwise we can merge it later today.

PR #6596, opened 2024-06-20
EventFiltering/PWGHF: Fix typo in indices for Ds* selection

fgrosa:
@zhangbiao-phy this fixes the issue in the Ds* selection


alibuild (approved):

PR #6597, opened 2024-06-20
[Multiplicity] Auto enable the processMC

njacazio:


ddobrigk (approved):
Thanks!  

Not sure if the CI will allow merging, let's see

PR #6604, opened 2024-06-20
PWGHF: Add XicPlus to Xi Pi Pi workflow

pstahlhu:
First version of XicPlus to Xi Pi Pi workflow
@creetz16


PWGHF/TableProducer/candidateCreatorXicToXiPiPi.cxx
```diff
@@ -0,0 +1,655 @@
...
+#include "PWGHF/Utils/utilsBfieldCCDB.h"
+
+using namespace o2;
+using namespace o2::analysis;
+using namespace o2::aod::hf_cand_xictoxipipi;
+using namespace o2::constants::physics;
+using namespace o2::framework;
+
+/// Reconstruction of heavy-flavour 3-prong decay candidates
+struct HfCandidateCreatorXic {
```
vkucera, line 52: The task name should correspond to the file name.

PWGHF/TableProducer/candidateCreatorXicToXiPiPi.cxx
```diff
@@ -0,0 +1,655 @@
...
+
+#ifndef HomogeneousField
+#define HomogeneousField
+#endif
+
+#include <KFParticleBase.h>
+#include <KFParticle.h>
+#include <KFPTrack.h>
+#include <KFPVertex.h>
+#include <KFVertex.h>
```
vkucera, line 26: Why do we need to support KFParticle?

PWGHF/TableProducer/candidateCreatorXicToXiPiPi.cxx
```diff
@@ -0,0 +1,655 @@
...
+  float massXiPi1{0.};
+  double bz{0.};
+
+  using CascadesLinked = soa::Join<aod::Cascades, aod::CascDataLink>;
+  using CascFull = soa::Join<aod::CascDatas, aod::CascCovs>;
+  using KFCascadesLinked = soa::Join<aod::Cascades, aod::KFCascDataLink>;
+  using KFCascFull = soa::Join<aod::KFCascDatas, aod::KFCascCovs>;
+
+  OutputObj<TH1F> hMass3{TH1F("hMass3", "3-prong candidates;inv. mass (#Xi #pi #pi) (GeV/#it{c}^{2});entries", 500, 2.3, 2.7)};
+  OutputObj<TH1F> hCovPVXX{TH1F("hCovPVXX", "3-prong candidates;XX element of cov. matrix of prim. vtx. position (cm^{2});entries", 100, 0., 1.e-4)};
```
vkucera, line 99: Use the histogram registry

PWGHF/TableProducer/candidateCreatorXicToXiPiPi.cxx
```diff
@@ -0,0 +1,655 @@
...
+  {
+    // initialize 3-prong vertex fitter
+    o2::vertexing::DCAFitterN<3> df;
+    df.setPropagateToPCA(propagateToPCA);
+    df.setMaxR(maxR);
+    df.setMaxDZIni(maxDZIni);
+    df.setMinParamChange(minParamChange);
+    df.setMinRelChi2Change(minRelChi2Change);
+    df.setUseAbsDCA(useAbsDCA);
+    df.setWeightedFinalPCA(useWeightedFinalPCA);
```
vkucera, line 145: Move this to `init`.

PWGHF/TableProducer/candidateCreatorXicToXiPiPi.cxx
```diff
@@ -0,0 +1,655 @@
...
+  Configurable<bool> constrainXicPlusToPv{"constrainXicPlusToPv", false, "Constrain XicPlus to PV"};
+  Configurable<int> kfConstructMethod{"kfConstructMethod", 0, "Construct method of XicPlus: 0 no mass constraint, 2 mass constraint"};
+  Configurable<bool> rejDiffCollTrack{"rejDiffCollTrack", true, "Reject tracks coming from different collisions (effective only for KFParticle w/o derived data)"};
+
+  Service<o2::ccdb::BasicCCDBManager> ccdb;
+  o2::base::MatLayerCylSet* lut;
+  o2::base::Propagator::MatCorrType matCorr = o2::base::Propagator::MatCorrType::USEMatCorrLUT;
+
+  double massXiMinusFromPdg = MassXiMinus;
+  double massPionFromPdg = MassPiPlus;
```
vkucera, line 85: Remove

PWGHF/TableProducer/candidateCreatorXicToXiPiPi.cxx
```diff
@@ -0,0 +1,655 @@
...
+    std::vector<int> arrDaughIndex;
+    std::array<int, 2> arrPDGDaugh;
+
+    std::array<int, 2> arrXiResonance = {3324, kPiPlus}; // 3324: Ξ(1530)
+    int pdgCodeXicPlus = Pdg::kXiCPlus;                  // 4232
+    int pdgCodeXiMinus = kXiMinus;                       // 3312
+    int pdgCodeLambda = kLambda0;                        // 3122
+    int pdgCodePiPlus = kPiPlus;                         // 211
+    int pdgCodePiMinus = kPiMinus;                       // -211
+    int pdgCodeProton = kProton;                         // 2212
```
vkucera, line 534: Remove

PWGHF/TableProducer/candidateSelectorXicToXiPiPi.cxx
```diff
@@ -0,0 +1,247 @@
...
+using namespace o2::aod;
+using namespace o2::framework;
+using namespace o2::analysis;
+
+struct HfCandidateSelectorXicToXiPiPi {
+  Produces<aod::HfSelXicToXiPiPi> hfSelXicToXiPiPiCandidate;
+
+  Configurable<double> ptCandMin{"ptCandMin", 0., "Lower bound of candidate pT"};
+  Configurable<double> ptCandMax{"ptCandMax", 36., "Upper bound of candidate pT"};
+  // topological cuts
```
vkucera, line 37: Use `float` instead of `double`.

PWGHF/TableProducer/candidateSelectorXicToXiPiPi.cxx
```diff
@@ -0,0 +1,247 @@
...
+  /// \param pidTrackPi0   PID status of trackPi0 (prong1 of Xic candidate)
+  /// \param pidTrackPi1   PID status of trackPi1 (prong2 of Xic candidate)
+  /// \param pidTrackPr    PID status of trackPr (positive daughter of V0 candidate)
+  /// \param pidTrackPiLam PID status of trackPiLam (negative daughter of V0 candidate)
+  /// \param pidTrackPiXi  PID status of trackPiXi (Bachelor of cascade candidate)
+  /// \param acceptPIDNotApplicable switch to accept Status::NotApplicable
+  /// \return true if prong1 and prong 2 of Xic candidate pass all selections
+  template <typename T1, typename T2, typename T3, typename T4, typename T5, typename T6 = bool>
+  bool selectionPid(const T1& pidTrackPi0, const T2& pidTrackPi1, const T3& pidTrackPr, const T4& pidTrackPiLam, const T5& pidTrackPiXi, const T6& acceptPIDNotApplicable)
+  {
```
vkucera, line 172: Why is this templated? You always call the method with the same types. Also there is no benefit passing by reference.

PWGHF/TableProducer/candidateSelectorXicToXiPiPi.cxx
```diff
@@ -0,0 +1,247 @@
...
+        auto trackPiFromXi = hfCandXic.bachelor_as<TracksPidWithSel>();
+        // assign proton and pion hypothesis to V0 daughters
+        auto trackPr = trackV0PosDau;
+        auto trackPiFromLam = trackV0NegDau;
+        if (hfCandXic.sign() < 0) {
+          trackPr = trackV0NegDau;
+          trackPiFromLam = trackV0PosDau;
+        }
+        // PID info
+        int pidTrackPi0 = selectorPion.statusTpcAndTof(trackPi0);
```
vkucera, line 224:
```suggestion
        TrackSelectorPID::Status pidTrackPi0 = selectorPion.statusTpcAndTof(trackPi0);
```

PWGHF/TableProducer/treeCreatorXicToXiPiPi.cxx
```diff
@@ -0,0 +1,762 @@
...
+DECLARE_SOA_COLUMN(InvMassXiPi1, invMassXiPi1, float);
+// PID daughters
+DECLARE_SOA_COLUMN(NSigTpcPi1, nSigTpcPi1, float); //! TPC Nsigma separation for prong1 with pion mass hypothesis
+DECLARE_SOA_COLUMN(NSigTofPi1, nSigTofPi1, float); //! TOF Nsigma separation for prong1 with pion mass hypothesis
+// Events
+DECLARE_SOA_COLUMN(IsEventReject, isEventReject, int); //! Event rejection flag
+DECLARE_SOA_COLUMN(RunNumber, runNumber, int);         //! Run number
+} // namespace full
+
+DECLARE_SOA_TABLE(HfCandXicLites, "AOD", "HFCANDXICLITE",
```
vkucera, line 103: These tables are already defined in `treeCreatorXicToPKPi.cxx`.

PWGHF/TableProducer/treeCreatorXicToXiPiPi.cxx
```diff
@@ -0,0 +1,762 @@
...
+  void fillEvent(const T& collision, int isEventReject, int runNumber)
+  {
+    rowCandidateFullEvents(
+      collision.bcId(),
+      collision.numContrib(),
+      collision.posX(),
+      collision.posY(),
+      collision.posZ(),
+      isEventReject, //! filled with 0
+      runNumber);    //! filled with 1
```
vkucera, line 363: Remove unused columns.

PWGHF/DataModel/CandidateReconstructionTables.h
```diff
@@ -1320,6 +1320,119 @@ DECLARE_SOA_TABLE(HfToOmegaKMCGen, "AOD", "HFTOOMEKMCGEN", //!
                   hf_cand_xic0_omegac0::FlagMcMatchGen, hf_cand_xic0_omegac0::DebugGenCharmBar, hf_cand_xic0_omegac0::DebugGenCasc, hf_cand_xic0_omegac0::DebugGenLambda,
                   hf_cand_xic0_omegac0::PtCharmBaryonGen, hf_cand_xic0_omegac0::RapidityCharmBaryonGen, hf_cand_xic0_omegac0::OriginGen, hf_cand::IdxBhadMotherPart, o2::soa::Marker<4>);
 
+// specific Xic to Xi Pi Pi candidate properties
+namespace hf_cand_xictoxipipi
```
vkucera, line 1324:
```suggestion
namespace hf_cand_xic_to_xi_pi_pi
```

PWGHF/DataModel/CandidateReconstructionTables.h
```diff
@@ -1320,6 +1320,119 @@ DECLARE_SOA_TABLE(HfToOmegaKMCGen, "AOD", "HFTOOMEKMCGEN", //!
...
+DECLARE_SOA_COLUMN(DcaPi1Xi, dcaPi1Xi, float);
+// MC matching result:
+DECLARE_SOA_COLUMN(FlagMcMatchRec, flagMcMatchRec, int8_t); // reconstruction level
+DECLARE_SOA_COLUMN(FlagMcMatchGen, flagMcMatchGen, int8_t); // generator level
+DECLARE_SOA_COLUMN(DebugMcRec, debugMcRec, int8_t);         // debug flag for mis-association reconstruction level
+DECLARE_SOA_COLUMN(DebugMcGen, debugMcGen, int8_t);
+DECLARE_SOA_COLUMN(OriginRec, originRec, int8_t);
+DECLARE_SOA_COLUMN(OriginGen, originGen, int8_t);
+// mapping of decay types
+enum DecayType { XicToXiPiPi = 0,
```
vkucera, line 1360:
```suggestion
DECLARE_SOA_COLUMN(OriginGen, originGen, int8_t);

// mapping of decay types
enum DecayType { XicToXiPiPi = 0,
```

PWGHF/D2H/Tasks/taskXicToXiPiPi.cxx
```diff
@@ -0,0 +1,356 @@
...
+using namespace o2::aod;
+using namespace o2::analysis;
+using namespace o2::framework;
+using namespace o2::framework::expressions;
+
+/// Xic analysis task
+struct HfTaskXicToXiPiPi {
+  Configurable<int> selectionFlagXic{"selectionFlagXic", 1, "Selection Flag for Xic"};
+  Configurable<double> yCandGenMax{"yCandGenMax", 0.5, "max. gen particle rapidity"};
+  Configurable<double> yCandRecoMax{"yCandRecoMax", 0.8, "max. cand. rapidity"};
```
vkucera, line 38: Use `float`

PWGHF/D2H/Tasks/taskXicToXiPiPi.cxx
```diff
@@ -0,0 +1,356 @@
...
+  }   // process
+
+  /// MC analysis and fill histograms
+  void processMc(soa::Filtered<soa::Join<aod::HfCandXic, aod::HfSelXicToXiPiPi, aod::HfCandXicMcRec>> const& candidates,
+                 soa::Join<aod::McParticles, aod::HfCandXicMcGen> const& mcParticles,
+                 aod::TracksWMc const&)
+  {
+    std::vector<int> arrDaughIndex;
+    int pdgCodeXiMinus = kXiMinus; // 3312
+    int pdgCodePiPlus = kPiPlus;   // 211
```
vkucera, line 223: Remove

PWGHF/D2H/Tasks/taskXicToXiPiPi.cxx
```diff
@@ -0,0 +1,356 @@
...
+    registry.add("hMassXiPi2", "#Xi^{#plus}_{c} candidates;inv. mass #Xi^{#mp} #pi^{#pm} (prong 2) (GeV/#it{c}^{2});#it{p}_{T} (GeV/#it{c})", {HistType::kTH2F, {axisMassXiRes, axisPt}});
+    /*
+    MC reconstructed
+    */
+    registry.add("hPtGenSig", "#Xi^{#plus}_{c} candidates (gen+rec);candidate #it{p}_{T}^{gen.} (GeV/#it{c});entries", {HistType::kTH1F, {{300, 0., 30.}}});
+    registry.add("hPtRecSig", "#Xi^{#plus}_{c} candidates (matched);candidate #it{p}_{T} (GeV/#it{c});entries", {HistType::kTH1F, {{300, 0., 30.}}});
+    registry.add("hPtRecBg", "#Xi^{#plus}_{c} candidates (unmatched);candidate #it{p}_{T} (GeV/#it{c});entries", {HistType::kTH1F, {{300, 0., 30.}}});
+    registry.add("hPtProng0RecSig", "#Xi^{#plus}_{c} candidates (matched);prong 0 (#Xi^{#mp}) #it{p}_{T} (GeV/#it{c});entries", {HistType::kTH1F, {{300, 0., 30.}}});
+    registry.add("hPtProng0RecBg", "#Xi^{#plus}_{c} candidates (unmatched);prong 0 (#Xi^{#mp}) #it{p}_{T} (GeV/#it{c});entries", {HistType::kTH1F, {{300, 0., 30.}}});
+    registry.add("hPtProng1RecSig", "#Xi^{#plus}_{c} candidates (matched);prong 1 (#pi^{#pm}) #it{p}_{T} (GeV/#it{c});entries", {HistType::kTH1F, {{300, 0., 20.}}});
```
vkucera, line 90: Consider adding MC histograms only when `processMc` is enabled.

PWGHF/Core/SelectorCuts.h
```diff
@@ -700,6 +700,185 @@ static const std::vector<std::string> labelsPt = {
...
+                                              0.4,
+                                              1e-5,
+                                              0.005,
+                                              0.005,
+                                              0.8,
+                                              0.8,
+                                              0.1,
+                                              0.1,
+                                              0.1,
+                                            }, /* 0  < pT < 1  */
```
vkucera, line 737: Make rows single lines.

vkucera (changes requested):

PWGHF/TableProducer/candidateCreatorXicToXiPiPi.cxx
```diff
@@ -0,0 +1,655 @@
...
+  OutputObj<TH1F> hCovPVXZ{TH1F("hCovPVXZ", "3-prong candidates;XZ element of cov. matrix of prim. vtx. position (cm^{2});entries", 100, -1.e-4, 1.e-4)};
+  OutputObj<TH1F> hCovSVXZ{TH1F("hCovSVXZ", "3-prong candidates;XZ element of cov. matrix of sec. vtx. position (cm^{2});entries", 100, -1.e-4, 0.2)};
+  OutputObj<TH1F> hCovPVZZ{TH1F("hCovPVZZ", "3-prong candidates;ZZ element of cov. matrix of prim. vtx. position (cm^{2});entries", 100, 0., 1.e-4)};
+  OutputObj<TH1F> hCovSVZZ{TH1F("hCovSVZZ", "3-prong candidates;ZZ element of cov. matrix of sec. vtx. position (cm^{2});entries", 100, 0., 0.2)};
+  OutputObj<TH2F> hDcaXYProngs{TH2F("hDcaXYProngs", "DCAxy of 3-prong candidates;#it{p}_{T} (GeV/#it{c};#it{d}_{xy}) (#mum);entries", 100, 0., 20., 200, -500., 500.)};
+  OutputObj<TH2F> hDcaZProngs{TH2F("hDcaZProngs", "DCAz of 3-prong candidates;#it{p}_{T} (GeV/#it{c};#it{d}_{z}) (#mum);entries", 100, 0., 20., 200, -500., 500.)};
+  OutputObj<TH1F> hVertexerType{TH1F("hVertexerType", "Use KF or DCAFitterN;Vertexer type;entries", 2, -0.5, 1.5)}; // See o2::aod::hf_cand::VertexerType
```
mfaggin, line 109: Consired to change from `TH1F`, `TH2F` to `TH1D`, `TH2D`. In this way you avoid facing floating-point precision issues (with the 24-bit mantissa of a `float` you are sensitive only up to `std::pow(2, 24)-1 = 16777215`, and this is your minimum precision for larger numbers).

PWGHF/TableProducer/candidateCreatorXicToXiPiPi.cxx
```diff
@@ -0,0 +1,655 @@
...
+    }
+    if ((doprocessXicplusWithDcaFitter == 1) && fillHistograms) {
+      hVertexerType->Fill(aod::hf_cand::VertexerType::DCAFitter);
+    }
+    if ((doprocessXicplusWithKFParticleFromDerivedData == 1) && fillHistograms) {
+      hVertexerType->Fill(aod::hf_cand::VertexerType::KfParticle);
+    }
+  }
+
+  void processXicplusWithDcaFitter(aod::Collisions const&,
```
mfaggin, line 130:
```suggestion
  void processXicPlusWithDcaFitter(aod::Collisions const&,
```
here and everywhere else

PWGHF/TableProducer/candidateCreatorXicToXiPiPi.cxx
```diff
@@ -0,0 +1,655 @@
...
+                       /*cascade specific columns*/
+                       vertexCasc[0], vertexCasc[1], vertexCasc[2],
+                       vertexV0[0], vertexV0[1], vertexV0[2],
+                       cpaXi, cpaXYXi, cpaLambda, cpaXYLambda,
+                       massXiPi0, massXiPi1);
+    } // loop over track triplets
+  }
+  PROCESS_SWITCH(HfCandidateCreatorXic, processXicplusWithDcaFitter, "Run candidate creator with DCAFitter.", true);
+
+  void processXicplusWithKFParticleFromDerivedData(aod::Collisions const&,
```
mfaggin, line 320:
```suggestion
  void processXicPlusWithKFParticleFromDerivedData(aod::Collisions const&,
```
and everywhere else

PWGHF/TableProducer/candidateCreatorXicToXiPiPi.cxx
```diff
@@ -0,0 +1,655 @@
...
+      std::array<float, 3> vertexCasc = {casc.x(), casc.y(), casc.z()};
+      std::array<float, 3> pVecCasc = {casc.px(), casc.py(), casc.pz()};
+      std::array<float, 21> covCasc = {0.};
+
+      //----------------create cascade track------------------------------------------------------------
+      constexpr int MomInd[6] = {9, 13, 14, 18, 19, 20}; // cov matrix elements for momentum component
+      for (int i = 0; i < 6; i++) {
+        covCasc[MomInd[i]] = casc.momentumCovMat()[i];
+        covCasc[i] = casc.positionCovMat()[i];
+      }
```
mfaggin, line 195: For my understanding:
 - the 1st line sets the elements with indices `9, 13, 14, 18, 19, 20`
 - the 2nd line sets the elements with indices `0, 1, 2, 3, 4, 5`
 
If I am not overlooking anything, it means that elements with indices `6, 7, 8, 10, 11, 12, 15, 16, 17, 21` are left to 0. Is it what you want?

PWGHF/TableProducer/candidateCreatorXicToXiPiPi.cxx
```diff
@@ -0,0 +1,655 @@
...
+
+      // get SV properties
+      const auto& secondaryVertex = df.getPCACandidate();
+      auto chi2SV = df.getChi2AtPCACandidate();
+      auto covMatrixSV = df.calcPCACovMatrixFlat();
+
+      // get track momenta
+      trackParCovCharmBachelor0 = df.getTrack(1);
+      trackParCovCharmBachelor1 = df.getTrack(2);
+      trackCasc = df.getTrack(2);
```
mfaggin, line 229: should not be this `df.getTrack(0)`?

PWGHF/TableProducer/candidateCreatorXicToXiPiPi.cxx
```diff
@@ -0,0 +1,655 @@
...
+      trackCasc.propagateToDCA(primaryVertex, bz, &impactParameterCasc);
+      trackParCovCharmBachelor0.propagateToDCA(primaryVertex, bz, &impactParameter0);
+      trackParCovCharmBachelor1.propagateToDCA(primaryVertex, bz, &impactParameter1);
+
+      // calculate cosine of pointing angle
+      std::array<float, 3> pvCoord = {collision.posX(), collision.posY(), collision.posZ()};
+      double cpaLambda = casc.v0cosPA(collision.posX(), collision.posY(), collision.posZ());
+      double cpaXYLambda = RecoDecay::cpaXY(pvCoord, vertexV0, pVecV0);
+      double cpaXi = casc.casccosPA(collision.posX(), collision.posY(), collision.posZ());
+      double cpaXYXi = RecoDecay::cpaXY(pvCoord, vertexCasc, pVecCasc);
```
mfaggin, line 258: not a strong opinion for this, just to tell you that `RecoDecay` implements also a `cpa` function (see here: https://github.com/AliceO2Group/O2Physics/blob/b06b9c01a6df7386eff1b0eb0a98ac643464996b/Common/Core/RecoDecay.h#L223C1-L241C4)

PWGHF/TableProducer/candidateCreatorXicToXiPiPi.cxx
```diff
@@ -0,0 +1,655 @@
...
+
+      // get invariant mass of Xi-pi pairs
+      auto arrayMomentaXiPi0 = std::array{pVecXi, pVecPi0};
+      massXiPi0 = RecoDecay::m(std::move(arrayMomentaXiPi0), std::array{massXiMinusFromPdg, massPionFromPdg});
+      auto arrayMomentaXiPi1 = std::array{pVecXi, pVecPi1};
+      massXiPi1 = RecoDecay::m(std::move(arrayMomentaXiPi1), std::array{massXiMinusFromPdg, massPionFromPdg});
+
+      // get uncertainty of the decay length
+      double phi, theta;
```
mfaggin, line 267: where are they defined?

PWGHF/TableProducer/candidateSelectorXicToXiPiPi.cxx
```diff
@@ -0,0 +1,247 @@
...
+
+#include "Framework/AnalysisTask.h"
+#include "Framework/runDataProcessing.h"
+
+#include "Common/Core/TrackSelectorPID.h"
+
+#include "PWGHF/Utils/utilsAnalysis.h" // findBin function
+#include "PWGHF/Core/SelectorCuts.h"
+#include "PWGHF/DataModel/CandidateReconstructionTables.h"
+#include "PWGHF/DataModel/CandidateSelectionTables.h"
```
mfaggin, line 25: try to put them in alphabetic order

PWGHF/TableProducer/candidateSelectorXicToXiPiPi.cxx
```diff
@@ -0,0 +1,247 @@
...
+
+    // candidate maximum decay length
+    if (hfCandXic.decayLength() > cuts->get(pTBin, "max decay length")) {
+      return false;
+    }
+
+    // candidate maximum decay length XY
+    if (hfCandXic.decayLengthXY() > cuts->get(pTBin, "max decay length XY")) {
+      return false;
+    }
```
mfaggin, line 138: 2 questions here:
 - is this upper cut intended to suppress the beauty component?
 - you have a 50 micron cut on this variable. Are you sure that this is not too tight for a secondary vertex built with from a cascade?

PWGHF/DataModel/CandidateSelectionTables.h
```diff
@@ -301,16 +301,25 @@ DECLARE_SOA_TABLE(HfSelChicToJpsiGamma, "AOD", "HFSELCHIC", //!
...
 } // namespace hf_sel_candidate_xic
 
 DECLARE_SOA_TABLE(HfSelXicToPKPi, "AOD", "HFSELXIC", //!
                   hf_sel_candidate_xic::IsSelXicToPKPi, hf_sel_candidate_xic::IsSelXicToPiKP);
 DECLARE_SOA_TABLE(HfMlXicToPKPi, "AOD", "HFMLXIC", //!
                   hf_sel_candidate_xic::MlProbXicToPKPi, hf_sel_candidate_xic::MlProbXicToPiKP);
+// XicPlus to Xi Pi Pi
+DECLARE_SOA_TABLE(HfSelXicToXiPiPi, "AOD", "HFSELXICTOXI2PI", //!
+                  hf_sel_candidate_xic::IsSelXicToXiPiPi);
+DECLARE_SOA_TABLE(HfMlXicToXiPiPi, "AOD", "HFMLXICTOXIPIPI", //!
```
mfaggin, line 321: this table here is not used anywhere at the moment, right? I do not see it in the candidate selector, where you are just filling `HfSelXicToXiPiPi`

PWGHF/TableProducer/treeCreatorXicToXiPiPi.cxx
```diff
@@ -0,0 +1,762 @@
...
+DECLARE_SOA_COLUMN(NSigTofPi1, nSigTofPi1, float); //! TOF Nsigma separation for prong1 with pion mass hypothesis
+// Events
+DECLARE_SOA_COLUMN(IsEventReject, isEventReject, int); //! Event rejection flag
+DECLARE_SOA_COLUMN(RunNumber, runNumber, int);         //! Run number
+} // namespace full
+
+DECLARE_SOA_TABLE(HfCandXicLites, "AOD", "HFCANDXICLITE",
+                  full::XiId,
+                  full::Pi0Id,
+                  full::Pi1Id,
```
mfaggin, line 106: Just a warning: if you goal is to create self-contained derived data then I think there are no problems. But beware that in case you just want slim derived data (e.g. a tree to train your ML model) then with `INDEX_COLUMNS` you will have issues with unbound indices in Hyperloop.

mfaggin (changes requested):
thanks a lot for the developments, I have some commets/suggestions

PWGHF/TableProducer/candidateCreatorXicToXiPiPi.cxx
pstahlhu: Yes. In the CascCovs table only the independent covariance matrices of the cascade decay vertex position and the cascade momenta are stored. The indices you mentioned are the ones that point to the covariance matrix elements containing the covariances between momenta and vertex coordinates (lower triangular form), which are not stored for the DCA fitter.
This is done analogously to the XicZero/OmegaCZero workflow.
As a remark, for the KF case, the full covariance matrix is stored.

PWGHF/TableProducer/candidateCreatorXicToXiPiPi.cxx
pstahlhu: Yes. These are dynamic columns that use the RecoDecay::cpa functions. I chose the dynamic columns out of convenience, as I don't have to pass as many arguments  in this case, the rest of them are already handled.

PWGHF/TableProducer/treeCreatorXicToXiPiPi.cxx
pstahlhu: Yes, similar tables are defined already there, although they are not identical to the ones I have currently.
One can probably converge on one format for the event (FullEvs) and generated MC-particle tables (FullPs).
However, I think that this is not possible for the candidate tables since in the Xic to Xi Pi Pi decay channel we need some cascade-specific variables, sometimes even KF-specific.

PWGHF/DataModel/CandidateSelectionTables.h
pstahlhu: Yes, I initially added the ML selection already, but decided to take it out since it is not needed without any trained models at the moment. Apparently I forgot to remove this before creating the PR.

PWGHF/TableProducer/candidateCreatorXicToXiPiPi.cxx
pstahlhu: In the Run 2 analysis we saw that the reconstruction of the XicPlus can be significantly improved with the KFParticle using topological and mass constraints. In addition to that, we also have the full covariance matrix of all the particles available throughout the reconstruction workflow such that the treatment of the uncertainties might be overall better.

For the moment, we added it in order to test if we can benefit from it in the Run 3 analysis as well. If not (significantly), we will probably transition fully to the DCAFitter.

PWGHF/TableProducer/candidateCreatorXicToXiPiPi.cxx
pstahlhu: Do you mean the variables to store the reconstructed masses? At the beginning of the struct, outside of the process functions.

PWGHF/TableProducer/candidateSelectorXicToXiPiPi.cxx
pstahlhu: To be honest, I just started from the Bs selector that I wrote as well and didn't think about it too much. If I didn't overlook something, the XicPlus has feeddown contributions only from the OmegaBMinus. Do you think this cut is useful in any way?
When I tested my code locally, I also set different values for this cut since it basically kills all the candidates as you mentioned. I modified the default values in the SelectorCuts.h accordingly.

PWGHF/TableProducer/treeCreatorXicToXiPiPi.cxx
pstahlhu: My goal was to compare different configurations (e.g. KF vs. DCA, KFs with different combination of constraints) in order to check how to optimize the reconstruction of the XicPlus. I added the indices to ensure that I would then compare the identical sample of XicPlus candidates.

1) Do you thin this would work? Or are there also problems with unbound indices?
2) Is this a problem also for running on locally stored derived data?
2)  Can I store the indices or do something similar without this problem of unbound indices? If yes, how?

pstahlhu:
Hi @vkucera and  @mfaggin,
thank you very much for your comments!! I implemented most of your requested changes now. 
For some I have some comments or questions myself, as you will see above...

vkucera:
@pstahlhu , thanks for the updates. Please do not mark suggestions as resolved. The reviewers have to check their implementation and only then they can approve that the conversation has been resolved.

PWGHF/TableProducer/treeCreatorXicToXiPiPi.cxx
vkucera: My point is that your tables are similar as those but have exactly the same names, which creates an ambiguity in the data model.

PWGHF/TableProducer/candidateCreatorXicToXiPiPi.cxx
vkucera: `Xicplus` was correct

PWGHF/TableProducer/candidateCreatorXicToXiPiPi.cxx
```diff
@@ -0,0 +1,655 @@
...
+    ccdb->setURL(ccdbUrl);
+    ccdb->setCaching(true);
+    ccdb->setLocalObjectValidityChecking();
+    lut = o2::base::MatLayerCylSet::rectifyPtrFromFile(ccdb->get<o2::base::MatLayerCylSet>(ccdbPathLut));
+    runNumber = 0;
+
+    // fill hVertexerType histogram
+    if ((doprocessXicPlusWithDcaFitter + doprocessXicPlusWithKFParticleFromDerivedData) != 1) {
+      LOGP(fatal, "Only one process function can be enabled at a time.");
+    }
```
vkucera, line 128: This should be done as the first thing in `init`.

PWGHF/TableProducer/treeCreatorXicToXiPiPi.cxx
```diff
@@ -0,0 +1,758 @@
...
+  {
+    rowCandidateFullEvents(
+      collision.bcId(),
+      collision.numContrib(),
+      collision.posX(),
+      collision.posY(),
+      collision.posZ());
+  }
+
+  template <bool doMc = false, bool noKf = true, typename T>
```
vkucera, line 362:
```suggestion
  template <bool doMc, bool doKf, typename T>
```
It's more transparent in the code when template arguments are always passed.
`noKf` is a negation of `Kf`, which you don't use here, so it's better to keep it simple.

PWGHF/TableProducer/candidateCreatorXicToXiPiPi.cxx
```diff
@@ -0,0 +1,655 @@
...
+                       /*cascade specific columns*/
+                       vertexCasc[0], vertexCasc[1], vertexCasc[2],
+                       vertexV0[0], vertexV0[1], vertexV0[2],
+                       cpaXi, cpaXYXi, cpaLambda, cpaXYLambda,
+                       massXiPi0, massXiPi1);
+    } // loop over track triplets
+  }
+  PROCESS_SWITCH(HfCandidateCreatorXicToXiPiPi, processXicPlusWithDcaFitter, "Run candidate creator with DCAFitter.", true);
+
+  void processXicPlusWithKFParticleFromDerivedData(aod::Collisions const&,
```
vkucera, line 326: Why do you need the suffix `FromDerivedData`? There is not alternative without it.

PWGHF/TableProducer/treeCreatorXicToXiPiPi.cxx
pstahlhu: Ok, I see. I will rename them.

PWGHF/TableProducer/candidateCreatorXicToXiPiPi.cxx
pstahlhu: I had another process function initially which did the track-index skimming and then built the candidates directly. I dediced not to remove the suffix to make clear that this process function (or more precisely, the KF version of the workflow) works only when one runs on derived data, otherwise the trackIndexSkimCreatorLfCascades will run over empty tables (CascCores+ CascCovs which are not filled before, instead of KFCascCores+KFCascCovs) and no output will be produced.
If you want, I can remove it.

pstahlhu:
Hi @vkucera, thanks for the additional comments, I will update the PR later. 
And sorry for marking the previous suggestions as resolved, I wanted to indicate which suggestions I implemented, but of course it makes much more sense that you mark them as resolved once you are happy with them.

PWGHF/TableProducer/treeCreatorXicToXiPiPi.cxx
vkucera: Still there

PWGHF/TableProducer/candidateCreatorXicToXiPiPi.cxx
vkucera: Yes please

PWGHF/TableProducer/candidateCreatorXicToXiPiPi.cxx
mfaggin: my bad, I was overlooking that their definition is in `getPointDirection`

PWGHF/TableProducer/treeCreatorXicToXiPiPi.cxx
mfaggin: If you want to be flexible and have the possibility to both include or esclude the index columns, maybe you can do a separate table only with these columns, to be filled only on demand at runtime (i.e. ruling it with a dedicated `Configurable`)

pstahlhu:
Hi @vkucera and @mfaggin, I have implemented your comments and updated the PR once again. Could you have another look? Many thanks in advance!!

vkucera:
> Hi @vkucera and @mfaggin, I have implemented your comments and updated the PR once again. Could you have another look? Many thanks in advance!!

Thanks @pstahlhu , I will have a look, but please do not force-push your branch when your PR is under review. It overwrites the commit history. See more details in the documentation.

mfaggin (dismissed):

PWGHF/TableProducer/candidateSelectorXicToXiPiPi.cxx
vkucera: Please propagate this change also to the PID configurables.

PWGHF/DataModel/CandidateReconstructionTables.h
```diff
@@ -1320,6 +1320,120 @@ DECLARE_SOA_TABLE(HfToOmegaKMCGen, "AOD", "HFTOOMEKMCGEN", //!
...
+DECLARE_SOA_COLUMN(FlagMcMatchRec, flagMcMatchRec, int8_t); // reconstruction level
+DECLARE_SOA_COLUMN(FlagMcMatchGen, flagMcMatchGen, int8_t); // generator level
+DECLARE_SOA_COLUMN(DebugMcRec, debugMcRec, int8_t);         // debug flag for mis-association reconstruction level
+DECLARE_SOA_COLUMN(DebugMcGen, debugMcGen, int8_t);
+DECLARE_SOA_COLUMN(OriginRec, originRec, int8_t);
+DECLARE_SOA_COLUMN(OriginGen, originGen, int8_t);
+
+// mapping of decay types
+enum DecayType { XicToXiPiPi = 0,
+                 XicToXiResPiToXiPiPi,
```
vkucera, line 1362: Please add comments on the meaning of each value.

PWGHF/TableProducer/candidateCreatorXicToXiPiPi.cxx
```diff
@@ -0,0 +1,656 @@
...
+      registry.add("hCovSVXZ", "3-prong candidates;XZ element of cov. matrix of sec. vtx. position (cm^{2});entries", {HistType::kTH1D, {{100, -1.e-4, 0.2}}});
+      registry.add("hCovPVZZ", "3-prong candidates;ZZ element of cov. matrix of prim. vtx. position (cm^{2});entries", {HistType::kTH1D, {{100, 0., 1.e-4}}});
+      registry.add("hCovSVZZ", "3-prong candidates;ZZ element of cov. matrix of sec. vtx. position (cm^{2});entries", {HistType::kTH1D, {{100, 0., 0.2}}});
+      registry.add("hVertexerType", "Use KF or DCAFitterN;Vertexer type;entries", {HistType::kTH1F, {{2, -0.5, 1.5}}}); // See o2::aod::hf_cand::VertexerType
+      registry.add("hDcaXYProngs", "DCAxy of 3-prong candidates;#it{p}_{T} (GeV/#it{c};#it{d}_{xy}) (#mum);entries", {HistType::kTH2D, {{100, 0., 20.}, {200, -500., 500.}}});
+      registry.add("hDcaZProngs", "DCAz of 3-prong candidates;#it{p}_{T} (GeV/#it{c};#it{d}_{z}) (#mum);entries", {HistType::kTH2D, {{100, 0., 20.}, {200, -500., 500.}}});
+    }
+
+    // fill hVertexerType histogram
+    if ((doprocessXicplusWithDcaFitter == 1) && fillHistograms) {
```
vkucera, line 123:
```suggestion
    if (doprocessXicplusWithDcaFitter && fillHistograms) {
```

PWGHF/TableProducer/candidateCreatorXicToXiPiPi.cxx
```diff
@@ -0,0 +1,656 @@
...
+      registry.fill(HIST("hVertexerType"), aod::hf_cand::VertexerType::DCAFitter);
+    }
+    if ((doprocessXicplusWithKFParticle == 1) && fillHistograms) {
```
vkucera, line 126:
```suggestion
    if (doprocessXicplusWithKFParticle && fillHistograms) {
```

PWGHF/TableProducer/candidateSelectorXicToXiPiPi.cxx
```diff
@@ -0,0 +1,252 @@
...
+  // TPC PID
+  Configurable<double> ptPidTpcMin{"ptPidTpcMin", 0.15, "Lower bound of track pT for TPC PID"};
+  Configurable<double> ptPidTpcMax{"ptPidTpcMax", 20., "Upper bound of track pT for TPC PID"};
+  Configurable<double> nSigmaTpcMax{"nSigmaTpcMax", 5., "Nsigma cut on TPC only"};
+  Configurable<double> nSigmaTpcCombinedMax{"nSigmaTpcCombinedMax", 5., "Nsigma cut on TPC combined with TOF"};
+  // TOF PID
+  Configurable<double> ptPidTofMin{"ptPidTofMin", 0.15, "Lower bound of track pT for TOF PID"};
+  Configurable<double> ptPidTofMax{"ptPidTofMax", 20., "Upper bound of track pT for TOF PID"};
+  Configurable<double> nSigmaTofMax{"nSigmaTofMax", 5., "Nsigma cut on TOF only"};
+  Configurable<double> nSigmaTofCombinedMax{"nSigmaTofCombinedMax", 5., "Nsigma cut on TOF combined with TPC"};
```
vkucera, line 54: Please use `float`.

vkucera:
> > Hi @vkucera and @mfaggin, I have implemented your comments and updated the PR once again. Could you have another look? Many thanks in advance!!
> 
> Thanks @pstahlhu , I will have a look, but please do not force-push your branch when your PR is under review. It overwrites the commit history. See more details in the documentation.

The implementation of comments looks fine to me, thanks. I just have a few remaining comments. Then it is ready for merging for me.

pstahlhu:
> > > Hi @vkucera and @mfaggin, I have implemented your comments and updated the PR once again. Could you have another look? Many thanks in advance!!
> > 
> > 
> > Thanks @pstahlhu , I will have a look, but please do not force-push your branch when your PR is under review. It overwrites the commit history. See more details in the documentation.
> 
> The implementation of comments looks fine to me, thanks. I just have a few remaining comments. Then it is ready for merging for me.

@vkucera thanks for your review. I also implemented your remaining comments now.
Sorry for any inconvenience regarding the force-push, I will avoid it in future.

vkucera (approved):

PR #6625, opened 2024-06-24
PWG-UD: Updates of  DGCutparHolder, DGPIDSelector.cxx, and DGCandProducer.cxx

pbuehler:
. Changed default values of CollisionSel
. Added setters in DGAnaparHolder
. Corrected bin-labels of histogram reco/Stat

### DIFF
--- a/PWGUD/Core/DGCutparHolder.h
+++ b/PWGUD/Core/DGCutparHolder.h
@@ -39,7 +39,7 @@ class DGCutparHolder
                  bool TOR = true,
                  float maxFITtime = 4,
                  std::vector<float> FITAmpLimits = {0., 0., 0., 0., 0.},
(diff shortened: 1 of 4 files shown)


rolavick (commented):
Do all the new setters following the Camel formatting style? Is it consistent with the rest of the code? For example compare getter cutPt() and the new setter setpt().

Whywas the Zvtx to PV matching cut released from default?

pbuehler:
The ZvtxFT0-to-PV matching has to be removed for the selection of DG events, where no FT0 is information is available.

The consistency of the method naming is indeed questionable. With the setters I am following the names used in DGCutparHolder and SGCutParHolder. I'm afraid we are currently not following a strict rule. Let's keep it for the moment.

rolavick (approved):

PR #6638, opened 2024-06-25
PWGHF: Move to ConfigurableGroup

ktf:


fcatalan92:
Hi @ktf, is there a particular reason for these changes? Are they transparent for what concern the configuration or we would need to update the related wagons on HY?

ktf:
This code is currently working by accident, since there is a limit of 100 elements per task which is not enforced.

The PR which enforces the limit (https://github.com/AliceO2Group/AliceO2/pull/12704) cannot be merged until this is fixed with the risk that more things go unnoticed.

The configuration is unchanged.

fcatalan92 (approved):

PR #6640, opened 2024-06-25
Use BasicCCDBManger::getRunDuration for better stability

mpuccio:
This avoids issues with runs with missing EOR information, see for instance 551230 (before it gets fixed)


ddobrigk (approved):
thanks a lot!

PR #6662, opened 2024-06-26
fix cluster size calculation + histogram filling

fmazzasc:


alibuild (approved):

PR #6577, opened 2024-06-18
[Event plane] add vectors columns for multiple event planes

jikim1290:
add vectors columns for multiple event planes
Q-vector from TPC full tracks

PR #6666, opened 2024-06-26
Common/TableProducer/qVectorTable.cxx: bug fix to support vector of Qvector

dsekihat:
This PR is to support vector of Qvector (https://github.com/AliceO2Group/O2Physics/pull/6577). In the previous version, only aod::QvectorXXXXs (FT0C, etc) or aod::Qvectors were accepted by comparing string. But, aod::QvectorXXXXVecs were not supported, and q vectors were -999.
In addition, suppress warning about comparison between int and size_t.

### DIFF
--- a/Common/TableProducer/qVectorsTable.cxx
+++ b/Common/TableProducer/qVectorsTable.cxx
@@ -156,7 +156,8 @@ struct qVectorsTable {
           goto allDetectorsInUse; // Added to break from nested loop if all detectors are in use.
         }
         for (auto det : useDetector) {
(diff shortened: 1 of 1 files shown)


dsekihat:
Dear @ddobrigk , @strogolo , @jikim1290 , could you check this PR, please? Thank you.

jikim1290:
@dsekihat Thank you for spotting this. It is indeed needed to call QvecXXX and QvecXXXVecs at the same time.

ddobrigk (approved):

PR #6677, opened 2024-06-27
[Centrality] add metadata information

njacazio:


njacazio:
@ddobrigk @strogolo

strogolo:
Hi @njacazio,
is this PR following the discussion we had on the metadata for the apass?
I see that in this PR the task produce a FATAL in case the calibration for that run and apass is not available. For me this option is fine but I think we should involve also @ekryshen @chiarazampolli @noferini @alcaliva, so that they are also aware of this change.

alcaliva:
Hi Stefano, 
Thanks! I was aware of this initiative, which I support very strongly

njacazio:
Hi, indeed there's the possibility to override the fatal and get the default values of the calibration, take the objects from file, override the metadata. In short this gives maximum freedom. Of course, we should test it with the new (metadata tagged) ccdb objects.

njacazio:
If metadata is available then this can also be merged @ddobrigk

ddobrigk (approved):

jgrosseo:
I think this broke RUN2 analysis. See for example here:
https://alimonitor.cern.ch/train-workdir/tests/0025/00256615/stdout.log

jgrosseo:
Please advise if it can be easily fixed or we revert. @njacazio @ddobrigk

ddobrigk:
@jgrosseo this should be fine now, it was a configuration matter only. Please write back in case something is still not ok! Thanks!

PR #6689, opened 2024-06-28
PWGJE: Fixed bugs and added histograms in PhotonIsolationQA.cxx

berend01:
The track_matching function has been improved: the tracks matched to the cluster are no longer hard-coded. Instead, the emcalcorrectiontask now fully handles track matching.
A bug in calculating the delta phi between tracks and clusters has been fixed.
A multidimensional histogram has been added for further analysis.

PWGJE: Fixed bugs and added histograms in PhotonIsolationQA.cxx

### DIFF
--- a/PWGJE/Tasks/PhotonIsolationQA.cxx
+++ b/PWGJE/Tasks/PhotonIsolationQA.cxx
@@ -1,4 +1,3 @@
-
 // Copyright 2019-2020 CERN and copyright holders of ALICE O2.
 // See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
 // All rights not expressly granted are reserved.


PWGJE/Tasks/PhotonIsolationQA.cxx
```diff
@@ -168,17 +174,9 @@ struct PhotonIsolationQA {
...
-      if (dPhi < 0) {
-        dPhi = 2. * M_PI - dPhi;
-      }
-      double distance = sqrt(pow(dEta, 2) + pow(dPhi, 2));
-      if (distance <= TrackMatchingRadius) {
-        double abs_p = abs(match.track_as<myGlobTracks>().p());
-        if ((cluster.energy() / abs_p) < 1.75) {
-          return true;
-        }
+      double abs_pt = abs(match.track_as<myGlobTracks>().pt());
```
fjonasALICE, line 177: I do not suggest to get rid of your track matching requirment in the task itself. What the EMCal track matching in the correction framework ist intended for is to give you the N closest tracks to a given cluster (which is done in a very efficienct way). This means that in your analysis task you do not have to each time loop over all tracks of the event but only those in the vicinity. So in this case, while you are now only looking at the tracks closest to the cluster, you should still make some cut on dEta and dPhi.

PWGJE/Tasks/PhotonIsolationQA.cxx
```diff
@@ -302,23 +301,23 @@ struct PhotonIsolationQA {
...
         // acces mc true info
         auto ClusterParticles = mccluster.mcParticle_as<aod::StoredMcParticles_001>();
         bool background = true;
         for (auto& clusterparticle : ClusterParticles) {
           if (clusterparticle.pdgCode() == 22) {
             MC_Info.fill(HIST("hMcParticlesToCluster"), clusterparticle.e());
+            MC_Info.fill(HIST("hMcParticleProcessCode"), clusterparticle.getProcess());
             MC_Info.fill(HIST("hMcParticleStatusCode"), clusterparticle.getGenStatusCode());
-            if ((clusterparticle.getGenStatusCode() >= 21) && (clusterparticle.getGenStatusCode() <= 29))
+            if ((clusterparticle.getProcess() >= 21) && (clusterparticle.getProcess() <= 29))
```
fjonasALICE, line 314: i am not completely sure how status codes work in O2, but please make sure that both fragmentation photons AND prompt photons get accepted as signal. In addition, signal photons should also fulfill and isolation on GENERATOR level

fjonasALICE (commented):
overall nice addition! Only have some comments. From a brief look nothing looks like it could break anything, so good to go from my side. Changes I suggested can also be implemented in the future

PWGJE/Tasks/PhotonIsolationQA.cxx
berend01: The emcalCorrection task has a configurable maxMatchingDistance (max matching distance between tracks and cluster). Which is adjustable to different radii. Is that not the same as encoding a dEta and dPhi cut in the PhotonIsolationQA task. Or am i missing something here?

fjonasALICE: the idea is that in the emcal correction framework, this value should be chosen large enough so that it works for every analysis. It is simply used to improve computational efficiency. This is important, because later multiple analyses can run in the same train, but there will be only one correction framework providing the output for all of them. This is why you should choose in the correction framework a large dR and cut in the task on dEta and dPhi of that subset

mfasDa (approved):

PR #6691, opened 2024-06-28
Use C++20 concepts to determine which Centrality to return

ktf:
Avoid recursive lookup in the schema while doing so.

### DIFF
--- a/Common/DataModel/Centrality.h
+++ b/Common/DataModel/Centrality.h
@@ -54,6 +54,23 @@ using CentFT0A = CentFT0As::iterator;
 using CentFT0C = CentFT0Cs::iterator;
 using CentFDDM = CentFDDMs::iterator;
 using CentNTPV = CentNTPVs::iterator;
+
+template <typename T>
+concept HasRun2Centrality = requires(T&& t) {
+  { t.centRun2V0M() };
+  { t.centRun2CL0() };
+  { t.centRun2CL1() };
+};
+
+template <typename T>
+concept HasCentrality = requires(T&& t) {
+  { t.centFV0A() };
+  { t.centFT0M() };
+  { t.centFT0A() };
+  { t.centFT0C() };
+  { t.centNTPV() };
+};
+
 } // namespace o2::aod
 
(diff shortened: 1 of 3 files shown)


ktf:
No rush, I just wanted it as an example of a good practice C++20 usage.

victor-gonzalez:
In the queue to be tested

victor-gonzalez (approved):
Thanks @ktf ! This works great! And it is really clean!
I only have rights on the PWGCF part
But I can fix the linter errors in `Centrality.h` if so you want
Just let me know

jgrosseo:
@ktf can you fix the clang error?
If it is fine for @victor-gonzalez , we can merge it then.

ktf:
I think my code is correct, actually. The issue is that cpplint / clang-format does not seem to understand requires. Is it ok if I bump them?

jgrosseo:
> I think my code is correct, actually. The issue is that cpplint / clang-format does not seem to understand requires. Is it ok if I bump them?

This is in O2. I am not the right person to answer...

ktf:
It's actually the tests of O2Physics which I need to touch.

jgrosseo:
Yes, true. Sorry for overlooking this. I don't mind.  @vkucera ?

vkucera:
I fully support bumping `clang-format`, although one has to be careful about potential discrepancies in the config file `.clang-format` that we currently use in many `AliceO2Group` repositories, because some parameters might have changed.

For `cpplint`, the version is picked up by MegaLinter and I'm not sure that one can require a specific version.

@ktf do you know which version of each we need to support the concepts?

ktf:
clang-format was bumped to use version 18.
cpplint needs to be bumped to a new version. As it is now, it's simply not working correctly.

Merging.

vkucera:
> clang-format was bumped to use version 18. cpplint needs to be bumped to a new version. As it is now, it's simply not working correctly.
> 
> Merging.

Thanks @ktf

ktf:
(I should mention the working version of cpplint does not yet exist)

PR #6697, opened 2024-06-29
PWGJE: Adding the ability to Run over ALICE 3 simulations in the JE f…

nzardosh:
…ramework


alibuild (approved):

PR #6705, opened 2024-07-01
PWGCF: Fill QA plots after selections in flowAnalysisGF

cnkoster:
The QA plots were filled after the occupancy selections, where they should be filled after all selections. This is fixed now. 
The changes also include the comments on my last PR.


victor-gonzalez (approved):
Thanks!
I hope you also agree that now it is easier to read, understand, maintain, and evolve the task code

cnkoster:
Thank you very much for the comments @victor-gonzalez, I agree indeed!

PR #6715, opened 2024-07-02
PWGHF: taskFlow.cxx: Uniformization of the name of the correlation containers

AlexianL:
Uniformization of the names of the variables which contains correlation containers according to the following rule : 

- sameDETECTOR1DETECTOR2TriggerparticleAssociatedparticle (and the same for "mixed" instead of "same")

Others changes : 
- the names of the correlation containers (!= name of the variable which contains correlation containers) has been changed to be the same as the name of the variable which contains correlation containers, as it should not cause problems for the good running of the code
- The names of the process functions have also been changed according to the rule above
- The folder organization of the AnalysisResults.root file have also been changed to be clearer (addition of a new layer of folder to detail the detectors used)

Known issues currently : 
- Issue for a long time : MixedEvent plots are filled in TPC-TPC h-h case even when running TPC-MFT h-h case -> planned to be resolved in a future PR when more QA plots will be added

### DIFF
--- a/PWGHF/HFC/Tasks/taskFlow.cxx
+++ b/PWGHF/HFC/Tasks/taskFlow.cxx
@@ -469,7 +469,7 @@ struct HfTaskFlow {
   //    process same event correlations: h-h case
   // =====================================
   void processSameTpcTpcChCh(FilteredCollisionsWSelMult::iterator const& collision,
-                           TracksWDcaSel const& tracks)
+                             TracksWDcaSel const& tracks)
   {
     if (!(isCollisionSelected(collision, true))) {
       return;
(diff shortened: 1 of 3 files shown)


vkucera:
Hi @AlexianL , just for your information, you don't need to run `clang-format` yourself. The automatic formatting PR linked above already applies the necessary changes. You just need to merge it in your branch.

AlexianL:
> Hi @AlexianL , just for your information, you don't need to run `clang-format` yourself. The automatic formatting PR linked above already applies the necessary changes. You just need to merge it in your branch. And please don't forget to add the prefix in the PR title when it's ready for review.

Thanks for the tip !

mmazzilli (approved):

PR #6716, opened 2024-07-02
Reversed occupancy cut 

cnkoster:
Task was excluding occupancy in the range that is was supposed to select the events. 
If statement is now reversed.


victor-gonzalez (approved):

victor-gonzalez:
You should rebase your local repo
In here you are referring to commits that are already incorporated to the central repo

PR #6729, opened 2024-07-03
PWGMM: Mult: dndeta: avoid putting iterators into vector

aalkin:


alibuild (approved):

PR #6739, opened 2024-07-05
SC fix, histogram binning adjustments

jaelpark:


PWGCF/JCorran/Core/JFFlucAnalysis.cxx
```diff
@@ -127,8 +127,9 @@ TComplex JFFlucAnalysis::Four(int n1, int n2, int n3, int n4)
 #undef C
 
 //________________________________________________________________________
-void JFFlucAnalysis::UserExec(Option_t* /*popt*/)
+void JFFlucAnalysis::UserExec(Option_t* popt)
 {
+  (void)popt; // unused var error
```
victor-gonzalez, line 132: `void JFFlucAnalysis::UserExec(Option_t*)`
will be less misleading

PWGCF/JCorran/Core/JFFlucAnalysis.cxx
```diff
@@ -282,33 +283,36 @@ void JFFlucAnalysis::UserExec(Option_t* /*popt*/)
...
+  }
 }
 
 //________________________________________________________________________
-void JFFlucAnalysis::Terminate(Option_t* /*popt*/)
+void JFFlucAnalysis::Terminate(Option_t* popt)
 {
-  //
+  (void)popt; // unused var error
 }
```
victor-gonzalez, line 318: Same here

victor-gonzalez (changes requested):
Please, have a look at my suggestions

PWGCF/JCorran/Core/JFFlucAnalysis.cxx
jaelpark: There is an issue when clang-format and cpplint are used together if I remove the `popt`. cpplint does not like if `popt` is removed. For now it'd be nice to keep this `popt` there for compatibility outside O2.

victor-gonzalez: @vkucera Any clue about this clang-format cpplint clash?

vkucera: I don't understand what is the issue that you are referring to.

- `clang-format` does not care if `popt` is used or not, it just formats your code.
- `cpplint` complained about casting in 4e48e7f
- "unused var error" is reported by the compiler

So what is the problem?

vkucera: Actually, I don't understand what the line `(void)popt;` is good for in the first place.

vkucera: @jaelpark What do you mean by "compatibility outside O2"? Your task is an obvious copy-paste from AliPhysics which is already very bad to start with. O2 is completely different and trying to keep it "compatible" with some code outside is a bad approach.

jaelpark: I think the issue has been that cpplint does not accept a single pointer argument that is unnamed. For some reason it thinks it's a C-style cast. I tested a couple of ways to put it before clang-format, so clang-format does not have a role in it in the end (correction to above). The line `(void)popt` is there to simply suppress the "unused variable" error. Because I can't seem to remove `popt` and leave only `Option_t *`

vkucera: > I think the issue has been that cpplint does not accept a single pointer argument that is unnamed. For some reason it thinks it's a C-style cast. I tested a couple of ways to put it before clang-format, so clang-format does not have a role in it in the end (correction to above). The line `(void)popt` is there to simply suppress the "unused variable" error. Because I can't seem to remove `popt` and leave only `Option_t *`

cpplint thinks it's a C-style cast because it IS a C-style cast. What you want is `void JFFlucAnalysis::UserExec(Option_t* /*popt*/)` and remove `(void)popt;`.
`clang-format` is irrelevant for the `unused-variable` error.

jaelpark: To clarify, `(void)popt` is not giving any cpplint issues, it is the line where the function is defined. We had it as you wrote that at one point (`void JFFlucAnalysis::Terminate(Option_t* /*popt*/)`) but unfortunately it results in the same error from cpplint. I just tested again. This error is why it originally had to change in the first place.

vkucera: Can you please push it in this PR so that I can see the error?

vkucera: Also, where do you see the unused var error? I don't see it in the build logs.

jaelpark: I just pushed and can see the error now. I have been testing locally with cpplint. This has always been consistent with the report here if there has been an issue. The unused var error comes only if `popt` is not commented and `(void)popt` is not in place.

vkucera: Indeed, it seems to be a known (and unfortunately not fixed) bug: https://github.com/cpplint/cpplint/issues/131

vkucera: I suggest to add a linting exception:
```c++
void JFFlucAnalysis::UserExec(Option_t* /*popt*/) // NOLINT(readability/casting) false positive: https://github.com/cpplint/cpplint/issues/131
```

jaelpark: Thanks! I added this. Seems to work.

victor-gonzalez (approved):

PR #6744, opened 2024-07-05
PWGJE + PWGDQ: Adding Dielectron Jets to framework

nzardosh:
Combining DQ and JE frameworks

For the DQ part just added the option to store tables linking back to the original AOD (not to be saved for DQ only derived data)

For the JE part added all the options to absorb the dielectron table into the framework and interface it with the derived JE tables

Strategy has been discussed with Ionut


iarsene (approved):

alibuild (approved):

PR #6753, opened 2024-07-06
fixed iterator: replaced globalIndex with int variable used as counter in the particle loop

alcaliva:


vkucera:
@alcaliva , please use more descriptive titles in your commits and pull requests.

alcaliva:
Done. I replaced the globalIndex with an integer that I define outside of the for loop. This is needed as the globalIndex does not correspond to the index of the for loop and I get a seg fault when I feed this index into the IteratorAt function

fmazzasc (approved):

vkucera:
> Done. I replaced the globalIndex with an integer that I define outside of the for loop. This is needed as the globalIndex does not correspond to the index of the for loop and I get a seg fault when I feed this index into the IteratorAt function

Thanks, but it would be better if you mentioned in the title PWGLF and the name of the analysis or file that you modified. Same for the commit titles. Now your merged commit appears in the upstream as "fixed iterator (#6753)" which tells me nothing about which part of O2Physics has been modified.

alcaliva:
ok. Can I still fix the title? if not, I could do one more commit and add a more descriptive message

vkucera:
It's already merged. You cannot modify the commits in upstream/master. It's for your future commits and PRs.

PR #6754, opened 2024-07-06
Updated macro for uploads of trigger aliases to ccdb

ekryshen:


alibuild (approved):

PR #6762, opened 2024-07-08
PWGCF: FemtoUniverse -- fixed typo

Eloviyo:


victor-gonzalez:
@Eloviyo Please, don't push code that does not compile locally

PR #6764, opened 2024-07-08
PWGCF: FemtoUniverse track-V0 task -- fix for a track filter

Eloviyo:


victor-gonzalez (changes requested):
I guess this is not a fix for a bug in the `FemtoUniverse` framework so the PR title is misleading. Please, use a more appropriate one

Eloviyo:
Ok, done.

victor-gonzalez (approved):

PR #6778, opened 2024-07-09
Common: better defaults for mult calibrator to avoid troubles

ddobrigk:


alibuild (approved):

PR #6781, opened 2024-07-09

vkucera:


PWGCF/JCorran/Core/JFFlucAnalysis.cxx
```diff
@@ -301,7 +301,7 @@ void JFFlucAnalysis::UserExec(Option_t* /*popt*/) // NOLINT(readability/casting)
   }
 
   for (UInt_t ih = 2; ih < kNH; ih++) {
-    for (UInt_t ihh = 2, mm = (ih < kcNH ? ih : kcNH); ihh < mm; ihh++) {
+    for (UInt_t ihh = 2, mm = (ih < kcNH ? ih : static_cast<UInt_t>(kcNH)); ihh < mm; ihh++) {
```
victor-gonzalez, line 304: Interesting that it does not complain in the comparison `ih < kcNH`

victor-gonzalez (approved):
Thanks! @vkucera

PWGCF/JCorran/Core/JFFlucAnalysis.cxx
vkucera: I guess the comparison `ih < kcNH` does not require the types to be same, but the condition has to return one type.

alibuild (approved):

PR #6788, opened 2024-07-09
[event plane] fix bug

jikim1290:
fix bug


dsekihat:
Dear @strogolo and @ddobrigk , It was me who spotted this. @jikim1290 explained to me that BPos is measured in 0.1 < eta < 0.8, BNeg is measured in -0.8 < eta < -0.1, and  BTot is measured  in - 0.8 < eta < 0.8. But, I found that code is inconsistent with his explanation. So, can you also check which eta cut is applied to tracks during the calibration? Thank you.

ddobrigk (approved):
Looks good, thanks! 
And the conclusion of the story is: `continue` is to be used with care (it's easy to slip up with this indeed :-D )

PR #6794, opened 2024-07-09
PWGHF: fix selection of MC reco candidates in correlatorDMesonPairs

atavirag:
Working on the addition of the reflections to the 2D inv. mass fit, I noticed that the reflection template had the wrong shape, so I changed the selection of MC reco candidates to match how it's done in task D0.

Cheers!


atavirag:
Hello! Sorry to insist, but would it be possible to review this PR?
Thank you!

mmazzilli (approved):

PR #6802, opened 2024-07-10
[PWGDQ] use float type for statistic histograms

feisenhu:
I ran into multiple occasions where the limit of an integer is not enough any more.
So I changed TH*I to TH*F


vkucera:
@feisenhu TH1F actually gives you even smaller range of integers than TH1I, because TH1I uses 32 bits for the integer bin content while TH1F uses `float` which has 23 bits for the mantissa, so once your bin content reaches 2^32 - 1, you can no longer increase it by 1. You might consider TH1L which uses 64-bit `int`.

alibuild (approved):

PR #6808, opened 2024-07-10
PWGMM: Mult: use event selection in reducer

aalkin:


alibuild (approved):

PR #6813, opened 2024-07-10
PWGHF: add missing line in ML training config file

AlexBigO:
Hi @fcatalan92 !

@VDiBella found a missing line in the ML training config file (the `tree_name` line). I must have accidentally deleted it when I modifed the config file to make it "dummy" for the `D2H/Macros` folder.

I also take the opportunity to add an extra `try... except` safety by defining one dedicated to `hipe4ml_converter` (as it is currently mixed with the `hipe4ml` one).

### DIFF
--- a/PWGHF/D2H/Macros/ML/train_models.py
+++ b/PWGHF/D2H/Macros/ML/train_models.py
@@ -37,10 +37,14 @@ try:
     from hipe4ml import plot_utils
     from hipe4ml.model_handler import ModelHandler
     from hipe4ml.tree_handler import TreeHandler
-    from hipe4ml_converter.h4ml_converter import H4MLConverter
     from sklearn.model_selection import train_test_split
 except ModuleNotFoundError:
     print("Module 'hipe4ml' is not installed. Please install it to run this macro")
+
+try:
+    from hipe4ml_converter.h4ml_converter import H4MLConverter
+except ModuleNotFoundError:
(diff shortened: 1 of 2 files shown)


PWGHF/D2H/Macros/ML/train_models.py
```diff
@@ -37,11 +37,15 @@
     from hipe4ml import plot_utils
     from hipe4ml.model_handler import ModelHandler
     from hipe4ml.tree_handler import TreeHandler
-    from hipe4ml_converter.h4ml_converter import H4MLConverter
     from sklearn.model_selection import train_test_split
```
fcatalan92, line 40: This line can be moved outside the try except

fcatalan92 (approved):
Hi @AlexBigO, thanks for the fix! All looks good to me. 
Looking at the code a second time, a small suggestion came to my mind. Moreover, I would also move the import of hipe4ml before the ones of matplotlib, numpy, etc., since these packages are dependencies of hipe4ml and should be there if the latter is installed (while if hipe4ml is not installed, they could not be too).

PWGHF/D2H/Macros/ML/train_models.py
AlexBigO: Thanks for the comments! I moved the mentioned dependencies after `hipe4ml` import as advised.

fcatalan92 (approved):

PR #6821, opened 2024-07-11
PWGCF: FemtoUniverse -- Changes in the two track cuts for the h-Phi analysis

zchochul:
Removing the kaon PID check for the hadron (mostly proton) in the two track cut section of the code


alibuild (approved):

PR #6823, opened 2024-07-11
PWGCF: FemtoUniverse - Changing mLambda to mParticle in FemtoDerived and adding bachelor track type

basiach:
Changing the invariant mass field name from mLambda to mParticle in FDCollisions in FemtoDerived and all tasks that use derived data. Adding kBachelor track type to FemtoDerived.


PWGCF/FemtoUniverse/DataModel/FemtoDerived.h
```diff
@@ -84,8 +85,8 @@ DECLARE_SOA_COLUMN(Cut, cut, cutContainerType);          //! Bit-wise container
 DECLARE_SOA_COLUMN(PIDCut, pidcut, cutContainerType);    //! Bit-wise container for the different PID selection criteria \todo since bit-masking cannot be done yet with filters we use a second field for the PID
 DECLARE_SOA_COLUMN(TempFitVar, tempFitVar, float);       //! Observable for the template fitting (Track: DCA_xy, V0: CPA)
 DECLARE_SOA_SELF_ARRAY_INDEX_COLUMN(Children, children); //! Field for the track indices to remove auto-correlations
-DECLARE_SOA_COLUMN(MLambda, mLambda, float);             //! The invariant mass of V0 candidate, assuming lambda
```
zchochul, line 87: Should we really change the name of the column? It will make our tables no longer compatible with FemtoDream ones

zchochul: We need to meet with FemtoDream group and talk about it, maybe both groups can change it. Can you move this PR to draft?

victor-gonzalez: @zchochul @basiach I agree, this cannot be unilaterally changed

zchochul: @victor-gonzalez @basiach I also think that the FemtoDream group might not want to change this name because they have produced skimmed data for their analysis in the Hyperloop and it would stop working if they changed it. So in my opinion better solution for the clarity of the data would be to add a comment explaining that we use MLambda column also for mass of other particles that are not Lambda

victor-gonzalez: I agree with the 'comment' suggestion for the time being
But also it can be checked with the FemtoDream team for when they have planned an update of the data model and have it included as part of the update

PR #6824, opened 2024-07-11
[PWGDQ/EM] track propagation to associated vtx in tableMaker

feisenhu:
Adding the track propagation to `skimTracks` of table maker. 
@rbailhac Could you have a second look at this?


alibuild (approved):

PR #6830, opened 2024-07-11
PWGHF: add MC tree for Lc background study.

mfaggin:


PWGHF/D2H/Tasks/taskCharmPolarisation.cxx
```diff
@@ -661,12 +705,35 @@ struct TaskPolarisationCharmHadrons {
     return true;
   }
 
+  template <typename TRK, typename PART>
+  void searchFirstLevelMother(TRK const& prongTrack, std::vector<int>& idMothers, PART const&)
```
fgrosa, line 709:
```suggestion
  template <typename Trk, typename Part>
  void searchFirstLevelMother(Trk const& prongTrack, std::vector<int>& idMothers, Part const&)
```

PWGHF/D2H/Tasks/taskCharmPolarisation.cxx
```diff
@@ -661,12 +705,35 @@ struct TaskPolarisationCharmHadrons {
...
+      return;
+    }
+    // loop over the mother particles of the analysed particle
+    for (auto iMother = prongParticle.mothersIds().front(); iMother <= prongParticle.mothersIds().back(); ++iMother) {
+      // if a mother is already present in the vector, do not check it again
+      if (std::find(idMothers.begin(), idMothers.end(), iMother) != idMothers.end()) {
+        continue;
+      }
+      idMothers.push_back(iMother);
+    }
```
fgrosa, line 727: why do you loop over all of them, if you in the end only use the first one?

PWGHF/D2H/Tasks/taskCharmPolarisation.cxx
```diff
@@ -874,26 +941,124 @@ struct TaskPolarisationCharmHadrons {
...
+            pVecPion = candidate.pVectorProng0();
+          }
+          const float massKPi = RecoDecay::m(std::array{pVecKaon, pVecPion}, std::array{o2::constants::physics::MassKPlus, o2::constants::physics::MassPiPlus});
+          const float massKProton = RecoDecay::m(std::array{pVecKaon, pVecProton}, std::array{o2::constants::physics::MassKPlus, o2::constants::physics::MassProton});
+          const float massPiProton = RecoDecay::m(std::array{pVecPion, pVecProton}, std::array{o2::constants::physics::MassPiPlus, o2::constants::physics::MassProton});
+
+          /// Fill the table for selected candidates
+          /// No need to check explicitly if candidates are selected, since the Filter is applied
+          rowCandLcBkg(invMassCharmHadForSparse, ptCharmHad, rapidity,
+                       cosThetaStarHelicity, cosThetaStarProduction, cosThetaStarBeam, cosThetaStarRandom,
```
fgrosa, line 1054: Do we want all the cost* axes, or maybe only one is sufficient (helicity, or configurable)?

fgrosa (changes requested):
Hi @mfaggin, thanks! I just have few minor comments, please have a look below

PWGHF/D2H/Tasks/taskCharmPolarisation.cxx
mfaggin: this is a refuse, at the beginning I was considering to save them all then I changed my mind. Fixed it now (I break immediately after the 1st mother)

PWGHF/D2H/Tasks/taskCharmPolarisation.cxx
mfaggin: good point, I made it configurable just for flexibility, but indeed we are more interested on helicity. Fixed it.

mfaggin:
Thanks @fgrosa , I addressed your comments

fgrosa (approved):

PR #6839, opened 2024-07-12
replaced Track_as with itsTrack_as to get ITS track with all properties

alcaliva:
…es stored


fmazzasc:
Hi Alberto, small suggestion: take only the cl size from the itsTrack and all the other parameters from the full strange track, which is much more precise

alcaliva:
Hi Francesco, done! Thanks for the suggestion.

fmazzasc (approved):

PR #6849, opened 2024-07-13
PWGUD: Personal task fix

rolavick:
fix to avoid calling undefined histograms


alibuild (approved):

PR #6874, opened 2024-07-16
[PWG-LF] fix bug in filling eff. histos

njacazio:


alibuild (approved):

PR #6886, opened 2024-07-17
PWGHF: Add multiplicity dependence

deependra170598:
Hi @fgrosa @fcatalan92 @mfaggin , In this PR I tried to assign FT0M amplitudes of reconstructed collisions to corresponding Generator level collisions (McCollisions).  This was needed to obtain the AccXEff for different centrality. Please provide your suggestions/comments to progress in the right direction. This PR is still in draft, after implementing all your suggestions, I will do formatting in last.

I found that some McCollisions have more than a Reconstructed collision. In this case, I assigned max FT0M amplitude to McCollision.


PWGHF/D2H/Tasks/taskDstarToD0Pi.cxx
```diff
@@ -15,18 +15,86 @@
...
+using HfCentMcCollision = HfCentMcCollisions::iterator;
+} // namespace o2::aod
+
 using namespace o2;
 using namespace o2::analysis;
 using namespace o2::framework;
 using namespace o2::framework::expressions;
+using namespace o2::soa;
+
+struct HfCentMcDstarToD0Pi {
```
fgrosa, line 51: Instead of producing a new table you can directly do it in the `processMC` only for the collisions with a Dstar meson. In particular, you can use `PresliceUnsorted` to preslice the reco collisions (joined with MM lables) according to the `mcCollisionId`, see an example here:
 https://github.com/AliceO2Group/O2Physics/blob/8f31b9faf32e2900e39216f9d597cad74e9d4e3f/Tutorials/PWGUD/UDTutorial_03b.cxx#L88
And then you loop over the slice of reco collisions.

PWGHF/D2H/Tasks/taskDstarToD0Pi.cxx
```diff
@@ -15,18 +15,86 @@
...
+          colCentVec.push_back(recCol.centFT0M());
+        }
+        std::sort(colCentVec.begin(), colCentVec.end());
+        auto centFT0MGen = colCentVec[colCentVec.size() - 1];
+
+        // filling centrality table at gen level if collision was reconstructed at REC level successfully
+        rowsCentMcCollision(centFT0MGen);
+      } else {
+        // filling centrality table at gen level if collision was not reconstructed at REC level
+        rowsCentMcCollision(0.0);
```
fgrosa, line 93: put a negative value if the collision is not reconstructed, since `0` is still a value in the valid range

fgrosa (changes requested):
Hi @deependra170598! I have a suggestion on how to do it more efficiently, please have a look. Thanks!

PWGHF/D2H/Tasks/taskDstarToD0Pi.cxx
deependra170598: Hi @fgrosa, I tried this way earlier but it was throwing some error:
When I tried `sliceByCachedUnsorted`, ' sliceByCached`, or `sliceBy`, Error: Index pointing to Collisions is not bound! Did you subscribe to the table?
While I have already subscribed to the collisions table.
I have attached here the link for the code with the method discussed above.
https://github.com/deependra170598/O2Physics/blob/Add-Multiplicity-Dependence-2/PWGHF/D2H/Tasks/taskDstarToD0Pi.cxx

There might be some mistakes in the code that I could not understand. Please, have a look at it. Thanks

fgrosa: Hi @deependra170598 thanks for the reply. Indeed it seems correct to me, I just have a doubt since in the case of the reco part you use the unsliced reco collision table, while for the gen part you use it sliced. Is the error raised here https://github.com/deependra170598/O2Physics/blob/9ec54ef50d236dad7e8f39c59a9450bc40bc41fe/PWGHF/D2H/Tasks/taskDstarToD0Pi.cxx#L212C1-L212C80 or here https://github.com/deependra170598/O2Physics/blob/9ec54ef50d236dad7e8f39c59a9450bc40bc41fe/PWGHF/D2H/Tasks/taskDstarToD0Pi.cxx#L278?

aalkin: Hi. The error "index is not bound" is produced when index is accessed while the pointer to the target table is not set. Slicing does not access the target table, it only uses the index values so it does not indeed produce this error. It comes from the fact that the partition `rowsSelectedCandDstarMcRec` does not have its index bound to collisions. You can do it manually by calling `.bindExternalIndices(&collisions);` on it once before the loop.

For accessing collisions, grouped on mc collisions, specifically, there is a also pattern that you can find in this example: https://github.com/AliceO2Group/O2Physics/blob/868dad1a441cc1cd026c86104067135fe8fe7beb/Tutorials/src/mcHistograms.cxx#L166

fgrosa: Thanks a lot @aalkin! @deependra170598 could you please try to implement Anton's suggestion?

deependra170598: Thanks to @aalkin & @fgrosa , I tested the above suggestion, and the code works now. I have the next question: if a McCollision corresponds to two RecCollisions, then how can correct centrality be assigned to McCollision? In the above code, I assigned the maximum one of the two (three or more) centralities.

aalkin: Indeed, in these special cases there is no obvious choice. The usual practice, however, is to consider the reconstructed vertex with the largest multiplicity as a primary. Which corresponds to minimal centrality.

PWGHF/D2H/Tasks/taskDstarToD0Pi.cxx
deependra170598: Hi @fgrosa, I created a [new PR](https://github.com/AliceO2Group/O2Physics/pull/6957) for all the above changes suggested. Please have a look at it. Thanks.

PR #6895, opened 2024-07-18
TrackV0 centrality

mm1707:
Added process for measuring centrality in TrackV0 femtoscopy Run3


PWGCF/FemtoUniverse/TableProducer/femtoUniverseProducerTask.cxx
```diff
@@ -1069,6 +1069,23 @@ struct femtoUniverseProducerTask {
     }
   }
 
+  template <bool isMC, typename V0Type, typename TrackType,
+            typename CollisionType>
+  void fillCollisionsAndTracksAndV0AndPhiCentRun3(CollisionType const& col, TrackType const& tracks, V0Type const& fullV0s)
+  {
+    if (ConfIsRun3)
```
zchochul, line 1076: I think that if you intend to use fillCollisionsAndTracksAndV0AndPhiCentRun3 you should not need ConfIsRun3, it could generate issues in the future.

PWGCF/FemtoUniverse/TableProducer/femtoUniverseProducerTask.cxx
```diff
@@ -1069,6 +1069,23 @@ struct femtoUniverseProducerTask {
...
+      fillCollisionsCentRun3<isMC>(col, tracks);
```
zchochul, line 1077: Also do we need a separate function for that? The only difference I see between this one and fillCollisionsAndTracksAndV0AndPhi is the function for filling collisions? What if we would just add an if in the old function?

PWGCF/FemtoUniverse/TableProducer/femtoUniverseProducerTask.cxx
```diff
@@ -1093,6 +1110,18 @@ struct femtoUniverseProducerTask {
...
 
+  void processTrackV0CentRun3(aod::FemtoFullCollisionCentRun3 const& col,
+                              aod::BCsWithTimestamps const&,
+                              soa::Filtered<aod::FemtoFullTracks> const& tracks,
+                              o2::aod::V0Datas const& fullV0s)
+  {
+    // get magnetic field for run
+    getMagneticFieldTesla(col.bc_as<aod::BCsWithTimestamps>());
+    // fill the tables
+    fillCollisionsAndTracksAndV0AndPhiCentRun3<false>(col, tracks, fullV0s);
```
zchochul, line 1121: Maybe a better option for this process function would be creating particles that you really need in the analysis. Try doing something like:
-getMagneticField
-fillCollisions
-fillTracks
-fillV0

zchochul (approved):

zchochul (approved):

PR #6905, opened 2024-07-19
PWGCF: Fix bug in fillMCParticle, check if motherparticlesMC is empty

alicja-pp:
Fixing a bug in fillMCParticle in femtoUniverseProducer to check if motherparticlesMC is empty or not before accesing its first element.


zchochul:
Hi @alicja-pp, 
Could you describe why we need this change in a greater detail? Could it affect any other analysis?

alicja-pp:
This array auto motherparticlesMC = particleMC.template mothers_as<aod::McParticles>(); might be empty and cause a fault in a task when trying to access its first element with .front(). This change adds a check to see if it is empty, if (!motherparticlesMC.empty()), and if it is not, the next step is the same as previously. It is just an additional check to make this code work for all kinds of particles, it should not affect the results of any analysis.

victor-gonzalez:
I agree with the changes
The return of `front` is undefined if the array is empty, any further use of it is unpredictable. Its use compromises the table producer task.

zchochul (approved):

PR #6916, opened 2024-07-19
PWGEM/Dilepton: change output format from THnSparse to TProfile3D for flow

dsekihat:
PWGEM/Dilepton: change output format from THnSparse to TProfile3D for flow


alibuild (approved):

PR #6921, opened 2024-07-21
Common: add MC Collision converter

ddobrigk:
This change follows O2 [#13287](https://github.com/AliceO2Group/AliceO2/pull/13287) and adds a MC collision converter that fills a dummy EP angle for MC collisions. Created as draft for now.


alibuild (approved):

PR #6922, opened 2024-07-21
PWGEM/Dilepton: add skimmerSecondaryElectron for tag-and-probe

dsekihat:
PWGEM/Dilepton: add skimmerSecondaryElectron for tag-and-probe


alibuild (approved):

PR #6926, opened 2024-07-22
PWGEM/PhotonMeson: add min/max. TPC n sigma El separately

dsekihat:
PWGEM/PhotonMeson: add min/max. TPC n sigma El separately


alibuild (approved):

PR #6929, opened 2024-07-22
PWGHF: DhCorr task update with ML option

apalasciano:
Update of the D+-h correlation workflow, based on the developments of Ds-h correlations (by @scattaru) .
It includes code for the evaluation of:
- Secondary particle contamination
- Monte Carlo Closure test
- Include ML scores (Bkg, Prompt)
- Variations of sideband definition


fcolamar (approved):
Ciao Antonio, I think that there are some refinements that could be done to the code and that we can discuss (as it's normal as it's a very large PR with lots of novelties). Anyway, since it correctly compiles and you told me that its funcitonality is correct, and since the new features are needed urgently to try to catch the Hard Probes approvals deadlines, I'm approving the PR now, with the idea that the code will be improved after the HP rush.
Cheers!

PR #6957, opened 2024-07-23
PWGHF: Add multiplicity dependence v2

deependra170598:
This PR is resulut of [old PR] (https://github.com/AliceO2Group/O2Physics/pull/6886). 
This assign the centrality to McCollision from Reconstructed Collisions.

### DIFF
--- a/PWGHF/D2H/Tasks/taskDstarToD0Pi.cxx
+++ b/PWGHF/D2H/Tasks/taskDstarToD0Pi.cxx
@@ -47,6 +47,9 @@ struct HfTaskDstarToD0Pi {
   using CollisionsWCent = soa::Join<aod::Collisions, aod::CentFT0Ms>;
   using CollisionsWCentMcLabel = soa::Join<aod::Collisions, aod::McCollisionLabels, aod::CentFT0Ms>;
 
+  PresliceUnsorted<CollisionsWCentMcLabel> colsPerMcCollision = aod::mccollisionlabel::mcCollisionId;
+  SliceCache cache;
+
   Partition<CandDstarWSelFlag> rowsSelectedCandDstar = aod::hf_sel_candidate_dstar::isSelDstarToD0Pi == selectionFlagDstarToD0Pi;
   Partition<CandDstarWSelFlagMcRec> rowsSelectedCandDstarMcRec = aod::hf_sel_candidate_dstar::isRecoD0Flag == selectionFlagHfD0ToPiK;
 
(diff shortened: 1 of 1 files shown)


PWGHF/D2H/Tasks/taskDstarToD0Pi.cxx
```diff
@@ -253,15 +280,45 @@ struct HfTaskDstarToD0Pi {
...
+
+        float centFT0MGen;
+        if (recCollisions.size()) {
+          // assigning centrality to MC Collision using max FT0M amplitute from Reconstructed collisions
+          std::vector<float> colCentVec;
+          for (const auto& recCol : recCollisions) {
+            colCentVec.push_back(recCol.centFT0M());
+          }
+          std::sort(colCentVec.begin(), colCentVec.end());
+          centFT0MGen = colCentVec[0]; // Selecting most central collision amplitute
```
fgrosa, line 312: Here you should select the collision with the larger number of PV contributors (`collision.numContrib()`) and not the larger centrality

PWGHF/D2H/Tasks/taskDstarToD0Pi.cxx
```diff
@@ -42,12 +44,16 @@ struct HfTaskDstarToD0Pi {
   using CandDstarWSelFlagMcRec = soa::Join<aod::HfD0FromDstar, aod::HfCandDstars, aod::HfSelDstarToD0Pi, aod::HfCandDstarMcRec>;
   using CandDstarMcGen = soa::Join<aod::McParticles, aod::HfCandDstarMcGen>;
 
+  using CollisionsWCent = soa::Join<aod::Collisions, aod::CentFT0Ms>;
+  using CollisionsWCentMcLabel = soa::Join<aod::Collisions, aod::McCollisionLabels, aod::CentFT0Ms>;
```
fgrosa, line 48: It is ok to use `CentFT0Ms` as default for pp collisions, but it would also be good in the future to be able to change estimator (for instance `CentFT0Cs` has to be used in Pb-Pb). For this you could templetise the process function and have multiple process functions for the different centrality estimators.
No need to do it now, but just keep it in mind for the future.

fgrosa (changes requested):
Hi @deependra170598 thanks! I just have a couple of comments, then the PR looks good to me.
Please also close the old PR.

PWGHF/D2H/Tasks/taskDstarToD0Pi.cxx
deependra170598: I tried to make the changes that were needed. Please look at this if the method is correct.

PWGHF/D2H/Tasks/taskDstarToD0Pi.cxx
deependra170598: Thanks Fabrizio.

PWGHF/D2H/Tasks/taskDstarToD0Pi.cxx
fgrosa: It looks good to me, thanks!

fgrosa (approved):

PR #6973, opened 2024-07-24
[PWGDQ] Adding extra zdc tables for SP calibrations

lucamicheletti93:
- Adding extra zdc tables for SP calibrations
- Adding histograms for offline checks


alibuild (approved):

PR #6977, opened 2024-07-24
[PWGUD] Adding 2D histos to better monitor multiplicity

rolavick:


alibuild (approved):

PR #6978, opened 2024-07-24
K*(892) optimized

murioni:
-adding a new task for K*(892)  analysis in PbPb using filters and partitions on ResonanceTables 
-adding TrackOccupancyInTimeRange and ImpactParameter to ResonanceTables

### DIFF
--- a/PWGLF/DataModel/LFResonanceTables.h
+++ b/PWGLF/DataModel/LFResonanceTables.h
@@ -53,14 +53,13 @@ enum {
   kAllCutsINELg010,
   kECend,
 };
-DECLARE_SOA_COLUMN(Cent, cent, float);                                         //! Centrality (Multiplicity) percentile (Default: FT0M)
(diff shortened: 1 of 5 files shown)


BongHwi:
Hello @murioni ,
Can we consider adding an option of trackOccupancyInTimeRange cut on the initializer or CollisionCuts?
One of the main idea of using the initializer is to handle all event selections on the initializer and minimize the analysis task.
If the trackOccupancyInTimeRange value is needed to be computed in the analysis task, we can save it on the collision table.

murioni:
> Hello @murioni , Can we consider adding an option of trackOccupancyInTimeRange cut on the initializer or CollisionCuts? One of the main idea of using the initializer is to handle all event selections on the initializer and minimize the analysis task. If the trackOccupancyInTimeRange value is needed to be computed in the analysis task, we can save it on the collision table.

Hello @BongHwi ,
I've added the occupancy cut to the CollisionCuts.h and added a configurable in ResonanceInitializer to apply this cut.
Thank you

BongHwi (approved):

PR #6996, opened 2024-07-25
PWGHF: Adapt process function for generated particles for cent cuts

fgrosa:


alibuild (approved):

PR #7008, opened 2024-07-26

yashpatley:


victor-gonzalez (approved):
I approve for the time being but there code that is replicated which I suggest to simplify 
Also there are control of right binning done at the most inner loop which I would suggest to improve it also. These checks should be done when the particles are selected, once they are it is not needed to further repeat any of the selection checks

PR #7009, opened 2024-07-26
Fixing minor issue: the MC mean must be subtracted to reco-gen DCA di…

arossi81:
…fference before scaling the latter by data/MC resolution ratio.


mfaggin:
thanks @arossi81 as we discussed this is fine with me. Can we merge it then @ddobrigk @iarsene @ktf @jgrosseo ?

jgrosseo (approved):

PR #7019, opened 2024-07-27
Common/TableProducer/qVectorsTable.cxx: q vector calibration up to cent. 90%

dsekihat:
Common/TableProducer/qVectorsTable.cxx: enable q vector calibration up to centrality 90%


dsekihat:
Dear @ddobrigk and @jikim1290 , I would like to ask to calibrate q vectors up to centrality 90%. If I understand correctly, 80% was hardcoded, because the centrality calibration was available only up to 80% a few years ago.

PR #7023, opened 2024-07-28
Common/TableProducer/qVectorsTable.cxx: change max cent from 80 to 90

dsekihat:
Common/TableProducer/qVectorsTable.cxx: change max cent from 80 to 90

### DIFF
--- a/Common/TableProducer/qVectorsTable.cxx
+++ b/Common/TableProducer/qVectorsTable.cxx
@@ -522,14 +522,14 @@ struct qVectorsTable {
       coll.centFV0A()};
     cent = centAllEstim[cfgCentEsti];
     bool IsCalibrated = true;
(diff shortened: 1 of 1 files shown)


dsekihat:
Dear @ddobrigk and @jikim1290 , I would like to ask to calibrate q vectors up to centrality 90%. If I understand correctly, 80% was hardcoded, because the centrality calibration was available only up to 80% a few years ago.

ddobrigk (approved):
Approving! Please note, in any case, that centrality goes all the way to 100% now (though beware, of course, of problems at low mult). Note also the current initiative to make a MC with flow (see the last AOT meeting) that might be relevant.

dsekihat:
Dear @ddobrigk, Thank you very much!

PR #7041, opened 2024-07-29
PWGHF: add monitoring histos from ZORRO for software triggers

fgrosa:


alibuild (approved):

PR #7044, opened 2024-07-29
PWGEM/PhotonMeson: Reduce size of derived skimmed photon AO2Ds

nstrangm:
- Remove columns from skimmedCluster table that are not needed for analysis to reduce size of derived data
- Remove index to collision table to avoid unbound indices
- Absorb matched tracks in cluster table
- Remove dedicated matched track arguments from GammaGamma task


mhemmer-cern (approved):
Looks good!

mhemmer-cern (approved):

PR #7056, opened 2024-07-30
[PWGDQ] Apply zorro selections for DQ analyses

lucamicheletti93:
Adding code to use zorro selections for DQ analyses


alibuild (approved):

alibuild (approved):

PR #7066, opened 2024-07-30
[Infrastructure] Add O2 linter

vkucera:
Adds a GitHub action that runs a script that checks for O2-specific issues in the code.
The script can run standalone, taking paths of files as arguments.
When running in the GitHub action on the pull_request event, the script tests files modified by the PR and generates error messages as GitHub annotations which appear as comments on the relevant lines or at the top of the file in case of per-file tests.
When running in the GitHub action on the push event, the script tests files modified in the branch w.r.t. the main branch and does not produce annotations.
False positives should be rare but, if any, they can be silenced per line for most of the tests by adding a comment in a given format.

Implements the following tests.

Bad practice:

- included `iostream`
- importing names from the `std` namespace
- using directives in headers
- missing `std::` prefix for common names from the `std` namespace
- unnecessary use of ROOT entities
- use of external pi
- adding/subtracting of 2 pi
- multiples/fractions of pi for existing equivalent constants
- use of `TDatabasePDG`
- use of hard-coded PDG codes
- unnecessary call of `Mass()` for a known PDG code
- non-O2 logging
- not using const refs in range-based `for` loops
- not using const refs in `process` function subscriptions
- usage of workflow options in `defineDataProcessing`

Documentation:

- mandatory documentation of C++ files

Naming conventions:

- functions
- variables
- macros
- `constexpr` constants
- namespaces
- defined types
- enumerators
- classes
- structs
- O2 columns
- O2 tables
- O2 workflows
- explicit task names
- workflow files
- configurables
- C++ files
- Python files

PWG-HF specific conventions:

- PWGHF: names of structs and classes
- PWGHF: names of task files
- PWGHF: order of struct members

### DIFF
--- a/.github/workflows/o2-linter.yml
+++ b/.github/workflows/o2-linter.yml
@@ -1,5 +1,5 @@
 ---
-# Find code issues in modified text files
+# Find issues in O2 code
 name: O2 linter
 
 'on': [pull_request, push]
@@ -23,6 +23,7 @@ jobs:
       - name: Run tests
         run: |
           # Diff against the common ancestor of the source branch and the target branch.
-          files=$(git diff --diff-filter d --name-only origin/${{ env.MAIN_BRANCH }}...)
+          readarray -t files < <(git diff --diff-filter d --name-only origin/${{ env.MAIN_BRANCH }}...)
           [ ${{ github.event_name }} == 'pull_request' ] && options="-g"
-          python Scripts/o2_linter.py $options $files
+          # shellcheck disable=SC2086 # Ignore unquoted options.
+          python Scripts/o2_linter.py $options "${files[@]}"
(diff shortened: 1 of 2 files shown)


ktf:
I like this, why is it still draft?

victor-gonzalez:
I have been running a previous version (two days old) on a few of my source files and found two things so far which I guess haven't been modified
- I use `process` in the name of few functions and that requires all parameter being passed by reference, even `int` or `float`. I guess I should use the `no lint` directive in these lines
- I use `using` directives in a header but only inside of `inline` functions. Is this as well not recommended

Thanks @vkucera for the tremendous work!!!

vkucera:
@ddobrigk Ready to go!

ddobrigk:
Merging following also discussion on Monday the 25th November. Thanks @vkucera !

ddobrigk (approved):

PR #7071, opened 2024-07-30
PWGCF: Fix in FemtoDream

ariedel-cern:
This PR addresses two fixes
- Remove filters used in pair-track-V0 task. When using filters, we encountered problems with applying deta-dphi cut. Applying selections of V0 and tracks separately with partitions has fixed the issue. 
- Use the new feature of configurable groups by declaring the prefix only once (in track-track and track-v0 task)


victor-gonzalez (approved):
Anton I'm in the USA and I missed this one for you to have it in tomorrow's tag
Sorry about that

PR #7086, opened 2024-07-31
Add Arrow dependency to EventFilteringUtils

mpuccio:
Add the dependency explicitly to try to fix the issue on hyperloop


alibuild (approved):

PR #7089, opened 2024-07-31
PWGHF: Ds-h correlation, effciency processes modified

scattaru:
Dear all,
I have modified the processes for evaluating both the candidate and associated particle efficiencies in order to consider at the kinematic level only those collisions which are reconstructed with the same selections applied on data. 

Let me know for any comments,
Cheers,
Samuele

### DIFF
--- a/PWGHF/HFC/TableProducer/correlatorDsHadrons.cxx
+++ b/PWGHF/HFC/TableProducer/correlatorDsHadrons.cxx
@@ -431,7 +431,7 @@ struct HfCorrelatorDsHadrons {
     // MC reco level
     bool isDsPrompt = false;
     bool isDsSignal = false;
-    bool isCorrectInvMassIpo = false;
+    bool isCorrectInvMassHypo = false;
     bool isDecayChan = false;
     bool isAlreadyFilledEvent = false;
     float multiplicityFT0M = collision.multFT0M();
--- a/PWGHF/HFC/TableProducer/correlatorDsHadrons.cxx
+++ b/PWGHF/HFC/TableProducer/correlatorDsHadrons.cxx
@@ -447,7 +447,7 @@ struct HfCorrelatorDsHadrons {
       }
 
       auto prong0McPart = candidate.template prong0_as<TracksWithMc>().template mcParticle_as<aod::McParticles>();
-      isCorrectInvMassIpo = ((std::abs(prong0McPart.pdgCode()) == kKPlus) && (candidate.isSelDsToKKPi() >= selectionFlagDs)) || ((std::abs(prong0McPart.pdgCode()) == kPiPlus) && (candidate.isSelDsToPiKK() >= selectionFlagDs));
+      isCorrectInvMassHypo = ((std::abs(prong0McPart.pdgCode()) == kKPlus) && (candidate.isSelDsToKKPi() >= selectionFlagDs)) || ((std::abs(prong0McPart.pdgCode()) == kPiPlus) && (candidate.isSelDsToPiKK() >= selectionFlagDs));
 
       double efficiencyWeightD = 1.;
       if (applyEfficiency) {
--- a/PWGHF/HFC/TableProducer/correlatorDsHadrons.cxx
+++ b/PWGHF/HFC/TableProducer/correlatorDsHadrons.cxx
@@ -64,7 +64,6 @@ struct HfCorrelatorDsHadronsSelCollision {
   using SelCollisions = soa::Join<aod::Collisions, aod::EvSels>;
   using CandDsData = soa::Filtered<soa::Join<aod::HfCand3Prong, aod::HfSelDsToKKPi>>;
   using CandDsMcReco = soa::Filtered<soa::Join<aod::HfCand3Prong, aod::HfSelDsToKKPi, aod::HfCand3ProngMcRec>>;
-  using CandDsMcGen = soa::Join<aod::McParticles, aod::HfCand3ProngMcGen>;
 
   Filter dsFilter = ((o2::aod::hf_track_index::hfflag & static_cast<uint8_t>(1 << aod::hf_cand_3prong::DecayType::DsToKKPi)) != static_cast<uint8_t>(0)) && (aod::hf_sel_candidate_ds::isSelDsToKKPi >= selectionFlagDs || aod::hf_sel_candidate_ds::isSelDsToPiKK >= selectionFlagDs);
 
@@ -140,7 +139,10 @@ struct HfCorrelatorDsHadrons {
   Configurable<bool> fillHistoData{"fillHistoData", true, "Flag for filling histograms in data processes"};
   Configurable<bool> fillHistoMcRec{"fillHistoMcRec", true, "Flag for filling histograms in MC Rec processes"};
   Configurable<bool> fillHistoMcGen{"fillHistoMcGen", true, "Flag for filling histograms in MC Gen processes"};
-  Configurable<int> selectionFlagDs{"selectionFlagDs", 7, "Selection Flag for Ds"};
+  Configurable<bool> removeCollWSplitVtx{"removeCollWSplitVtx", false, "Flag for rejecting the splitted collisions"};
+  Configurable<bool> useSel8{"useSel8", true, "Flag for applying sel8 for collision selection"};
+  Configurable<bool> selNoSameBunchPileUpColl{"selNoSameBunchPileUpColl", true, "Flag for rejecting the collisions associated with the same bunch crossing"};
+  Configurable<int> selectionFlagDs{"selectionFlagDs", 7, "Selection Flag for Ds (avoid the case of flag = 0, no outputMlScore)"};
   Configurable<int> numberEventsMixed{"numberEventsMixed", 5, "Number of events mixed in ME process"};
   Configurable<int> decayChannel{"decayChannel", 1, "Decay channels: 1 for Ds->PhiPi->KKpi, 2 for Ds->K0*K->KKPi"};
   Configurable<bool> applyEfficiency{"applyEfficiency", true, "Flag for applying D-meson efficiency weights"};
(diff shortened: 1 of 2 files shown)


PWGHF/HFC/TableProducer/correlatorDsHadrons.cxx
```diff
@@ -447,6 +431,7 @@ struct HfCorrelatorDsHadrons {
     // MC reco level
     bool isDsPrompt = false;
     bool isDsSignal = false;
+    bool isCorrectInvMassIpo = false;
```
fgrosa, line 434:
```suggestion
    bool isCorrectInvMassHypo = false;
```

PWGHF/HFC/TableProducer/correlatorDsHadrons.cxx
```diff
@@ -461,14 +446,17 @@ struct HfCorrelatorDsHadrons {
         continue;
       }
 
+      auto prong0McPart = candidate.template prong0_as<TracksWithMc>().template mcParticle_as<aod::McParticles>();
+      isCorrectInvMassIpo = ((std::abs(prong0McPart.pdgCode()) == kKPlus) && (candidate.isSelDsToKKPi() >= selectionFlagDs)) || ((std::abs(prong0McPart.pdgCode()) == kPiPlus) && (candidate.isSelDsToPiKK() >= selectionFlagDs));
```
fgrosa, line 450:
```suggestion
      isCorrectInvMassHypo = ((std::abs(prong0McPart.pdgCode()) == kKPlus) && (candidate.isSelDsToKKPi() >= selectionFlagDs)) || ((std::abs(prong0McPart.pdgCode()) == kPiPlus) && (candidate.isSelDsToPiKK() >= selectionFlagDs));
```

PWGHF/HFC/TableProducer/correlatorDsHadrons.cxx
```diff
@@ -461,14 +446,17 @@ struct HfCorrelatorDsHadrons {
...
+
       double efficiencyWeightD = 1.;
       if (applyEfficiency) {
         efficiencyWeightD = 1. / efficiencyD->at(o2::analysis::findBin(binsPtEfficiencyD, candidate.pt()));
       }
 
       std::vector<float> outputMl = {-1., -1., -1.};
 
-      if (isDsSignal && isDecayChan) {
+      if (isDsSignal && isDecayChan && isCorrectInvMassIpo) {
```
fgrosa, line 459:
```suggestion
      if (isDsSignal && isDecayChan && isCorrectInvMassHypo) {
```

PWGHF/HFC/TableProducer/correlatorDsHadrons.cxx
```diff
@@ -525,7 +513,7 @@ struct HfCorrelatorDsHadrons {
         bool isPhysicalPrimary = false;
         int trackOrigin = -1;
         // DsToKKPi and DsToPiKK division
-        if (candidate.isSelDsToKKPi() >= selectionFlagDs) {
+        if (isCorrectInvMassIpo && candidate.isSelDsToKKPi() >= selectionFlagDs) {
```
fgrosa, line 516:
```suggestion
        if (isCorrectInvMassHypo && candidate.isSelDsToKKPi() >= selectionFlagDs) {
```

PWGHF/HFC/TableProducer/correlatorDsHadrons.cxx
```diff
@@ -550,7 +538,7 @@ struct HfCorrelatorDsHadrons {
             }
           }
           entryTrackRecoInfo(track.dcaXY(), track.dcaZ(), track.tpcNClsCrossedRows());
-        } else if (candidate.isSelDsToPiKK() >= selectionFlagDs) {
+        } else if (isCorrectInvMassIpo && candidate.isSelDsToPiKK() >= selectionFlagDs) {
```
fgrosa, line 541:
```suggestion
        } else if (isCorrectInvMassHypo && candidate.isSelDsToPiKK() >= selectionFlagDs) {
```

PWGHF/HFC/TableProducer/correlatorDsHadrons.cxx
```diff
@@ -582,68 +570,80 @@ struct HfCorrelatorDsHadrons {
...
+      bool isDsPrompt = false;
+      bool isDecayChan = false;
+      int trackOrigin = -1;
+
+      const auto groupedMcParticles = mcParticles.sliceBy(perCollisionCandMc, mcCollision.globalIndex());
+
+      // MC gen level
+      for (const auto& particle : groupedMcParticles) {
+        // check if the particle is Ds
+        if (std::abs(particle.pdgCode()) != Pdg::kDS) {
```
fgrosa, line 598: This should not be necessary, since it is already included in `flagMcMatchGen`

PWGHF/HFC/Tasks/taskCorrelationDsHadrons.cxx
```diff
@@ -523,15 +522,126 @@ struct HfTaskCorrelationDsHadrons {
...
+      // TODO: add correctly multiplicity and posZ dependence
+      float multiplicityReco = collision.multFT0M();
+      float multiplicityGen = mcCollision.multMCFT0A() + mcCollision.multMCFT0C(); // multFT0M = multFt0A + multFT0C
+
+      const auto groupedMcParticles = mcParticles.sliceBy(perCollisionCandMc, mcCollision.globalIndex());
+      const auto groupedCandidates = candidates.sliceBy(perCollisionCand, collision.globalIndex());
+
+      // generated candidate loop
+      for (const auto& mcParticle : groupedMcParticles) {
+        if (std::abs(mcParticle.pdgCode()) == Pdg::kDS) {
```
fgrosa, line 566: Again this is probably not necessary, since it is already included in `(std::abs(mcParticle.flagMcMatchGen()) == 1 << aod::hf_cand_3prong::DecayType::DsToKKPi)`

PWGHF/HFC/Tasks/taskCorrelationDsHadrons.cxx
```diff
@@ -523,15 +522,126 @@ struct HfTaskCorrelationDsHadrons {
...
+  void processMcCandEfficiency(soa::Join<aod::Collisions, aod::FT0Mults, aod::EvSels, aod::McCollisionLabels> const& collisions,
                                soa::Join<aod::McCollisions, aod::MultsExtraMC> const&,
                                CandDsMcGen const& mcParticles,
                                CandDsMcReco const& candidates,
                                aod::TracksWMc const&)
   {
     auto hCandidates = registry.get<StepTHn>(HIST("hCandidates"));
 
+    /// loop over reconstructed collisions
+    for (const auto& collision : collisions) {
```
fgrosa, line 539: I am not sure about your logic; if you have two reconstructed collisions that point to the same generated collision (split vertex), then you consider twice the same generated particles. Is this really what you want to do? Wouldn't it be better to loop over mcCollisions and check if they have at least a reconstructed collision associated to it?

PWGHF/HFC/Tasks/taskCorrelationDsHadrons.cxx
```diff
@@ -523,15 +522,126 @@ struct HfTaskCorrelationDsHadrons {
...
+        std::vector<float> outputMl = {-1., -1., -1.};
+        if (candidate.isSelDsToKKPi() >= selectionFlagDs) {
+          for (unsigned int iclass = 0; iclass < classMl->size(); iclass++) {
+            outputMl[iclass] = candidate.mlProbDsToKKPi()[classMl->at(iclass)];
+          }
+        } else if (candidate.isSelDsToPiKK() >= selectionFlagDs) {
+          for (unsigned int iclass = 0; iclass < classMl->size(); iclass++) {
+            outputMl[iclass] = candidate.mlProbDsToPiKK()[classMl->at(iclass)];
+          }
+        }
```
fgrosa, line 611: be careful that this would crash if the value of `selectionFlagDs` chosen by the user is `0`, since for candidates that do not pass the topological preselections in the selector no ML output score is computed and so the vector stored in the table is empty

PWGHF/HFC/Tasks/taskCorrelationDsHadrons.cxx
```diff
@@ -608,149 +718,114 @@ struct HfTaskCorrelationDsHadrons {
...
-          } else if (std::abs(mcParticle.pdgCode()) == kMuonMinus) {
-            registry.fill(HIST("hPtPrmMuonMcGen"), mcParticle.pt());
-          }
-          int trackOrigin = RecoDecay::getCharmHadronOrigin(mcParticles, mcParticle, true);
-          if (trackOrigin == 1) { // charm orgin
-            registry.fill(HIST("hPtPrmPromptPartMcGen"), mcParticle.pt());
-          } else if (trackOrigin == 2) { // beauty origin
-            registry.fill(HIST("hPtPrmNonPromptPartMcGen"), mcParticle.pt());
+    /// loop over reconstructed collisions
+    for (const auto& collision : collisions) {
```
fgrosa, line 735: Same comment as above about the logic (I would reverse it)

PWGHF/HFC/TableProducer/correlatorDsHadrons.cxx
```diff
@@ -582,68 +570,80 @@ struct HfCorrelatorDsHadrons {
...
-    for (const auto& particle : mcParticles) {
-      // check if the particle is Ds
-      if (std::abs(particle.pdgCode()) != Pdg::kDS) {
-        continue;
-      }
-      if ((std::abs(particle.flagMcMatchGen()) == 1 << aod::hf_cand_3prong::DecayType::DsToKKPi) && (particle.flagMcDecayChanGen() == decayChannel)) {
-        double yD = RecoDecay::y(particle.pVector(), MassDS);
-        if (std::abs(yD) > yCandGenMax || particle.pt() < ptCandMin || particle.pt() > ptCandMax) {
+
+    for (const auto& collision : collisions) {
```
fgrosa, line 582: I am not sure about your logic; if you have two reconstructed collisions that point to the same generated collision (split vertex), then you consider twice the same generated particles. Is this really what you want to do? Wouldn't it be better to loop over mcCollisions and check if they have at least a reconstructed collision associated to it?

fgrosa (changes requested):
Hi @scattaru, apart from some minor comments, I have a substantial one about the logic of your implementation. You first loop over reconstructed collisions, you get the corresponding MC collision and then you loop over the generated particles associated to the MC collision. However, in case of split vertices you will have double counts (because you will have multiple times the same MC collision and hence you will loop multiple times on the same generated particles). I would rather reverse the logic and loop over the MC collisions, check if they have at least a reco collision, and in this case loop over generated collisions.

PWGHF/HFC/Tasks/taskCorrelationDsHadrons.cxx
scattaru: Thanks for the comment, for the moment to avoid this I have put a comment on the configurable

scattaru:
> Hi @scattaru, apart from some minor comments, I have a substantial one about the logic of your implementation. You first loop over reconstructed collisions, you get the corresponding MC collision and then you loop over the generated particles associated to the MC collision. However, in case of split vertices you will have double counts (because you will have multiple times the same MC collision and hence you will loop multiple times on the same generated particles). I would rather reverse the logic and loop over the MC collisions, check if they have at least a reco collision, and in this case loop over generated collisions.

Hi @fgrosa, thanks a lot for the suggestions! You are right, I have not considered this case of the split vertices. I have reversed the logic in order to loop on the generated collision and then take the reconstructed collision. I have defined a configurable `removeCollWSplitVtx`to have also the possibility to explicitly remove the case in which for one generated collision there are more than 1 associated reconstructed collisions, so we can check what happen if we consider or not also those split collisions.

fgrosa (approved):

fgrosa:
> > Hi @scattaru, apart from some minor comments, I have a substantial one about the logic of your implementation. You first loop over reconstructed collisions, you get the corresponding MC collision and then you loop over the generated particles associated to the MC collision. However, in case of split vertices you will have double counts (because you will have multiple times the same MC collision and hence you will loop multiple times on the same generated particles). I would rather reverse the logic and loop over the MC collisions, check if they have at least a reco collision, and in this case loop over generated collisions.
> 
> Hi @fgrosa, thanks a lot for the suggestions! You are right, I have not considered this case of the split vertices. I have reversed the logic in order to loop on the generated collision and then take the reconstructed collision. I have defined a configurable `removeCollWSplitVtx`to have also the possibility to explicitly remove the case in which for one generated collision there are more than 1 associated reconstructed collisions, so we can check what happen if we consider or not also those split collisions.

Thanks @scattaru!

PR #7113, opened 2024-08-01
PWGHF: fix bug in MC table processing for Omegac/Xic creator

fgrosa:
@ZFederica @stefanopolitano


alibuild (approved):

PR #7115, opened 2024-08-01
Production of more histograms in PIDML producers

mytkom:


saganatt:
Since we have time, we can refactor/shorten the code:
1. Could you make a single fillHist() function ? Either merge the two producers into one analysis task with 4 process functions (a much better solution, but you would need to update all your scripts and configs) or create a helper header file. Then, fillHistMC would just call fillHist(), and fill a few more histograms.

2. You can also make fillHist templated with an int to avoid code repetition. The int would be an index to an array of { "plus", "minus"}.
Here is a templated example: https://github.com/AliceO2Group/O2Physics/blob/master/DPG/Tasks/AOTTrack/qaEfficiency.cxx#L243
pdgSign is used to calculate histogramIndex and Form a HIST string: https://github.com/AliceO2Group/O2Physics/blob/master/DPG/Tasks/AOTTrack/qaEfficiency.cxx#L307
PDGs is an array of constexpr ints: https://github.com/AliceO2Group/O2Physics/blob/master/DPG/Tasks/AOTTrack/qaEfficiency.cxx#L71
You can make an array of constexpr char* like on line 69.

You would then use the templated fillHist like:
```
if (track.sign() < 0) {
  fillHistos<1> (track, pdgCode, isPrimary); // fills histograms "plus/..."
}
else {
  fillHistos<0> (track, pdgCode, isPrimary);
}
```
Or perhaps you could directly pass the const string in the template (I am not sure this is possible): `fillHistos<"plus">(...)`  or `fillHistos<HIST("plus")>(...)`

saganatt (approved):

PR #7128, opened 2024-08-02
[PWGEM] rework of efficiency task, include eff for tracks only recons…

feisenhu:
…tructed only once

Include changes of a local backup version of the efficiency task.
Differentiate between tracks which are reconstructed only once and those reconstructed multiple times.
Create different histograms and do a separation in double counting and collision information.


alibuild (approved):

PR #7144, opened 2024-08-03
PWGEM/Dilepton: avoid double counting of pairs when using TTCA

dsekihat:
PWGEM/Dilepton: avoid double counting of pairs when using TTCA


alibuild (approved):

PR #7162, opened 2024-08-05
PIDML: evaluate FSE + self-attention network

mytkom:
* remove detector count setting and reorder network arguments (with NaNs if detector not available)

* update README.md

* markdownlint changes

### DIFF
--- a/Tools/PIDML/README.md
+++ b/Tools/PIDML/README.md
@@ -54,7 +54,7 @@ Currently, only models for run 285064 (timestamp interval: 1524176895000 - 15242
 
 Both model and interface analysis examples can be run with a script:
 
-### Script for Run2 Converted to Run3 data:
+### Script for Run2 Converted to Run3 data
 ```bash
 #!/bin/bash
 
@@ -77,7 +77,7 @@ o2-analysis-tracks-extra-converter --configuration json://$config_file -b |
 ```
 Remember to set every setting, which states that helper task should process Run2 data to `true`.
 
-### Script for Run3 data:
+### Script for Run3 data
 ```bash
 #!/bin/bash
 
--- a/Tools/PIDML/README.md
+++ b/Tools/PIDML/README.md
@@ -14,6 +14,7 @@ This class represents a single ML model from an ONNX file. It requires the follo
 - timestamp of the input analysis data -- neded to choose appropriate model
 - PID to be checked
 - minimum certainty for accepting a track to be of given PID
+- *p* limits array - specifiying p limits for each detector configuration (TPC, TPC+TOF, TPC+TOF+TRD)
 
 Let's assume your `PidONNXModel` instance is named `pidModel`.
 Then, inside your analysis task `process()` function, you can iterate over tracks and call: `pidModel.applyModel(track);` to get the certainty of the model.
@@ -36,10 +37,10 @@ This is a wrapper around PidONNXModel that contains several models. It has the p
 
 Then, obligatory parameters for the interface:
 - a vector of int output PIDs
-- a vector of *p*T cuts for each PID.
+- a 2-dimensional LabeledArray of *p* limits for each PID, for each detector configuration. It describes the minimum *p* values at which each next detector should be included for predicting given PID
 - a vector of minimum certainties for each PID for accepting a track to be of this PID
 - boolean flag: whether to switch on auto mode. If true, then *p*T limits and minimum certainties can be passed as an empty array and an empty vector, and the interface will fill them with default configuration:
-  - *p*T limits: 0.5 for all PIDs
+  - *p* limits: same values for all PIDs: 0.0 (TPC), 0.5 (TPC + TOF), 0.8 (TPC + TOF + TRD)
   - minimum certainties: 0.5 for all PIDs
 
 You can use the interface in the same way as the model, by calling `applyModel(track)` or `applyModelBoolean(track)`. The interface will then call the respective method of the model selected with the aforementioned interface parameters.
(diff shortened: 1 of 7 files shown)


Tools/PIDML/README.md
```diff
@@ -31,10 +36,10 @@ This is a wrapper around PidONNXModel that contains several models. It has the p
 
 Then, obligatory parameters for the interface:
 - a vector of int output PIDs
-- a 2-dimensional LabeledArray of *p*T limits for each PID, for each detector configuration. It describes the minimum *p*T values at which each next detector should be included for predicting given PID
+- a vector of *p*T cuts for each PID.
 - a vector of minimum certainties for each PID for accepting a track to be of this PID
 - boolean flag: whether to switch on auto mode. If true, then *p*T limits and minimum certainties can be passed as an empty array and an empty vector, and the interface will fill them with default configuration:
-  - *p*T limits: same values for all PIDs: 0.0 (TPC), 0.5 (TPC + TOF), 0.8 (TPC + TOF + TRD)
+  - *p*T limits: 0.5 for all PIDs
```
saganatt, line 42: I would still keep 0.0, 0.5 and 0.8 for all the 3 configs (see my comment on PidOnnxInterface)

Tools/PIDML/pidOnnxInterface.h
```diff
@@ -100,11 +89,9 @@ struct PidONNXInterface {
   bool applyModelBoolean(const T& track, int pid)
   {
     for (std::size_t i = 0; i < mNPids; i++) {
-      if (mModels[i * kNDetectors].mPid == pid) {
-        for (uint32_t j = 0; j < kNDetectors; j++) {
-          if (track.pt() >= mPTLimits[i][j] && (j == kNDetectors - 1 || track.pt() < mPTLimits[i][j + 1])) {
-            return mModels[i * kNDetectors + j].applyModelBoolean(track);
-          }
+      if (mModels[i].mPid == pid) {
+        if (track.pt() >= mPTLimits[i]) {
```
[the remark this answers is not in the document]
saganatt: So, you don't need to skip applyModelBoolean when pT is below the limit.
That will maybe improve the results.

Tools/PIDML/pidOnnxModel.h
```diff
@@ -131,27 +126,22 @@ struct PidONNXModel {
     return getModelOutput(track) >= mMinCertainty;
   }
 
-  PidMLDetector mDetector;
+  bool mUseCCDB;
```
saganatt, line 129: Is it used anywhere?

Tools/PIDML/pidOnnxInterface.h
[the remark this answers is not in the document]
saganatt: And, also, the mPTLimits would make now sense in PidOnnxModel, not the Interface.
Now, we will decide per-model, when what detectors it should use, instead of deciding which model to pick.

Tools/PIDML/pidOnnxModel.h
```diff
@@ -203,60 +193,72 @@ struct PidONNXModel {
     }
   }
 
+  static constexpr float kTOFMissingSignal = -999.0f;
+  static constexpr float kTOFPCut = 0.5f;
```
saganatt, line 197: I would prefer user-defined pT cuts, a configurable array. mPTLimits transferred here from the PidOnnxInterface.

Tools/PIDML/qaPidML.cxx
```diff
@@ -326,15 +324,9 @@ struct pidml {
   void pidML(const T& track, const int pdgCodeMC)
   {
     float pidCertainties[3];
-    if (track.p() < pSwitchValue[i]) {
```
saganatt, line 329: After moving mPTLimits to PidOnnxModel, you could use pSwitchValues as configurables here.

saganatt (changes requested):
Thanks a lot for upgrading the code.
Here is my batch of comments.

Tools/PIDML/pidOnnxModel.h
```diff
@@ -203,89 +169,83 @@ struct PidONNXModel {
...
+    float scaledTPCNClsShared = (static_cast<float>(track.tpcNClsShared()) - mScalingParams.at("fTPCNClsShared").first) / mScalingParams.at("fTPCNClsShared").second;
+    float scaledDcaXY = (track.dcaXY() - mScalingParams.at("fDcaXY").first) / mScalingParams.at("fDcaXY").second;
+    float scaledDcaZ = (track.dcaZ() - mScalingParams.at("fDcaZ").first) / mScalingParams.at("fDcaZ").second;
 
-    return inputValues;
-  }
+    inputValues.insert(inputValues.end(), {track.p(), track.pt(), track.px(), track.py(), track.pz(), static_cast<float>(track.sign()), scaledX, scaledY, scaledZ, scaledAlpha, static_cast<float>(track.trackType()), scaledTPCNClsShared, scaledDcaXY, scaledDcaZ});
 
-  // FIXME: Temporary solution, new networks will have sigmoid layer added
-  float sigmoid(float x)
```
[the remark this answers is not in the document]
mytkom: On the Python side, we do not have a sigmoid layer added on the network class, but then we export it to `.onnx` with a sigmoid layer added using `nn.Sequential` in the export code. It is neither bug-prone nor hard to do. The comment says `FIXME`, so why do not fix it now if it isn't a problem to prepare such networks?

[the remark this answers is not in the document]
saganatt: Alright, I didn't know it worked for you without this sigmoid here.

Tools/PIDML/pidOnnxModel.h
mytkom: No, I would remove it

saganatt (approved):

PR #7169, opened 2024-08-05
Full Jet spectra analysis in pp in RUN 3

Archita-Dash:
Hi @mfasDa and @nzardosh .

With this PR, 
- I create a new task for my "fulljet spectra analysis in Run 3 in pp"
- the task includes histograms (as of now) to obtain the particle level jet energy resolution, jet energy scale and response matrices
- it loops over my selected EMCCollisions table
- the task is configured to run over the Data as well as MC (configurable also included when using MB Gap-Triggered MC in  near future)
- this also includes a configurable for the "work-around" for the MC datasets that are missing out the required CTP trigger aliases (in my analysis, it's kTVXinEMC) by looking for a cell content in the EMCAL
- for the MC simulations with proper trigger flags in them, the task is configured to do the default EMCAL event selection by directly picking up the trigger bit/flag information 
- there are quite some more TO DOs (ongoing) but FOR NOW from HARD PROBES 2024 POV and for the upcoming PWG approvals POV, I would like to get this task merged so that I would be able to run my task over the full statistics of the available MC!

### DIFF
--- a/ALICE3/DataModel/A3DecayFinderTables.h
+++ b/ALICE3/DataModel/A3DecayFinderTables.h
@@ -43,7 +43,10 @@ enum a3selectionBit : uint32_t { kDCAxy = 0,
                                  kTruePrPlusFromLc,
                                  kTruePiMinusFromLc,
                                  kTrueKaMinusFromLc,
-                                 kTruePrMinusFromLc };
+                                 kTruePrMinusFromLc,
+                                 kTrueXiFromXiC,
+                                 kTruePiFromXiC,
+                                 kTruePiFromXiCC };
 
 namespace o2::aod
 {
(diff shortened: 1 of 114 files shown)


nzardosh:
Hi Archita,
As we discussed previously we should add EMCALMatchedCollisions to the jet framework derived data. I have made a PR here : https://github.com/AliceO2Group/O2Physics/pull/7170
So now you can directly join JCollisions to JEMCALMatchedCollisions>

You should change this and remove any dependency of the task on aod::Collisions

nzardosh:
Hi Archita,

I am happy to approve this. Just to understand other than the event selection workaround for the EMCal, how is this task different to the jetfinderfullQA task? is there anything it does differently?

Thanks,
Nima

Archita-Dash:
Hi Nima,

thanks for your comment. As of now, it's the workaround + a configurable for the EMC gap triggered MC + all histograms in 1D / 2D for my easy analysis. As I have tested these myself and it works just fine so I pushed these for now. 
In the next days, I would like to push the developments specific to my analysis task (some of which I am testing locally but would commit them only after they look reasonable with a bigger sample and) once I present them at the meetings.

Thanks,
Archita

nzardosh (approved):

PR #7173, opened 2024-08-06
PWGEM/Dilepton: fix Join table in MC skimming task

dsekihat:
PWGEM/Dilepton: fix Join table in MC skimming task


alibuild (approved):

PR #7190, opened 2024-08-06
[PWGEM]: Add CutSetting with 30 cut variations

feisenhu:
Adding setting of cut variations.


dsekihat (commented):
Hi Florian, do you really need 30 hardcoded variations? You should think about other better ways.

alibuild (approved):

PR #7194, opened 2024-08-07
One more fix to missing std::

ktf:
Apparently recently reintroduced. Required by ROOT 6.32.XX


ktf:
Trivial and not tested. Merging.

PR #7199, opened 2024-08-07
(QVecDev) Added tables with new labeling

MaximVirta:
New labeling added for the Q-vector tables, with backwards compatibility with the old labeling convetion.
Internal variables have been changed directly.

### DIFF
--- a/Common/DataModel/Qvectors.h
+++ b/Common/DataModel/Qvectors.h
@@ -50,7 +50,6 @@ DECLARE_SOA_COLUMN(QvecTPCnegImVec, qvecTPCnegImVec, std::vector<float>);
 DECLARE_SOA_COLUMN(QvecTPCallReVec, qvecTPCallReVec, std::vector<float>);
 DECLARE_SOA_COLUMN(QvecTPCallImVec, qvecTPCallImVec, std::vector<float>);
 
-
 DECLARE_SOA_COLUMN(QvecFT0CRe, qvecFT0CRe, float);
 DECLARE_SOA_COLUMN(QvecFT0CIm, qvecFT0CIm, float);
 DECLARE_SOA_COLUMN(QvecFT0ARe, qvecFT0ARe, float);
--- a/Common/DataModel/Qvectors.h
+++ b/Common/DataModel/Qvectors.h
@@ -43,12 +43,13 @@ DECLARE_SOA_COLUMN(QvecFT0MReVec, qvecFT0MReVec, std::vector<float>);
 DECLARE_SOA_COLUMN(QvecFT0MImVec, qvecFT0MImVec, std::vector<float>);
 DECLARE_SOA_COLUMN(QvecFV0AReVec, qvecFV0AReVec, std::vector<float>);
 DECLARE_SOA_COLUMN(QvecFV0AImVec, qvecFV0AImVec, std::vector<float>);
-DECLARE_SOA_COLUMN(QvecBPosReVec, qvecBPosReVec, std::vector<float>);
-DECLARE_SOA_COLUMN(QvecBPosImVec, qvecBPosImVec, std::vector<float>);
-DECLARE_SOA_COLUMN(QvecBNegReVec, qvecBNegReVec, std::vector<float>);
-DECLARE_SOA_COLUMN(QvecBNegImVec, qvecBNegImVec, std::vector<float>);
-DECLARE_SOA_COLUMN(QvecBTotReVec, qvecBTotReVec, std::vector<float>);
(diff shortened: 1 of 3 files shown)


Common/DataModel/Qvectors.h
```diff
@@ -43,12 +43,12 @@ DECLARE_SOA_COLUMN(QvecFT0MReVec, qvecFT0MReVec, std::vector<float>);
...
-DECLARE_SOA_COLUMN(QvecBNegReVec, qvecBNegReVec, std::vector<float>);
-DECLARE_SOA_COLUMN(QvecBNegImVec, qvecBNegImVec, std::vector<float>);
-DECLARE_SOA_COLUMN(QvecBTotReVec, qvecBTotReVec, std::vector<float>);
-DECLARE_SOA_COLUMN(QvecBTotImVec, qvecBTotImVec, std::vector<float>);
+DECLARE_SOA_COLUMN(QvecTPCposReVec, qvecTPCposReVec, std::vector<float>);
+DECLARE_SOA_COLUMN(QvecTPCposImVec, qvecTPCposImVec, std::vector<float>);
+DECLARE_SOA_COLUMN(QvecTPCnegReVec, qvecTPCnegReVec, std::vector<float>);
+DECLARE_SOA_COLUMN(QvecTPCnegImVec, qvecTPCnegImVec, std::vector<float>);
+DECLARE_SOA_COLUMN(QvecTPCallReVec, qvecTPCallReVec, std::vector<float>);
+DECLARE_SOA_COLUMN(QvecTPCallImVec, qvecTPCallImVec, std::vector<float>);
```
vkucera, line 51: Please use proper camelCase for good readability. `Qvec`, `Cpos`, `Cneg`, `Call` are not words nor abbreviations.
```suggestion
DECLARE_SOA_COLUMN(QVecTpcPosReVec, qVecTpcPosReVec, std::vector<float>);
DECLARE_SOA_COLUMN(QVecTpcPosImVec, qVecTpcPosImVec, std::vector<float>);
DECLARE_SOA_COLUMN(QVecTpcNegReVec, qVecTpcNegReVec, std::vector<float>);
DECLARE_SOA_COLUMN(QVecTpcNegImVec, qVecTpcNegImVec, std::vector<float>);
DECLARE_SOA_COLUMN(QVecTpcAllReVec, qVecTpcAllReVec, std::vector<float>);
DECLARE_SOA_COLUMN(QVecTpcAllImVec, qVecTpcAllImVec, std::vector<float>);
```
This applies to all other name changes too.

MaximVirta: For the "Qvec" -> "QVec" part, I would prefer not to change those as there are many tasks which are already dependent on these labeling and would need to be changed as well. The "TPCpos" -> "TpcPos" I can change, but why should we change only TPC to be in lower case, when other detectors are in capitals?

MaximVirta: Would "TPCPos" be a better option?

vkucera: Your PR itself is titled `QVecDev`, includes files `qVectorsTable.cxx`, `qVectorsCorrection.cxx`, and defines variables like `qVectorBPosVec`. So apparently the capitalisation of `v` is very common in your code and not an issue. So I would say that the occurrences of lower-case `v` (e.g. `Qvectors.h`) should be fixed to avoid inconsistencies.

vkucera: Concerning the capitalisation of TPC, while it might be fine at the beginning of a word, it becomes less readable when in the middle of the word and very unreadable when combined with another detector acronym, like TPCTOF, see the examples `ITSReco` vs `numberOfDnsConnections` in the [naming conventions](https://rawgit.com/AliceO2Group/CodingGuidelines/master/naming_formatting.html?showone=General_Naming_Rules#General_Naming_Rules).
Other detectors are not exceptions, of course.

vkucera: So `TPCPos` is definitely better than `TPCpos` but `TpcPos` is even better.

strogolo: Dear @vkucera,
we prefer not to change the capitalisation of the `v` because for the other detectors used in the Q-vector calculation this is the convention. So it would create troubles and differences. 

Please @MaximVirta keep the `v` not capital and be consistent with the rest of the Q-vector framework. 
This PR is intended to fix a renaming issue and not to change the convention adopted by us.

ddobrigk (approved):

ddobrigk:
Thank you @MaximVirta !

PR #7201, opened 2024-08-07
PWGLF: Prevent crush of collisicon cut in run2 case

BongHwi:
`trackOccupancyInTimeRange_noCut` histogram has to be filled only in run3 case.


alibuild (approved):

PR #7205, opened 2024-08-07
DPG/AOTTracks: fix ConfigurableAxis.

mfaggin:
- move the `ConfigurableAxis` outside the `init`, otherwise they are not seen as configurable
- add `VARIABLE_WIDTH` at the beginning for those supposed to define bins of different widths by default


alibuild (approved):

PR #7217, opened 2024-08-08
PWGLF: Fix CollCutCount hist for Run2 dataset

BongHwi:
`CollCutCounts` was not defined in Run2 case.


alibuild (approved):

PR #7219, opened 2024-08-08
[PWGUD] Exclusive phi revisited with powered up ITS psuedo-PID

siragoni:
Exclusive (in)coherent phi search in the KK channel
Powered up ITS pseudo-PID, following suggestions from ITS experts
Adapted to use with both SG and DG skimmed data


siragoni:
Hi @abylinkin @rolavick it can be merged as far as I am concerned, cheers

rolavick (approved):

PR #7221, opened 2024-08-08
[PWGEM]: fix of cut variations, add missing 'return cut;'

feisenhu:
Fixing cuts by adding `return cut;`


alibuild (approved):

PR #7235, opened 2024-08-08
PWGCF: FemtoUniverse track-V0 task -- created a config group for V0 selections

Eloviyo:


zchochul:
This is not longer needed. Basia is adding those changes in her PR, which is being merged as we speak.

PR #7250, opened 2024-08-09
PWGCF: Add nSigmaTOFDe to femtoUniversePairTaskTrackTrackMultKtExtend…

a-m-andrushko:
Add a separate nSigmaTOF configurable for De.

### DIFF
--- a/PWGCF/FemtoUniverse/Tasks/femtoUniversePairTaskTrackTrackMultKtExtended.cxx
+++ b/PWGCF/FemtoUniverse/Tasks/femtoUniversePairTaskTrackTrackMultKtExtended.cxx
@@ -84,6 +84,8 @@ struct femtoUniversePairTaskTrackTrackMultKtExtended {
   /// Table for linear cut for TPC Deuteron Sigma
   struct : o2::framework::ConfigurableGroup {
     Configurable<bool> ConfIsLine{"ConfIsLine", false, "Enable a separation line for clearer TPC Deuteron Sigma"};
+    Configurable<float> pTlow{"pTlow", 0.0f, "Lower limit of pT for linear cut of TPC Deuteron Sigma"};
+    Configurable<float> pThigh{"pThigh", 1.4f, "Higher limit of pT for linear cut of TPC Deuteron Sigma"};
     Configurable<float> a{"a", -167.0f, "Parameter 'a' of a linear function 'y = a * x + b'"};
     Configurable<float> b{"b", 300.0f, "Parameter 'b' of a linear function 'y = a * x + b'"};
   } lincut;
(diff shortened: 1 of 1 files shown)


PWGCF/FemtoUniverse/Tasks/femtoUniversePairTaskTrackTrackMultKtExtended.cxx
```diff
@@ -254,7 +254,6 @@ struct femtoUniversePairTaskTrackTrackMultKtExtended {
         return false;
       }
     }
-    return false;
```
saganatt, line 257: You can make it shorter:
```
if (mom < ...) {
  return TMath::Abs(nsigmaTPCPr) < twotrackconfigs.ConfNsigmaTPC;
}
return TMath::Hypot(nsigmaTOFPr, nsigmaTPCPr) < twotrackconfigs.ConfNsigmaCombined;
```

PWGCF/FemtoUniverse/Tasks/femtoUniversePairTaskTrackTrackMultKtExtended.cxx
```diff
@@ -321,29 +316,27 @@ struct femtoUniversePairTaskTrackTrackMultKtExtended {
...
   }
 
   /// Linear cut for clearer TPC Deuteron Sigma
   bool IsDeuteronNSigmaLinearCut(float mom, float nsigmaTPCDe, float nsigmaTOFDe, float tpcSignal)
   {
     if (lincut.ConfIsLine == true) {
-      if (mom > 0.8 && mom < 1.4) {
+      if (mom > 0 && mom < 1.4) {
```
saganatt, line 339: Maybe you want 0 and 1.4 to be configurables? Then, next time, you could update them in JSON and have no need to do PRs.

PWGCF/FemtoUniverse/Tasks/femtoUniversePairTaskTrackTrackMultKtExtended.cxx
```diff
@@ -254,7 +254,6 @@ struct femtoUniversePairTaskTrackTrackMultKtExtended {
         return false;
       }
     }
-    return false;
   }
 
   bool IsKaonNSigma(float mom, float nsigmaTPCK, float nsigmaTOFK)
```
saganatt, line 259: This is a copy of the function IsProtonSigma(). Couldn't all the functions Is...Sigma() be merged into one?
This way, in the future you will need to modify the code just in 1 place instead of 4.

PWGCF/FemtoUniverse/Tasks/femtoUniversePairTaskTrackTrackMultKtExtended.cxx
a-m-andrushko: Right, I will change them to configurables

PWGCF/FemtoUniverse/Tasks/femtoUniversePairTaskTrackTrackMultKtExtended.cxx
a-m-andrushko: Yes, but those functions are later used separately in a 'switch-case' construction in IsParticleNSigma() function depending on PDG codes, I should not think that merging them all into one would be a successful idea.

saganatt: I still don't see a problem.
In IsParticleNSigma(), you would have lines like `return IsNSigma(mom, nsigmaTPCPr, nsigmaTOFPr);`, `IsNSigma(mom, nsigmaTPCK, nsigmaTOFK);` instead of `return IsProtonNSigma(mom, nsigmaTPCPr, nsigmaTOFPr);`, `IsKaonNSigma(mom, nsigmaTPCK, nsigmaTOFK);`.
It is just a change in the name of the function. The rest works the same way.

saganatt (approved):

PR #7286, opened 2024-08-13
PWGJE: Task for tagging the b-jet using ML model

hahassan7:
A new task for tagging the b-jet using ML model loaded from CCDB in ONNX format.
I also wrote a task that converts the dataframes taken the treeCreator into TTrees that will be used for the ML training and testing. I also updated the tagging framework to support a meson definition for jet flavors.

Also in order to minimize the size of the out tree of TreeCreator some of the lf-jet will be dropped with a certain percentage.


Tools/ML/model.h
```diff
@@ -111,6 +111,39 @@ class OnnxModel
     return evalModel<T>(inputTensors);
   }
 
+  // For multi-dimensional inputs
```
fcatalan92, line 114:
```suggestion
  // For 2D inputs
```

Tools/ML/model.h
```diff
@@ -111,6 +111,39 @@ class OnnxModel
...
+  template <typename T>
+  T* evalModel(std::vector<std::vector<T>>& input)
+  {
+    std::vector<Ort::Value> inputTensors;
+
+#if !__has_include(<onnxruntime/core/session/onnxruntime_cxx_api.h>)
+    Ort::MemoryInfo mem_info = Ort::MemoryInfo::CreateCpu(OrtAllocatorType::OrtArenaAllocator, OrtMemType::OrtMemTypeDefault);
+#endif
```
fcatalan92, line 122: You can drop this if guard since it was needed to support an old ONNXRuntime version

Tools/ML/model.h
```diff
@@ -111,6 +111,39 @@ class OnnxModel
...
+      std::vector<int64_t> inputShape{static_cast<int64_t>(size / totalSize)};
+      for (size_t idim = 1; idim < mInputShapes[iinput].size(); idim++) {
+        inputShape.push_back(mInputShapes[iinput][idim]);
+      }
+
+#if __has_include(<onnxruntime/core/session/onnxruntime_cxx_api.h>)
+      inputTensors.emplace_back(Ort::Experimental::Value::CreateTensor<T>(input[iinput].data(), size, inputShape));
+#else
+      inputTensors.emplace_back(Ort::Value::CreateTensor<T>(mem_info, input[iinput].data(), size, inputShape.data(), inputShape.size()));
+#endif
```
fcatalan92, line 141: You can drop this if guard since it was needed to support an old ONNXRuntime version

fcatalan92 (changes requested):
Hi @hahassan7, I reviewed only the part concerning `model.h` and have a few minor comments. I'm assuming you tested it locally and it works, right?

@ChSonnabend Could you have a look too?

Tools/ML/model.h
ChSonnabend: For now we can also leave it in, not sure how many people might be using an older build stack still

ChSonnabend:
I'm not a huge fan of adding this new function... In principle we should keep the code simple and leave it up to the user to flatten the vector before passing it to evalModel (as far as I can see this is just done in the function itself by emplace_back into inputTensors). What do you think @fcatalan92 ?

fcatalan92:
> I'm not a huge fan of adding this new function... In principle we should keep the code simple and leave it up to the user to flatten the vector before passing it to evalModel (as far as I can see this is just done in the function itself by emplace_back into inputTensors). What do you think @fcatalan92 ?

To me having this additional function looks fine. I would not expect users to work with `evalModel(std::vector<Ort::Value>& input)` directly. Looking at the current use case, each instance (the information of the jet to be classified) has 3 tensors as input to the model. I'm not sure this is already supported by `evalModel(std::vector<T>& input)`.

hahassan7:
Hi @fcatalan92, @ChSonnabend,
Thanks for reviewing the code. Fabio you're right, in this case my model is made of three branches and each branch takes a separate input. So in total my input is made of three vectors (std::vector<std::vector<float>>). `T* evalModel(std::vector<T>& input)` doesn't work here, I tested it, because if `T` is a std::vector then the return should also be std::vector, while this method expects T* which will crash. Then `evalModel(std::vector<Ort::Value>& input)` might work, but I actually prefer to do all the ONNX procedure in  `model.h` and not the user code.

By the way, I tested everything and it works. It produces exactly the same results as tensorflow predict.

ChSonnabend:
Perfect, then it's fine from my side

hahassan7:
Thanks @ChSonnabend. In the end should I remove the guard `#if` or keep it ?

ChSonnabend:
> Thanks @ChSonnabend. In the end should I remove the guard `#if` or keep it ?

I leave the final decision up to @fcatalan92 but I would keep them in still for now

fcatalan92:
> Thanks @ChSonnabend. In the end should I remove the guard `#if` or keep it ?
You can keep it, and we will remove all the guards all together later on.
I'm approving but you need to solve the alibuild error

fcatalan92 (dismissed):

fcatalan92 (dismissed):

PWGJE/Core/JetTaggingUtilities.h
```diff
@@ -305,19 +323,48 @@ int16_t getJetFlavor(AnyJet const& jet, AllMCParticles const& mcparticles)
       if (dR < jet.r() / 100.f) {
         if (TMath::Abs(pdgcode) == 5) {
           return 2; // Beauty jet
-        } else {
```
nzardosh, line 308: Hi Hadi, whats the new logic on this part of the code?

nzardosh: why isnt 1 returned for charm quark directly in the loop if the pdgcode is 4 instead of setting charmQuark=true

PWGJE/Core/JetTaggingUtilities.h
```diff
@@ -305,19 +323,48 @@ int16_t getJetFlavor(AnyJet const& jet, AllMCParticles const& mcparticles)
...
+  for (auto& mcpart : mcparticles) {
+    int pdgcode = mcpart.pdgCode();
+    if (isBHadron(pdgcode) || isCHadron(pdgcode)) {
+      double dR = jetutilities::deltaR(jet, mcpart);
+
+      if (dR < jet.r() / 100.f) {
+        if (isBHadron(pdgcode)) {
+          return 2; // Beauty jet
+        } else if (isCHadron(pdgcode)) {
+          charmHadron = true;
```
nzardosh, line 360: same as above comment here

PWGJE/Tasks/bjetTaggingML.cxx
```diff
@@ -0,0 +1,592 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
```
nzardosh, line 2: why is there a separate task for ML here? why can the ML functionality not be added to the previous task

PWGJE/Tasks/bjetTreeMerger.cxx
```diff
@@ -0,0 +1,347 @@
...
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+/// \file bjetTreeMerger.cxx
+/// \brief Task for merging time frames with b-jet information. Can not be used on hyperloop
```
nzardosh, line 13: why do you need this task instead of using the merger tool locally?

hahassan7: Removed

PWGJE/Tasks/bjetTaggingML.cxx
hahassan7: This is different from the TreeCreator, this doesn't produce any trees or tables. I would like to keep the TreeCreator only for creating Trees so it stays clean and simple. And we apply the tagging in a different task. At some point later (after HP), this tagging will be done differently.

nzardosh (dismissed):

PWGJE/Core/JetTaggingUtilities.h
hahassan7: This copied from AliPhysics with modification. We can't return a charm if finds it in the stack. Because it could be that there is a case in which the charm appears first in the stack and then comes the beauty, then this will be mistaken as a charm. To be checked later if this is really the case in O2.

hahassan7:
@fcatalan92 Can you please approve this pull request again ? I didn't make any changes in the `model.h` in the last commit. Thanks

fcatalan92 (approved):

nzardosh (approved):

PR #7312, opened 2024-08-14
Common: add multNGlobalTracks in multiplicity table

dsekihat:
Common: add multNGlobalTracks in multiplicity table

### DIFF
--- a/Common/DataModel/Multiplicity.h
+++ b/Common/DataModel/Multiplicity.h
@@ -72,6 +72,9 @@ DECLARE_SOA_COLUMN(MultNTracksITSTPC, multNTracksITSTPC, int);       //!
 DECLARE_SOA_COLUMN(MultAllTracksTPCOnly, multAllTracksTPCOnly, int); //!
 DECLARE_SOA_COLUMN(MultAllTracksITSTPC, multAllTracksITSTPC, int);   //!
 DECLARE_SOA_COLUMN(MultNTracksGlobal, multNTracksGlobal, int);       //!
+DECLARE_SOA_COLUMN(MultNGlobalTracksPV, multNGlobalTracksPV, int);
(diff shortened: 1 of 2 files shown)


ddobrigk:
Hi @dsekihat , from what I can see this adds a dependency on track selection in all process functions. Did I understand that correctly? That is probably not a good feature, since we avoided that dependency on purpose... we should try to compartmentalize this or we will have to even change a lot of dependencies in hyperloop. Can you please adapt? Thank youo!

dsekihat:
Hi @ddobrigk , I agree. I don't like the extra dependency, either. Do you have any idea about how to compartmentalize them?

dsekihat:
Hi @altsybee , Do you mind, if I modify your process function "processGlobalTrackingCounters"?

ddobrigk:
> Hi @altsybee , Do you mind, if I modify your process function "processGlobalTrackingCounters"?

Hi @dsekihat , yes, this is what I would suggest and in fact this extra process function is exactly disabled by default to avoid creating that dependency. You could add your counters also to the table produced by that process function or so. Can you look into that? Thank you!

dsekihat:
Dear @ddobrigk and @altsybee , Thank you for your suggestions. I modified "processGlobalTrackingCounters". Can you check it again? Thank you.

altsybee:
> Hi @altsybee , Do you mind, if I modify your process function "processGlobalTrackingCounters"?

Hi @dsekihat , I think I never touched this file, and the processGlobalTrackingCounters is not "mine" ..  :)

dsekihat:
Hi @altsybee , I see. Thank you for clarification. I thought it is yours, because I found a comment "counter from Igor".
Anyway, nGlobalTracks is not modified.

ddobrigk (approved):
thanks, this looks good and now it's nicely modular!

PR #7314, opened 2024-08-15
feat(EventFiltering/Zorro): Add getTriggerOfInterestResults method

mpuccio:
The changes add a new method `getTriggerOfInterestResults` to the `Zorro` class. This method returns a vector of boolean values indicating which Triggers of Interest (TOIs) were selected in the last run of the `isSelected` method. This functionality is useful for providing the results of the TOI selection to the users.


alibuild (approved):

PR #7329, opened 2024-08-16
Add codes for PP flow analysis.

JiangTao-thu:
Add `FlowPbPbTask.cxx` for PP flow analysis and modify the corresponding `CMakeLists.txt`.

### DIFF
--- a/PWGCF/Flow/Tasks/CMakeLists.txt
+++ b/PWGCF/Flow/Tasks/CMakeLists.txt
@@ -38,3 +38,8 @@ o2physics_add_dpl_workflow(flow-pbpb-pikp-task
                     SOURCES FlowPbPbpikp.cxx
                     PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::AnalysisCore O2Physics::GFWCore
                     COMPONENT_NAME Analysis)
+
+o2physics_add_dpl_workflow(flow-pp-task
+                    SOURCES FlowPPTask.cxx
+                    PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::AnalysisCore O2Physics::AnalysisCCDB O2Physics::GFWCore
+                    COMPONENT_NAME Analysis)
--- a/PWGCF/Flow/Tasks/FlowPPTask.cxx
+++ b/PWGCF/Flow/Tasks/FlowPPTask.cxx
@@ -1,0 +1,700 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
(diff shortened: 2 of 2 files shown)


victor-gonzalez:
Is there any reason for not using the already existing `FlowPbPb` task?
The code is exactly the same plus a few additions that probably are also valid for the `FlowPbPb` task
The name of the existing task `FlowPbPb` is misleading because nothing prevents its use also for pp
In some cases fine tuning of histogram axes, if needed, can be done by mean of a configurable but nothing extra should in principle be needed

I'm in favor of adding the modified/incorporated lines (two or three) to the existing task and change its name to `Flow` or something like that

JiangTao-thu:
Hello Victor, indeed, the current version of the code is almost the same as the `FlowPbPb` version, but the subsequent analysis directions are different. Moreover, the current `FlowPbPb` code cannot be applied to the pp system. After discussing with Zhiyong, we decided to separate them to facilitate future changes.

Luzhiyongg:
> Is there any reason for not using the already existing `FlowPbPb` task? The code is exactly the same plus a few additions that probably are also valid for the `FlowPbPb` task The name of the existing task `FlowPbPb` is misleading because nothing prevents its use also for pp In some cases fine tuning of histogram axes, if needed, can be done by mean of a configurable but nothing extra should in principle be needed
> 
> I'm in favor of adding the modified/incorporated lines (two or three) to the existing task and change its name to `Flow` or something like that

Hi victor,
    Currently there is no different cut between pp and PbPb, but I'm afraid some special cuts may be added separately in pp or PbPb in the future. The `FlowPbPb` task I am using is under an PbPb analysis and I will add PbPb cut in this code. At that moment, this task is not a general flow task.

victor-gonzalez:
That's not the principle under which O2Physics was conceived
The idea is precisely having a single task which can be configured in different ways for doing different analysis
This principle was chosen on the base that replicating code is error prone and difficult to maintain and evolve

The whole execution scenario under hyperloop supports that

You have a single task but you have multiple wagons defined on it 
Each wagon has its complete own configuration
You can define a set of cut values to be used for pp analyses and you will have a wagon for your pp analyses
You can define a completely different set of cut values to be used for PbPb analyses and you will have a wagon for your PbPb analyses

So I strongly recommend starting with what is there and evolving it until there is something that really forbids keeping a single task

Just for you to know, in most of O2Physics analyses people usually use a single task for pp, PbPb, MC, and converted Run1/Run2 data

Of course there are things that are impossible but let's face that things

Luzhiyongg:
Hi victor,
    Thanks for your explanation. I agree with you that we should use a single script since we just change the system from PbPb to pp.
    I will modify the code to be more general and help @JiangTao-thu to run it on pp dataset.

victor-gonzalez:
Thanks!
Please, let me know in case you run in trouble and need any help
Thanks again!

PR #7330, opened 2024-08-16
[PWGEM]: efficiency task add Selection for injected events

feisenhu:
Add cfg to only select injected Events, use subGenerator for selection


alibuild (approved):

PR #7332, opened 2024-08-16
PWGHF: change track table for LF daughters

ZFederica:
Change table used to load information about cascade/V0 daughters (candidateCreator and candidateSelectors): Tracks -> TracksIU
@mfaggin

### DIFF
--- a/PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx
+++ b/PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx
@@ -1024,7 +1024,6 @@ struct HfCandidateCreatorXic0Omegac0 {
   void processOmegacToOmegaPiWithKFParticle(aod::Collisions const& collisions,
                                             aod::BCsWithTimestamps const& bcWithTimeStamps,
                                             MyKfTracks const& tracks,
-                                            MyLFTracksWCov const &lfTracks,
                                             MyKfCascTable const& cascades,
                                             KFCascadesLinked const& cascadeLinks,
                                             aod::HfCascLf2Prongs const& candidates)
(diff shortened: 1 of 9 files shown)


PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx
```diff
@@ -317,9 +319,9 @@ struct HfCandidateCreatorXic0Omegac0 {
       }
       auto casc = cascAodElement.cascData_as<MyCascTable>();
       hCascadesCounter->Fill(1);
-      auto trackCascDauCharged = casc.bachelor_as<TracksWCovDca>(); // pion <- xi track
-      auto trackV0Dau0 = casc.posTrack_as<TracksWCovDca>();         // V0 positive daughter track
-      auto trackV0Dau1 = casc.negTrack_as<TracksWCovDca>();         // V0 negative daughter track
+      auto trackCascDauCharged = casc.bachelor_as<MyLFTracksWCov>(); // pion <- xi track
+      auto trackV0Dau0 = casc.posTrack_as<MyLFTracksWCov>();         // V0 positive daughter track
+      auto trackV0Dau1 = casc.negTrack_as<MyLFTracksWCov>();         // V0 negative daughter track
```
mfaggin, line 325: Casting your tracks to `MyLFTracksWCov`, then your `pseudorapV0Dau0`, `pseudorapV0Dau1` and `pseudorapCascBachelor` are taken from the IU point ofthe track. Viceversa, casting them to `TracksWCovDca` as you were doing before means that you take the eta values from the PV.
None of this is correct, since you need the eta of these tracks at the Xi vertex. You should probably use the getters from the LF data model (see here https://github.com/AliceO2Group/O2Physics/blob/master/PWGLF/DataModel/LFStrangenessTables.h#L1096-L1104)

Actually, for eta this should not create a big issue, but for pt it does so be careful. I do not see other places where this issue happens, but be sure not to take any variable of the daughter at the wrong vertex (i.e. use the methods from the LF data models, but it seems that you are using them correctly in the other places)

mfaggin (commented):
Ciao Federica,

I have one major comment (easy to be solved). Try to cross check once more if the issue that I highlight is somewhere else in your code (I have not found it anywhere else, I hope not to have skipped any)

mfaggin (approved):

PR #7334, opened 2024-08-16
PWGHF: Automatically detect centrality estimator using C++20 concepts

fchinu:
This PR exploits C++20 `concepts` to automatically detect which centrality estimator to use.

This simplifies the code, as it is not anymore necessary to use many `if constexpr`s to choose the correct centrality estimator, and it not necessary to "transport" the information on which detector to use via, for example, function templates.


mfaggin (approved):
thanks @fchinu for the very nice implementation. We should take it as a good practice :)

PR #7337, opened 2024-08-16
[PWGDQ] reduce number of standard selected histos

feisenhu:
Introduce subgroup for PID histograms using postCalibration


alibuild (approved):

PR #7345, opened 2024-08-18
Updated EMCAL event selection for Full Jet Finders

Archita-Dash:
In this PR, I have updated the EMCAL event selection for the full jet finder processes functions. Now the fulljet finder at the DATA and MCD (also for the MCD to MCP matching) should only see those events that are readout by the EMCAL. For clarity, I have added more bin labels in the event histograms to clearly distinguish among the selected events in each process functions.


mfasDa (approved):
Looks good!

PR #7352, opened 2024-08-19
[PWGEM]: efficiency task add cfg to set subGenID being used

feisenhu:
Add cfg to provide a list of subGenerator being selected


alibuild (approved):

PR #7353, opened 2024-08-19
[PWGDQ] Implement time from SOR in VarManager and update definition of SEL8

iarsene:
1) implement the time from start of run as a variable in VarManager
2) modify the VarManager SEL8 to include also the cut on the ITS ROF border and the time frame border cut, similar to the event selection


alibuild (approved):

PR #7387, opened 2024-08-22
Add trigger for double cascades

fmazzasc:
@mpuccio , @ChiaraDeMartin95 , @ercolessi 

I have also removed the processRun2 function, which was not up to date.
Could you please have a look at this?


EventFiltering/PWGLF/strangenessFilter.cxx
```diff
@@ -1044,19 +795,43 @@ struct strangenessFilter {
...
     }
 
-    // 4Xi trigger definition
-    if (xicounter > 3) {
-      keepEvent[4] = true;
+    // Omega + Xi trigger definition
+    if (v0sFromOmegaID.size() > 0 && v0sFromXiID.size() > 0) {
+      std::set<std::array<int, 2>> uniqueOmegas = {v0sFromOmegaID.begin(), v0sFromOmegaID.end()};
+      std::set<std::array<int, 2>> uniqueXis = {v0sFromXiID.begin(), v0sFromXiID.end()};
+      if (uniqueOmegas.size() != 0 && uniqueXis.size() != 0) {
```
mpuccio, line 824: I am not sure this makes sense: if the uniqueOmegas and uniqueXis are each 1 then you will get the trigger true. I think here the condition should be:
```cpp
if (uniqueOmegas.size() > 1 || uniqueXis.size() > 1) {
  keepEvent[11] = true;
}
```

mpuccio (approved):

PR #7404, opened 2024-08-23
PWGHF: Reduce code repetitions in Ds Task

fchinu:
This PR reduces the amount of repeated code in the Ds Task. This facilitates future modifications and reduces the possibility to introduce bugs in the code

- The code in the different process functions has been moved to the `runMcAnalysisPerCollision` and `runDataAnalysisPerCollision` functions, and the partition for the candidates per collision is now passed as an argument to these functions
- The `hasDsMlInfo` concept is now added to automatically infer if the ML table is available, so that this information is not required to be "transported" across the different functions as a template argument. Two `fillSparse` functions have been added to fill the mass sparse, and differ on the requirements on the ML information


fgrosa (approved):
Thanks @fchinu, very nice reduction of code!

PR #7452, opened 2024-08-27
clang-format: Insert a newline at end of file if missing.

vkucera:
New option which fixes the missing end of line, which is reported as error by `cpplint`.


vkucera:
@ktf Is it fine for you?

PR #7456, opened 2024-08-27
PWGCF: FemtoUniverse -- Adding reject PID for kaons + reco and truth for Phi

zchochul:


alibuild (approved):

PR #7458, opened 2024-08-27
PWGJE: restrict true jets to evnet selected level as well as reco jets

joonsukbae:
Due to the difference between event selections in Data and MC currently, only TVX-selected true jets seem to make the difference when unfolding Data and MC (response matrix: selMC -> kIsTVXtriggered != sel8 -> kIsTVXtriggered).

### DIFF
--- a/PWGJE/Tasks/jetfinderQA.cxx
+++ b/PWGJE/Tasks/jetfinderQA.cxx
@@ -736,7 +736,7 @@ struct JetFinderQATask {
     }
     if (checkMcCollisionIsMatched) {
       auto collisionspermcpjet = collisions.sliceBy(CollisionsPerMCPCollision, jet.mcCollisionId());
-      if (collisionspermcpjet.size() == 1 && jetderiveddatautilities::selectCollision(collisionspermcpjet.begin(), eventSelection)) {
+      if (collisionspermcpjet.size() >= 1 && jetderiveddatautilities::selectCollision(collisionspermcpjet.begin(), eventSelection)) {
         fillMCPHistograms(jet);
       }
     } else {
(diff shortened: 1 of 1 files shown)


nzardosh:
hi joonsuk, can you explain the reason for taking only non split collisions?

joonsukbae:
Hi Nima, there isn’t a strong reason for enforcing that confinement for split collisions. I was hoping it might reduce the discrepancy between the data and MC, as well as with Run 2. However, if it’s preferable to leave the split collisions as they are, I’ll ensure that their size remains equal to or larger than 1.

nzardosh:
To me i think we should not remove the split collisions as they are a real part of the analysis

nzardosh:
If its really a 1% effect then its not too bad, but i am surprised its not more

nzardosh:
Lets keep them for now and we can always do a test later by removing them

joonsukbae:
Alright! Then I just force McCollisions only to the event selection level. Thanks for the quick response:)

nzardosh:
Thank you too :) also here you are checking that the mcCollision is matched to a detector level collision which is selected? Is that the intention?

joonsukbae:
Yes, exactly. I though collision table used for slicing was already selected, but they are not. I think reducing the difference of selection schemes for true and reco info would make it more reasonable, particulary under the condition that we use different event selection for data and MC.

nzardosh:
I guess you would need to slice the Filtered<collisions> table to get the selections. But ok cool, I will approve it now then

joonsukbae:
Ahh, thanks a lot

nzardosh (approved):

PR #7459, opened 2024-08-28
[PWG-DQ] Implementation of a few TPC pileup estimators

iarsene:
1) TPC pileup estimators
2) VarManager fill function which takes as arguments the collision and the table of tracks


alibuild (approved):

PR #7476, opened 2024-08-28
PWGHF+JE: Differentiate D0 PID variables based on the expected species

nzardosh:
Save PID variables under names that reflect the expected species of a prong, regardless of the difference of position of the prong for D0 and D0 bar. This should allow to train the ML model with mixed D0 and D0 bar signal while improving the separation of the correct PID hypotheses and reflections.


vkucera (approved):

alibuild (approved):

PR #7500, opened 2024-08-29
[Mult] add pos Z per BC

njacazio:


njacazio:
Hi @ddobrigk I think we can start with this, for the MC I will have to think what's best

PR #7516, opened 2024-08-30
ALICE 3: minimal sels for no MC assoc added

ddobrigk:
@jesgum this adds minimal selections for us to be able to do Pb-Pb with zero MC requirements - crucially, a XiC mass window requirement was still missing. I've also added a flag that disables the filling of `MCharmCores` so as to not use memory excessively when not producing derived data and changed the key output histos to be THxDs instead of THxFs to accommodate giant data samples properly (the generation is still extremely fast).


ddobrigk:
@hscheid @njacazio can you please take a look? Thanks!!

njacazio (approved):

PR #7517, opened 2024-08-30
Fix cosPA calculation

romainschotter:
Fix cosPA calculation when using the cascade momentum at the prim. vtx (useCascadeMomentumAtPrimVtx = on), in order to always the cosPA computed using the decay daughters momenta


njacazio (approved):
Good catch

fmazzasc (approved):

PR #7519, opened 2024-08-30
[PWGEM] Add MC signal of direct B->c->e decay

feisenhu:
Adding an additional MC signal


alibuild (approved):

PR #7526, opened 2024-09-01
Add UPC info in the derived data producer & V0 analysis task

romainschotter:
- Add UPC info in the derived data producer 
      @ddobrigk, can you please have a look at this PR? As you can see, I have introduced dynamical columns in the StraRawCents table in order to be able to use the method trueGap from the SGSelector class (which requires specific getters...)
- Add UPC info in the V0 analysis + separation between UPC and hadronic collisions


romainschotter:
@ddobrigk and @nepeivodaRS can you please have a look at that PR? I included the use of UD tables to fill the strangeness tables. O2 compiles and the whole analysis chain (derived data + analysis task) works like a charm

njacazio (approved):

PR #7532, opened 2024-09-02
[EPDev] add occupancy study

jikim1290:
add occupancy study

### DIFF
--- a/Common/Tasks/qVectorsCorrection.cxx
+++ b/Common/Tasks/qVectorsCorrection.cxx
@@ -73,6 +73,8 @@ struct qVectorsCorrection {
 
   Configurable<bool> cfgQAAll{"cfgQAAll", false, "draw all q-vector steps"};
   Configurable<bool> cfgQAFinal{"cfgQAFinal", false, "draw final q-vector steps"};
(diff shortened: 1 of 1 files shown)


jikim1290:
Hi @strogolo @ddobrigk, I don't know why this is so much pending here. Could you please remind me of whatever you think about why this should not be merged? Note that this PR is not affecting anything on the current analyzers as this utilizes already produced tables.

dsekihat:
Dear @jikim1290 , Thank you for preparing this PR. Indeed, this is necessary. I have a minor thing about your occupancy bins. If occupancy is not calculated, the  default value -1. So, it is better to have [-1, 0, XXX]  in your occupancy bins when you submit your wagon.

jikim1290:
HI @dsekihat , Thanks for the comment, it is added

jgrosseo:
@ddobrigk @strogolo I will merge this unless there are objections until noon

strogolo:
@jikim1290 for me this PR is fine if the default is not to use the occupancy cut but it is just used for testing.
If this is the case, then @jgrosseo can approve and merge it.

jgrosseo (approved):

PR #7550, opened 2024-09-03
PWGHF: Update LcToPKPi tree creator

phymanshu:
Update HfCandLcFullPs table in treeCreatorLcToPKPi.cxx

   - Remove `McCollisionId` and `McParticleId` in HfCandLcFullPs to account for the Unbound indices problem.


fcatalan92 (changes requested):
Hi @phymanshu, the changes look good to me, I would just remove the code instead of commenting it.

fcatalan92 (approved):

PR #7556, opened 2024-09-04
PWGHF: D0selector Mass Cut from Triggers

apalasciano:
CandidateSelectorD0.cxx modified to allow selections from trigger class


fgrosa (approved):
Thanks!

PR #7575, opened 2024-09-04
[PWGDQ] TtCA, adding propagation of track to collisions on pair level

feisenhu:
Adding recalculation of pair values like DCA based on propagated tracks when using Track to Collision association.

@iarsene You can have a second look on this.


alibuild (approved):

PR #7577, opened 2024-09-04

yashpatley:


yashpatley:
Dear @victor-gonzalez , 

This is a temporary workaround for the current problem we face on hyperloop for not being able to create subwagons because we the previous task has derived data output. 

Please provide your valuable comments and suggestions so we can further modify this new version of the code. The part to update lambda list is still under construction. I will make a PR in future with that part modified. For the time being I want to run this one to get some results and comparison plots.

victor-gonzalez (approved):

victor-gonzalez (changes requested):
Yash, I aproved it by mistake before

victor-gonzalez:
The rule in O2 is to use the information that is stored in a really efficient way in the tables that are resident in memory and don't create new complex structures that mimic in some way that information. This structures have quite inefficient memory access and, of course, occupy memory 

I would not suggest to modify an already functional task because it cannot be run with subwagons
Subwagons are also limited resources

yashpatley:
Hello Victor,

I understand what you are implying. I will change the task as before. Thank you.

victor-gonzalez (approved):

PR #7580, opened 2024-09-05
[TOF] switch based on coll system

njacazio:
Reopening https://github.com/AliceO2Group/O2Physics/pull/7559


njacazio:
@noferini

njacazio:
@ddobrigk based on the GRP info I fetch if this is pp or Pb-Pb, this is a wider use case than the TOF. Could you please check if this is ok for e.g. the multiplicity case?

ddobrigk (approved):
approving, does not act in main core service but in alternate dev task AFAIK

ddobrigk:
and thank you @njacazio !

PR #7583, opened 2024-09-05
PWGJE: give a maximum-pT cut on jet constituents

joonsukbae:
Hi Nima,
Existing track pT max cut can't allow jet rejection after jets finding w/o track pT cut both in jetfinderQA nor jetfinder tasks, thus I added this selection in jet acceptance task.

### DIFF
--- a/PWGJE/Tasks/jetfinderQA.cxx
+++ b/PWGJE/Tasks/jetfinderQA.cxx
@@ -66,6 +66,7 @@ struct JetFinderQATask {
   Configurable<int> nBinsEta{"nBinsEta", 200, "number of bins for eta axes"};
   Configurable<float> jetAreaFractionMin{"jetAreaFractionMin", -99.0, "used to make a cut on the jet areas"};
   Configurable<float> leadingConstituentPtMin{"leadingConstituentPtMin", -99.0, "minimum pT selection on jet constituent"};
+  Configurable<float> leadingConstituentPtMax{"leadingConstituentPtMax", 200.0, "maximum pT selection on jet constituent"};
   Configurable<float> randomConeR{"randomConeR", 0.4, "size of random Cone for estimating background fluctuations"};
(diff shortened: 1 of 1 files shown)


nzardosh:
Hi Joonsuk, i think the function is cleaner like this:

template <typename T, typename U>
  bool isAcceptedJet(U const& jet)
  {

    if (jetAreaFractionMin > -98.0) {
      if (jet.area() < jetAreaFractionMin * M_PI * (jet.r() / 100.0) * (jet.r() / 100.0)) {
        return false;
      }
    }
    bool checkConstituentPt = true;
    bool checkConstituentMinPt = (leadingConstituentPtMin > -98.0);
    bool checkConstituentMaxPt = (leadingConstituentPtMax < 9998.0);
    if (!checkConstituentMinPt && !checkConstituentMaxPt) {
      checkConstituentPt = false;
    }

    if (checkConstituentPt){
      bool isMinLeadingConstituent = !checkConstituentMinPt;
      bool isMaxLeadingConstituent = true;

    for (const auto& constituent : jet.template tracks_as<T>()) {
      double pt = constituent.pt();

      if (checkConstituentMinPt && pt >= leadingConstituentPtMin) {
        isMinLeadingConstituent = true;
      }
      if (checkConstituentMaxPt && pt > leadingConstituentPtMax) {
        isMaxLeadingConstituent = false;
      }
    }
    return isMinLeadingConstituent && isMaxLeadingConstituent;
    }

    return true;
  }
  
  can you please check it works and change accoridngly?

nzardosh:
I think the problem with the early condition was that if we want to add a condition not related to the pT later then it would be skipped

joonsukbae:
Please ignore this fetch at 21:23. Because my O2 installation failed in local, I pulled upstream.

joonsukbae:
I checked your solution works very well in local, but due to some git conflics, I closed the commit.

joonsukbae:
I checked your solution and it works very well in local

PR #7624, opened 2024-09-09
Common: easier indexing to avoid HY warning

ddobrigk:


alibuild (approved):

PR #7633, opened 2024-09-09
Common: interlink MC extras with reco coll info

ddobrigk:
It seems this also requires to have an `s` at the end of the table that is being indexed, which is extremely unfortunate - and has required that I change the name of the table as well. `using` has been used to avoid backwards incompatibility.


alibuild (approved):

PR #7634, opened 2024-09-09
fixing abs rounding with fabs

glromane:
Dear @victor-gonzalez 

in this commit we're attempting to fix a bug I spotted recently: O2Physics tags after 15 Aug in the HL were producing strange distributions on k*  -- discrete values like the values that are supposed to be float were rounded to int. I didn't manage to reproduce the issue on my machine but I assume that might be due to the usage of abs() function that might return "int" or "float" depending on if it was overloaded, compiler settings etc. Maybe in the middle of August the building settings for O2 were changed and this caused the bug. Maybe in that case it can also be helpful to notify other analysers.

So, I changed it everywhere where floats needed to fabs(). On my local machine it doesn't make any difference (I used today's build of O2 on a Mac), but hopefully it'll solve the issue in the HL.

Sincerely yours, Gleb.

@njacazio @ercolessi

### DIFF
--- a/PWGCF/Femto3D/Core/femto3dPairTask.h
+++ b/PWGCF/Femto3D/Core/femto3dPairTask.h
@@ -109,14 +109,14 @@ float GetKstarFrom4vectors(TLorentzVector& first4momentum, TLorentzVector& secon
 {
   if (isIdentical) {
     TLorentzVector fourmomentadiff = first4momentum - second4momentum;
-    return 0.5 * abs(fourmomentadiff.Mag());
+    return 0.5 * fabs(fourmomentadiff.Mag());
   } else {
     TLorentzVector fourmomentasum = first4momentum + second4momentum;
     TLorentzVector fourmomentadif = first4momentum - second4momentum;
 
     fourmomentadif.Boost((-1) * fourmomentasum.BoostVector());
 
-    return 0.5 * abs(fourmomentadif.Vect().Mag());
+    return 0.5 * fabs(fourmomentadif.Vect().Mag());
   }
 }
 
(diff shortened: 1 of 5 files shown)


victor-gonzalez:
Hi Gleb,
thanks!

Interesting. 
I reported this issue (https://mattermost.web.cern.ch/alice/pl/w3z4a5q9uighzjhba1j179dwqc) on the Analysis channel without any success. But now with your information it starts to make sense. Our compiled unit was not touched since the middle of July and things worked fine regarding the DCA distribution. Then, on the 16 of August there was a PR from our side not related at all with the DCA stuff but which, for sure, forced the re-compilation of the compiled unit. Since then nothing has worked correctly regarding the DCA distribution. 

Did you reproduce the problem locally? 
Did your changes fix it?

@ktf @jgrosseo @ddobrigk, do we have any other information that could bring light on this?
Thanks!

@glromane, if the local tests showed it solved I will approve the PR. If not probably it would be better to wait a bit

jgrosseo:
The code is fully build every night, regardless of PRs or not.

Wouldn't `std::fabs` be even better?

jgrosseo:
@victor-gonzalez The things you can check are change of ROOT or gcc. @ktf may also know by heart... ;-)

victor-gonzalez:
@glromane @ktf @jgrosseo @ddobrigk I confirm the issue Gleb found, the fix is correct, and the issue also match the one I reported on the `O2Analysis` channels a few days ago

Since August 16 `abs` return an integer in Linux machines which I guess are most of the ones in the grid
It seems Macs are not affected

The central `TrackSelection` is also affected in the DCA cut
I will incorporated its fix as par of the same PR in which I will fix the correlations corresponding one

victor-gonzalez (approved):
Thanks Gleb!!!

PR #7647, opened 2024-09-10
Fix unexpected behavior in DCA cuts due to `abs` returning an integer

victor-gonzalez:
See also https://github.com/AliceO2Group/O2Physics/pull/7634


ddobrigk (approved):

alibuild (approved):

PR #7663, opened 2024-09-11
Common: switch debug interlinks to foundBC

ddobrigk:


alibuild (approved):

PR #7670, opened 2024-09-12
PWGDQ: Add the process for electron-muon analysis in the tableMaker

MaolinZH:
1. Add a new process for producing electron-muon table without TOF-related variables to avoid large memory consumption;
2. Group the cfgs of adding histograms to avoid reaching the element limits for a structure.

### DIFF
--- a/PWGDQ/TableProducer/tableMaker.cxx
+++ b/PWGDQ/TableProducer/tableMaker.cxx
@@ -95,6 +95,8 @@ using MyBarrelTracksWithV0AndDalitzBits = soa::Join<aod::Tracks, aod::TracksExtr
                                                     aod::pidTPCFullKa, aod::pidTPCFullPr,
                                                     aod::pidTOFFullEl, aod::pidTOFFullMu, aod::pidTOFFullPi,
                                                     aod::pidTOFFullKa, aod::pidTOFFullPr, aod::pidTOFbeta, aod::V0Bits, aod::DalitzBits>;
+using MyBarrelTracksForElectronMuon = soa::Join<aod::Tracks, aod::TracksExtra, aod::TrackSelection, aod::TracksDCA,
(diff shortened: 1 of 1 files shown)


PWGDQ/TableProducer/tableMaker.cxx
```diff
@@ -1326,6 +1339,13 @@ struct TableMaker {
     }
   }
 
+  // Produce barrel + muon tables for the eletron-muon analysis, without PIDTOF----------------------------------------------------------------------
+  void processFullForElectronMuon(MyEvents::iterator const& collision, aod::BCsWithTimestamps const& bcs,
+                                  soa::Filtered<MyBarrelTracksForElectronMuon> const& tracksBarrel, soa::Filtered<MyMuons> const& tracksMuon)
+  {
+    fullSkimming<gkEventFillMap, gkTrackFillMapForElectronMuon, gkMuonFillMap>(collision, bcs, tracksBarrel, tracksMuon, nullptr, nullptr);
+  }
+
```
iarsene, line 1348: Hi Maolin, I am not sure I understand the need of this process function. Could you explain? One can get already the muons and electrons in the data model with the other "Full" process functions.

MaolinZH: Hi Ionut, this is our current solution for the memory consumption problem on the hyperloop. With other full processes, the memory consumption is always larger than the limit 3.6 G. So I add this new "slight" process with only TPCPID, and remove all the unnecessary PID dependencies.

iarsene: Hi Maolin, ok i will approve this, but we have to look more into this. Not having TOF in the skims can be problematic if the TPC pid is not enough.

iarsene (approved):

PR #7674, opened 2024-09-12
[PWGDQ] Adding control histograms for zorro

lucamicheletti93:
@iarsene 
This modification requires the inclusion of the histogramRegistry, so that the automatic zorro histograms are produced in the table-maker. The same modification will be done in the table-maker-with-association after few checks.


alibuild (approved):

PR #7675, opened 2024-09-13
Calculate cascade flow using GFW

fuchuncui:


PWGCF/Flow/Tasks/flowGFWOmegaXi.cxx
```diff
@@ -0,0 +1,409 @@
...
+    int nMassBins = 0;
+    TAxis* fMass = nullptr;
+    if (partical == 1) {
+      nMassBins = nXiMassBins;
+      fMass = fXiMass;
+    } else if (partical == 2) {
+      nMassBins = nOmegaMassBins;
+      fMass = fOmegaMass;
+    } else {
+      std::cout << "Error, partical = 1 for Xi and 2 for Omega" << std::endl;
```
victor-gonzalez, line 225: Please, no `std::cout`, use `LOGF(error, ...)` or `LOGF(fatal, ...)`

PWGCF/Flow/Tasks/flowGFWOmegaXi.cxx
```diff
@@ -0,0 +1,409 @@
...
+      mEfficiency = ccdb->getForTimeStamp<TH1D>(cfgEfficiency, timestamp);
+      if (mEfficiency == nullptr) {
+        LOGF(fatal, "Could not load efficiency histogram for trigger particles from %s", cfgEfficiency.value.c_str());
+      }
+      LOGF(info, "Loaded efficiency histogram from %s (%p)", cfgEfficiency.value.c_str(), (void*)mEfficiency);
+    }
+    correctionsLoaded = true;
+  }
+
+  bool setCurrentParticleWeights(float& weight_nue, float& weight_nua, float phi, float eta, float pt, float vtxz)
```
victor-gonzalez, line 263: It is more efficient to template this function passing a track object (reco track or detector level track) and use it directly without storing its `eta`, `phi`, and `pT` values. Let the compiler to decide that

victor-gonzalez (changes requested):
Please have a look at my suggestions

PWGCF/Flow/Tasks/flowGFWOmegaXi.cxx
```diff
@@ -0,0 +1,410 @@
...
+      mEfficiency = ccdb->getForTimeStamp<TH1D>(cfgEfficiency, timestamp);
+      if (mEfficiency == nullptr) {
+        LOGF(fatal, "Could not load efficiency histogram for trigger particles from %s", cfgEfficiency.value.c_str());
+      }
+      LOGF(info, "Loaded efficiency histogram from %s (%p)", cfgEfficiency.value.c_str(), (void*)mEfficiency);
+    }
+    correctionsLoaded = true;
+  }
+
+  template <typename aodTracks>
```
victor-gonzalez, line 263: Could a different `typename` name for the template parameter be used?
That name has been already used in line 110 for your filtered tracks table
Also the passed object is a single one so its type name should not be plural

victor-gonzalez (changes requested):

victor-gonzalez (approved):
Thanks for considering my comments
For the next iteration I would suggest as `typename` `TrackObject` because it is a track what is being passed and not a track weight. It would be easier for reading and understand the function functionality

PR #7678, opened 2024-09-13
Common: save only non-empty BCs, add BC flags creator

ddobrigk:


alibuild (approved):

PR #7682, opened 2024-09-13
Added a Fake V0 Filter

JStaa:


victor-gonzalez (approved):
I overlooked this one thinking that was related with the other you opened not related to CF. Sorry about that
Please, for the next time, if you see no reaction from my side within two days, ping me. Don't wait that long

PR #7693, opened 2024-09-16
DPG: AOTTrack: added track position at IU info

scattaru:
Dear all,
I have added the track position at inner most update information. Let me know for any comment.

Cheers,
Samuele

### DIFF
--- a/DPG/Tasks/AOTTrack/qaImpPar.cxx
+++ b/DPG/Tasks/AOTTrack/qaImpPar.cxx
@@ -65,12 +65,16 @@ struct QaImpactPar {
   ConfigurableAxis binningPhi{"binningPhi", {24, 0.f, TMath::TwoPi()}, "Phi binning"};
   ConfigurableAxis binningPDG{"binningPDG", {5, -1.5f, 3.5f}, "PDG species binning (-1: not matched, 0: unknown, 1: pi, 2: K, 3: p)"};
   ConfigurableAxis binningCharge{"binningCharge", {2, -2.f, 2.f}, "charge binning (-1: negative; +1: positive)"};
+  ConfigurableAxis binningIuPosX{"binningIuPosX", {100, -10.f, 10.f}, "Track IU x position"};
+  ConfigurableAxis binningIuPosY{"binningIuPosY", {100, -10.f, 10.f}, "Track IU y position"};
+  ConfigurableAxis binningIuPosZ{"binningIuPosZ", {100, -10.f, 10.f}, "Track IU z position"};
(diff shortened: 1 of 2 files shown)


DPG/Tasks/AOTTrack/qaImpPar.cxx
```diff
@@ -168,13 +173,14 @@ struct QaImpactPar {
   void processMC(o2::soa::Filtered<collisionMCRecoTable>::iterator& collision,
                  trackTable const& tracksUnfiltered,
                  o2::soa::Filtered<trackMCFullTable> const& tracks,
+                 const o2::aod::TracksIU& tracksIU,
```
mfaggin, line 176: in this way the `TracksIU` table is not filtered, is it what you want?

DPG/Tasks/AOTTrack/qaImpPar.cxx
```diff
@@ -154,11 +158,12 @@ struct QaImpactPar {
   void processData(o2::soa::Filtered<collisionRecoTable>::iterator& collision,
                    const trackTable& tracksUnfiltered,
                    const o2::soa::Filtered<trackFullTable>& tracks,
+                   const o2::aod::TracksIU& tracksIU,
```
mfaggin, line 161: in this way the `TracksIU` table is not filtered, is it what you want?

mfaggin (commented):
ciao @scattaru all good for me. I've just added a comment in the code just to be sure that what you added is what you need. But from the code point of view it is fine

DPG/Tasks/AOTTrack/qaImpPar.cxx
scattaru: Ciao @mfaggin, indeed this is one of my doubts I would prefer to filter the table in order to reduce its size to speed up the execution. Is it equivalent applying the track filter defined in the code on the track table with track parameters at collision vertex or on the track table with track parameters at IU? Just to be sure, because I was wandering that some parameters could have different values among the two tables

mfaggin: Ciao @scattaru , indeed this is a good point especially because you extract `x`, `y`, `z` for the `trackIU` associated to each selected global track, and filtering the `tableIU` you might miss the corresponding `trackIU` for your global track (and instead if you analyse a `trackIU` not associated to any global track the `x`, `y`, `z` remain `-999.f`, so you recognize them offline).
Fair enough, let's go ahead in this way. I approve it.

mfaggin (approved):

PR #7701, opened 2024-09-16
PWGLF: optionally require trigger particle as ev sel

ddobrigk:
@lhusova @ChiaraDeMartin95 @kcui1 will still cross-check in the morning ...


ddobrigk:
This is now ready to be merged. It adds the possibility to require a trigger particle within a given (configurable) bin of the `pTtrigger` axis and makes that requirement in both efficiency numerators and denominators. It is not likely that we can truly extract a 2D efficiency with these requirements, but presumably, studying any global factor that impacts the efficiency as a whole (or studying how constant it is with respect to the trigger particle presence requirement) is already possible if we integrate eta out and look only at efficiency vs pT...

ChiaraDeMartin95 (approved):

PR #7722, opened 2024-09-18
PWGHF: Add finding leading particle in correlatorD0Hadrons for both McGen and McRec

1481014945:
Hello，experts.
Sorry for bothering you again.I would like to add functionality to find the leading particle in correlatorD0Hadrons for both McGen and McRec. Can you give me some suggestions?


fgrosa (approved):
Hi @1481014945, it looks good to me!

PR #7736, opened 2024-09-19
[STS] add converter

njacazio:
adds a converter to move to the new format from v1 and fixes the unpacking
@glromane @sofiatomassini @ercolessi


njacazio:
@victor-gonzalez ready to merge, this handles the conversion and a fix for the dynamic columns only

victor-gonzalez (approved):

PR #7738, opened 2024-09-19
[PWGDQ ]Adding exceptions to exit the init in unused process functions [PWGDQ-83]

lucamicheletti93:
@iarsene  
Addition of exceptions to exit the init in unused process functions. This should allow to reduce the memory usage for the various tasks.
If it works it will be extended to all the tasks.


alibuild (approved):

PR #7743, opened 2024-09-20
PWGHF/HFL:  Add 2Dhistogram and remove Thnsparce in electronselection Table

rashigupt:
remove some THnsparce and add 2Dhistogram  in electronselection Table


mfaggin:
Here a `rebase` is required before merging

PR #7784, opened 2024-09-24
[FT0] add bypass of FT0 in MC

njacazio:


njacazio:
tagging FT0 @afurs @andreasmolander

Common/TableProducer/ft0CorrectedTable.cxx
```diff
@@ -10,38 +10,86 @@
...
+    histos.add("t0C", "t0C", kTH1D, {{1000, -1, 1, "t0C (ns)"}});
+    histos.add("t0AC", "t0AC", kTH1D, {{1000, -1000, 1000, "t0AC (ns)"}});
+    histos.add("deltat0AC", "deltat0AC", kTH1D, {{1000, -10, 10, "#Deltat0AC (ns)"}});
+    if (doprocessWithBypassFT0timeInMC) {
+      histos.add("MC/deltat0A", "t0A", kTH1D, {{1000, -50, 50, "t0A (ps)"}});
+      histos.add("MC/deltat0C", "t0C", kTH1D, {{1000, -50, 50, "t0C (ps)"}});
+      histos.add("MC/deltat0AC", "t0AC", kTH1D, {{1000, -50, 50, "t0AC (ps)"}});
+    }
+  }
+  void process(aod::BCs const&) {}
```
afurs, line 74: Why it is empty? for which purpose we keep it?

afurs: I see that it is not registered by `PROCESS_SWITCH` macro, but maybe better just to remove it?

njacazio: That's on purpose, in case no process function is called the task doesn't do anything

afurs: > That's on purpose, in case no process function is called the task doesn't do anything

but there are always at least one task should be called, otherwise this executable should appear in `o2-` chain. Or is it to keep executable presence in chain but change functionality via json cfg?

njacazio: I removed the default case

afurs (approved):

ddobrigk (approved):

PR #7798, opened 2024-09-26
PWGJE: Custom DCA cuts for quality tracks

KotliarovAr:
Due to an issue with the phi distribution of high pT tacks, it was proposed to study tracks with relaxed conditions on the DCA. 
The updated code implements functionality to adjust DCA cuts for the quality tracks:
1) DcaZ cut is const and it can be tuned in JSON file
2) DcaXY cut has pT dependence of a track (taking here: https://github.com/AliceO2Group/O2Physics/blob/master/Common/Core/TrackSelectionDefaults.cxx#L37); Additional multiplication factor is introduced to relax this condition


KotliarovAr:
@fkrizek As an owner of the code, could you review it and give your consent?

KotliarovAr:
@nzardosh Could you approve the code?

nzardosh (approved):

PR #7800, opened 2024-09-26
PWGCF: Fixing configurables for efficiency 

zchochul:


victor-gonzalez:
Hi @zchochul, have you checked it locally?
If reordering configurables did work that means that, after reordering, other configurables will be affected although perhaps you don't see the effect now because you are using their default values

zchochul:
Hi @victor-gonzalez, you are right I will put configurables in the structures. Thank you!

alibuild (approved):

PR #7816, opened 2024-09-27
PWGLF: added flag for filling histograms and code for SP polarization

prottayCMT:


prottayCMT:
@BongHwi need your approval here as well.

PR #7839, opened 2024-10-01
[Ev. Sel.] handle Run2 vs Run3

njacazio:


njacazio:
@ekryshen @ddobrigk as we discussed this could be the way to handle the Run2 vs Run3, it might not be the prettiest as it involves some duplication and we have to see how this is handled by HL (therefore I copied the configuration in all tasks) but this could be a way to go.

PR #7851, opened 2024-10-02
Modify track selection

ilikmeta:


PWGCF/Flow/Tasks/FlowGFWPbPb.cxx
```diff
@@ -392,25 +400,28 @@ struct FlowGFWPbPb {
...
+    if (occupancy < cfgCutOccupancyLow || occupancy > cfgCutOccupancyHigh)
+      return 0;
 
     // V0A T0A 5 sigma cut
     if (abs(collision.multFV0A() - fT0AV0AMean->Eval(collision.multFT0A())) > 5 * fT0AV0ASigma->Eval(collision.multFT0A()))
-      return false;
+      return 0;
 
-    return true;
+    return 1;
```
victor-gonzalez, line 424: Please, for the next iteration, keep the `bool` returns
Your function returns a boolean and it is more readable if `true` and `false` are in place and not `0` or `1`

victor-gonzalez (approved):
I approve it but please have a look at my comment for the next iteration

PWGCF/Flow/Tasks/FlowGFWPbPb.cxx
ilikmeta: Ahh my bad. I remember that you had mentioned this in the past. I will fix that in the next iteration, thanks!

PR #7853, opened 2024-10-03
Add nSigmaMC.

mfaggin:
Adding definition of missing histo.
NB for @njacazio : beware that, as written now, for `nSigmaMC` histogram the code always fills `hnsigma/El/*` histograms. This is because you loop from 0 to 8 here https://github.com/AliceO2Group/O2Physics/blob/master/DPG/Tasks/AOTTrack/PID/TPC/qaPIDTPCMC.cxx#L580-L583 and therefore always these histos are filled: https://github.com/AliceO2Group/O2Physics/blob/master/DPG/Tasks/AOTTrack/PID/TPC/qaPIDTPCMC.cxx#L83-L85

I'm not fixing it, I'd leave it to you to avoid any clash in case somebody is massively using it in your crew.


alibuild (approved):

njacazio (approved):

PR #7857, opened 2024-10-03
PWGHF: taskFlow.cxx: Add process functions for data TPC-MFT HF-h case, and MC Rec/Gen for TPC-TPC h-h case

AlexianL:
QA functions and histograms for the data TPC-MFT HF-h case were also added
In both TPC-MFT h-h and TPC-MFT HF-h case, histograms have been more clearly separated to investigate tracks/candidates from the TPC and tracks from the MFT
QA functions and histograms for the MC Rec/Gen TPC-TPC h-h case were also added
Also mixCollisions function had to be duplicated to be adapted for MC Gen
Added invariant mass plot for D0


vkucera:
Duplicate of https://github.com/AliceO2Group/O2Physics/pull/7857
Please update the original PR instead of creating a new one.

PR #7864, opened 2024-10-03
Added intermediate loop to match collisions with high-pt tracks

ekryshen:


alibuild (approved):

PR #7867, opened 2024-10-03
[PWGJE] V0 QA: Collision association

GijsvWeelden:
* Added process for collision association of V0s in jets and Xis in jets

### DIFF
--- a/PWGJE/Tasks/v0qa.cxx
+++ b/PWGJE/Tasks/v0qa.cxx
@@ -144,6 +144,36 @@ struct V0QA {
       registry.add("collisions/XiPlusPtYAntiLambdaPt", "#Xi^{+} Pt, Y, #bar{#Lambda} Pt", HistType::kTH3D, {axisV0Pt, axisEta, axisV0Pt});
       registry.add("collisions/XiPlusPtYAntiLambdaPtWrongColl", "#Xi^{+} Pt, Y, #bar{#Lambda} Pt, wrong collision", HistType::kTH3D, {axisV0Pt, axisEta, axisV0Pt});
     }
+    if (doprocessCollisionAssociationJets) {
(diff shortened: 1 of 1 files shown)


nzardosh:
Hi Gijs,

Here are you aiming to do the reassociation only for the V0s or all particles in the jet?

Thanks,
Nima

GijsvWeelden:
Hi Nima,
The collision association is done for V0s and Xis in jets (the inclusive case was already included in the task).
I want to check how large the effect is of V0s being assigned to the wrong collision.
The method is based on the `derivedlambdakzeroanalysis` task in LF

nzardosh:
Hi Gjis, just to understand fully. Are the V0s coming from the LF tasks not already accounting for the possible reassociations? Because if they are then this is already taken care for by defaukt fir the jets

GijsvWeelden:
When reconstructing V0s, they are not a priori required to be matched to a particular collision, [see lambdakzerofinder](https://github.com/AliceO2Group/O2Physics/blob/6bc9e02bdbe83af9e9079af9a2059782586f8c63/PWGLF/TableProducer/Strangeness/lambdakzerofinder.cxx#L272). When matching reconstructed V0s with generated K0S, Lambda(bar), the collision of the V0 is also not required to correspond to the appropriate MC collision, [see lambdakzeromcbuilder](https://github.com/AliceO2Group/O2Physics/blob/6bc9e02bdbe83af9e9079af9a2059782586f8c63/PWGLF/TableProducer/Strangeness/lambdakzeromcbuilder.cxx#L247).
This means that reconstructed V0s may be associated with collisions that are unrelated to the actual MC event they originate from. This is by design, to account for the finite lifetime of V0s.

With the collision association processes, I aim to quantify this effect of V0s being associated with the "wrong" collision. I also want to check whether this is different for reconstructed V0s that end up being clustered into ch+V0 jets

PR #7870, opened 2024-10-04
PWGCF - New task for zdc energy gain equalisation and calculating q-vectors

cnkoster:
Task will be used to calculate and recenter Q-vectors in the ZDC. 
The different steps to recenter the q-vectors will be added one by one.


victor-gonzalez:
Dear @cnkoster ,
the QVector functionality is being added at the `Common` level and not at the `PWG` level
Could you please check with @jikim1290 and @strogolo if the functionality you wanted to add is already implemented?
If not, then you can implement it there according to the structure used for other detectors
If this were the case, functionality not implemented, I would also suggest to give a presentation about it in Tursday's DPG AOT meetings
Thanks!

cnkoster:
Dear @victor-gonzalez, 
Thank you for the comment, I will contact them.

PR #7871, opened 2024-10-04
[DPG] add check on the abs

njacazio:
@victor-gonzalez


njacazio:
Unable to test Mac but failing standard build. Closing

PR #7883, opened 2024-10-04
Add collision flags to StraEvSels derived table

romainschotter:
Add collision flags to the StraEvSels derived table in order to allow the separation between events with and without UPC mode


romainschotter:
Add converters for ML scores

njacazio (approved):

PR #7885, opened 2024-10-04
PWGHF: Adding centrality information in mc-validation task

stefanopolitano:
This PR adds the histograms storing the centrality information at the generation and reconstruction levels.


PWGHF/Tasks/taskMcValidation.cxx
```diff
@@ -497,6 +512,7 @@ struct HfTaskMcValidationRec {
   Configurable<int> eventGeneratorType{"eventGeneratorType", -1, "If positive, enable event selection using subGeneratorId information. The value indicates which events to keep (0 = MB, 4 = charm triggered, 5 = beauty triggered)"};
 
   std::array<std::shared_ptr<TH1>, nChannels> histDeltaPt, histDeltaPx, histDeltaPy, histDeltaPz, histDeltaSecondaryVertexX, histDeltaSecondaryVertexY, histDeltaSecondaryVertexZ, histDeltaDecayLength;
+  std::array<std::array<std::shared_ptr<TH2>, 2>, nChannels> histPtCentReco;
   std::array<std::array<std::array<std::shared_ptr<TH1>, 5>, 2>, nChannels> histPtDau, histEtaDau, histImpactParameterDau;
   std::array<std::array<std::shared_ptr<TH1>, 2>, nChannels> histPtReco;
```
fgrosa, line 517: This one should not be needed anymore once we have those with pT vs cent (pT can be projected from the 2D)

fgrosa (changes requested):
Hi @stefanopolitano the PR looks good to me, I would just remove one output since there are already a huge number of histograms in the output (see below)

PWGHF/Tasks/taskMcValidation.cxx
stefanopolitano: Hi @fgrosa ! Thank you for the suggestion, I will remove the TH1. Thanks!

PR #7886, opened 2024-10-04
PWGHF: fix L(1520) PDG code.

mfaggin:
Fixing Lambda(1520) PDG code in candidateCreator3Prong.cxx, aligning it to the expected value, as done in O2DPG here: [PR #1673](https://github.com/AliceO2Group/O2DPG/pull/1673).

Tested on 1 file of `HF_LHC24g5_All` MC production (see bin centered at 3 below):

image

### DIFF
--- a/PWGHF/TableProducer/candidateCreator3Prong.cxx
+++ b/PWGHF/TableProducer/candidateCreator3Prong.cxx
@@ -529,7 +529,7 @@ struct HfCandidateCreator3ProngExpressions {
     std::array<int, 2> arrPDGDaugh;
     std::array<int, 2> arrPDGResonant1 = {kProton, 313};      // Λc± → p± K*
     std::array<int, 2> arrPDGResonant2 = {2224, kKPlus};      // Λc± → Δ(1232)±± K∓
-    std::array<int, 2> arrPDGResonant3 = {3124, kPiPlus};     // Λc± → Λ(1520) π±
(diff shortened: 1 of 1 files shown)


fgrosa (approved):

vkucera (commented):
@mfaggin , I think this requires a thorough cross check between the MC numbering scheme and the ROOT numbering scheme, because these two should be identical and if they are not, it can be a big problem.

The code 3124 corresponds to the PDG numbering scheme used in the particle table of TDatabasePDG in ROOT, from where we get the particle properties.

See https://root.cern/doc/master/TPDGCode_8h.html
and `grep 1520 $ROOTSYS/etc/pdg_table.txt`:

```
  522 Lambda(1520)0      3124  1 100 Baryon      0 1.519000e+00 1.600000e-02 -100 -1 -100 -1   -1  12
  523 Lambda(1520)_bar  -3124     522     0
```
If `particle.pdgCode()` returns the MC code instead of the ROOT one, we have a problem.

fgrosa:
Hi @vkucera indeed the pdg code for the L(1520) is an issue if one uses `TDatabasePDG`, because it was modified in pythia but not in ROOT. I think it would be worth to open an issue / PR in ROOT, what do you think?

vkucera:
> Hi @vkucera indeed the pdg code for the L(1520) is an issue if one uses `TDatabasePDG`, because it was modified in pythia but not in ROOT. I think it would be worth to open an issue / PR in ROOT, what do you think?

I agree. Can you link the change in PYTHIA here?

PR #7901, opened 2024-10-06
Fix boolian function and track filter

ilikmeta:


victor-gonzalez (approved):
Thanks for considering my comments

PR #7912, opened 2024-10-07
added a configurable

prottayCMT:


BongHwi (approved):

vkucera:
@prottayCMT @BongHwi The commit title is too vague. It can be referring to any file in O2Physics. Please add information about the location of the changes.

BongHwi:
@vkucera Agree. 
Could you update the PR? and maybe we could check the old PRs as well. @prottayCMT

vkucera:
Thanks for your reply @BongHwi .
Unfortunately, renaming the merged PRs now does not help. What matters are the titles of commits in the upstream master branch made when the PR is merged. Once the PR is merged, the master history cannot be modified.
For future, please at least enforce to have the "PWGLF" prefix in the titles and, ideally, some information that indicates which analysis or file is concerned.
And be aware that when a PR has only one commit, it is the title of the commit, not the title of the PR, that becomes the title of the new commit in the upstream master branch.

PR #7916, opened 2024-10-08
replaced abs->TMath::Abs + configurable DCA cuts

alcaliva:


maciacco (approved):

vkucera:
@alcaliva , there is no benefit in replacing `std::abs` with `TMath::Abs`. You just increase dependency of your code on ROOT for no reason.

vkucera:
@maciacco @alcaliva The commit title is too vague. It can be referring to any file in O2Physics. Please add information about the location of the changes in the title of the commit. Please note that changing the title of the PR is not enough.

alcaliva:
@vkucera I don't understand your comment. Should I specify the name of the task in the comment field to avoid a reference to "any file in O2Physics"? How can I specify the location of the changes? I changed several lines ...

vkucera:
First of all, please add a "PWGLF" prefix in your commit title. Second, adding e.g. "Nuclei in jets" in the commit title is already specific enough to give clear information about the concerned analysis or file.

PR #7933, opened 2024-10-09
PWGHF: replace LOG(info) with LOG(debug) to avoid large outputs

fgrosa:


alibuild (approved):

PR #7934, opened 2024-10-09
Update Track Filter

ilikmeta:


victor-gonzalez (changes requested):
Could you please split this commit according to the functionality incorporated
- update track filter
- new task

Additionally
- put `PWGCF: ` as a prefix of the commit headers
- the mention to the new task say new task for global tracks. Probably that is not the intention because using a different kind of tracks can be done by mean of configuration and some additional code and not with a complete new task. Please be more explicit in the naming of the commit

ilikmeta:
Hello @victor-gonzalez 
My intention is to have two different Track Filter, One for Global tracks and one for Global+ITSonly. However, I could not do this in the same task, therefore I made two. I tried to do it using an if (Global==true) do global tracks and if (Global_ITSonly==true) do Global+ITSonly tracks (with Global and Global_ITSonly to be configurable). But, I saw that I cannot have a Filter inside an if statement (compiler was complaining). Ofcourse, I would prefere to have both track filters in the same task. 

Do you have a suggestion on how to do that? Thanks

victor-gonzalez:
Filters don't necessarily need to incorporate all selection criteria
Additional criteria might always be applied over the filtered objects
In case you want inspiration on flexible filters have a look at  https://github.com/AliceO2Group/O2Physics/blob/783c9012a7ec7f8a796dee64ab96d9e2c3b0d52f/PWGMM/Mult/Tasks/dndeta.cxx#L430 and followings

ilikmeta:
Please, allow me to ask a question: In lines 430-439 you have 5 track Filters. Then you join <aod::Tracks, aod::TracksExtra, aod::TrackSelection, aod::TracksDCA tables and then Filtered them on FiTracks. So, the Filtered tracks will contain tracks that was filtered with all 5 expressions::Filter? 

In my case I have the ITS+TPC and ITSonly in the same expressions::Filter `Filter trackFilter` . Then, with the new task I wanted to have `requireGlobalTrackInFilter() || isGlobalTrackSDD` track Filter. I do not think these two kind of track filters can exist on the same task as they produce different track samples. I was trying to have the Filter to be configurable, so I could enable it on the fly, but this did not work.

victor-gonzalez:
OK. I was talking about inspiration and not exact mapping
The idea is the flexibility that the `ifnode` construction provides when checking configuration inputs for altering the filters
Was that what you were looking for, right?
And there are a few examples there

ilikmeta:
@victor-gonzalez Can we close this PL without merging as I will modify the current analysis task to inlcude new track selection?

victor-gonzalez:
Sure! Just close it

PR #7946, opened 2024-10-10
[PWGLF] Update nuclei-proton correlation task: add purity and DCA pT dependent cut

ercolessi:


alibuild (approved):

PR #7913, opened 2024-10-07
PWGHF: add possibility to swap PID hypo for prongs 0 and 2 in ML application for Ds, Lc->pKpi, and Xic->pKpi 

fgrosa:

PR #7963, opened 2024-10-11
PWGHF: Add information on swapped candidates and considered mass hypotesis in Ds tree creator

fchinu:


PWGHF/TableProducer/treeCreatorDsToKKPi.cxx
```diff
@@ -294,13 +296,15 @@ struct HfTreeCreatorDsToKKPi {
     int8_t flagMc = 0;
     int8_t originMc = 0;
     int8_t channelMc = 0;
+    int8_t isSwapped = 1 - massHypo; // 0 if PiKK, 1 if KKPi
```
fcatalan92, line 299: Why you initialise `isSwapped` in this way?
BTW, it seems to me the convention is the opposite: `1 if PiKK, 0 if KKPi`
https://github.com/AliceO2Group/O2Physics/blob/master/PWGHF/TableProducer/candidateCreator3Prong.cxx#L586

fchinu: Hi @fcatalan92. Sorry I put the wrong numbers in the comments. 
Regarding the reason for initialising `isSwapped` like this is that, we would like to train and apply ML models using only  the KKPi mass hypotesis for the PID variables, i.e., when we have a PiKK candidate we swap the prongs 0 and 2 (see also #7913). This initialisation allows for an easier implementation of that: while for MC we swap based on the generated-level information of the prongs, for the data we swap based on the mass hypothesis that passed the selections.

fcatalan92: Thanks for the explanation @fchinu. After your last iteration, all looks good to me

fcatalan92 (approved):

PR #7971, opened 2024-10-12
PWGCF: flow-task: fix mistakes and add option to consume less memory

Luzhiyongg:


PWGCF/Flow/Tasks/FlowTask.cxx
```diff
@@ -559,18 +565,22 @@ struct FlowTask {
...
+      case 2:
+        return (track.isGlobalTrackWoTPCCluster() && track.tpcNClsFound() >= cfgCutTPCclu);
+      case 3:
+        return (track.isGlobalTrackWoPtEta() && (abs(track.eta()) < cfgCutEta) && (track.pt() > cfgCutPtMin) && (track.pt() < cfgCutPtMax));
+      case 4:
+        return (track.isGlobalTrackWoDCA() && abs(track.dcaZ()) <= cfgCutDCAz && abs(track.dcaXY()) <= cfgCutDCAxy * pow(track.pt(), -1.1));
+      case 5:
+        return (track.isGlobalTrackWoDCATPCCluster() && abs(track.dcaZ()) <= cfgCutDCAz && abs(track.dcaXY()) <= cfgCutDCAxy * pow(track.pt(), -1.1) && track.tpcNClsFound() >= cfgCutTPCclu);
+      default:
+        return false;
```
victor-gonzalez, line 582: Please, don't use `abs`. It turned out that now it is behaving as it should have, i.e. return an `int`
Use instead `std::fabs`

victor-gonzalez (changes requested):
Please, use `std::fabs` instead of `abs`. The last one will not work as you would expect

Luzhiyongg:
Hi @victor-gonzalez , thank you for your remind for the abs! I have fixed this.
Besides those using abs(), In Filter
```
Filter collisionFilter = nabs(aod::collision::posZ) < cfgCutVertex;
Filter trackFilter = (nabs(aod::track::eta) < cfgCutEta) && (aod::track::pt > cfgCutPtMin) && (aod::track::pt < cfgCutPtMax) && (aod::track::tpcChi2NCl < cfgCutChi2prTPCcls) && (aod::track::itsChi2NCl < cfgCutChi2prITScls);
```
I am using o2 inline node function nabs(). Is it corrected?

victor-gonzalez (approved):
The use of `nabs` in filters is correct

PR #7973, opened 2024-10-12
PWGEM/Dilepton: add a cut to ITS cluster size for hadron rejection

dsekihat:
PWGEM/Dilepton: add a cut to ITS cluster size for hadron rejection


alibuild (approved):

PR #7985, opened 2024-10-14
PWGCF: FemtoUniverse -- Correct selection functions in MultKt

a-m-andrushko:
Previously introduced simplifications resulted in selection functions cutting too many particles. This modification aims to resolve this issue.

### DIFF
--- a/PWGCF/FemtoUniverse/Tasks/femtoUniversePairTaskTrackTrackMultKtExtended.cxx
+++ b/PWGCF/FemtoUniverse/Tasks/femtoUniversePairTaskTrackTrackMultKtExtended.cxx
@@ -119,15 +119,15 @@ struct femtoUniversePairTaskTrackTrackMultKtExtended {
     Configurable<int> ConfPDGCodePartOne{"ConfPDGCodePartOne", 211, "Particle 1 -- PDG code"};
     // Configurable<uint32_t> ConfCutPartOne{"ConfCutPartOne", 5542474, "Particle 1 -- Selection bit from cutCulator"};
     Configurable<int> ConfPIDPartOne{"ConfPIDPartOne", 2, "Particle 1 -- Read from cutCulator"}; // we also need the possibility to specify whether the bit is true/false ->std>>vector<std::pair<int, int>>int>>
-    Configurable<float> ConfpLowPart1{"ConfpLowPart1", 0.14, "Lower limit for Pt for the first particle"};
+    Configurable<float> ConfPtLowPart1{"ConfPtLowPart1", 0.14, "Lower limit for Pt for the first particle"};
     Configurable<float> ConfPtHighPart1{"ConfPtHighPart1", 1.5, "Higher limit for Pt for the first particle"};
     Configurable<int> ConfChargePart1{"ConfChargePart1", 1, "Particle 1 sign"};
   } trackonefilter;
 
   /// Partition for particle 1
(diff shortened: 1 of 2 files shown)


PWGCF/FemtoUniverse/Tasks/femtoUniversePairTaskTrackTrackMultKtExtended.cxx
```diff
@@ -258,44 +258,78 @@ struct femtoUniversePairTaskTrackTrackMultKtExtended {
     // ConfNsigmaCombined -> TPC and TOF Sigma (combined) for momentum > ConfTOFpMin
 
     if (mom < twotracksconfigs.ConfTOFpMin) {
-      return TMath::Abs(nsigmaTPC) < twotracksconfigs.ConfNsigmaTPC;
+      if (TMath::Abs(nsigmaTPC) < twotracksconfigs.ConfNsigmaTPC) {
```
saganatt, line 261: lines 261-265 are equivalent to the old line 261. I don't see how that changes anything.
The only difference could be made by introducing line 274.

PWGCF/FemtoUniverse/Tasks/femtoUniversePairTaskTrackTrackMultKtExtended.cxx
```diff
@@ -258,44 +258,78 @@ struct femtoUniversePairTaskTrackTrackMultKtExtended {
...
+    return false;
   }
 
   /// TPC Kaon Sigma selection (stricter cuts for K+ and K-) -- based on Run2 results
   bool IsKaonNSigma(float mom, float nsigmaTPCK, float nsigmaTOFK)
   {
     if (twotracksconfigs.IsKaonNsigma == true) {
       if (mom < 0.4) {
-        return TMath::Abs(nsigmaTPCK) < 2;
+        if (TMath::Abs(nsigmaTPCK) < 2) {
```
saganatt, line 282: same as above -- I don't see any difference except line 312. Could you keep the old code except line 312?

PWGCF/FemtoUniverse/Tasks/femtoUniversePairTaskTrackTrackMultKtExtended.cxx
```diff
@@ -258,44 +258,78 @@ struct femtoUniversePairTaskTrackTrackMultKtExtended {
...
+    return false;
   }
 
   /// TPC Deuteron Sigma selection
   bool IsDeuteronNSigma(float mom, float nsigmaTPCDe, float nsigmaTOFDe)
   {
-    if (mom > deuteronconfigs.ConfPLowDe && mom < deuteronconfigs.ConfPHighDe) {
-      if (mom < deuteronconfigs.ConfTOFpMinDe) {
-        return TMath::Abs(nsigmaTPCDe) < deuteronconfigs.ConfNsigmaTPCDe;
+    if (mom < deuteronconfigs.ConfTOFpMinDe) {
```
saganatt, line 318: OK. Here you indeed change the logics a bit. You no longer check for pLow and pHigh before checking TOFpMin and nsigmas. Probably this is the one true fix that enables you to get more particles.

Could you also here use the convention `return TMath::Abs(nsigmaTPCDe) < deuteronconfigs.ConfNsigmaTPCDe` instead of repeating if...else ? Same for line 325. Trust me, it is exactly the same logics and it is not the cause of your problems :)

PWGCF/FemtoUniverse/Tasks/femtoUniversePairTaskTrackTrackMultKtExtended.cxx
saganatt: Could you please keep the old code here except line 274?

PR #7991, opened 2024-10-14
PWG LF: add boolean to store only true cascades for efficiency calculation in cascadeflow task

ChiaraDeMartin95:


alibuild (approved):

ercolessi (approved):

PR #7996, opened 2024-10-14
[PWGJE] Task for hadron-jet correlations in O-O collisions

KotliarovAr:
A new task designed to study hadron-jet correlations in the upcoming OO collisions.
It is the first draft of the code; new functionality and fixes will be added

### DIFF
--- a/PWGJE/Tasks/jetHadronRecoil_OO.cxx
+++ b/PWGJE/Tasks/jetHadronRecoil_OO.cxx
@@ -1,0 +1,524 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+// h+jet analysis task for OO collisions
+/// \author: Kotliarov Artem <artem.kotliarov@cern.ch>
+
+#include <vector>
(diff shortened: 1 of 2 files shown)


PWGJE/Tasks/jetHadronRecoil_OO.cxx
```diff
@@ -0,0 +1,524 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
```
KotliarovAr, line 10: Check

KotliarovAr:
@fkrizek It is the skeleton of the code; it is much easier to follow the code update and also you can provide your comments line-by-line, which is helpful. Currently, there is no possibility of estimating rho for the MC particle level, so this part of the code has been commented out. However, I suggest to start discussing the histograms and observables, which we will study.

KotliarovAr:
Dear @nzardosh, 
Could you review the code? Also, I want to ask 2 questions:

1) we discussed some time ago that the JE framework does not contain functionality to estimate rho background for particle level MC. Do we plan to add such functionality?

2) when we want to get a weight for MC detector level simulations, we use ``auto weight = collision.mcCollision().weight()``. There is a function that checks whether a collision has an associated MC collision: ``collision.has_mcCollision()``. So, not all detector-level collisions have associated particle-level collisions? Are we protected from the output not being garbage due to the absence of the associated MC collision when calling ``collision.mcCollision().weight()``? Should we also introduce a check (``collision.has_mcCollision()``) that discards the MC detector level collisions, which do not have associated MC particle level collisions?

PR #8007, opened 2024-10-15
PWGLF: added v1 calculations for ZDC Callibration

prottayCMT:
Added the scalar product v1calculations to compare with RUN1 results as suggested by FLOW PAG.


BongHwi (approved):

BongHwi:
Hello @prottayCMT , it will be even better if you add some descriptions.

PR #8015, opened 2024-10-15
PWGHF: add columns of candidate type and origin for ML training in D0 derived table creator

Mingyu3360715:


Mingyu3360715:
Hi @vkucera ,  I'm using this new task to produce derived table for ML training. 
At the moment, the candidateSelFlag, flagMcMatchRec and originMcRec are stored in separate tables. In order to build training samples for different candidate types, I put them in the same output table HfD0Pars. If it is not necessarily needed or there is another way to do it, please leave a comment on it. 
Thanks a lot!

vkucera:
Hi @Mingyu3360715 , the separation of MC properties is intentional to allow for modularity in running on MC and real data. Adding redundant MC columns in the table dedicated to candidate parameters goes against this design. Please join the two tables in your processing code.

Mingyu3360715:
> Hi @Mingyu3360715 , the separation of MC properties is intentional to allow for modularity in running on MC and real data. Adding redundant MC columns in the table dedicated to candidate parameters goes against this design. Please join the two tables in your processing code.

I get it. Thank you very much.

PR #8017, opened 2024-10-15
Common: move mult/cent table producers to getForRun CCDB queries

ddobrigk:
* changes CCDB queries in mult and cent tasks to `getForRun` (without metadata) and `getSpecificForRun` (with metadata information) 
* Adds a healthy printout explaining which object is being loaded in case of metadata use with reconstruction pass
* Rearranges the `mRunNumber = bc.runNumber();` call such that operation without crash over uncalibrated data doesn't result in unnecessary CCDB queries: attempt CCDB load only once when run number changes and switch run number regardless of outcome.


ddobrigk:
Now tested successfully locally. Further development will make use of the reco-pass-aware load (for Pb-Pb pass4), coming soon.

alibuild (approved):

PR #8020, opened 2024-10-16
GH Actions: Bypass PEP 668 in runners

singiamtel:
This is needed in order to install pip packages globally in the new Ubuntu images, the other choice would be to create a venv and install the deps there instead

https://peps.python.org/pep-0668/

https://mattermost.web.cern.ch/alice/pl/f75f33zosf8rbd9dnagywfinac

CC: @mpuccio @vkucera @ktf


vkucera:
Unfortunately this throws `no such option: --break-system-packages` on Ubuntu 22.04.5 LTS workers.

PR #8025, opened 2024-10-16
[PWGEM] Single track QC: Increase range of DCA axis

feisenhu:
Increase the range of the DCA axis from 10 to 20 sigma


dsekihat:
can you make this configurable? default should be up to 10.

dsekihat (approved):

PR #8028, opened 2024-10-16
PWGHF: Add the QA plots for the mixed-event and Change the strategy of mixed event

zhangbiao-phy:
Hi @1994ra and @fgrosa, The commit I implemented last week is still not satisfactory, since the number of events that contain lc (with selection) in one DF are few. The mixed event flow can only process in one DF for the current O2. so I made further modifications. Now, the producer no longer stores only the Lc-proton pair for collisions. Instead, I’ve set a mask to tag the collisions that contain at least one Lc or at least one selected track. Let me know if you have any questions or comments. Thanks!

### DIFF
--- a/PWGHF/HFC/TableProducer/femtoDreamProducer.cxx
+++ b/PWGHF/HFC/TableProducer/femtoDreamProducer.cxx
@@ -434,7 +434,7 @@ struct HfFemtoDreamProducer {
       bitCand |= 1 << 0;
     }
 
-    Masks(static_cast<aod::femtodreamcollision::BitMaskType>(bitTrack),
+    rowMasks(static_cast<aod::femtodreamcollision::BitMaskType>(bitTrack),
           static_cast<aod::femtodreamcollision::BitMaskType>(bitCand),
           0);
   }
--- a/PWGHF/HFC/TableProducer/femtoDreamProducer.cxx
+++ b/PWGHF/HFC/TableProducer/femtoDreamProducer.cxx
@@ -412,8 +412,7 @@ struct HfFemtoDreamProducer {
                 if constexpr (isMc) {
                   rowCandMcCharmHad(
                     candidate.flagMcMatchRec(),
-                    candidate.originMcRec());
-                }           
+                    candidate.originMcRec());}
       } };
 
       fillTable(0, candidate.isSelLcToPKPi(), outputMlPKPi.at(0), outputMlPKPi.at(1), outputMlPKPi.at(2));
--- a/PWGHF/HFC/TableProducer/femtoDreamProducer.cxx
+++ b/PWGHF/HFC/TableProducer/femtoDreamProducer.cxx
@@ -46,7 +46,7 @@ using namespace o2::analysis::femtoDream;
 struct HfFemtoDreamProducer {
 
   Produces<aod::FDCollisions> outputCollision;
-  Produces<aod::FDColMasks> Masks;
+  Produces<aod::FDColMasks> rowMasks;
   Produces<aod::FDHfCand> rowCandCharmHad;
   Produces<aod::FDHfCandMC> rowCandMcCharmHad;
   Produces<aod::FDHfCandMCGen> rowCandCharmHadGen;
--- a/PWGHF/HFC/TableProducer/femtoDreamProducer.cxx
+++ b/PWGHF/HFC/TableProducer/femtoDreamProducer.cxx
@@ -388,7 +388,6 @@ struct HfFemtoDreamProducer {
                            float BDTScorePrompt,
                            float BDTScoreFD) {
         if (FunctionSelection >= 1){
-    
                 rowCandCharmHad(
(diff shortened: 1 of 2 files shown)


PWGHF/HFC/Tasks/taskCharmHadronsFemtoDream.cxx
```diff
@@ -12,6 +12,7 @@
 /// \file taskCharmHadronsFemtoDream.cxx.cxx
 /// \brief Tasks that reads the track tables used for the pairing and builds pairs of two tracks
 /// \author Ravindra SIngh, GSI, ravindra.singh@cern.ch
+/// \author Biao Zhang, Heidelburg University, biao.zhang@cern.ch
```
vkucera, line 15:
```suggestion
/// \author Biao Zhang, Heidelberg University, biao.zhang@cern.ch
```

PWGHF/HFC/TableProducer/femtoDreamProducer.cxx
```diff
@@ -46,6 +46,7 @@ using namespace o2::analysis::femtoDream;
 struct HfFemtoDreamProducer {
 
   Produces<aod::FDCollisions> outputCollision;
+  Produces<aod::FDColMasks> Masks;
```
vkucera, line 49:
```suggestion
  Produces<aod::FDColMasks> rowMasks;
```

PWGHF/HFC/Tasks/taskCharmHadronsFemtoDream.cxx
```diff
@@ -154,13 +155,15 @@ struct HfTaskCharmHadronsFemtoDream {
...
+  femtodreamcollision::BitMaskType BitMask = 1 << 0;
+
   /// Histogramming for particle 1
   FemtoDreamParticleHisto<aod::femtodreamparticle::ParticleType::kTrack, 1> trackHistoPartOne;
   /// Histogramming for Event
   FemtoDreamEventHisto eventHisto;
   /// Histogram output
   HistogramRegistry registry{"CorrelationsAndQA", {}, OutputObjHandlingPolicy::AnalysisObject};
-
+  HistogramRegistry MixQaRegistry{"MixQaRegistry", {}, OutputObjHandlingPolicy::AnalysisObject};
```
vkucera, line 166:
```suggestion
  HistogramRegistry registryMixQa{"registryMixQa"};
```

PWGHF/HFC/Tasks/taskCharmHadronsFemtoDream.cxx
```diff
@@ -203,17 +206,29 @@ struct HfTaskCharmHadronsFemtoDream {
...
+    MixQaRegistry.add("MixingQA/hMECollisionBins", ";bin;Entries", kTH1F, {{120, -0.5, 119.5}});
     pairCleaner.init(&registry);
     if (useCPR.value) {
       pairCloseRejectionSE.init(&registry, &registry, cprDeltaPhiMax.value, cprDeltaEtaMax.value, cprPlotPerRadii.value, 1);
       pairCloseRejectionME.init(&registry, &registry, cprDeltaPhiMax.value, cprDeltaEtaMax.value, cprPlotPerRadii.value, 2);
     }
   }
 
+  template <typename CollisionType>
+  void fillCollision(CollisionType col)
```
vkucera, line 220:
```suggestion
  void fillCollision(CollisionType const& col)
```

vkucera (changes requested):

fgrosa:
Hi @zhangbiao-phy, thanks a lot!
I remember that we chose the strategy to save only events that contain a p-Lc pair to reduce the disk space. In case derived data are produced, DF are merged and therefore we should have enough protons and Lcs to pair.
Is this not the case because you are doing all in one go, without producing derived data?

zhangbiao-phy:
> Hi @zhangbiao-phy, thanks a lot! I remember that we chose the strategy to save only events that contain a p-Lc pair to reduce the disk space. In case derived data are produced, DF are merged and therefore we should have enough protons and Lcs to pair. Is this not the case because you are doing all in one go, without producing derived data?

Hi @fgrosa, yes, this is the previous strategy in the producer, but we don't save the derived table at this stage. So, from what I tested locally, I did not find it would increase the disk space a lot.  Then I would say It's not merged in the task process, then the event pool is arround ~10-25, so it's hard to mix 10 events in that pool. 

another method is to save the collision index and vz and mul in the final derived table, then mix it locally. These are the two methods also used in HFC currently.

zhangbiao-phy:
> > Is this not the case because you are doing all in one go, without producing derived data?

It's not the case, the point is still the number of collisions that include Lc-p pair (satisfy the Lc selection) are very few.

vkucera (approved):
Thanks @zhangbiao-phy for implementing my suggestions.
I let @deepathoms @apalasciano have the final look and merge.

alibuild (approved):

PR #8035, opened 2024-10-17
PWGHF: Unify different partitions in Ds task

fchinu:
This PR fixes the filling of the number of contributors in the Ds task output. 
After an interesting chat with @aalkin, I was suggested to drop the usage of the `std::map` to indicate the partition to be used, as it increases both compilation and run time, and remove redundant multiple definitions of the partitions, as it is not needed to have different specialisations for the different `Join`s.

### DIFF
--- a/PWGHF/D2H/DataModel/ReducedDataModel.h
+++ b/PWGHF/D2H/DataModel/ReducedDataModel.h
@@ -221,14 +221,21 @@ DECLARE_SOA_TABLE(HfRed3ProngsCov, "AOD", "HFRED3PRONGSCOV", //! Table with 3pro
                   HFTRACKPARCOV_COLUMNS,
                   o2::soa::Marker<2>);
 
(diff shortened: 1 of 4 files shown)


fgrosa (dismissed):
Nice solution, thanks!

aalkin:
@fchinu I took a closer look at the code, and what you are doing with selecting preslice is in fact redundant. Since you are always slicing on the MC collision index, there is no reason to have several instances of preslice, it is actually enough to have `PresliceUnsorted<CollisionsMc> colPerMcCollision = aod::mccollisionlabel::mcCollisionId;` or even `PresliceUnsorted<aod::McCollisionLabels> colPerMcCollision = aod::mccollisionlabel::mcCollisionId;` and it is usable for any join that contains it.

fchinu:
Thanks a lot @aalkin for your time and suggestions! I have changed the code following your advice, it is now much cleaner!

mfaggin (approved):
all good for me, I approve it. @aalkin if you can confirm that it's ok for you as well I merge it. Thanks

aalkin:
The code looks good, please go ahead and merge it.

PR #8051, opened 2024-10-18
PWGHF: Add centrality (FT0C as default) in Lc task

zhangbiao-phy:
To make the centrality differential analysis more convenient for offline process

### DIFF
--- a/PWGHF/D2H/Tasks/taskLc.cxx
+++ b/PWGHF/D2H/Tasks/taskLc.cxx
@@ -43,7 +43,7 @@ struct HfTaskLc {
   ConfigurableAxis thnConfigAxisPt{"thnConfigAxisPt", {72, 0, 36}, ""};
   ConfigurableAxis thnConfigAxisMass{"thnConfigAxisMass", {300, 1.98, 2.58}, ""};
   ConfigurableAxis thnConfigAxisPtProng{"thnConfigAxisPtProng", {100, 0, 20}, ""};
-  ConfigurableAxis thnConfigAxisMultiplicity{"thnConfigAxisMultiplicity", {100, 0, 1000}, ""};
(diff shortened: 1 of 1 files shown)


zhangbiao-phy:
Hi @vkucera, I would like to add the centrality information in the ThnSparse of the Lc task. Any comments and suggestions are welcome! Thanks for your review!

fgrosa (dismissed):
@zhangbiao-phy looks good to me! Thanks for addressing our comment, it will improve the usage of resources on hyperloop a lot. Could you please just add a process function that does not consume any centrality table, so that we do not force to have it in the dependencies in case of MB analyses? Thanks!

vkucera (changes requested):
Hi @zhangbiao-phy , looks good to me provided you implement the request by @fgrosa 
If `float cent = collision.centFT0C();` is the only line requiring the centrality table, you can also take it out of the function and provide it as an argument.

fgrosa:
> Hi @zhangbiao-phy , looks good to me provided that you implement the request by @fgrosa If `float cent = collision.centFT0C();` is the only line requiring the centrality table, you can also take it out of the function and provide it as an argument.

Actually I think it would be even better to implement the method that uses the `concepts` defined by @fchinu (see Ds task https://github.com/AliceO2Group/O2Physics/blob/master/PWGHF/D2H/Tasks/taskDs.cxx#L252-L259)

fgrosa:
@zhangbiao-phy thanks for the fast implementation! Let's see if @vkucera has other comments, otherwise it can be merged in my opinion.

vkucera (dismissed):

zhangbiao-phy:
Hi @vkucera, there are typos in the name of the process function. I fixed it now. Could you approve the PR again? Thanks a lot!

vkucera (approved):

PR #8055, opened 2024-10-18
[PWGLF] Add histograms to check inv. mass Vs Phi Vs Eta

romainschotter:
- Copy UPC info in the aod::StraEvSels_002 converter task (before they were set to dummy values by mistake)
- Correct typo in the histogram axis title
- Add a histogram to monitor the number of reconstructed K0s, Lambda and AntiLambda per collision
- Add a QC histogram to check the invariant mass as a function of phi and eta to check the overall detector alignment


njacazio (approved):
Thanks @romainschotter

PR #8073, opened 2024-10-21
Update JE codeowners

ddobrigk:
Actually in this case I do not know what label I should put (suggestions appreciated, of course :-) )


alibuild (approved):

PR #8082, opened 2024-10-21
PWGHF: taskFlow.cxx: Solved problems with TPC-MFT HF-h mixed event

AlexianL:
Solved problems with TPC-MFT HF-h mixed event, some plots re-organization, solves the problem concerning the hEventSelection plot being filled twice when doing reference flow


AlexianL:
Hello @apalasciano @deepathoms , could any of you please review this small PR ? I would need to run TPC-MFT HF-h case on hyperloop soon so I am able to discuss results with my supervisors before presenting at PAG meeting. Thank you !

apalasciano (approved):

PR #8095, opened 2024-10-22
[PWGDQ]: update Zorro

zjxiongOvO:


EventFiltering/Zorro.h
```diff
@@ -44,6 +45,11 @@ class Zorro
   bool isNotSelectedByAny(uint64_t bcGlobalId, uint64_t tolerance = 100);
 
   void populateHistRegistry(o2::framework::HistogramRegistry& histRegistry, int runNumber, std::string folderName = "Zorro");
+  TH2* mZorroHisto = nullptr;
+  TH2* mToiHisto = nullptr;
+  void populateHist(int runNumber);
```
mpuccio, line 50: I would change this function to be called `populateExternalHists` and pass to the function the pointers to the histograms to be populated

EventFiltering/Zorro.cxx
```diff
@@ -191,6 +253,9 @@ bool Zorro::isSelected(uint64_t bcGlobalId, uint64_t tolerance)
         mAnalysedTriggersOfInterest->Fill(i);
         mZorroSummary.increaseTOIcounter(mRunNumber, i);
       }
+      if (mToiHisto && lastSelectedIdx != mLastSelectedIdx) {
```
mpuccio, line 256: Instead of having `mToiHisto` as a datamember, you can add an optional histogram pointer to `isSelected` and fill it only if it is provided

mpuccio (changes requested):
Dear @zjxiongOvO, thanks a lot for your PR! Please have a look at my comments and tell me if you agree with those.

EventFiltering/Zorro.cxx
zjxiongOvO: Dear @mpuccio , Thanks for your comments. I agree with them and have submitted a PR to make the modifications.

mpuccio (approved):

lucamicheletti93 (approved):

PR #8102, opened 2024-10-22
Update the default base CCDB path for Zorro

mpuccio:
Changes the base CCDB path for the Zorro class from "Users/m/mpuccio/EventFiltering/OTS/" to
"Users/m/mpuccio/EventFiltering/OTS/Chunked/". This is necessary to point to the correct location of the CCDB files for the chunked data.


alibuild (approved):

PR #8103, opened 2024-10-22
PWGHF: Add histograms directories to Xic task

phymanshu:
- Add prompt/non-prompt directory structure in MC reco and gen histogram directories


mfaggin (changes requested):
change all the `TH1F` to `TH1D` to avoid issues related to the floating-point precision when the number of bins in a histogram is low

mfaggin (approved):

PR #8109, opened 2024-10-23
pull request for new workflow femtoDreampHe3ProducerTask.cxx

blacwovie:
Tables pidTPCFullHe3 and pidTOFFullHe3 are added in producer task and saved as a new workflow femtoDreampHe3ProducerTask.cxx to producer femto derived data with He3 particle.

### DIFF
--- a/PWGCF/DataModel/FemtoDerived.h
+++ b/PWGCF/DataModel/FemtoDerived.h
@@ -37,7 +37,7 @@ enum CollisionBinning {
   kMultMultPercentile, //! Bin collision in number of charged tracks and multiplicity percentile for mixing
   kNCollisionBinning
 };
-
+//DECLARE_SOA_COLUMN 是一个宏定义，通常用于声明 结构化数组
 DECLARE_SOA_COLUMN(MultV0M, multV0M, float);       //! V0M multiplicity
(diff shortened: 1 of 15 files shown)


victor-gonzalez:
Probably this is going to affect the `FemtoDream` datamodel
@blacwovie Please, set it as draft while you agree on it with the `FemtoDream` people @ariedel-cern @lauraser @sartozza

ariedel-cern:
Hi @blacwovie,
copying and creating a new producer task is not the way to go to add He3 PID.
I went ahead and created another draft to extend the data model to include the PID for He3 and triton. The PR is not quite finished yet because to see the PID, I also need to add a histogram to the debug task.

@victor-gonzalez : He3 and trition have not been in the data model so far because there was no analysis yet and the PID task did not have the autoconfigre feature back then. The changes are rather straight forward, but if you could also have a short look at the draft, that would be nice.

Draft https://github.com/AliceO2Group/O2Physics/pull/8123

blacwovie:
Hi,@ariedel-cern So sorry for causing troubles,I really didn't think this through. May I ask when will the new draft with He3 complete? And is it possible to add histogram of particles dE/dx in the task?

ariedel-cern:
Hi, I should be done with the draft today or tomorrow. And we should be able to get it merged soon after. A histogram for the energy loss is already implemented. Only a histogram with the nsigma distribution for the triton/He3 PID was missing from the debug task

PR #8119, opened 2024-10-23
[PWGDQ] Removing useless part to the PID task

lucamicheletti93:
Removing useless part to the PID task


alibuild (approved):

PR #8123, opened 2024-10-23
PWGCF: add triton and He3 PID to femtodream data model

ariedel-cern:
Add triton and He3 PID to femtodream data model.

### DIFF
--- a/PWGCF/FemtoDream/Core/femtoDreamParticleHisto.h
+++ b/PWGCF/FemtoDream/Core/femtoDreamParticleHisto.h
@@ -100,16 +100,22 @@ class FemtoDreamParticleHisto
       mHistogramRegistry->add((folderName + folderSuffix + "/nSigmaTPC_K").c_str(), "n#sigma_{TPC}^{K}", kTH2F, {pTAxis, NsigmaTPCAxis});
       mHistogramRegistry->add((folderName + folderSuffix + "/nSigmaTPC_p").c_str(), "n#sigma_{TPC}^{p}", kTH2F, {pTAxis, NsigmaTPCAxis});
       mHistogramRegistry->add((folderName + folderSuffix + "/nSigmaTPC_d").c_str(), "n#sigma_{TPC}^{d}", kTH2F, {pTAxis, NsigmaTPCAxis});
+      mHistogramRegistry->add((folderName + folderSuffix + "/nSigmaTPC_tr").c_str(), "n#sigma_{TPC}^{tr}", kTH2F, {pTAxis, NsigmaTPCAxis});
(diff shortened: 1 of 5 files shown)


PWGCF/FemtoDream/Core/femtoDreamParticleHisto.h
```diff
@@ -360,10 +372,18 @@ class FemtoDreamParticleHisto
...
-          case 1000010020:
+          case 1000010020: // Deuteron
             pidTPC = part.tpcNSigmaDe();
             pidTOF = part.tofNSigmaDe();
             break;
+          case 1000010030: // Triton
+            pidTPC = part.tpcNSigmaTr();
+            pidTOF = part.tofNSigmaTr();
+            break;
+          case 1000020030: // Helium3
```
victor-gonzalez, line 383: Could we move to use named constants as per
https://github.com/AliceO2Group/AliceO2/blob/dev/Common/Constants/include/CommonConstants/PhysicsConstants.h
it will be less error prone and no comments are needed

As it is commented there, the previous file expands the definitions in `<TPDGCode_8.h>` as visible in https://root.cern/doc/master/TPDGCode_8h.html

victor-gonzalez (changes requested):
Anton, please, have a look at my only comment

Is that everything that is needed to incorporate He3 to the derived data?

PWGCF/FemtoDream/Core/femtoDreamParticleHisto.h
ariedel-cern: Hi Victor,
yes. I just need to add the PID to the extra tables and some additional histograms. The framework which calculates the bitmask did already support all possible track PIDs.
Thanks for the comment. I swich the PDG codes to the name constants.

ariedel-cern:
Hi @mfaggin @fgrosa @1994ra ,
I got the request to add triton and He3 PID to the femto dream extra table. Since the femto dream producer in the HFC framework is also subscribing to this table, I had to make a small adjustment (i.e. adding 4 more dummy values in your producer since you only use track PID for proton and lighter particles). As far as I can tell, these changes should not affect the other HF tasks. But it would be good if you could also take a look.

fgrosa (approved):
Hi @AntonRiedel I am approving it not to block your work, but I think it would be better to have joinable tables for the PID hypotheses, not to have many unused dummy columns when not needed

victor-gonzalez:
@ariedel-cern, @fgrosa made a good point!
Joinable tables will reduce the memory footprint for concrete analyses
I think we should consider it for the ongoing re-organization, if feasible

victor-gonzalez (approved):

PR #8126, opened 2024-10-23
[PWGCF] Add files via upload

rahulverma012:
h Phi Correlation Task by rahul.verma@cern.ch


victor-gonzalez:
Thanks for moving forward this functionality

It would be more appropriate to set the PR as draft because it will need a few iterations before it can be merged

The `MultiparticleCorrelations` directory is in principle intended for correlations among several particles, usually more than two. As the PR code is today, it is only addressing two-particle correlations although of different species. Would not be more appropriate to have this code under the `TwoParticleCorrelations` directory?

You made a huge effort coding your own mixed events functionality. Is there any reason that made you decide not to use the standard mixed events functionality available in O2?

For implementing the mixed events functionality, all the information stored in tables is replicated in internal structures. This goes against the way O2 was conceived: the information is stored, in a very efficient way, in tables that are resident in memory. Then from there the information is read, in a very efficient way, as many times as needed but, in principle it should not be replicated and much less in internal structures which are not supporting an efficient access

The code needs to be re-structured in a way that the information in the tables is not replicated
I would suggest to use the standard mixed events functionality which already provides the functionality adapted to the O2 framework

rahulverma012:
Dear Victor,

>I would like to commit my code to the TwoParticleCorrelations directory.

>While working on event mixing with the existing method, I encountered an issue with insufficient collisions in certain mixing bins across dataframes. I’m dividing my events into 40 bins (10 z-vertex positions and 4 centrality bins) and conducting three types of correlations: hadron-Phi, hadron-hadron, and Phi-Phi.

In the case of hadron-Phi, I get only 0, 1, or 2 collisions per dataframe. However, some of the 40 bins are very abundant (for example, bin 10, with vtxZ in [0,2] and centrality in [0,20], consistently has collisions after 2-3 dataframes), while others are quite rare (like bin 40, with vtxZ in [8,10] and centrality in [80,100], which has two collisions after analyzing a thousand dataframes).

If i try to derive the merged data for mixing, to achieve six collisions for each bin (mixing one collision with five others), the size of the merged dataframes becomes excessively large. Including collisions for hadron-hadron and Phi-Phi mixing would further increase the size.

To address this, I  come with idea of storing minimal collision and track information to minimize memory usage. This information will remain in memory only as long as needed, being read five times for mixing, with the sixth read updated with the next collision's data. This approach mirrors the strictly upper index policy used in standard functionality.

I have structured the stored data in a way that it behaves like tables and partitions. And it consumes less or equal memory compared to the memory consumed by tables when loaded into the memory.

victor-gonzalez:
> I would like to commit my code to the TwoParticleCorrelations directory

It sounds reasonable. Let's iterate on this layout and at the end do the relocation

> While working on event mixing with the existing method, I encountered an issue with insufficient collisions in certain mixing bins across dataframes. I’m dividing my events into 40 bins (10 z-vertex positions and 4 centrality bins) and conducting three types of correlations: hadron-Phi, hadron-hadron, and Phi-Phi

This is something other groups have already confronted, for example D-h correlations in HF, and they kept using the standard mixed events framework from O2

> In the case of hadron-Phi, I get only 0, 1, or 2 collisions per dataframe. However, some of the 40 bins are very abundant (for example, bin 10, with vtxZ in [0,2] and centrality in [0,20], consistently has collisions after 2-3 dataframes), while others are quite rare (like bin 40, with vtxZ in [8,10] and centrality in [80,100], which has two collisions after analyzing a thousand dataframes)

Most probably that is expected although going to 10 cm in zvtx and 100% centrality/multiplicity is a bit extreme

> If i try to derive the merged data for mixing, to achieve six collisions for each bin (mixing one collision with five others), the size of the merged dataframes becomes excessively large. Including collisions for hadron-hadron and Phi-Phi mixing would further increase the size

Derive data involving hadrons will never have the proper reduction factor so it should not be constructed 

> To address this, I come with idea of storing minimal collision and track information to minimize memory usage. This information will remain in memory only as long as needed, being read five times for mixing, with the sixth read updated with the next collision's data. This approach mirrors the strictly upper index policy used in standard functionality

With the proposed implementation, the information is replicated and stored across data frames which tends to be quite large
The amount of memory available in hyperloop is limited because the goal is to execute as much tasks in parallel as possible having all of them accessing to the same memory in a quite efficient way
In the standard functionality the information is not replicated

> I have structured the stored data in a way that it behaves like tables and partitions. And it consumes less or equal memory compared to the memory consumed by tables when loaded into the memory

Table and partitions use the SOA architecture that makes them extremely efficient for massive sequential access. That is the core of O2 architecture. Internal structures not use that architecture 

In my opinion as it is today it will not run on hyperloop due to memory constraints and, probably, to CPU usage.
I would suggest
- As mentioned above, other groups have confronted this situation keeping use of the O2 mixed event framework
- For h-h correlations there should not be any issue. A group within CF is already doing h-h correlations with the standard framework
- For Phi-Phi correlations derived data with a large data frame integration should work. Femto group has been doing this since a while
- For h-Phi correlations are in the same foot as D-h correlations. The DH group works with derived data, for the D, linked to original data for the h. The details of both were presented on the latest O2 analysis tutorial

rahulverma012:
Dear Victor,  I am working on the changes that you have suggested for event mixing. For now, I have only kept the same event correlation part in this draft and have removed the mixing part which was duplicating data. Can you please review the code for merging in main repository
Best Regards
Rahul

PR #8136, opened 2024-10-24
PWGCF - Calibration study of ZDC SP for charge dependent flow studies 

cnkoster:
For the analysis of v1 these calibrations are needed to get the correct spectator plane. 
The task creates derived data to enable offline analysis and creation of the calibration histograms.


victor-gonzalez:
The code as is now, produces a table, but so far there are not subscribers to that table
What is the aim, to use the produced table as derived data to do some post-processing? If that is the case, I would suggest to put the data model within the own task source file
If the aim is to produce a table whose data will be used by other task/s then I would suggest moving the task source file to the `TableProducer` directory

Regarding the table as such, it is really needed to have the run number for each collision?

cnkoster:
For now table is indeed needed for post-processing. At a later stage the output will be used by another task to calculate v1. So I could move inside the task for now and extend it to the table producer at a later stage, or would it be better to move it there directly?  For the post processing, al q-vectors need to be analysed on a run-by-run basis and it seemed convenient to add the run numbers to the table.

victor-gonzalez:
I see. Then I would suggest moving the table producer source code to the `TableProducer` directory renaming it appropriately and reserve the source code name for the v1 task that will be placed in the `Tasks` directory

victor-gonzalez (approved):

PR #8141, opened 2024-10-24
[PWGCF] singletrackselector: Add TPC nsigma electron and Zorro utils

ercolessi:
This PR is to include the TPC nsigma electron and the possibility to use a trigger mask on the skimmed datasets.

### DIFF
--- a/PWGCF/Femto3D/DataModel/singletrackselector.h
+++ b/PWGCF/Femto3D/DataModel/singletrackselector.h
@@ -483,7 +483,7 @@ DECLARE_SOA_TABLE(SingleTrkExtras, "AOD", "SINGLETRKEXTRA",
                   singletrackselector::TPCSignal,
                   singletrackselector::Beta);
 
-DECLARE_SOA_TABLE(SinglePIDExtras, "AOD", "SINGLEPIDEXTRA",
+DECLARE_SOA_TABLE(SinglePIDEls, "AOD", "SINGLEPIDEL",
                   singletrackselector::StoredTPCNSigmaEl,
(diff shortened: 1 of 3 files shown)


ercolessi:
Dear @victor-gonzalez,

this PR is ready for review.

Thank you in advance,
Francesca 

Cc: @glromane

PWGCF/Femto3D/DataModel/singletrackselector.h
```diff
@@ -479,6 +483,10 @@ DECLARE_SOA_TABLE(SingleTrkExtras, "AOD", "SINGLETRKEXTRA",
                   singletrackselector::TPCSignal,
                   singletrackselector::Beta);
 
+DECLARE_SOA_TABLE(SinglePIDEls, "AOD", "SINGLEPIDEL",
+                  singletrackselector::StoredTPCNSigmaEl,
+                  singletrackselector::TPCNSigmaEl<singletrackselector::StoredTPCNSigmaEl>);
+
```
victor-gonzalez, line 489: Why a separated table?
Derived data will be produced together with the other tables and as such has to be coordinated
Should not be then better to have it as a natural extension of the other table which stores the same information about Pi, Ka, ...?

victor-gonzalez (commented):

PWGCF/Femto3D/DataModel/singletrackselector.h
ercolessi: Hi @victor-gonzalez , this is the first step towards the splitting of PID tables for different types of particles in order to use them as joinable. Since we needed to urgently add this new column we decided to move already in that direction.

victor-gonzalez (approved):

PR #8144, opened 2024-10-25
PWGLF:: To add the namspace for abs

navneetkumar231295:
Dear Experts,

I am trying to run the o2-analysis-lf-k892pmanalysis task for testing. However, I have noticed that abs(v0.mLambda() - massLambda0) always returns a zero value. When I include the namespace, it works fine. Therefore, I kindly request that these changes be committed.

Thanks,
With Regards, 
Navneet Kumar

### DIFF
--- a/PWGLF/Tasks/Resonances/k892pmanalysis.cxx
+++ b/PWGLF/Tasks/Resonances/k892pmanalysis.cxx
@@ -382,7 +382,7 @@ struct k892pmanalysis {
         }
 
         // Checking whether the mid-rapidity condition is met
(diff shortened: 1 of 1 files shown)


BongHwi:
Hello @navneetkumar231295 , we have more `abs()` without `std::` in this task, such as in line 385.
Could you apply the update to all them together?

navneetkumar231295:
> Hello @navneetkumar231295 , we have more `abs()` without `std::` in this task, such as in line 385. Could you apply the update to all them together?

Hello, Bong. I made the changes

BongHwi (approved):

PR #8153, opened 2024-10-25
PWGJE/EMCal: Add option to shift cluster time in MC

jokonig:
- A shift in the cluster time was observed, caused by the tof to the EMCal surface.
- Discussed in https://indico.cern.ch/event/1469237/
- Two options are implemented:
- Case 1: Constant shift of 15ns
- Case 2: Eta dependent shift. Based on the lenght from vertex to the specific cell If applyCellTimeShift is set to 0, no shift is applied.


nzardosh:
@fjonasALICE  can i approve this?

fjonasALICE:
Yes, please approve. It was discussed in the emcal meeting. Should be enabled per default for MC

nzardosh (approved):

PR #8156, opened 2024-10-25
PWGLF: Add occupancy in the reso collision table

BongHwi:
- Add `collision.trackOccupancyInTimeRange()` in the `ResoCollisions` table for the occupancy study.


alibuild (approved):

PR #8146, opened 2024-10-25
PWGEM:LMee: added new cuts and histogram for opening angle

jjungIKF:

PR #8170, opened 2024-10-28
PWGEM/Dilepton: bug fix after #8100 and #8146

dsekihat:
PWGEM/Dilepton: bug fix
#8100
sign of M_PI/2. was wrong.
.2 was wrong.

#8146
Histogram name was wrong.
if-else logic was wrong.
The linear cut is part of the elliptic cut. Namely, the linear deta cut is equivalent at the limit of dphi-> infinity.
If one wants to apply linear deta cut, one should set cfg_min_dphi = 1e+10, and vice versa.


alibuild (dismissed):

alibuild (approved):

PR #8175, opened 2024-10-28
PWGCF: Added nogap method to FlowPtContainer

EmilGorm:


victor-gonzalez:
@EmilGorm many thanks for this!

Two main points from my side
- please, prefix the PR title with `PWGCF: ` (this is a small one but helps a lot)
- nowadays there are already a few active users of the Generic framework. I guess all the modifications will be backwards compatible. But it would be good if you announce the changes in the `CF towards o2` mattermost channel for everyone in CF to know and also asking for feedback when potential issues happen

EmilGorm:
Hi @victor-gonzalez 

Thanks for the review. I have modified the title now.
The code should be backward compatible, except for removing some of the constructors, which should be called through the initialize method instead. 

The only problem is that I have been removed from all my mattermost channels (seems my contract length had not been updated, but I have fixed this now). Would you happen to have, or do you know who has, the ability to add me back?

victor-gonzalez:
Hi @EmilGorm 
I tried to add you to the `CF towards o2` channel and got a message saying that you must be a member of the `ALICE team`
Try to join the `ALICE team` in mattermost and perhaps all the subscriptions will be back
I see you in the `CF towards o2` channel but as `offline`

PR #8176, opened 2024-10-28

yashpatley:


victor-gonzalez (approved):
Have in mind that you are basically replicating the information in the V0 table
At some point it would be more efficient to produce just a table with the strictly new information and join it to the V0 table. Unless you are pursuing stored derived data

PR #8188, opened 2024-10-28
[PWGEM] Add AnalysisType kDCA to analyse DCA 3D,xy,z together

feisenhu:
Include an additional AnalysisType `kDCA` to use when studding `DCAxy`, `DCAz` and `DCA3D` together in the same output.

To keep the option of choosing which component "3D,xy" or now also "z" should be used as DCA axis, I changed from a boolean to use DCAxy, to a string in which the component can be written `cfgDCAcase`.

### DIFF
--- a/PWGEM/Dilepton/Core/Dilepton.h
+++ b/PWGEM/Dilepton/Core/Dilepton.h
@@ -978,9 +978,6 @@ struct Dilepton {
         fRegistry.fill(HIST("Pair/") + HIST(event_pair_types[ev_id]) + HIST("lsmm/hs"), v12.M(), v12.Pt(), pair_dca, dphi, deta, weight);
       }
     } else if (cfgAnalysisType == static_cast<int>(o2::aod::pwgem::dilepton::utils::pairutil::DileptonAnalysisType::kDCA)) {
-      float deta = v1.Eta() - v2.Eta();
-      float dphi = v1.Phi() - v2.Phi();
-      o2::math_utils::bringToPMPi(dphi);
       if (t1.sign() * t2.sign() < 0) { // ULS
         fRegistry.fill(HIST("Pair/") + HIST(event_pair_types[ev_id]) + HIST("uls/hs"), v12.M(), v12.Pt(), pair_dca3D, pair_dcaxy, pair_dcaz, weight);
         if constexpr (pairtype == o2::aod::pwgem::dilepton::utils::pairutil::DileptonPairType::kDielectron) {
--- a/PWGEM/Dilepton/Core/Dilepton.h
+++ b/PWGEM/Dilepton/Core/Dilepton.h
@@ -115,7 +115,7 @@ struct Dilepton {
   Configurable<int> cfgNtracksPV08Min{"cfgNtracksPV08Min", -1, "min. multNTracksPV"};
   Configurable<int> cfgNtracksPV08Max{"cfgNtracksPV08Max", static_cast<int>(1e+9), "max. multNTracksPV"};
   Configurable<bool> cfgApplyWeightTTCA{"cfgApplyWeightTTCA", false, "flag to apply weighting by 1/N"};
-  Configurable<std::string> cfgDCAcase{"cfgDCAcase", "3D", "use string to choose which DCA component is used (default:3D, xy, z)"};
+  Configurable<int> cfgDCAcase{"cfgDCAcase", 0, "DCA component for output (3D:0, xy:1, z:2, all:3)"};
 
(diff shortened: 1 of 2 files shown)


PWGEM/Dilepton/Core/Dilepton.h
```diff
@@ -115,7 +115,7 @@ struct Dilepton {
   Configurable<int> cfgNtracksPV08Min{"cfgNtracksPV08Min", -1, "min. multNTracksPV"};
   Configurable<int> cfgNtracksPV08Max{"cfgNtracksPV08Max", static_cast<int>(1e+9), "max. multNTracksPV"};
   Configurable<bool> cfgApplyWeightTTCA{"cfgApplyWeightTTCA", false, "flag to apply weighting by 1/N"};
-  Configurable<bool> cfgUseDCAxy{"cfgUseDCAxy", false, "flag to use DCAxy, instead of DCA3D"};
+  Configurable<std::string> cfgDCAcase{"cfgDCAcase", "3D", "use string to choose which DCA component is used (default:3D, xy, z)"};
```
dsekihat, line 118: Can you use int, instead of string? 0 : 3D, 1:xy, 2:z.
Default should be 0.

hscheid: I agree with Daiki. The C++ string comparison is rather slow. I think this would not be a big hit in performance but I would consider it a good style to use int or an enum.
Some more optimisations should be taken care of in the code like unused variables.
However, a more fundamental question from my side would be if one needs all 3 pair-DCA values in one histogram. This is only necessary if the correlation between the three is to be studied. If I recall correctly this was done before. I am also unsure if we should not use 3D and adjust the templates properly. Also far as I remember the main problem was that the prompt templates (Jpsi) needed to be constructed at the reconstructed mass to account for a bias in the DCA xy due to Bremsstrahlung effects.

feisenhu: I have adapted it to use int.

From my point of view I think that we should keep looking into all three components. We already have prioritised different components for different reasons and it changed already a few times and I don't think that we should neglect any of those, yet. Especially when looking at the effect of the Bremsstrahlung. I agree that at some point we just use once of the component which might be the 3D case. But I don't think we are at that point yet. 

Furthermore, this change is just adding an additional AnalysisTyp. It adds the two additional DCA components only for this AnalysisType and not for all, so it will not affect other analysis if they don't use it.

alibuild (dismissed):

PWGEM/Dilepton/Core/Dilepton.h
```diff
@@ -115,7 +115,7 @@ struct Dilepton {
   Configurable<int> cfgNtracksPV08Min{"cfgNtracksPV08Min", -1, "min. multNTracksPV"};
   Configurable<int> cfgNtracksPV08Max{"cfgNtracksPV08Max", static_cast<int>(1e+9), "max. multNTracksPV"};
   Configurable<bool> cfgApplyWeightTTCA{"cfgApplyWeightTTCA", false, "flag to apply weighting by 1/N"};
-  Configurable<bool> cfgUseDCAxy{"cfgUseDCAxy", false, "flag to use DCAxy, instead of DCA3D"};
+  Configurable<int> cfgDCAcase{"cfgDCAcase", 0, "DCA component for output (3D:0, xy:1, z:2, all:3)"};
```
dsekihat, line 118: cfgDCAcase = 3 should not exist.

PWGEM/Dilepton/Core/Dilepton.h
```diff
@@ -946,7 +977,26 @@ struct Dilepton {
       } else if (t1.sign() < 0 && t2.sign() < 0) { // LS--
         fRegistry.fill(HIST("Pair/") + HIST(event_pair_types[ev_id]) + HIST("lsmm/hs"), v12.M(), v12.Pt(), pair_dca, dphi, deta, weight);
       }
-
+    } else if (cfgAnalysisType == static_cast<int>(o2::aod::pwgem::dilepton::utils::pairutil::DileptonAnalysisType::kDCA)) {
+      float deta = v1.Eta() - v2.Eta();
```
hscheid, line 981: In case of the `kDCA`, I dont think you need to calculate the dphi and data, right?

PWGEM/Dilepton/Core/Dilepton.h
```diff
@@ -805,15 +825,26 @@ struct Dilepton {
...
+      if (cfgDCAcase == 1 || cfgDCAcase == 3) {
+        dcaxy_t1 = t1.dcaXY() / std::sqrt(t1.cYY());
+        dcaxy_t2 = t2.dcaXY() / std::sqrt(t2.cYY());
+        pair_dcaxy = std::sqrt((dcaxy_t1 * dcaxy_t1 + dcaxy_t2 * dcaxy_t2) / 2.);
+        pair_dca = pair_dcaxy;
+      }
+      if (cfgDCAcase == 2 || cfgDCAcase == 3) {
+        dcaz_t1 = t1.dcaZ() / std::sqrt(t1.cZZ());
+        dcaz_t2 = t2.dcaZ() / std::sqrt(t2.cZZ());
+        pair_dcaz = std::sqrt((dcaz_t1 * dcaz_t1 + dcaz_t2 * dcaz_t2) / 2.);
```
hscheid, line 846: https://github.com/AliceO2Group/O2Physics/pull/8188/files#r1820689972

PWGEM/Dilepton/Core/Dilepton.h
```diff
@@ -805,15 +825,26 @@ struct Dilepton {
...
-        pair_dca = std::sqrt((dca_t1 * dca_t1 + dca_t2 * dca_t2) / 2.);
+      dca3D_t1 = dca3DinSigma(t1);
+      dca3D_t2 = dca3DinSigma(t2);
+      pair_dca3D = std::sqrt((dca3D_t1 * dca3D_t1 + dca3D_t2 * dca3D_t2) / 2.);
+      pair_dca = pair_dca3D;
+      if (cfgDCAcase == 1 || cfgDCAcase == 3) {
+        dcaxy_t1 = t1.dcaXY() / std::sqrt(t1.cYY());
+        dcaxy_t2 = t2.dcaXY() / std::sqrt(t2.cYY());
+        pair_dcaxy = std::sqrt((dcaxy_t1 * dcaxy_t1 + dcaxy_t2 * dcaxy_t2) / 2.);
+        pair_dca = pair_dcaxy;
```
hscheid, line 841: Using this syntax makes case `cfgDCAcase = 3` obsolete. You only fill the histogram once.

PWGEM/Dilepton/Core/Dilepton.h
hscheid: So. The most interesting is the 3D DCA; from my point of view, this should be the one we follow up on. The only need to have all 3 versions of the DCA in the histogram is if you want to do a correlation of DCAz and DCAxy for example. If not, a setup with 3 sub wagons for each DCA is more efficient. 3 histograms with 3 dimensions should take less disc space than a 5D histogram.
And the problem with 3D was only the Jpsi peak moving in the tail right? This can be dealt with by constructing the template at the correct mass.
So: Is there any reason to have DCAxy, DCAz and DCA3D in the same histogram, or do you need one histogram for each one?
I made some more comments on the readability and performance of the code. This can be done later and is not necessary now.

hscheid: To make the problem clear. At the moment the default value of bins is 30 bins in DCA. So adding two axes with 30 bins increases the memory consumption by this Sparse up to a factor 900. And this is only for the standard up to DCA = 10. I guess you have more bins in your analysis.
In ULS and LS this might be Ok since a lot of bins are empty and in the THnSparse implementation, this is not counting then. However, in mixed events, most bins should be filled and the memory will go through the roof. So, if not absolutely necessary (and right now I don't see why it is) I suggest using sub wagons.

PR #8192, opened 2024-10-28
[PWGCF] Use global track selection

ilikmeta:


victor-gonzalez (approved):

victor-gonzalez:
I would suggest using named constants for the bins of the event count histogram
The code will be more readable and less error prone

ilikmeta:
Could you please ellaborate on your comment? Are you referring to this one for example: `registry.fill(HIST("hEventCount"), 0.5);`  Thanks

victor-gonzalez:
Yes, I was referring to that

Say that there is defined an `enum`
```
enum eventprogress {
  kFILTERED,
  kSEL8,
  kOCCUPANCY,
  kTVXINTRD,
  ...  
  kNOOFEVENTSTEPS
};
```
then the event histogram can be defined as of kind
```
{HistType::kTH1D, {{kNOOFEVENTSTEPS, - 0.5, kNOOFEVENTSTEPS - 0.5}}}
```
The filling then will be done
```
  registry.fill(HIST("hEventCount"), kFILTERED);
```
As I said, easier to read, easier to expand, and less error prone

PR #8197, opened 2024-10-29
EventFiltering/PWGHF: fix B 3prong in case of double mass hypo for D0

fgrosa:


alibuild (approved):

PR #8199, opened 2024-10-29
Add the PR cleaner workflow

ktf:


ktf:
Not really checked. Merging.

vkucera:
The workflow is failing `checkov` with
```
CKV_GHA_7: "The build output cannot be affected by user parameters other than the build entry point and the top-level source location. GitHub Actions workflow_dispatch inputs MUST be empty. "
```

PR #8201, opened 2024-10-29
PWGCF: flow: add Run3ITSall7Layers option and a new run-by-run task

Luzhiyongg:
Hi reviewers, 
    The `FlowTask.cxx` I am using has already consume a lot of memory (>500 MB per wagon). So I'd like to create to new task to hold the run-by-run output. This task would be lightweight for each run.


victor-gonzalez (approved):
I would suggest to have a look at my comments for the next iteration

PR #8202, opened 2024-10-29
[PWGDQ] add TPCnCRoverFindCls as variable 

feisenhu:
Additionally it  is adding a cut setting to do a comparison between DQ and EM framework


alibuild (approved):

PR #8226, opened 2024-10-29
PWGCF: Add PDG code selections for the MC particle correlations

jaelpark:


PWGCF/Tasks/correlations.cxx
```diff
@@ -724,8 +741,11 @@ struct CorrelationTask {
       case 2212: // proton
       case -2212:
         return 2;
-      default:
+      case 421: // D0
+      case -421:
         return 3;
+      default:
+        return 4;
```
victor-gonzalez, line 748: I would suggest using named constants from `TPDGCode.h` extended by Common/Constants/include/CommonConstants/PhysicsConstants.h if needed

victor-gonzalez (approved):

jgrosseo:
This looks good to me. I let you see with Victor's comment.

PWGCF/Tasks/correlations.cxx
jaelpark: Yes, this is something we should probably change eventually.

PR #8231, opened 2024-10-30
DPL Analysis: rename `aod::Hashes` for upcoming O2 update

aalkin:
Hello!

Upcoming O2 update will introduce `aod::Hash` template used in the core of refactored table types and providing mapping between string hash and the string. Unfortunately, it would clash with the aliases already defined in O2Physics. To ensure that O2 update would be seamless and do not require any interventions, I've took the liberty of renaming the `aod::Hashes` tables into `aod::MixingHashes` and the clashing alias to `using MixingHash = aod::MixingHashes::iterator;`.

### DIFF
--- a/PWGCF/DataModel/FemtoDerived.h
+++ b/PWGCF/DataModel/FemtoDerived.h
@@ -451,8 +451,8 @@ namespace hash
 {
 DECLARE_SOA_COLUMN(Bin, bin, int); //! Hash for the event mixing
 } // namespace hash
-DECLARE_SOA_TABLE(Hashes, "AOD", "HASH", hash::Bin);
-using Hash = Hashes::iterator;
+DECLARE_SOA_TABLE(MixingHashes, "AOD", "HASH", hash::Bin);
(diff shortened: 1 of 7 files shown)


ddobrigk (approved):

saganatt:
Thanks Anton, no obstacles from my side.
I take the opportunity to ask @zchochul @prchakra @Eloviyo (FemtoWorld/FemtoUniverse) and @lauraser @ariedel-cern (FemtoDream) - are you actually using mixing hash tasks in your analysis workflows?

Eloviyo:
I am not using it in my analysis, so no problem from my side.

prchakra:
Okay from my side too!

ariedel-cern:
Hi, I think a previous developer planned it, but we did not use it. So, also an okay from my side.

aalkin:
This still needs an approval on behalf of CF code owner to be merged.

saganatt (approved):
Hi, so I am approving it.

I think we could delete the CF hash codes soon if not now.

PR #8244, opened 2024-10-30
[PWGHF] Modified Resonance Data Creator to include muons

Luca610:
This PR introduces some small changes to the MC matching to treat the case in which one of daughter tracks is a muon from pi decay.


fgrosa (approved):
Ciao @Luca610 thanks!

PR #8267, opened 2024-11-01
[PWGDQ,DPG] adding a maximum pT for downscaling

amaringarcia:


amaringarcia:
I will try to solved the issues

PR #8268, opened 2024-11-01
[DPG,PWGDQ] adding a configurable, maximum Pt for applying downsampling factor  p…

amaringarcia:
…ions, protons electrons


iarsene (approved):

noferini (approved):
Hi saw that you applied a 1000 downscale factor which is common to all particles and not-configurable. If this is enough for you purpose it is fine with me. Otherwise, please consider to make it more general. In the meanwhile I approve it

chiarazampolli:
Hello @amaringarcia , 

For what is this task used? For the AODs from reco, of for derived data? I assume the latter, please confirm, so that we are sure it does *not* affect production (there is also no label).

Thanks,

Chiara

amaringarcia:
Hi Chiara,


this is for derived data.  Skimmed trees for TPC dEdx calibration. The default value (pt=100GeV)  should let the output untouched.


Regards,

Ana

ddobrigk:
Dear Ana, sorry to jump in. Please remember to reserve the word "skimming" for the software triggers only. This use, as far as I understand, is standard derived data. Thank you!

chiarazampolli:
Thanks, and yes, "skimming" is a word that we use in production, to create the skimmed CTFs, so it'd be good to avoid it if used in a different context.
Chiara

PR #8279, opened 2024-11-02
[PWGCF] FemtoUniverse : Adding randomness in mixing

prchakra:

PR #8289, opened 2024-11-03
[PWGCF] FemtoUniverse : Adding randomness in mixing for SH 

prchakra:
In this commit, I fixed the memory allocation and de-allocation issue that was earlier present in the pair-loop for same events. Also, I added randomness in making pair from mixed events (similar to the PR: https://github.com/AliceO2Group/O2Physics/pull/8279).


victor-gonzalez (approved):
@prchakra  It sounds good to me

PR #8299, opened 2024-11-04
[PWGLF] Add cuts pT max and min for tree saving

m-rasa:
Adding cuts on pT value, both min and max, in order to reduce the output of the derived table produced.


njacazio (approved):

vkucera:
@m-rasa Missing includes reported by cpplint.

PR #8304, opened 2024-11-04
[PWGHF] add protection in case of non filled metadata

fgrosa:


mfaggin (dismissed):

alibuild (approved):

PR #8308, opened 2024-11-05
[PWGHF] Ds-h correlation derived data - correction table names

scattaru:
Dear all,
I modified the table names in the DataModel to address an error with unbound indices. Let me know for any comments or suggestions.

Cheers,
Samuele

### DIFF
--- a/PWGHF/HFC/DataModel/DerivedDataCorrelationTables.h
+++ b/PWGHF/HFC/DataModel/DerivedDataCorrelationTables.h
@@ -26,7 +26,7 @@ DECLARE_SOA_COLUMN(Multiplicity, multiplicity, float); //! Event multiplicity
 DECLARE_SOA_COLUMN(PosZ, posZ, float);                 //! Primary vertex z position
 
 } // namespace hf_collisions_reduced
-DECLARE_SOA_TABLE(HfRedCollisions, "AOD", "COLLREDUCED", //! Table with collision info
+DECLARE_SOA_TABLE(HfRedCollisions, "AOD", "HFREDCOLLISION", //! Table with collision info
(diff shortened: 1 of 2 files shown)


PWGHF/HFC/DataModel/DerivedDataCorrelationTables.h
```diff
@@ -26,7 +26,7 @@ DECLARE_SOA_COLUMN(Multiplicity, multiplicity, float); //! Event multiplicity
 DECLARE_SOA_COLUMN(PosZ, posZ, float);                 //! Primary vertex z position
 
 } // namespace hf_collisions_reduced
-DECLARE_SOA_TABLE(HfRedCollisions, "AOD", "COLLREDUCED", //! Table with collision info
+DECLARE_SOA_TABLE(HfRedCollisions, "AOD", "HFREDCOLLISION", //! Table with collision info
```
apalasciano, line 29: Can you please modify the name of **HfRedCollisions**. It is already used in another data model ([PWGHF/D2H/DataModel/ReducedDataModel.h](https://github.com/AliceO2Group/O2Physics/blob/46201c5e858faad90f1dd0b6ecec2c1a40f7e694/PWGHF/D2H/DataModel/ReducedDataModel.h#L54)) and it would be better to avoid duplications. For example, also something like this can work:
```suggestion
DECLARE_SOA_TABLE(HfcRedCollisions, "AOD", "HFCREDCOLLS", //! Table with collision info
```

apalasciano (changes requested):
Ciao Samuele, 
I have only a minor comment:

apalasciano (approved):

PR #8310, opened 2024-11-05
[PWGCF] FemtoUniverse: Replacing MutlNtr with MultV0M in colBinning

prchakra:
In this PR, I replace MutlNtr with MultV0M in colBinning for event mixing.


alibuild (approved):

PR #8312, opened 2024-11-05
[PWGDQ] Update tableMaker_withAssoc.cxx

shreyasiacharya:
1. Added a histogram to determine the percentage of orphan tracks.
2. Now, we store the original event ID in the "trackBasic" table instead of the first amongst the associated collisions. This is necessary for calculation of the proper delta-time (track time - collision time with BC offset), example when using the TableReader_withAssoc


iarsene (approved):

vkucera:
@shreyasiacharya Many errors reported by cpplint.

iarsene:
@vkucera We will try to fix them in a separate commit

PR #8322, opened 2024-11-06
[Common,PWGCF,PWGDQ,PWGHF,PWGLF,Trigger] DPL Analysis: more compatibility changes for the upcoming O2 update

aalkin:
This is (hopefully) the last compatibility change in O2Physics to unblock updating of O2.

Important notes:
* `aod-data-model-graph` is temporarily disabled, as it is impossible to make it compile with both versions of O2 - it uses a lot of internal functions. After O2 is update, I will move framework-specific back into O2, so that data model graph can use abstracted API instead of relying on internals
* CEPF task reverted to use `persistent_columns_t` temporarily, same as with the data model graph I will move it to dedicated API that hides the framework internals
* PWGCF/Femto3D: declaration of versioned tables using existing `_VERSIONED` macro
* PWGDQ: the table description together with the table will become the primary distinctive feature of the table declaration, it will define the set of columns. Therefore different tables should have different descriptions. Additionally, the declarations are repeated 3 times in different tasks, consider moving them to a single header. Please also note that changed description also means you need to change the configuration of aod-writer if you want to store these tables.
* PWGHF:  declaration of versioned tables using existing `_VERSIONED` macro
* PWGLF:  declaration of staged/versioned tables using existing `_VERSIONED` macro; `Spawns<>` requires a table type, and not an iterator alias, it was functional essentially by accident.


njacazio:
Thanks a lot Anton

njacazio (approved):

vkucera (approved):

ddobrigk (approved):
Approving on behalf of common (small and uncontroversial)

aalkin:
The approvals for trigger, DQ and CF are still required. 
@mpuccio, @iarsene, @saganatt could you please add your review?

victor-gonzalez (approved):
Thanks Anton!

mpuccio (approved):

aalkin:
An approval for DQ is still required.
@iarsene @dsekihat @feisenhu @lucamicheletti93 could you please take a look?

iarsene (approved):

PR #8342, opened 2024-11-07
[PWGDQ] Add quarkonia to hyperon-antihyperon analysis tasks

romainschotter:
- Add quarkonia to hyperon-antihyperon analysis task (`qqbar2hypahyp.cxx`)
- Add task to store the number of generated quarkonia (`generatedQuarkonia.cxx`)


romainschotter:
@ddobrigk

iarsene:
Hi @romainschotter,
I was looking through the code you add with this PR, but I see no connection to the framework we have implemented in PWGDQ. There is no interconnection to the existing framework, so I don't see the reason why should this be commited in DQ. You can make a presentation for tomorrow's DQ O2 meeting so we can discuss the posibilities.
Ionut

ddobrigk:
Hi @iarsene , of course, we would be happy to discuss different approaches for this analysis to be carried out. However, let me point out that PWG-DQ is a physics working group to deal with quarkonia analysis and this code targets quarkonia analysis, and there's a priori no requirement that all analyses in a given PWG be carried out with one single approach, at least to my knowledge. I should also point out this effort and the approach we are taking to it were discussed with PWG-LF and PWG-DQ conveners as well as jpsi2ee and strangeness PAG coordinators via mail already and it was already agreed that the code should go to DQ, with further discussion to take place as the effort evolves. You have also been added to this mail thread by Romain for completeness.

In this particular case, and on a more technical note, please note that we are taking advantage of a long-standing LF effort in generating general-purpose derived data containing V0s and Cascades, complete with track indexing and complementary collision information, using the standard V0 and Cascade data model that is already used in several hundreds of files in O2Physics. It is also important to note that having derived data with both V0s and Cascades has to be a central effort, as the entire data volume is not negligible and we want to avoid duplication of stored derived data. 

That being said, we can certainly discuss if there are alternate approaches that could be advantageous, as there may be further use cases of LF/DQ cross-pwg analyses or other possibilities that Romain and I may not have considered: we would be very happy to hear your input on what we are doing and plan for the next steps.  Note only that Romain and I will attend the DPG AOT meeting tomorrow morning, so unfortunately this Thursday we are not available. Thanks a lot!

iarsene (approved):

PR #8344, opened 2024-11-07
[PWGHF] fix the bug of pid and add QA histograms

zhangbiao-phy:


zhangbiao-phy:
Hi @fgrosa, there is the PR to fix the bug and add QA plots

fgrosa (approved):
Hi @zhangbiao-phy, good catch, thanks!

PR #8345, opened 2024-11-07
[Common] RecoDecay: Add matching for decaying tracks in getMatchedMCRec

Luca610:
This PR introduces the option to match cases with one or more decayed tracks in the function getMatchedMCRec.
@vkucera  let me know if it works for you!

### DIFF
--- a/Common/Core/RecoDecay.h
+++ b/Common/Core/RecoDecay.h
@@ -690,10 +690,8 @@ struct RecoDecay {
     // Printf("MC Rec: Expected mother PDG: %d", PDGMother);
     int8_t coefFlavourOscillation = 1;     // 1 if no B0(s) flavour oscillation occured, -1 else
     int8_t sgn = 0;                        // 1 if the expected mother is particle, -1 if antiparticle (w.r.t. PDGMother)
-    if constexpr (acceptTrackDecay) {
-      int8_t nPiToMuLocal = 0; // number of pion prongs decayed to a muon
-      int8_t nKaToPiLocal = 0; // number of kaon prongs decayed to a pion
-    }
+    int8_t nPiToMuLocal = 0; // number of pion prongs decayed to a muon
+    int8_t nKaToPiLocal = 0; // number of kaon prongs decayed to a pion
     int indexMother = -1;                  // index of the mother particle
     std::vector<int> arrAllDaughtersIndex; // vector of indices of all daughters of the mother of the first provided daughter
     std::array<int, N> arrDaughtersIndex;  // array of indices of provided daughters
--- a/Common/Core/RecoDecay.h
+++ b/Common/Core/RecoDecay.h
@@ -24,6 +24,7 @@
 #include <vector>    // std::vector
 
 #include "TMCProcess.h" // for VMC Particle Production Process
+#include "TPDGCode.h" // for PDG codes
 #include "CommonConstants/MathConstants.h"
 
 /// Base class for calculating properties of reconstructed decays
@@ -688,6 +689,8 @@ struct RecoDecay {
     // Printf("MC Rec: Expected mother PDG: %d", PDGMother);
     int8_t coefFlavourOscillation = 1;     // 1 if no B0(s) flavour oscillation occured, -1 else
     int8_t sgn = 0;                        // 1 if the expected mother is particle, -1 if antiparticle (w.r.t. PDGMother)
+    int8_t piToMu = 0;                     // number of pions decayed to a muon >=1 if any are found
+    int8_t kaToPi = 0;                     // number of kaons decayed to a pion >=1 if any are found
     int indexMother = -1;                  // index of the mother particle
     std::vector<int> arrAllDaughtersIndex; // vector of indices of all daughters of the mother of the first provided daughter
     std::array<int, N> arrDaughtersIndex;  // array of indices of provided daughters
--- a/Common/Core/RecoDecay.h
+++ b/Common/Core/RecoDecay.h
@@ -665,7 +665,7 @@ struct RecoDecay {
   /// Checks whether the reconstructed decay candidate is the expected decay.
   /// \param checkProcess  switch to accept only decay daughters by checking the production process of MC particles
   /// \param acceptIncompleteReco  switch to accept candidates with only part of the daughters reconstructed
-  /// \param acceptTrackDecay  switch to accept candidates with one or more tracks decayed during transport
+  /// \tparam acceptTrackDecay  switch to accept candidates with daughter tracks of pions and kaons which decayed
   /// \param particlesMC  table with MC particles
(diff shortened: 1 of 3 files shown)


vkucera:
Hi @Luca610 , please elaborate more about the problem you are trying to tackle with this implementation and why `getMatchedMCRec` is the right place to put it. To be honest, from the first look, I don't like the idea of adding this very species-specific feature inside a generic matching method if it doesn't solve a problem that cannot be done outside and doesn't improve the matching result either.

Luca610:
Hi @vkucera, when trying to match some decay such as K0s->pi+pi- it was observed that some reconsructed particles where not correctly matched causing a visible peak in the invariant mass distribution of unmatched candidates. This is caused by the fact that one of the pions decayed into a muon during the simulation and the track is then associated to a mu. For a reference, this is the case of around 2% of K0s. I beleive the same could happen with K->pi.  I think that this could be the place to add it because many analyses could profit from it, preventing code duplication.

Luca610:
Hi @vkucera, yes this is the case. I don't know if it is an issue or a feature of simulations: these tracks degrade the resolution on the reconstructed invariant mass, so it can be usefull to tag them separately to study this effect.  Anyway this changes allow to account for this effect and easily match them with RecoDecay.

vkucera:
Thanks @Luca610 , I understand the problem now. Your solution looks fine to me. Let me just suggest some improvements in the code.

Common/Core/RecoDecay.h
```diff
@@ -663,26 +665,33 @@ struct RecoDecay {
...
-                             int depthMax = 1)
+                             int depthMax = 1,
+                             int8_t* nPiToMu = nullptr,
+                             int8_t* nKaToPi = nullptr)
   {
     // Printf("MC Rec: Expected mother PDG: %d", PDGMother);
     int8_t coefFlavourOscillation = 1;     // 1 if no B0(s) flavour oscillation occured, -1 else
     int8_t sgn = 0;                        // 1 if the expected mother is particle, -1 if antiparticle (w.r.t. PDGMother)
+    int8_t piToMu = 0;                     // number of pions decayed to a muon >=1 if any are found
+    int8_t kaToPi = 0;                     // number of kaons decayed to a pion >=1 if any are found
```
vkucera, line 694:
```suggestion
    if constexpr (acceptTrackDecay) {
      int8_t nPiToMuLocal = 0; // number of pion prongs decayed to a muon
      int8_t nKaToPiLocal = 0; // number of kaon prongs decayed to a pion
    }
```

Common/Core/RecoDecay.h
```diff
@@ -708,6 +717,17 @@ struct RecoDecay {
...
+      if (acceptTrackDecay) {
+        if (std::abs(particleI.pdgCode()) == std::abs(kMuonPlus) && std::abs(motherI.pdgCode()) == std::abs(kPiPlus)) {
+          // LOGF(info, "found muon %d with mother %d", particleI.pdgCode(), motherI.pdgCode());
+          piToMu += 1;
+          particleI = motherI;
+        } else if (std::abs(particleI.pdgCode()) == std::abs(kPiPlus) && std::abs(motherI.pdgCode()) == std::abs(kKPlus)) {
+          kaToPi += 1;
+          particleI = motherI;
+        }
+      }
```
vkucera, line 730:
```suggestion
      if constexpr (acceptTrackDecay) {
        // Replace the MC particle associated with the prong by its mother for π → μ and K → π.
        auto motherI = particleI.template mothers_first_as<T>();
        auto pdgI = std::abs(particleI.pdgCode());
        auto pdgMotherI = std::abs(motherI.pdgCode());
        if (pdgI == kMuonMinus && pdgMotherI == kPiPlus) {
          // π → μ
          nPiToMuLocal++;
          particleI = motherI;
        } else if (pdgI == kPiPlus && pdgMotherI == kKPlus) {
          // K → π
          nKaToPiLocal++;
          particleI = motherI;
        }
      }
```

Common/Core/RecoDecay.h
```diff
@@ -663,26 +665,33 @@ struct RecoDecay {
   /// Checks whether the reconstructed decay candidate is the expected decay.
   /// \param checkProcess  switch to accept only decay daughters by checking the production process of MC particles
   /// \param acceptIncompleteReco  switch to accept candidates with only part of the daughters reconstructed
+  /// \param acceptTrackDecay  switch to accept candidates with one or more tracks decayed during transport
```
vkucera, line 668:
```suggestion
  /// \tparam acceptTrackDecay  switch to accept candidates with daughter tracks of pions and kaons which decayed
```

Common/Core/RecoDecay.h
```diff
@@ -780,6 +800,12 @@ struct RecoDecay {
     if (sign) {
       *sign = sgn;
     }
+    if (nPiToMu) {
+      *nPiToMu = piToMu;
+    }
+    if (nKaToPi) {
+      *nKaToPi = kaToPi;
+    }
```
vkucera, line 808:
```suggestion
    if constexpr (acceptTrackDecay) {
      if (nPiToMu) {
        *nPiToMu = nPiToMuLocal;
      }
      if (nKaToPi) {
        *nKaToPi = nKaToPiLocal;
      }
    }
```

Common/Core/RecoDecay.h
```diff
@@ -663,26 +665,33 @@ struct RecoDecay {
...
+  /// \param acceptTrackDecay  switch to accept candidates with one or more tracks decayed during transport
   /// \param particlesMC  table with MC particles
   /// \param arrDaughters  array of candidate daughters
   /// \param PDGMother  expected mother PDG code
   /// \param arrPDGDaughters  array of expected daughter PDG codes
   /// \param acceptAntiParticles  switch to accept the antiparticle version of the expected decay
   /// \param sign  antiparticle indicator of the found mother w.r.t. PDGMother; 1 if particle, -1 if antiparticle, 0 if mother not found
   /// \param depthMax  maximum decay tree level to check; Daughters up to this level will be considered. If -1, all levels are considered.
+  /// \param nPiToMu  number of pions decayed in a muon during transport
+  /// \param nKaToPi  number of kaons decayed in a pion during transport
```
vkucera, line 677:
```suggestion
  /// \param nPiToMu  number of pion prongs decayed to a muon
  /// \param nKaToPi  number of kaon prongs decayed to a pion
```

vkucera (changes requested):

Common/Core/RecoDecay.h
Luca610: Testing locally this change causes the compilation to break. Can I define these variables outside the if statement?

vkucera (approved):

ddobrigk (approved):

PR #8355, opened 2024-11-08
[PWGLF/Nuspex] added configurable for pt coverage of TPC nsigma

chiarapinto:


maciacco (dismissed):

maciacco (approved):

vkucera:
@chiarapinto @maciacco I fixed the title prefix to avoid the duplication. The generated prefix was automatically added because no valid prefix was recognised, as the original prefix "[PWGLF]/Nuspex: " does not have a valid syntax. Please check these before enabling merging.

chiarapinto:
@vkucera well, also with the duplication it was clear that it was a PWGLF commit :) but ok, I'll be more careful, thanks!

PR #8369, opened 2024-11-10
[PWGCF] [PWG-CF] new utility functions, clean up, improved log messages, supp…

abilandz:
…ort for occupancy

### DIFF
--- a/PWGCF/MultiparticleCorrelations/Core/MuPa-MemberFunctions.h
+++ b/PWGCF/MultiparticleCorrelations/Core/MuPa-MemberFunctions.h
@@ -120,8 +120,9 @@ void DefaultConfiguration()
   tc.fDryRun = cf_tc.cfDryRun;
   tc.fVerbose = cf_tc.cfVerbose;
   if (tc.fVerbose) {
-    LOGF(info, "\033[1;32m%s\033[0m", __FUNCTION__); // yes, here
+    StartFunction(__FUNCTION__); // yes, here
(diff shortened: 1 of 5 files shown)


PWGCF/MultiparticleCorrelations/Core/MuPa-MemberFunctions.h
```diff
@@ -8561,6 +8841,27 @@ void Trace(const char* functionName, Int_t lineNumber)
...
+} // void StartFunction(const char* functionName)
+
+//============================================================
+
+void ExitFunction(const char* functionName)
+{
+  // A simple utility wrapper, used when tc.fVerbose = kTRUE. It merely ensures uniform formatting of notification when the function exits.
+
+  LOGF(info, "\033[1;32mExit %s\033[0m\n", functionName); // prints in green
+  sleep(0.1);
```
victor-gonzalez, line 8859: Is this really needed in here?
I don't know if we are allowed to use `sleep`
Especially after a `LOGF` which is automatically handled by the Fair log framework
I need to check it

victor-gonzalez (approved):

PWGCF/MultiparticleCorrelations/Core/MuPa-MemberFunctions.h
abilandz: This was used only during debugging, but I have removed it now.

PR #8371, opened 2024-11-10
[PWGLF] Handle kinks in association for V0/casc labeler

ddobrigk:
This PR ensures the labeler service provided to help with V0 and cascade analysis still labels V0s/cascades correctly in case a kink decay (pion -> muon + antineutrino) is found in the MC association step. This happens frequently, as the TPC track carries the label and thus a plain de-reference of a V0 that includes a pion that decayed into a muon will have a muon+X in its association. Note that such cases will also be flagged as fake associations by corresponding bits of the track table (if checked). 

After this PR is merged, the label will still point to the originating particle regardless, meaning a case in which e.g. Lambda -> p + pi -> p + mu + antineutrino, with the V0 made up of (p+mu), will be correctly labeled with the `mcParticle` label for the Lambda. For general use, *please note that the corresponding prong will still irrevocably carry the muon information*, so care must be taken if the prong pdg is checked explicitly in analysis. 

Please note that this PR has been marked draft pending some further checks and verifications.

### DIFF
--- a/PWGLF/TableProducer/Strangeness/cascademcbuilder.cxx
+++ b/PWGLF/TableProducer/Strangeness/cascademcbuilder.cxx
@@ -94,32 +94,33 @@ struct cascademcbuilder {
   mcCascinfo thisInfo;
   //*+-+*+-+*+-+*+-+*+-+*+-+*+-+*+-+*+-+*+-+*
 
-  // kink handling 
-  template <typename mcpart> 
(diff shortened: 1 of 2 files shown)


ddobrigk:
Tagging @romainschotter @fgrosa @ChiaraDeMartin95 @ercolessi @Luca610 @nepeivodaRS @GijsvWeelden as people with which this topic has been discussed recently.

PR #8373, opened 2024-11-10

yashpatley:
Added ccdb service to read correction factors from TH3F histograms 
Added new TH3F histograms to compute pT, eta, posZ dependent efficiencies

### DIFF
--- a/PWGCF/TwoParticleCorrelations/Tasks/lambdaR2Correlation.cxx
+++ b/PWGCF/TwoParticleCorrelations/Tasks/lambdaR2Correlation.cxx
@@ -25,6 +25,7 @@
 #include "PWGLF/DataModel/LFStrangenessTables.h"
 #include "CommonConstants/PhysicsConstants.h"
 #include "Common/Core/RecoDecay.h"
+#include "CCDB/BasicCCDBManager.h"
 #include "TPDGCode.h"
 
 using namespace o2;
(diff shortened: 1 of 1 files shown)


PWGCF/TwoParticleCorrelations/Tasks/lambdaR2Correlation.cxx
```diff
@@ -1136,74 +1101,67 @@ struct lambdaCorrelationAnalysis {
...
-        get_corr_factor<kLambda>(eff_2, p2.pt());
-      } else if (part_pair == kAntiLambdaAntiLambda) {
-        get_corr_factor<kAntiLambda>(eff_1, p1.pt());
-        get_corr_factor<kAntiLambda>(eff_2, p2.pt());
-      }
-    }
+    float corfac1 = 1., corfac2 = 1.;
+
+    get_corr_factor<rec_gen>(corfac1, col, p1);
+    get_corr_factor<rec_gen>(corfac2, col, p2);
```
victor-gonzalez, line 1107: This is highly inefficient
Each track correction factor is search a number of times equal to the multiplicity
If the multiplicity is only the number of Lambdas then the impact will be small
But if there are execution time issues definitely this would need to be modified

victor-gonzalez (approved):
Have a look at my comment

PWGCF/TwoParticleCorrelations/Tasks/lambdaR2Correlation.cxx
yashpatley: Dear Victor,
I understand this. I will try to add these correction factors in the table itself while producing tables. Would that be efficient ?

victor-gonzalez: Initially, that sound a good approach

PR #8376, opened 2024-11-11
[PWGJE] Updated PhotonIsolation.QA

berend01:
For MC data the collision weights are now processed


nzardosh:
Hi Berend, Is there a reason the jet derived data model is not used in this task?

nzardosh (approved):

PR #8390, opened 2024-11-11
[PWGLF] piHe3 analysis

blacwovie:
based on lithium4analysis code. Add pion pid.

### DIFF
--- a/PWGLF/DataModel/piHe3Tables.h
+++ b/PWGLF/DataModel/piHe3Tables.h
@@ -1,0 +1,124 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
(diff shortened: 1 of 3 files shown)


fmazzasc (changes requested):
Hi @blacwovie , thank you very much for your work! 
My main suggestion is to avoid code duplication and modify directly the li4 task in the following way:
- Rename `lithium4analysis.cxx` into` he3HadronFemto.cxx` and` LFLithiumTables.h` into `LFhe3HadronTables.h`
- Change in` LFhe3HadronTables.h` the proton columns like: `NSigmaTPCPr --> NSigmaTPCHad` or `Chi2TPCPr --> Chi2TPCHad`
- In `he3HadronFemto.cxx` add a switch flag to either select pions or protons

blacwovie:
Hi @fmazzasc,Thank you for the suggestion. I will modify directly the li4 task as you suggested.Then I'll creat another PR.

PR #8393, opened 2024-11-11
[PWGCF] Efficiency corrections for pt and centrality

ilikmeta:


PWGCF/Flow/Tasks/GFWEfficiency.cxx
```diff
@@ -0,0 +1,148 @@
...
+  {
+    if (abs(pdg) == 211)
+      return true;
+    if (abs(pdg) == 321)
+      return true;
+    if (abs(pdg) == 2212)
+      return true;
+    if (abs(pdg) == 11)
+      return true;
+    if (abs(pdg) == 13)
```
victor-gonzalez, line 70: Please, use named constants from `TPDGCodes.h` (https://root.cern/doc/v632/TPDGCode_8h.html)

victor-gonzalez (changes requested):
Although having a separate task for efficiency extraction is possible it has a big drawback in the fact that the configuration must replicate the one in the actual analysis

I would suggest to consider having the efficiency extraction as an additional, optional, process function on the own analysis task. This will easy the configuration consistency between analysis and efficiency extraction. The involved histograms can be easily incorporated, or not, according to the activation, or not, of the efficiency extraction process

PR #8397, opened 2024-11-11
[Trigger,PWGHF] update produceSample task

apalasciano:


EventFiltering/PWGHF/HFFilterPrepareMLSamples.cxx
```diff
@@ -54,8 +55,12 @@ struct HfFilterPrepareMlSamples { // Main struct
 
   // parameters for production of training samples
   Configurable<bool> fillSignal{"fillSignal", true, "Flag to fill derived tables with signal for ML trainings"};
-  Configurable<bool> fillBackground{"fillBackground", true, "Flag to fill derived tables with background for ML trainings"};
+  Configurable<bool> fillOnlyBackground{"fillOnlyBackground", true, "Flag to fill derived tables with background for ML trainings"};
   Configurable<float> downSampleBkgFactor{"downSampleBkgFactor", 1., "Fraction of background candidates to keep for ML trainings"};
+  Configurable<float> leftLimitLSB{"leftLimitLSB", 1.72, "Left Sideband Lower Minv limit"};
+  Configurable<float> rightLimitLSB{"rightLimitLSB", 1.78, "Left Sideband Upper Minv limit"};
+  Configurable<float> leftLimitRSB{"leftLimitRSB", 1.94, "Right Sideband Lower Minv limit"};
+  Configurable<float> rightLimitRSB{"rightLimitRSB", 1.98, "Right Sideband Upper Minv limit"};
```
fgrosa, line 63: specify in the name of the variables and/or description that this applies to the 2 prongs only

fgrosa (changes requested):
Hi @apalasciano it looks good to me, thanks!
In the future we could foresee to have two template functions (one for 2 prongs and one for 3 prongs) with a flag for MC, instead of duplicating most of the code in two process functions, but we can do it in a second step. I just have a minor comment for this round, then we can merge it.

EventFiltering/PWGHF/HFFilterPrepareMLSamples.cxx
vkucera: Also if you adopt the `<object><attribute>` naming scheme, word capitalisation, and name variables after the quantities they represent, you will get more searchable, sortable and readable names, also grouped in the Hyperloop menu (which sorts alphabetically) : `massSbLeftMin`, `massSbLeftMax`, `massSbRightMin`, `massSbRightMax`.

fgrosa (approved):

PR #8399, opened 2024-11-11
[PWGCF] Add efficiency corrections

ilikmeta:


PWGCF/Flow/Tasks/FlowGFWPbPb.cxx
```diff
@@ -645,15 +664,74 @@ struct FlowGFWPbPb {
...
+        }
+      }
+    }
+  }
+  PROCESS_SWITCH(FlowGFWPbPb, processReco, "process reconstructed information", false);
+
+  // Filter for MCParticle
+  Filter particleFilter = (nabs(aod::mcparticle::eta) < cfgCutEta) && (aod::mcparticle::pt > cfgCutPtMin) && (aod::mcparticle::pt < cfgCutPtMax);
+  using myMcParticles = soa::Filtered<aod::McParticles>;
+  using myMcCollisionsFT0Cs = soa::Join<o2::aod::Collisions, o2::aod::EvSels, o2::aod::CentFT0Cs>;
```
victor-gonzalez, line 709: I would suggest being careful with the variable naming
This seems to not be an MC object, perhaps a reco MC object. The name is quite misleading
Also, for the above case `EvSels` table is not joined and here it is. Is that what is expected?

PWGCF/Flow/Tasks/FlowGFWPbPb.cxx
```diff
@@ -645,15 +664,74 @@ struct FlowGFWPbPb {
...
       FillFC(corrconfigs.at(l_ind), centrality, l_Random);
     }
 
   } // End of process
+  PROCESS_SWITCH(FlowGFWPbPb, processData, "Process analysis for Run 3 data", false);
+
+  // Filter the Reco tracks
+  Filter mytrackFilter = (nabs(aod::track::eta) < cfgCutEta) && (aod::track::pt > cfgCutPtMin) && (aod::track::pt < cfgCutPtMax) && (nabs(aod::track::dcaXY) < cfgCutDCAxy);
+  using myTracks = soa::Filtered<soa::Join<aod::Tracks, aod::TracksExtra, aod::TrackSelection, aod::TracksDCA, aod::McTrackLabels>>;
+  using myCollision = soa::Join<aod::Collisions, aod::CentFT0Cs>;
```
victor-gonzalez, line 682: Here `myCollision` is a table so keep the plural, if not the name is misleading
See also the below comment

PWGCF/Flow/Tasks/FlowGFWPbPb.cxx
```diff
@@ -645,15 +664,74 @@ struct FlowGFWPbPb {
...
+          registry.fill(HIST("hPtMCGen"), mcParticle.pt());
+          if (centrality > 0 && centrality <= 5) {
+            registry.fill(HIST("hPtMCGen05"), mcParticle.pt());
+          }
+          if (centrality >= 50 && centrality <= 60) {
+            registry.fill(HIST("hPtMCGen5060"), mcParticle.pt());
+          }
+        }
+      }
+    }
```
victor-gonzalez, line 731: Have you checked locally that this gives you what you expect?

victor-gonzalez (changes requested):
Please have a look at my comments

PWGCF/Flow/Tasks/FlowGFWPbPb.cxx
ilikmeta: I see what you mean. Here I wanted to get the centrality  for MC produced collisions, but as I see now, I am using again the FT0c centrality of reco events.  When I tried to use `o2::aod::HepMCHeavyIon = o2::aod::HepMCHeavyIons::iterator
` with the `centrality ` getter the histogram came empty.  My intension is to get the efficiency plotted as a function of centrality.

PWGCF/Flow/Tasks/FlowGFWPbPb.cxx
ilikmeta: You are right, I will change it to `myCollisions ` Thanks!

PWGCF/Flow/Tasks/FlowGFWPbPb.cxx
ilikmeta: To be more specific, from McCollisionLabels table I found this: `o2::aod::pid_tpc_tof_utils::McCollisionsFT0Cs = soa::Join<o2::aod::Collisions, o2::aod::EvSels, o2::aod::McCollisionLabels, o2::aod::CentFT0Cs>` and tried to use it here. That's why `EvSels` was included.

victor-gonzalez: Well, this is not form the `McCollisionsLabels` table, I guess
And the fact that you found it somewhere does not mean you would need it

My curiosity came because 
```
using myCollision = soa::Join<aod::Collisions, aod::CentFT0Cs>;
```
vs
```
using myMcCollisionsFT0Cs = soa::Join<o2::aod::Collisions, o2::aod::EvSels, o2::aod::CentFT0Cs>;
```
The naming mislead me and I inferred you wanted at generator level more information on the collision that what you required at reco level

ilikmeta: A concern that I have here is if I am getting the centrality for each produced MC event correctly by using:  `using myMcCollisionsFT0Cs = soa::Join<o2::aod::Collisions, o2::aod::EvSels, o2::aod::CentFT0Cs>;` Do you agree with this?

victor-gonzalez: Again, the naming is misleading
```
soa::Join<o2::aod::Collisions, o2::aod::EvSels, o2::aod::CentFT0Cs>;
```
is a table at the reconstruction level with information about reconstructed collisions, the different flags to apply at the level of event selection, and the centrality/multiplicity percentile based on the information of the FT0C

If the table is originated from  reconstructed data and the multiplicity/centrality available is properly calibrated the multiplicity/centrality percentile will be 'correct' in the measure that the event selection performed is similar to the one when the multiplicity/centrality was calibrated. This can be checked keeping track of the multiplicity/centrality percentile distribution after event/collision selection. It should be relatively flat

If the above table is originated from reconstructed MC, i.e. MC at the detector level, then, as the available multiplicity/centrality information has been calibrated on data, it cannot be expected that the mulitplicity/centrality percentile distribution after event/collision selection being flat. This is known

At some point in the future the proper MC calibration on the calibration on data will be available 
But not yet

PR #8410, opened 2024-11-12
[PWGCF]: Fix bug in event mixing

ariedel-cern:
Initialization of the column mixing policy has to be done in the init function. Otherwise, the mixing bins set in the hyperloop interface are not applied, and always the default values are used.


alibuild (approved):

victor-gonzalez (approved):

PR #8418, opened 2024-11-12
[PWGHF] code for v1 of D mesons using SP

prottayCMT:
code for analysis of directed flow of D mesons

### DIFF
--- a/PWGHF/D2H/Tasks/CMakeLists.txt
+++ b/PWGHF/D2H/Tasks/CMakeLists.txt
@@ -74,6 +74,11 @@ o2physics_add_dpl_workflow(task-flow-charm-hadrons
                     PUBLIC_LINK_LIBRARIES O2Physics::AnalysisCore O2Physics::EventFilteringUtils
                     COMPONENT_NAME Analysis)
 
+o2physics_add_dpl_workflow(task-directed-flow-charm-hadrons
+                    SOURCES taskDirectedFlowCharmHadrons.cxx
+                    PUBLIC_LINK_LIBRARIES O2Physics::AnalysisCore O2Physics::EventFilteringUtils
+                    COMPONENT_NAME Analysis)
+
 o2physics_add_dpl_workflow(task-lb
                     SOURCES taskLb.cxx
                     PUBLIC_LINK_LIBRARIES O2Physics::AnalysisCore
--- a/PWGHF/D2H/Tasks/CMakeLists.txt
+++ b/PWGHF/D2H/Tasks/CMakeLists.txt
@@ -58,7 +58,7 @@ o2physics_add_dpl_workflow(task-directed-flow-charm-hadrons
                     SOURCES taskDirectedFlowCharmHadrons.cxx
                     PUBLIC_LINK_LIBRARIES O2Physics::AnalysisCore O2Physics::EventFilteringUtils
                     COMPONENT_NAME Analysis)
-		  
+
 o2physics_add_dpl_workflow(task-dplus
                     SOURCES taskDplus.cxx
                     PUBLIC_LINK_LIBRARIES O2Physics::AnalysisCore
--- a/PWGHF/D2H/Tasks/CMakeLists.txt
+++ b/PWGHF/D2H/Tasks/CMakeLists.txt
@@ -49,7 +49,6 @@ o2physics_add_dpl_workflow(task-charm-reso-reduced
                     PUBLIC_LINK_LIBRARIES O2Physics::AnalysisCore
                     COMPONENT_NAME Analysis)
 
-
 o2physics_add_dpl_workflow(task-d0
                     SOURCES taskD0.cxx
                     PUBLIC_LINK_LIBRARIES O2Physics::AnalysisCore
@@ -79,7 +78,6 @@ o2physics_add_dpl_workflow(task-flow-charm-hadrons
                     SOURCES taskFlowCharmHadrons.cxx
                     PUBLIC_LINK_LIBRARIES O2Physics::AnalysisCore O2Physics::EventFilteringUtils
                     COMPONENT_NAME Analysis)
-
 
 o2physics_add_dpl_workflow(task-lb
                     SOURCES taskLb.cxx
--- a/PWGHF/D2H/Tasks/CMakeLists.txt
+++ b/PWGHF/D2H/Tasks/CMakeLists.txt
@@ -49,11 +49,17 @@ o2physics_add_dpl_workflow(task-charm-reso-reduced
                     PUBLIC_LINK_LIBRARIES O2Physics::AnalysisCore
                     COMPONENT_NAME Analysis)
 
+
 o2physics_add_dpl_workflow(task-d0
                     SOURCES taskD0.cxx
                     PUBLIC_LINK_LIBRARIES O2Physics::AnalysisCore
                     COMPONENT_NAME Analysis)
 
+o2physics_add_dpl_workflow(task-directed-flow-charm-hadrons
+                    SOURCES taskDirectedFlowCharmHadrons.cxx
+                    PUBLIC_LINK_LIBRARIES O2Physics::AnalysisCore O2Physics::EventFilteringUtils
+                    COMPONENT_NAME Analysis)
+		  
 o2physics_add_dpl_workflow(task-dplus
                     SOURCES taskDplus.cxx
                     PUBLIC_LINK_LIBRARIES O2Physics::AnalysisCore
@@ -74,10 +80,6 @@ o2physics_add_dpl_workflow(task-flow-charm-hadrons
                     PUBLIC_LINK_LIBRARIES O2Physics::AnalysisCore O2Physics::EventFilteringUtils
                     COMPONENT_NAME Analysis)
 
-o2physics_add_dpl_workflow(task-directed-flow-charm-hadrons
-                    SOURCES taskDirectedFlowCharmHadrons.cxx
-                    PUBLIC_LINK_LIBRARIES O2Physics::AnalysisCore O2Physics::EventFilteringUtils
-                    COMPONENT_NAME Analysis)
 
 o2physics_add_dpl_workflow(task-lb
                     SOURCES taskLb.cxx
--- a/PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
+++ b/PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
@@ -136,7 +136,8 @@ struct HfTaskDirectedFlowCharmHadrons {
   /// \param candidates are the selected candidates
   template <DecayChannel channel, typename T1, typename Trk>
   void runFlowAnalysis(CollsWithQvecs::iterator const& collision,
-                       T1 const& candidates, Trk const& /*tracks*/)
+                       T1 const& candidates,
+                       Trk const& /*tracks*/)
   {
     double cent = getCentrality(collision);
     if (cent < centralityMin || cent > centralityMax) {
@@ -205,7 +206,8 @@ struct HfTaskDirectedFlowCharmHadrons {
   }
   // Dplus with ML
   void processDplusMl(CollsWithQvecs::iterator const& collision,
-                      CandDplusDataWMl const& candidatesDplus, TracksWithExtra const& tracks)
+                      CandDplusDataWMl const& candidatesDplus,
+                      TracksWithExtra const& tracks)
   {
     runFlowAnalysis<DecayChannel::DplusToPiKPi>(collision, candidatesDplus, tracks);
   }
@@ -213,7 +215,8 @@ struct HfTaskDirectedFlowCharmHadrons {
 
   // Dplus with rectangular cuts
   void processDplusStd(CollsWithQvecs::iterator const& collision,
-                       CandDplusData const& candidatesDplus, TracksWithExtra const& tracks)
+                       CandDplusData const& candidatesDplus,
+                       TracksWithExtra const& tracks)
   {
     runFlowAnalysis<DecayChannel::DplusToPiKPi>(collision, candidatesDplus, tracks);
   }
--- a/PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
+++ b/PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
@@ -14,6 +14,27 @@
 ///
 /// \author Prottay Das, prottay.das@cern.ch
 
+#include <TH1F.h>
+#include <TDirectory.h>
+#include <THn.h>
+#include <TLorentzVector.h>
+#include <TMath.h>
+#include <TObjArray.h>
+#include <TFile.h>
+#include <TH2F.h>
+#include <TLorentzVector.h>
+#include <TPDGCode.h>
+#include <TDatabasePDG.h>
+#include <cmath>
+#include <array>
+#include <cstdlib>
+
+#include "TRandom3.h"
+#include "Math/Vector3D.h"
+#include "Math/Vector4D.h"
+#include "Math/GenVector/Boost.h"
+#include "TF1.h"
+
 #include "CCDB/BasicCCDBManager.h"
 #include "Framework/AnalysisTask.h"
 #include "Framework/HistogramRegistry.h"
--- a/PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
+++ b/PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
@@ -14,6 +14,8 @@
 ///
 /// \author Prottay Das, prottay.das@cern.ch
 
+#include <string>
+#include <vector>
 #include <TH1F.h>
 #include <TDirectory.h>
 #include <THn.h>
--- a/PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
+++ b/PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
@@ -1,0 +1,281 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+/// \file taskv1CharmHadrons.cxx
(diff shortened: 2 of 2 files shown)


PWGHF/D2H/Tasks/CMakeLists.txt
```diff
@@ -74,6 +74,11 @@ o2physics_add_dpl_workflow(task-flow-charm-hadrons
                     PUBLIC_LINK_LIBRARIES O2Physics::AnalysisCore O2Physics::EventFilteringUtils
                     COMPONENT_NAME Analysis)
 
+o2physics_add_dpl_workflow(task-directed-flow-charm-hadrons
+                    SOURCES taskDirectedFlowCharmHadrons.cxx
+                    PUBLIC_LINK_LIBRARIES O2Physics::AnalysisCore O2Physics::EventFilteringUtils
+                    COMPONENT_NAME Analysis)
```
vkucera, line 80: Respect alphabetic order.
Where do you use `EventFilteringUtils`?

PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,302 @@
...
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+/// \file taskv1CharmHadrons.cxx
+/// \brief Analysis task for charm hadron flow
```
vkucera, line 13: Fix the name.
Why do you not use the existing charm flow task?
Be more specific about the difference w.r.t. the other charm flow task.

PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,302 @@
...
+///
+/// \author Prottay Das, prottay.das@cern.ch
+
+#include <TH1F.h>
+#include <TDirectory.h>
+#include <THn.h>
+#include <TLorentzVector.h>
+#include <TMath.h>
+#include <TObjArray.h>
```
vkucera, line 24: Include only what you use.

PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,304 @@
...
+#include "PWGHF/Utils/utilsEvSelHf.h"
+
+using namespace o2;
+using namespace o2::aod;
+using namespace o2::framework;
+using namespace o2::framework::expressions;
+using namespace o2::hf_centrality;
+using namespace o2::hf_evsel;
+
+enum DecayChannel { DplusToPiKPi = 0 };
```
vkucera, line 61: Are you expecting to extend this to other species?

PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,304 @@
...
+
+struct HfTaskDirectedFlowCharmHadrons {
+  Configurable<int> centEstimator{"centEstimator", 2, "Centrality estimation (FT0A: 1, FT0C: 2, FT0M: 3, FV0A: 4)"};
+  Configurable<int> selectionFlag{"selectionFlag", 1, "Selection Flag for hadron (e.g. 1 for skimming, 3 for topo. and kine., 7 for PID)"};
+  Configurable<float> centralityMin{"centralityMin", 0., "Minimum centrality accepted in SP computation"};
+  Configurable<float> centralityMax{"centralityMax", 100., "Maximum centrality accepted in SP computation"};
+  Configurable<bool> storeMl{"storeMl", false, "Flag to store ML scores"};
+  Configurable<std::string> ccdbUrl{"ccdbUrl", "http://alice-ccdb.cern.ch", "url of the ccdb repository"};
+  Configurable<std::vector<int>> classMl{"classMl", {0, 2}, "Indexes of BDT scores to be stored. Two indexes max."};
```
vkucera, line 70:
```suggestion
  Configurable<std::vector<int>> classMl{"classMl", {0, 2}, "Indices of BDT scores to be stored. Two indexes max."};
```

PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,304 @@
...
+  }; // end init
+
+  /// Fill THnSparse
+  /// \param mass is the invariant mass of the candidate
+  /// \param pt is the transverse momentum of the candidate
+  /// \param cent is the centrality of the collision
+  /// \param cosNPhi is the cosine of the n*phi angle
+  /// \param cosDeltaPhi is the cosine of the n*(phi - evtPl) angle
+  /// \param sp is the scalar product
+  /// \param outputMl are the ML scores
```
vkucera, line 133: Fix

PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,304 @@
...
+      registry.fill(HIST("hpuxQxtvscentpteta"), mass, cent, pt, eta, uxQxt, sign);
+      registry.fill(HIST("hpuyQytvscentpteta"), mass, cent, pt, eta, uyQyt, sign);
+    }
+  }
+
+  /// Get the centrality
+  /// \param collision is the collision with the centrality information
+  double getCentrality(CollsWithQvecs::iterator const& collision)
+  {
+    double cent = -999.;
```
vkucera, line 162: Can you not reuse the existing utility function with `concepts`?

PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,304 @@
...
+  }
+
+  /// Check if the collision is selected
+  /// \param collision is the collision with the Q vector information
+  /// \param bc is the bunch crossing with timestamp information
+  /// \return true if the collision is selected, false otherwise
+  template <o2::hf_centrality::CentralityEstimator centEstimator>
+  bool isCollSelected(CollsWithQvecs::iterator const& collision,
+                      aod::BCsWithTimestamps const&)
+  {
```
vkucera, line 191: This method is not used anywhere.

PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,304 @@
...
+    double qxAQVec = 0.0;
+    double qyAQVec = 0.0;
+    double qxCQVec = 0.0;
+    double qyCQVec = 0.0;
+    qxAQVec = collision.qxZDCA();
+    qyAQVec = collision.qyZDCA();
+    qxCQVec = collision.qxZDCC();
+    qyCQVec = collision.qyZDCC();
+    return {qxAQVec, qyAQVec, qxCQVec, qyCQVec};
+  }
```
vkucera, line 213: Unnecessary since used only once

PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,304 @@
...
+
+    if (!collision.triggerevent()) { // for selecting only callibrated events
+      return;
+    }
+
+    std::vector<double> qVecs = getQvec(collision);
+    double qxAQVec = qVecs[0];
+    double qyAQVec = qVecs[1];
+    double qxCQVec = qVecs[2];
+    double qyCQVec = qVecs[3];
```
vkucera, line 235: Can be assigned directly from `collision`.

PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,304 @@
...
+    registry.fill(HIST("hpQxtQxpvscent"), cent, QxtQxp);
+    registry.fill(HIST("hpQytQypvscent"), cent, QytQyp);
+    registry.fill(HIST("hpQxpQytvscent"), cent, QxpQyt);
+    registry.fill(HIST("hpQxtQypvscent"), cent, QxtQyp);
+
+    for (const auto& candidate : candidates) {
+      double massCand = 0.;
+      std::vector<double> outputMl = {-999., -999.};
+      if constexpr (std::is_same_v<T1, CandDplusData> || std::is_same_v<T1, CandDplusDataWMl>) {
+        massCand = hfHelper.invMassDplusToPiKPi(candidate);
```
vkucera, line 257: Do you need this because you want to support multiple species?

PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,304 @@
...
+  // Dplus with ML
+  void processDplusMl(CollsWithQvecs::iterator const& collision,
+                      CandDplusDataWMl const& candidatesDplus, TracksWithExtra const& tracks)
+  {
+    runFlowAnalysis<DecayChannel::DplusToPiKPi>(collision, candidatesDplus, tracks);
+  }
+  PROCESS_SWITCH(HfTaskDirectedFlowCharmHadrons, processDplusMl, "Process Dplus candidates with ML", false);
+
+  // Dplus with rectangular cuts
+  void processDplus(CollsWithQvecs::iterator const& collision,
```
vkucera, line 292:
```suggestion
  void processDplusStd(CollsWithQvecs::iterator const& collision,
```

PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,304 @@
...
+  EventPlaneHelper epHelper;
+  HfEventSelection hfEvSel; // event selection and monitoring
+  o2::framework::Service<o2::ccdb::BasicCCDBManager> ccdb;
+
+  HistogramRegistry registry{"registry", {}, OutputObjHandlingPolicy::AnalysisObject};
+
+  void init(InitContext&)
+  {
+    const AxisSpec thnAxisInvMass{thnConfigAxisInvMass, "Inv. mass (GeV/#it{c}^{2})"};
+    const AxisSpec thnAxisPt{thnConfigAxisPt, "#it{p}_{T} (GeV/#it{c})"};
```
vkucera, line 99: Add check for process switches.

PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,304 @@
...
+  void processDplusMl(CollsWithQvecs::iterator const& collision,
+                      CandDplusDataWMl const& candidatesDplus, TracksWithExtra const& tracks)
+  {
+    runFlowAnalysis<DecayChannel::DplusToPiKPi>(collision, candidatesDplus, tracks);
+  }
+  PROCESS_SWITCH(HfTaskDirectedFlowCharmHadrons, processDplusMl, "Process Dplus candidates with ML", false);
+
+  // Dplus with rectangular cuts
+  void processDplus(CollsWithQvecs::iterator const& collision,
+                    CandDplusData const& candidatesDplus, TracksWithExtra const& tracks)
```
vkucera, line 293: Put one argument per line.

vkucera (changes requested):

PWGHF/D2H/Tasks/CMakeLists.txt
prottayalice: [Reply]: I updated the CMakeLists.txt....Is it ok now?
              I used PWGHF/Utils/utilsEvSelHf.h which i guess needs EventFilteringUtils. It is same as used in https://github.com/AliceO2Group/O2Physics/blob/master/PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx.

PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
prottayalice: [Reply]: Name is fixed now.
             This code is for v1 of D mesons where we need to use spectator plane. The earlier code was for v2 which uses the event plane. The physics and techniques are very much different so I prefer to use my own task and nt modify the existing one.

PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
prottayalice: [Reply]: Done.

prottayalice: [Reply]: Done as suggested.

PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
prottayalice: [Reply]: Currently I plan to only perform the measurement for D mesons but yes in future it can be extended (by me or by other analyzer) to other particle species as well.

PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
prottayalice: [Reply]: Indexes -> Indices

PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
prottayalice: [Reply]: Fixed in the updated code.

PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
prottayalice: [Reply]: At present would like to keep it as it is.

PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
prottayalice: [Reply]: You are correct. I have removed it now from the updated code.

PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
prottayalice: [Reply]: Removed the function.

PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
prottayalice: [Reply]: Done as suggested.

PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
prottayalice: [Reply]: Yes for that reason I would like to keep it.

PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
prottayalice: [Reply]: processDplus -> processDplusStd

PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
prottayalice: [Reply]: Added few arguments.

PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,249 @@
...
+  void fillThn(double& mass,
+               double& pt,
+               double& eta,
+               double& cent,
+               double& uxQxp,
+               double& uyQyp,
+               double& uxQxt,
+               double& uyQyt,
+               std::vector<double>& outputMl,
+               double& sign)
```
vkucera, line 125: You won't save any memory by passing simple numbers by reference. In fact, this is bad because you miss `const` so the function can modify the variables you passed. The reference here only makes sense for the vector, in which case you should add `const` for the same reason.

vkucera (changes requested):

PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
vkucera: I don't see a change.

PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
prottayalice: [Reply]: I have now removed the function.

PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
prottayalice: [Reply]: I have added few comments in the code such as "// correlations in the denominators for SP calculation" in Line 160 and so on....What specific comments/arguments you want in each line of code?

vkucera: I didn't mention comments. I meant the arguments of the process function. It's easier to read the table subscriptions when they are broken into one argument per line.

prottayalice: Is it okay now?

PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
```diff
@@ -0,0 +1,230 @@
...
+  PROCESS_SWITCH(HfTaskDirectedFlowCharmHadrons, processDplusMl, "Process Dplus candidates with ML", false);
+
+  // Dplus with rectangular cuts
+  void processDplusStd(CollsWithQvecs::iterator const& collision,
+                       CandDplusData const& candidatesDplus,
+                       TracksWithExtra const& tracks)
+  {
+    runFlowAnalysis<DecayChannel::DplusToPiKPi>(collision, candidatesDplus, tracks);
+  }
+  PROCESS_SWITCH(HfTaskDirectedFlowCharmHadrons, processDplusStd, "Process Dplus candidates", true);
```
vkucera, line 223:
```suggestion
  PROCESS_SWITCH(HfTaskDirectedFlowCharmHadrons, processDplusStd, "Process Dplus candidates with rectangular cuts", true);
```

prottayalice: Done

vkucera (approved):

PR #8419, opened 2024-11-12
[Common] [PID] add ITS PID

njacazio:


njacazio:
@fmazzasc I think this is ready

PR #8420, opened 2024-11-13
[Common] Add unified converter

njacazio:


jgrosseo:
@njacazio Could we have this in draft for now?

We were discussing this with @ktf earlier and decided it is not possible without a special treatment in O2. Because how do you know that a certain table is actually needed? If I would add this converter now, it will spawn all converters based on the current data, regardless if any downstream device is actually needing the converted table. So we would also read all data... What do you think?

njacazio:
Hi @jgrosseo many thanks for the comment! I was testing it yesterday and decided to try it, I did not know that this was already happening :)
I understand your concern, indeed.
Are you worried about the memory impact of having all the converters or are you referring to the filling of non-needed tables?

jgrosseo:
Mostly about reading not-needed data, but the rest is relevant as well.

One needs access to both the metadata tables (as you did) and to the subscribed tables in the topology (which is only available in init). Giulio was looking into how to generalize it such that we can have a general converter.

njacazio:
I see, yes.
Currently what we can do is move the process function behind a switch and have the process function enabled only if the table is actually needed by the workflow. This solves the reading and the filling of the tables but not the memory footprint of the device.

jgrosseo:
Unfortunately, the problem remains. Even with a process switch the statement:
`Produces<aod::BCs_001> bc_001;`
stays active and this device will be used as source of the data, *instead* of reading it from the AO2D...

jgrosseo:
Yes, you are right! It just means we have to separate all cases, e.g. if we go from v0 -> v1 -> v2.
The downside is only some memory waste per converter which is spawned but where the table is actually not needed. Do you know why much this is per device?

There seems to be a bug in your lines for V0, it points to the same converter for v0->v1 and v1->v2.

Also @ktf was looking into conditional Produces statements. Maybe you can check with him, if this is already there. If that existed one could put everything in one device...

njacazio:
This should be constant and depend on the number of devices, I don't know the overhead tbh.
But we can do the following: 1) perform the test to prove that it works 2) measure the overhead and 3) try to put it in production (provided that it works) in parallel to the others and 4) move there once things are set

jgrosseo:
Yes, we can do 1) and 2) before merging.
One would need to find a sufficient number of tests with different type of data.
It would be also good to measure the memory consumption. Possibly it can be done by producing the performance metrics and visualizing them here: https://alimonitor.cern.ch/hyperloop/performance-graphs

victor-gonzalez:
If I may comment, perhaps one of the best use cases would be converted Run2 data (I'm returning back to them right now :))
Thanks for moving this forward!

njacazio:
This does not seem as simple as I imagined in the beginning, it needs a bit more testing :(

njacazio:
Hi @jgrosseo I managed to have a working example, with a full workflow, if you agree we can merge this one and see the rest with HL.
The only issue I observed is that it seems that this task has to be the last one in the workflow.

ddobrigk:
After the discussion at today's WP4+14, I will approve this PR in the afternoon in case there are no further comments before that. Thanks a lot @njacazio !

jgrosseo:
> Hi @jgrosseo I managed to have a working example, with a full workflow, if you agree we can merge this one and see the rest with HL. The only issue I observed is that it seems that this task has to be the last one in the workflow.

I am sorry, I was on leave today.

At present we cannot use this on HY if it has to be at the end of the workflow.

Also we should measure the additional memory. Has this been done?

I see no need to merge it for measuring the memory. The metrics can be enabled and we can use the page I posted above to visualize. However @ddobrigk if you agreed today to merge, I don't want to object.

Common/TableProducer/Converters/converter.cxx
```diff
@@ -0,0 +1,615 @@
...
+  }
+  PROCESS_SWITCH(MftTracksConverter, processConverter, "Process converter (autoset)", true);
+};
+
+/// Spawn the extended table for MFTTracks001 to avoid the call to the internal spawner and a consequent circular dependency
+struct MFTTracksSpawner {
+  Spawns<aod::MFTTracks_001> mftTracks_001;
+};
+
+struct MultsExtraConverter {
```
jgrosseo, line 478: This doesn't seem to produce something in the central data model? Where is this converter from?

Common/TableProducer/Converters/converter.cxx
```diff
@@ -0,0 +1,615 @@
...
+        LOG(info) << "  - AOD converter: Table " << t << " needs no converter";
+      } else if (t == "O2zdc_000" || t == "O2zdc") {
+        LOG(info) << "  + AOD converter: Table " << t << " found, adding converter task";
+        workflow.push_back(adaptAnalysisTask<zdcConverter>(cfgc));
+      } else {
+        LOG(debug) << "AOD converter: Versioned table not covered " << t;
+      }
+    }
+  } else {
+    LOG(warning) << "AOD converter: No tables found in the workflow";
```
jgrosseo, line 612:
```suggestion
    LOG(warning) << "AOD converter: No tables found in the meta data";
```

Common/TableProducer/Converters/converter.cxx
```diff
@@ -0,0 +1,615 @@
...
+// Converts bc_000 into bc_001
+struct bcConverter {
+  Produces<aod::BCs_001> bc_001;
+  void init(o2::framework::InitContext& initContext)
+  {
+    doprocessConverter.value = isTableRequiredInWorkflow(initContext, "BCs_001");
+    if (doprocessConverter) {
+      LOG(info) << "Enabling converter for table BCs_001";
+    }
+  }
```
jgrosseo, line 43: This part is very similar in all converters. Can it be put written only once with the argument of the table name?

Common/TableProducer/Converters/converter.cxx
```diff
@@ -0,0 +1,615 @@
...
+
+WorkflowSpec defineDataProcessing(ConfigContext const& cfgc)
+{
+  auto workflow = WorkflowSpec{};
+  if (0) {
+    for (auto c : cfgc.options().specs()) {
+      LOG(info) << "Option: " << c.name << " " << static_cast<int>(c.type);
+      // LOG(info) << "Option: " << c.name << " " << cfgc.options().get<c.type>(c.name);
+    }
+  }
```
jgrosseo, line 531: Remove this...

Common/TableProducer/Converters/converter.cxx
```diff
@@ -0,0 +1,615 @@
...
+  if (cfgc.options().hasOption("aod-metadata-tables")) {
+    const std::vector<std::string> tables = cfgc.options().get<std::vector<std::string>>("aod-metadata-tables");
+    for (auto t : tables) {
+      // if (t.find("_0") == std::string::npos) {
+      //   continue;
+      // }
```
jgrosseo, line 537: ... and this

jgrosseo (commented):
Thanks for your work @njacazio 

I went through and have a few comments. 

Let me know. :-)

njacazio:
Hi, I resumed the project, I tested the memory footprint and it is comparable (in my test actually a bit less). This covers most of the converters from Run 3 and Run 2. I would start merging this so that we can start the cleanup of the converters and the dev on the HL side.

njacazio:
> For me it is fine to merge. However, does it still need to be at the end of the command line? This will not be available on Hyperloop then.

Hi, yes this needs to be the last workflow as confirmed by Giulio. But isn't the workflow ordering something that can be handled at the HL time?

njacazio:
Hi, sounds good. I don't know how it works currently. If there is an order e.g. alphabetical one we can tune the name accordingly such that it's the latest workflow.

PR #8426, opened 2024-11-13
[PWGEM] add PID scenario: TPC only selection hadron rej, TOF required"

feisenhu:
Adding a PID scenario which doesn't use TOF nSigma for electron selection while applying TPC electron selection.
Used for DQ and EM comparison.


alibuild (approved):

PR #8427, opened 2024-11-13
[PWGHF] filling DstarTrack hist with delta Mass

cterrevo:


PWGHF/D2H/TableProducer/candidateCreatorCharmResoReduced.cxx
```diff
@@ -312,7 +312,7 @@ struct HfCandidateCreatorCharmResoReduced {
 
         if constexpr (channel == DecayChannel::DstarTrack) {
           invMassReso = RecoDecay::m(std::array{pVecD, pVecV0Tr}, std::array{massDstar, massProton});
-          registry.fill(HIST("hMassDstarTrack"), invMassReso, ptReso);
+          registry.fill(HIST("hMassDstarTrack"), invMassReso - invMassD, ptReso);
```
fgrosa, line 315: if you compute the delta Mass, the invariant mass of the resonance `invMassReso` should be computed from the daughter tracks and it should not be the PDG one (otherwise the purpose of calculating the delta mass is not fulfilled, since we do not cancel any momentum resolution)

fgrosa (changes requested):

PWGHF/D2H/TableProducer/candidateCreatorCharmResoReduced.cxx
```diff
@@ -311,8 +315,12 @@ struct HfCandidateCreatorCharmResoReduced {
         float ptReso = RecoDecay::pt(RecoDecay::sumOfVec(pVecV0Tr, pVecD));
 
         if constexpr (channel == DecayChannel::DstarTrack) {
-          invMassReso = RecoDecay::m(std::array{pVecD, pVecV0Tr}, std::array{massDstar, massProton});
-          registry.fill(HIST("hMassDstarTrack"), invMassReso, ptReso);
+          if (candD.dType() > 0) {
```
vkucera, line 318: A suggestion (not major): You are unnecessarily calling the `pVector` methods twice. You can instead save the array of all 4 momenta in a single variable and reuse it for bot calls of `m`, while just swapping the masses.

PWGHF/D2H/TableProducer/candidateCreatorCharmResoReduced.cxx
```diff
@@ -311,8 +315,12 @@ struct HfCandidateCreatorCharmResoReduced {
...
+            invMassReso = RecoDecay::m(std::array{candD.pVectorProng0(), candD.pVectorProng1(), candD.pVectorProng2(), pVecV0Tr}, std::array{massPion, massKaon, massPion, massProton});
+          } else {
+            invMassReso = RecoDecay::m(std::array{candD.pVectorProng1(), candD.pVectorProng0(), candD.pVectorProng2(), pVecV0Tr}, std::array{massPion, massKaon, massPion, massProton});
+          }
```
vkucera, line 322: A suggestion (not major): You are unnecessarily calling the 3 `pVector` methods twice. You can instead save the array of all 4 momenta in a single variable and reuse it for both calls of `m`, while just swapping the masses. This is not an issue in this case because we store the xyz components but for example in case of tracks, it would cause unnecessary calculations.

cterrevo: Thanks Vit for the suggestion! I would implement it in the next PR, if for you it is fine!

fcatalan92:
@ktf @singiamtel It looks like the `O2Physics/o2` build is failing due to the machine having the disk full, could you check?

fcatalan92:
> @ktf @singiamtel It looks like the `O2Physics/o2` build is failing due to the machine having the disk full, could you check?

I noticed only now that the issue is already followed up in https://github.com/alisw/ali-bot/pull/1423, thanks a lot!

PR #8434, opened 2024-11-13
[PWGHF] add converter for table with Dstar indices

fgrosa:
This PR adds a converter to allow the analyses of HF derived data for D* after the fix in https://github.com/AliceO2Group/O2Physics/pull/8322


vkucera:
Thanks @fgrosa !

PR #8438, opened 2024-11-13
[Common,PWGJE] Add getParticleOrigin and modified init problem

hanseopark:
To use jet-flavour definition from jet group

Modification of problem about initialize for track and particle

### DIFF
--- a/Common/Core/RecoDecay.h
+++ b/Common/Core/RecoDecay.h
@@ -1035,6 +1035,119 @@ struct RecoDecay {
     }
     return OriginType::None;
   }
+
+  /// based on getCharmHardronOrigin in order to extend general particle
+  /// Finding the origin (from charm hadronisation or beauty-hadron decay) of paritcle (b, c and others)
+  /// \param particlesMC  table with MC particles
+  /// \param particle  MC particle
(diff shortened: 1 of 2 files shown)


hanseopark:
@hahassan7 @vkucera @fgrosa
A review would be appreciated

nzardosh:
Hi Hanseo,

Why is this better to be added here rather than to recodecay?

Thanks,
Nima

hanseopark:
Hi @nzardosh 
If we use searchUpToQuark==false, recodecay function have to return prompt and non-prompt only, not none
I should have implemented for tagging

nzardosh:
Hi Hanseo, but my quesion is cant we modify reco decay instead?

hanseopark:
Hi Nima @nzardosh
This function is currently under development because quark level doesn't work well, and `getCharmHadronOrigin` is a function developed specifically for Charm hadrons (at hardron level). Modification the recodecay function might solve it for now, but I'm thinking that updating the quark level part might change the code a lot in the process.

hanseopark:
@vkucera @nzardosh 
I move to the function of `getParticleOrigin`() on `RecoDecay.h`

PR #8441, opened 2024-11-13
[PWGCF] Add PID with ITS to STS

njacazio:


victor-gonzalez (approved):
I guess you have considered potential backward compatibility with already existing derived data

njacazio:
Hi Victor, yes, with split tables there's no need to have changes in the analysis code. Many thanks!

PR #8449, opened 2024-11-14
[PWGUD] update DGCandProducer to fill proper globalBC number into UDMcCollisions table

pbuehler:


rolavick (dismissed):

vkucera:
@pbuehler @rolavick PWG names with hyphens are not recognised as valid tags by the title prefix checker which makes the bot add a PWG prefix in the valid format which effectively leads to the PWG name mentioned twice in the PR title. If you avoid the hyphens, you will avoid the duplication.

rolavick (approved):

rolavick (approved):

rolavick (dismissed):

rolavick (approved):

PR #8452, opened 2024-11-14
[PWGHF] include number of contributors histos in correlatorDMesonPair.cxx

atavirag:


apalasciano (dismissed):

PWGHF/HFC/TableProducer/correlatorDMesonPairs.cxx
```diff
@@ -750,7 +758,12 @@ struct HfCorrelatorDMesonPairs {
         registry.fill(HIST("hStatusSinglePartMcGen"), 4);
       }
 
-      registry.fill(HIST("hPtVsYMcGen"), particle1.pt(), particle1.y());
+      for (auto const& collision : collisions) {
```
fgrosa, line 761: with this implementation, in case of multiple reco collisions corresponding to the same MC collision (split vertex) you would fill the histogram for the same particle multiple times (biasing the efficiency to lower values), while you never fill it in case of no reco collision (while it would be good to monitor the number of generated signals in events that are not reconstructed --> signal loss).

The convention in case of split vertices is to pick the number of PV contributors of the reco collision with the largest number of PV contributors. You can check what is done here for instance: https://github.com/AliceO2Group/O2Physics/blob/a861eb45508e6cf97659fc0c4500b1264e1bc330/PWGHF/D2H/Tasks/taskCharmPolarisation.cxx#L1734-L1743

fgrosa (changes requested):

PWGHF/HFC/TableProducer/correlatorDMesonPairs.cxx
atavirag: I just changed it, thanks!

vkucera (approved):
Thanks @atavirag , looks good to me.

fgrosa (approved):

atavirag:
Hi, the merge does not seem to go through due to an error in the o2 build that does not seem related to the commit. Is there anything I can do?

vkucera:
Hi @atavirag , a PR that is supposed to fix the space issue was merged this morning. Feel free to complain in the support channel if the build does not finish successfully today.

atavirag:
All checks have passed now, but it still needs someone with write access to this repo to merge the PR. Could that be done? Thanks!

PR #8475, opened 2024-11-15
[PWGHF] Occupancy studies in Pb-Pb collisions for D mesons

Marcellocosti:
This PR adds to `enum EventRejection` of `utilsEvSelHf.h` the following event selection flags:

-  `NoCollInTimeRangeNarrow`
-  `NoCollInTimeRangeStrict`
-  `NoCollInTimeRangeVzDependent`
-  `NoCollInRofStrict`
-  `NoCollInRofStandard`

according to the linked [PR](https://github.com/AliceO2Group/O2Physics/pull/8124). 
Moreover, additional modifications are made to the task `taskFlowCharmHadrons.cxx` to perform occupancy studies and introduce the FT0C-based occupancy estimation.

### DIFF
--- a/PWGHF/D2H/TableProducer/dataCreatorCharmResoReduced.cxx
+++ b/PWGHF/D2H/TableProducer/dataCreatorCharmResoReduced.cxx
@@ -992,7 +992,7 @@ struct HfDataCreatorCharmResoReduced {
     }
     registry.fill(HIST("hEvents"), 1 + Event::DV0Selected);
     float centrality = -1.f;
-    uint32_t hfRejMap = hfEvSel.getHfCollisionRejectionMask<true, o2::hf_centrality::CentralityEstimator::None, aod::BCsWithTimestamps>(collision, centrality, ccdb, registry);
+    uint16_t hfRejMap = hfEvSel.getHfCollisionRejectionMask<true, o2::hf_centrality::CentralityEstimator::None, aod::BCsWithTimestamps>(collision, centrality, ccdb, registry);
     // fill collision table if it contains a DPi pair a minima
     hfReducedCollision(collision.posX(), collision.posY(), collision.posZ(), collision.numContrib(), hfRejMap, bz);
   } // run data creation
--- a/PWGHF/D2H/TableProducer/dataCreatorCharmResoReduced.cxx
+++ b/PWGHF/D2H/TableProducer/dataCreatorCharmResoReduced.cxx
@@ -992,7 +992,7 @@ struct HfDataCreatorCharmResoReduced {
     }
     registry.fill(HIST("hEvents"), 1 + Event::DV0Selected);
     float centrality = -1.f;
-    uint16_t hfRejMap = hfEvSel.getHfCollisionRejectionMask<true, o2::hf_centrality::CentralityEstimator::None, aod::BCsWithTimestamps>(collision, centrality, ccdb, registry);
+    uint32_t hfRejMap = hfEvSel.getHfCollisionRejectionMask<true, o2::hf_centrality::CentralityEstimator::None, aod::BCsWithTimestamps>(collision, centrality, ccdb, registry);
     // fill collision table if it contains a DPi pair a minima
     hfReducedCollision(collision.posX(), collision.posY(), collision.posZ(), collision.numContrib(), hfRejMap, bz);
   } // run data creation
--- a/PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
+++ b/PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
@@ -15,6 +15,9 @@
 /// \author S. Politanò, INFN Torino, Italy
 /// \author Wu Chuntai, CUG, China
 
+#include <string>
+#include <vector>
+
 #include "CCDB/BasicCCDBManager.h"
 #include "Framework/AnalysisTask.h"
 #include "Framework/HistogramRegistry.h"
--- a/PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
+++ b/PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
@@ -131,12 +131,12 @@ struct HfTaskFlowCharmHadrons {
     const AxisSpec thnAxisMlTwo{thnConfigAxisMlTwo, "FD score"};
     const AxisSpec thnAxisOccupancyITS{thnConfigAxisOccupancyITS, "OccupancyITS"};
     const AxisSpec thnAxisOccupancyFT0C{thnConfigAxisOccupancyFT0C, "OccupancyFT0C"};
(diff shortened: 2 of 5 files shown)


PWGHF/Utils/utilsEvSelHf.h
```diff
@@ -142,9 +158,9 @@ struct HfEventSelection : o2::framework::ConfigurableGroup {
   /// \param registry reference to the histogram registry needed for zorro
   /// \return bitmask with the event selection criteria not satisfied by the collision
   template <bool useEvSel, o2::hf_centrality::CentralityEstimator centEstimator, typename BCs, typename Coll>
-  uint16_t getHfCollisionRejectionMask(const Coll& collision, float& centrality, o2::framework::Service<o2::ccdb::BasicCCDBManager> const& ccdb, o2::framework::HistogramRegistry& registry)
+  uint32_t getHfCollisionRejectionMask(const Coll& collision, float& centrality, o2::framework::Service<o2::ccdb::BasicCCDBManager> const& ccdb, o2::framework::HistogramRegistry& registry)
   {
-    uint16_t rejectionMask{0}; // 16 bits, in case new ev. selections will be added
+    uint32_t rejectionMask{0}; // 16 bits, in case new ev. selections will be added
```
fgrosa, line 163: If you change the type of the mask, you should modify it everywhere it's used. For instance, this is also used in the https://github.com/AliceO2Group/O2Physics/blob/master/PWGHF/TableProducer/trackIndexSkimCreator.cxx task and stored in this table https://github.com/AliceO2Group/O2Physics/blob/a2a96905974a13d9c77cb15b0f3f2eeac487e3f2/PWGHF/DataModel/CandidateReconstructionTables.h#L194 (which also has to be modified, otherwise it gets truncated). 
You should carefully check everywhere it is used and make sure that it is updated to 32 bits everywhere

PWGHF/D2H/TableProducer/dataCreatorCharmResoReduced.cxx
```diff
@@ -992,7 +992,7 @@ struct HfDataCreatorCharmResoReduced {
     }
     registry.fill(HIST("hEvents"), 1 + Event::DV0Selected);
     float centrality = -1.f;
-    uint16_t hfRejMap = hfEvSel.getHfCollisionRejectionMask<true, o2::hf_centrality::CentralityEstimator::None, aod::BCsWithTimestamps>(collision, centrality, ccdb, registry);
+    uint32_t hfRejMap = hfEvSel.getHfCollisionRejectionMask<true, o2::hf_centrality::CentralityEstimator::None, aod::BCsWithTimestamps>(collision, centrality, ccdb, registry);
```
fgrosa, line 995: This variable is stored in the tables produced by these data producers (which are written on disk): https://github.com/AliceO2Group/O2Physics/blob/a2a96905974a13d9c77cb15b0f3f2eeac487e3f2/PWGHF/D2H/DataModel/ReducedDataModel.h#L44 
This means that the definition of the column should also change and a converter should be implemented in order to be able to analyse the already produced derived data (otherwise this change will break the analysis of already produced derived data of B mesons and Ds resonances)

fgrosa (changes requested):
Hi @Marcellocosti, thanks a lot for the useful implementation! I have however a major comment; when changing the the bitmap with the rejection response from `uint16_t` to `uint32_t` you have to make sure that the correct type is modified everywhere, including in the table definitions. Moreover, if it is used in derived data, which are already written on disk, it is necessary to provide converters that allow the analyses of already produced derived data (see below)

Marcellocosti:
Hi @fgrosa, thanks for the crucial input. I will reshape the number of selections in order to fit the variable type `uint16_t`.

PWGHF/Utils/utilsEvSelHf.h
vkucera: Why is making the type bigger a problem? When reading data with `uint16_t`, are they not implicitly converted to `uint32_t`?

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
```diff
@@ -360,6 +449,13 @@ struct HfTaskFlowCharmHadrons {
     if (cent < centralityMin || cent > centralityMax) {
       return;
     }
+    float occupancy = 0.;
+    uint16_t hfevflag;
+    if (storeOccupancy) {
+      occupancy = getOccupancy(collision);
+      registry.fill(HIST("trackOccVsFT0COcc"), collision.trackOccupancyInTimeRange(), collision.ft0cOccupancyInTimeRange());
+      hfevflag = static_cast<int>(hfEvSel.getHfCollisionRejectionMask<true, o2::hf_centrality::CentralityEstimator::None, aod::BCsWithTimestamps>(collision, cent, ccdb, registry));
```
vkucera, line 457: `getHfCollisionRejectionMask` returns `uint16_t`. Why would you cast it to `int` and back?

mfaggin:
Thanks a lot @Marcellocosti for the important implementation!
I have just one comment: according to [this](https://indico.cern.ch/event/1479094/contributions/6230143/attachments/2967245/5220832/occupancy_inROF_pileup_AOT_Nov14_IA.pdf) presentation, and also as written by you in the original comment to this PR, there are 5 event selections that can be added to our HF code, but only 3 of them are implemented in `PWGHF/Utils/utilsEvSelHf.h`.
In particular, I see that `kNoCollInRofStrict` and `kNoCollInTimeRangeStrict` are not implemented. Is there a specific reason for it? Otherwise, can you please add them as well, so that in case of need they are available? 
Thanks

Marcellocosti:
Hi @mfaggin, inserting all the flags would require an extension to the `uint32_t` type for the variable type associated to the event selection bitmap. @fgrosa was mentioning above that this could cause issues with the processing of already produced tables, thus after an offline iteration we decided to select only a subsample of flags.

mfaggin:
> Hi @mfaggin, inserting all the flags would require an extension to the `uint32_t` type for the variable type associated to the event selection bitmap. @fgrosa was mentioning above that this could cause issues with the processing of already produced tables, thus after an offline iteration we decided to select only a subsample of flags.

Thanks @Marcellocosti , I did not understand this detail from the above discussion.
Fine for me then, let's go in this way.
Please fix the formatting issues before we can merge.

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
Marcellocosti: It was indeed not necessary. Thanks for noticing!

mfaggin (approved):

vkucera:
@Marcellocosti Can you please also add the missing headers?

fgrosa (approved):

fgrosa:
@Marcellocosti @vkucera I approve it for the time being, but let's figure out if the conversion from `uint16_t` to `uint32_t` it's not an issue for tables already written, in case we want to further extend it.

vkucera:
@singiamtel What kind of approval are the macOS builds waiting for?

PR #8479, opened 2024-11-15
[Common] Split PIDResponse

njacazio:


njacazio:
tagging also @amaringarcia

njacazio:
@fgrosa @victor-gonzalez (I tag only you too, redirect in case needed, apologies) this also has changes in the PWG code but is needed to decouple the development and streamline the code. Could you check that it's fine for you as well?

PR #8482, opened 2024-11-15
[PWGLF] add EP angle to stra der data

ddobrigk:
@ChiaraDeMartin95 this tiny change will allow for the closure tests to be done for cascade v2 straight from the strangeness derived data (it is a simple passthrough of the event plane angle from the generator, added to the main data model for this purpose).


ddobrigk:
Moving to ready for review - now added the converter I forgot to add before :-P

skundu692 (approved):

PR #8485, opened 2024-11-15
[PWGHF] Add self-contained derived-data format for B+

vkucera:
FYI: @nzardosh @qgp @DelloStritto

### DIFF
--- a/PWGHF/D2H/Tasks/taskBplus.cxx
+++ b/PWGHF/D2H/Tasks/taskBplus.cxx
@@ -263,7 +263,7 @@ struct HfTaskBplus {
 
         float ptProngs[2], yProngs[2], etaProngs[2];
         int counter = 0;
-        for (const auto& daught : particle.daughters_as<aod::McParticles>()) {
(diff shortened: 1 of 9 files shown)


vkucera:
> Hi @vkucera, did you consider to add a joinable table with the D0 BDT output scores? I think it would be very beneficial to have them, since you will anyway have to apply the BDT models used in the trigger, and from my experience in the B0 cross section measurement, they can be very useful to train a second BDT model to classify B-meson candidates.

Good point, @fgrosa , thanks for the suggestion.

vkucera:
> > Hi @vkucera, did you consider to add a joinable table with the D0 BDT output scores? I think it would be very beneficial to have them, since you will anyway have to apply the BDT models used in the trigger, and from my experience in the B0 cross section measurement, they can be very useful to train a second BDT model to classify B-meson candidates.
> 
> Good point, @fgrosa , thanks for the suggestion.

Added in [f14b809](https://github.com/AliceO2Group/O2Physics/pull/8485/commits/f14b809c629c4445b8751943c07d69ca07d5327c)

PR #8487, opened 2024-11-15
[PWGLF] Fix V0 DCA cut in Lambda polarisation task

jikim1290:
fix  bug for wrong v0 selection


skundu692 (approved):

vkucera:
@jikim1290 @skundu692 The original title "fix bug" was a perfect example of a bad commit message. Please take some effort to make meaningful titles.

PR #8490, opened 2024-11-16
[PWGLF] Add lower order cumulant terms for resonance flow

skundu692:
Add lower order cumulant terms to study the effect of non-uniform detector acceptance and event plane distribution on resonance flow measurement


alibuild (approved):

PR #8504, opened 2024-11-18
[PWGLF] AngularCorrelationsInJets.cxx: tiny update for new build check

ljoergen:


maciacco:
hi @ljoergen @ktf looks like the build fails due to insufficient disk space on the test machine

PR #8508, opened 2024-11-18
[Infrastructure] Update labeler.yml tag DM changes

njacazio:


njacazio:
@ddobrigk this would help us in reviewing the PRs in the LF. Could we make it part of the labeler?

.github/labeler.yml
```diff
@@ -23,6 +23,10 @@ infrastructure:
           - 'packaging/**'
           - 'pyproject.toml'
 
+datamodel:
+  - changed-files:
+      - any-glob-to-any-file: ['DataModel/**', '*/DataModel/**']
```
vkucera, line 28: @njacazio The first pattern can never be matched.

PR #8524, opened 2024-11-18
[Tutorial] Fix int8_t configurables

vkucera:
Default values of configurables of type int8_t are not correctly read from the JSON because they are interpreted as literal char values and the character code is used instead. It is therefore better to avoid them completely and use int instead.


vkucera:
Ping

PR #8529, opened 2024-11-18
[PWGCF] Add some selections to pt-effeciency code

JiangTao-thu:
Summary of Changes:

- Added additional track selection criteria, including:

  - Chi2 per TPC clusters
  - TPC crossed rows
  - Pt-dependent cuts for DCAxy and DCAz, tailored for Run 3 pp
  - `GlobalTrackRun3ITSMatching::Run3ITSall7Layers` selection

- Implemented selection criteria for vertex z.

- Enabled processing for specific runs of interest only.

### DIFF
--- a/PWGCF/Flow/Tasks/flowPtEfficiency.cxx
+++ b/PWGCF/Flow/Tasks/flowPtEfficiency.cxx
@@ -102,8 +102,7 @@ struct flowPtEfficiency {
     }
     if (cfgCutDCAxyppPass3Enabled) {
       myTrackSel.SetMaxDcaXYPtDep([](float pt) { return 0.004f + 0.013f / pt; });
-    }
-    else {
+    } else {
       myTrackSel.SetMaxDcaXY(cfgCutDCAxy);
     }
     myTrackSel.SetMinNClustersTPC(cfgCutTPCclu);
--- a/PWGCF/Flow/Tasks/flowPtEfficiency.cxx
+++ b/PWGCF/Flow/Tasks/flowPtEfficiency.cxx
@@ -17,8 +17,11 @@
 #include "Framework/runDataProcessing.h"
 #include "Framework/AnalysisTask.h"
 #include "Common/DataModel/TrackSelectionTables.h"
+#include "Common/DataModel/EventSelection.h"
 #include "Framework/ASoAHelpers.h"
 #include "Framework/HistogramRegistry.h"
+#include "Common/Core/TrackSelection.h"
+#include "Common/Core/TrackSelectionDefaults.h"
 
 using namespace o2;
 using namespace o2::framework;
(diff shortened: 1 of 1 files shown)


PWGCF/Flow/Tasks/flowPtEfficiency.cxx
```diff
@@ -71,13 +95,46 @@ struct flowPtEfficiency {
...
   {
     registry.fill(HIST("eventCounter"), 0.5);
+    if (!collision.sel8())
+      return;
+    if (tracks.size() < 1)
+      return;
+    if (cfgSelRunNumberEnabled) {
+      auto bc = collision.bc_as<aod::BCsWithTimestamps>();
+      int RunNumber = bc.runNumber();
+      if (!std::count(cfgRunNumberList.value.begin(), cfgRunNumberList.value.end(), RunNumber))
```
victor-gonzalez, line 133: Doing this on a per `bc` level may incur in a high CPU usage depending on the number of runs to consider
If the run number has already been checked and the coming `bc` brings the same run number there is not point for checking it again
Please, consider it for the next iteration

PWGCF/Flow/Tasks/flowPtEfficiency.cxx
```diff
@@ -89,8 +146,14 @@ struct flowPtEfficiency {
   }
   PROCESS_SWITCH(flowPtEfficiency, processReco, "process reconstructed information", true);
 
-  void processSim(aod::McCollision const&, soa::SmallGroups<soa::Join<aod::McCollisionLabels, aod::Collisions>> const& collisions, myMcParticles const& mcParticles)
+  void processSim(myMcCollisions::iterator const& collision, aod::BCsWithTimestamps const&, soa::SmallGroups<soa::Join<aod::McCollisionLabels, aod::Collisions>> const& collisions, myMcParticles const& mcParticles)
   {
+    if (cfgSelRunNumberEnabled) {
+      auto bc = collision.bc_as<aod::BCsWithTimestamps>();
```
victor-gonzalez, line 152: Same here
Also consider that the run number should be the same at detector level as at generator level

PWGCF/Flow/Tasks/flowPtEfficiency.cxx
```diff
@@ -89,8 +146,14 @@ struct flowPtEfficiency {
...
+      int RunNumber = bc.runNumber();
+      if (!std::count(cfgRunNumberList.value.begin(), cfgRunNumberList.value.end(), RunNumber))
```
victor-gonzalez, line 154: Same here
Also consider that the run number should be the same at detector level as at generator level

victor-gonzalez (approved):

PWGCF/Flow/Tasks/flowPtEfficiency.cxx
JiangTao-thu: Hi Victor, thank you for your comments! However, I have one question: how can I check the run number at the generator level? I temporarily can't find any methods in the AO2D tables or in the previous code to achieve this. Could you please provide some guidance?

victor-gonzalez: Well, I guess you are doing it in here
This is your generator level process function, right?
You are getting your generator level collision which brings the `bc` from which you extract the run number out of the `BCWithTimestamps` table
Are you still missing something?

PR #8546, opened 2024-11-19
[PWGCF] flow: add FlowContainer in FlowRunbyRun; add d2{4} in FlowTask

Luzhiyongg:


PWGCF/Flow/Tasks/FlowRunbyRun.cxx
```diff
@@ -164,6 +225,7 @@ struct FlowRunbyRun {
     // detect run number
     auto bc = collision.bc_as<aod::BCsWithTimestamps>();
     int runNumber = bc.runNumber();
+    float l_Random = fRndm->Rndm();
     if (std::find(RunNumbers.begin(), RunNumbers.end(), runNumber) == RunNumbers.end()) {
```
victor-gonzalez, line 229: If the run number has not changed with respect to the previous collision there is not point for searching again for it
Please consider it for the next iteration

victor-gonzalez (approved):
Please, consider my comment for the next iteration

PR #8537, opened 2024-11-19
Add single-track variables in B0 reduced task

fgrosa:

PR #8564, opened 2024-11-20
[PWGHF] Add single track variables in B+ reduced task

apalasciano:
Following modifications introduced in PR #8537

### DIFF
--- a/PWGHF/D2H/Tasks/taskBplusReduced.cxx
+++ b/PWGHF/D2H/Tasks/taskBplusReduced.cxx
@@ -35,58 +35,99 @@ namespace o2::aod
 {
 namespace hf_cand_bplus_lite
 {
-DECLARE_SOA_COLUMN(PtProng0, ptProng0, float);                               //! Transverse momentum of prong0 (GeV/c)
-DECLARE_SOA_COLUMN(PtProng1, ptProng1, float);                               //! Transverse momentum of prong1 (GeV/c)
-DECLARE_SOA_COLUMN(MProng0, mProng0, float);                                 //! Invariant mass of prong0 (GeV/c)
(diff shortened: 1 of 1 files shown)


PWGHF/D2H/Tasks/taskBplusReduced.cxx
```diff
@@ -511,54 +555,73 @@ struct HfTaskBplusReduced {
...
+          candD0.absEtaProngMin(),
+          candD0.itsNClsProngMin(),
+          candD0.tpcNClsCrossedRowsProngMin(),
+          candD0.tpcChi2NClProngMax(),
+          candD0.tpcNSigmaPiProng0(),
+          candD0.tofNSigmaPiProng0(),
+          candD0.tpcTofNSigmaPiProng0(),
+          candD0.tpcNSigmaKaProng1(),
+          candD0.tofNSigmaKaProng1(),
+          candD0.tpcTofNSigmaKaProng1(),
```
fgrosa, line 591: In case you do not distinguish here between D0 and D0bar, you should add both NsigmaPi and NsigmaKa for both prongs. Alternatively (better in my opinion) you can always have pion in position 0 and kaon in position 1, but for this you should check if it is D0 (so B-) or D0bar (so B+) to put the correct ones (`NsigmaPiProng1`, `NsigmaKaProng0` in the first case and viceversa in the second case)

fgrosa (changes requested):

PWGHF/D2H/Tasks/taskBplusReduced.cxx
apalasciano: Thanks for spotting this!

fgrosa (approved):

PR #8565, opened 2024-11-20
[PWGJE] additional features for GammaJetTrees

fjonasALICE:
This PR adds multiple features to the GammaJetTree for prompt photon analyses
- track matching including eta and phi information on emcal surface
- capability to store jets of any given set of radii
- added leading hadron pt to jet information
- added event occupancy information (crucial for Pb-Pb)
- capability to estimate UE in perpendicular cone also perpendicular to jets


nzardosh:
Hi @fjonasALICE I think the aod:: before the namespace for the jet tables needs to be added now

fjonasALICE:
> Hi @fjonasALICE I think the aod:: before the namespace for the jet tables needs to be added now

Fixed! Please apologize, this commit also snuck in one additional histogram to the emc correction framework

PR #8572, opened 2024-11-20
[PWGDQ,PWGEM] Completion of flow calculation with on-the-flight resolution factors

Elros60:
- Resolution factors provide by event-level task will be uploaded to CCDB and then read by Same-Event/Event-Mixing task in table-reader for flow analysis
- Some news QA histograms/profiles added

### DIFF
--- a/PWGDQ/Core/VarManager.h
+++ b/PWGDQ/Core/VarManager.h
@@ -310,6 +310,15 @@ class VarManager : public TObject
     kR2SP_AB,
     kR2SP_AC,
     kR2SP_BC,
+    kWR2SP_AB,
+    kWR2SP_AC,
+    kWR2SP_BC,
+    kR2SP_AB_Im,
+    kR2SP_AC_Im,
+    kR2SP_BC_Im,
+    kWR2SP_AB_Im,
+    kWR2SP_AC_Im,
+    kWR2SP_BC_Im,
     kR2SP_FT0CTPCPOS,
     kR2SP_FT0CTPCNEG,
     kR2SP_FT0ATPCPOS,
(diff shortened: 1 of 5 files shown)


PWGDQ/Core/VarManager.h
```diff
@@ -2939,10 +2956,27 @@ void VarManager::FillPairME(T1 const& t1, T2 const& t2, float* values)
...
+  values[kV2SP2] = values[kU2Q2Ev2] / values[kTwoR2SP2];
+  values[kV2EP1] = values[kCos2DeltaPhiEv1] / values[kTwoR2EP1];
+  values[kV2EP2] = values[kCos2DeltaPhiEv2] / values[kTwoR2EP2];
+
+  float V2ME_SP = values[kV2SP1] * values[kCos2DeltaPhiMu1] + values[kV2SP2] * values[kCos2DeltaPhiMu2];
+  float V2ME_EP = values[kV2EP1] * values[kCos2DeltaPhiMu1] + values[kV2EP2] * values[kCos2DeltaPhiMu2];
+  values[kV2ME_SP] = std::isnan(V2ME_SP) || std::isinf(V2ME_SP) ? 0. : V2ME_SP;
+  values[kWV2ME_SP] = std::isnan(V2ME_SP) || std::isinf(V2ME_SP) ? 0. : 1.0;
+  values[kV2ME_EP] = std::isnan(V2ME_EP) || std::isinf(V2ME_EP) ? 0. : V2ME_EP;
+  values[kWV2ME_EP] = std::isnan(V2ME_EP) || std::isinf(V2ME_EP) ? 0. : 1.0;
```
iarsene, line 2979: All these calculations have to be done only if they are requested, otherwise they will be run by all the analyses which will slow down analyses. So please add a protection so they are computed only in the case they are needed.

Elros60: Thanks a lot for your comment. So then I will need to revert my last commit; I removed it because it touched also part of PWGEM code where they used also table-reader

Elros60: I reverted my last commit, and changed the label as well to pass the `Pull Request Labeler` check

iarsene (approved):

iarsene:
@rbailhac  Could you approve this please?

ivorobye (approved):

PR #8582, opened 2024-11-20
[PWGHF] add the v1 for D0

zhangbiao-phy:
Hi @adubla, Here is the commit for D0 v1. Thanks, @prottayCMT, for creating the task! I’ve also extended the BDT score to include the prompt in addition to adding the D0 case, as we sometimes need this control to achieve results with a high prompt fraction.

Let me know if you have any feedback! Thanks!

### DIFF
--- a/PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
+++ b/PWGHF/D2H/Tasks/taskDirectedFlowCharmHadrons.cxx
@@ -13,6 +13,7 @@
 /// \brief Analysis task for charm hadron directed flow
 ///
 /// \author Prottay Das, prottay.das@cern.ch
+/// \author Biao Zhang, biao.zhanng@cern.ch
 
 #include <string>
 #include <vector>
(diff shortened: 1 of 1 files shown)


fgrosa:
Hi @zhangbiao-phy couldn't high prompt contribution achieved by requiring low non-prompt BDT score? In Pb-Pb the sparses might become very large, so avoiding one axis might be very beneficial ;-)

zhangbiao-phy:
> Hi @zhangbiao-phy couldn't high prompt contribution achieved by requiring low non-prompt BDT score? In Pb-Pb the sparses might become very large, so avoiding one axis might be very beneficial ;-)
indeed, It can be, but it is dependent on your preselections, sometimes, even the non-prompt score is 0, the prompt fraction is still not very high. Ok, but I agree with you,  to avoid the Sparse being too large,  I keep 2 Axises for that. we can add it once we need it

fgrosa (approved):

PR #8598, opened 2024-11-21
[PWGCF] Added a procedure for calculating the PID purity in MC

JStaa:


victor-gonzalez (approved):
I observed that the `TrackPID` routine is repeated for each track for each `V0` (L.528)
This is quite inefficient and scales in the wrong way with the number of `V0`
I guess in the case of event mixing the situation would be even worse
Try to consider it for the next iteration

PR #8603, opened 2024-11-21
[DPG] Add occupancy information in qa efficiency task

stefanopolitano:
This PR adds the centrality and occupancy information in the efficiency task for qa


mfaggin (approved):

chiarazampolli:
Hello @stefanopolitano , 

This is breaking the analysis QC, can you tell us how you run the tests before you committed the changes (including how you configured the centrality task)? We can try to port them to the analysis QC. Otherwise we'll have ti either revert this change, or exclude this task from the analysis QC of apass1 PbPb 2024, which would not be that good either (@mfaggin ).

Thanks for your understanding and help. 

Chiara

@ekryshen

mfaggin:
from our side, this is not used for TRK, so not a big deal is this is excluded.
But if you point me to the failing jobs I can try to have a look

ekryshen:
@mfaggin It is failing in Analysis QC tests because centrality task is not added as a dependency. If you don't use the output of this task in the AOTTrack QC, I would suggest to set this line to false:
https://github.com/AliceO2Group/O2DPG/blob/master/MC/config/analysis_testing/json/analyses_config.json#L15

If you think it is crucial to have it, we should add a dependency on centrality task and corresponding config in:
https://github.com/AliceO2Group/O2DPG/tree/master/MC/config/analysis_testing/json/dpl

PR #8610, opened 2024-11-22
[PWGCF] FemtoUniverse - D0 task update

kgwizdzi:
Changes:
- pT configurables for tracks
- pT cuts moved from process function to the partitions


vkucera:
@kgwizdzi If you miss the space after the PWG prefix, it is not a prefix any more and cannot be recognised by the bot which leads to duplication.

zchochul (approved):

PR #8611, opened 2024-11-22
[Common] add track extra converter to v002

maciacco:
@mpuccio This PR adds a new converter for the tracks extra table to v002, which includes the difference between findable and PID clusters in TPC https://github.com/AliceO2Group/AliceO2/pull/13617. This converter contains two process functions to act on both v000 and v001.


ddobrigk (approved):
thanks a lot, looks good to me!

mpuccio:
Looks good to me too!

PR #8612, opened 2024-11-22
[PWGLF] Initial commit of KF hypernuclei task

mhartung71:
Initial commit of KF hypernuclei task


fmazzasc:
HI @mhartung71 , thanks for your work. I have two questions:
- Why did't you use the 2- and 3- body structures that come out from the SVertexer? This has two pros: speed-up of your code as would avoid to redo the combinatorics, recover doblets/triplets where one of the prongs is not associated to the right collision
- Could you delete redundant data members for your output tree? For example you store all the momenta of the daughters + the one of the mother

mhartung71:
Dear Francesco,

thank you very much for your quick reply. The code is created to reconstruct both single and double hypernuclei (e.g. 4LLH->4LHe+pi). However, the SVertexer is not optimized for the decay of double hypernuclei, so we decided to use the KFParticle package (KFP). Furthermore, we also want to check whether KFP achieves a higher reconstruction efficiency compared to the SVertexer (discussed by Maximiliano and Benjamin). 

The different momenta of mother and daughter particles are not equivalent, since the mother momentum is optimized during the reconstruction by KFP. Calculating the mother momentum from the daughter momenta can lead to a reduction in resolution, with double hypernuclei being affected in particular, as the effect occurs twice. But we cannot yet say exactly how strong the impact is. Our plan is to start with a run over MC and a part of data (LHC23zzh_pass4_small). After the evaluation, we will revise the code and delete information from the tree that is not needed.

Best regards,
Michael and Janik

fmazzasc (approved):

njacazio:
Hi, the macOS test is failing, I would start fixing this before going on with the merge

njacazio (dismissed):
Thanks, hopefully it should fix the CI

njacazio (approved):

PR #8630, opened 2024-11-23
[PWGHF]: Add QA histograms for tracks and charm hadron before pairing

zhangbiao-phy:
- To enable more convenient comparison with CF femtoscopy analysis: adding QA histograms for tracks and charm hadrons before pairing.


fgrosa (approved):
Good for HF part, @lauraser @ariedel-cern @victor-gonzalez could you please check the CF one? Thanks!

lauraser (approved):

PR #8645, opened 2024-11-25
[PWGCF] FemtoUniverse: From modular to nested filling of trks

prchakra:
In this PR, I changed the modular or independent (of the collision table) filling of tracks to the nested filling of tracks. The track table will be filled only if the collision is selected. The changes are made only for the centrality analyses (i.e. Pb-Pb system). Next, I'll implement the same changes for multiplicity analyses also.


alibuild (approved):

PR #8655, opened 2024-11-25
[Trigger] Add option to skip untriggered events

mpuccio:
This change adds a new configurable option `cfgSkipUntriggeredEvents` to the
`centralEventFilterTask` in the `EventFiltering` module. When this option is
enabled, the task will skip processing events where both the trigger and
decision flags are false, reducing the overall processing output.


alibuild (approved):

PR #8661, opened 2024-11-26
[Common] To fix mistake for getParticleOrigin

hanseopark:
I have mistaken to write this function.
The direction I want to go is that if the mother does not have a parton, then I should check the particle's pdgcode.
Because if a hadron does not have a parton as its mother, but it is not a simple break, we can assign the origin of the particle based on the pdgcode of the hadron.
if it breaks, it move to `OriginType::None`


hanseopark:
I would like to check this PR

PR #8664, opened 2024-11-26
[PWGCF] Updated pair cleaner

lauraser:
Fixed the Pair Cleaner, to check if the daughter is really not the same track as primary track in the pairs based on information stored at the producer level


alibuild (approved):

PR #8669, opened 2024-11-26
[PWGLF] Fix KF cascade daughter DCA calculation in cascadebuilder

creetz16:
Hi @ddobrigk 
You probably remember that I added this part some time ago. After another detailed discussion with the KF authors it turned out that we should not use the SetProductionVertex in this case. I changed the code accordingly. 
Let me know if you have any comments on this. :)
Cheers
Carolina


njacazio (approved):

njacazio:
Thanks Carolina

njacazio:
CI seems unrelated

PR #8688, opened 2024-11-27
[PWGJE] Task for hadron-jet correlations in O-O collisions

KotliarovAr:
A new task designed to study hadron-jet correlations in the upcoming OO collisions.

Dear @nzardosh,
Could you review the code? Also, I want to ask 2 questions:

1. we discussed some time ago that the JE framework does not contain functionality to estimate rho background for particle level MC. Do we plan to add such functionality?

2. when we want to get a weight for MC detector level simulations, we use ``auto weight = collision.mcCollision().weight()``. There is a function that checks whether a collision has an associated MC collision: ``collision.has_mcCollision()``. So, not all detector-level collisions have associated particle-level collisions? Are we protected from the output not being garbage due to the absence of the associated MC collision when calling ``collision.mcCollision().weight()``? Should we also introduce a check (``collision.has_mcCollision()``) that discards the MC detector level collisions, which do not have associated MC particle level collisions?

### DIFF
--- a/PWGJE/Tasks/CMakeLists.txt
+++ b/PWGJE/Tasks/CMakeLists.txt
@@ -60,6 +60,10 @@ if(FastJet_FOUND)
                         SOURCES jetSubstructureLc.cxx
                         PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::PWGJECore O2Physics::AnalysisCore
                         COMPONENT_NAME Analysis)
+o2physics_add_dpl_workflow(jet-substructure-bplus
+                        SOURCES jetSubstructureBplus.cxx
+                        PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::PWGJECore O2Physics::AnalysisCore
+                        COMPONENT_NAME Analysis)
     o2physics_add_dpl_workflow(jet-substructure-dielectron
                         SOURCES jetSubstructureDielectron.cxx
                         PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::PWGJECore O2Physics::AnalysisCore
@@ -70,6 +74,10 @@ if(FastJet_FOUND)
                         COMPONENT_NAME Analysis)
     o2physics_add_dpl_workflow(jet-substructure-lc-output
                         SOURCES jetSubstructureLcOutput.cxx
+                        PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::PWGJECore O2Physics::AnalysisCore
+                        COMPONENT_NAME Analysis)
+    o2physics_add_dpl_workflow(jet-substructure-bplus-output
+                        SOURCES jetSubstructureBplusOutput.cxx
                         PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::PWGJECore O2Physics::AnalysisCore
                         COMPONENT_NAME Analysis)
     o2physics_add_dpl_workflow(jet-substructure-dielectron-output
@@ -102,6 +110,10 @@ if(FastJet_FOUND)
                         COMPONENT_NAME Analysis)
     o2physics_add_dpl_workflow(jet-finder-lc-qa
                         SOURCES jetFinderLcQA.cxx
+                        PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::PWGJECore O2Physics::AnalysisCore
+                        COMPONENT_NAME Analysis)
+    o2physics_add_dpl_workflow(jet-finder-bplus-qa
+                        SOURCES jetFinderBplusQA.cxx
                         PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::PWGJECore O2Physics::AnalysisCore
                         COMPONENT_NAME Analysis)
     o2physics_add_dpl_workflow(jet-finder-dielectron-qa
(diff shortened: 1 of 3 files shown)


PWGJE/Tasks/jetHadronRecoil_OO.cxx
```diff
@@ -0,0 +1,541 @@
...
+  }
+  PROCESS_SWITCH(jetHadronRecoil_OO, processJetsMatched_Weighted, "process matching of MC jets (weighted)", false);
+
+  //------------------------------------------------------------------------------
+  // Auxiliary functions
+  template <typename Collision>
+  bool skipEvent(const Collision& coll)
+  {
+    /// \TODO: Do need !jetderiveddatautilities::selectTrigger(collision, triggerMaskBits) flag to skip an event?
+    return !coll.selection_bit(aod::evsel::kNoTimeFrameBorder) || !jetderiveddatautilities::selectCollision(coll, eventSelection);
```
nzardosh, line 518: hi Artem, isnt the NoTimeFrameBorder cut already included in sel8?
For this one selection this means that you are not fully using the derived data and are instead relying on the evsels table which we should find a way of avoiding. If the timeframe border cut is not included we can add this as a selection in the jet framework that can be accessed with jetderiveddatautilities::selectCollision

PWGJE/Tasks/jetHadronRecoil_OO.cxx
```diff
@@ -0,0 +1,541 @@
...
+  }
+
+  template <typename Jet>
+  std::tuple<double, bool> isRecoilJet(const Jet& jet,
+                                       double phi_TT)
+  {
+    double dphi = fabs(TVector2::Phi_mpi_pi(jet.phi() - phi_TT));
```
nzardosh, line 525: for this use the RecoDecay::constrainAngle function

PWGJE/Tasks/jetHadronRecoil_OO.cxx
nzardosh: If you could fix it now it would be great, because i think this way if you remove the need for it then you can remove the dependency on the Event selection tables and then you can run directly on the JE derived data

PWGJE/Tasks/jetHadronRecoil_OO.cxx
KotliarovAr: Added
Interesting to know why we should use ``RecoDecay::constrainAngle`` instead of ``TVector2::Phi_mpi_pi``?

nzardosh: recodecay is the official place where these operations are done so the recommendation from O2 is to use that where possible. @vkucera can comment more

nzardosh (approved):

nzardosh:
My final comment is, why do you call this task OO? In principle it can be used for any system right? It would be good to rename it

PWGJE/Tasks/jetHadronRecoil_OO.cxx
vkucera: @KotliarovAr O2 is C++ code, not ROOT code. There is no reason to use ROOT features if equivalents exist in standard C++ or O2.

KotliarovAr:
> My final comment is, why do you call this task OO? In principle it can be used for any system right? It would be good to rename it

You are right, the code can be also used for pp collisions. However, there is a similar code for h+jet analysis written by Danny and Jaime, which analyses in addition WTA jet axis. So, renaming could confuse.

nzardosh:
> > My final comment is, why do you call this task OO? In principle it can be used for any system right? It would be good to rename it
> 
> You are right, the code can be also used for pp collisions. However, there is a similar code for h+jet analysis written by Danny and Jaime, which analyses in addition WTA jet axis. So, renaming could confuse.

It would be good if we could find a different name than OO though to make them distinct.

PWGJE/Tasks/CMakeLists.txt
```diff
@@ -168,8 +168,8 @@ o2physics_add_dpl_workflow(jet-substructure-bplus
                         SOURCES jetHadronRecoil.cxx
                         PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::PWGJECore O2Physics::AnalysisCore
                         COMPONENT_NAME Analysis)
-    o2physics_add_dpl_workflow(jet-hadron-recoil-oo
-                        SOURCES jetHadronRecoil_OO.cxx
+    o2physics_add_dpl_workflow(trigger-hadron-recoil-jet
+                        SOURCES recoilJets.cxx
```
vkucera, line 172: These don't match.

PWGJE/Tasks/recoilJets.cxx
```diff
@@ -538,4 +538,4 @@ struct jetHadronRecoil_OO {
   }
 };
 
-WorkflowSpec defineDataProcessing(ConfigContext const& cfgc) { return WorkflowSpec{adaptAnalysisTask<jetHadronRecoil_OO>(cfgc, TaskName{"jet-hadron-recoil-oo"})}; }
+WorkflowSpec defineDataProcessing(ConfigContext const& cfgc) { return WorkflowSpec{adaptAnalysisTask<recoilJets>(cfgc, TaskName{"trigger-hadron-recoil-jet"})}; }
```
vkucera, line 541: There is no reason to use `TaskName`.

vkucera: Not addressed

KotliarovAr: Could you clarify a bit more what you suggest?

vkucera: It is all written in the linter warnings. Which message is not clear?

KotliarovAr: I see your point. So, the part ``TaskName{"recoil-jets"}`` can be omitted. It will be implemented in the next PR, thank you

vkucera: Yes, not only it can be omitted but it should be omitted. Please also fix the `struct` name to `RecoilJets` together with it. Thanks!

PR #8692, opened 2024-11-27
[Common] [TOF] streamline ccdb configuration

njacazio:


njacazio:
@fmazzasc I removed the template from the class and moved it to the static function for `GetBeta` this way it is easier to use, just double check the code pls.

njacazio:
This also catches the case where the run is changed in the same workflow and the pass from the metadata is taken and stored in the ccdb manager. Thanks to @rolavick

ddobrigk (approved):

alibuild (approved):

PR #8700, opened 2024-11-27
[PWGHF] Adding occupcany vs centrality info in HF ev. sel. utils

stefanopolitano:
This PR adds the possibility of storing occupancy vs centrality in the HF vent selection utils

### DIFF
--- a/PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
+++ b/PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
@@ -631,7 +631,7 @@ struct HfTaskFlowCharmHadrons {
   void processResolution(CollsWithQvecs::iterator const& collision,
                          aod::BCsWithTimestamps const& bcs)
   {
-    float centrality {-1.f};
+    float centrality{-1.f};
     if (!isCollSelected<o2::hf_centrality::CentralityEstimator::None>(collision, bcs, centrality)) {
       // no selection on the centrality is applied on purpose to allow for the resolution study in post-processing
       return;
--- a/PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
+++ b/PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
@@ -64,8 +64,7 @@ struct HfTaskFlowCharmHadrons {
   Configurable<float> centralityMax{"centralityMax", 100., "Maximum centrality accepted in SP/EP computation (not applied in resolution process)"};
   Configurable<bool> storeEP{"storeEP", false, "Flag to store EP-related axis"};
   Configurable<bool> storeMl{"storeMl", false, "Flag to store ML scores"};
-  Configurable<int> occEstimator{"occEstimator", 1, "Occupancy estimation (1: ITS, 2: FT0C)"};
-  Configurable<bool> storeOccupancy{"storeOccupancy", false, "Flag to store TH2 occITS/occFT0C + HfEvSelBitMasks"};
+  Configurable<int> occEstimator{"occEstimator", 0, "Occupancy estimation (0: None, 1: ITS, 2: FT0C)"};
   Configurable<bool> saveEpResoHisto{"saveEpResoHisto", false, "Flag to save event plane resolution histogram"};
   Configurable<std::string> ccdbUrl{"ccdbUrl", "http://alice-ccdb.cern.ch", "url of the ccdb repository"};
   Configurable<std::vector<int>> classMl{"classMl", {0, 2}, "Indexes of BDT scores to be stored. Two indexes max."};
@@ -147,7 +146,7 @@ struct HfTaskFlowCharmHadrons {
     if (storeMl) {
       axes.insert(axes.end(), {thnAxisMlOne, thnAxisMlTwo});
     }
-    if (storeOccupancy) {
+    if (occEstimator != 0) {
       if (occEstimator == 1) {
         axes.insert(axes.end(), {thnAxisOccupancyITS, thnAxisNoSameBunchPileup, thnAxisNumTracksInTimeRange,
                                  thnAxisNoCollInTimeRangeNarrow, thnAxisNoCollInTimeRangeStandard, thnAxisNoCollInRofStandard});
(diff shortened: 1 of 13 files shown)


PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -188,7 +188,7 @@ struct HfTrackIndexSkimCreatorTagSelCollisions {
   PROCESS_SWITCH(HfTrackIndexSkimCreatorTagSelCollisions, processTrigSel, "Use trigger selection", false);
 
   /// Event selection without trigger selection
-  void processNoTrigSel(aod::Collision const& collision, aod::BCsWithTimestamps const& bcs)
+  void processNoTrigSel(soa::Join<aod::Collisions, aod::EvSels>::iterator const& collision, aod::BCsWithTimestamps const& bcs)
```
mfaggin, line 191: is it necessary? In principle everything is templetized properly, and even w/o the `Join` the compilation should not complain. Even, this is the function w/o "trigger sleections" meant as "event selections" (see the `false` template parameter, which consequently invokes `hfEvSel.getHfCollisionRejectionMask<applyTrigSel, ...>()` that does not require the `Join` with `aod::EvSels`). I'd leave this piece untouched.

mfaggin (changes requested):

PWGHF/TableProducer/trackIndexSkimCreator.cxx
stefanopolitano: Hi @mfaggin , the `processNoTrigSel` function calls `hfEvSel.getHfCollisionRejectionMask` which with this PR needs the occupancy information stored in `aod::EvSels`. If I do not join the tables, it breaks at compilation time. If I understood correctly, the joining should not have any effect since the  `applyTrigSel` is still false and avoids the event selections. Of course if you have a better solution I will implement it rightaway!

mfaggin: Ciao @stefanopolitano are you sure that the compilation fails? Have you explicitly seen it? This would be surprising to me, since the function is templetized (see `typename Coll` here https://github.com/AliceO2Group/O2Physics/blob/2363ac8e965e9b6f15a218926b31c32b20bfe32b/PWGHF/Utils/utilsEvSelHf.h#L162-L163). This would surprise me, honestly.

What I see that it can be causing some troubles is the fact that with this PR you add the `TH2` to monitor the selected events vs. centrality and occupancy, and the occupancy info is indeed in the `EvSel` table. However, this issue affects only the function `fillHistograms`, and in my opinion you can overcome it easily as for the `centrality`, i.e. defining a default argument `float occupancy = -1` that you agnostically use to fill the `TH2` (in other words: the decision to use `collision.ft0cOccupancyInTimeRange()` or `collision.trackOccupancyInTimeRange()` is moved when calling the function itself).

stefanopolitano: Hi, I see your point. Ok then, I will modify the code to have a default occupancy value. I am adding the occupancy info in the creators and in the flow task to proceed with the occupancy study. Thanks!

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
```diff
@@ -386,10 +386,11 @@ struct HfTaskFlowCharmHadrons {
                       aod::BCsWithTimestamps const&)
   {
     float centrality{-1.f};
+    float occupancy = getOccupancy(collision);
```
mfaggin, line 389: I think that you can define this function in `HfEventSelection` so that everyone can call it in the same way

mfaggin (changes requested):

mfaggin (approved):
thanks a lot @stefanopolitano

PWGHF/Utils/utilsEvSelHf.h
```diff
@@ -268,11 +259,51 @@ struct HfEventSelection : o2::framework::ConfigurableGroup {
     return rejectionMask;
   }
 
+  /// Get the occupancy
+  /// \param collision is the collision with the occupancy information
+  /// \param occEstimator is the occupancy estimator (1: ITS, 2: FT0C)
+  template <typename Coll>
+  float getOccupancy(Coll const& collision, int occEstimator = 1)
+  {
```
vkucera, line 267: Please try to avoid literal numbers as options. Define the cases in `enum`.

PWGHF/Utils/utilsEvSelHf.h
```diff
@@ -268,11 +259,51 @@ struct HfEventSelection : o2::framework::ConfigurableGroup {
...
+        return collision.trackOccupancyInTimeRange();
+        break;
+      case 2: // FT0c
+        return collision.ft0cOccupancyInTimeRange();
+        break;
+      default:
+        LOG(warning) << "Occupancy estimator not valid. Possible values are ITS or FT0C. Fallback to ITS";
+        return collision.trackOccupancyInTimeRange();
+        break;
+    }
```
vkucera, line 279: Returning a fallback value just with a warning when an invalid value is provided look like a bad idea to me. Invalid input should cause a fatal error.

PWGHF/Utils/utilsEvSelHf.h
```diff
@@ -107,6 +107,8 @@ struct HfEventSelection : o2::framework::ConfigurableGroup {
   o2::framework::Configurable<std::string> softwareTrigger{"softwareTrigger", "", "Label of software trigger. Multiple triggers can be selected dividing them by a comma. Set None if you want bcs that are not selected by any trigger"};
   o2::framework::Configurable<uint64_t> bcMarginForSoftwareTrigger{"bcMarginForSoftwareTrigger", 100, "Number of BCs of margin for software triggers"};
   o2::framework::Configurable<std::string> ccdbPathSoftwareTrigger{"ccdbPathSoftwareTrigger", "Users/m/mpuccio/EventFiltering/OTS/", "ccdb path for ZORRO objects"};
+  o2::framework::ConfigurableAxis th2ConfigAxisCent{"th2ConfigAxisCent", {100, 0., 100.}, ""};
+  o2::framework::ConfigurableAxis th2ConfigAxisOccupancy{"th2ConfigAxisOccupancy", {14, 0, 140000}, ""};
```
vkucera, line 111: This is a huge number of bins. Is it really needed?

PR #8705, opened 2024-11-27
[PWGDQ] Enable running over derived data from skimmed datasets

romainschotter:
- add option to produce same-sign pairs
- allow to run over derived data produced from skimmed datasets
- remove warnings


iarsene (approved):

iarsene:
@romainschotter
I approved the PR and we can merge it like this since it already passed the o2 tests. But for your next PR, please consider to fix the MegaLinter errors. I know these are not formally required, but all these errors are there due to deprecated behaviour, which means at some point in the future these might break compilation. So better to take care of them now.

PR #8717, opened 2024-11-28
[PWGHF] Add PV contributors in D0 sparse for MC and remove unnecessary axes

fgrosa:
Hi @Mingyu3360715 this is the change that we discussed offline:
- I added PV contributors (needed for JPsi-D efficiency maps) in the sparses for MC
- I removed a couple of unnecessary axes for data, to reduce data size (note that I also changed slightly the order of the axes, to have the common one in data and MC in the same position)

Informing also @atavirag


alibuild (approved):

PR #8581, opened 2024-11-20
[PWGHF] Refactor derived-data creators

vkucera:

PR #8726, opened 2024-11-29
[PWGHF,PWGJE] Adding B+ to the jet workflow

nzardosh:


PWGHF/TableProducer/derivedDataCreatorBplusToD0Pi.cxx
```diff
@@ -127,70 +129,11 @@ struct HfDerivedDataCreatorBplusToD0Pi {
...
-    if (fillMcRCollId) {
-      // Fill the table with the vector of indices of derived reconstructed collisions matched to mcCollision.globalIndex()
-      rowMcRCollId(
-        matchedCollisions[mcCollision.globalIndex()]);
-    }
-  }
-
   template <typename T, typename U, typename V>
-  void fillTablesCandidate(const T& candidate, const U& prongCharm, const V& prongBachelor, int candFlag, double invMass,
+  void fillTablesCandidate(const T& candidate, const U& prongCharm, const V& prongBachelor, int candFlag, float invMass,
```
vkucera, line 133: This is wrong.

PWGHF/DataModel/DerivedTables.h
```diff
@@ -585,47 +542,24 @@ DECLARE_SOA_TABLE_STAGED(HfBplusParEs, "HFBPPARE", //! Table with additional can
                          hf_cand_par::Ct,
                          o2::soa::Marker<MarkerBplus>);
 
-DECLARE_SOA_TABLE_STAGED(HfBplusMls, "HFBPML", //! Table with candidate selection ML scores
-                         hf_cand_mc::MlScoreSig,
+DECLARE_SOA_TABLE_STAGED(HfBplusMls, "HFBPML",   //! Table with candidate selection ML scores
+                         hf_cand_mc::MlScoreSig, // why is this the signal ML score instead of the full one?
```
vkucera, line 546: Please remove the comment.

vkucera (changes requested):
This PR is mixing at least 4 completely independent features and must be adjusted:

- Adding B+ in the jet workflow
  - This part has nothing to do with the HF code and should be the only content of this PR (as the title suggests).
- Renaming the common and 3P derived HF tables
  - This is not required to add the B+ in jets. It should be done in a separate PR before or after this one.
- Declaring derived HF tables with macros
  - This is not required to add the B+ in jets. This part is only relevant to making the HF tables unique as a part of the table content unification. It should be done in a separate PR.
- Factoring out common methods for filling derived HF tables into a utility class
  - This is not required to add the B+ in jets and it is duplicating features already covered in the unification PR https://github.com/AliceO2Group/O2Physics/pull/8581 . It should be dropped completely.

PR #8742, opened 2024-11-29
[PWGHF] Implement ML models in D0-h correlation

samrangy:
Use of ML models has been implemented for data only. Yet to be taken care for MC.


apalasciano:
Dear Samrangy, please rebase your branch to include the fix in #8762 which will solve the current errors in compilation.

apalasciano (approved):

PR #8746, opened 2024-11-30
[PWGCF] minor additions and fixes

abilandz:


PWGCF/MultiparticleCorrelations/Core/MuPa-MemberFunctions.h
```diff
@@ -4257,12 +4276,37 @@ void CheckCurrentRunNumber(T1 const& collision, T2 const&)
...
+    if (!(tc.fRunTime[eEndOfRun] == endOfRun)) {
+      LOGF(error, "\033[1;33m%s tc.fRunTime[eEndOfRun] changed within process(). This most likely indicates that a given masterjob is processing 2 or more different runs in one go.\033[0m", __FUNCTION__);
+      LOGF(fatal, "tc.fRunTime[eEndOfRun] = %d, endOfRun = %d", tc.fRunTime[eEndOfRun], endOfRun);
+    }
+
+    // **) Insanity check on run duration:
+    if (!(tc.fRunTime[eDurationInSec] == durationInSec)) {
+      LOGF(error, "\033[1;33m%s tc.fRunTime[eDurationInSec] changed within process(). This most likely indicates that a given masterjob is processing 2 or more different runs in one go.\033[0m", __FUNCTION__);
+      LOGF(fatal, "tc.fRunTime[eDurationInSec] = %d, durationInSec = %d", tc.fRunTime[eDurationInSec], durationInSec);
+    }
```
victor-gonzalez, line 4308: If this is being done per BC is useless because the object from the CCDB is always the same it only changes when the run number changes and this usually does not happen at the grid nodes
Of course, locally it can be enforced

victor-gonzalez (approved):
Have a look at my comment

BTW, Vit has made a tremendous effort to put available a set of checks that will help to follow the O2 coding guidelines, to increase code readability, and to catch some kind of execution errors which sometimes don't directly manifest themselves
As you could have seen they produce a failure in the `linter` CI test providing a detailed information on what the error is and how to fix it
As today they are not mandatory for getting the PR approved and merged
But it would be good to try to start implementing the different suggestions at each own pace
As it is said in the final line of the CI test failure report under `Details`, the tests can be run locally
They are really good suggestions!!

PWGCF/MultiparticleCorrelations/Core/MuPa-MemberFunctions.h
abilandz: I do NOT use this check regularly: "  // Used only during validation."

It's a protection that I do not process in one masterjob data from different runs. Once that is checked, the call to this function is switched off. 

In practice, this function is never called, unless weird things start to happen...

PR #8752, opened 2024-12-01
Update NucleiHistTask.cxx

RafRaf11:
Added on the fly (track-by-track) shifting of TPCnSigma distributions using TF1.


RafRaf11:
Hi, do I need to restart the PR? Yesterday the macOS-arm check didn't have the "required" flag

PR #8754, opened 2024-12-01
[PWGCF] DptDpt - Splitting the PID QC information

victor-gonzalez:
Also getting rid of the use of `TDatabasePDG`


alibuild (approved):

PR #8761, opened 2024-12-02
[PWGUD] Modify MC saved tables

ariffero:
- add mass to the variables saved in recom MC
- save table for gen MC


ariffero:
The O2 linter check continues to fail, but the only two warnings left imply the change of the name of the task and of the workflow, and if I change them the CMakeList.txt file is included in the PR and a lot of other warnings (not related to my task) appear.

rolavick (approved):

PR #8762, opened 2024-12-02
Fix wrong type for receiving container

ktf:


ekryshen:
@ktf Thanks for the fix

PR #8769, opened 2024-12-02
[PWGJE]Add separat HG/LG cell time QC histos for time calib

nstrangm:
- For now the time correction parameters are the same for LG and HG cells, will be adjusted as soon as possible


mhemmer-cern (approved):
Looks good from my side.
O2 linter changes should be looked into at some point, but not necessary right now.

vkucera:
Disabling auto-merge because of a wrong prefix.

PR #8772, opened 2024-12-02
[Infrastructure] Don't try to lint when no files are found

singiamtel:
Currently it fails the check without files

CC @vkucera


vkucera (commented):
Thanks @singiamtel , I was about make the same fix.
Maybe this could be formatted as a `notice::` GitHub annotation so that it appears in the summary of the action run.

vkucera:
Indeed, I think your branch setup is not very common. Most analysers use the master branch to run their analysis and if they experience problems, updating it is usually the first advice they get. ;-)

vkucera (changes requested):
Thanks @singiamtel

vkucera (approved):

vkucera:
@singiamtel @ktf Is this ready for merging?

PR #8779, opened 2024-12-02

yashpatley:
1. Modified the lambda selection methods
2. Added correction factor histograms from ccdb

### DIFF
--- a/PWGCF/TwoParticleCorrelations/Tasks/lambdaR2Correlation.cxx
+++ b/PWGCF/TwoParticleCorrelations/Tasks/lambdaR2Correlation.cxx
@@ -9,7 +9,7 @@
 // granted to it by virtue of its status as an Intergovernmental Organization
 // or submit itself to any jurisdiction.
 
-/// \file lambdaCorrelationAnalysis.cxx
+/// \file lambdaR2Correlation.cxx
 /// \brief R2 correlation of Lambda baryons.
 /// \author Yash Patley <yash.patley@cern.ch>
 
@@ -33,6 +33,7 @@ using namespace o2;
 using namespace o2::framework;
 using namespace o2::framework::expressions;
 using namespace o2::constants::physics;
+using namespace o2::constants::math;
 
 namespace o2::aod
 {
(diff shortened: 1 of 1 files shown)


PWGCF/TwoParticleCorrelations/Tasks/lambdaR2Correlation.cxx
```diff
@@ -440,55 +459,123 @@ struct lambdaCorrTableProducer {
...
+    } else if (obj->InheritsFrom("TH3F")) {
+      TH3F* hist = reinterpret_cast<TH3F*> obj->Clone();
+      int pt_bin = hist->GetXaxis()->FindBin(v0.pt());
+      int rap_bin = hist->GetYaxis()->FindBin(v0.yLambda());
+      int vz_bin = hist->GetZaxis()->FindBin(col.posZ());
+      return hist->GetBinContent(pt_bin, rap_bin, vz_bin);
+    } else {
+      return 1.;
     }
   }
```
victor-gonzalez, line 573: Please, run it locally before pushing the functionality
There are histograms that are cloned on a per V0 basis and that are left there, allocated, which creates memory leaks

victor-gonzalez (changes requested):
Please, have a look at my comment
Functionality should be pushed once it has been tested locally

PWGCF/TwoParticleCorrelations/Tasks/lambdaR2Correlation.cxx
yashpatley: Dear Victor, I will keep this in mind. I missed compiling the task after the last change, it will not happen again.
I understand the thing about memory leak. I will rectify it in next PR. I am not going to use CCDB object this time because I want to get the correction factors with this task. I added the ccdb method for future purposes. I will keep the memory leak thing in mind and thank you for your suggestion.

victor-gonzalez (approved):
Have in mind that all histograms are derived from `TH1` and that it has a method for knowing the number of dimensions one histogram actually has

PR #8783, opened 2024-12-02
[PWGCF] DptDpt - Efficiency and QC - Fixing linter errors

victor-gonzalez:


alibuild (approved):

victor-gonzalez:
Hi @vkucera, I've modified the `dptdptEfficiencyAndQc.cxx` source file to fix all `linter` errors/warnings but I reach a contradictory (false positive?) situation. The check of the `CMakeLists.txt` suggests to use `dptdptEfficiencyAndQc` as file name but if it is used we get the error that the main `struct` should have the same name but at the same time the name of the struct must be `UpperCamelCase`
BTW, the source file has several `struct` structures on it but only two of them are actually tasks

vkucera:
> Hi @vkucera, I've modified the `dptdptEfficiencyAndQc.cxx` source file to fix all `linter` errors/warnings but I reach a contradictory (false positive?) situation. The check of the `CMakeLists.txt` suggests to use `dptdptEfficiencyAndQc` as file name but if it is used we get the error that the main `struct` should have the same name but at the same time the name of the struct must be `UpperCamelCase` BTW, the source file has several `struct` structures on it but only two of them are actually tasks

Hi @victor-gonzalez The message is "Name of a workflow file must match the name of the main struct in it." It does not say "same as". That means that the naming conventions for `struct` still apply and it should comply with the `UpperCamelCase` style.

victor-gonzalez:
> > Hi @vkucera, I've modified the `dptdptEfficiencyAndQc.cxx` source file to fix all `linter` errors/warnings but I reach a contradictory (false positive?) situation. The check of the `CMakeLists.txt` suggests to use `dptdptEfficiencyAndQc` as file name but if it is used we get the error that the main `struct` should have the same name but at the same time the name of the struct must be `UpperCamelCase` BTW, the source file has several `struct` structures on it but only two of them are actually tasks
> 
> Hi @victor-gonzalez The message is "Name of a workflow file must match the name of the main struct in it." It does not say "same as". That means that the naming conventions for `struct` still apply and it should comply with the `UpperCamelCase` style.

@vkucera But then, I don't understand why we are getting that message 
The workflow file name is `dptdptEfficiencyAndQc` and the main `struct` in the workflow file is `DptDptEfficiencyAndQc` 
Or, how is the main `struct` decided when there are a few?

vkucera:
> `DptDptEfficiencyAndQc`

`DptDptEfficiencyAndQc` is not an `UpperCamelCase` counterpart to `dptdptEfficiencyAndQc` so the linter is correct about the mismatch.

victor-gonzalez:
> > `DptDptEfficiencyAndQc`
> 
> `DptDptEfficiencyAndQc` is not an `UpperCamelCase` counterpart to `dptdptEfficiencyAndQc` so the linter is correct about the mismatch.

I see, thanks!

vkucera:
> `dptdptEfficiencyAndQc`

You have to either call the file `dptDptEfficiencyAndQc.cxx` or the `struct` `DptdptEfficiencyAndQc`. I think the former is better.

victor-gonzalez:
> > `dptdptEfficiencyAndQc`
> 
> You have to either call the file `dptDptEfficiencyAndQc.cxx` or the `struct` `DptdptEfficiencyAndQc`. I think the former is better.

Agree!

PR #8787, opened 2024-12-03
[PWGLF] occupancy checks for MC

sarjeetagami:
PID and occupancy checks for MC


vkucera:
> @sarjeetagami Consider using the `RecoDecay` methods for all your momentum vector operations. It will simplify your code very significantly. Also get the momenta directly from the tables with `.pVector()` if available.

PR #8815, opened 2024-12-04
[PWGJE,EMCAL-670] Add LG cell time correction parameters

nstrangm:
- Add parameters for MC broadening for LG and HG cells, as well as data shift for LG and HG cells
- Made the EMCalCorrectionTask o2linter compliant


fjonasALICE (approved):
Good to got and locally tested. Thanks for making O2linter happy!

PR #8827, opened 2024-12-05
[PWGCF] Disable IsGlobalTrackSDD for the purposes of checking

JiangTao-thu:
Disable the track selection `IsGlobalTrackSDD`, to check if it is the cause of the missing statistics in the high multiplicity range of the multiplicity distribution in pp.

### DIFF
--- a/PWGCF/Flow/Tasks/FlowRunbyRun.cxx
+++ b/PWGCF/Flow/Tasks/FlowRunbyRun.cxx
@@ -72,7 +72,7 @@ struct FlowRunbyRun {
   ConfigurableAxis axisIndependent{"axisIndependent", {VARIABLE_WIDTH, 0, 5, 10, 20, 30, 40, 50, 60, 70, 80, 90}, "X axis for histograms"};
 
   Filter collisionFilter = nabs(aod::collision::posZ) < cfgCutVertex;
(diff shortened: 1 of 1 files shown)


JiangTao-thu:
o2-linter: disable=const-ref-in-for-loop, const-ref-in-for-loop, name/function-variable, name/type, name/file-cpp, name/configurable

victor-gonzalez (changes requested):
Please, check the details of the `linter` errors with its notations in the `Files changed` tab
At least the `const-ref-in-for-loop` are mandatory right now because they have an impact in the execution
For the rest I would suggest to implement them in a timely manner with the next iterations. They are really helpful for increasing the readability of the code which increases its maintainability and easy evolution

BTW, you are making an `or` of global tracks and `isGlobalTrackSDD` which seems correct, if you remove the latest one probably you will get a slightly lower number of tracks, as I said it is only marginal (you did not mention the global tracks selection in your e-mail)
I would suggest to check the number of tracks before and after selection and the effect of each of the cuts you set in the filter

JiangTao-thu:
Hello Victor, you are right, and I apologize for forgetting to mention the global tracks selection. I have also decided to use your method to check.  Additionally, the owner of this code, Zhiyong, informed me that there are some bugs that need to be fixed in this code. However, I am currently facing some issues with the local compilation of O2Physics, which makes it inconvenient for me to test the modified code locally. After our discussion, I have decided to temporarily close this pull request and will resubmit it after making the necessary modifications. Thank you again for your suggestions.

PR #8834, opened 2024-12-05
[PWGCF] FemtoUniverse -- Add minus-plus combination to track-nucleus task.

a-m-andrushko:
Add minus-plus combination building to femtoUniversePairTaskTrackNucleus.cxx.


a-m-andrushko:
@victor-gonzalez, is it possible to move forward with this PR considering warnings given by O2-linter?

victor-gonzalez (changes requested):
Remove 
```
#include "TDatabasePDG.h"
```
it is not being used and could impact the memory layout. The PDG database O2 service is in use as recommended by the `linter`

Regarding the `root-entity` `linter` suggestions, I would suggest to move them to the standard versions (`std::` prefixed) in a timely manner in future iterations. ROOT in most of the cases rely on the standard implementation so the use of the ROOT version implies an additional step in the function call path

victor-gonzalez (approved):

PR #8838, opened 2024-12-05
[PWGCF] Addition of nSigma PID function

Preet-Bhanjan:
Added the function to provide PIDs for tracks using the TPC and TOF nSigma


PWGCF/Flow/Tasks/FlowPbPbpikp.cxx
```diff
@@ -169,6 +170,38 @@ struct GfwPidflow {
...
 
+  enum Particles {
+    PIONS,
+    KAONS,
+    PROTONS
+  };
+
+  enum Particles pion = PIONS;
+  enum Particles kaon = KAONS;
+  enum Particles proton = PROTONS;
```
victor-gonzalez, line 181: These variables are not actually needed 
The comparisons can be done directly with the `enum` named constants

PWGCF/Flow/Tasks/FlowPbPbpikp.cxx
```diff
@@ -191,7 +224,7 @@ struct GfwPidflow {
     int maxProb[3] = {80, 80, 80};
     int pidID = -1;
     std::pair<int, int> idprob = GetBayesID(track);
-    if (idprob.first == 0 || idprob.first == 1 || idprob.first == 2) { // 0 = pion, 1 = kaon, 2 = proton
+    if (idprob.first == pion || idprob.first == kaon || idprob.first == proton) { // 0 = pion, 1 = kaon, 2 = proton
```
victor-gonzalez, line 227: See previous comment

PWGCF/Flow/Tasks/FlowPbPbpikp.cxx
```diff
@@ -277,14 +310,15 @@ struct GfwPidflow {
       histos.fill(HIST("hEta"), track.eta());
       histos.fill(HIST("hPt"), pt);
 
-      histos.fill(HIST("TofTpcNsigma"), 0, track.tpcNSigmaPi(), track.tofNSigmaPi());
-      histos.fill(HIST("TofTpcNsigma"), 1, track.tpcNSigmaKa(), track.tofNSigmaKa());
-      histos.fill(HIST("TofTpcNsigma"), 2, track.tpcNSigmaPr(), track.tofNSigmaPr());
+      histos.fill(HIST("TofTpcNsigma"), pion, track.tpcNSigmaPi(), track.tofNSigmaPi(), pt);
+      histos.fill(HIST("TofTpcNsigma"), kaon, track.tpcNSigmaKa(), track.tofNSigmaKa(), pt);
+      histos.fill(HIST("TofTpcNsigma"), proton, track.tpcNSigmaPr(), track.tofNSigmaPr(), pt);
```
victor-gonzalez, line 315: See the above comments

victor-gonzalez (changes requested):
Have a look at my comments for the next iterations
You should start implementing `linter` suggestions, that you can see in the `Files changed` tab, in a timely manner. They are intended to increase the readability of the code which impacts in its maintenance and also its evolution

But most important, `linter` suggestions/errors catch error situations which are really difficult to identify with other means
In here we have one of them in the line L.231 of the code. The usage of `abs` in here is wrong. `abs` returns an integer which means that the conditional is almost useless. You should move to either `std::abs` or to `std::fabs`, the latest one being more explicit, as suggested by the `linter`

Perhaps it is a good time to unify the usage of asking for the absolute value along the whole source file. There are still a few occurrences of `TMath::Abs` which should migrate to `std::abs` or `std::fabs` as suggested by the `linter`

PWGCF/Flow/Tasks/FlowPbPbpikp.cxx
```diff
@@ -8,11 +8,13 @@
 // In applying this license CERN does not waive the privileges and immunities
 // granted to it by virtue of its status as an Intergovernmental Organization
 // or submit itself to any jurisdiction.
+/// \file FlowPbPbpikp.cxx
```
victor-gonzalez, line 11: Leave an empty line above the `/// \file` directive

PWGCF/Flow/Tasks/FlowPbPbpikp.cxx
```diff
@@ -106,28 +110,28 @@ struct GfwPidflow {
...
     int nPtBins = axis.binEdges.size() - 1;
-    double* PtBins = &(axis.binEdges)[0];
-    fPtAxis = new TAxis(nPtBins, PtBins);
+    double* ptBins = &(axis.binEdges)[0];
+    fPtAxis = new TAxis(nPtBins, ptBins);
 
     TObjArray* oba = new TObjArray();
     oba->Add(new TNamed("Ch08Gap22", "Ch08Gap22"));
-    for (Int_t i = 0; i < fPtAxis->GetNbins(); i++)
+    for (std::int i = 0; i < fPtAxis->GetNbins(); i++)
```
victor-gonzalez, line 123: The `std::` prefix is not needed for standard types as `int`

PWGCF/Flow/Tasks/flowPbPbpikp.cxx
```diff
@@ -8,6 +8,7 @@
 // In applying this license CERN does not waive the privileges and immunities
 // granted to it by virtue of its status as an Intergovernmental Organization
 // or submit itself to any jurisdiction.
+
 /// \file FlowPbPbpikp.cxx
```
victor-gonzalez, line 12: As the filename was changed use the new one in here

victor-gonzalez: Also, the `lint` warning referring to the `name/workflow-file` is due to that it is expected that the name of the main task within the file should match the name of the file so you might use 
```
struct FlowPbPbpikp
```
for your task in the new L.61

victor-gonzalez (approved):
Consider may comments for future iterations

victor-gonzalez:
Please, never push changes that have not been previously compiled and tested locally

victor-gonzalez:
You have to reflect in `CMakeLists.txt` the change in source filename

Preet-Bhanjan:
As suggested by O2 linter, I had to change the function name 'FillProfile' to 'fillProfile'. Now, this gives an error since for GFW flow container the function is defined as 'FillProfile'.

victor-gonzalez:
Well, you were suggested to change the name of 'your' function 
If you are calling a function from any other place of the O2/O2Physics framework you have to use the name of that function as it is, if not your function call will not be recognized

Preet-Bhanjan:
All of the linter issues regarding my file have been fixed. Please suggest any changes to resolve the rest of the issues.

victor-gonzalez (approved):
Thanks!
The other `linter` errors have to be fixed by other analyzers from the `Flow` functional area

PR #8854, opened 2024-12-07
[PWGLF] fill also zero in gen histogram

maciacco:


alibuild (approved):

PR #8859, opened 2024-12-08
[PWGCF] additions to IV, validations of banishment loop, support for ES, etc.

abilandz:


PWGCF/MultiparticleCorrelations/Core/MuPa-MemberFunctions.h
```diff
@@ -2485,8 +2525,8 @@ void BookQAHistograms()
       for (Int_t ba = 0; ba < 2; ba++) // before/after cuts
       {
         qa.fQAParticleHistograms2D[t][rs][ba] = new TH2F(
-          Form("fQAParticleHistograms2D[%s][%s][%s]", qa.fParticleHistogramsName2D[t].Data(), srs[rs].Data(), sba[ba].Data()),
-          Form("%s, %s, %s", "__RUN_NUMBER__", srs_long[rs].Data(), sba_long[ba].Data()), // __RUN_NUMBER__ is handled in PropagateRunNumber(...)
+          Form("fQAParticleHistograms2D[%s][%s][%s]", qa.fParticleHistogramsName2D[t].Data(), cc.srs[rs].Data(), cc.sba[ba].Data()),
+          Form("%s, %s, %s", "__RUN_NUMBER__", cc.srs_long[rs].Data(), cc.sba_long[ba].Data()), // __RUN_NUMBER__ is handled in PropagateRunNumber(...)
```
victor-gonzalez, line 2529: I would be careful with using `Form`
It has a limited, although relatively large, circular buffer which means that under certain conditions its content maybe overwritten
It gave me bad, really bad (hard to fix bugs), experiences in the past
Since then I always use `TString::Format`

PWGCF/MultiparticleCorrelations/Core/MuPa-MemberFunctions.h
```diff
@@ -2826,35 +2856,31 @@ void BookParticleCutsHistograms()
...
   for (Int_t rs = 0; rs < 2; rs++) // reco/sim
   {
 
     if (Skip(rs)) {
       continue;
     }
 
-    for (Int_t cc = 0; cc < eCutCounter_N; cc++) // enum eCutCounter
+    for (Int_t cuco = 0; cuco < eCutCounter_N; cuco++) // cut counter. I have to use "cuco", because "cc." is already reserved for struct "common cosmetics"
     {
```
victor-gonzalez, line 2868: > I have to use "cuco", because "cc." is already reserved

Don't limit yourself and be more readable! Suggestion: `ixCutCounter`

victor-gonzalez (approved):

PWGCF/MultiparticleCorrelations/Core/MuPa-MemberFunctions.h
abilandz: OK, thanks, noted! Form and TString::Format() have different return type, which will trigger additional modifications, but I will move gradually towards TString::Format().

PWGCF/MultiparticleCorrelations/Core/MuPa-MemberFunctions.h
abilandz: I use this one as an array index, so its name preferably has to stay short. But I will re-think the whole thing

PR #8881, opened 2024-12-09
[PWGLF] feat: Add new features to the NonPromptCascadeTask

mpuccio:
This commit introduces several new features to the NonPromptCascadeTask:

1. Calculates the delta PT between the cascade and the ITS track, and stores it in the `DeltaPtITSCascade` column of the `NPCascTable`.
2. Adds a new column `HasReassociatedCluster` to the `NPCascTable` to indicate if the ITS track has reassociated clusters.
3. Modifies the `cpa` calculation to use the cascade momentum instead of a separate array.
4. Adds the `deltaPt` and `hasReassociatedCluster` fields to the `NPCascCandidate` struct and the corresponding `emplace_back` calls.

These changes aim to provide more detailed information about the non-prompt cascade candidates, which can be useful for further analysis and selection.

### DIFF
--- a/PWGLF/DataModel/LFNonPromptCascadeTables.h
+++ b/PWGLF/DataModel/LFNonPromptCascadeTables.h
@@ -25,7 +25,9 @@ namespace o2::aod
 namespace NPCascadeTable
 {
 DECLARE_SOA_COLUMN(MatchingChi2, matchingChi2, float);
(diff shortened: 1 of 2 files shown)


skundu692 (approved):

vkucera:
@mpuccio Please wait a few seconds for the bot to fix the PR title before you enable auto-merge. GitHub uses the title at the moment auto-merge was enabled and does not consider any later changes.

PR #8883, opened 2024-12-09
[PWGJE,EMCAL-670] Bugfix in EMCalCorrectionTask for ShaperCorrection

nstrangm:
- Shaper correction should only be applied for LG cells (oversight in O2Physics implementation), see 
https://github.com/alisw/AliPhysics/blob/master/PWG/EMCAL/EMCALbase/AliEMCALRecoUtils.cxx#L458


mhemmer-cern (approved):
Good from EMCal side

PR #8886, opened 2024-12-09
[Common] Add MC analysis function (check if correct or wrong matching)

syano0822:
Added new workflow of "match-mft-mch-data-mc.cxx".
This is the same as match-mft-mch-data.cxx but it can handle the MC information.


syano0822:
I don't understand the conflict because the conflict is between the PR and master branch. RP is basically changed from master branch. So all PR is different from the master brancha and never happens between PR and master branch.

PR #8891, opened 2024-12-09
[PWGHF] Add D0 D0bar reflection infos tagged from data in D0 task

zhangbiao-phy:
**It's good to know how large the reflection contribution in the data**
- Add D0 D0bar reflection column into the D0 candidate selection table
- Flii D0 D0bar reflection into the D0 task ThnSparse of Data

### DIFF
--- a/PWGHF/D2H/Tasks/taskD0.cxx
+++ b/PWGHF/D2H/Tasks/taskD0.cxx
@@ -337,7 +337,6 @@ struct HfTaskD0 {
       } else {
         if (candidate.isSelD0() >= selectionFlagD0) {
           registry.fill(HIST("hMassVsPtVsPtBVsYVsOriginVsD0Type"), massD0, ptCandidate, hfHelper.yD0(candidate), SigD0);
-
           if (candidate.isSelD0bar()) {
             registry.fill(HIST("hMassVsPtVsPtBVsYVsOriginVsD0Type"), massD0, ptCandidate, hfHelper.yD0(candidate), ReflectedD0);
           } else if (!candidate.isSelD0bar()) {
--- a/PWGHF/D2H/Tasks/taskD0.cxx
+++ b/PWGHF/D2H/Tasks/taskD0.cxx
@@ -35,12 +35,12 @@ using namespace o2::framework::expressions;
 namespace
 {
 enum CandTypeSel {
-  SigD0 = 0,     // Signal D0
-  SigD0bar,      // Signal D0bar
-  PureSigD0,     // Signal D0 without Reflected D0bar
-  PureSigD0bar,  // Signal D0bar without Reflected D0
-  ReflectedD0,   // Reflected D0
-  ReflectedD0bar // Reflected D0bar
+  SigD0 = 0,      // Signal D0
+  SigD0bar,       // Signal D0bar
+  ReflectedD0,    // Reflected D0
+  ReflectedD0bar, // Reflected D0bar
+  PureSigD0,      // Signal D0 without Reflected D0bar
+  PureSigD0bar    // Signal D0bar without Reflected D0
 };
 } // namespace
 struct HfTaskD0 {
--- a/PWGHF/D2H/Tasks/taskD0.cxx
+++ b/PWGHF/D2H/Tasks/taskD0.cxx
@@ -318,16 +318,24 @@ struct HfTaskD0 {
       if constexpr (applyMl) {
         if (candidate.isSelD0() >= selectionFlagD0) {
           registry.fill(HIST("hBdtScoreVsMassVsPtVsPtBVsYVsOriginVsD0Type"), candidate.mlProbD0()[0], candidate.mlProbD0()[1], candidate.mlProbD0()[2], massD0, ptCandidate, hfHelper.yD0(candidate), SigD0);
+          if (candidate.isSelD0Refl())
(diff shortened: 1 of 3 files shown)


vkucera:
Hi @zhangbiao-phy , can you please elaborate on the rationale and logic of your changes?
I don't understand why you cannot use the existing flags.

zhangbiao-phy:
> Hi @zhangbiao-phy , can you please elaborate on the rationale and logic of your changes? I don't understand why you cannot use the existing flags.

Hi @vkucera, I would like to directly estimate the D0 reflection in data. Indeed it can be done in the D0 task by `candidate.isSelD0bar()  && candidate.isSelD0()`, but it can not be separated to D0 and D0 bar reflection. So I implement it into the selector. Let me know if you have better idea for this. Thanks a  lot!

vkucera:
> > Hi @zhangbiao-phy , can you please elaborate on the rationale and logic of your changes? I don't understand why you cannot use the existing flags.
> 
> Hi @vkucera, I would like to directly estimate the D0 reflection in data. Indeed, it can be done in the D0 task by `candidate.isSelD0bar() && candidate.isSelD0()`, but it can not be separated to D0 and D0 bar reflection. So I implement it into the selector. Let me know if you have a better idea for this. Thanks a lot!

How would you identify a reflection in data?

zhangbiao-phy:
> > > Hi @zhangbiao-phy , can you please elaborate on the rationale and logic of your changes? I don't understand why you cannot use the existing flags.
> > 
> > 
> > Hi @vkucera, I would like to directly estimate the D0 reflection in data. Indeed, it can be done in the D0 task by `candidate.isSelD0bar() && candidate.isSelD0()`, but it can not be separated to D0 and D0 bar reflection. So I implement it into the selector. Let me know if you have a better idea for this. Thanks a lot!
> 
> How would you identify a reflection in real data?

The candidates be identified as both D0 and D0 bar, It also be saved in some run2 AliPhysics D0 tasks. I attached one plot from local test based on data.
image

vkucera:
I'm sorry, I'm not getting it.
A D0 reflection is an actual D0 meson whose invariant mass under the swapped daughter mass hypotheses is within the analysis mass range. This cannot be decided with certainty in real data. If it was possible, we would reject all reflections from the selection. The degree of certainty that a candidate is a reflection is equal to the level of certainty that a candidate is a D0, which is quantified in the selection flag. I don't see how any combination of other flags can give you any additional information.

fcatalan92:
@zhangbiao-phy Also to me it is not clear how your changes to the selector add any additional information to what is already available. I agree with Vit, by definition we cannot disentangle reflections in data.

zhangbiao-phy:
> @zhangbiao-phy Also to me it is not clear how your changes to the selector add any additional information to what is already available. I agree with Vit, by definition we cannot disentangle reflections in data.

Hi @vkucera and @fcatalan92, Thanks for the discussion! We can call that "reflection candidates" instead of "reflection D0 or D0 bar" since it includes signal and bkg. Tagging reflections in real data by requiring the overlap between D0 and D0bar selections is a valid approach to identifying reflection candidates. For some D0 charge-dependent analysis, it will be helpful to understand the reflection contribution and the response in data and bkg. Reflection studies can indeed shed light on the systematic effects and detector responses (PID) that might otherwise bias such measurements. Indeed, further analysis is necessary to separate the true signal from background contributions within this tagged sample. This ensures accurate signal extraction and proper treatment of reflections in the physics analysis. What do you think?

fcatalan92:
@zhangbiao-phy I see your motivation for performing these checks.

> Tagging reflections in real data by requiring the overlap between D0 and D0bar selections is a valid approach to identifying reflection candidates
On this I agree, and it corresponds to requiring `candidate.isSelD0bar() && candidate.isSelD0()` in the task. I do not think what you are doing in the selector now can provide more information than this, which is already available. Is there a difference?

fgrosa:
@vkucera @fcatalan92 indeed the reflected-over-signal can be determined in data by extracting separately the D0 signal for candidates that are compatible with one hypothesis, the other, or both (see e.g. slide 6 https://indico.cern.ch/event/1043402/contributions/4394133/attachments/2258188/3832203/promptD0_update_2018PbPb.pdf), but I agree that the information provided by `candidate.isSelD0bar()` and `candidate.isSelD0()` should be sufficient.

zhangbiao-phy:
> @zhangbiao-phy I see your motivation for performing these checks.
> 
> > Tagging reflections in real data by requiring the overlap between D0 and D0bar selections is a valid approach to identifying reflection candidates
> 
> On this I agree, and it corresponds to requiring `candidate.isSelD0bar() && candidate.isSelD0()` in the task. I do not think what you are doing in the selector now can provide more information than this, which is already available. Is there a difference?
hi @fcatalan92, there are some differences in the selector we have the function `selectionTopolConjugate`, here we can do the selection for D0 and D0bar. Then, if you tag the reflection in the D0 task, it already merged for D0 D0bar reflection, So the difference between D0 with D0bar reflection only is the Inv. mass hypothesis. But I think even we do the different topo selection, it should be always reflected since the PID not perfect always. So I will do that in the task only this time.

zhangbiao-phy:
> @vkucera @fcatalan92 indeed the reflected-over-signal can be determined in data by extracting separately the D0 signal for candidates that are compatible with one hypothesis, the other, or both (see e.g. slide 6 https://indico.cern.ch/event/1043402/contributions/4394133/attachments/2258188/3832203/promptD0_update_2018PbPb.pdf), but I agree that the information provided by `candidate.isSelD0bar()` and `candidate.isSelD0()` should be sufficient.

Hi @vkucera @fcatalan92 @fgrosa all, after the discussion, I implemented this only in the D0 task but added two types for pure D0 and pure D0 bar for the checks. You can see the plot below, they are third and fourth-bin content. Let me know if you have any comments and suggestions! Thanks a lot!
image

fcatalan92 (approved):

PR #8892, opened 2024-12-09
[PWGHF] PID inclusion for associated particles in Lc correlation

1994ra:


vkucera:
@1994ra Your PR title is too vague as it has no information about where the changes were made. Please add it.

vkucera:
The `[pi-multiple-fraction]` warning by O2 linter is a false positive which can be ignored for now.

apalasciano (approved):
Dear Ravindra, thanks for the modification.
To me it looks fine, I'm approving the PR. 
In future, together with other analysers using these tasks, we should modify a bit them to avoid large code duplication. This can be done at a later stage to not interfere with QM approvals

PR #8898, opened 2024-12-10
[PWGDQ]: Updates for X(3872) analysis

zjxiongOvO:


PWGDQ/Core/VarManager.h
```diff
@@ -4556,6 +4558,40 @@ void VarManager::FillDileptonTrackTrack(T1 const& dilepton, T2 const& hadron1, T
   }
 }
 
+//__________________________________________________________________
+template <int candidateType, typename T1, typename T2, typename T3>
+void VarManager::FillQaudMC(T1 const& dilepton, T2 const& track1, T3 const& track2, float* values)
```
iarsene, line 4563: T2 and T3 types are the same, so use just one in the template definition

zjxiongOvO: Thanks for comment. I already remove it.

iarsene (approved):

PR #8903, opened 2024-12-10
[PWGDQ] fix configQA option in analysis-event-selection

iarsene:


alibuild (approved):

PR #8927, opened 2024-12-11
[PWGCF] Changed from THnSparseD to THnSparseF because of memory issue

Thorkj:


Thorkj:
I have no issue building locally

victor-gonzalez (approved):

PR #8938, opened 2024-12-11
[PWGCF] Added PID, Added RBR NUA

EmilGorm:
Updated task:
 - Handles PID flow based on TPC+TOF Nsigma cuts
 - Added run-by-run weights and run-averaged weights capabilities
 - Weight filling now no longer performs the rest of the analysis
 - Weights stored in TList so it can hold PID weights + RBR weights
 - Updated reading of weights from ccdb

### DIFF
--- a/PWGCF/GenericFramework/Core/FlowPtContainer.cxx
+++ b/PWGCF/GenericFramework/Core/FlowPtContainer.cxx
@@ -21,10 +21,11 @@ FlowPtContainer::FlowPtContainer() : TNamed("name", "name"),
                                      fillCounter(0),
                                      fEventWeight(kEventWeight::kUnity),
                                      fUseCentralMoments(true),
+                                     fUseGap(false),
                                      sumP(),
                                      corrNum(),
                                      corrDen(),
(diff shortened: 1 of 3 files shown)


EmilGorm:
Right, I recall being asked to add the prefix before. Good to know the bot handles it now.

victor-gonzalez (approved):
Hi Emil!
There are already a few users of the GFW. Will these changes affect the previous behavior or is just new functionality?
In any case, please, consider presenting the changed in a ten minutes slot in one of the CF towards o2 meetings

For the next iterations, please, consider `linter` suggestions. They are really good to increase code readability and also unveil tricky to catch errors. Also consider moving towards using named constants from `TPDGCode.h` instead of numbers

EmilGorm:
Thanks, Victor; I will keep in mind the O2 linter suggestions for next time.
The changes are mainly to the GFW analysis task, so there are no new functionalities for others to consider, and it will not change the behavior of existing analyses.
However, I would be open to a short discussion in the CF towards O2 on Monday if we have a meeting.

PR #8942, opened 2024-12-11
[PWGEM/PhotonMeson] Fix Cluster-Collision assignment in emcQC task

nstrangm:
- In EMCalQC task: Change SkimEMCCluster to MyEMCCluster, containing the EMEventId
- Template "fillClusterHistograms" function to allow for this different cluster class


mhemmer-cern (approved):

vkucera:
The `[std-prefix]` warning for `abs` is reporting a bug.

PR #8957, opened 2024-12-12
[PWGEM/PhotonMeson] Fix abs in emcalQC task

nstrangm:
- Implemented bugfix for "abs-bug" reported by Vit via the o2linter (https://github.com/AliceO2Group/O2Physics/pull/8942)

### DIFF
--- a/PWGEM/PhotonMeson/Tasks/emcalQC.cxx
+++ b/PWGEM/PhotonMeson/Tasks/emcalQC.cxx
@@ -169,7 +169,7 @@ struct EmcalQC {
       fRegistry.fill(HIST("Event/hEMCCollisionCounter"), 1);
       if (collision.selection_bit(o2::aod::evsel::kIsTriggerTVX)) {
         fRegistry.fill(HIST("Event/hEMCCollisionCounter"), 2);
-        if (collision.posZ() < eventcuts.cfgZvtxMax && collision.posZ() > -eventcuts.cfgZvtxMax) {
+        if (std::abs(collision.posZ()) < eventcuts.cfgZvtxMax) {
           fRegistry.fill(HIST("Event/hEMCCollisionCounter"), 3);
           if (collision.sel8()) {
             fRegistry.fill(HIST("Event/hEMCCollisionCounter"), 4);
--- a/PWGEM/PhotonMeson/Tasks/emcalQC.cxx
+++ b/PWGEM/PhotonMeson/Tasks/emcalQC.cxx
@@ -9,10 +9,14 @@
 // granted to it by virtue of its status as an Intergovernmental Organization
 // or submit itself to any jurisdiction.
 //
-// ========================
-//
-// This code runs loop over EMCal clusters for EMCal QC.
-//    Please write to: nicolas.strangmann@cern.ch
+/// EMCAL QC Task
+///
+/// \file emcalQC.cxx
+///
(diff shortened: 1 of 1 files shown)


PWGEM/PhotonMeson/Tasks/emcalQC.cxx
```diff
@@ -156,16 +160,16 @@ struct emcalQC {
...
-      if (eventcuts.onlyKeepWeightedEvents && fabs(collision.weight() - 1.) < 1E-10) {
+      if (eventcuts.onlyKeepWeightedEvents && std::fabs(collision.weight() - 1.) < 1E-10) {
         continue;
       }
 
       fRegistry.fill(HIST("Event/hEMCCollisionCounter"), 1);
       if (collision.selection_bit(o2::aod::evsel::kIsTriggerTVX)) {
         fRegistry.fill(HIST("Event/hEMCCollisionCounter"), 2);
-        if (abs(collision.posZ()) < eventcuts.cfgZvtxMax) {
+        if (collision.posZ() < eventcuts.cfgZvtxMax && collision.posZ() > -eventcuts.cfgZvtxMax) {
```
vkucera, line 172: Why did you remove the `abs`?

vkucera (changes requested):

PWGEM/PhotonMeson/Tasks/emcalQC.cxx
nstrangm: I got a warning about the abs by itself and then checked out how it was done in the PWGEM/Dilepton/Core/EMEventCut.h and wanted to make it more similar to the way it was implemented there

vkucera: But `EMEventCut` has `mMinZvtx` and `mMaxZvtx`. You only have `cfgZvtxMax`.

nstrangm: Yes, however, I don't plan on ever using an asymmetric z vertex cut.
I don't really understand the point of this change request. Do you not like the style of my if statement, or would you prefer EMCal QC to be done in an asymmetric z vertex region, or what is blocking this PR?

vkucera: I understood the subject of this PR was to fix the abs bug. What you did instead is to change the condition to support asymmetric selection without actually using it.

nstrangm: Sorry, maybe I don't fully understand: Does this not implicitly solve the issue?

vkucera: It's not clear to me why you would prefer
```c++
collision.posZ() < eventcuts.cfgZvtxMax && collision.posZ() > -eventcuts.cfgZvtxMax
```
instead of
```c++
std::abs(collision.posZ()) < eventcuts.cfgZvtxMax
```
if don't use and will never use an asymmetric cut.

nstrangm: To be honest, I do not have a preference at all. But I am happy to comply with the request

vkucera: Thanks

vkucera (approved):

mhemmer-cern (approved):

PR #8965, opened 2024-12-12
[PWGJE] Add histograms for data-driven methods

hahassan7:
Adding a THnSparse with jetpT score, jet mass, SV mass, and SVfE histograms wich will be used for data-driven approaches for efficiency and purity estimation.

### DIFF
--- a/PWGJE/Tasks/bjetTaggingML.cxx
+++ b/PWGJE/Tasks/bjetTaggingML.cxx
@@ -43,7 +43,7 @@ using namespace o2::framework::expressions;
 
 struct BJetTaggingML {
 
-  struct bjetParams {
+  struct BJetParams {
     float mJetpT = 0.0;
     float mJetEta = 0.0;
     float mJetPhi = 0.0;
@@ -52,7 +52,7 @@ struct BJetTaggingML {
     float mJetMass = 0.0;
   };
 
-  struct bjetTrackParams {
+  struct BJetTrackParams {
     double mTrackpT = 0.0;
     double mTrackEta = 0.0;
     double mDotProdTrackJet = 0.0;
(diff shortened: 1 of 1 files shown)


nzardosh:
Hi Hadi, why not use a table?

nzardosh:
have you checked how large they would be? it seems you already have most of this information in trees anyways?

hahassan7:
The Trees are in the TreeCreator, there are no trees here. I also think THnSparse since it has binnings it will be much smaller than TTrees.

nzardosh:
ok fair enough

hahassan7:
Yes, I totally agree with this. It will make tagging much easier with just taking the score from the extended jet table. I think we edit the one from Hanseo and add one score colunm to it and a function that evaluates the model and gives the score it gives. Though Hanseo's code needs so much cleaning.

nzardosh:
Ok that sounds great then, lets go in that direction :) Yeah I agree the code needs alot of cleaning. I think this would be a good thing to spend a week or so on since the code needs to be used in many workflows. I had discussed this at some point also with Hanseo that for Hard probes I approved alot of code changes that were urgent but with the understanding that they would be tidied up after so its a good moment perhaps to do this.

vkucera:
@nzardosh @hahassan7 If it's fine with you, I'm converting this to a draft to avoid unnecessary recompilations since this is still in progress.

hahassan7:
Hi @vkucera @nzardosh , this is done. There's nothing more to be done here. It should be merged.

nzardosh:
Hi Hadi,

I think we decided on changing the strategy right? to have the output such that the ML score would be joinable to the the jet table instead of how it is done here.
I think it would be useful if we could all meet before the christmas holidays to discuss this HF Jet part of the framework a bit. I will ping you all on mattermost

PR #8968, opened 2024-12-12
[PWGUD] Added Rho prime resonance analysis and Angular distributions

ananthapadmanabhan18:


ananthapadmanabhan18:
This Code compiles successfully. This is for the analysis of the rho' meson which decays to 4pions. The code aims to measure the resonance mass and the pair angular distributions of the pions.

ananthapadmanabhan18:
Edits req

PR #8969, opened 2024-12-12
[PWGUD] Rho' resonance analysis and 4-pi pair angular distributions

ananthapadmanabhan18:


abylinkin (changes requested):
Few comments: 1. Usual Eta cut for tracks is 0.9 not 0.5. 2. With this cut applied before track selections you are removing good events with fake tracks. Probably one can cut at value <10 or so, but it had to be justified and/or studied. For example, if we refer to the 2015 paper on rho' the selection was exactly 4 GOOD tracks - we didn't remove the event if it had bad ones additionally. "    if (collision.numContrib() != 4) {
      return;
    }" 3. I believe one can get Pb mass (and probably snn) from the o2 utils without hardcoding the value.

PR #8980, opened 2024-12-13
[Common] add extra MFT counters to mult table

ddobrigk:


alibuild (approved):

PR #8994, opened 2024-12-14
[PWGUD] Resonance analysis of rho' and 4-pi pair angular distributions.

ananthapadmanabhan18:
Considering your requested changes:
<ol>
<li> I have changed the $\eta$ cut to 0.9. (it was a mistake)
<li> I have made a bool such that of its true, it will reject the event if it has fake tracks or bad tracks even if it has 4 pion good tracks. If its set to false, it will select those events. The reason that i have created it is that, the expected results as in the paper is not obtained if I don't use strict event selection.  I will push it as soon as I find it
<li> I could not find the inbuilt method of getting the Pb208 Mass. All the other codes present in the repo has used the same method as mine. Same for the beam energy.
</ol>

The below showed image is the event  $p_T$ distribution obtained without rejecting events with bad tracks:
Untitled

The below showed image is the event  $p_T$ distribution obtained with rejecting events with bad tracks (That is events with exactly 4 good quality pion tracks only):
Untitled

## **The Second one is the expected result for the coherent photoproduction of the $\rho'$ and its decay to $\pi^+\pi^-\pi^+\pi^-$**
Since there is a switch that controls which one to use, I can switch from one to another easily.


abylinkin:
Thank you for the plot! 
What I wanted to say is that you will likely reject events with pt>200 MeV later when selecting coherent events. 
If you look at the coherent peak (2nd bin for example) before the NumContrib cut you have 40 events, after - 28. I doubt that all of the difference is the background. So that is great that you have a switch, to select one or another option.

I am approving this PR.
For the future I might think of saving NumContrib to the output or passing the cut value as a configurable for more detailed studies.

abylinkin (approved):

ananthapadmanabhan18:
Thank you so much!. Yes, I will be putting a cut around 200MeV. I only thought about the $p_T$ distribution looking same :)
Never thought about getting more signal after the cut!. There are lot of stuffs to be done!
Cheers
Anantha

PR #8997, opened 2024-12-14
[PWGHF] Ds-h correlations, ME offline on self-contained derived data

scattaru:
Dear all,
With this commit I add the last part of the code for performing ME offline, I have tested it and adding the Preslice it works.
I have also introduced two different process functions for  producing self-contained derived data as a test, since I have noticed that using the two different functions locally I obtain two different results.

Let me know for any comment,
Cheers,
Samuele

### DIFF
--- a/PWGHF/HFC/TableProducer/correlatorDsHadrons.cxx
+++ b/PWGHF/HFC/TableProducer/correlatorDsHadrons.cxx
@@ -61,8 +61,8 @@ struct HfCorrelatorDsHadronsSelCollision {
   Configurable<float> yCandMax{"yCandMax", 0.8, "max. cand. rapidity"};
   Configurable<float> ptCandMin{"ptCandMin", 1., "min. cand. pT"};
 
-  SliceCache cache;
   HfHelper hfHelper;
+  SliceCache cache;
 
   using SelCollisions = soa::Join<aod::Collisions, aod::EvSels>;
   using CandDsData = soa::Filtered<soa::Join<aod::HfCand3Prong, aod::HfSelDsToKKPi>>;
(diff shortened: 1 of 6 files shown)


vkucera:
@scattaru Please fix the O2 linter errors.

PWGHF/HFC/TableProducer/correlatorDsHadrons.cxx
```diff
@@ -689,32 +692,77 @@ struct HfCorrelatorDsHadrons {
   }
   PROCESS_SWITCH(HfCorrelatorDsHadrons, processMcGen, "Process MC Gen mode", false);
 
-  void processDerivedDataDs(SelCollisionsWithDs::iterator const& collision,
+  int hfcReducedCollisionIndex = 0;
```
vkucera, line 695: Don't declare data members between methods. It makes the code structure hard to navigate.

vkucera (changes requested):

PWGHF/HFC/TableProducer/correlatorDsHadrons.cxx
scattaru: Hi @vkucera, if the process function is called for each DF analyzed in this way the index of the collisions restarts the counting for each collision every DF. I'm just wandering if in this case there will be candidates from different collisions but with the same index.

vkucera: I don't understand the relevance of your comment. This line declares a data member with an initial value of 0. It has nothing to do with resetting the value during processing. The variable is initialised to 0 when an object of type `HfCorrelatorDsHadrons` is created. If you want to reset its value for each call of a member method, you have to do it inside that method.

scattaru: I tested the code locally by inserting the data member declaration inside the process function and attached some plots to illustrate the results. As shown, 6 collisions are saved, but the indices of the Ds candidates are all set to 0. I was wondering whether, in this case, all the candidates are pointing to the first collision in the table, or if each candidate correctly points to its respective collision.
Schermata del 2024-12-17 10-20-09
Schermata del 2024-12-17 10-20-26
Schermata del 2024-12-17 10-20-44

On the other hand, when I declare the data member outside the process function, I obtain the second set of plots, where 21 candidates are reconstructed over 20 collisions, and the candidate indices span from 0 to 19. To me, this suggests that each candidate is correctly associated with the collision it belongs to.
Schermata del 2024-12-17 10-21-01
Schermata del 2024-12-17 10-21-09

vkucera: I'm afraid there have been a few misunderstandings.
- I did not ask you to declare the member inside the process function which would make it a local variable. I asked you not to declare it between methods, referring to the order of declaration of `struct `members.
- Your process function has an explicit loop over collisions. And you increment `hfcReducedCollisionIndex` at the end of each iteration. If the result of your local test was the same value for each collision, I suspect that you wrongly declared it inside the loop instead of outside.
- It is common to increment counters with the `++` operator in C++.

PWGHF/HFC/TableProducer/correlatorDsHadrons.cxx
```diff
@@ -689,32 +695,75 @@ struct HfCorrelatorDsHadrons {
...
+      for (const auto& track : tracks) {
+        if (!track.isGlobalTrackWoDCA()) {
+          continue;
+        }
+        assocTrackReduced(hfcReducedCollisionIndex, track.phi(), track.eta(), track.pt());
       }
-      assocTrackReduced(collReduced.lastIndex(), track.phi(), track.eta(), track.pt());
+
+      collReduced(collision.multFT0M(), collision.posZ());
+      hfcReducedCollisionIndex += 1;
```
vkucera, line 727:
```suggestion
      hfcReducedCollisionIndex++;
```

vkucera (approved):

PR #8998, opened 2024-12-14
[PWGCF] new group of QA histos + minor fixes

abilandz:


PWGCF/MultiparticleCorrelations/Core/MuPa-Configurables.h
```diff
@@ -49,6 +49,9 @@ struct : ConfigurableGroup {
...
   Configurable<vector<string>> cfBookQAParticleHistograms2D{"cfBookQAParticleHistograms2D", {"1-Pt_vs_dcaXY"}, "book (1) or do not book (0) this QA 2D particle histogram"};
+  Configurable<bool> cfFillQAParticleEventHistograms2D{"cfFillQAParticleEventHistograms2D", false, "if false, all QA 2D particle event histograms are not filled. if true, only the ones for which fBookQAParticleEventHistograms2D[...] is true, are filled"};
+  Configurable<vector<string>> cfBookQAParticleEventHistograms2D{"cfBookQAParticleEventHistograms2D", {"1-CurrentRunDuration_vs_itsNCls", "1-CurrentRunDuration_vs_itsNClsNegEtaEbyE", "1-CurrentRunDuration_vs_itsNClsPosEtaEbyE", "1-CurrentRunDuration_vs_Eta0804EbyE", "1-CurrentRunDuration_vs_Eta0400EbyE", "1-CurrentRunDuration_vs_Eta0004EbyE", "1-CurrentRunDuration_vs_Eta0408EbyE", "1-CurrentRunDuration_vs_Pt0005EbyE", "1-CurrentRunDuration_vs_Pt0510EbyE", "1-CurrentRunDuration_vs_Pt1050EbyE"}, "book (1) or do not book (0) this QA 2D particle event histogram"};
```
vkucera, line 53: Names `vector` and `string` are declared inside the `std` namespace. Since you do not specify the `std::` prefix, you force the users of the header to have a using directive in their files.

vkucera (changes requested):

vkucera (changes requested):

PWGCF/MultiparticleCorrelations/Core/MuPa-Configurables.h
abilandz: I am fixng that gradually in the code, as that came just recently. Btw, are you not concerned with the trivial code bloat, if we now start prepending std:: all over the place?

PWGCF/MultiparticleCorrelations/Core/MuPa-DataMembers.h
```diff
@@ -84,27 +84,43 @@ struct EventByEventQuantities {
...
+  TH2F* fQAParticleEventHistograms2D[eQAParticleEventHistograms2D_N][2][2] = {{{NULL}}};     //! [ type - see enum eQAParticleEventHistograms2D ][reco,sim][before, after cuts]
+  bool fFillQAParticleEventHistograms2D = true;                                              // if false, all 2D histograms in this category are not filled. If true, the ones for which fBookQAParticleEventHistograms2D[...] is true, are filled
+  Bool_t fBookQAParticleEventHistograms2D[eQAParticleEventHistograms2D_N] = {kTRUE};         // book or not this 2D histogram, see configurable cfBookQAParticleEventHistograms2D
+  Float_t fQAParticleEventHistogramsBins2D[eQAParticleEventHistograms2D_N][2][3] = {{{0.}}}; // [type - see enum][x,y][nBins,min,max]
+  TString fQAParticleEventHistogramsName2D[eQAParticleEventHistograms2D_N] = {""};           // name of fQAParticleEventHistograms2D, determined programatically from other 1D names, to ease bookkeeping
+  TProfile* fQAParticleEventProEbyE[2][2] = {{{NULL}}};                                      // helper profile to calculate <some-particle-property> event-by-event
```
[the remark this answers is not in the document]
abilandz: Fixed.

PWGCF/MultiparticleCorrelations/Core/MuPa-Configurables.h
vkucera: It is not trivial code bloat at all, so no, I am not concerned.
If you don't use the full names in your header, it's a bug, because the header is not valid, does not compile and requires unjustified action from anyone who includes it.
If you are concerned about the prefix in the `cxx` files, it is very useful to know where entities with common names are taken from, plus it allows to avoid bugs like `abs` vs `std::abs`.

abilandz: Does this mean that we should gradually abandon the usage of 'namespace'? 

I am aware of your O2Linter development, but since it came out of blue just recently, please give us some time to adjust the code to O2Linter requirements. There are much more pressing things, I already promised Victor to align my code with O2Linter gradually, I cannot really do it in one go.

vkucera: No, it does not mean we should abandon using namespace at all. We should just use it correctly. The `std` namespace is a bit special though, hence the O2 linter warning.
Nevertheless, the missing prefix in the header is a problem for any namespace. It's like selling cars without wheels and expecting the customers to buy them themselves.

Thanks for addressing the warnings gradually. No one expects it to be done in one go of course.

victor-gonzalez (approved):

vkucera (approved):

PR #9006, opened 2024-12-16
[PWGJE] Adjusted logic for TC histogram filling

hyungjun-lee:
The TC histogram was creating all histograms related to xy, z, and xyz regardless of the fillIPxy, fillIPz, and fillIPxyz switches.
  This could lead to memory waste, so I modified the switch conditions to address this issue.


nzardosh:
Thanks for the PR to save memory. Please fix the mega linter issues since they seem straightforward and then I can approve it :)

hyungjun-lee:
I'm sorry for late check this request. I'll request again after adding some modification of logic for TC histogram filling. thanks to comment me, Nima.

PR #9018, opened 2024-12-17
[PWGLF] Added generated and reconstructed MC

sawankumawat:
o2-linter: disable=name/workflow-file


sawankumawat:
o2-linter: disable=name/o2-workflow

PR #9032, opened 2024-12-17
[PWGLF] add its nsigma cut + add trigger class cut

maciacco:


alibuild (approved):

PR #9033, opened 2024-12-17
[PWGLF] TPC PID info in f1 table, improved PID and background for f1

skundu692:
Add TPC PID info in f1 table + add momentum dependent PID cut to select f1 daughters + improved rotational background estimation


alibuild (approved):

njacazio (approved):

PR #9037, opened 2024-12-18
[PWGCF] flow: modify code according to O2 linter

Luzhiyongg:


Luzhiyongg:
for the other workflow names, I think I don't have the permission to change someone else's code

victor-gonzalez:
> for the other workflow names, I think I don't have the permission to change someone else's code

Yes, you have it, but it is a task for every analyzer. It will happen
Thanks!!!

victor-gonzalez (approved):
Thank you!!!

PR #9046, opened 2024-12-18
[PWGCF] FemtoUniverse -- Fixing major O2 linter errors for FemtoUniverse directory

zchochul:
This will minimize the number of O2 linter errors, making it easier for other FemtoUniverse users to reduce them in their tasks further.


alibuild (approved):

PR #9055, opened 2024-12-18
[PWGCF] DptDpt -- Tracking the TPC sector borders

victor-gonzalez:
Also
- TOF signal in global momentum
- TPC signal from MC tune on data when required


alibuild (approved):

PR #9059, opened 2024-12-19
[PWGEM/PhotonMeson] Change event counter precision (float -> double)

nstrangm:
- Incrementing by one sometimes did not work because of float precision with high numbers of events


mhemmer-cern (approved):
Looks good

PR #9076, opened 2024-12-20
[PWGCF] LambdaR2Correlation Update

yashpatley:


PWGCF/TwoParticleCorrelations/Tasks/lambdaR2Correlation.cxx
```diff
@@ -581,12 +588,16 @@ struct LambdaCorrTableProducer {
     float retVal = 0.;
 
     if (std::string(obj->ClassName()) == "TH1F") {
+      histos.fill(HIST("Tracks/h1f_tracks_info"), 21.5);
```
victor-gonzalez, line 591: I would recommend to use an `enum` with named constants for the different `tracks_info` cases
If the corresponding histogram `x` axis starts with `-0.5` the `enum` named constants can be used straight on
Additionally appropriate `string` labels can be used to identify each bin when shown

victor-gonzalez (approved):
Have a look at my comment in case you want to consider it for further iterations

PWGCF/TwoParticleCorrelations/Tasks/lambdaR2Correlation.cxx
yashpatley: Dear Victor, Thank you for this suggestion. I will try to update accordingly in next PR.

PR #9080, opened 2024-12-20
[PWGLF] store centrality with high granularity for HM-triggered events

maciacco:


alibuild (approved):

PR #9084, opened 2024-12-20
[PWGJE] Implementation of Hadronic Correction task

Archita-Dash:
Hadronic Correction in the EMCAL framework is implemented to avoid the double counting of the charged particles' contribution in jets.


Archita-Dash:
Hi Nima. yes, that's right. The jet finder needs to use the corrected energy values (we have 4 corrected energy columns- depends on what the analyser wants). everything else remains the same.

Archita-Dash:
BTW Nima, I think the more efficient way to achieve this would be somehow to set up 4 different jet-finders for each corrected energy value such that the user can use whichever jet finder they want corresponding to the desired corrected energy.

nzardosh:
> BTW Nima, I think the more efficient way to achieve this would be somehow to set up 4 different jet-finders for each corrected energy value such that the user can use whichever jet finder they want corresponding to the desired corrected energy.

This would be good if the user needs to run all 4 jet finders at the same time for some reason. If they would instead each only be needed independently then i would make it a configurable in the jet finder to select which is used but only have it happen once. Let me know which is best

PWGJE/TableProducer/emcalClusterHadronicCorrectionTask.cxx
```diff
@@ -0,0 +1,309 @@
...
+#include "PWGJE/Core/JetUtilities.h"
+#include "TVector2.h"
+
+#include "CommonDataFormat/InteractionRecord.h"
+
+using namespace o2;
+using namespace o2::framework;
+using namespace o2::framework::expressions;
+using collisionEvSelIt = o2::soa::Join<o2::aod::Collisions, o2::aod::EvSels>::iterator;
+using selectedClusters = o2::soa::Filtered<o2::aod::EMCALClusters>;
```
nzardosh, line 56: one thing i would recommend here is not performing this on filtered clusters, otherwise this table would not be joinable with the original cluster table, unless exactly the same filters are used everywhere.

Archita-Dash:
> > BTW Nima, I think the more efficient way to achieve this would be somehow to set up 4 different jet-finders for each corrected energy value such that the user can use whichever jet finder they want corresponding to the desired corrected energy.
> 
> This would be good if the user needs to run all 4 jet finders at the same time for some reason. If they would instead each only be needed independently then i would make it a configurable in the jet finder to select which is used but only have it happen once. Let me know which is best

Okay, you are right. Since each would be needed independently by the user, it makes more sense to define a configurable in the jet finder instead.

PWGJE/TableProducer/emcalClusterHadronicCorrectionTask.cxx
```diff
@@ -0,0 +1,309 @@
...
+          continue;
+        } // end CASE 1
+
+        // CASE 2:
+        //  a) If one matched track -> 100% energy subtraction
+        //  b) If more than one matched track -> 100% energy subtraction
+        //  c) If you want to do systematic studies -> perform the above two checks a) and b), and then subtract 70% energy instead of 100%
+
+        // Perform dEta/dPhi matching
+        double dEta = match.track_as<myTracks>().eta() - cluster.eta();
```
fjonasALICE, line 196: this should use eta on the emcal surface instead of eta at vertex

PWGJE/TableProducer/emcalClusterHadronicCorrectionTask.cxx
```diff
@@ -0,0 +1,309 @@
...
+        double dPhi = TVector2::Phi_mpi_pi(match.track_as<myTracks>().phi() - cluster.phi());
```
fjonasALICE, line 197: likewise this should use phi propagated to emcal surface and not at vertex

PWGJE/TableProducer/emcalClusterHadronicCorrectionTask.cxx
```diff
@@ -0,0 +1,309 @@
...
+        // dEta and dPhi cut : ensures that the matched track is within the desired eta/phi window
+
+        // Do pT-dependent track matching
+        if (doMomDepMatching) {
+          auto trackEtaMax = funcPtDepEta.Eval(mom);
+          auto trackPhiHigh = +funcPtDepPhi.Eval(mom);
+          auto trackPhiLow = -funcPtDepPhi.Eval(mom);
+
+          if ((dPhi < trackPhiHigh && dPhi > trackPhiLow) && fabs(dEta) < trackEtaMax) {
+            if (Nmatches == 0) {
```
fjonasALICE, line 210: while the list is indeed sorted to contain the closest track first, for safety reasons I would explicitly check that it is the closest of the ones you looped over

PWGJE/TableProducer/emcalClusterHadronicCorrectionTask.cxx
```diff
@@ -0,0 +1,309 @@
...
+          // Do M02-based correction if enabled
+          Ecluster1 = subtractM02ClusterEnergy(cluster.m02(), Ecluster1, Nmatches, closestTrkP, fHadCorr1, UseFraction1);
+          EclusterAll1 = subtractM02ClusterEnergy(cluster.m02(), EclusterAll1, Nmatches, totalTrkP, fHadCorralltrks1, UseFraction1);
+        } else {
+          // Default energy subtraction (100% and 70%)
+          Ecluster1 = subtractClusterEnergy(Ecluster1, closestTrkP, fHadCorr1, Nmatches, UseFraction1);
+          EclusterAll1 = subtractClusterEnergy(EclusterAll1, totalTrkP, fHadCorralltrks1, Nmatches, UseFraction1);
+        }
+
+        if (UseM02SubtractionScheme2) {
```
fjonasALICE, line 244: what is the M02 subtraction scheme and what is the difference bewteen one and two?

PWGJE/TableProducer/emcalClusterHadronicCorrectionTask.cxx
```diff
@@ -0,0 +1,309 @@
...
+      registry.fill(HIST("h_Ecluster2"), Ecluster2);
+      registry.fill(HIST("h_EclusterAll1"), EclusterAll1);
+      registry.fill(HIST("h_EclusterAll2"), EclusterAll2);
+      registry.fill(HIST("h2_ClsEvsEcluster1"), cluster.energy(), Ecluster1);
+      registry.fill(HIST("h2_ClsEvsEcluster2"), cluster.energy(), Ecluster2);
+      registry.fill(HIST("h2_ClsEvsEclusterAll1"), cluster.energy(), EclusterAll1);
+      registry.fill(HIST("h2_ClsEvsEclusterAll2"), cluster.energy(), EclusterAll2);
+
+      // Fill the table with all four corrected energies
+      hadroniccorrectedclusters(Ecluster1, Ecluster2, EclusterAll1, EclusterAll2);
```
fjonasALICE, line 264: how does one match this table to the cluster table? What is the index? collision id?

PWGJE/TableProducer/emcalClusterHadronicCorrectionTask.cxx
```diff
@@ -0,0 +1,309 @@
...
+    registry.add("h2_ClsEvsEcluster1", "; Cls E w/o correction (GeV); Ecluster1 (GeV)", {HistType::kTH2F, {{350, 0., 350.}, {350, 0., 350.}}});
+    registry.add("h2_ClsEvsEcluster2", "; Cls E w/o correction (GeV); Ecluster2 (GeV)", {HistType::kTH2F, {{350, 0., 350.}, {350, 0., 350.}}});
+    registry.add("h2_ClsEvsEclusterAll1", "; Cls E w/o correction (GeV); EclusterAll1 (GeV)", {HistType::kTH2F, {{350, 0., 350.}, {350, 0., 350.}}});
+    registry.add("h2_ClsEvsEclusterAll2", "; Cls E w/o correction (GeV); EclusterAll2 (GeV)", {HistType::kTH2F, {{350, 0., 350.}, {350, 0., 350.}}});
+
+    // Matched-Track histograms
+    registry.add("h_matchedtracks", "Total matched tracks; track status;entries", {HistType::kTH1F, {{1, 0.5, 1.5}}});
+  }
+
+  Filter clusterDefinitionSelection = (o2::aod::emcalcluster::definition == mClusterDefinition) && (o2::aod::emcalcluster::time >= minTime) && (o2::aod::emcalcluster::time <= maxTime) && (o2::aod::emcalcluster::m02 > minM02) && (o2::aod::emcalcluster::m02 < maxM02);
```
fjonasALICE, line 124: Why are we doing any cluster cuts here? I think then the resulting table will no longer be joinable to the existing cluster table. I think this task should simply provide the values for all clusters. Cluster selection should then be done in the full jet finder

fjonasALICE (changes requested):

PWGJE/TableProducer/emcalClusterHadronicCorrectionTask.cxx
Archita-Dash: Done

PWGJE/TableProducer/emcalClusterHadronicCorrectionTask.cxx
```diff
@@ -0,0 +1,306 @@
...
+  }
+
+  // Helper function for M02-based energy subtraction (based on cluster-shape)
+  double subtractM02ClusterEnergy(double m02, double Ecluster, int Nmatches, double totalTrkP, double correctionFactor, bool UseFraction) {
+    double Ecorr = Ecluster;
+
+    // For M02 in the single photon region, the signal is primarily: Single photons, single electrons, single MIPs
+    if (m02 > 0.1 && m02 < 0.4) {   //circular clusters(electron/photon)
+      Ecorr = Ecluster;  // Single electron, single MIP
+    } else if (m02 > 0.4) {
```
fjonasALICE, line 289: Normally (I think) if M02 is bigger than 0.4 and the number of matches is >1, there should be nothing subtracted. Could you explain why you changed the logic from the function we had in Run2?

PWGJE/TableProducer/emcalClusterHadronicCorrectionTask.cxx
Archita-Dash: These I have defined for the cluster filter but in the main process function, I am now looping over all the EMCAL clusters and only only the selectedclusters. So, no cuts are applied anymore in this case on the clusters.

PWGJE/TableProducer/emcalClusterHadronicCorrectionTask.cxx
FriederikeBock: How would you think it go resorted? Because if she's doing a subselection of the already matched tracks if at all (which are in order of distance to the cluster). Skipping those which don't match her criteria should in principle not change the order.

fjonasALICE: It is not a big point. I just meant that if the list would be unsorted, the Nmatches number would be wrong

PWGJE/TableProducer/emcalClusterHadronicCorrectionTask.cxx
fjonasALICE: they should become not joinable if you apply cuts here that do not get applied to the other table. This is why I suggested to remove all cuts in the hadronic correction task and apply cuts later (e.g. in full jet finder)

fjonasALICE:
HI @Archita-Dash, just checking in with the status of this PR. Are you still working on the suggested changes or fighting with the mega linter and o2 linter? Can we help?

PWGJE/TableProducer/emcalClusterHadronicCorrectionTask.cxx
Archita-Dash: well if this comment is made after comparing the present version of the code with that of the existing one in Run 2 then one also needs to recall that in Run 2 the M02 correction scheme was never correctly implemented or atleast not most likely used in analyses for the same reason I guess. Because if the value is 0.4 and then it could also be an electron in this case where if you don't subtract any energy, your correction is incorrect. That's why this has been taken care of here. But how this needs to be implemented for heavy-flavour jet analyses is something needs to be discussed and followed up later on.

Archita-Dash:
Hi @fjonasALICE . Your suggested changes have been taken care of. Kindly ignore the above repeatedly triggered o2-linter and mega-linter stuffs. This was because I was making some remote changes to my task and constantly updated the master branch and since this PR was still open, it kept on triggering these checks every time.

HI @nzardosh . I would need some help here. I wanted to create another PR to push some changes to my exisitng fuljet analysis task; however yesterday when I did so, unknowingly I kind of pushed those changes in this PR as I made them from the same branch itself. Could this be accepted here? I simply added event selection to reject MB events and added some more QA histos for the JJ MC checks. If this PR gets approved then it would be fairly easily for me to go ahead with the JJ MC QA checks.

nzardosh:
Hi @Archita-Dash for me its fine but i am adding a central cut on the event weight soon so you will then need to convert your task to use that

nzardosh (approved):

Archita-Dash:
> Hi @Archita-Dash for me its fine but i am adding a central cut on the event weight soon so you will then need to convert your task to use that

Thank you, Nima!

Archita-Dash:
@fjonasALICE this now only awaits your approval. Once done, I should be able to go ahead with JJ MC checks with more statistics on hyperloop.

fjonasALICE (approved):

PR #9085, opened 2024-12-20
[PWGHF] Add Lc and Hadron table

zz951:
1. -Add LambdaC and Hadron table which will be used for mixed event calculation offline.


apalasciano (approved):

vkucera:
@zz951 You have to add a space after the prefix if you want it to be recognised.

zz951:
> @zz951 You have to add a space after the prefix if you want it to be recognised.

Ok, thank you!

PR #9094, opened 2024-12-21
[PWGLF] Update of occupancy bins

sarjeetagami:
added upper and lower limit for occupancy


sarjeetagami:
@skundu692 please approve

vkucera:
@sarjeetagami If you want your prefix to be recognised, you have to put a space after it. Otherwise, don't add it at all and it will be added automatically.

vkucera:
You include plenty of ROOT header which are not needed or even used.

skundu692 (approved):

PR #9098, opened 2024-12-21
[Common] add fatal if tracksExtra  converter has no process enabled

ddobrigk:


alibuild (approved):

Common/TableProducer/Converters/tracksExtraV002Converter.cxx
```diff
@@ -19,6 +19,13 @@ using namespace o2::framework;
 struct TracksExtraV002Converter {
   Produces<aod::StoredTracksExtra_002> tracksExtra_002;
 
+  void init(InitContext const&)
+  {
+    if (doprocessV000ToV002 == false && doprocessV001ToV002 == false) {
+      LOGF(fatal, "Neither processV000ToV002 nor processV001ToV002 is enabled. Please choose one!");
```
vkucera, line 25: Maybe better condition would be
```c++
if (doprocessV000ToV002 == doprocessV001ToV002)
```
which also covers the case when both process functions are enabled.

PR #9100, opened 2024-12-21
[PWGLF] addition of derived data path for lambda polarization

ddobrigk:
As discussed with @prottayCMT, this PR adds a process function to the lambda polarization task that uses the centrally produced strangeness derived data for the analysis. The algorithm should match precisely what is done in original AO2Ds. A few relevant comments: 

* the 'check published' option is not available, since it requires all original tracks and these are not saved. 
* the 'use global momentum' is not available when using derived data, since daughter track momenta are only stored at the decay point. 
* in order to be able to fully port the code to derived data, a change had to be done to the derived data format itself:
   * the `DauTracksExtra` did not previously contain findable TPC clusters, which had to be added. Furthermore, I have revamped the data model such that it now more closely matches the original TracksExtra table, making it easier to match template functions in terms of getters in the future. 

I have a question, if I may: 

* I have noticed that there's an escape condition of `if (LambdaTag == aLambdaTag) continue;`. From my understanding, this will mean that a candidate V0 is rejected if it passes _both_ the lambda and the antilambda hypotheses. Looking at the conditions that are checked with the lambda and antilambda tag, it is quite possible that a V0 passes both once the TPC bands of protons and pions merge, and effectively, you may be missing a very large amount of candidates above the band merging. This is also relatively unusual, since TPC dE/dx is traditionally used as a compatibility check (not a rejection check). Is my understanding correct? Is this intentional? I apologize if I misunderstood.

Please, take a look at the code carefully (don´t trust me! I may have misunderstood something or made a mistake somewhere!). Once merged, you can give it a try and see what you get. I hope this helps: running over 2023 Pb-Pb is  much faster when using derived data, and it will hopefully make your life much easier especially for cross-checks and systematics :-)

Tagging also @gianniliveraro @romainschotter @mpuccio for completeness. Marking as draft for the time being.

### DIFF
--- a/PWGLF/DataModel/LFStrangenessTables.h
+++ b/PWGLF/DataModel/LFStrangenessTables.h
@@ -230,7 +230,7 @@ DECLARE_SOA_TABLE(StraTPCQVs, "AOD", "STRATPCQVS", //! tpc Qvec
 DECLARE_SOA_TABLE(StraFT0CQVsEv, "AOD", "STRAFT0CQVSEv", //! events used to compute t0c Qvec
                   epcalibrationtable::TriggerEventEP);
 DECLARE_SOA_TABLE(StraZDCSP, "AOD", "STRAZDCSP", //! ZDC SP information
-                  spcalibrationtable::TriggerEventSP, 
+                  spcalibrationtable::TriggerEventSP,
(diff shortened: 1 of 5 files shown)


ddobrigk:
Moving to review following discussion with @prottayCMT

njacazio (approved):

PR #9108, opened 2024-12-22
[PWGUD] Added few more TPC N Sigmas

ananthapadmanabhan18:
Could anyone tell me how to remove the error caused by O2 linter? I will fix it and update it in PR


rolavick (commented):
With the remaining O2 linter you might want to consult with @vkucera . Currently, I do not see it as an obstacle for the merge, so I approve

rolavick (approved):

ananthapadmanabhan18:
Sure, I will consult with him. Thanks for letting me know!

vkucera:
Hi @ananthapadmanabhan18 , how can I help with the O2 linter errors?

ananthapadmanabhan18:
Hello @vkucera . 90% of the errors in the O2 linter is due to the following error shown in the picture:

image

The problem is...for the data types say `PtEtaPhiVector`, which is the types in standard C++ or O2 replaces these ?

( I have checked the lines in which the error is showing and found out one of them was `PtEtaPhiVector`. There are many others like this only )

vkucera:
> `PtEtaPhiVector`

Those errors are not referring to `PtEtaPhiVector`, they are referring to `Double_t` which should be replace with `double`.

vkucera:
> Hello @vkucera . 90% of the errors in the O2 linter is due to the following error shown in the picture:
> 
> image
> 
> The problem is...for the data types say `PtEtaPhiVector`, which is the types in standard C++ or O2 replaces these ?
> 
> ( I have checked the lines in which the error is showing and found out one of them was `PtEtaPhiVector`. There are many others like this only )

Those errors are not referring to `PtEtaPhiVector`, they are referring to `Double_t` which should be replace with `double`.

PR #9109, opened 2024-12-22
[DPG] adding checks after ROF occupancy sel.bits

altsybee:
added true/false flags for in-ROF and prevROF event selection for occupancy histos


alibuild (approved):

alibuild (approved):

PR #9114, opened 2024-12-23
[Common] cover versioned tables in table autodetect

ddobrigk:
@njacazio @aalkin I'm unsure if there are better ways to do this, but this should already do the trick to first order. This fixes table autodetect in case a table that is being detected is actually versioned. @aalkin: the capturing of the versioned table binding is very primitive, so if there are better ways to do this, please do let me know.


ddobrigk:
Marking as draft until better tested.

PR #9128, opened 2024-12-25
[PWGHF] Add centrality information to Xic0

Tao-Fang:
The FT0M, FT0A, FT0C and numContrib has been added to Xic0 event lever

### DIFF
--- a/PWGHF/TableProducer/treeCreatorToXiPi.cxx
+++ b/PWGHF/TableProducer/treeCreatorToXiPi.cxx
@@ -507,7 +507,7 @@ struct HfTreeCreatorToXiPi {
   }
   PROCESS_SWITCH(HfTreeCreatorToXiPi, processDataFull, "Process data with full information w/o centrality", true);
 
-  void processDataFullWithCentrality(MyEventTable const& collisions, MyTrackTable const&,
+  void processDataFullWithCentrality(MyEventTableWithCent const& collisions, MyTrackTable const&,
                                      soa::Join<aod::HfCandToXiPi, aod::HfSelToXiPi> const& candidates)
   {
     // Filling event properties
--- a/PWGHF/TableProducer/treeCreatorToXiPi.cxx
+++ b/PWGHF/TableProducer/treeCreatorToXiPi.cxx
@@ -271,12 +271,14 @@ struct HfTreeCreatorToXiPi {
     float centFT0M = -1.f;
     float centFV0A = -1.f;
     float centFDDM = -1.f;
-    if (useCentrality) {
+    float multZeqNTracksPV = -1.f;
+    if constexpr (useCentrality) {
       centFT0A = collision.centFT0A();
       centFT0C = collision.centFT0C();
       centFT0M = collision.centFT0M();
       centFV0A = collision.centFV0A();
       centFDDM = collision.centFDDM();
+      multZeqNTracksPV = collision.multZeqNTracksPV();
     }
 
     rowEv(
@@ -293,7 +295,7 @@ struct HfTreeCreatorToXiPi {
       centFT0M,
       centFV0A,
       centFDDM,
-      collision.multZeqNTracksPV());
+      multZeqNTracksPV);
   }
 
   template <typename T>
--- a/PWGHF/TableProducer/treeCreatorToXiPi.cxx
+++ b/PWGHF/TableProducer/treeCreatorToXiPi.cxx
@@ -163,7 +163,6 @@ DECLARE_SOA_COLUMN(TofNSigmaPrFromLambda, tofNSigmaPrFromLambda, float);
 
 DECLARE_SOA_TABLE(HfToXiPiEvs, "AOD", "HFTOXIPIEV",
                   full::IsEventSel8, full::IsEventSelZ,
-                  full::CollisionId,
                   full::McCollisionId,
                   collision::NumContrib,
                   collision::PosX,
@@ -214,7 +213,6 @@ DECLARE_SOA_TABLE(HfToXiPiFulls, "AOD", "HFTOXIPIFULL",
                   full::FlagMcMatchRec, full::DebugMcRec, full::OriginRec, full::CollisionMatched);
 
 DECLARE_SOA_TABLE(HfToXiPiLites, "AOD", "HFTOXIPILITE",
-                  full::CollisionId,
                   full::XPv, full::YPv, full::ZPv, collision::NumContrib, collision::Chi2,
                   full::XDecayVtxCharmBaryon, full::YDecayVtxCharmBaryon, full::ZDecayVtxCharmBaryon,
                   full::XDecayVtxCascade, full::YDecayVtxCascade, full::ZDecayVtxCascade,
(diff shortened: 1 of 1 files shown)


PWGHF/TableProducer/treeCreatorToXiPi.cxx
```diff
@@ -232,13 +265,42 @@ struct HfTreeCreatorToXiPi {
...
+      centFV0A,
+      centFDDM,
+      collision.multZeqNTracksPV());
   }
 
   template <typename T>
   void fillCandidate(const T& candidate, int8_t flagMc, int8_t debugMc, int8_t originMc, bool collisionMatched)
   {
     rowCandidateFull(
+      candidate.collisionId(),
```
fgrosa, line 303: Why do you need to keep the collision Id? Please note that this doesn't allow you to link it to your event table, since the same collisionId can be found in different original DF, and you loose the info of the original DF in self-contained derived data due to AO2D merging. In fact, the addition of this index implies the error of `unbound indices` since the original collision table is not stored.
Also, even in case you want to store the collision Id (which is however not useful for the reason mentioned above), there is no need to declare the column as index column (because it doesn't point to any table that you store).

PWGHF/TableProducer/treeCreatorToXiPi.cxx
```diff
@@ -356,6 +418,7 @@ struct HfTreeCreatorToXiPi {
     if (candidate.resultSelections() && candidate.statusPidCharmBaryon() && candidate.statusInvMassLambda() && candidate.statusInvMassCascade() && candidate.statusInvMassCharmBaryon()) {
 
       rowCandidateLite(
+        candidate.collisionId(),
```
fgrosa, line 421: The addition of this index implies the error of `unbound indices` since the original collision table is not stored. We should not add it here (especially in the `lite` table which is typically meant for the production of samples for ML trainings and does not require the collisionId info)

PWGHF/TableProducer/treeCreatorToXiPi.cxx
```diff
@@ -218,9 +249,11 @@ struct HfTreeCreatorToXiPi {
   Produces<o2::aod::HfToXiPiEvs> rowEv;
 
   Configurable<float> zPvCut{"zPvCut", 10., "Cut on absolute value of primary vertex z coordinate"};
+  Configurable<bool> useCentrality{"useCentrality", false, "Decide whether to use centrality information"};
```
fgrosa, line 252: Instead of adding a configurable, I would templetize the process function in order to have one version that consumes the centrality tables, and another one that doesn't (otherwise regardless the value of this configurable, the user must always add the centrality task since this workflow always consumes the centrality tables)

fgrosa (changes requested):
Hi @Tao-Fang! I have a couple of comments, please take a look at them (especially regarding the unbound indices, I believe that you need to find another strategy)

PWGHF/TableProducer/treeCreatorToXiPi.cxx
fgrosa: `candidate.collisionId()` and `collision.globalIndex()` are the same (provided that `collision` is the collision where the candidate id reconstructed), but they are unique only if the column is declared as index column. The issue is that if you declare it as index column it will break the merging because the collision table is not saved in the derived AO2D, while if you declare it as a regular column, it doesn't ensure uniqueness.

Tao-Fang: Hello Fabrizio, the strategy of link event table is updated which is similar to Lc analysis, used sliceBy function get how much candidate from each collision.

PWGHF/TableProducer/treeCreatorToXiPi.cxx
Tao-Fang: In candidatelite table the collision id is removed

PWGHF/TableProducer/treeCreatorToXiPi.cxx
Tao-Fang: Use template parameters to determine whether to execute the section containing centrality

PWGHF/TableProducer/treeCreatorToXiPi.cxx
```diff
@@ -219,8 +248,12 @@ struct HfTreeCreatorToXiPi {
 
   Configurable<float> zPvCut{"zPvCut", 10., "Cut on absolute value of primary vertex z coordinate"};
 
+  SliceCache cache;
+  Preslice<aod::HfCandToXiPi> candXicPerCollision = aod::hf_cand_xic0_omegac0::collisionId;
+
+  using Cents = soa::Join<aod::CentFV0As, aod::CentFT0Ms, aod::CentFT0As, aod::CentFT0Cs, aod::CentFDDMs>;
   using MyTrackTable = soa::Join<aod::Tracks, aod::TrackSelection, aod::TracksExtra>;
-  using MyEventTable = soa::Join<aod::Collisions, aod::EvSels>;
+  using MyEventTable = soa::Join<aod::Collisions, aod::EvSels, aod::PVMultZeqs, Cents>;
```
fgrosa, line 256:
```suggestion
  using MyEventTable = soa::Join<aod::Collisions, aod::EvSels>;  
  using MyEventTableWithCent = soa::Join<aod::Collisions, aod::EvSels, aod::PVMultZeqs, Cents>;
```

PWGHF/TableProducer/treeCreatorToXiPi.cxx
```diff
@@ -443,15 +504,35 @@ struct HfTreeCreatorToXiPi {
       fillCandidate(candidate, -7, -7, RecoDecay::OriginType::None, false);
     }
   }
-  PROCESS_SWITCH(HfTreeCreatorToXiPi, processDataFull, "Process data with full information", true);
+  PROCESS_SWITCH(HfTreeCreatorToXiPi, processDataFull, "Process data with full information w/o centrality", true);
+
+  void processDataFullWithCentrality(MyEventTable const& collisions, MyTrackTable const&,
```
fgrosa, line 509:
```suggestion
  void processDataFullWithCentrality(MyEventTableWithCent const& collisions, MyTrackTable const&,
```

PWGHF/TableProducer/treeCreatorToXiPi.cxx
```diff
@@ -460,15 +541,35 @@ struct HfTreeCreatorToXiPi {
       fillCandidate(candidate, candidate.flagMcMatchRec(), candidate.debugMcRec(), candidate.originRec(), candidate.collisionMatched());
     }
   }
-  PROCESS_SWITCH(HfTreeCreatorToXiPi, processMcFullXic0, "Process MC with full information for xic0", false);
+  PROCESS_SWITCH(HfTreeCreatorToXiPi, processMcFullXic0, "Process MC with full information for xic0 w/o centrality", false);
+
+  void processMcFullXic0WithCentrality(MyEventTable const& collisions, MyTrackTable const&,
```
fgrosa, line 546:
```suggestion
  void processMcFullXic0WithCentrality(MyEventTableWithCent const& collisions, MyTrackTable const&,
```

fgrosa (changes requested):
@Tao-Fang ok for me if removed from the "lite" table, even though I am still convinced that this will give you the error about unbound indices if you select the "full" table. I think that for the Lc nobody uses the "full" version, but I believe that it also produces the error about unbound indices.
In any case we can merge it as soon as you implement the two remaining comments (if you consume the table with the centrality in the process functions without centrality, they anyway require the centrality task as dependency)

PWGHF/TableProducer/treeCreatorToXiPi.cxx
Tao-Fang: The code has been modified.

fgrosa (dismissed):

ktf:
Looks like a genuine error. Please check.

ktf:
Please check the build.

Tao-Fang:
> Please check the build.
The bug is fix and local build is passed

fgrosa (approved):

PR #9133, opened 2024-12-26
[PWGJE] detailed QA for ch-jet spectra in pp

joonsukbae:
independent further tasks to avoid too much consumption on memory and CPU for core QA task


nzardosh:
Since only 1 histogram is being filled here i think it does not warrant an extra task. Could you please make the PR in the normal QA task instead?

fjonasALICE:
Hi @joonsukbae ,

What is the status of this PR? Will you move to the QA task or extend this task in the future with further histograms? Should this PR stay open or should it be closed?

joonsukbae:
Hi Florian, sorry for not to response, debugging took more than expected. I'm going to make a new PR with many more updates including perpendicular cone method subtraction. Thank you for reminding me!

PR #9171, opened 2024-12-31
[Common] Only load matLUT once, avoid reload when switching runs 

ddobrigk:
...meant as safety feature only


alibuild (approved):

PR #9175, opened 2025-01-01
[PWGHF,Trigger] Build V0s and cascades in HF triggers

fgrosa:
This PR removes the dependence on the strangeness builders from the HF filters. This allows us to reduce the memory consumption and to build only V0s and cascades in interesting events.
It also fixes a couple of details in the B vertex reconstruction.


alibuild (approved):

PR #9185, opened 2025-01-03
[PWGUD] Fix global forward tracks selection and store tan lambda information

diana0x0f:
The global forward tracks (MFT-MCH-MID type) were not selected properly for each BC, now this is done using the global BC info. 
The tangent lambda information is stored in a new table to be able to perform the global muon refit as is done in DQ.

### DIFF
--- a/PWGUD/DataModel/UDTables.h
+++ b/PWGUD/DataModel/UDTables.h
@@ -429,14 +429,9 @@ DECLARE_SOA_TABLE_VERSIONED(UDFwdTracksExtra_001, "AOD", "UDFWDTRACKEXTRA", 1,
 
 using UDFwdTracksExtra = UDFwdTracksExtra_001;
 
-// Tangent lambda information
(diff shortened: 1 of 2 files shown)


rolavick (changes requested):
Why do you introduce this new table? Apparently, there is already a table with this parameter here https://github.com/AliceO2Group/O2Physics/blob/a5e761a8a67d4bf0ff037876376e6820bd34c176/PWGUD/DataModel/UDTables.h#L456
Another approach how to obtain track.tgl is to use the px, py, px information from https://github.com/AliceO2Group/O2Physics/blob/a5e761a8a67d4bf0ff037876376e6820bd34c176/PWGUD/DataModel/UDTables.h#L396 which are actually calculated based on tgl see here https://github.com/AliceO2Group/AliceO2/blob/64dd90cfb914805f3334aa32022581e7a05971b7/Framework/Core/include/Framework/AnalysisDataModel.h#L785

rolavick (approved):

PR #9194, opened 2025-01-06
[PWGHF] Add mc collision counter vs centrality

DelloStritto:
Add histogram with the number of generated collisions vs centrality.

This histogram is needed to compute the efficiency collision reconstruction as a function of the multiplicity

### DIFF
--- a/PWGHF/TableProducer/candidateCreator2Prong.cxx
+++ b/PWGHF/TableProducer/candidateCreator2Prong.cxx
@@ -689,6 +689,7 @@ struct HfCandidateCreator2ProngExpressions {
   using McCollisionsNoCents = soa::Join<aod::Collisions, aod::EvSels, aod::McCollisionLabels>;
   using McCollisionsFT0Cs = soa::Join<aod::Collisions, aod::EvSels, aod::McCollisionLabels, aod::CentFT0Cs>;
   using McCollisionsFT0Ms = soa::Join<aod::Collisions, aod::EvSels, aod::McCollisionLabels, aod::CentFT0Ms>;
+  using McCollisionsCentFT0Ms = soa::Join<aod::McCollisions, aod::McCentFT0Ms>;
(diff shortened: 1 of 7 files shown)


DelloStritto:
> Thanks @DelloStritto . I'm sure I'm missing something, but I have a question: can you please point me to the service wagon in Hyperloop that fills these centrality tables for generated collisions? I see that for this you are using some infrastructure from LF, I'm curious to know what this histogram `h1dFT0M` is
> 
> https://github.com/AliceO2Group/O2Physics/blob/1e27587074fe4ca2d0f3f364cb737d4032ac1a19/PWGLF/TableProducer/Common/mcCentrality.cxx#L108
> 
> )
> Thanks a lot.

Dear @mfaggin, I am not sure there is a service wagon to produce the histo that you are pointing to. At the moment, it is provided to the task uploading it on ccdb and providing the path as an argument. What was done by the LF people, as far as I understood, is to use simulations to convert the number of primary particles at midrapidity in the MC into a FT0 percentile. The histogram you are pointing should contain this correspondence.
I was also planning to contact them after the break to ask about the details.

Thanks!

mfaggin:
Hi @DelloStritto many thanks for the explanation. I was indeed thinking to this histogram as the result of a sort of correlation between charged particles and FT0 signal (I guess done only on reconstructed MC collisions), to be used then with all MC collisions. 
From the code point of view all is good for me, I was just curious about the ingredients.
Do you want it to be merged, or do you prefer to get in contact with LF people first?

DelloStritto:
Thanks a lot for the review @mfaggin. Maybe it's safer if I ask them first. So we don't need to revert the PR if something is changed since their analysis.

I will let you know when you can go ahead with the merging.

DelloStritto:
@mfaggin green light from my side!
Thanks again for the review

mfaggin (approved):

PR #9238, opened 2025-01-08
[PWGJE] Removing single track jets

dajones2:
As Delta R caries no meaning for single track jets, we remove them


nzardosh:
Hi Danny. I am happy to approve this but out of interest wouldnt it be better to keep these so as to have the total number of jets for normalisation purposes and just the single track jets with deltaR = -1.0 or something?

nzardosh (approved):

dajones2:
Hi Nima, I don't think this is needed as we normalise these distributions per trigger rather than per jet. Thank you for the suggestion though, it's worth considering.

PR #9240, opened 2025-01-09
[Common] Cent study: fix histo names in collision processing

ddobrigk:


alibuild (approved):

PR #9241, opened 2025-01-09
[PWGCF] Changed the track PID to use TOF only and added a rudimentary pT-based PID filter

JStaa:


victor-gonzalez:
Any reason for not following the linter recomendations?

JStaa:
> Any reason for not following the linter recommendations?

I didn't know that the linter check would be enforced. Where can I find what the recommendations are?

victor-gonzalez:
Well, as today they are not enforced but they are intended for improving the readability of the code for people to learn from others' experiences, to early catch errors that have been already identified and which will take a while to otherwise debug them, and to improve in efficiency and code sharing

You can find the suggestions under the `Details` link of the `O2 linter` test and also inline on the modified source files under `File changed` tab. While the first source is complete, the second one only shows the initial subset for not over populate the github view

PWGCF/MultiparticleCorrelations/Tasks/ThreeParticleCorrelations.cxx
```diff
@@ -363,28 +363,28 @@ struct ThreePartCorr {
...
-        } else if (particle.pdgCode() == kKMinus) { // Neg kaons
-          MCRegistry.fill(HIST("hRecKaonN"), track.pt());
-        } else if (particle.pdgCode() == kProton) { // Pos protons
-          MCRegistry.fill(HIST("hRecProtonP"), track.pt());
-        } else if (particle.pdgCode() == kProtonBar) { // Neg protons
-          MCRegistry.fill(HIST("hRecProtonN"), track.pt());
-        }
+      if (TrackFilters(track)) {
+        auto particle = track.mcParticle();
+        if (particle.isPhysicalPrimary()) {
```
victor-gonzalez, line 368: I would suggest being careful in here
If you are analyzing reco at this point, you should replicate what you are able to do in data and this information will not be available in data
Of course if you are just collecting information it is perfectly OK

JStaa: At the moment, the MC is used exclusively for calculating the efficiency and the purity of the tracks. It is not (yet) used for building correlations. Is that what you mean by collecting information?

victor-gonzalez: That's what I meant
In my opinion for calculating the efficiency the same criteria as in data should be used for MC reco
In data, the check `isPhysicalPrimary` is not present. Usually a set of tracks cuts are used to try to be close to that condition. Again, in my opinion, the same approach should be used in MC reco
As a side effect, I see that if that condition is required, the contamination from secondaries stays on the air, so to say. Something equivalent I guess will happen with the wrong identification if the associated particle pdg code is considered as part of the MC reco procedure

victor-gonzalez (approved):
Please, consider the linter recommendation in the next iterations

PR #9261, opened 2025-01-10
[PWGLF] Lambda polarization induced by jet in pp collision in 13.6 TeV

15071832337:
The recent code is mainly used to obtain the Lambda signal and then extract the signal and jet construction.
The local laptop tests are successful.
Thank you!


vkucera:
@15071832337 Your PR title had 4 leading spaces which led to a duplication of the PWG prefix. Please either provide a valid prefix or do not provide it at all and it will be added automatically.

ChiaraDeMartin95 (approved):

njacazio (approved):

ktf:
This does not seem to pass tests on linux. Could you have a look?

njacazio (approved):

PR #9268, opened 2025-01-10
[PWGUD] Modification of global track selection in UPCCandidateProducer

diana0x0f:
Modification of global track selection in UPCCandidateProduceR:

- check the correct number of tracks before finding the corresponding MCH-MID tracks
- check whether *both* global tracks are in the MFT acceptance and handle the four cases separately


diana0x0f:
Hi @rolavick , thanks a lot for your review! I have pushed the changes as you requested.

rolavick (approved):

PR #9274, opened 2025-01-11
[PWGCF] adjusted O2 linter and added 2D histograms

hernasab:


victor-gonzalez:
Any reason for not following the linter recommendations?

hernasab:
> Any reason for not following the linter recommendations?

the following don't make sense to me. it doesn't agree with what is in the tutorial? "Use constant references for non-modified iterators in range-based for loops. [const-ref-in-for-loop]"

victor-gonzalez:
> > Any reason for not following the linter recommendations?
> 
> the following don't make sense to me. it doesn't agree with what is in the tutorial? "Use constant references for non-modified iterators in range-based for loops. [const-ref-in-for-loop]"

Well, the recommendations are there for concrete reasons
The `for` loops that are for the amplitudes of the FT0 are safe by not using the `const` reference because they get basically a number. Probably they should be considered as false positives (@vkucera) 
But for the one on `tracks`, using the `const` qualifier secures at compile time that the `tracks` table rows are not modified at all by mistake. So it is a good practice for protecting against mistakes which could affect others analyses in a really really hard way to identify/fix 

The ones suggesting concrete case are intended to improve the readability of the code. People are learning from others' code and a well written code is easy to maintain, not only by the own coder but also for the ones that inherit it

BTW, if you leave an empty line between the copyright notice and the `\file` clause you will get rid of that linter 'error'

hernasab:
> ng the `const` qualifier secures at compile time that the `tracks` table rows are not modified at all by mistake. So it is a good practice for protecting against mistakes which could affect others analyses in a really really hard way to identify/fix

thanks. and why the capitalization and lower case variables importance? where is the documentation on this?

victor-gonzalez:
> > ng the `const` qualifier secures at compile time that the `tracks` table rows are not modified at all by mistake. So it is a good practice for protecting against mistakes which could affect others analyses in a really really hard way to identify/fix
> 
> thanks. and why the capitalization and lower case variables importance? where is the documentation on this?

That was my second point 
> The ones suggesting concrete case are intended to improve the readability of the code. People are learning from others' code and a well written code is easy to maintain, not only by the own coder but also for the ones that inherit it

https://indico.cern.ch/event/1482467/contributions/6246612/attachments/2973662/5235085/2024-11-25_O2-linter.pdf

vkucera:
> > ng the `const` qualifier secures at compile time that the `tracks` table rows are not modified at all by mistake. So it is a good practice for protecting against mistakes which could affect others analyses in a really really hard way to identify/fix
> 
> thanks. and why the capitalization and lower case variables importance? where is the documentation on this?

The official reference are the [O2 conventions](https://rawgit.com/AliceO2Group/CodingGuidelines/master/naming_formatting.html) but you can find a summary on slide 24 of the presentation linked by @victor-gonzalez

hernasab:
> > > ng the `const` qualifier secures at compile time that the `tracks` table rows are not modified at all by mistake. So it is a good practice for protecting against mistakes which could affect others analyses in a really really hard way to identify/fix
> > 
> > 
> > thanks. and why the capitalization and lower case variables importance? where is the documentation on this?
> 
> That was my second point
> 
> > The ones suggesting concrete case are intended to improve the readability of the code. People are learning from others' code and a well written code is easy to maintain, not only by the own coder but also for the ones that inherit it
> 
> https://indico.cern.ch/event/1482467/contributions/6246612/attachments/2973662/5235085/2024-11-25_O2-linter.pdf

this is new to me and will take me a long time to fix since i've had this piece written for almost a year now. i will fully fix it on the next iteration. in the meantime since it is not required, please approve.

PWGCF/Flow/Tasks/FlowZDCtask.cxx
```diff
@@ -229,7 +241,7 @@ struct FlowZDCtask {
 
     if (cent < 0.0 && cent > 70)
       return;
-    TComplex qTPC(0, 0); // Starting with a q-vector of zero
+    std::complex<double> qTPC(0, 0); // Starting with a q-vector of zero
 
     for (auto& track : tracks) {
```
victor-gonzalez, line 246: Please, qualify it as const reference

victor-gonzalez (changes requested):
The const reference in the tracks loop is for security of other analyses
The other can be implemented on the next iterations

victor-gonzalez (approved):
Thanks!!!

hernasab:
> Thanks!!!
Thank you for the help!

PR #9284, opened 2025-01-12
[PWGDQ] Adding 2D histogram for Axe calculation

lucamicheletti93:
Adding 2D histogram for Axe calculation


alibuild (approved):

PR #9286, opened 2025-01-12
[PWGCF] DptDpt -- Fix for wrong behavior excluding good ITS

victor-gonzalez:


alibuild (approved):

PR #9291, opened 2025-01-13
[PIDML] use dynamic columns instead of hardcoding

mytkom:


saganatt:
We should ask (on our team Teams channel) if anyone uses KaonPidTask and whether we can remove it.

PR #9292, opened 2025-01-13
[ALICE3] Add ms & eloss to fasttracker

jesgum:
- Addition of ms and eloss corrections to the fasttracker
- Strangeness tracking now gets resolution from the current layer from the fasttracker
- Grouped cascade configurables


jesgum:
Hi @njacazio @hscheid Could you please have look and approve if possible? Many thanks!

njacazio (approved):

PR #9301, opened 2025-01-13
[PWGCF] Jcorran mass dependence

jaelpark:


jaelpark:
Thanks! I'll look into the other recommendations eventually.

I assume I'm going to have to do a empty commit to rerun the checks once the fix for the macOS-arm check is in place. I might need another approval once that happens.

victor-gonzalez:
I'm in the same shape with one of mine so I will keep an eye

victor-gonzalez (approved):

PR #9310, opened 2025-01-14
[PWGJE] charged jet v2 code

YubiaoWang:
add rho as function of phi, jet pt distribution of in-plane and out-pf-plane using rho_local, delta pt test, and table test of event plane


nzardosh (dismissed):

nzardosh:
Hi @YubiaoWang . It would be great if you could build your O2Physics before pushing to make sure it will build for the git tests. Thanks!

nzardosh (approved):

PR #9319, opened 2025-01-15
General types of inputs and outputs for MlResponse and OnnxModel

choich08365:
`MlResponse.h`
- I have updated ln 161 to resolve the template ambiguity issue first.

I plan to make a draft `MlResponse` to be capable of utilizing the most general types of inputs and outputs (multiple, variable-sized, multi-dimensional tensors). However, since I would like to use `MlResponse` as soon as possible, I would appreciate it if you could merge the current changes first, if possible.

Thank you.


fmazzasc:
Looks good to me (trivial change), thank you!

PR #9324, opened 2025-01-15
[PWGUD] Update MC workflow in fwdMuonsUPC.cxx

ariffero:
In reco workflow:
- Do not get gen candidates.
- Check that the method gets the same gen particles as the one use by Nazar.
- Check that all the particles are muons.

In gen workflow:
- change mu+mu- test


ariffero:
I marked the pull request as ready even if there are some O2linter warnings since they are only related to the name of the task and of the workflow (and changing them will trigger a lot of other warnings).

vkucera:
> I marked the pull request as ready even if there are some O2linter warnings since they are only related to the name of the task and of the workflow (and changing them will trigger a lot of other warnings).

Thanks for addressing the warning. What do you mean by "changing them will trigger a lot of other warnings"? From what I can see, all you have to do is to change `struct fwdMuonsUPC` to `struct FwdMuonsUPC`.

rolavick (changes requested):
Hi Andrea, these LOGF(info) are dangerous when operating on hyperloop. There are written into stdout. When you run over a large sample and you get many of these, the size of stdout grows to much eventually casing the train to fail, effectively preventing you from running on hyperloop. I would like you to consider this again and maybe you want to change to logic, perhaps replacing info with debug.

ariffero:
> Hi Andrea, these LOGF(info) are dangerous when operating on hyperloop. There are written into stdout. When you run over a large sample and you get many of these, the size of stdout grows to much eventually casing the train to fail, effectively preventing you from running on hyperloop. I would like you to consider this again and maybe you want to change to logic, perhaps replacing info with debug.

Hi Roman, thanks for the tip, I have just modified the LOGF(info) in LOGF(debug).

ariffero:
Does anyone know why the building is failing? [In this commit](https://github.com/AliceO2Group/O2Physics/pull/9324/commits/c09f68072dc11daa879dcf62d366b87902d7914b) I should have solved the issue, meaning that now the variable `nzTrMc2` is used (as much as nzTrMc1, for which I do not get complaints) and I have commented `mcCandidates` so that it should not be used.

PWGUD/Tasks/FwdMuonsUPC.cxx
```diff
@@ -921,7 +921,7 @@ struct FwdMuonsUPC {
   // process reco MC (gen info included)
   void processMcReco(CandidatesFwd const& eventCandidates,
                      CompleteFwdTracks const& fwdTracks,
-                     aod::UDMcCollisions const& mcCandidates,
+                     /*aod::UDMcCollisions const& mcCandidates,*/
```
rolavick, line 924: Hi Andrea,
this looks dangerous, because now you are not requesting UDMcCollisions table in this scope. Is it true when you call processMcRecoCand()? Because if you need this table in that function, that you should comment it the following way
aod::UDMcCollisions const& /*mcCandidate)*/,
this way you keep requiring the table, but you avoid the complaint about unused variable. In fact, you can also delete just the "mcCandidate" and everything  should be fine.

rolavick (changes requested):

rolavick (approved):

ktf:
Please have a look at the linter messages and fix them.

PR #9326, opened 2025-01-15
[PWGJE] Add event rejection for MB-gap productions

jaimenorman:
- option to reject events with event weight = 1 - required for MB-gap productions where only the weighted events should be analysed
- event selection fix for particle-level analysis

### DIFF
--- a/PWGJE/Tasks/jetFinderQA.cxx
+++ b/PWGJE/Tasks/jetFinderQA.cxx
@@ -72,6 +72,7 @@ struct JetFinderQATask {
   Configurable<bool> checkMcCollisionIsMatched{"checkMcCollisionIsMatched", false, "0: count whole MCcollisions, 1: select MCcollisions which only have their correspond collisions"};
   Configurable<int> trackOccupancyInTimeRangeMax{"trackOccupancyInTimeRangeMax", 999999, "maximum occupancy of tracks in neighbouring collisions in a given time range; only applied to reconstructed collisions (data and mcd jets), not mc collisions (mcp jets)"};
   Configurable<int> trackOccupancyInTimeRangeMin{"trackOccupancyInTimeRangeMin", -999999, "minimum occupancy of tracks in neighbouring collisions in a given time range; only applied to reconstructed collisions (data and mcd jets), not mc collisions (mcp jets)"};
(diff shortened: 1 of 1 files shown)


fjonasALICE:
This looks reasonable to me and like a straight forward implementation. I was just wondering if it would make sense to put the functionality to "reject MB events" somewhere into the deriveddatautilities so that JE tasks (including the jetfinder) can call that function instead of checking for the weights in each task explicitly ?

nzardosh:
> This looks reasonable to me and like a straight forward implementation. I was just wondering if it would make sense to put the functionality to "reject MB events" somewhere into the deriveddatautilities so that JE tasks (including the jetfinder) can call that function instead of checking for the weights in each task explicitly ?

For now we are checking with @jaimenorman  if there is a better way in the MC to check this, so for now its only a placeholder. But yes we could also put it in the deriveddatautilities. The thinking with putting this in the jetfinder was that you then dont need to really change any of your tasks down the line (if you are using jets. For photons it would indeed still be useful to have a common function). Once Jaime has confirmed the MC thing hes looking into then i think I will make this into a more official cut indeed and also try to apply it at particle level.
 I was also wondering if it would be useful to just outright reject these events in the deriveddatacreator but i think it would also cause some issues for people

fjonasALICE:
> > This looks reasonable to me and like a straight forward implementation. I was just wondering if it would make sense to put the functionality to "reject MB events" somewhere into the deriveddatautilities so that JE tasks (including the jetfinder) can call that function instead of checking for the weights in each task explicitly ?
> 
> For now we are checking with @jaimenorman if there is a better way in the MC to check this, so for now its only a placeholder. But yes we could also put it in the deriveddatautilities. The thinking with putting this in the jetfinder was that you then dont need to really change any of your tasks down the line (if you are using jets. For photons it would indeed still be useful to have a common function). Once Jaime has confirmed the MC thing hes looking into then i think I will make this into a more official cut indeed and also try to apply it at particle level. I was also wondering if it would be useful to just outright reject these events in the deriveddatacreator but i think it would also cause some issues for people

Ok fine by me! I would not outright remove these events. I did some calculations and depending on what analysis you do (and if pileup from the JJ events matters to you) one could actually experiment using the MB events taken for the gap. Since the EMCal is biased to be always on, I think for EMCal these gaps actually contain more statistics than the entire general purpose MC. But this is just a loose idea with many "ifs"

PR #9340, opened 2025-01-15
[PWGLF] Add strangeness builder service

ddobrigk:
* added `strangenessbuilder` task to (eventually) replace the `lambdakzerobuilder`, `cascadebuilder`, `lambdakzeromcbuilder` and `cascademcbuilder` tasks. Using a number of tricks and with autodetect, the strangeness builder will self-configure to generate V0s, Cascades or any exclusive combination thereof depending on the requirements of the tasks that connect to it. Further, a key advantage of this strangeness builder is significantly reduced memory consumption as only one single copy of the material LUT is used (as opposed to two when building V0s and cascades with two independent builders). At the moment, this task is provided for larger-scale testing: it has not yet been fully tested. 
* Added `strangenessBuilderHelper`, a class conveniently able to create V0s and Cascades from any Track table iterator (row) combination provided by any user. The key advantage of having a helper for this is that no additional copy of the material LUT is required when using this tool (in case the task using it already has a copy of the LUT in memory), and it will at the same time provide standard values of V0/Cascade-related quantities. 

Tagging @romainschotter @gianniliveraro


njacazio (approved):

ddobrigk:
@romainschotter we should probably also merge a number of tables currently being used for derived data to simplify life for users. I'll take a look at this when I have time and when I start to prepare this `strangenessbuilder` to generate also those other tables as well...

ChiaraDeMartin95 (approved):

njacazio (approved):

PR #9354, opened 2025-01-16
[PWGDQ] Adding BC tolerance for zorro and new Ev. sel.

lucamicheletti93:
Adding:

- BC tolerance for zorro as a configurable
- and new Ev. sel. for Pb-Pb studies


alibuild (approved):

PR #9367, opened 2025-01-17
[PWGCF] Fix MC processing in FemtoDream

ariedel-cern:
Recent PR, which introduced ITS PID femtodream, introduced breaking changes for MC processing. Fix with this PR.


alibuild (approved):

PR #9383, opened 2025-01-18
[PWGEM,PWGJE] Bug fix for generated neutral mesons

nstrangm:
- Added std:: in front of all abs() calls in associateMCinfoPhoton
- Corrected PDG code requirement in MCGeneratorStudies (should not have caused any issues)
- Added option to only count generated pi0 and eta that decay into gammagamma
- Cosmetic changes in associateMCinfo


mhemmer-cern (approved):
looks good

PR #9384, opened 2025-01-18
[PWGLF] Library cleanup for strangeness builder

ddobrigk:
Meant to do further reduction of memory footprint of the new strangeness builder via removal of the unnecessary `O2::DetectorsBase`. 

For the record, current numbers indicate that the replacement of the `lambdakzerobuilder` and `cascadebuilder` as well as of the labeler services ('mc' builders) leads to a reduction of up to 1GB of memory use already, even still pulling in the entire `DetectorsBase` (which is unnecessary). Speed is also significantly better and is currently being tested. V0s have been proven to be fully equivalent to the ones provided by the canonical `lambdakzerobuilder`; cascades to be tested.


njacazio (approved):

njacazio (approved):

njacazio:
I think ci might be unrelated

PR #9387, opened 2025-01-18
[PWGCF] FemtoUniverse: Fixing default process to FALSE

prchakra:


alibuild (approved):

PR #9391, opened 2025-01-19
[Common] Fix matching to MC for daughters interacting with material

fgrosa:
Together with @mfaggin we realised that whenever a daughter particle (i.e. pion or kaon from a D0 decay) interacts with the material, the particle id changes and in particular another particle with same pdg of the original one is produced by GEANT, and it is set as daughter of the original mother. This can happen several times, due to several interactions with material.
When this happens, `getMatchedMCRec` does not match correctly the signals, since the track can be associated to the particle from material and not the original daughter:
```
[INFO]    pdg code 421
[INFO]        - daughter PDG -321 | idx 58356
[INFO]             - grandDaughter PDG 1000120250 | idx 58518
[INFO]             - grandDaughter PDG 11 | idx 58519
[INFO]        - daughter PDG 211 | idx 58357
[INFO]             - grandDaughter PDG 1000120260 | idx 58514
[INFO]             - grandDaughter PDG 211 | idx 58515
[INFO]             - grandDaughter PDG 11 | idx 58516
[INFO]             ++++  track found! 4963 | idxDaug 58356
[INFO]             ++++  track found! 4981 | idxDaug 58515
[INFO]    idx matched reco particle:   -1
```
To overcome this issue, I added a functionality similar to the one added by @Luca610 for the decays into muons, with the only difference that it searches the mothers iteratively, since multiple interactions might have happen.

The effect is anyway very small (< 1%) in case global tracks are required, since often if the particle interacts with material, there are two track segments reconstructed (only one of the two having hits in the ITS IB, i.e. the one produced by decay itself) or only a track for the particle produced by the material that has no ITS IB hit most of the times.
However, if no global track requirement is applied to the daughters (which might be in case of non-HF decays) the fraction is very high (in the case of the D0, about 25% of the D0 have at least a daughter that interacted with material), so the fraction of non matched signals becomes also very high.

@vkucera could you please have a look to check if it's good for you? If it's the case, I would ask @ddobrigk to merge this so that I will include it in the HF code. Thanks!

### DIFF
--- a/Common/Core/RecoDecay.h
+++ b/Common/Core/RecoDecay.h
@@ -695,14 +695,14 @@ struct RecoDecay {
                              int8_t* nInteractionsWithMaterial = nullptr)
   {
     // Printf("MC Rec: Expected mother PDG: %d", PDGMother);
-    int8_t coefFlavourOscillation = 1;         // 1 if no B0(s) flavour oscillation occured, -1 else
-    int8_t sgn = 0;                            // 1 if the expected mother is particle, -1 if antiparticle (w.r.t. PDGMother)
-    int8_t nPiToMuLocal = 0;                   // number of pion prongs decayed to a muon
-    int8_t nKaToPiLocal = 0;                   // number of kaon prongs decayed to a pion
-    int8_t nInteractionsWithMaterialLocal = 0; // number of interactions with material
-    int indexMother = -1;                      // index of the mother particle
-    std::vector<int> arrAllDaughtersIndex;     // vector of indices of all daughters of the mother of the first provided daughter
-    std::array<int, N> arrDaughtersIndex;      // array of indices of provided daughters
+    int8_t coefFlavourOscillation = 1;          // 1 if no B0(s) flavour oscillation occured, -1 else
+    int8_t sgn = 0;                             // 1 if the expected mother is particle, -1 if antiparticle (w.r.t. PDGMother)
+    int8_t nPiToMuLocal = 0;                    // number of pion prongs decayed to a muon
+    int8_t nKaToPiLocal = 0;                    // number of kaon prongs decayed to a pion
+    int8_t nInteractionsWithMaterialLocal = 0;  // number of interactions with material
+    int indexMother = -1;                       // index of the mother particle
+    std::vector<int> arrAllDaughtersIndex;      // vector of indices of all daughters of the mother of the first provided daughter
+    std::array<int, N> arrDaughtersIndex;       // array of indices of provided daughters
     if (sign) {
       *sign = sgn;
     }
@@ -746,14 +746,17 @@ struct RecoDecay {
         auto process = particleI.getProcess();
         auto pdgI = std::abs(particleI.pdgCode());
         auto pdgMotherI = std::abs(particleI.pdgCode());
-        while (process != kPDecay && process != kPPrimary && pdgI == pdgMotherI) {
+        while (process != TMCProcess::kPDecay && process != TMCProcess::kPPrimary && pdgI == pdgMotherI) {
+          if (!particleI.has_mothers()) {
+            break;
+          }
           auto motherI = particleI.template mothers_first_as<T>();
           pdgI = std::abs(particleI.pdgCode());
           pdgMotherI = std::abs(motherI.pdgCode());
           if (pdgI == pdgMotherI) {
             particleI = motherI;
             process = particleI.getProcess();
-            if (process == kPDecay || process == kPPrimary) { // we found the original daughter that interacted with material
+            if (process == TMCProcess::kPDecay || process == TMCProcess::kPPrimary) { // we found the original daughter that interacted with material
               nInteractionsWithMaterialLocal++;
             }
           }
@@ -840,7 +843,7 @@ struct RecoDecay {
       }
     }
     if constexpr (acceptTrackIntWithMaterial) {
-      if (nInteractionsWithMaterialLocal) {
+      if (nInteractionsWithMaterial) {
         *nInteractionsWithMaterial = nInteractionsWithMaterialLocal;
       }
     }
--- a/Common/Core/RecoDecay.h
+++ b/Common/Core/RecoDecay.h
@@ -695,14 +695,14 @@ struct RecoDecay {
                              int8_t* nInteractionsWithMaterial = nullptr)
   {
     // Printf("MC Rec: Expected mother PDG: %d", PDGMother);
-    int8_t coefFlavourOscillation = 1;     // 1 if no B0(s) flavour oscillation occured, -1 else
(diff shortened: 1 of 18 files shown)


Common/Core/RecoDecay.h
```diff
@@ -737,6 +740,25 @@ struct RecoDecay {
...
+        auto pdgI = std::abs(particleI.pdgCode());
+        auto pdgMotherI = std::abs(particleI.pdgCode());
+        while (process != kPDecay && process != kPPrimary && pdgI == pdgMotherI) {
+          auto motherI = particleI.template mothers_first_as<T>();
+          pdgI = std::abs(particleI.pdgCode());
+          pdgMotherI = std::abs(motherI.pdgCode());
+          if (pdgI == pdgMotherI) {
+            particleI = motherI;
+            process = particleI.getProcess();
+            if (process == kPDecay || kPPrimary) { // we found the original daughter that interacted with material
```
vkucera, line 756: Bug?
```suggestion
            if (process == kPDecay || process == kPPrimary) { // we found the original daughter that interacted with material
```

Common/Core/RecoDecay.h
```diff
@@ -737,6 +740,25 @@ struct RecoDecay {
           particleI = motherI;
         }
       }
+      if constexpr (acceptTrackIntWithMaterial) {
+        // Replace the MC particle associated with the prong by its mother for part → part due to material interactions.
+        // It keeps looking at the mother iteratively, until it finds a particle from decay or primary
+        auto process = particleI.getProcess();
+        auto pdgI = std::abs(particleI.pdgCode());
+        auto pdgMotherI = std::abs(particleI.pdgCode());
+        while (process != kPDecay && process != kPPrimary && pdgI == pdgMotherI) {
```
vkucera, line 749:
```suggestion
        while (process != TMCProcess::kPDecay && process != TMCProcess::kPPrimary && pdgI == pdgMotherI) {
```

Common/Core/RecoDecay.h
```diff
@@ -737,6 +740,25 @@ struct RecoDecay {
...
+          auto motherI = particleI.template mothers_first_as<T>();
+          pdgI = std::abs(particleI.pdgCode());
+          pdgMotherI = std::abs(motherI.pdgCode());
+          if (pdgI == pdgMotherI) {
+            particleI = motherI;
+            process = particleI.getProcess();
+            if (process == kPDecay || process == kPPrimary) { // we found the original daughter that interacted with material
```
vkucera, line 756:
```suggestion
            if (process == TMCProcess::kPDecay || process == TMCProcess::kPPrimary) { // we found the original daughter that interacted with material
```

vkucera (changes requested):

Common/Core/RecoDecay.h
```diff
@@ -680,8 +680,9 @@ struct RecoDecay {
   /// \param depthMax  maximum decay tree level to check; Daughters up to this level will be considered. If -1, all levels are considered.
   /// \param nPiToMu  number of pion prongs decayed to a muon
   /// \param nKaToPi  number of kaon prongs decayed to a pion
+  /// \param nInteractionsWithMaterial  number of daughter particles that interacted with material
   /// \return index of the mother particle if the mother and daughters are correct, -1 otherwise
-  template <bool acceptFlavourOscillation = false, bool checkProcess = false, bool acceptIncompleteReco = false, bool acceptTrackDecay = false, std::size_t N, typename T, typename U>
+  template <bool acceptFlavourOscillation = false, bool checkProcess = false, bool acceptIncompleteReco = false, bool acceptTrackDecay = false, bool acceptTrackIntWithMaterial = false, std::size_t N, typename T, typename U>
```
vkucera, line 685: Please add documentation for
```
/// \tparam acceptTrackIntWithMaterial ...
```

vkucera (changes requested):

Common/Core/RecoDecay.h
fgrosa: I confirm that it works!

Common/Core/RecoDecay.h
fgrosa: Done

Common/Core/RecoDecay.h
fgrosa: Done

Common/Core/RecoDecay.h
fgrosa: Done

vkucera (dismissed):
Thanks @fgrosa

mfaggin (commented):
hi @fgrosa many thanks for implementing the fix. Just a question: looking better into the code, I do not get the role of `nInteractionsWithMaterial` (see lines https://github.com/AliceO2Group/O2Physics/pull/9391/files#diff-7fb78a667de645413dabf3d3cb49ee16a67780bbb11e1f3f839d503906af9aefR847-R851) which is not used neither to constrain the number of iterations with material that one wants to check, nor to monitor things later e.g. with a histogram or anything else. Am I missing something?

fgrosa:
> hi @fgrosa many thanks for implementing the fix. Just a question: looking better into the code, I do not get the role of `nInteractionsWithMaterial` (see lines https://github.com/AliceO2Group/O2Physics/pull/9391/files#diff-7fb78a667de645413dabf3d3cb49ee16a67780bbb11e1f3f839d503906af9aefR847-R851) which is not used neither to constrain the number of iterations with material that one wants to check, nor to monitor things later e.g. with a histogram or anything else. Am I missing something?

It is needed to monitor when the matched candidate had a particle interacting with material, similarly to what done for the decays into muons. We will store this information in the HF tables to allow the analysers to evaluate the effect at the analysis level (excluding or including the signals with a decay daughter that interacted with material). Also consider that the momentum resolution might be worse when the particles interact with material and might effect the mass distribution of signals (as we observed for pi->mu)

mfaggin:
> > hi @fgrosa many thanks for implementing the fix. Just a question: looking better into the code, I do not get the role of `nInteractionsWithMaterial` (see lines https://github.com/AliceO2Group/O2Physics/pull/9391/files#diff-7fb78a667de645413dabf3d3cb49ee16a67780bbb11e1f3f839d503906af9aefR847-R851) which is not used neither to constrain the number of iterations with material that one wants to check, nor to monitor things later e.g. with a histogram or anything else. Am I missing something?
> 
> It is needed to monitor when the matched candidate had a particle interacting with material, similarly to what done for the decays into muons. We will store this information in the HF tables to allow the analysers to evaluate the effect at the analysis level (excluding or including the signals with a decay daughter that interacted with material). Also consider that the momentum resolution might be worse when the particles interact with material and might effect the mass distribution of signals (as we observed for pi->mu)

thanks @fgrosa , indeed I was overlooking this thing (to be propagated to candidate creators)

Many thanks, all good for me

ddobrigk (approved):

ktf:
Please fix the linter in a subsequent PR.

PR #9401, opened 2025-01-20
[PWGJE] updates on background subtraction methods

joonsukbae:
Add histograms to see leading jet pT versus averaged rho,
activate perpendicular cone methods via enabling table generation,
and debugged randome cone methods by limiting the number of jets in collision loop, when filling the histograms that fill delta pT_RC without one or two leading jets.


nzardosh:
Hi Joonsuk. I am not sure I understand this PR. First it seems you remove a bunch of the HF rho tables which shouldnt be done. Secondly the perp cone method relies on jets being found first and then the perpendicular cone is made perpendicular to the jet axis. So your use of it here doesnt make sense I think because for each analysis a different jet finding configuration might be used

PR #9427, opened 2025-01-21
[PWGCF] Latest event selection recommendations for PbPb

jaelpark:


victor-gonzalez (approved):
At some point you have to get rid of the usage of `TDatabasePDG.h` as `linter` suggests because it is extremely inefficient

ktf:
Indeed, please get rid of TDatabasePDG.h ASAP.

jaelpark:
Thanks. It seems that the `TDatabasePDG.h` is included, but not used. We will get rid of it next time.

PR #9443, opened 2025-01-22
[PWGDQ] Adding FT0C histogram for Pb-Pb generated

lucamicheletti93:
Adding FT0C histogram for Pb-Pb generated


alibuild (approved):

PR #9461, opened 2025-01-22
[Infrastructure] Add pre-commit hooks

vkucera:
Adds automatic formatting at the time of `git commit`.
See https://pre-commit.com/


ddobrigk (approved):

ddobrigk:
Approving, thanks! As discussed with Vit, this is a nice addition in case pre-commit is installed locally but current operation will stay unchanged in case pre-commit is not installed.

ktf (approved):

PR #9465, opened 2025-01-23
[PWGLF] modify MC truth tagging part

limhyunji999:
Modify MC truth tagging part to calculate signal loss and resolve some error


smaff92 (approved):
It is strange that the O2 linter failed, because AFAIK it seems that you're using the same name for the struct as the workflow name... I'm approving for now!

PR #9472, opened 2025-01-23
[PWGJE,EMCAL-670] EMCalCorrectionTask fix for unordered collions + New pi0 task

nstrangm:
- EMCalCorrectionTask:
   - Skip collisions where the collisionID of a previous BC is higher than that in a later BC
   - Absolute edge case with only a handful occurrences in all of run 3, but causes an issue in the new JJ (~0.01% of collisions affected)
   - Add QC histogram to monitor this fraction
- BC wise pi0 task
   - Add a new pi0 task that can run without any central barrel detectors (only TVX + EMCal)


fjonasALICE (dismissed):
Urgent commit for processing new JJ MC. Checked implementation and looks good to me. Local tests by Nicolas show 0.02% affected by ordering issue (this had been previously discovered in rare cases also in data). Approved on behalf of EMCal

fjonasALICE (approved):

PR #9502, opened 2025-01-24
[PWGCF] ppCumRun2: Code for calculating efficiency and MC closure check

SwatiSaha-1997:


victor-gonzalez:
Do you really want to use a new source file for your MC analysis?
In principle the detector level processing of the MC analysis must be identical as the processing on data. So you have to have two copies of the same code which is error prone and difficult to maintain/evolve. For the generator level processing of the MC analysis, it is usually smaller and can be easily handled as an additional process on the main source file without the need of replicating anything

SwatiSaha-1997:
Dear Victor, I will need a separate MC at this moment because I want to verify everything here, once verified the observable closure, I will merge the MC part in the Data code as you suggested. It will be very helpful if you could approve it for the time being. Many thanks in advance.
Best regards,
Swati

otonvd (approved):
I approve it on request of Swati. Please follow Victor's instructions to unify the code also for MC asap

shouqiye (approved):

PR #9506, opened 2025-01-24
[PWGHF] Remove unnecessary gen bkg histos and add configurable to fill or not reco bkg histos

fgrosa:
@Marcellocosti @fchinu @scattaru this PR should fix the issue of the high CPU time for the Ds MC and it adds a configurable to choose whether to fill bkg histos or not in Ds and D+ tasks in case of MC


alibuild (approved):

PR #9510, opened 2025-01-24
[PWGCF,PWGLF] Data model update

glromane:
@njacazio 
@ercolessi

### DIFF
--- a/PWGCF/Femto3D/DataModel/singletrackselector.h
+++ b/PWGCF/Femto3D/DataModel/singletrackselector.h
@@ -23,6 +23,7 @@
 #include "Framework/ASoA.h"
 #include "Framework/AnalysisDataModel.h"
 #include "Common/DataModel/PIDResponse.h"
+#include "Common/DataModel/PIDResponseITS.h"
 #include "Framework/Logger.h"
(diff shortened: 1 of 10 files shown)


njacazio (approved):
Thanks Gleb

njacazio:
Just fix the mega linter

njacazio (approved):

njacazio:
Hi @victor-gonzalez could you please have a look?

victor-gonzalez (approved):
Fine from my side
Please, for the next iterations start implementing `linter` recommendations

njacazio:
Tip top! Thanks!

ktf:
This is actually breaking the CI because it now creates an empty workflow if the tables are not present (like when we create the json in the builds). I can remove the protection, however that will result in an empty json. @jgrosseo will that create issues to HyperLoop?

njacazio:
Hi @ktf this is a sort of pilot and uses the same procedure that we wanted to have in the unified converter and could be relevant also there.

jgrosseo:
@ktf An empty workflow means an empty json?
In this case, Hyperloop will not show any configurables...

PR #9524, opened 2025-01-25
[PWGCF] NetchargeFluctuations

nidamalikk:
This code creates QA plots for the Net-charge Fluctuations using an observable #nu_dyn at energy 13.6 TeV for pp collision system.


victor-gonzalez:
So, which one is valid, this one, or https://github.com/AliceO2Group/O2Physics/pull/9525

victor-gonzalez:
Please, close the one that is not valid

PR #9528, opened 2025-01-25
[PWGCF] Kaon isospin fluctuation

rahulverma012:
New code for studying Kaon Isospin Fluctuations in pp collisions


PWGCF/EbyEFluctuations/Tasks/kaonIsospinFluctuations.cxx
```diff
@@ -0,0 +1,1409 @@
...
+
+    if (posIsPion && negIsPion) {
+      v0TagValue += 1; // It is K0s
+    }
+    if (posIsProton && negIsPion) {
+      v0TagValue += 2; // It is Lambda
+    }
+    if (posIsPion && negIsProton) {
+      v0TagValue += 4; // It is AntiLambda
+    }
```
victor-gonzalez, line 724: I would suggest using named constants and flags instead
For instance you can use an integer variable as having 32 flags, one per bit, and you use named constants for the meaning of each bit, for instance if you want the second bit to mean that the pair is a Lambda you might use
```
#define kPAIRBIT_ISLAMBDA (1 << 1)
```
you can set the corresponding bits in your variable and check for them on any place in your code
With the current approach you have to check the whole arithmetic

PWGCF/EbyEFluctuations/Tasks/kaonIsospinFluctuations.cxx
```diff
@@ -0,0 +1,1409 @@
...
+      v0daughterCollisionIndexTag = +1;
+    }
+    if (v0.collisionId() == negDaughterTrack.collisionId()) {
+      v0daughterCollisionIndexTag = +2;
+    }
+    if (posDaughterTrack.collisionId() == negDaughterTrack.collisionId()) {
+      v0daughterCollisionIndexTag = +4;
+    }
+    return v0daughterCollisionIndexTag;
+  }
```
victor-gonzalez, line 742: Something similar here

victor-gonzalez (approved):
Please, have a look at my comments for future iterations in case you want to consider them

PR #9533, opened 2025-01-25
[Common] Analysis-level muon re-alignment task and new tables definition

Elros60:
1. Add new tables definition for re-aligned forward tracks keeping the same getters as the original forward tracks
2. Add a new table producer which will do muon re-alignment at analysis-level, providing re-aligned tracks information that can be accessed by any PWG analysers.

### DIFF
--- a/Common/DataModel/CMakeLists.txt
+++ b/Common/DataModel/CMakeLists.txt
@@ -23,4 +23,5 @@ o2physics_add_header_only_library(DataModel
                                           MatchMFTFT0.h
                                           MftmchMatchingML.h
                                           ZDCInterCalib.h
-                                          EseTable.h)
+                                          EseTable.h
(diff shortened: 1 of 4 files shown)


Elros60:
MegaLinter still complains about `filesystem` library. 
The calling of `filesystem::exists` is useful in case of using this task locally where you may need to use your local geometry file or grp file instead of reading them from CCDB; this option will be quite important for alignement study.

ddobrigk:
Hi @ktf , seems like the CI is unrelated. Can you please take a look? Thank you!

ddobrigk (approved):

Elros60:
> Hi @ktf , seems like the CI is unrelated. Can you please take a look? Thank you! 

Seems like we are having a common issue with git? My another PR is in the same trouble.

ddobrigk (approved):

PR #9542, opened 2025-01-27
[PWGDQ] Adding useful cut ITS + Occupancy

victorvalenciatorres:


lucamicheletti93 (commented):
We have already differente selection for occupancy:

https://github.com/AliceO2Group/O2Physics/blob/28e9474c5e1c19736b99fb7eb751cf97a5048b32/PWGDQ/Core/CutsLibrary.cxx#L3739C1-L3766C4

Before adding this other cuts it is better to check the impact of the exinsting cuts and then implement new cuts

PR #9548, opened 2025-01-27
[PWGCF] Fixes and improvements for FemtoDream Cascades

gmantzar:
- Fixing selection bug in the Cascades
- Adding "BeforeCuts" and "AfterCuts" in the selection class of femtodream (for tracks and cascades only for now)
- chaning std::abs() to std::fabs() in the track and cascade selection
- adding the Omega selection flag in the producer task to select Omegas as Cascades
- Charge of the Cascade Candidates is determined by the cascade builder


victor-gonzalez (approved):
I approve it for the time being but the following linter recommendations need to be addressed in the next iteration
- the inclusion of `<iostream>`, impact compilation time and should not be used in O2Physics
- the use of `<TDatabasePDG.h>` is quite inefficient regarding memory usage
- use constant references for non-modified iterators in loops, that prevent inefficient memory management and access and also accidental inter-task common data modification which is really hard to catch

The rest of the linker recommendations needs to be addressed in a timely manner 
And, please, remove all commented code, they reduce readability

PR #9571, opened 2025-01-28
[PWGCF] STS: fix dca min selection

sofiatomassini:


victor-gonzalez (approved):
Please, consider the linter recommendations for the next iterations

PR #9595, opened 2025-01-28
Fix linter issues in data model

vkucera:


ddobrigk:
Hi @vkucera, thanks for the efforts! If I may, I would suggest to keep this in draft until after quark matter: at this point, any unforeseen issues when renaming variables will cause delays, and we are right in the middle of QM approvals. (BTW, just to check: I was looking through the changed files and I don't think you actually touch the names of _configurables_, do you? That might be a bit dangerous for HY settings)

vkucera:
> Hi @vkucera, thanks for the efforts! If I may, I would suggest to keep this in draft until after quark matter: at this point, any unforeseen issues when renaming variables will cause delays, and we are right in the middle of QM approvals. (BTW, just to check: I was looking through the changed files and I don't think you actually touch the names of _configurables_, do you? That might be a bit dangerous for HY settings)

@ddobrigk I am not touching anything exposed outside O2, i.e. no changes in table types written on disk or names of executables or configurables.

vkucera:
@fgrosa @zhangbiao-phy Can you please approve?

fgrosa (approved):

PR #9600, opened 2025-01-29
[PWGCF] Adding v2 four particle cumulants

Preet-Bhanjan:
Added the v2{4PC} after aligning local repository with O2 master


victor-gonzalez:
It looks much better
Thanks!

victor-gonzalez (approved):

PR #9616, opened 2025-01-29
[PWGHF] Add methods to compute fraction, provided both efficiencies and cut-variation results externally

fgrosa:
@atavirag this PR adds two methods to the cut-variation class, to compute fractions provided efficiencies and corrected yields from cut variation as inputs (so externally). This allows the calculation of the raw fraction for any set of selections, without requiring to re-run the minimisation
It also adds a 4-panel plot in the script (useful for presentations)


alibuild (approved):

vkucera:
@fgrosa There are several import issues. Please check.

PR #9631, opened 2025-01-30
[PWGEM] LMee: added TTCA option to resolution map creation

jjungIKF:


dsekihat:
Dear @jjungIKF , This approach does not work, because this task runs over original AO2D.

PR #9639, opened 2025-01-30
[PWGCF] Push correct task

ilikmeta:


ilikmeta:
Additionally, I test O2Linter locally, and it passed all tests. Here though is failed. 
`O2 linter result: All tests passed.`

PR #9643, opened 2025-01-30
[PWGLF] Move heavyionMultiplicity task from PWGMM/Mult to PWGLF/GlobalEventProperties

abmodak:


gbencedi (approved):
@abmodak thanks for the migration!

PR #9655, opened 2025-01-31
[PWGJE] Fix EMC rotation BG bug

nstrangm:


mhemmer-cern (changes requested):
Can you also add the `&` to make the loop use const ref to make the O2linter happy?

mhemmer-cern (approved):
Nvm the O2 linter is wrong here,
Should be okay.

PR #9656, opened 2025-01-31

yashpatley:
1. Updated pair efficiency histograms
2. Added a process function for computation of correlation function from mix event method

### DIFF
--- a/PWGCF/TwoParticleCorrelations/Tasks/lambdaR2Correlation.cxx
+++ b/PWGCF/TwoParticleCorrelations/Tasks/lambdaR2Correlation.cxx
@@ -22,7 +22,6 @@
 #include "Framework/AnalysisTask.h"
 #include "Framework/ASoAHelpers.h"
 #include "Framework/runDataProcessing.h"
-#include "PWGLF/DataModel/LFResonanceTables.h"
 #include "PWGLF/DataModel/LFStrangenessTables.h"
 #include "CommonConstants/PhysicsConstants.h"
 #include "Common/Core/RecoDecay.h"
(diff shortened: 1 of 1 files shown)


PWGCF/TwoParticleCorrelations/Tasks/lambdaR2Correlation.cxx
```diff
@@ -1216,84 +1237,156 @@ struct LambdaR2Correlation {
...
+    // get ccdb object
+    TObject* obj = reinterpret_cast<TObject*>(ccdbObj->FindObject(Form("%s", vPairCorrFactStrings[part_pair].c_str())));
+    TH1F* hist = reinterpret_cast<TH1F*>(obj->Clone());
+    float retVal = 0., dRap = p1.rap() - p2.rap();
+
+    if (std::string(obj->ClassName()) == "TH2F") {
+      retVal = hist->GetBinContent(hist->FindBin(p1.pt(), p2.pt()));
+    } else if (std::string(obj->ClassName()) == "TH3F") {
+      retVal = hist->GetBinContent(hist->FindBin(p1.pt(), p2.pt(), dRap));
+    } else {
```
victor-gonzalez, line 1321: I think I already suggested at some point to use the histogram number of dimensions instead of comparing strings at the pair loop level

victor-gonzalez (approved):
Have in mind that only the collisions in the same data frame are going to be mixed by the mixed event framework. If you really need mixed event probably you would need to consider derived data

PWGCF/TwoParticleCorrelations/Tasks/lambdaR2Correlation.cxx
yashpatley: Dear Victor, It completely slipped my mind, I will do it in next PR. Thank you for reminding.

PR #9657, opened 2025-01-31
[DPG] Fix track from which to compute the signal

njacazio:


njacazio:
@giovannimalfattore

alibuild (approved):

PR #9672, opened 2025-02-01
[PWGDQ] Removing temporarily v3 histos from flow-dimuon & res-flow-dimuon

lucamicheletti93:
Removing temporarily v3 histos from flow-dimuon & res-flow-dimuon to study the memory issye


alibuild (approved):

PR #9679, opened 2025-02-01
[PWGHF] added tracks efficiency calculation block

1994ra:


PWGHF/HFC/TableProducer/correlatorLcHadrons.cxx
```diff
@@ -486,6 +497,7 @@ struct HfCorrelatorLcHadrons {
     bool isLcPrompt = false;
     bool isLcNonPrompt = false;
     bool isLcSignal = false;
+    int cntLc = 1;
```
vkucera, line 500: What is `cnt`? The name is not clear enough to tell the meaning of the variable.

vkucera:
Please fix the PR title.

PWGHF/HFC/TableProducer/correlatorLcHadrons.cxx
1994ra: @vkucera, this variable is from the previous version, however, cnt means count here

apalasciano (approved):

PR #9680, opened 2025-02-01
[PWGJE] hybrid track cuts for Run2 converted data

sweyh99:
Matching the 2011 definition (https://twiki.cern.ch/twiki/bin/viewauth/ALICE/HybridTracks). Verified consistency with AliPhysics via local tests

### DIFF
--- a/PWGJE/Core/JetDerivedDataUtilities.h
+++ b/PWGJE/Core/JetDerivedDataUtilities.h
@@ -558,7 +558,8 @@ enum JTrackSel {
   globalTrack = 1,
   qualityTrack = 2,
   qualityTrackWDCA = 3,
-  hybridTrack = 4
+  hybridTrack = 4,
+  hybridTrackRun2 = 5
 };
 
 template <typename T>
@@ -589,6 +590,8 @@ int initialiseTrackSelection(std::string trackSelection)
     return JTrackSel::qualityTrackWDCA;
   } else if (trackSelection == "hybridTracks") {
     return JTrackSel::hybridTrack;
+  } else if (trackSelection == "hybridTracksRun2") {
+    return JTrackSel::hybridTrackRun2;
   }
   return -1;
 }
@@ -614,6 +617,10 @@ uint8_t setTrackSelectionBit(T const& track, float trackDCAZ, float maxDCAZ)
   if (track.trackCutFlagFb5()) {
     SETBIT(bit, JTrackSel::hybridTrack);
   }
+  if (track.trackCutFlagFb5()) { //Change get FB6 to work
+    SETBIT(bit, JTrackSel::hybridTrackRun2);
+    LOG(info) << "In JetDerivedDataUtilities setTrackSelectionBit: The bit was set to HYBRID_RUN2";
+  } //CHANGE
   return bit;
 }
 
--- a/PWGJE/Core/JetDerivedDataUtilities.h
+++ b/PWGJE/Core/JetDerivedDataUtilities.h
@@ -617,10 +617,9 @@ uint8_t setTrackSelectionBit(T const& track, float trackDCAZ, float maxDCAZ)
   if (track.trackCutFlagFb5()) {
     SETBIT(bit, JTrackSel::hybridTrack);
   }
-  if (track.trackCutFlagFb5()) { //Change get FB6 to work
+  if (track.trackCutFlagFb5()) {
     SETBIT(bit, JTrackSel::hybridTrackRun2);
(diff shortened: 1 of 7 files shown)


PWGJE/Core/JetDerivedDataUtilities.h
```diff
@@ -561,6 +564,9 @@ uint8_t setTrackSelectionBit(T const& track, float trackDCAZ, float maxDCAZ)
   if (track.trackCutFlagFb5()) {
     SETBIT(bit, JTrackSel::hybridTrack);
   }
+  if (track.trackCutFlagFb5()) {
```
nzardosh, line 567: This seems to check the same things as the if statement above?
So hybridTrack and hybridTrackRun2 are the same? If the original hybridTrack has just been updated (i.e trackCutFlagFb5 has been updated), then we can just keep on using hybridTrack as it was and not add hybridTrackRun2

PWGJE/TableProducer/derivedDataProducer.cxx
```diff
@@ -184,62 +186,85 @@ struct JetDerivedDataProducerTask {
...
   void processMcCollisions(aod::McCollision const& mcCollision)
   {
-    jMcCollisionsTable(mcCollision.posX(), mcCollision.posY(), mcCollision.posZ(), mcCollision.weight(), mcCollision.getSubGeneratorId());
-    jMcCollisionsParentIndexTable(mcCollision.globalIndex());
+    products.jMcCollisionsTable(mcCollision.posX(), mcCollision.posY(), mcCollision.posZ(), mcCollision.weight(), mcCollision.getSubGeneratorId());
+    products.jMcCollisionsParentIndexTable(mcCollision.globalIndex());
   }
   PROCESS_SWITCH(JetDerivedDataProducerTask, processMcCollisions, "produces derived MC collision table", false);
 
+  void processTracksRun2(soa::Join<aod::Tracks, aod::TracksExtra, aod::TracksCov, aod::TracksDCA, aod::TrackSelection, aod::TrackSelectionExtension>::iterator const& track)
```
nzardosh, line 242: could you please move this process function to below processTracksWithCollisionAssociator and also remove the commented blocks of code out

nzardosh (commented):
Thanks for the PR @sweyh99 . Just a few small comments from the JE side but otherwise its good to go. Depending on how long the common code takes to review you might have to rebase to the latest master as there are some other PRs changing the derivedDataProducer that might be added

PWGJE/Core/JetDerivedDataUtilities.h
sweyh99: Yes it does. trackCutFlagFb5 has been updated, so hybridTrack and hybridTrackRun2 are the same. I have removed hybridTrackRun2 in the latest PR so we can just use hybridTrack.

PWGJE/TableProducer/derivedDataProducer.cxx
sweyh99: Done

nzardosh:
Thanks @sweyh99 for making the changes. Once you rebase to master the producesGroup change should dissapear since this was already added in a previous PR. After that from the jet side all looks good.
@ddobrigk does this look ok from your side?

nzardosh:
@ddobrigk could you please check if this is ok from your side?

ddobrigk:
Thanks @sweyh99 , this looks good to me! Since this touches also track selection, maybe it would be good if perhaps the DPG AOT tracks could confirm that this looks good to them? Pinging also @f3sch @lbariogl @iouribelikov. Thanks a lot!

lbariogl:
Thanks @sweyh99. @ddobrigk I went through the changes and I think that they are ok. Green light on my side

jgrosseo:
Was a test done if one gets identical selected tracks with default settings (and without the new cut)?

f3sch:
LGTM, but I do not know this code, just quickly skimmed through it. From what I can tell the changes in the common code should only affect JE selection since the default is set high. But of course, I agree with @jgrosseo that a quick validation would be prudent.

sweyh99:
@nzardosh This has been rebased to master and the producesGroup changes in the derivedDataProducer disappeared. 
@jgrosseo @f3sch I checked and confirmed that you get exactly the same default tracks with and without these cuts. I tested it with the monitorfilterbit task and the FB0 histograms corresponding to using the standard global track selection are exactly the same before and after this push.

ddobrigk (approved):

nzardosh (approved):

PR #9687, opened 2025-02-02
[PWGHF] refined soft pion selection in tcc tree creator

zhangbiao-phy:


alibuild (approved):

PR #9692, opened 2025-02-02
[PWGDQ] Adding functionality to define histograms, cuts and MC signals in configurables via JSON formatted strings

iarsene:
Allows the definition of ad-hoc histograms, cuts and MC signals.


alibuild (approved):

PR #9699, opened 2025-02-03
[PWGDQ] FT0C occupancy added to PWGDQ code

sjzhu0v0:
FT0C occupancy has been added to PWGDQ/DataModel. The corresponding modification has been added to tableMaker.cxx.


iarsene:
Hi @sjzhu0v0,
I don't really like creating a new table with just one column. This could lead to lots of tables in the future. So I suggest to add this column into the multPV table. As a temporary solution, in order to not break backward compatibility for older skims, you can make a temporary converter task for this table.

sjzhu0v0:
> Hi @sjzhu0v0, I don't really like creating a new table with just one column. This could lead to lots of tables in the future. So I suggest to add this column into the multPV table. As a temporary solution, in order to not break backward compatibility for older skims, you can make a temporary converter task for this table.

Hi, @iarsene . Thanks for your suggestion! Which folder should I add such a converter to? Please kindly give me a recommendation.

iarsene:
Hi, you can add it to Tasks

iarsene:
Also, @sjzhu0v0, please be aware that there is another change for this table being prepared. The PR is nearly ready, it should be merged immediately after we merge your change. So, after you merge your PR, wait a bit before yiu create new skims so that the other updates are merged

sjzhu0v0:
Hi, @iarsene . I've created a new pull requeset. https://github.com/AliceO2Group/O2Physics/pull/9760 . Please take a look and close this pull requeset.

PR #9710, opened 2025-02-03
[PWGCF] DptDpt - Event selection tune

victor-gonzalez:
Additional flags incorporated to the MB selection
- kNoCollInTimeRangeStandard
- kNoCollInRofStandard

Incorporated minimum value for the occupancy selection


alibuild (approved):

PR #9712, opened 2025-02-03
[PWGHF] Fix the bug of the thnsparse axis number in Lc task

zhangbiao-phy:


alibuild (approved):

PR #9717, opened 2025-02-04
[PWGCF] adjusted contents for o2 linter and added event selection and correlations

hernasab:


victor-gonzalez (changes requested):
You have to rebase your changes to the latest version of O2Physics

PR #9747, opened 2025-02-04
[PWGUD] PWG-UD: Test on rho' in UPCs using UD tables

CesarOmarRA:


rolavick (dismissed):

rolavick:
Could you please use more describing words for your commits?
Also, why you call your task RhoF? Why not i.e. RhoPrime or some other word more distinctive from names of other tasks?

rolavick (approved):

rolavick (approved):

PR #9761, opened 2025-02-05
[PWGHF] Add ML variables in treeCreatorD0ToKPi

atavirag:


fcatalan92:
Hi @atavirag, the build is currently failing. Please have a look and fix the issues, which look sensible to me.

atavirag:
Hi @fcatalan92, I was missing one line, but it should be fixed now. Thanks!

PWGHF/TableProducer/treeCreatorD0ToKPi.cxx
```diff
@@ -401,6 +426,20 @@ struct HfTreeCreatorD0ToKPi {
...
+            candidate.mlProbD0()[1],
+            candidate.mlProbD0()[2]);
+        }
+        if (candidate.isSelD0bar()) {
+          rowCandidateMl(
+            candidate.mlProbD0bar()[0],
+            candidate.mlProbD0bar()[1],
+            candidate.mlProbD0bar()[2]);
+        }
+      }
```
fgrosa, line 442: I would suggest to move this to the `fillTable` function (adding the `applyMl` template argument), to have the filling of the tables in the same place

PWGHF/TableProducer/treeCreatorD0ToKPi.cxx
```diff
@@ -478,6 +538,20 @@ struct HfTreeCreatorD0ToKPi {
...
```
fgrosa, line 554: I would suggest to move this to the `fillTable` function (adding the `applyMl` template argument), to have the filling of the tables in the same place

fgrosa (changes requested):
Hi @atavirag I just have a small suggestion before merging, thanks!

PWGHF/TableProducer/treeCreatorD0ToKPi.cxx
atavirag: Hi @fgrosa, I just made the change, thanks!

fgrosa (approved):

PR #9762, opened 2025-02-05
[PWGJE] Hadron jet recoil

KotliarovAr:
Edited material:

1.    introduce conditions to prevent the creation of empty histograms when it is not needed
2.    change the range of histograms
3.    add selection cuts on MC particles
4.    fix the issue when a neutral MC particle could be accepted for the analysis


KotliarovAr:
Hi @nzardosh 

All merging conflicts should be resolved

nzardosh (approved):

PR #9769, opened 2025-02-05
[PWGDQ,Trigger] Proposal for the electron-muon pp trigger

MaolinZH:
1. Extend the dq trigger group in filterTable
2. Activate the CEFP related code for e-mu trigger in filterPPwithAssociation
3. Modify the filterPP correspondingly for the compile
4. Add related cuts in CutsLibrary


mpuccio (dismissed):
Approving for the core trigger part

mpuccio (approved):

PR #9772, opened 2025-02-05
[PWGLF] updated event average to particle average

prottayCMT:


ChiaraDeMartin95 (approved):

vkucera:
@prottayCMT You have unused PDG stuff (includes and the service).

PR #9780, opened 2025-02-05
[PWGLF] Add ctpRateFetcher service

romainschotter:
- Add ctpRateFetcher service
- Add event selections on local IR


njacazio (approved):

njacazio:
We should fix mega linter in a coming pr

PR #9801, opened 2025-02-06
[PWGHF,Trigger] Add option to select only signal in MC and improve MC matching

fgrosa:


alibuild (approved):

PR #9805, opened 2025-02-07
[PWGDQ] Adding Single muon Correlators

victorvalenciatorres:
Adding single muon and antimony correlators needed for event mixing. (For cumulant's method)


PWGDQ/Core/VarManager.h
```diff
@@ -688,6 +690,23 @@ class VarManager : public TObject
...
+    kM0111POIoverMpminus,
+    kM0111POIoverMpplus,
+    kCORR2POIplus,
+    kCORR2POIminus,
+    kCORR4POIplus,
+    kCORR4POIminus,
+    kM11REFoverMpplus,
+    kM1111REFoverMpplus,
+    kM11REFoverMpminus,
+    kM1111REFoverMpminus,
```
iarsene, line 709: you should add these as well in the fgVarNamesMap, otherwise they will not be available via JSON configurables

lucamicheletti93 (approved):

PR #9818, opened 2025-02-07
[PWGCF] add ML score into CF filter

jikim1290:
add ML score into CF filter


jaelpark:
> I guess this prevents backward compatibility with previous derived data @jaelpark is that assumed?

It will break it, but that is probably unavoidable. We will work around that for sometime with existing data.

My question is does the added ML score subscription break the compatibility with non-ML based selections? Additionally, would something less precise than float be enough to save space?

jaelpark:
Actually, the alternative might be to also do entirely different ML specific table? In that case, compatibility will remain if I'm not wrong. Would that be better?

victor-gonzalez:
A compromise solution could be to have an additional joinable table with only the ML information

jaelpark:
> A compromise solution could be to have an additional joinable table with only the ML information

Yes, this is what I meant.

PR #9837, opened 2025-02-08
[PWGLF] added q vectors information in strangeness derived data tables

prottayCMT:
@ddobrigk updated the strangeness table with ZDC q vectors


ddobrigk:
Looks okay to me, thanks @prottayCMT !
@romainschotter note we may have to redo the small pass4 Pb-Pb with this change for Prottay to be able to use the derived data with the Q vectors. Can you look into that? Thanks!

skundu692 (approved):

PR #9847, opened 2025-02-09
[PWGHF] Simplify derived-data macros

vkucera:


alibuild (approved):

PR #9848, opened 2025-02-10
[PWGHF] Fix bug of the initial type of generated pt of B mother in Lc task

zhangbiao-phy:


alibuild (approved):

PR #9850, opened 2025-02-10
[PWGCF] a new table with ML score

jikim1290:
a new table with ML score

### DIFF
--- a/PWGCF/DataModel/CorrelationsDerived.h
+++ b/PWGCF/DataModel/CorrelationsDerived.h
@@ -123,19 +123,10 @@ namespace cf2prongtrackml
 {
 DECLARE_SOA_COLUMN(MlProbD0, mlProbD0, std::vector<float>);                       //!
 DECLARE_SOA_COLUMN(MlProbD0bar, mlProbD0bar, std::vector<float>);                 //!
-enum ParticleDecay {
-  D0ToPiK,
-  D0barToKPi,
-  JPsiToEE,
-  JPsiToMuMu
-};
 } // namespace cf2prongtrackml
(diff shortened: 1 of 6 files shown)


jaelpark:
Hi Junlee, thanks! I think we were thinking about some table with just the ML related columns. This table could then be joined with the existing `CF2ProngTracks`. That would be the most ideal. So for example, you could only keep
```
DECLARE_SOA_COLUMN(MlProbD0, mlProbD0, std::vector<float>);
DECLARE_SOA_COLUMN(MlProbD0bar, mlProbD0bar, std::vector<float>); 
```
for your new table and then just use `soa::Join` to combine `CF2ProngTracks` and your new table when you do the correlations in the task.

jaelpark:
Additionally, to support non-ML D0 selectors, `HfMlD0` needs to be optional. As discussed, two process could do the job with one with the original `HFCandidates` and then another `HFCandidatesWithML`. `processData` can be a template with `template<class HFCandidateType>` so that both can be supported. `constexpr if` together with `std::experimental::is_detected` works to automatically see when you need to output the ML scores.

victor-gonzalez (approved):

victor-gonzalez:
For the next time, you didn't need to close the previous PR
In fact the comments there, now lost, were also valuable
This one could have been just an update of the previous one

jaelpark:
I didn't quite mean to duplicate the entire `processData`, but okay, se can compact is someday.

victor-gonzalez:
> I didn't quite mean to duplicate the entire `processData`, but okay, we can compact is someday.

You are right @jaelpark! I overlooked it!
@jikim1290 Could you please take care of simplify the new `process` function to only produce the `ML` scores
You don't need to do it now with this PR but plan it for your next changes. Thanks!

PR #9851, opened 2025-02-10
[PWGCF] Fix bug in collision masker

ariedel-cern:
Fix bug in collision masker for the track-track pairing. Due to the bug, collisions with tracks which passed the selections were not properly masked.


alibuild (approved):

PR #9863, opened 2025-02-10
[PWGHF,Trigger] Add ITSPID slection on the deuteron of LcDe femto trigger

zhangbiao-phy:
hi @fgrosa, here is the PR for the ITSPID of the deuterons, let me know if you have any comments. Thanks a lot!


alibuild (approved):

PR #9881, opened 2025-02-11
[PWGDQ] Add Pole-mass to pseudoproper decay time and TOF-PID to cut library

shreyasiacharya:


iarsene (commented):
I suggest to add the new variable also in the fgVarNamesMap, so that it can be used in json cuts and histograms

iarsene (approved):

PR #9889, opened 2025-02-11
[PWGLF] initial commit for LStar v2

prottayCMT:
code to calculate elliptic flow for lstar resonance


dmallick2 (commented):
Please fix o2linter issue. Also define  structure and naming configuration set up accordingly, as well for TMath::functions and o2::constant. This fix is not major issue that can be ignored for task aiming for QM depend on urgent basis. It is always good to keep consistent.

njacazio (approved):

PR #9894, opened 2025-02-11
[PWGLF] Add PID flag and change main loop using two for loop

sangwoo184:
Add PID flag and change the loop of filling histogram using two for loop


smaff92 (approved):
Dear sangwoo,

Please, until next time, try to correct some of the warnings produced in the O2linter.
The main one would be to try to migrate away from ROOT classes, and use standard C++ classes instead.

Best Regards,
Adrian.N

PR #9885, opened 2025-02-11
[PWGHF] Fix ML variable addition to tree creator

atavirag:
This PR aims to fix the way HfCandD0Mls is filled. In the previous version, candidates saved as D⁰ and D⁰bar simultaneously were counted four times, two as D⁰ and two as D⁰bars, causing a mismatch between the number of candidates in HfCandD0Lites and HfCandD0Mls.

PR #9919, opened 2025-02-12
[PWGHF] Cleanup table filling in D0 tree creator

fcatalan92:
@vkucera @atavirag This is the followup of our discussion in https://github.com/AliceO2Group/O2Physics/pull/9885


atavirag:
Thank you, @fcatalan92! :)

PR #9937, opened 2025-02-13
[PWGUD] Task to test generatorID of UD MC production on hyperloop

rolavick:
Creating a task, which test MC production of standard tables by plotng some standard variables against generatorID


alibuild (approved):

PR #9940, opened 2025-02-13
[PWGLF] Use concepts everywhere, not only for MC

mpuccio:


maciacco (approved):

maciacco (approved):

mpuccio:
@singiamtel here alidoctor seems to have a problem on macOS, could you please have a look?

PR #9950, opened 2025-02-13
[Common] Fix event selection for not-anchored simulations

fmazzasc:
@ekryshen , the introduction of the `mapRCT` object broke the not-anchored simulations, where the default runNumber is set to 300000 and the `mapRCT` is not properly set. Does this trivial fix make sense to you? I checked it locally and it seems to be working.

### DIFF
--- a/Common/TableProducer/eventSelection.cxx
+++ b/Common/TableProducer/eventSelection.cxx
@@ -262,7 +262,7 @@ struct BcSelectionTask {
 
     int run = bcs.iteratorAt(0).runNumber();
 
-    if (run != lastRun && run >= 500000) {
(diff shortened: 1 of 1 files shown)


ekryshen:
@fmazzasc Thank you for spotting this issue. I am actually surprised that your fix works, because the requirement run>=500000 was set because we don't expect to find valid ccdb objects for unanchored simulations. May be run=300000 is a special case, and it will fail for run=499999, for example. I think I need to initialize the map properly for runs<500000. I'll try to fix it asap.

ekryshen:
Should be fixed in this PR: https://github.com/AliceO2Group/O2Physics/pull/10032

PR #9968, opened 2025-02-14
[PWGCF] adding ML selection in the correlations task

jikim1290:
adding ML selection in the correlations task


victor-gonzalez (commented):
@jikim1290 @jaelpark Didn't we ask for some changes on the derived data producer task in the previous PR?

jaelpark:
Mostly my request from last time regarding the `filter2Prong.cxx` was to not have excessively duplicated code across the two process functions. This can be avoided by templating the process function `HFCandidates` type and compile-time checking if the ML specific attributes are present with this templated type. The idea is similar to what is already done with `fillCorrelations` in `correlations.cxx`. But if it's not clear, I'm happy to adjust it slightly later myself also.

victor-gonzalez (approved):
OK! I approve it then for the time being

PR #9974, opened 2025-02-14
[PWGCF] debug, fix GFW dictionart problem

Luzhiyongg:
GFW_Core dictionary is broken when include AnalysisTask.h

### DIFF
--- a/PWGCF/GenericFramework/Core/GFWWeights.cxx
+++ b/PWGCF/GenericFramework/Core/GFWWeights.cxx
@@ -10,6 +10,7 @@
 // or submit itself to any jurisdiction.
 
 #include "GFWWeights.h"
+#include "TMath.h"
 #include <cstdio>
 
 GFWWeights::GFWWeights() : TNamed("", ""),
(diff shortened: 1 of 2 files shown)


PWGCF/GenericFramework/Core/GFWWeights.cxx
```diff
@@ -108,7 +109,7 @@ void GFWWeights::fill(double phi, double eta, double vz, double pt, double /*cen
   TH3D* th3 = reinterpret_cast<TH3D*>(tar->FindObject(getBinName(0, 0, pf))); // pT bin 0, V0M bin 0, since all integrated
   if (!th3) {
     if (!htype)
-      tar->Add(new TH3D(getBinName(0, 0, pf), ";#varphi;#eta;v_{z}", 60, 0, TwoPI, 64, -1.6, 1.6, 40, -10, 10)); // 0,0 since all integrated
+      tar->Add(new TH3D(getBinName(0, 0, pf), ";#varphi;#eta;v_{z}", 60, 0, TMath::TwoPi(), 64, -1.6, 1.6, 40, -10, 10)); // 0,0 since all integrated
```
victor-gonzalez, line 112: Any reason for not using `linter` recommendation?

victor-gonzalez (approved):

victor-gonzalez:
Interesting that this breaks the dictionary
This reinforce the `linter` rules that sometimes we don't get too seriously
One of the rules is that one should not use generic `namespace`s in header files

PWGCF/GenericFramework/Core/GFWWeights.cxx
Luzhiyongg: after remoce "AnalysisTask.h", there is no o2::constants::math namesapce so I have to replace the TwoPI with previous version.

Luzhiyongg:
I guess the reason for this breaks is that we include "AnalysisTask.h" in a file which do not have a task.
Then O2 will fail to fetch a task object from this file and then breaks

PWGCF/GenericFramework/Core/GFWWeights.cxx
victor-gonzalez: Well, a `using o2::constants::math` clause at the beginning of the source file or `o2::constants::math::TwoPI` directly, would do the work

victor-gonzalez:
> I guess the reason for this breaks is that we include "AnalysisTask.h" in a file which do not have a task. Then O2 will fail to fetch a task object from this file and then breaks

Right! Thanks!

PR #9977, opened 2025-02-14

yashpatley:
Modified reco and gen level processes.
They are combined into one process function with a method that select gen event that is uniquely associated to only one reco event that passes the event selection criterion so we have same number of events at reco and gen level.

### DIFF
--- a/PWGCF/TwoParticleCorrelations/Tasks/lambdaR2Correlation.cxx
+++ b/PWGCF/TwoParticleCorrelations/Tasks/lambdaR2Correlation.cxx
@@ -23,6 +23,7 @@
 #include "Framework/ASoAHelpers.h"
 #include "Framework/runDataProcessing.h"
 #include "PWGLF/DataModel/LFStrangenessTables.h"
+#include "PWGLF/DataModel/mcCentrality.h"
 #include "CommonConstants/PhysicsConstants.h"
 #include "Common/Core/RecoDecay.h"
 #include "CCDB/BasicCCDBManager.h"
@@ -134,8 +135,7 @@ using LambdaMcGenTrack = LambdaMcGenTracks::iterator;
 enum CollisionLabels {
   kTotColBeforeHasMcCollision = 1,
   kTotCol,
-  kPassSelCol,
-  kPassMultSelCol
+  kPassSelCol
 };
 
 enum TrackLabels {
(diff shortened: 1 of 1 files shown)


PWGCF/TwoParticleCorrelations/Tasks/lambdaR2Correlation.cxx
```diff
@@ -1074,6 +1041,33 @@ struct LambdaTableProducer {
...
+    }
+    histos.fill(HIST("McGen/h1f_collisions_info"), kTotCol);
+    // Check the reco collision
+    if (!collisions.begin().has_mcCollision() || !selCollision<run>(collisions.begin()) || collisions.begin().mcCollisionId() != mcCollision.globalIndex()) {
+      return;
+    }
+    histos.fill(HIST("McGen/h1f_collisions_info"), kPassSelCol);
+    histos.fill(HIST("McGen/h2f_collision_posZ"), mcCollision.posZ(), collisions.begin().posZ());
+    auto v0Tracks = V0s.sliceBy(perCollision, collisions.begin().globalIndex());
+    fillLambdaRecoTables<run, dmc>(collisions.begin(), v0Tracks, tracks);
```
victor-gonzalez, line 1064: Have you checked that locally you obtain what you expect regarding the reconstructed V0?
In here the `mcParticles` are grouped by `mcCollision` but the `V0`s are not grouped neither are the `tracks`

victor-gonzalez (approved):
I approve it but have a look at my comment

PWGCF/TwoParticleCorrelations/Tasks/lambdaR2Correlation.cxx
yashpatley: Yes Victor, it was not giving right multiplicity when I started in this direction. But after that "sliceBy", it worked. I think that is grouping the V0s associated to that collision. Well, I need to check it on Hyperloop, because I do not have much statistics locally. But the results seemed okay !

victor-gonzalez: > Yes Victor, it was not giving right multiplicity when I started in this direction. But after that "sliceBy", it worked. I think that is grouping the V0s associated to that collision. Well, I need to check it on Hyperloop, because I do not have much statistics locally. But the results seemed okay !

All right! I overlooked that `sliceBy`! It will work!

PR #9999, opened 2025-02-17
[PWGJE] - updating stat prompt photon 

smaff92:
Update of stat prompt photon code to incorporate raw data-level analysis


PWGJE/Tasks/statPromptPhoton.cxx
```diff
@@ -128,26 +206,73 @@ struct statPromptPhoton {
...
-  Filter mcPosZFilter = nabs(aod::mccollision::posZ) < cfgVtxCut;
+  Filter PosZFilter = nabs(aod::jcollision::posZ) < cfgVtxCut;
+  Filter mcPosZFilter = nabs(aod::jmccollision::posZ) < cfgVtxCut;
 
   using MCCells = o2::soa::Join<aod::Calos, aod::McCaloLabels_001>;
-
   using MCClusters = o2::soa::Join<o2::aod::EMCALMCClusters, o2::aod::EMCALClusters>;
-  using selectedCollisions = soa::Join<aod::Collisions, aod::EvSels>;
-  using selectedMCCollisions = aod::McCollisions;
+  using Clusters = o2::aod::EMCALMCClusters;
```
fjonasALICE, line 248: @nzardosh is this one the correct one to use for derived JE data? Not sure but we so far anyways do not have derived data for MC? so should be fine?

PWGJE/Tasks/statPromptPhoton.cxx
```diff
@@ -374,8 +509,71 @@ struct statPromptPhoton {
...
+                    momindex3 = photon_momom.globalIndex();
+                    momstatus3 = photon_momom.getGenStatusCode();
+                  }
+                }
+                break;
+              }
+            } // 3rd photon loop
+          } // 2nd photon check
+
+          std::cout << "We have a GEN prompt photon" << std::endl;
```
fjonasALICE, line 561: you probably do not want to havethese debug messages for every photon when running on hyperloop

PWGJE/Tasks/statPromptPhoton.cxx
```diff
@@ -374,8 +509,71 @@ struct statPromptPhoton {
...
+        int momstatus1 = 0;
+        for (auto& photon_mom : mcPhoton.mothers_as<aod::JMcParticles>()) {
+          if (mompdg1 == 0) {
+            mompdg1 = photon_mom.pdgCode();
+            momindex1 = photon_mom.globalIndex();
+            momstatus1 = photon_mom.getGenStatusCode();
+          }
+        } // first photon loop
+
+        if (std::fabs(mompdg1) < 40 && std::fabs(mompdg1) > 0) {
```
fjonasALICE, line 527: can you explain the loop logic? it is not fully clear to me

PWGJE/Tasks/statPromptPhoton.cxx
```diff
@@ -415,37 +604,92 @@ struct statPromptPhoton {
...
+    }
     histos.fill(HIST("REC_nEvents"), 2.5);
 
+    bool noTrk = true;
+    for (auto& track : tracks) {
+      auto ogtrack = track.track_as<TrackCandidates>();
+      if (!trackSelection(ogtrack)) {
+        continue;
+      }
+      if (!ogtrack.isGlobalTrack()) {
```
fjonasALICE, line 627: since you are in any case using derived data, I suggest to use the track selection from the jeutilities and make the type of track you want to filter on a configurable

PWGJE/Tasks/statPromptPhoton.cxx
```diff
@@ -455,104 +699,218 @@ struct statPromptPhoton {
...
+          phitrigger = true;
+        }
+
+        if (chargetrigger) {
+          histos.fill(HIST("REC_Track_v_Cluster_Phi_C"), phidiff);
+          histos.fill(HIST("REC_Track_v_Cluster_Eta_C"), etadiff);
+          histos.fill(HIST("REC_Track_v_Cluster_Phi_Eta_C"), phidiff, etadiff);
+        } else {
+          if (etatrigger && chargetrigger) {
+            std::cout << "????????????????????" << std::endl;
```
fjonasALICE, line 796: remove cout

PWGJE/Tasks/statPromptPhoton.cxx
```diff
@@ -571,20 +947,65 @@ struct statPromptPhoton {
           }
         } // photon check
       } // clusterparticle loop
+      /*
```
fjonasALICE, line 950: remove?

PWGJE/Tasks/statPromptPhoton.cxx
```diff
@@ -571,20 +947,65 @@ struct statPromptPhoton {
...
         histos.fill(HIST("REC_Trigger_Energy_FAKE"), mccluster.energy());
+        histos.fill(HIST("REC_Trigger_Purity_v_Energy"), 2.5, mccluster.energy());
       }
     } // cluster loop
 
+    auto bc = collision.bc_as<BcCandidates>();
+    int rnr = bc.runNumber();
+
+    std::string rnrstring = std::to_string(rnr);
+    if (runs.find(rnrstring) == std::string::npos) {
```
fjonasALICE, line 980: remove debug output

PWGJE/Tasks/statPromptPhoton.cxx
```diff
@@ -593,6 +1014,31 @@ struct statPromptPhoton {
       if (!trackSelection(ogtrack)) {
         continue;
       }
+      if (!ogtrack.isGlobalTrack()) {
```
fjonasALICE, line 1017: see previous comment about track selection

PWGJE/Tasks/statPromptPhoton.cxx
```diff
@@ -623,6 +1069,246 @@ struct statPromptPhoton {
 
   PROCESS_SWITCH(statPromptPhoton, processMCRec_JE, "processJE  MC data", false);
 
+  int nEventsData = 0;
+  void processData(jfilteredCollisions::iterator const& collision, jfilteredClusters const& clusters, jDataTrackCandidates const& tracks, soa::Join<aod::JTracks, aod::JTrackExtras, aod::JTrackPIs> const& caltracks, TrackCandidates const&, BcCandidates const&)
+  {
+
+    nEventsData++;
+    if ((nEventsData + 1) % 10000 == 0) {
```
fjonasALICE, line 1077: remove debug output, not needed when on hyperloop

PWGJE/Tasks/statPromptPhoton.cxx
```diff
@@ -623,6 +1069,246 @@ struct statPromptPhoton {
...
+    }
+    histos.fill(HIST("DATA_nEvents"), 2.5);
+
+    bool noTrk = true;
+    for (auto& track : tracks) {
+      auto ogtrack = track.track_as<TrackCandidates>();
+      if (!trackSelection(ogtrack)) {
+        continue;
+      }
+      if (!ogtrack.isGlobalTrack()) {
```
fjonasALICE, line 1103: see previous comment about track selection

PWGJE/Tasks/statPromptPhoton.cxx
```diff
@@ -623,6 +1069,246 @@ struct statPromptPhoton {
...
+        double etaC = cluster.eta();
+        double phiT = ogtrack.trackPhiEmcal();
+        double phiC = cluster.phi();
+        double ptT = ctrack.pt();
+        bool etatrigger = false;
+        bool phitrigger = false;
+        double phidiff = TVector2::Phi_mpi_pi(cluster.phi() - ogtrack.phi());
+        double etadiff = cluster.eta() - ogtrack.eta();
+
+        if (fabs(etaT - etaC) < (0.010 + pow(ptT + 4.07, -2.5))) {
```
fjonasALICE, line 1189: suggestion: make configurable

PWGJE/Tasks/statPromptPhoton.cxx
```diff
@@ -623,6 +1069,246 @@ struct statPromptPhoton {
...
+          etatrigger = true;
+        }
+
+        if (fabs(TVector2::Phi_mpi_pi(phiT - phiC)) < (0.015 + pow(ptT + 3.65, -2.0))) {
+          phitrigger = true;
+        }
+
```
fjonasALICE, line 1196: E/p veto for track matching missing?

fjonasALICE (dismissed):
Hi,
I left some comments for your consideration. After they are implemented it should be okay for a first approval @nzardosh should check as well.
Best,
Florian

PWGJE/Tasks/statPromptPhoton.cxx
smaff92: No, E/p veto for track matching is later down the code (1213-1223)

PWGJE/Tasks/statPromptPhoton.cxx
smaff92: Done

PWGJE/Tasks/statPromptPhoton.cxx
smaff92: Done

PWGJE/Tasks/statPromptPhoton.cxx
smaff92: I would prefer not to do this. I prefer the flexibility to also be able to run over non-derived data.
Furthermore, I find this solution better when later on doing systematics on track quality, instead of using a magnitude of several restricted filters

PWGJE/Tasks/statPromptPhoton.cxx
smaff92: Done

PWGJE/Tasks/statPromptPhoton.cxx
smaff92: I am currently anyway not utilizing these non-JE derived classes, so it should be fine if I also remove it.

PWGJE/Tasks/statPromptPhoton.cxx
nzardosh: For this i would agree with Florian. If the only reason you need the original track table is track selection then you should instead use the je utilities and remove the subscription to the original tracks. Unless there is a type of track selection you need which we dont have (we can also add that to the JE framework). Using the JE utilities doesnt mean you are limited to the derived data, as you can still run over the original aod but it means you have the possibility of running over the derived too. If you keep the original tracks table you cant run over derived data

PWGJE/Tasks/statPromptPhoton.cxx
nzardosh: For this we have a JE derived mode which is here:
https://github.com/AliceO2Group/O2Physics/blob/af190f20700d5e6323309f08789c99f4bd4c404b/PWGJE/DataModel/Jet.h#L208C1-L208C6
so this should be used

smaff92: These non-JE cluster classes are now removed

PWGJE/Tasks/statPromptPhoton.cxx
smaff92: I must insist on using these track tables for the track cuts at this moment. I've already found that this is analysis is extremely sensitive to cuts (especially absence of a geometric TPC cut), and I need more statistics to properly QA different selections, especially for data.

I will migrate to these je util filters once the "golden" cuts + suitable sys. variations have been found, but doing so now would simply be inefficient. I would have to add potentially 20-30+ filter variations that in a couple of months might be redundant.

smaff92: For now, I have to insist on using the original track table. 

I've already found that this analysis is extremely sensitive to the track selection used, (geometrical TPC cut is required, and the sensitivity of this cut is correlated to TPC crossed-rows etc) and I need more statistics to do proper QA in order to investigate which configuration of cuts work best.

It would at this point be inefficient use of my time to transfer to the je utilities, as I at this point don't know what kind of selection I would like to have, which would mean that I have to add 20-30 track selections to the je utilities that might be redundant in a couple of weeks.
 
Once the "golden cuts" + suitable systematic variations are found for the analysis, I will migrate to the je utilities and add any required track selection to the JE framework (as well as adding a flag for a built-in geometric TPC cut)

nzardosh: If we can set a time, lets say a month where you can find the correct tracking cuts and then move to je utilities i think that is fine. But it would be quite important to do so as soon as possible, particularly before any analysis level long trains

smaff92: Yep, one month would be more than enough, I would essentially need like 10-20 or so subwagons running over the small apass7 dataset + small Jet-Jet anchored MC to estimate this. So one month would be fine for me.

nzardosh:
Florian let me know when you are happy with the PR

fjonasALICE:
@nzardosh happy with the changes, you can approve!

nzardosh (approved):

PR #10003, opened 2025-02-17
[PWGLF] moved UE calculation outside of loop on tracks

alcaliva:


maciacco:
@alcaliva can we close this one? As I understand, the content of this PR was implemented in https://github.com/AliceO2Group/O2Physics/pull/10017

PR #10015, opened 2025-02-17
Updates in HF filters

fgrosa:
@zhangbiao-phy @mpuccio this PR:
- fixes a couple of mistakes in the Xi+bachelors triggers
- adds the H_b -> JPsi(->mumu) + X triggers
- fixes a small bug in the deuteron selection with ITS PID (or at least this is what looked to me, but please @zhangbiao-phy take a look)

For the new B triggers, I added one flag per channel, even if there is quite large overlap among them. @mpuccio let me know if this is fine!


zhangbiao-phy:
> @zhangbiao-phy @mpuccio this PR:
> 
> * fixes a couple of mistakes in the Xi+bachelors triggers
> * adds the H_b -> JPsi(->mumu) + X triggers
> * fixes a small bug in the deuteron selection with ITS PID (or at least this is what looked to me, but please @zhangbiao-phy take a look)
> 
> For the new B triggers, I added one flag per channel, even if there is quite large overlap among them. @mpuccio let me know if this is fine!

Hi @fgrosa, Sorry for that mistake! Thanks for spotting that bug! Looks good to me for your changes now

mpuccio (approved):

PR #10024, opened 2025-02-17
[PWGLF] Add findable mode for new strangeness builder

ddobrigk:
Adds a test version of the strangeness builder in which two findable modes exist: 1) generate complementary findables and add to existing V0s and 2) generate only findables, disregard existing V0s. The addition is temporarily provided in a new task so that large-scale memory and CPU tests can be performed. The new task will be removed and the changes merged with the regular `strangenessbuilder` soon.


njacazio (approved):

njacazio:
Thanks a lot @ddobrigk please fix the megalinter in a coming PR

PR #10040, opened 2025-02-18
[PWGHF] change the tag value of reflection case for D0 bar in Tcc tree creator

zhangbiao-phy:
Changing the tag value of reflection case for D0 bar in Tcc tree creator for avoiding double counting of reflection


alibuild (approved):

PR #10052, opened 2025-02-18
[PWGHF] Fixed bug on resonance data creator

Luca610:
This PR fixes a bug on the resonance data creator workflow with ML


fgrosa (approved):
Thanks for spotting the bug!

PR #10055, opened 2025-02-18
[PWGHF] Fix and improve D+ selection in derived-data creator

vkucera:
- Prefilter candidates which passed cuts at least until the step before ML.
- Reject candidates which did not pass ML immediately inside the candidate loop when ML selection is required.

FYI: @nzardosh @DelloStritto


alibuild (approved):

PR #10063, opened 2025-02-19
[PWGHF] Add table with ML outputs for reco MC

Marcellocosti:
This PR adds a table to the `treeCreatorDplusToPiKPi.cxx`, containing ML scores, mass, pt and centrality for MC signal candidates. This allows a more detailed study of the correlated backgrounds visible in the D+ invariant mass spectrum. Tagging also @stefanopolitano, @fgrosa, @fchinu as you may be interested. Please let me know if you need further entries in the table! 

Edit: As asked, the D* has been added to the candidate creator 3 prongs, using the same structure as for the other particles but avoiding to insert the decay channel in the enum used for the other 3-prong candidates, as this would require a modification of the trackIndexSkimCreator. Hope this solution is fine for you.

### DIFF
--- a/PWGHF/TableProducer/candidateCreator3Prong.cxx
+++ b/PWGHF/TableProducer/candidateCreator3Prong.cxx
@@ -85,7 +85,7 @@ struct HfCandidateCreator3Prong {
   Configurable<bool> createDs{"createDs", false, "enable Ds+/- candidate creation"};
   Configurable<bool> createLc{"createLc", false, "enable Lc+/- candidate creation"};
   Configurable<bool> createXic{"createXic", false, "enable Xic+/- candidate creation"};
-  Configurable<bool> createDstarDplusBkg{"createDstarDplusBkg", false, "enable D* candidate creation"};
+  Configurable<bool> createDstarToPiKPiBkg{"createDstarToPiKPiBkg", false, "enable D* candidate creation"};
 
   HfEventSelection hfEvSel;        // event selection and monitoring
   o2::vertexing::DCAFitterN<3> df; // 3-prong vertex fitter
@@ -145,7 +145,7 @@ struct HfCandidateCreator3Prong {
       }
     }
 
-    std::array<bool, 5> creationFlags = {createDplus, createDs, createLc, createXic, createDstarDplusBkg};
+    std::array<bool, 5> creationFlags = {createDplus, createDs, createLc, createXic, createDstarToPiKPiBkg};
     if (std::accumulate(creationFlags.begin(), creationFlags.end(), 0) == 0) {
       LOGP(fatal, "At least one particle specie should be enabled for the creation.");
     }
@@ -792,7 +792,7 @@ struct HfCandidateCreator3ProngExpressions {
   bool createDs{false};
   bool createLc{false};
   bool createXic{false};
-  bool createDstarDplusBkg{false};
+  bool createDstarToPiKPiBkg{false};
 
   HfEventSelectionMc hfEvSelMc; // mc event selection and monitoring
   using BCsInfo = soa::Join<aod::BCs, aod::Timestamps, aod::BcSels>;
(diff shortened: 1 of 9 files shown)


PWGHF/TableProducer/treeCreatorDplusToPiKPi.cxx
```diff
@@ -79,8 +81,20 @@ DECLARE_SOA_COLUMN(Ct, ct, float);
...
-
+DECLARE_SOA_TABLE(HfCandDpMls, "AOD", "HFCANDDPML",
+                  full::M,
+                  full::Pt,
+                  full::Centrality,
+                  full::BkgScore,
+                  full::FdScore,
+                  hf_cand_3prong::FlagMcMatchRec,
+                  hf_cand_3prong::OriginMcRec,
+                  hf_cand_3prong::FlagMcDecayChanRec)
```
fgrosa, line 97: Why don't we just put the BDT scores here? In this case we would have a table which is joinable with the `Lite` one and we do not duplicate information

fgrosa (changes requested):
Hi @Marcellocosti, thanks!
I have a comment about the new table, I would avoid to repeat variables and I would only add the BDT scores (anyway it is joinable with the `Lite` table, so you can retrieve all the other information from there)

I would then suggest you to add a flag for D* -> PiKPi in the creator3prongs, adding another block like the existing ones, see e.g. https://github.com/AliceO2Group/O2Physics/blob/99b5f0299e8e42fba6b4937e8c3a4ccc65ffffab/PWGHF/TableProducer/candidateCreator3Prong.cxx#L996-L1010.
This requires to add a bit in the enum for 3 prongs for the D* -> PiKPi here https://github.com/AliceO2Group/O2Physics/blob/99b5f0299e8e42fba6b4937e8c3a4ccc65ffffab/PWGHF/DataModel/CandidateReconstructionTables.h#L950

Also be careful that `N3ProngDecays` is used also in the `trackIndexSkimCreator` and there we should not add a channel.

fgrosa (approved):
I remove the "Request changes" mark since I will be offline in the next few days to avoid blocking this PR, but this doesn't mean that it can be merged as it is ;-)

PWGHF/TableProducer/candidateCreator3Prong.cxx
```diff
@@ -85,6 +85,7 @@ struct HfCandidateCreator3Prong {
   Configurable<bool> createDs{"createDs", false, "enable Ds+/- candidate creation"};
   Configurable<bool> createLc{"createLc", false, "enable Lc+/- candidate creation"};
   Configurable<bool> createXic{"createXic", false, "enable Xic+/- candidate creation"};
+  Configurable<bool> createDstarToPiKPiBkg{"createDstarToPiKPiBkg", false, "enable D* candidate creation"};
```
fgrosa, line 88: To avoid a new configurable and considering that this channel doesn't imply to create candidates that have a different flag than the D+ from the track index skim creator (which was the original idea behind the `create` configurables, to reduce the CPU time in Pb-Pb) I would simply enable the MC matching of D* everytime the D+ is enabled. Also, we could consider to do the MC matching for all the channels regardless the `create` configurables (which in the end arr important to pre-select the candidates in the spirit of reducing the CPU time)

fgrosa (changes requested):
Hi @Marcellocosti, only one remaining comment and then good for me, thanks!

fgrosa (approved):

vkucera:
@Marcellocosti Please fix the issues introduce by this PR.

vkucera:
> Dear @vkucera, can you please point me to the issues you are referring to? I tested the pr locally and everything looked fine.

I am referring to the issues reported by the MegaLinter test in this PR.

PWGHF/TableProducer/candidateCreator3Prong.cxx
vkucera: I'm not sure I follow the reasoning behind removing the `create` flags. Enabling matching only for analysed species still does save CPU, especially in `candidateCreatorMcGen.cxx` when we definitely not need all species to be matched.

PR #8443, opened 2024-11-14
[PWGHF] Add Non Hfe selection task in HfeSelection task

rashigupt:

PR #10064, opened 2025-02-19
[PWGHF] Add Non-HFE Selection and Correlation with Hadron

rashigupt:
This pull request introduces  for selecting Non-HFE electrons and calculating their correlation with hadrons, specifically for both the dataset and Monte Carlo (MC) generation.


vkucera:
@rashigupt This is a duplicate of https://github.com/AliceO2Group/O2Physics/pull/8443. Please avoid this, as it is wasting collaboration resources and people's time.

PR #10065, opened 2025-02-19
[PWGUD] Study of nContribs per ITSROFs vs nCollisions per ITSROF

rolavick:
Adding plot to count PV contributors per ITSROF against n collisions per ITSROF


alibuild (approved):

PR #10095, opened 2025-02-19
[PWGCF] Fix bug in Data Model for FemtoDream

lauraser:
Does not affect any results for FemtoDream users as the px and py values are not used in the analysis.


alibuild (approved):

PR #10102, opened 2025-02-20
[Infrastructure] Bump MegaLinter to v8.4.2

vkucera:
Also disable black, flake8, isort because they are replaced by ruff.


vkucera:
@ddobrigk Ping

PR #10134, opened 2025-02-21
Fix abs to std::abs

lauraser:


alibuild (approved):

PR #10135, opened 2025-02-21
[Common] Fill multNContribs variables for Run 2 converted data

romainschotter:
Fill multNContribs variables for Run 2 converted data, allowing to apply INEL > 0, > 1, etc in consistent manner as what we do in Run 3


ddobrigk (approved):
@romainschotter thanks! (also for taking care of O2 linter) 

Harmless for Run 3, and in fact this just fixes a missing variable for Run 2 - approving.

PR #10154, opened 2025-02-23
[PWGJE] Adding event weight to THnSparse

hahassan7:
Adding the event weight to the THnSparse. In Addition to formatting changes.

### DIFF
--- a/PWGJE/Core/JetTaggingUtilities.h
+++ b/PWGJE/Core/JetTaggingUtilities.h
@@ -63,50 +63,50 @@ namespace jettaggingutilities
 const int cmTomum = 10000; // using cm -> #mum for impact parameter (dca)
 
 struct BJetParams {
-  float JetpT = 0.0;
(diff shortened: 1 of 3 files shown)


vkucera:
@hahassan7 Please fix the name of `JetTaggingUtilities.h` to `jetTaggingUtilities.h`, as reported by the O2 linter.

hahassan7:
@vkucera , it worked. But then the name would be different from the other files in `PWGJE/Core/` I think it would be better to make them consistent. What do you think ? I don't really want to change the name of the other files. I can only change this one.

vkucera:
> @vkucera , it worked. But then the name would be different from the other files in `PWGJE/Core/` I think it would be better to make them consistent. What do you think ? I don't really want to change the name of the other files. I can only change this one.

It can be fixed all at once in a separate PR.

nzardosh (approved):

PR #10164, opened 2025-02-24
[PWGJE] Fix a bug in JetMatching function

wefeng1110:
Fix a bug in JetMatching (with bkg subtraction) function and change few histograms' name


wefeng1110:
Hi, I would like to do a change in my processQC function, please do not approved for the moment. Thanks!

nzardosh (approved):

PR #10170, opened 2025-02-24
[PWGDQ] Adding minitree to esimate offline Axe

lucamicheletti93:
@iarsene @MauriceCoke 

These 2 extra trees are added for:
-  the offline calculation of the Axe
- to perform iterative tuning of the Axe as done in Run 2 analyses

Some checks is necessary on hyperloop on large samples


alibuild (approved):

PR #10184, opened 2025-02-25
[PWGCF] Adding 2 and 4 particle flow for resonance

Preet-Bhanjan:


PWGCF/Flow/Tasks/resonancesGfwFlow.cxx
```diff
@@ -248,7 +336,7 @@ struct ResonancesGfwFlow {
   }
 
   template <typename TTrack, typename vector, char... chars>
```
victor-gonzalez, line 338: Why are you templating on `vector`?

PWGCF/Flow/Tasks/resonancesGfwFlow.cxx
```diff
@@ -271,10 +359,20 @@ struct ResonancesGfwFlow {
       minusdaug = ROOT::Math::PxPyPzMVector(partminus.px(), partminus.py(), partminus.pz(), plusmass);
       mom = plusdaug + minusdaug;
 
+      double pt = mom.Pt();
+      double invMass = mom.M();
```
victor-gonzalez, line 363: I would suggest using the methods in `Common/Core/RecoDecay.h` they are by far more efficient than using `PxPyPzMVector`

PR #10193, opened 2025-02-25
[PWGJE] Adding shift in rho for reference events

dajones2:


dajones2:
Can this be merged please

PR #10200, opened 2025-02-26
Allows to fill Run 2 centrality tables with dummy values

romainschotter:
Allows to fill Run 2 centrality tables with dummy values in case the converted calibration objects do not exist on ccdb


ddobrigk (approved):
thanks for adding this extra option @romainschotter

PR #10231, opened 2025-02-27
Fix missing std prefix

vkucera:


alibuild (approved):

PR #10254, opened 2025-02-28
[PWGLF] AngularCorrelationsInJets.cxx: added MC rec process using jet tables

ljoergen:


ljoergen:
I'll fix the remaining O2 linter complaints, didn't catch them all apparently

maciacco (approved):

PR #10297, opened 2025-03-03
[PWGDQ] Added truth multiplicity in the table-maker-mc

Debadatta3337:
[PWGDQ] Added truth multiplicity in the table-maker-mc


iarsene:
In addition to the comments i made to the specific places where your changes are, you have to produce the extra MC multiplicity table in the skimMCCollisions function. That is just trivial, becase if you use the join of MCCollisions and the MC extra mult table, the columns you need are available as methods to the existing event table argument.

PR #10320, opened 2025-03-04
[Common] Final non-linarity fix

peressounko:


peressounko:
Dear @ivorobye could you please approve this PR? We found another bug in previous parameterisaion...

ivorobye:
hi @peressounko, I can approve this PR, but as it was also mentioned in the previous case, you should seriously consider to implement these parameters as configurables, to avoid exactly this kind of fixes in the future; also, are you sure you need the vertex position in LOG(info)?

peressounko:
dear @ktf could you please approve this PR from core side? Thanks, Dmitri

PR #10341, opened 2025-03-05
PWGHF: Add track quality selection in D0 selector

Mingyu3360715:


fcatalan92 (changes requested):
Hi @Mingyu3360715, looking at the builder error, you have to update the selection table definition to accommodate the new column. In general, it would be good to move to a bitmap at a certain point, to avoid the need of adding new flags each time.

PR #10347, opened 2025-03-05
[PWGEM/Dilepton] fix uint8_t and uint32_t

dsekihat:
PWGEM/Dilepton: fix uint8_t and uint32_t


alibuild (approved):

PR #10404, opened 2025-03-08
[PWGCF] Fix hEventCounter histogram

ilikmeta:
Apply last comments (Thank you) and take out of doProcessData the EventCounter histogram.


ilikmeta:
@victor-gonzalez Following your comments, I removed PDG_t:: prefix, no problems on the local test, but here Megalinter failed due to ITS PID leaks. Then, I put back the prefix, but still Megalinter fails. Not sure how to fix this.

victor-gonzalez (approved):
MegaLinter error does not have anything to do with your PR
So you can safely remove the prefixes in the next iteration

PR #10428, opened 2025-03-10
[PWGDQ] Added option to reject collision splitting candidates

iarsene:
table-reader-with-assoc: event-selection
a) added additional protection in the init function for when the dummy function is enabled
table-reader-with-assoc: same-event-pairing
a) option to reject split collision candidates, as determined by the event selection task via two-event correlations;
b) added additional process function for barrel only analysis, but with extra multiplicity estimators being available, mainly for TPC occupancy checks


alibuild (approved):

PR #10430, opened 2025-03-10
Update 3body EM and reduced data creator

wang-yuanzhe:


wang-yuanzhe:
@creetz16

wang-yuanzhe:
For failing checks of o2_linter, the name of configurables/variables to be changed. The part of mc analysis will be rearranged soon.

vkucera:
@wang-yuanzhe You also have cpplint errors. Those should be fixed before merging.

wang-yuanzhe:
> @wang-yuanzhe You also have cpplint errors. Those should be fixed before merging.

Thanks a lot for the reminder! For the cpplint errors, what we would like to do is to keep the struct of the reduced data as same as the one in AnalysisDataModel.h. Therefore we just copied this part from the header file in O2 to avoid any potential mistakes. It is fine?

vkucera:
> > @wang-yuanzhe You also have cpplint errors. Those should be fixed before merging.
> 
> Thanks a lot for the reminder! For the cpplint errors, what we would like to do is to keep the struct of the reduced data as same as the one in AnalysisDataModel.h. Therefore we just copied this part from the header file in O2 to avoid any potential mistakes. Is it fine?

Why don't you use that column directly in your tables then?

wang-yuanzhe:
> > > @wang-yuanzhe You also have cpplint errors. Those should be fixed before merging.
> > 
> > 
> > Thanks a lot for the reminder! For the cpplint errors, what we would like to do is to keep the struct of the reduced data as same as the one in AnalysisDataModel.h. Therefore we just copied this part from the header file in O2 to avoid any potential mistakes. Is it fine?
> 
> Why don't you use that column directly in your tables then?

Thank for your advice! The cpplint errors are fixed now.

wang-yuanzhe:
@njacazio @skundu692 Could you have a look? Thanks!

njacazio (approved):

PR #10433, opened 2025-03-10
[PWGCF] Remove rapidity cut and update efficiency correction

TGahlaut1:


victor-gonzalez (approved):
Please, remember not leaving code commented

PR #10451, opened 2025-03-11
[DPG] adding two proccess to save the trackQA variables to the standar…

amaringarcia:
…d derived datafor PID


amaringarcia:
@alcaliva would it be possible for you to check this pull request quickly? 
Thanks in advance.

alcaliva (approved):

PR #10460, opened 2025-03-12
[PWGCF] Pair Fractions

mm1707:
Adding process filling histogram of mother particle PDG of pair obtained while mixing to get Pair Fractions


prchakra (approved):

ktf:
@davidrohr there is a number of checks which fail on macOS with errors in the GPU Cmake, similar (but not unique) to what happens here. Could you have a look?

PR #10465, opened 2025-03-12
[ALICE3] Add hit efficiency correction to the fasttracker

jesgum:
Adding the hit efficiency corrections used in the SolveTrack method in Delphes that is used to produce the LUTs for the OTF simulations 

link: https://github.com/AliceO2Group/DelphesO2/blob/master/src/DetectorK/DetectorK.cxx#L1421


jesgum:
Hi @njacazio @hscheid! Could you have a look when you have a moment? Thanks!

njacazio (approved):

PR #10473, opened 2025-03-12
[PWGCF] DptDpt - Incorporating ITS only track selection

victor-gonzalez:


victor-gonzalez:
> @victor-gonzalez Please remove the `disable=name/function-variable` comment. It is not needed.

@vkucera  Thanks!

PR #10494, opened 2025-03-14
[PWGCF] updates on Systematic checking

zhengqi-w:
updates on Systematic checking

### DIFF
--- a/PWGCF/Flow/Tasks/pidcme.cxx
+++ b/PWGCF/Flow/Tasks/pidcme.cxx
@@ -86,6 +86,8 @@ struct FillPIDcolums {
   Configurable<float> cfgMaxChi2NClITS{"cfgMaxChi2NClITS", 2.5, "Maximum chi2 per cluster ITS for PID if not use costom track cuts"};
   Configurable<float> cfgMinTPCCls{"cfgMinTPCCls", 70, "Minimum TPC clusters for PID if not use costom track cuts"};
   Configurable<float> cfgMinITSCls{"cfgMinITSCls", 1, "Minimum ITS clusters for PID if not use costom track cuts"};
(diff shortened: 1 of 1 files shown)


shouqiye (approved):

PWGCF/Flow/Tasks/pidcme.cxx
```diff
@@ -1667,7 +1924,7 @@ struct QAProcessCent {
   }
 };
 
-struct pidcme { // o2-linter: disable=name/struct
+struct pidcme { // o2-linter: disable=name/struct(keep the saving dir name for offline analysis)
```
vkucera, line 1927: What does the comment mean?

zhengqi-w: Hi vkucera, thanks for your comment. If I wang to correct this O2 Linter warning, I need to rename my worflow in both .cxx file and CMakelist and also it will change the Analysisresults.root dir-name for my analysis,  which will causing much more online and offline effort for this analysis. As this analysis is now urgent for QM, I do not want to pay extra effort for this. By the way, I will correct it after QM next month for aftertoward using.

zhengqi-w: > What does the comment mean?
Hi vkucera, thanks for your comment. If I wang to correct this O2 Linter warning, I need to rename my worflow in both .cxx file and CMakelist and also it will change the Analysisresults.root dir-name for my analysis, which will causing much more online and offline effort for this analysis. As this analysis is now urgent for QM, I do not want to pay extra effort for this. By the way, I will correct it after QM next month for aftertoward using.

vkucera: OK, thanks!

PR #10497, opened 2025-03-14
[PWGHF] Add swapping info in XicToPKPi tree creator

phymanshu:


mfaggin (commented):
any strict reason to change `int8_t` into `int`?

mfaggin (approved):

phymanshu:
> any strict reason to change `int8_t` into `int`?

yes, the decision was based on combination of some local tests. Printing `swappedCand` values on terminal with `int8_t` there was nothing stored while it worked with `int`. I needed to have it on hyperloop quickly so I chose to stay with `int`.

mfaggin:
ok, I think that this is easily solvable by casting it into a `int` when printing

phymanshu:
> ok, I think that this is easily solvable by casting it into a `int` when printing

Thanks Mattia, I didnot explore it.

PR #10510, opened 2025-03-15
[PWGDQ] Changes to MCProng/MCSignal and dq-effieciency for generator level signal counting

iarsene:
Adding the HEPMC status code 11 (final state) in the MCProng source list; 
Fix a bug in the MCSignal check sources; 
Adding a function to count the generator level signal only for selected reconstructed collisions in Dilepton-Hadron task


alibuild (approved):

PR #10526, opened 2025-03-16
[PWGCF] DptDpt - Improving QC data layout for HL execution

victor-gonzalez:


alibuild (approved):

PR #10546, opened 2025-03-17
[PWGUD] Added new Process function and modified the tables

ananthapadmanabhan18:
- Added a new Process function `processDataFast` which uses Filtered UD tables instead of the full ones, and it only produces derived data except 2-3 histograms. 
-  Removed the configurables `fddaCut`, `fddcCut` and `StrictEventSelection ` which is not required any more
- Small modifications in code


ananthapadmanabhan18:
- Also added new process function for MC reconstruction with filters to speedup the process. 
- I am unable to fix the o2-linter error for the pdg of rho-prime-1700.
- FIxed the explicit pdg code for pions

rolavick (approved):

PR #10564, opened 2025-03-18
[PWGHF] Added flag to reject reflection for Lc at reco level

1994ra:


PWGHF/HFC/TableProducer/correlatorLcHadrons.cxx
```diff
@@ -527,6 +528,9 @@ struct HfCorrelatorLcHadrons {
       auto trackPos1 = candidate.template prong0_as<TracksWithMc>(); // positive daughter (negative for the antiparticles)
       auto trackPos2 = candidate.template prong2_as<TracksWithMc>();
       int8_t chargeLc = trackPos1.sign();                            // charge of 1st prong will be the charge of Lc candidate
+      auto mcParticleProng0 = candidate.template prong0_as<TracksWithMc>().template mcParticle_as<aod::McParticles>();
+      auto pdgCodeProng0 = std::abs(mcParticleProng0.pdgCode());
```
fgrosa, line 532: Since this is done for both signal and background, you should first check is the track `has_mcParticle()`
Also, you could consider to use the `IsCandidateSwapped` flag instead of accessing the particle table, which is more efficient https://github.com/AliceO2Group/O2Physics/blob/bf9dfac5dc045824fb82193a35df9cc00f865b45/PWGHF/DataModel/CandidateReconstructionTables.h#L1061

fgrosa (changes requested):

PR #10567, opened 2025-03-18
[PWGHF] add deviation of pi from Omega_c for Omega_c -> OmegaPi and remove the cuts on negative chi2

Yunfan-Liu:
- add deviation of pi from Omega_c from KFParticle

-  keep negative Chi2 and NDF from KFParticle


vkucera:
@Yunfan-Liu Please either add a correct prefix or do not add it at all.

Yunfan-Liu:
> @Yunfan-Liu Please either add a correct prefix or do not add it at all.

I'm very sorry about that. I'll be careful.

zhangbiao-phy (approved):

PR #10572, opened 2025-03-19
[PWGHF,Trigger] Add the possibility to use TPC PID in BDT for HFfilters + minor changes to selections

fgrosa:
@zhangbiao-phy @apalasciano in this PR I:
1) added the possibility to use TPC PID variables in the BDTs for Lc (and possibly Xic) in the software triggers
2) slightly modified some selections: for D0Pr and D*+Pr I added minimal track selections for the proton, while for the femto I modified the deuteron PID selections in order to be one sided only (for ITS or TPC we only keep > -Nsigma, while for TOF we only keep < Nsigma). Please note the signs, that imply that in the configurable the values should always be **without sign**


alibuild (approved):

PR #10586, opened 2025-03-20
[PWGCF] Fix a bug

fuchuncui:
Fix a bug about histogram name.


vkucera:
@fuchuncui The PR title is too vague.

PR #10587, opened 2025-03-20
Fix missing centrality axis in hEff histos

gbencedi:


alibuild (approved):

PR #10602, opened 2025-03-21
[PWGCF] Changed histogram Doubles to histogram Floats

JStaa:


JStaa:
@vkucera I don't believe so. The resolution of my histograms is not close to the precision limit. There is a concern I have regarding `float`s that may occur when two large histograms are subtracted, in which case I may hit the precision limit. Did you see anything in the code that concerned you?

vkucera:
> @vkucera I don't believe so. The resolution of my histograms is not close to the precision limit. There is a concern I have regarding `float`s that may occur when two large histograms are subtracted, in which case I may hit the precision limit. Did you see anything in the code that concerned you?

As soon as your bin count hits 16,777,216, your counter will not be able to represent any higher integer.

victor-gonzalez:
@JStaa , @vkucera is right, be careful generalizing the transition to `float`
For the correlation function for sure you will not hit precision issues
But for the other histograms you changed, I would suggest, in line of Vit, to consider them carefully
I will approve it for the time being but roll back the histograms you consider whenever you want

victor-gonzalez (approved):

PR #10610, opened 2025-03-21
[PWGHF] KFParticle based 3 prong decay reconstruction development

lubynets:
In this PR usage of KF Particle for 3 prong decay (in particular `Lc->PKPi`) reconstruction was developed as well as some changes needed for `Lc` lifetime measurement were implemented:
1. Application of topological and invariant mass constraints implemented.
2. Enabled application of cuts on Kf-based variables in the `candidateSelectorLc` task.
3. Added lifetime-relevant fields to the Mc particles table of the `treeCreatorLcToPKPi` task.
4. List of pid-related variables in the `treeCreatorLcToPKPi` task reduced to what is really needed according to the order of prongs in the candidate - `PKPi` or `PiKP`.


PWGHF/Core/SelectorCuts.h
```diff
@@ -493,6 +494,19 @@ constexpr double Cuts[NBinsPt][NCutVars] = {{0.4, 0.4, 0.4, 0.4, 0., 0.005, 0.,
                                             {0.4, 0.4, 0.4, 0.4, 0., 0.005, 0., 0., 0., 1e+10, -1.},  /* 12 < pT < 24 */
                                             {0.4, 0.4, 0.4, 0.4, 0., 0.005, 0., 0., 0., 1e+10, -1.}}; /* 24 < pT < 36 */
 
+// default value for the cuts                    Chi2Prim      Chi2Geo       DCA, cm     Chi2Geo  Chi2Topo
+//                                               P   K  Pi  KPi PPi  PK  KPi    PPi   PK   ↓  LdL  ↓
+constexpr double kfCuts[NBinsPt][NKfCutVars] = {{3., 3., 3., 3., 3., 3., 0.01, 0.01, 0.01, 3., 5., 5.},  /* 0  < pT < 1  */
```
vkucera, line 499:
```suggestion
constexpr double CutsKf[NBinsPt][NKfCutVars] = {{3., 3., 3., 3., 3., 3., 0.01, 0.01, 0.01, 3., 5., 5.},  /* 0  < pT < 1  */
```

PWGHF/Core/SelectorCuts.h
```diff
@@ -508,6 +522,7 @@ static const std::vector<std::string> labelsPt = {
 
 // column labels
 static const std::vector<std::string> labelsCutVar = {"m", "pT p", "pT K", "pT Pi", "Chi2PCA", "decay length", "cos pointing angle", "decLengthXY", "normDecLXY", "impParXY", "mass (Kpi)"};
+static const std::vector<std::string> labelsKfCutVar = {"kfChi2PrimPr", "kfChi2PrimKa", "kfChi2PrimPi", "kfChi2GeoKaPi", "kfChi2GeoPrPi", "kfChi2GeoPrKa", "kfDcaKaPi", "kfDcaPrPi", "kfDcaPrKa", "kfChi2Geo", "kfLdL", "kfChi2Topo"};
```
vkucera, line 525:
```suggestion
static const std::vector<std::string> labelsCutKfVar = {"kfChi2PrimPr", "kfChi2PrimKa", "kfChi2PrimPi", "kfChi2GeoKaPi", "kfChi2GeoPrPi", "kfChi2GeoPrKa", "kfDcaKaPi", "kfDcaPrPi", "kfDcaPrKa", "kfChi2Geo", "kfLdL", "kfChi2Topo"};
```

PWGHF/Core/SelectorCuts.h
```diff
@@ -465,6 +465,7 @@ namespace hf_cuts_lc_to_p_k_pi
 {
 static constexpr int NBinsPt = 10;
 static constexpr int NCutVars = 11;
+static constexpr int NKfCutVars = 12;
```
vkucera, line 468:
```suggestion
static constexpr int NCutKfVars = 12;
```

PWGHF/TableProducer/treeCreatorLcToPKPi.cxx
```diff
@@ -704,13 +718,29 @@ struct HfTreeCreatorLcToPKPi {
...
+        auto p = particle.p();
+        const float p2m = p / MassLambdaCPlus;
+        const float gamma = std::sqrt(1 + p2m * p2m); // mother's particle Lorentz factor
+        const float pvX = mcCollision.posX();
+        const float pvY = mcCollision.posY();
+        const float pvZ = mcCollision.posZ();
+        const float svX = mcDaughter0.vx();
+        const float svY = mcDaughter0.vy();
+        const float svZ = mcDaughter0.vz();
+        const float l = std::sqrt((svX - pvX) * (svX - pvX) + (svY - pvY) * (svY - pvY) + (svZ - pvZ) * (svZ - pvZ));
```
vkucera, line 732: This looks like a distance calculation, which can be done simply with `RecoDecay::distance`.

PWGHF/Core/SelectorCuts.h
lubynets: Done

PWGHF/Core/SelectorCuts.h
lubynets: Done

PWGHF/Core/SelectorCuts.h
lubynets: Done

PWGHF/TableProducer/treeCreatorLcToPKPi.cxx
lubynets: Done

lubynets:
After implementations of changes I was not able to push the development branch (although I added only a single commit and did not have any other discrepancies between local and remote branches). And after trying to synchronize the local branch with remote, I got plenty of commits in it (therefore my PR now has so many PWG labels).
I am sorry if I made something wrong and I hope that it's possible to fix it from now.

vkucera:
Hi @lubynets , you did not update your branch correctly. Try to fix it with these commands:
```bash
git switch KFdev
git switch -c backup
git switch KFdev
git reset --hard 030fdbd
git cherry-pick 35872d0
git push -f origin KFdev
```

lubynets:
Hi @vkucera,
Thanks, it worked!

PWGHF/TableProducer/treeCreatorLcToPKPi.cxx
```diff
@@ -685,7 +699,7 @@ struct HfTreeCreatorLcToPKPi {
               svX = mcParticleProng0.vx();
               svY = mcParticleProng0.vy();
               svZ = mcParticleProng0.vz();
-              l = std::sqrt((svX - pvX) * (svX - pvX) + (svY - pvY) * (svY - pvY) + (svZ - pvZ) * (svZ - pvZ));
+              l = RecoDecay::distance(std::vector<float>{svX, svY, svZ}, std::vector<float>{pvX, pvY, pvZ});
```
vkucera, line 702: Use `std::array` when the number of elements is fixed.

lubynets: Done

lubynets:
> @lubynets Can you please fix the remaining issues? I know you did not introduce most of them.

@vkucera, do you mean failed two checks - [build/O2Physics/o2/macOS](https://ali-ci.cern.ch/alice-build-logs/AliceO2Group/O2Physics/10610/446eb222ff0bfb99a513deb9ada76371f50ccd94/build_O2Physics_o2_macOS/pretty.html) and [O2 linter / O2 linter (pull_request)](https://github.com/AliceO2Group/O2Physics/actions/runs/14023451376/job/39258416477?pr=10610)?
What about the first one, it seems this is a common problem for all PRs?

vkucera:
> > @lubynets Can you please fix the remaining issues? I know you did not introduce most of them.
> 
> @vkucera, do you mean failed two checks - [build/O2Physics/o2/macOS](https://ali-ci.cern.ch/alice-build-logs/AliceO2Group/O2Physics/10610/446eb222ff0bfb99a513deb9ada76371f50ccd94/build_O2Physics_o2_macOS/pretty.html) and [O2 linter / O2 linter (pull_request)](https://github.com/AliceO2Group/O2Physics/actions/runs/14023451376/job/39258416477?pr=10610)? What about the first one, it seems this is a common problem for all PRs?

I mean the O2 linter issues. The failing build is not related to this PR.

vkucera:
@lubynets Thanks a lot for fixing the issues!

PWGHF/TableProducer/treeCreatorLcToPKPi.cxx
```diff
@@ -154,14 +159,16 @@ DECLARE_SOA_TABLE(HfCandLcMCs, "AOD", "HFCANDLCMC",
                   mc_match::XEvent, mc_match::YEvent, mc_match::ZEvent)
 
 DECLARE_SOA_TABLE(HfCandLcKFs, "AOD", "HFCANDLCKF",
+                  full::CollisionId,
```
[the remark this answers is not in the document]
lubynets: I need the collision index at the post-processing steps, when I work with produced trees and want to split decay candidates between different events.
But how can it prevent from data merging? Does the data merger recognize the `CollisionId` name, i.e. it is somehow reserved? If  I replace it with own-defined name like `CollisionId1`, will it work?

PWGHF/TableProducer/treeCreatorLcToPKPi.cxx
```diff
@@ -502,35 +526,29 @@ struct HfTreeCreatorLcToPKPi {
...
               candidate.impactParameter1(),
               candidate.impactParameter2(),
-              trackPos1.tpcNSigmaPi(),
-              trackPos1.tpcNSigmaPr(),
-              trackPos1.tofNSigmaPi(),
-              trackPos1.tofNSigmaPr(),
+              candFlag == 0 ? trackPos1.tpcNSigmaPi() : trackPos2.tpcNSigmaPi(),
               trackNeg.tpcNSigmaKa(),
+              candFlag == 0 ? trackPos2.tpcNSigmaPr() : trackPos1.tpcNSigmaPr(),
+              candFlag == 0 ? trackPos1.tofNSigmaPi() : trackPos2.tofNSigmaPi(),
```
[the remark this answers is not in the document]
lubynets: Did I understand correctly that you propose
1) Not to use `candFlag == 0 ? /* ... */ : /* ... */`, but use `if-else`
2) Do it not inside `rowCandidateLite()` but in advance, like `functionCt, functionY, functionE` are defined earlier?

PWGHF/TableProducer/candidateCreator3Prong.cxx
```diff
@@ -487,6 +494,15 @@ struct HfCandidateCreator3Prong {
       kfCandPiKK.SetConstructMethod(2);
       kfCandPiKK.Construct(kfDaughtersPiKK, 3);
 
+      const float chi2topo = kfCalculateChi2ToPrimaryVertex(kfCandPKPi, kfpV);
+
+      if (applyTopoConstraint) { // constraints applied after chi2topo getter - to preserve unbiased value of chi2topo
+        for (const auto& kfCand : std::array<KFParticle*, 5>{&kfCandPKPi, &kfCandPiKP, &kfCandPiKPi, &kfCandKKPi, &kfCandPiKK}) {
+          kfCand->SetProductionVertex(kfpV);
+          kfCand->TransportToDecayVertex();
+        }
```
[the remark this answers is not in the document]
lubynets: I was also doubting if I should have `const` there. The functions `SetProductionVertex` and `TransportToDecayVertex` are supposed to modify the `kfCand` object.
1) But with this `const` the code compiled successfully.
2) As far as I understand, this `const` prevents modifying the `KFParticle*` pointers themselves but still allow to modify the `KFParticle` objects they point to through non-const member functions.

PWGHF/TableProducer/treeCreatorLcToPKPi.cxx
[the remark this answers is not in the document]
vkucera: It is possible to set a train to avoid DF merging, but AFAIK it is allowed only for the production of linked derived data whose indices point to the parent files. You can declare the collision ID as a normal column, but the values will be ambiguous after the merging.

PWGHF/TableProducer/candidateCreator3Prong.cxx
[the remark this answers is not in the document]
vkucera: You are right. The elements are pointers, so it should be fine.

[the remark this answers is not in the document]
lubynets: P.S. I checked it with a toy-example and convinced myself that the functions under loop work as I suppose (even with `const`).

PWGHF/TableProducer/treeCreatorLcToPKPi.cxx
```diff
@@ -704,13 +718,29 @@ struct HfTreeCreatorLcToPKPi {
...
+        const float p2m = p / MassLambdaCPlus;
+        const float gamma = std::sqrt(1 + p2m * p2m); // mother's particle Lorentz factor
+        const float pvX = mcCollision.posX();
+        const float pvY = mcCollision.posY();
+        const float pvZ = mcCollision.posZ();
+        const float svX = mcDaughter0.vx();
+        const float svY = mcDaughter0.vy();
+        const float svZ = mcDaughter0.vz();
+        const float l = RecoDecay::distance(std::array<float, 3>{svX, svY, svZ}, std::array<float, 3>{pvX, pvY, pvZ});
+        const float t = mcDaughter0.vt() * NanoToPico / gamma; // from ns to ps * from lab time to proper time
```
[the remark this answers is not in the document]
lubynets: Done

PWGHF/TableProducer/candidateSelectorLc.cxx
```diff
@@ -268,14 +285,97 @@ struct HfCandidateSelectorLc {
...
+        return false;
+      }
     }
 
-    /// cut on the Kpi pair invariant mass, to study Lc->pK*(->Kpi)
-    const double cutMassKPi = cuts->get(pTBin, "mass (Kpi)");
-    if (cutMassKPi > 0 && std::abs(massKPi - massK0Star892) > cutMassKPi) {
-      return false;
+    if constexpr (reconstructionType == aod::hf_cand::VertexerType::KfParticle) {
+      if (applyKfCuts) {
```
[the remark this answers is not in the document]
lubynets: Inside this if-statement I use getters which are KF-specific. And without if **constexpr** the code does not compile.

PWGHF/TableProducer/treeCreatorLcToPKPi.cxx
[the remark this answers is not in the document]
lubynets: Sorry, I still do not understand what the issue is and thus how to fix it.
In my postprocessing (which I mentioned above) I indeed do not merge files. But what will be wrong if one merges files? Is `CollisionId` somehow critical for merging, does it differ from any arbitrary int field?

PWGHF/TableProducer/candidateSelectorLc.cxx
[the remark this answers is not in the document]
vkucera: OK, that makes sense.

PWGHF/TableProducer/treeCreatorLcToPKPi.cxx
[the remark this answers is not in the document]
vkucera: The automatic output merging on Hyperloop does not only merge files but also data frames. The indices have to be renumbered to keep them pointing to the correct rows of the merged tables when different data frames are merged. This is not possible when the index is pointing outside the output tables, i.e. is unbound. This condition is detected by Hyperloop and prevents the train submission. The only exception is the production of linked derived data, where the output tables are linked to the parent AODs and are supposed to be used as input for other workflows.

PWGHF/TableProducer/treeCreatorLcToPKPi.cxx
```diff
@@ -469,23 +489,51 @@ struct HfTreeCreatorLcToPKPi {
...
+            valueTpcTofNSigmaPi = trackPos2.tpcTofNSigmaPi();
           } else {
-            FunctionInvMass = CandFlag == 0 ? candidate.kfMassPKPi() : candidate.kfMassPiKP();
-            FunctionInvMassKPi = CandFlag == 0 ? candidate.kfMassKPi() : candidate.kfMassPiK();
+            valueTpcNSigmaPr = trackPos2.tpcNSigmaPr();
+            valueTpcNSigmaPi = trackPos1.tpcNSigmaPi();
+            valueTofNSigmaPr = trackPos2.tofNSigmaPr();
+            valueTofNSigmaPi = trackPos1.tofNSigmaPi();
+            valueTpcTofNSigmaPr = trackPos2.tpcTofNSigmaPr();
+            valueTpcTofNSigmaPi = trackPos1.tpcTofNSigmaPi();
```
fgrosa, line 535: While I understand the logic behind that, unfortunately this would change the behaviour for all the Lc analyses based on BDT models. In fact, the features here should match those that are then used in the ML inference https://github.com/AliceO2Group/O2Physics/blob/9857c8bc064c980c3a13fe07862ea24160853fdf/PWGHF/Core/HfMlResponseLcToPKPi.h#L161-L196. Currently what you are doing here it is done with the "expected hypothesis" features in the ML inference https://github.com/AliceO2Group/O2Physics/blob/9857c8bc064c980c3a13fe07862ea24160853fdf/PWGHF/Core/HfMlResponseLcToPKPi.h#L115-L120, but the default is to have p and pi hypotheses for prongs 0 and 2 regardless the pKpi and piKp mass hypotheses.
I would then ask you not to change this, or if you want, do it only if a configurable is enabled. Please keep also in mind that:
1) you can always do it when processing the trees locally (because you have the info of the mass hypothesis for each candidate)
2) it is not necessarily bad to have proton and pion PID hypotheses even if you have the candidate that it is preselected by only one of the two, because the wrong one can be used by the BDTs to reject the wrong hypothesis.
Thanks!

fgrosa (changes requested):

PWGHF/TableProducer/treeCreatorLcToPKPi.cxx
lubynets: Hello, @fgrosa,
Thanks for your comment! I did not know that the ML-oriented part is explicitly based on those variables which I removed.
1. Evaluating "expected PID hypothesis" locally is what I was doing before, but wanted to avoid. In my local ML setup I use "expected", not "raw" `nSigma` values.
2. Honestly I do not understand how working with "raw" `nSigma`s can be useful for BDT: the number of sigmas from, let's say, protons location at `dE/dx` plot has completely opposite impact if the candidate is `PKPi` or `PiKP`. In first case it is "good" (for signal) to have small values, and in second case it should be far from proton, because it is close to pion (again, I'm talking about signal). Does BDT manage to distinguish these two different patterns is distributions of `nSigma` proton and pion?

P.S. The second item is rather my thinking in loud. Of course if these variables are needed I will find a solution how not to remove them. But I would also prefer to have "expected" values. You mentioned including them by a configurable, but as far as I understand it's impossible to variate the tables content in runtime? Will it be a solution if I
- return remove variables as they were;
- add "expected" variables into another table, `HFCANDLCKF`, KFParticle related (although `nSigma` is not a KFParticle-specific thing, but at least one can have a possibility not to write these variables by default, when running in the DCAFitter mode)?

PWGHF/TableProducer/candidateSelectorLc.cxx
```diff
@@ -268,14 +285,97 @@ struct HfCandidateSelectorLc {
...
-    if (std::abs(massLc - o2::constants::physics::MassLambdaCPlus) > cuts->get(pTBin, "m")) {
-      return false;
+    if (reconstructionType == aod::hf_cand::VertexerType::DCAFitter || (reconstructionType == aod::hf_cand::VertexerType::KfParticle && applyNonKfCuts)) {
+      // cut on daughter pT
+      if (trackProton.pt() < cuts->get(pTBin, "pT p") || trackKaon.pt() < cuts->get(pTBin, "pT K") || trackPion.pt() < cuts->get(pTBin, "pT Pi")) {
+        return false;
+      }
+
+      // cut on Lc->pKpi, piKp mass values
+      if (std::abs(massLc - o2::constants::physics::MassLambdaCPlus) > cuts->get(pTBin, "m")) {
```
mfaggin, line 295: I'd like to ask you to put these lines where they were before, i.e. before the inv. masses calculations. And leave the comments that were there before, please.

PWGHF/TableProducer/treeCreatorLcToPKPi.cxx
mfaggin: Hi,
the current version of the tree creator always allows you to process locally your trees to have the "expected" version of the PID variables, so I agree with Fabrizio not to touch it

PWGHF/TableProducer/candidateSelectorLc.cxx
lubynets: Hello, @mfaggin,
thanks for the feedback! I admit that I accidentally removed comments; I will recover them.
Concerning the order of prong pT and inv. mass cuts - is it important due to performance reasons, i.e. not to evaluate invariant mass if the pT condition returns false?

PWGHF/TableProducer/treeCreatorLcToPKPi.cxx
lubynets: Hi, 
I'll return the previously existed variables. But what about the "expected" ones - do you object against their placing e.g. in a different table? I understand that they can be evaluated later, but it will complicate the post-processing operations. Moreover, if one needs to add "expected" PID variables into ML (inside selector task), then delayed evaluation of them won't be possible anymore.

PWGHF/TableProducer/candidateSelectorLc.cxx
mfaggin: indeed if the candidate is already outside the kinematic range of interest there is no need to calculate those quantities. This for sure reduces the CPU time (I cannot quantify precisely the impact of it, but it's a good practice to avoid unnecessary calculations)

PWGHF/TableProducer/treeCreatorLcToPKPi.cxx
mfaggin: if you want to place them in a different table, which can be saved only upon request as the others, then I think it's ok. But indeed I'd keep the current tables as they are (as pointed out by Fabrizio)

PWGHF/TableProducer/candidateSelectorLc.cxx
lubynets: Hi @mfaggin,
Now I do not understand which comments you mean. Did I remove anything in my PR?

vkucera: @mfaggin is right. Getting a track pt from the table is much faster than calculating the candidate's invariant mass. Also the pt spectrum is steeply falling so a cut on the minimum pt rejects a significant fraction of candidates. Both factors matter for performance.
I suspect that the comment you removed is
```c++
/// cut on the Kpi pair invariant mass, to study Lc->pK*(->Kpi)
```
which seems relevant.

PWGHF/TableProducer/treeCreatorLcToPKPi.cxx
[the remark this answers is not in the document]
vkucera: Correct.
As long as `full::CollisionId` is in any of the tables you save, the train will not be allowed to run.

vkucera:
Converted to draft because of unresolved breaking changes.

PWGHF/TableProducer/treeCreatorLcToPKPi.cxx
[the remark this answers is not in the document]
lubynets: Thanks for the clarification! I fixed it.

PWGHF/TableProducer/candidateSelectorLc.cxx
lubynets: Hi @mfaggin, @vkucera,
I recovered the `pT -> minv` order of cut application.
What about the comment, I did not remove them, I just split comments about `PKPi` and `KPi` invariant mass since the cuts are applied in two separated lines now.

fgrosa (dismissed):
Good for me now, thanks @lubynets for reverting the PID variables!

fgrosa:
Hi @lubynets, is this PR ready? Should we revert it to `Ready for review`? Thanks!

lubynets:
Hi @fgrosa,
Yes, it's ready for review. Sorry for the silence from my side.

vkucera:
@lubynets Please fix the issues.

lubynets:
> @lubynets Please fix the issues.

@vkucera, I noticed the "magic number" issues, but I could not invent a meaningful name for the variable 2, because I do not understand the physics case - why a 3-prong decay particle has only two daughters. Can you please briefly hint what it means, so I could create a proper alias for 2?

PWGHF/TableProducer/treeCreatorLcToPKPi.cxx
```diff
@@ -966,44 +1097,83 @@ struct HfTreeCreatorLcToPKPi {
...
+            const float z = candidate.zSecondaryVertex();
+            const float errX = candidate.kfXError();
+            const float errY = candidate.kfYError();
+            const float errZ = candidate.kfZError();
+            const float errPVX = candidate.kfXPVError();
+            const float errPVY = candidate.kfYPVError();
+            const float errPVZ = candidate.kfZPVError();
+            const float chi2primKaon = candidate.kfChi2PrimProng1();
+            const float dcaProtonPion = candidate.kfDcaProng0Prong2();
+            const float chi2GeoProtonPion = candidate.kfChi2GeoProng0Prong2();
```
vkucera, line 1176: The amount of code duplication is increasing with every addition. Please take a moment for a proper factorisation soon.

vkucera:
> > @lubynets Please fix the issues.
> 
> @vkucera, I noticed the "magic number" issues, but I could not invent a meaningful name for the variable 2, because I do not understand the physics case - why a 3-prong decay particle has only two daughters. Can you please briefly hint what it means, so I could create a proper alias for 2?

It's the number of daughters in the resonant decay channel.
```c++
constexpr std::size_t NDaughtersResonant{2u};
```

lubynets:
> @lubynets Please fix the issues.

Done

PWGHF/TableProducer/treeCreatorLcToPKPi.cxx
fgrosa: Hi @lubynets @vkucera if you agree, I would postpone it to a separated PR, since it would touch only one file while the current one is open since a while and it is implementing modifications to a few files that we would like to modify for a couple of central developments (see for event selection https://github.com/AliceO2Group/O2Physics/pull/10949, and another one regards the PID variables). If you confirm that it is ok for you, I would merge this PR. Thanks!

lubynets: Hi @fgrosa (@vkucera cc),
From my side it's ok to prepare a separate PR dedicated to code duplication reduction (if Vit does not object).
I would also like to clarify whether my idea expressed above (replace "full" table with "additional / extended / etc") makes sense? Or should I contact those who is marked as code authors - Nicolo' and Luigi?

fgrosa: @lubynets thanks! 
While I understand the reasoning, I would keep `full` simply because it is the same for all channels and it's easier for newcomers to have similar names (or you change it for all the channels), but not a strong opinion ;-)

lubynets: No, my idea is not about names change! (the name can be approved later)
The point is that there are two possibilities - to write `Lite` or `Full` table on user's request according to the Configurable flag. But the `Lite` is a subset of `Full`, so there are a lot of identical code lines, which we want to reduce.
My suggestion is to replace `Full` with `Full*` (the name can be a subject of discussion) which will contain the _difference_ between `Full` and `Lite` - then there will be no code duplication. Then the `Lite` will be written always, by default, and `Full*` - only by user's request of one needs to save more information.
But I am not sure if it will be compatible with other tasks, and if it's OK to have two output tables instead of one.

fgrosa: I see, sorry for the misunderstanding! In principle it is OK to have 2 tables instead of one, however it is not very user friendly for local processing (since the tree creator is meant to produce a flat tree for analysis). If you use O2 or python (e.g. pandas) is easy to join two "tables", but otherwise not much, so I would still keep the `lite` and `full` so that it's easier for less expert users (but honestly speaking I also don't know how many users use the `full` version..)

vkucera: I was not suggesting to do the factorisation in this PR, so I'm fine with the idea of a dedicated PR after the central features are merged.

fgrosa (approved):

PR #10620, opened 2025-03-23
[PWGLF] add setting_cutChi2NClITS configurable

blacwovie:
add setting_cutChi2NClITS configurable


maciacco (approved):
please fix the O2 linter errors in the next PR, they are mostly caused by the usage of PDG codes.

PR #10632, opened 2025-03-24
[DPG] Update detectorOccupancyQa.cxx - add configs to suppress heavy histos

altsybee:
add configs to suppress heavy histos


alibuild (approved):

PR #10641, opened 2025-03-25
[DPG] adding dummy proccess, and ITS PID to the derived data

amaringarcia:


amaringarcia:
@catalinristea @alcaliva could you have a look at the pull request and approved as soon as possible?
Thanks in advance
Ana

alcaliva (approved):

PR #10642, opened 2025-03-25
[PWGHF] Fix the sliceby issue in xic0 vs mult task

Tao-Fang:


Tao-Fang:
1, Try to fix the issue that the CandidateTable only fills in the last collision 2, Process the collisionIndex independently from the Candidate table

PR #10656, opened 2025-03-26
[PWGCF] Update corrSparse.cxx addressing possible memory leak

Thorkj:


victor-gonzalez (approved):
I would not say `possible`, it is actually a memory leak!

PR #10665, opened 2025-03-26
[Trigger] Add doublephi filter in cefp task

skundu692:


vkucera:
@skundu692 You are using a workflow option `train_config` to customise workflow topology composition in `defineDataProcessing` which is not supported on AliHyperloop. How is it useful?

mpuccio (approved):

mpuccio:
Hi @vkucera, this is my code, Sourav kindly modified it to add his new trigger in the configurable. We inspect the train configuration json to subscribe to different tables depending on the tasks present in the topology. The configurable is meant to be used outside hyperloop, so I would not bother now changing it. Actually, this gives me the occasion to ask if we can customise the o2-linter warnings per directory, in the trigger directory I would like to disable those I am anyway ignoring for approving the PRs. Is that possible?

PR #10666, opened 2025-03-26
[PWGHF] fix a mistake in previous commit in tcc tree creator

zhangbiao-phy:


alibuild (approved):

PR #10679, opened 2025-03-27
[PWGLF,Trigger] Omegamult

lietava:
HM omegas selected using normalised multiplicity - code copied from multFilter.cxx.


lietava:
Hi @mpuccio , @ChiaraDeMartin95 - there is one thing I'd like to clarify - around 15% of collisions does not have FT0.
(collision.has_FoundFT0).

lietava:
I will do one more check - comparison for rejected collisions.

lietava:
The histo of collision.multFT0M() for those which not has_foundFT0 or amplitude is zero looks like:
T0M_no_FT0
The FT0M cut is at 3100 (def in code). So it looks okish although it would be good to understamd what they are.

lietava:
Hi @aortizve , do you know or where can I find relation between multFT0M and FT0 ?

ChiaraDeMartin95:
Hello @lietava,
thank you for implementing this in the code, it all looks fine to me!
I was naively expecting that events with well-defined FT0M multiplicity also have the flag has_foundT0() set to one.  From your AnalysisResult.root, it seems this is not the case for 11% of events, but I don't know why.

ChiaraDeMartin95 (approved):

PR #10681, opened 2025-03-27
[PWGCF/FemtoUniverse] Add pT histo with only primaries.

davkk:
This PR adds a new $p_T$ histogram to `FemtoUniverseParticleHisto` only from primaries, for correct reconstruction efficiency calculation.


zchochul (approved):

vkucera:
@zchochul There are many directives disabling O2 linter tests without any given reason. Why?

PR #10684, opened 2025-03-27
[PWGLF] doubleResonanceScan - Add new task

BongHwi:
- New task that can scan general two resonance decay mode into four daugthers using Resonance Derived dataset
    - pion, kaon, proton daughters can be chosen.

### DIFF
--- a/PWGLF/Tasks/Resonances/doubleResonanceScan.cxx
+++ b/PWGLF/Tasks/Resonances/doubleResonanceScan.cxx
@@ -1,0 +1,549 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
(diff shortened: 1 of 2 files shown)


PWGLF/Tasks/Resonances/doubleResonanceScan.cxx
```diff
@@ -0,0 +1,548 @@
...
+// or submit itself to any jurisdiction.
+///
+/// \file doubleResonanceScan.cxx
+/// \brief Resonance Scanner with ResoTracks and ResoMicroTracks
+/// \author Bong-Hwi Lim <bong-hwi.lim@cern.ch>
+/// \since 27/03/2025
+///
+
+#include <vector>
+#include <TLorentzVector.h>
```
vkucera, line 19: `TLorentzVector` is a legacy class and should not be used.

BongHwi: Thanks for the suggestion, I'd try with `ROOT::Math::LorentzVector`

BongHwi: I temporarily reverted to use the original method. To use the other method, it needs a time to apply and test.

dmallick2 (approved):

PR #10689, opened 2025-03-28
[Common] Always initialize listCalib as it is serialized unconditionally

aalkin:
This fixes 
```
 Cannot read class info from buffer of ATSK/12c99cc07f5f____/0
 Cannot read class info from buffer of ATSK/12c99cc0863a____/0
 ```
errors that are the result of a `nullptr` being serialized. Now the empty list is always created, even if `produceHistograms` is `false`. I will add protection for `OutputObj` in a separate PR so we can avoid further issues like this. 

@njacazio @ktf @mhemmer-cern


mhemmer-cern:
Thank you so much @aalkin @ktf  🥳

njacazio (approved):

PR #10691, opened 2025-03-28
[PWGUD] Modify FwdMuonsUPC.cxx

ariffero:
- Add a pT histo with bins with variable width.
- Modify the names of saved gen varaibles to match the names of gen variables in the reco tree.


ariffero:
Searching on the internet it seems that the error I got from MegaLinter is a false positive, furthermore it is not related to a code that I wrote (and committed).

vkucera:
@ariffero Don't start the PR title with a space.

vkucera:
> Searching on the internet it seems that the error I got from MegaLinter is a false positive, furthermore it is not related to a code that I wrote (and committed).

If you update your branch, you won't get it.

vkucera:
> > Searching on the internet it seems that the error I got from MegaLinter is a false positive, furthermore it is not related to a code that I wrote (and committed).
> 
> If you update your branch, you won't get it.

The O2 linter warnings are true positives though.

vkucera:
Given that you modify only your own file, I would suggest to fix the errors.

PR #10694, opened 2025-03-28
[PWGCF] FemtoUniverse - Separating nSigma cuts for tracks and daughters

basiach:
Separating nSigma cuts for tracks and daughters into different configurables for better selection.


vkucera:
@basiach The `o2-linter: disable` comments should be removed and the issues should be fixed.

prchakra (approved):

PR #10702, opened 2025-03-30
[PWGDQ] Adding muonBasic process function for code optimization

lucamicheletti93:
Adding muonBasic process function for code optimization. The menory consumption of Multiplicity table is ~200 MB for basic muon task, bringing beyond the limit the PSS memory for the dqEfficiency_withAssoc. Including this minimal muon process function should optimize the preocessing.

immagine


alibuild (approved):

PR #10716, opened 2025-04-01
[PWGUD] MC related error Solved

sdudi123:
The modification in SG Producer is done by adding the line 347 if(!collision.has_mcCollision()) return; to overcome the error "o2-analysis-ud-sgcand-producer, device shutting down. Reason: Accessing invalid index for mcCollision". It gives an error when there is no associated mc collision to a collision index.

### DIFF
--- a/PWGUD/TableProducer/SGCandProducer.cxx
+++ b/PWGUD/TableProducer/SGCandProducer.cxx
@@ -344,6 +344,8 @@ struct SGCandProducer {
                      aod::FT0s const& ft0s, aod::FDDs const& fdds)
   {
     // select specific processes with the GeneratorID
+    if (!collision.has_mcCollision())
+      return;
     auto mccol = collision.mcCollision();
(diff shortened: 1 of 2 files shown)


rolavick:
Hi @sdudi123 ,
thanks for introducing the fix of the bug. Please, add more detailed description of the error here in the PR, so it is trackable what issue has been solved.
@abylinkin @pbuehler , could you please check the modification of SG Producer is within the code intentions? Thank you.
Roman

sdudi123:
The modification in SG Producer is done by adding the line 347  if(!collision.has_mcCollision()) return;  to overcome the error "o2-analysis-ud-sgcand-producer, device shutting down. Reason: Accessing invalid index for mcCollision". It gives an error when there is no associated mc collision to a collision index.

pbuehler (approved):

rolavick (approved):

PR #10726, opened 2025-04-01
[PWGLF] PWGMM:Lumi

JosueMtzGar:
adding leading BC pattern and corrections for rate.


JosueMtzGar:
Hi Reviewers, 

I saw the log of errors message for the build on macOS but I didn't see any error in relation with my specific task. How can I fix this fails in the pull request process?

PR #10728, opened 2025-04-02
[PWGLF] enable DCA fitter

blacwovie:
enable DCA fitter


maciacco (approved):
please fix o2 linter errors in the next PR

PR #10733, opened 2025-04-02
[PWGDQ] Save the information of ambiguous pairs

yuanz-USTC:
Define histograms to save the information of mix event ambiguous pairs and also ambiguous pairs which are double counted for both same event and mix event.


PWGDQ/Tasks/tableReader_withAssoc.cxx
```diff
@@ -1178,6 +1179,8 @@ struct AnalysisSameEventPairing {
   std::map<int, std::vector<TString>> fMuonHistNames;
   std::map<int, std::vector<TString>> fTrackMuonHistNames;
   std::vector<AnalysisCompositeCut> fPairCuts;
+  std::vector<AnalysisCompositeCut> fTrackCuts;
```
iarsene, line 1182: Here you create a vector of composite cuts, but you actually just need the name of the cuts. So I suggest to make a vector of TString only.

PWGDQ/Tasks/tableReader_withAssoc.cxx
```diff
@@ -1229,6 +1233,9 @@ struct AnalysisSameEventPairing {
     TObjArray* objArrayTrackCuts = nullptr;
     if (!trackCutsStr.IsNull()) {
       objArrayTrackCuts = trackCutsStr.Tokenize(",");
+      for (int icut = 0; icut < objArrayTrackCuts->GetEntries(); ++icut) {
+        fTrackCuts.push_back(*dqcuts::GetCompositeCut(objArrayTrackCuts->At(icut)->GetName()));
+      }
```
iarsene, line 1238: If you get the vector of cuts at this point, then this would mix-up the track bits and corresponding histogram directories. The problem is that these are the track cuts specified in the same event pairing configurable. This is not guaranteed to be in the same order (or they don't even exist in the track-selection task) as the cuts actually defined in the track selection task. As you see later in the init and the run function, for the track bits, we rely on the order in which the user provided the cuts in the track-selection task, not in this task. So I think its probably enough to actually initialize this vector a bit lower in the init function, where we loop over the track cuts defined in the track-selection task and check if these where required in the same-event-pairing task.

iarsene (approved):

PR #10740, opened 2025-04-02
[PWGDQ] Added a few electron cuts for the filter pp

iarsene:


alibuild (approved):

PR #10746, opened 2025-04-03
[PWGCF] Add closure test

fuchuncui:
Add closure test
Optimize some code


victor-gonzalez (approved):
I would suggest following `linter` recommendations
They are really useful when something does not work as expected and the code needs to be reviewed

PR #10749, opened 2025-04-03
Add PID in MC process functions

cnkoster:


victor-gonzalez (approved):

victor-gonzalez:
Does the code compile locally?

cnkoster:
I fixed the problem (there was a "{" somewhere where it shouldn't) and now it builds locally. 
However, the build on macOS still seems to fail

victor-gonzalez (approved):

PR #10754, opened 2025-04-03
Fix -Wmissing-template-arg-list-after-template-kw

mpuccio:
I suspect that this might break the CI, as I don't think people put template keywords around out of whimsical creativity. But either this works, or we will have to disable that warning from the list of -Wall


dsekihat (approved):

dsekihat:
Thank you for checking this.

PR #10760, opened 2025-04-04
[Common] Add TrackQAConverter003

shahor02:
Goes with https://github.com/AliceO2Group/AliceO2/pull/14144 and substitutes https://github.com/AliceO2Group/O2Physics/pull/10748 from @f3sch


f3sch:
@ddobrigk even though its stale this should still be merged to keep backward comp. for the table.

ddobrigk (approved):

PR #10776, opened 2025-04-07
[PWGLF] Add and fix for code test using kaon

sangwoo184:
Add and fix for code test using kaon instead of pion


vkucera:
@sangwoo184 Please fix the issues.
Keep in mind that you should really not use `std::cout` for logging messages.

smaff92 (approved):

PR #10779, opened 2025-04-07
[Tutorial] Add an example of a configurable expression column

aalkin:


ktf:
Very nice. We should probably review tasks using expressions already and see if this does not allow more stuff to be moved inside.

PR #10786, opened 2025-04-08
[PWGCF] Add MCRec for closure test

fuchuncui:


victor-gonzalez (approved):

victor-gonzalez:
I would suggest following `linter` recommendation regarding 'magic numbers'
Use named constants instead. The code will be more readable which will make it better maintainable and easy to evolve it

PR #10798, opened 2025-04-09
Adding a process function for V0 tagged tracks in pp, without TOF dependency

iarsene:


alibuild (approved):

PR #10802, opened 2025-04-09
[PWGHF] Add tagging of D* -> D0 pi -> K pi pi pi0 at gen level

fgrosa:


alibuild (approved):

PR #10813, opened 2025-04-10
[PWGHF] Patches for first Xic autoencoder study

macamerl:
Minimal changes in XicToPKPi specific files (task and selector) to evaluate the MSE between input (externally scaled) and internal scaled output of autoencoder. At moment, one can not run Autoencoder and BDT at the same time since they competitively fill candidate.mlProbXicToPKPi()[0].

### DIFF
--- a/PWGHF/TableProducer/candidateSelectorXicToPKPi.cxx
+++ b/PWGHF/TableProducer/candidateSelectorXicToPKPi.cxx
@@ -54,7 +54,7 @@ using namespace o2::framework;
 struct HfCandidateSelectorXicToPKPi {
   Produces<aod::HfSelXicToPKPi> hfSelXicToPKPiCandidate;
   Produces<aod::HfMlXicToPKPi> hfMlXicToPKPiCandidate;
-  Produces<aod::HfMseXicToPKPi> hfMseXicToPKPiCandidate; 
+  Produces<aod::HfMseXicToPKPi> hfMseXicToPKPiCandidate;
   Produces<aod::HfAeOutXicToPKPi> hfAeXicToPKPiCandidate;
 
   Configurable<double> ptCandMin{"ptCandMin", 0., "Lower bound of candidate pT"};
@@ -87,7 +87,7 @@ struct HfCandidateSelectorXicToPKPi {
   Configurable<bool> applyMSE{"applyMSE", false, "Flag to calculate MSE for autoencoders"};
   Configurable<bool> applyMinMax{"applyMinMax", false, "Flag to MinMax feature preprocessing"};
   /// Parameter vectors for feature preprocessing - external scaling
-  Configurable<std::vector<float>> scaleMin{"scaleMin", {0.,0.,0.}, "vector of scaling parameter min"};  
+  Configurable<std::vector<float>> scaleMin{"scaleMin", {0.,0.,0.}, "vector of scaling parameter min"};
   Configurable<std::vector<float>> scaleMax{"scaleMax", {1.,1.,1.}, "vector of scaling parameter max"};
   // CCDB configuration
   Configurable<std::string> ccdbUrl{"ccdbUrl", "http://alice-ccdb.cern.ch", "url of the ccdb repository"};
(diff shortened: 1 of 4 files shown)


zhangbiao-phy:
Hi @macamerl, Thanks for the development! First of all, let me ping the code owner to see if they have any comments on this implementation. @mfaggin @cterrevo and @phymanshu.  Then I only have a few comments, see the review. Then could you also fix magalinter errors since they come from this PR?

phymanshu:
Hi @macamerl @zhangbiao-phy 
Since you are adding a configurable for these studies, I have no problem as long as it is not conflicting with the standard BDT analysis. 
Best wishes

macamerl:
Thanks @phymanshu. Yes, I locally checked that the BDT inference  has no problem after the changes that I made.  The flags "applyMSE" and "applyMinMax" are set false/0 by default. In any case, I will privately send you the config file that I used for the check.

PWGHF/TableProducer/candidateSelectorXicToPKPi.cxx
```diff
@@ -35,11 +35,38 @@
...
+        yf_true.push_back((y_true.at(j) - ScaleMin.at(j)) / (ScaleMax.at(j) - ScaleMin.at(j)));
+        LOG(debug) << "feature = " << j << " ----> input = " << y_true.at(j) << " scaled feature = " << yf_true.at(j);
+      } else {
+        yf_true.push_back(y_true.at(j));
+      }
+      sum += pow(((yf_true).at(j) - (y_pred).at(j)), 2); // has dimensions
+      LOG(debug) << "feature = " << j << " ----> input = " << yf_true.at(j) << " AE prediction = " << y_pred.at(j);
+    }
+    mse = sum / y_pred.size(); // MSE of a candidate
+  }
```
fgrosa, line 62: Since this is not HF specific, I would move this in the ML folder. You could consider to add it as a function of the https://github.com/AliceO2Group/O2Physics/blob/master/Tools/ML/MlResponse.h class, or to add a header file with utils functions.
Please also follow the naming conventions of O2Physics when you move it (camel case), thanks!

fgrosa (changes requested):

PWGHF/TableProducer/candidateSelectorXicToPKPi.cxx
```diff
@@ -234,13 +270,19 @@ struct HfCandidateSelectorXicToPKPi {
...
 
       if (!TESTBIT(candidate.hfflag(), aod::hf_cand_3prong::DecayType::XicToPKPi)) {
         hfSelXicToPKPiCandidate(statusXicToPKPi, statusXicToPiKP);
         if (applyMl) {
-          hfMlXicToPKPiCandidate(outputMlXicToPKPi, outputMlXicToPiKP);
+          if (applyMSE) {
+            hfMlXicToPKPiCandidate(outputMSEXicToPKPi, outputMSEXicToPiKP);
+          } else {
+            hfMlXicToPKPiCandidate(outputMlXicToPKPi, outputMlXicToPiKP);
+          }
```
vkucera, line 285: Can you not avoid all the branching and just make the decision once?

vkucera (changes requested):

macamerl:
Hi, I'm implementing the comment received by @fgrosa .  However, we had Italian national holidays as soon as after Easter and I had to replace one of my colleagues for other more urgent tasks of the project. Then, I have slowed down the debugging of the new version. I will give you an update next week.

vkucera:
@macamerl Is there any activity on this?

macamerl:
@vkucera Yes, I implemented the ideas (and small adjustment) of the ideas in the offline mail in the old O2Physics version. it worked at the beginning of June. Now, I think to see some conflicts btw the onnx versions (see error below). Then, I decide to update my local O2Physics version and to arrive at the point that my branch works again.  

I also attached an example of the error: "Schema error: Trying to register schema with name Abs (domain:  version: 1) from file /local/workspace/DailyBuilds/DailyO2Physics-slc9/daily-tags.uO7Ig4H9DC/SOURCES/onnx/v1.17.0-alice2/v1.17.0-alice2/onnx/defs/math/old.cc line 2729, but it is already registered from file /local/workspace/DailyBuilds/DailyO2Physics-slc9/daily-tags.uO7Ig4H9DC/SOURCES/onnx/v1.17.0-alice2/v1.17.0-alice2/onnx/defs/math/old.cc

PR #10815, opened 2025-04-11
[PWGLF] PWGMM:Lumi: Changes onn rate histo axis

JosueMtzGar:


aalkin:
The active tasks really should be moved under PWGLF directory, so that PWGMM can be archived. In addition, the review and merge should be done by the respective PAG or PAG/PWG conveners, rather than me.

aalkin (approved):
This is an exception, please move your task into PWGLF directory as soon as you can.

PR #10819, opened 2025-04-11
[DPG] Update eventSelectionQa.cxx - add dEdx vs occupancy QA histo

altsybee:
add dEdx vs occupancy QA histo, to monitor dEdx in a narrow momentum bin


alibuild (approved):

PR #10820, opened 2025-04-11
[PWGHF,Trigger] Add Bc decay to D0K into the HF filter

zhangbiao-phy:
Hi @fgrosa, Is it Ok? I only change the mass hypothesis from pion to kaon.


joao-krauchenco-mottu (approved):

fgrosa:
Hi @zhangbiao-phy, thanks!
I think that you need to adjust a bit the enum though; the code currently assumes that we have first "3-prong" beauty candidates (those with D0) and then "4-prong". In fact, you can see that for the 4-prong ones, we loop from B0 to Xib:
https://github.com/AliceO2Group/O2Physics/blob/7972d4be62372842de3661c85c09c2f2517089a0/EventFiltering/PWGHF/HFFilter.cxx#L1264
So you have to put the `Bc` after the `B0toDstar` and you have to pay very attention to the indices for the 4-progns (especially you have to replace `kNBeautyParticles - 2` with `kNBeautyParticles - 3`), otherwise we screw up all the 4-prong channels.

alibuild (approved):

PR #10824, opened 2025-04-11
Improve O2 linter

vkucera:
- Rename classes to match the test names better.
- `std-prefix`: Simplify test.
- `root-entity`: Rename to `root/entity`.
- `magic-number`: Extend pattern. Fix false positves. Support multiple matches per line.
- Add `root/lorentz-vector`.
- Add `pdg/explicit-mass`.


iarsene (approved):

alibuild (approved):

iarsene (approved):

PR #10854, opened 2025-04-15
[PWGLF] angularCorrelationsInJets.cxx: improved resource use

ljoergen:


ljoergen:
What is this, I'm not even explicitly including cmath? My task compiles on linux and it runs, this error isn't my doing.

maciacco (approved):

PR #10857, opened 2025-04-15
[PWGHF] Fixes for OmegaC0 workflows

fcatalan92:
Fixed some issues in histogram filling and selections. In particular, the reconstruction flag was not checked before filling the trees, and the current boolean flag that is provided by the candidate selector has some problems when used with partitions. Please @Yunfan-Liu have a look.

@mtorresc15


alibuild (approved):

fcatalan92:
Hi @vkucera, fixing the naming issue means changing the name of the workflow and it would affect various analyses on Hyperloop. So, I would avoid it for the moment.

PR #10862, opened 2025-04-15
PWGHF: added table for DeltaY between Lc-hadrons

1994ra:


vkucera:
Hi @1994ra , please define a constant struct member `static const size_t nDaughters{3u}` and use it everywhere instead of `3`.

PWGHF/HFC/TableProducer/correlatorLcHadrons.cxx
vkucera, line 1: Please fix the issues.

vkucera (changes requested):

PWGHF/HFC/TableProducer/correlatorLcHadrons.cxx
1994ra: Hi @vkucera , Which issue, you suggested one in previous comment but I don't see anything here

1994ra: it is done

PWGHF/HFC/DataModel/CorrelationTables.h
```diff
@@ -212,7 +213,8 @@ DECLARE_SOA_TABLE(TrkRecInfoLc, "AOD", "TRKRECINFOLC", //! Tracks Reconstructed
                   aod::hf_correlation_lc_hadron::TrackDcaXY,
                   aod::hf_correlation_lc_hadron::TrackDcaZ,
                   aod::hf_correlation_lc_hadron::TrackTPCNClsCrossedRows);
-
+DECLARE_SOA_TABLE(LcHadronPairY, "AOD", "LCHPAIRY", //! Lc candidates Generated Information
+                  aod::hf_correlation_lc_hadron::DeltaY);
```
vkucera, line 217: Why is this not part of the `LcHadronPairY` table?

1994ra: you mean `LcHadronPair` table?

1994ra: We were using `DeltaEta` information, now I wanted to check `DeltaY` distribution if it is same as `DeltaEta` and if there any advantage over `DeltaEta`. At the same time I don't want to overload the existing table as we are not sure yet about `DeltaY`, though  this table can be  join  to `LcHadronPair`

vkucera: OK, fair enough?

vkucera (approved):

PWGHF/HFC/TableProducer/correlatorLcHadrons.cxx
```diff
@@ -453,6 +456,7 @@ struct HfCorrelatorLcHadrons {
                             track.pt() * track.sign(),
                             poolBin,
                             correlationStatus);
+          entryLcHadronPairY(track.y() - hfHelper.yLc(candidate));
```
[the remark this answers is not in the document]
1994ra: I agree, In future maybe we can create a function to calculate and fill these values for all the processes

PR #10864, opened 2025-04-15
[PWGCF] DptDpt - Tracking the insertion of pT bins of interest

victor-gonzalez:
In hyperloop, modifying the pT bins of interest makes the tests to fail while locally they don't fail
Tracking the insertion of the pT bins of interest will allow to check what the issue is


alibuild (approved):

PR #10876, opened 2025-04-16
[PWGHF] Added centrality to D+ tree tables

Marcellocosti:
This PR adds the centrality information in the D+ tables in the `treeCreatorDplusToPiKPi.cxx` workflow


zhangbiao-phy:
Ciao @Marcellocosti, Thanks for the development! looks good to me! I only have one comment, could you also keep the default process case without the centrality option for data and mc? since we always want to keep it for the people who are interested in the MB analysis.

Marcellocosti:
Hi @zhangbiao-phy, thanks for the suggestion! I tried to implement your request, by introducing two additional process functions. Please let me know if this is what you meant or further adjustments are needed!

vkucera:
If you can't get that commit, try ed51ab4 which is still in your remote branch.

Marcellocosti:
Dear @vkucera, indeed I managed to recover the original PR, thanks for the help!

zhangbiao-phy (approved):

PR #10887, opened 2025-04-16
[PWGCF] DptDpt - Increasing the level of pT bins tracking

victor-gonzalez:
Also
- fixed `linter` magic number usage recommendations
- don't process individual tracks if only pairs are required


PWGCF/TwoParticleCorrelations/Tasks/dptDptEfficiencyAndQc.cxx
```diff
@@ -391,6 +392,9 @@ struct QADataCollectingEngine {
     using namespace analysis::dptdptfilter;
     using namespace o2::aod::track;
 
+    constexpr int kFiftyPerCent = 50.0f;
+    constexpr int kHundredPerCent = 100.0f;
```
vkucera, line 396: Shouldn't these be `float`?

PR #10949, opened 2025-04-23
[PWGHF] Add RCT selection in HF event selection utils

fgrosa:
Hi @mfaggin @zhangbiao-phy, this PR adds the possibility to apply the RCT event selection in the HF code. 
I still keep it draft for the pending questions I asked to the experts in the mail thread.

@pstahlhu @creetz16 @JaeYoonCHO @sh-lim (I could not find Krista, please add her if you know her github account) while implementing this I noticed that there is no event selection applied at all for the Xic+ -> XiPiPi, could you please add it in a dedicated PR, or do you prefer me to do it?

### DIFF
--- a/PWGHF/Utils/utilsEvSelHf.h
+++ b/PWGHF/Utils/utilsEvSelHf.h
@@ -207,7 +207,8 @@ struct HfEventSelection : o2::framework::ConfigurableGroup {
     }
   }
 
-  void initRctChecker() {
+  void initRctChecker()
+  {
     rctChecker.init(rctLabel.value, rctCheckZDC.value, rctTreatLimitedAcceptanceAsBad.value);
   }
 
--- a/PWGHF/Utils/utilsEvSelHf.h
+++ b/PWGHF/Utils/utilsEvSelHf.h
@@ -207,7 +207,9 @@ struct HfEventSelection : o2::framework::ConfigurableGroup {
   void init(o2::framework::HistogramRegistry& registry)
   {
     // we initialise the RCT checker
-    rctChecker.init(rctLabel.value, rctCheckZDC.value, rctTreatLimitedAcceptanceAsBad.value);
+    if (requireGoodRct) {
+      rctChecker.init(rctLabel.value, rctCheckZDC.value, rctTreatLimitedAcceptanceAsBad.value);
+    }
 
     // we initialise the summary object
     if (softwareTrigger.value != "") {
@@ -439,7 +441,9 @@ struct HfEventSelectionMc {
     configureFromDevice(device);
 
     // we initialise the RCT checker
-    rctChecker.init(rctLabel, rctCheckZDC, rctTreatLimitedAcceptanceAsBad);
+    if (requireGoodRct) {
+      rctChecker.init(rctLabel, rctCheckZDC, rctTreatLimitedAcceptanceAsBad);
+    }
 
     // we initialise histograms
     addHistograms(registry);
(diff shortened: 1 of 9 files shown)


PWGHF/Utils/utilsEvSelHf.h
```diff
@@ -197,6 +207,10 @@ struct HfEventSelection : o2::framework::ConfigurableGroup {
     }
   }
 
+  void initRctChecker() {
```
vkucera, line 210: Maybe we can take the opportunity to define a common `init` function and move the Zorro initialisation and `addHistograms` inside.
```suggestion
  void init(o2::framework::HistogramRegistry& registry) {
```

PWGHF/Utils/utilsEvSelHf.h
```diff
@@ -383,8 +408,17 @@ struct HfEventSelectionMc {
...
+        requireGoodRct = option.defaultValue.get<bool>();
+      } else if (option.name.compare("hfEvSel.rctLabel") == 0) {
+        rctLabel = option.defaultValue.get<std::string>();
+      } else if (option.name.compare("hfEvSel.rctCheckZDC") == 0) {
+        rctCheckZDC = option.defaultValue.get<bool>();
+      } else if (option.name.compare("hfEvSel.rctTreatLimitedAcceptanceAsBad") == 0) {
+        rctTreatLimitedAcceptanceAsBad = option.defaultValue.get<bool>();
       }
     }
+    rctChecker.init(rctLabel, rctCheckZDC, rctTreatLimitedAcceptanceAsBad);
```
vkucera, line 421: Should be done only if `requireGoodRct`.

PWGHF/Utils/utilsEvSelHf.h
```diff
@@ -197,6 +207,10 @@ struct HfEventSelection : o2::framework::ConfigurableGroup {
     }
   }
 
+  void initRctChecker() {
+    rctChecker.init(rctLabel.value, rctCheckZDC.value, rctTreatLimitedAcceptanceAsBad.value);
```
vkucera, line 211: Should be done only if `requireGoodRct`.

PWGHF/Utils/utilsEvSelHf.h
vkucera, line 421: Also here, we could have a common `init` functions that includes `addHistograms` and `configureFromDevice`.

PWGHF/Utils/utilsEvSelHf.h
fgrosa: it should not matter, because if `requireGoodRct` is false then the selection is not applied

PWGHF/Utils/utilsEvSelHf.h
fgrosa: Done

PWGHF/Utils/utilsEvSelHf.h
fgrosa: Done

PWGHF/Utils/utilsEvSelHf.h
vkucera: That's fine, but my point is that `rctChecker.init` should not be called unnecessarily.

fgrosa: Done

pstahlhu:
Hi @fgrosa,
thanks for noticing this issue. I will add the event selection for the Xic+ -> Xi Pi Pi  in a separate PR.
Tagging @creetz16 @JaeYoonCHO @sh-lim @klsmith15k for information.

mfaggin (approved):

PR #10967, opened 2025-04-24
[PWGDQ] Process using realigned muon track in table-maker-MC-with-assoc

ppoava:
1. This follows the previous 826faec which implemented a process in standard table-maker using realigned muon tracks; the same implementation is now made for table-maker-mc-with-assoc.
2. Two new process (pp and PbPb) are added into table-maker-mc-with-assoc using realigned muons.


ppoava:
1. This follows the previous https://github.com/AliceO2Group/O2Physics/commit/826faec55b2bfc1e870642a3abfb2e117df9fb84 which implemented a process in standard table-maker using realigned muon tracks; the same implementation is now made for table-maker-mc-with-assoc.
2. Two new process (pp and PbPb) are added into table-maker-mc-with-assoc using realigned muons.

PR #10972, opened 2025-04-24
[PWGHF] Add KFParticle Quality Assurance in XicToXiPi candidate creator

Rrantu:
Evaluate the resolution of the particles reconstructed using the KFParticle method.

### DIFF
--- a/PWGHF/DataModel/CandidateReconstructionTables.h
+++ b/PWGHF/DataModel/CandidateReconstructionTables.h
@@ -1369,9 +1369,9 @@ DECLARE_SOA_COLUMN(MassCascChi2OverNdf, massCascChi2OverNdf, float);
 DECLARE_SOA_COLUMN(CascRejectInvmass, cascRejectInvmass, float);
 
 // Kf QA results:
-DECLARE_SOA_COLUMN(InvMassV0Sig, invMassV0Sig, float);
-DECLARE_SOA_COLUMN(InvMassXiSig, invMassXiSig, float);
-DECLARE_SOA_COLUMN(InvMassXic0Sig, invMassXic0Sig, float);
+DECLARE_SOA_COLUMN(InvMassV0Err, invMassV0Err, float);
+DECLARE_SOA_COLUMN(InvMassXiErr, invMassXiErr, float);
+DECLARE_SOA_COLUMN(InvMassXic0Err, invMassXic0Err, float);
 DECLARE_SOA_COLUMN(V0DauPosX, v0DauPosX, float);
 DECLARE_SOA_COLUMN(V0DauPosY, v0DauPosY, float);
 DECLARE_SOA_COLUMN(V0DauPosZ, v0DauPosZ, float);
@@ -1611,7 +1611,7 @@ DECLARE_SOA_TABLE(HfCandToXiPiKf, "AOD", "HFCANDTOXIPIKF", //!
 
 DECLARE_SOA_TABLE(HfCandToXiPiKfQa, "AOD", "HFCANDTOXIPIKFQA",
                   o2::soa::Index<>,
-                  hf_cand_xic0_omegac0::InvMassLambda, hf_cand_xic0_omegac0::InvMassCascade, hf_cand_xic0_omegac0::InvMassCharmBaryon, hf_cand_xic0_omegac0::InvMassV0Sig, hf_cand_xic0_omegac0::InvMassXiSig, hf_cand_xic0_omegac0::InvMassXic0Sig,
+                  hf_cand_xic0_omegac0::InvMassLambda, hf_cand_xic0_omegac0::InvMassCascade, hf_cand_xic0_omegac0::InvMassCharmBaryon, hf_cand_xic0_omegac0::InvMassV0Err, hf_cand_xic0_omegac0::InvMassXiErr, hf_cand_xic0_omegac0::InvMassXic0Err,
                   hf_cand_xic0_omegac0::CollisionId, hf_track_index::V0Id, v0data::PosTrackId, v0data::NegTrackId, hf_cand_xic0_omegac0::CascadeId, hf_cand_xic0_omegac0::BachelorFromCharmBaryonId, cascdata::BachelorId,
                   hf_cand_xic0_omegac0::V0DauPosX, hf_cand_xic0_omegac0::V0DauPosY, hf_cand_xic0_omegac0::V0DauPosZ, hf_cand_xic0_omegac0::V0DauPosXError, hf_cand_xic0_omegac0::V0DauPosYError, hf_cand_xic0_omegac0::V0DauPosZError, hf_cand_xic0_omegac0::V0DauPosPt,
                   hf_cand_xic0_omegac0::V0DauNegX, hf_cand_xic0_omegac0::V0DauNegY, hf_cand_xic0_omegac0::V0DauNegZ, hf_cand_xic0_omegac0::V0DauNegXError, hf_cand_xic0_omegac0::V0DauNegYError, hf_cand_xic0_omegac0::V0DauNegZError, hf_cand_xic0_omegac0::V0DauNegPt,
--- a/PWGHF/DataModel/CandidateReconstructionTables.h
+++ b/PWGHF/DataModel/CandidateReconstructionTables.h
@@ -1368,6 +1368,58 @@ DECLARE_SOA_COLUMN(MassV0Chi2OverNdf, massV0Chi2OverNdf, float);
 DECLARE_SOA_COLUMN(MassCascChi2OverNdf, massCascChi2OverNdf, float);
 DECLARE_SOA_COLUMN(CascRejectInvmass, cascRejectInvmass, float);
 
+// Kf QA results:
+DECLARE_SOA_COLUMN(InvMassV0Sig, invMassV0Sig, float);
+DECLARE_SOA_COLUMN(InvMassXiSig, invMassXiSig, float);
+DECLARE_SOA_COLUMN(InvMassXic0Sig, invMassXic0Sig, float);
+DECLARE_SOA_COLUMN(V0DauPosX, v0DauPosX, float);
+DECLARE_SOA_COLUMN(V0DauPosY, v0DauPosY, float);
+DECLARE_SOA_COLUMN(V0DauPosZ, v0DauPosZ, float);
+DECLARE_SOA_COLUMN(V0DauPosXError, v0DauPosXError, float);
+DECLARE_SOA_COLUMN(V0DauPosYError, v0DauPosYError, float);
+DECLARE_SOA_COLUMN(V0DauPosZError, v0DauPosZError, float);
+DECLARE_SOA_COLUMN(V0DauPosPt, v0DauPosPt, float);
+DECLARE_SOA_COLUMN(V0DauNegX, v0DauNegX, float);
+DECLARE_SOA_COLUMN(V0DauNegY, v0DauNegY, float);
+DECLARE_SOA_COLUMN(V0DauNegZ, v0DauNegZ, float);
+DECLARE_SOA_COLUMN(V0DauNegXError, v0DauNegXError, float);
+DECLARE_SOA_COLUMN(V0DauNegYError, v0DauNegYError, float);
+DECLARE_SOA_COLUMN(V0DauNegZError, v0DauNegZError, float);
+DECLARE_SOA_COLUMN(V0DauNegPt, v0DauNegPt, float);
+
+DECLARE_SOA_COLUMN(V0VtxX, v0VtxX, float);
+DECLARE_SOA_COLUMN(V0VtxY, v0VtxY, float);
+DECLARE_SOA_COLUMN(V0VtxZ, v0VtxZ, float);
+DECLARE_SOA_COLUMN(V0XError, v0XError, float);
+DECLARE_SOA_COLUMN(V0YError, v0YError, float);
+DECLARE_SOA_COLUMN(V0ZError, v0ZError, float);
+DECLARE_SOA_COLUMN(V0Pt, v0Pt, float);
+DECLARE_SOA_COLUMN(XiBachelorX, xiBachelorX, float);
+DECLARE_SOA_COLUMN(XiBachelorY, xiBachelorY, float);
+DECLARE_SOA_COLUMN(XiBachelorZ, xiBachelorZ, float);
+DECLARE_SOA_COLUMN(XiBachelorPt, xiBachelorPt, float);
+DECLARE_SOA_COLUMN(XiBachelorXError, xiBachelorXError, float);
+DECLARE_SOA_COLUMN(XiBachelorYError, xiBachelorYError, float);
+DECLARE_SOA_COLUMN(XiBachelorZError, xiBachelorZError, float);
+DECLARE_SOA_COLUMN(XiX, xiX, float);
+DECLARE_SOA_COLUMN(XiY, xiY, float);
+DECLARE_SOA_COLUMN(XiZ, xiZ, float);
+DECLARE_SOA_COLUMN(XiXError, xiXError, float);
+DECLARE_SOA_COLUMN(XiYError, xiYError, float);
+DECLARE_SOA_COLUMN(XiZError, xiZError, float);
+DECLARE_SOA_COLUMN(XiPt, xiPt, float);
+DECLARE_SOA_COLUMN(Xic0BachelorX, xic0BachelorX, float);
+DECLARE_SOA_COLUMN(Xic0BachelorY, xic0BachelorY, float);
+DECLARE_SOA_COLUMN(Xic0BachelorZ, xic0BachelorZ, float);
+DECLARE_SOA_COLUMN(Xic0BachelorPt, xic0BachelorPt, float);
+DECLARE_SOA_COLUMN(Xic0BachelorXError, xic0BachelorXError, float);
+DECLARE_SOA_COLUMN(Xic0BachelorYError, xic0BachelorYError, float);
+DECLARE_SOA_COLUMN(Xic0BachelorZError, xic0BachelorZError, float);
+DECLARE_SOA_COLUMN(Xic0Pt, xic0Pt, float);
+DECLARE_SOA_COLUMN(Xic0XError, xic0XError, float);
+DECLARE_SOA_COLUMN(Xic0YError, xic0YError, float);
+DECLARE_SOA_COLUMN(Xic0ZError, xic0ZError, float);
+
 // MC matching result:
 DECLARE_SOA_COLUMN(FlagMcMatchRec, flagMcMatchRec, int8_t); // reconstruction level
 DECLARE_SOA_COLUMN(DebugMcRec, debugMcRec, int8_t);         // debug flag for mis-association reconstruction level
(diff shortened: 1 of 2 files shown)


vkucera:
Hi @Rrantu , such a huge PR deserves some comments.
- First of all, what is your goal?
- Second, why do you need so many variables, new structs, histograms and tables?
- Why do you need to do further MC matching?
- How much more memory does the extra stuff cost?
- Last but not least, fix the issues first, only then make the PR ready.

Rrantu:
Hi @vkucera , I'm trying to check the resolution of the particles reconstructed using the KFParticle method. For this, I need the information saved in runXic0Omegac0Creator function, Therefore, I’ve added a new tree within the runXic0Omegac0Creator function to store these variables.
Because KFParticle reconstruction starts from V0s, my MC matching also needs to begin from V0s. So, I’ve written a new process function to read the new tree and do the MC matching accordingly.
I’m not exactly sure how much memory this will consume, but it should be minimal because the new tree is relatively small.
And for the O2 linter issues, they were not introduced by my changes.

zhangbiao-phy:
hi @Rrantu, Thanks for the implementation! I have a few comments. See the review

PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx
```diff
@@ -2386,6 +2639,392 @@ struct HfCandidateCreatorXic0Omegac0Mc {
...
+    float xic0YDelta;
+    float xic0ZDelta;
+    float xic0Pt;
+    float xic0XPull;
+    float xic0YPull;
+    float xic0ZPull;
+
+    float cascXDelta;
+    float cascYDelta;
+    float cascZDelta;
```
zhangbiao-phy, line 2722: Can you simplify the initialization and assignment of these variables? You can do it all inside the candidate loop. For example: `auto v0MassPull = (candidate.invMassLambda() - MassLambda0) / candidate.invMassV0Sig();`

Rrantu: Thanks for the feedback, I’ve updated the code accordingly.

zhangbiao-phy (approved):

zhangbiao-phy:
> Hi @vkucera , I'm trying to check the resolution of the particles reconstructed using the KFParticle method. For this, I need the information saved in runXic0Omegac0Creator function, Therefore, I’ve added a new tree within the runXic0Omegac0Creator function to store these variables. Because KFParticle reconstruction starts from V0s, my MC matching also needs to begin from V0s. So, I’ve written a new process function to read the new tree and do the MC matching accordingly. I’m not exactly sure how much memory this will consume, but it should be minimal because the new tree is relatively small. And for the O2 linter issues, they were not introduced by my changes.

Hi @Rrantu, Thanks! Please fix the O2 linter issues next time!

PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx
```diff
@@ -236,6 +237,65 @@ struct HfCandidateCreatorXic0Omegac0 {
     float chi2MassCasc;
     float etaXic;
   } kfXic0Candidate;
+
```
fgrosa, line 240: following @vkucera, why do you need a struct here? Can't you directly fill the table?

PWGHF/DataModel/CandidateReconstructionTables.h
```diff
@@ -1368,6 +1368,58 @@ DECLARE_SOA_COLUMN(MassV0Chi2OverNdf, massV0Chi2OverNdf, float);
 DECLARE_SOA_COLUMN(MassCascChi2OverNdf, massCascChi2OverNdf, float);
 DECLARE_SOA_COLUMN(CascRejectInvmass, cascRejectInvmass, float);
 
+// Kf QA results:
+DECLARE_SOA_COLUMN(InvMassV0Sig, invMassV0Sig, float);
+DECLARE_SOA_COLUMN(InvMassXiSig, invMassXiSig, float);
+DECLARE_SOA_COLUMN(InvMassXic0Sig, invMassXic0Sig, float);
```
fgrosa, line 1374: What is the difference between these invariant masses and those in the usual table?

PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx
```diff
@@ -1576,6 +1638,63 @@ struct HfCandidateCreatorXic0Omegac0 {
       registry.fill(HIST("hKfXiC0_ldl"), kfXic0Candidate.ldlXic);
       registry.fill(HIST("hDcaXYCascadeToPVKf"), kfXic0Candidate.kfDcaXYCascToPv);
 
+      // QA
+      kfXic0Qa.v0DauPosX = kfPos.GetX();
```
fgrosa, line 1642: why filling first the struct and then the table? Can't you directly fill the table?

PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx
```diff
@@ -1896,6 +2027,128 @@ struct HfCandidateCreatorXic0Omegac0Mc {
...
     hGenCharmBaryonPtRapidityLooseOmegacToOmegaK = registry.add<TH1>("hGenCharmBaryonPtRapidityLooseOmegacToOmegaK", "Generated charm baryon #it{p}_{T};#it{p}_{T} (GeV/#it{c});entries", {HistType::kTH1D, {{20, 0.0, 20.0}}});
+
+    // QA
+    if (doprocessMcXicToXiPiKfQa) {
+      // mass over pt
+      registry.add("hV0MassPullVsPt", "m_{PULL}(V0) vs. p_{T}", HistType::kTH2F, {{20, 0., 20.}, {2000, -10., 10.}});
+      registry.add("hXiMassPullVsPt", "m_{PULL}(#Xi^{-}) vs. p_{T}", HistType::kTH2F, {{20, 0., 20.}, {2000, -10., 10.}});
+      registry.add("hXic0MassPullVsPt", "m_{PULL}(#Xic0) vs. p_{T}", HistType::kTH2F, {{20, 0., 20.}, {2000, -10., 10.}});
+      // delta
+      registry.add("hV0DauPosXDelta", "x^{p} - x^{MC}", kTH1F, {{2000, -10., 10.}});
```
fgrosa, line 2038: Replace all `TH1F` with `TH1D`
Do you also need all these histograms since you are filling a table with the same variables?

PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx
```diff
@@ -1617,6 +1736,18 @@ struct HfCandidateCreatorXic0Omegac0 {
                          v0NDF, cascNDF, charmbaryonNDF, v0Ndfm, cascNdfm,
                          v0Chi2OverNdf, cascChi2OverNdf, charmbaryonChi2OverNdf, v0Chi2OverNdfm, cascChi2OverNdfm);
 
+      // fill QA table
```
fgrosa, line 1739: This should be filled only if the QA is enabled and not by default

fgrosa (changes requested):
Hi @Rrantu, thanks for the development! I have however a few comments (some of them follow the doubts raised by @vkucera). You can always retrieve the info about the matching from the "default" table and, if you want to add more flags for matching for intermediate states, you can do it in the MC process function as well without duplicating

PWGHF/DataModel/CandidateReconstructionTables.h
Rrantu: Hi @fgrosa, Thanks for your comments, these columns contain the uncertainty of the invariant mass, which is not included in the default table.

PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx
zhangbiao-phy: I agree, it's useless, we can remove it and fill the table directly

Rrantu: Thank you for your feedback, I've update the code as you suggested.

PWGHF/DataModel/CandidateReconstructionTables.h
fgrosa: Thanks for the clarification! Maybe use a more meaningful name (to make explicit that it is an uncertainty)

Rrantu: Thanks, I've changed the column name.

PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx
Rrantu: Hi @fgrosa, I’ve made the changes as you suggested. Could you please approve them?

PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx
```diff
@@ -1896,6 +1913,128 @@ struct HfCandidateCreatorXic0Omegac0Mc {
 
     hGenCharmBaryonPtRapidityTightOmegacToOmegaK = registry.add<TH1>("hGenCharmBaryonPtRapidityTightOmegacToOmegaK", "Generated charm baryon #it{p}_{T};#it{p}_{T} (GeV/#it{c});entries", {HistType::kTH1D, {{20, 0.0, 20.0}}});
     hGenCharmBaryonPtRapidityLooseOmegacToOmegaK = registry.add<TH1>("hGenCharmBaryonPtRapidityLooseOmegacToOmegaK", "Generated charm baryon #it{p}_{T};#it{p}_{T} (GeV/#it{c});entries", {HistType::kTH1D, {{20, 0.0, 20.0}}});
+
+    // QA
+    if (doprocessMcXicToXiPiKfQa) {
+      // mass over pt
+      registry.add("hV0MassPullVsPt", "m_{PULL}(V0) vs. p_{T}", HistType::kTH2D, {{20, 0., 20.}, {2000, -10., 10.}});
+      registry.add("hXiMassPullVsPt", "m_{PULL}(#Xi^{-}) vs. p_{T}", HistType::kTH2D, {{20, 0., 20.}, {2000, -10., 10.}});
+      registry.add("hXic0MassPullVsPt", "m_{PULL}(#Xic0) vs. p_{T}", HistType::kTH2D, {{20, 0., 20.}, {2000, -10., 10.}});
```
vkucera, line 1922: Avoid hard-coded duplication of axis definitions by using axis objects.

PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx
```diff
@@ -1896,6 +1913,128 @@ struct HfCandidateCreatorXic0Omegac0Mc {
...
+      registry.add("hXic0BachelorYPull", "y^{PULL}", kTH1D, {{4000, -20., 20.}});
+      registry.add("hXic0BachelorZPull", "z^{PULL}", kTH1D, {{4000, -20., 20.}});
+
+      registry.add("hXic0XPull", "x^{PULL}(#Xi_{c}^{0})", kTH1D, {{4000, -20., 20.}});
+      registry.add("hXic0YPull", "y^{PULL}(#Xi_{c}^{0})", kTH1D, {{4000, -20., 20.}});
+      registry.add("hXic0ZPull", "z^{PULL}(#Xi_{c}^{0})", kTH1D, {{4000, -20., 20.}});
+      // pull over pt
+      registry.add("hV0DauPosXPullVsPt", "x_{PULL} vs. p_{T}", HistType::kTH2D, {{20, 0., 20.}, {4000, -20., 20.}});
+      registry.add("hV0DauPosYPullVsPt", "y_{PULL} vs. p_{T}", HistType::kTH2D, {{20, 0., 20.}, {4000, -20., 20.}});
+      registry.add("hV0DauPosZPullVsPt", "z_{PULL} vs. p_{T}", HistType::kTH2D, {{20, 0., 20.}, {4000, -20., 20.}});
```
vkucera, line 2005: That's a lot of bins. Please evaluate the extra RAM consumption when running.

PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx
```diff
@@ -2386,6 +2525,261 @@ struct HfCandidateCreatorXic0Omegac0Mc {
     } // close loop on MCCollisions
   } // close process
 
+  template <o2::hf_centrality::CentralityEstimator centEstimator, int decayChannel, typename TMyRecoCand>
+  void runXic0Omegac0McQa(TMyRecoCand const& candidates,
+                          MyTracksWMc const&,
+                          aod::McParticles const& mcParticles,
+                          BCsInfo const&)
+  {
```
vkucera, line 2533: Can the body of this function be done inside `runXic0Omegac0Mc` instead?

PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx
```diff
@@ -2386,6 +2525,261 @@ struct HfCandidateCreatorXic0Omegac0Mc {
     } // close loop on MCCollisions
   } // close process
 
+  template <o2::hf_centrality::CentralityEstimator centEstimator, int decayChannel, typename TMyRecoCand>
```
vkucera, line 2528:
```suggestion
  template <o2::hf_centrality::CentralityEstimator centEstimator, aod::hf_cand_xic0_omegac0::DecayType decayChannel, typename TMyRecoCand>
```
Please fix this elsewhere too.

PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx
```diff
@@ -2386,6 +2525,261 @@ struct HfCandidateCreatorXic0Omegac0Mc {
...
+
+      auto mcV0DauPos = arrayDaughtersV0[0].mcParticle();
+      auto mcV0DauNeg = arrayDaughtersV0[1].mcParticle();
+      auto mcXiBachelor = arrayDaughtersCasc[0].mcParticle();
+      auto mcXic0Bachelor = arrayDaughters[0].mcParticle();
+
+      // Xic0 -> xi pi matching
+      if constexpr (decayChannel != aod::hf_cand_xic0_omegac0::DecayType::XiczeroToXiPi) {
+        LOGF(info, "ERROR: Quality validation is restricted to Xic0 → Xi Pi decay processes at this stage");
+      }
```
vkucera, line 2566: I guess you want to throw a `LOGF(fatal,` here?

PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx
```diff
@@ -2386,6 +2525,261 @@ struct HfCandidateCreatorXic0Omegac0Mc {
...
+            }
+          }
+        }
+        // Check whether the charm baryon is non-prompt (from a b quark).
+        if (flag != 0) {
+          auto particle = mcParticles.rawIteratorAt(indexRecCharmBaryon);
+        }
+        if (debug == McMatchFlag::CascUnmatched || debug == McMatchFlag::V0Unmatched) {
+          LOGF(info, "WARNING: Xic0ToXiPi decays in the expected final state but the condition on the intermediate states are not fulfilled");
+        }
```
vkucera, line 2778: Use logging and severity properly. Do not write `WARNING`, use the correct logging level.
What do you want to do when his happens?

vkucera (changes requested):

PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx
vkucera: Also what is the purpose of the `debug` variable? Usually, it's a flag for logging level which is not the case here.

Rrantu: Hi @vkucera , thanks for your feedback. I’ve updated the code accordingly, and the debug variable wasn’t needed, so I removed it.

PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx
vkucera: No, they  don't:
```c++
  void runXic0Omegac0Mc(TMyRecoCand const& candidates,
                        MyTracksWMc const&,
                        aod::McParticles const& mcParticles,
                        Colls const& collsWithMcLabels,
                        McCollisions const& mcCollisions,
                        BCsInfo const&)
```

PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx
vkucera: It's still there.

PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx
```diff
@@ -118,6 +119,7 @@ struct HfCandidateCreatorXic0Omegac0 {
   Configurable<int> kfConstructMethod{"kfConstructMethod", 2, "KF Construct Method"};
   Configurable<bool> kfUseV0MassConstraint{"kfUseV0MassConstraint", false, "KF: use Lambda mass constraint"};
   Configurable<bool> kfUseCascadeMassConstraint{"kfUseCascadeMassConstraint", false, "KF: use Cascade mass constraint"};
+  Configurable<bool> kfResolutionQA{"kfResolutionQA", true, "KF: KFParticle Quality Assurance"};
```
vkucera, line 122:
```suggestion
  Configurable<bool> kfResolutionQA{"kfResolutionQA", false, "KF: KFParticle Quality Assurance"};
```
Do not make new features activated by default unless they are needed for a correct functionality of the code.

PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx
Rrantu: Sorry, Now the debug variable is properly removed.

PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx
Rrantu: Sorry for the earlier unclear explanation. I didn’t use runXic0Omegac0Mc because it fills MCGen and MCRec tables that I don’t need, and they consume a large amount of memory.
Also, the matching in runXic0Omegac0Mc goes from Xic0 to Λ0, while for my QA I actually need matching from Λ0 to Xic0, so the matching setup is different.
Since runXic0Omegac0Mc is used in many other analyses, modifying it directly could affect others. So I added an independent runXic0Omegac0McQA function to minimize the impact.

PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx
```diff
@@ -2386,6 +2536,239 @@ struct HfCandidateCreatorXic0Omegac0Mc {
...
+
+      auto mcV0DauPos = arrayDaughtersV0[0].mcParticle();
+      auto mcV0DauNeg = arrayDaughtersV0[1].mcParticle();
+      auto mcXiBachelor = arrayDaughtersCasc[0].mcParticle();
+      auto mcXic0Bachelor = arrayDaughters[0].mcParticle();
+
+      // Xic0 -> xi pi matching
+      if constexpr (decayChannel != aod::hf_cand_xic0_omegac0::DecayType::XiczeroToXiPi) {
+        LOGF(fatal, "ERROR: Quality validation is restricted to Xic0 → Xi Pi decay processes at this stage");
+      }
```
vkucera, line 2570:
```suggestion
      if constexpr (decayChannel != aod::hf_cand_xic0_omegac0::DecayType::XiczeroToXiPi) {
        LOGF(fatal, "Quality validation is restricted to Xic0 → Xi Pi decay processes at this stage");
      }
```
- The severity level is already defined with `fatal`. Adding a text "ERROR" is not only redundant but also misleading because `error` is a different severity level.
- Any sanity check should be done asap before any unrelated calculations.
- The result of this check is known at compile time. Making it fail at run time is too late. 
- What is the point of making `decayChannel` a parameter when only one value is valid?

vkucera (changes requested):

PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx
Rrantu: Thank you for the suggestion. The decayChannel selection was already applied before the match, so I have removed the check accordingly.

PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx
Rrantu: Thanks for the feedback. I have reduced the number of bins.

PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx
Rrantu: Hi @vkucera, I’ve made the changes as you suggested. Could you please approve them?

fgrosa:
@vkucera @Rrantu have all the comments been implemented? Can we merge this?

Rrantu:
> @vkucera @Rrantu have all the comments been implemented? Can we merge this?
I believe I’ve addressed all the comments. Please let me know if anything else is needed before merging.

PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx
vkucera: Let's see if it runs fine, but please check the extra memory consumption. It's still a lot of bins.

vkucera (approved):

PR #10973, opened 2025-04-24
[PWGCF] Add class members back into ROOT streamer

EmilGorm:


victor-gonzalez (approved):
At some point the `linter` recommendation about not using `using` directives at global scope in headers have to be fulfilled

PR #11005, opened 2025-04-28
[Trigger] Fix 2phi filter table descriptor for real

mpuccio:


alibuild (approved):

PR #11011, opened 2025-04-28
[PWGHF] Add HF event selection and centrality process functions to Xic to Xi Pi Pi

pstahlhu:
@fgrosa @JaeYoonCHO @sh-lim @klsmith15k

### DIFF
--- a/PWGHF/TableProducer/candidateCreatorXicToXiPiPi.cxx
+++ b/PWGHF/TableProducer/candidateCreatorXicToXiPiPi.cxx
@@ -243,8 +243,8 @@ struct HfCandidateCreatorXicToXiPiPi {
       std::array<float, 21> covCasc = {0.};
 
       //----------------create cascade track------------------------------------------------------------
-      constexpr int MomInd[6] = {9, 13, 14, 18, 19, 20}; // cov matrix elements for momentum component
-      for (int i = 0; i < 6; i++) {
+      constexpr std::array<int, 6> MomInd = {9, 13, 14, 18, 19, 20}; // cov matrix elements for momentum component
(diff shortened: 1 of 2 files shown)


vkucera:
Looks like more than just event selection. ;-)

pstahlhu:
Hi @vkucera,@fgrosa, yes it became a bit more than expected :sweat_smile:
Do you have any comments?

vkucera:
> Hi @vkucera,@fgrosa, yes it became a bit more than expected 😅 Do you have any comments?

Please update the PR title accordingly.

vkucera (approved):

PWGHF/TableProducer/candidateCreatorXicToXiPiPi.cxx
```diff
@@ -741,7 +943,7 @@ struct HfCandidateCreatorXicToXiPiPiExpressions {
           }
           if (indexRec > -1) {
             RecoDecay::getDaughters(mcParticles.rawIteratorAt(indexRecXicPlus), &arrDaughIndex, std::array{0}, 1);
-            if (arrDaughIndex.size() == 2) {
+            if (arrDaughIndex.size() == arrXiResonance.size()) {
```
vkucera, line 946: It is not needed to call `size` every time since the array size is a compile time constant which can be declared as
```c++
constexpr std::size_t NDaughtersResonant{2u};
```
and used in the declaration of `arrXiResonance`.

vkucera (changes requested):

vkucera:
Thanks for the update. Please also fix the missing include.

vkucera (approved):

PR #11018, opened 2025-04-29
[PWGLF] renamed file and fixed cluster size calculation

alcaliva:
ITS clusters belonging to the bachelor's track are excluded from the cluster size calculation of tracked cascades


njacazio (approved):
Looks good, thx!

PR #11027, opened 2025-04-29
[Infrastructure] Updating codeowner rights for PAG Strangeness and GEP

romainschotter:
Updating codeowner rights according to the current Strangeness and Global Event Properties PAG coordinators: @romainschotter and @abmodak


njacazio:
Thanks @romainschotter indeed, this reflects the change in coordination.
@ddobrigk could you see that both are added to the repository users?

ddobrigk (approved):

PR #11030, opened 2025-04-29
[PWGEM/PhotonMeson] Bug fixes for BC wise pi0 task

nstrangm:
- Filters in emcalBcWisePi0.cxx now correctly use the configurable values and not just the defaults -> changed a few data types from unsigned to signed int to allow for filter comparison
- z vertex position now correctly read as int16 (instead of int8)
- Added MC pi0 min pT configurable to reduce derived data size
- Added check for mc particle ID in map before calling its value (map does not contain all pi0s, as not all pi0s are stored)


mhemmer-cern (approved):
Looks good

PR #11031, opened 2025-04-29
[PWGLF] Fix bug in to-build cascade list mapping

ddobrigk:
@fcatalan92 this fixes the bug that manifested itself in the omegaC when building in the KF path. 

The issue is that in a recent commit I introduced a check for TPC-only V0s being used in cascades in the first entry point of the builder and removed those cascades from the pool, which accidentally renders the joinability between `Cascades` and the `Cascades`-to-[]Data interlinks false because a tiny amount of candidates is rejected and the interlink tables are smaller. This is inadequately reported by the framework for some reason in the particular use case of the OmegaC building, since there was a segmentation fault instead of a message of "table size mismatch" (tagging also @aalkin - I suspect it may be because we are dealing with linked derived data here, not sure). I found out the table size mismatch by isolating the joined table combination that led to the crash manually.

In the fix, I simply don't filter out cascades that use tpc-only V0s; these are filtered from actual candidate generation anyhow a posteriori via a build call that includes the protection added by #10908, but by not removing the cascades from the build pool, the interlinks will have the same size as the original `Cascades` table. 

Tagging also @romainschotter @gianniliveraro; further tagging @f3sch - we should still find out why cascades seem to be using TPC-only V0s as this should not happen in the svertexer from what I remember.

### DIFF
--- a/PWGLF/TableProducer/Strangeness/strangenessbuilder.cxx
+++ b/PWGLF/TableProducer/Strangeness/strangenessbuilder.cxx
@@ -29,7 +29,7 @@
 //                           task will adapt algorithm to spare / spend CPU accordingly
 //  -- mc_findableMode ....: 0: only found (default), 1: add findable to found, 2: all findable
 //                           When using findables, refer to FoundTag bools for checking if found
-//  -- v0builderopts ......: V0-specific building options (topological, etc)
+//  -- v0builderopts ......: V0-specific building options (topological, deduplication, etc)
 //  -- cascadebuilderopts .: cascade-specific building options (topological, etc)
 
 #include <string>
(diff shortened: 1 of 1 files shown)


ddobrigk:
The compilation error seems unrelated :-(

aalkin:
Hello @ddobrigk, any kind of Join would produce an error about different size, there is no other code path, so there may be something else at play here. Do you `reserve()` the size of the produced tables anywhere?

ddobrigk:
> Hello @ddobrigk, any kind of Join would produce an error about different size, there is no other code path, so there may be something else at play here. Do you `reserve()` the size of the produced tables anywhere?

Hi @aalkin, I'm afraid in this case, the error is not reported properly - that's what I mean. I'll contact you privately with the log and the test I did to illustrate that the tables are, in fact, not joinable. About `reserve()`, I did not call it but it's irrelevant in the code path because the bug of different sizes would still be there if `reserve()` was called...

fcatalan92:
Thanks @ddobrigk for the fix!

@mtorresc15

njacazio (approved):

PR #11037, opened 2025-04-30
[PWGDQ] Fixing issue for dqEfficiency_withAssoc

lucamicheletti93:
Issue of dqEfficiency_withAssoc when running on hypeloop (not when running locally on MacOs). Test moving to a single process function to take into accout all and selected only MC collisions.


alibuild (approved):

PR #11051, opened 2025-05-02
[ALICE3] A3TOF: compute multiplicity in any case, add possibility to switch between p and pt

njacazio:


alibuild (approved):

PR #11076, opened 2025-05-04
[PWGUD] Add some cuts

zhaomr13:


PWGUD/Tasks/flowCorrelationsUpc.cxx
```diff
@@ -233,7 +233,8 @@ struct FlowCorrelationsUpc {
     }
 
     int gapSide = collision.gapSide();
-    if (gapSide < 0 || gapSide > 2) {
+    // The gapSide are the selection criteria from the UPC from, not `magic number`
+    if (gapSide < 0 || gapSide > 2) { // o2-linter: disable=magic-number
```
vkucera: Why are you trying to silence a true positive instead of fixing it?!

vkucera (changes requested):

rolavick (approved):

PR #11089, opened 2025-05-05
[PWGUD] Implement luminosity calculation in SG candidate producer

nburmaso:
@rolavick, could you please check?

NB: To keep original code structure (with BCs, tracks etc. grouped by collision), I have created a separate process, since we need to loop over all BCs

Cheers,
Nazar


nburmaso:
Also, should I fix linter warnings? There are a lot of errors from it, with some caused by struct and variable naming

abylinkin (approved):

PR #11116, opened 2025-05-07
[PWGJE] Modify cuts for mother hadrons and decay photons

pestratm:
I removed the requirement that the daughter photons from neutral hadrons be in the acceptance (for the decay kinematics) and introduced the requirement that MC particles do not have a generator status code -1


pestratm:
Fixed a bug where for some particles, associated particles and trigger particles are switched when calculating the Δη and Δφ

pestratm:
Made more axes configurable

pestratm:
Histograms are now filled using MC weights. 
Added Sigma0 (Sigma0_bar) to possible sources of photons.

nzardosh (approved):

PR #11125, opened 2025-05-08
[PWGHF] Update MC matching in XicToXiPiPi workflow

pstahlhu:
@JaeYoonCHO @sh-lim @klsmith15k

- Update MC matching to include the latest updates to RecoDecay functions from recent PRs
- remove debug flags from the output trees (unnecessary memory consumption, different logic between reconstructed + generated), instead add counter histogram for the matching of the reconstructed candidates
- add histograms for the number of decayed pions and number of material interactions of daughter particles from matched $\Xi_{\rm{c}}^{+} \rightarrow \Xi^{-} \pi^{+} \pi^{+}$ decays
- Use a filter for filling of generated particle tables in the tree creator


zhangbiao-phy:
Thanks @pstahlhu, Looks good to me!

zhangbiao-phy (approved):

PR #11137, opened 2025-05-09
[PWGCF] fixed typo and added TOF tables

Eloviyo:


victor-gonzalez (approved):
`linter` recommendations are meaningful because they have an impact in the execution performance
Consider its implementation in future iterations

PR #11146, opened 2025-05-09
[PWGHF] fix linter error and only apply ml when it needed in the Lcp femto producer

zhangbiao-phy:


PWGHF/HFC/TableProducer/femtoDreamProducer.cxx
```diff
@@ -70,6 +70,12 @@ enum MlMode : uint8_t {
   kFillMlFromNewBDT
 };
 
+enum ProcessType {
+  kUndefined = -1,
+  kProcessDecay = 4,
+  kProcessInelasticHadronic = 23
+};
```
vkucera, line 77: Do not redefine constants that already exist in `TMCProcess`.

vkucera, line 77: Remove the prefixes `k` from all the `enum` values.

vkucera (changes requested):

PWGHF/HFC/TableProducer/femtoDreamProducer.cxx
zhangbiao-phy: fixed

zhangbiao-phy: fixed

vkucera (approved):

PR #11151, opened 2025-05-09
[PWGCF] Add directory Femto and pi-d analysis workflow

blacwovie:


shouqiye (approved):

victor-gonzalez:
@blacwovie Thanks for moving forward this analysis|
This, I guess, implies new derived data datamodel
Have you shared with Femto PAG coordinators (@lauraser and Dimitar) the analysis aim and the proposed datamodel

victor-gonzalez:
> > @blacwovie Thanks for moving forward this analysis| This, I guess, implies new derived data datamodel Have you shared with Femto PAG coordinators (@lauraser and Dimitar) the analysis aim and the proposed datamodel
> 
> Thanks for checking, Victor. Yes, the Femto PAG has been informed, and to avoid interfering with the existing structure, we have created this new one

@shouqiye @otonvd So, the limiting on the amount of produced derived data is over then

PR #11160, opened 2025-05-12
[PWGCF] FemtoUniverse -- Adding DCA plots

zchochul:
Adding DCA plots needed for the template fit for the p-phi analysis. This will add this option also for other correlations.


alibuild (approved):

PR #11174, opened 2025-05-13
[PWGHF] Fix names of Omegac0 tree creator files

vkucera:


vkucera:
@fcatalan92 @Yunfan-Liu Please pick a moment to fix the name of the struct.

PR #11187, opened 2025-05-14
[PWGHF] Unify PID method configuration in selectors

vkucera:


zhangbiao-phy (approved):
Thanks @vkucera! good to me!

PR #11188, opened 2025-05-14
[PWGJE] removing std::optional

nzardosh:


alibuild (approved):

mhemmer-cern:
Hey @nzardosh, why the change from std::optional to references? Is there a performance reason?

nzardosh:
Hi @mhemmer-cern it was pointed out to me by Vit that std::optional creates a copy of the variable which might not be the best performance wise

PR #11191, opened 2025-05-14
[PWGHF,PWGCF] Split same-event table and add the collision infos into lcp femto task 

zhangbiao-phy:
For convenience in doing the mixed event locally, I added the collision index to the table 

- Separate into 3 tables: charm, track, and collision
- Add eta and phi into the table --> calculate k*, kT, mT locally
- Remove table filling in the mix-event function (if the test is smooth, this part will be deleted later)

### DIFF
--- a/PWGCF/DataModel/FemtoDerived.h
+++ b/PWGCF/DataModel/FemtoDerived.h
@@ -312,7 +312,7 @@ DECLARE_SOA_TABLE(FDHfCand, "AOD", "FDHFCAND", //! Table to store the derived da
                   fdhf::Phi<fdhf::Prong0Pt, fdhf::Prong0Phi, fdhf::Prong0Eta, fdhf::Prong1Pt, fdhf::Prong1Phi, fdhf::Prong1Eta, fdhf::Prong2Pt, fdhf::Prong2Phi, fdhf::Prong2Eta>,
                   fdhf::Pt<fdhf::Prong0Pt, fdhf::Prong0Phi, fdhf::Prong0Eta, fdhf::Prong1Pt, fdhf::Prong1Phi, fdhf::Prong1Eta, fdhf::Prong2Pt, fdhf::Prong2Phi, fdhf::Prong2Eta>);
 
-DECLARE_SOA_TABLE(FDResultsHFCharm, "AOD", "FDRESULTSHFCHARM", //! table to store results for HF femtoscopy
+DECLARE_SOA_TABLE(FDResultsCharm, "AOD", "FDRESULTSCHARM", //! table to store results for HF femtoscopy
                   fdhf::GIndexCol,
                   fdhf::TimeStamp,
                   fdhf::CharmM,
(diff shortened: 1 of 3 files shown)


zhangbiao-phy:
The linter error will not be fixed in this pr

PWGCF/DataModel/FemtoDerived.h
```diff
@@ -210,7 +210,8 @@ enum CharmHadronMassHypo {
   lcToPKPi = 1,
   lcToPiKP = 2
 };
-
+DECLARE_SOA_COLUMN(GIndexCol, gIndexCol, int);                      //! Global index for the collision
```
vkucera, line 213: Why is this a normal column and how do you plan to use it, specifically outside or inside O2Physics?

zhangbiao-phy: yes, I use it outside for mixing events locally. I will merge the DF directly, so I also add timestamp

PWGCF/DataModel/FemtoDerived.h
```diff
@@ -305,31 +312,39 @@ DECLARE_SOA_TABLE(FDHfCand, "AOD", "FDHFCAND", //! Table to store the derived da
...
-                  fdhf::Mult,
-                  fdhf::MultPercentile,
-                  fdhf::Charge,
-                  fdhf::PairSign,
-                  fdhf::ProcessType,
                   fdhf::FlagMc,
                   fdhf::OriginMcRec);
 
-DECLARE_SOA_TABLE(FDResultsHFTrkInfo, "AOD", "FDRESULTSHFTRKINFO", //! table to store results for HF femtoscopy
+DECLARE_SOA_TABLE(FDResultsTrk, "AOD", "FDRESULTSTRK", //! table to store results for HF femtoscopy
```
victor-gonzalez, line 329: Please, keep `HF` as before in the table declaration

PWGCF/DataModel/FemtoDerived.h
```diff
@@ -305,31 +312,39 @@ DECLARE_SOA_TABLE(FDHfCand, "AOD", "FDHFCAND", //! Table to store the derived da
...
+                  fdhf::TrkEta,
+                  fdhf::TrkPhi,
+                  femtodreamparticle::Sign,
                   femtodreamparticle::TPCNClsFound,
                   track::TPCNClsFindable,
                   femtodreamparticle::TPCNClsCrossedRows,
                   femtodreamparticle::TPCNSigmaPr,
                   femtodreamparticle::TOFNSigmaPr);
 
+DECLARE_SOA_TABLE(FDResultsColl, "AOD", "FDRESULTSCOLL", //! table to store results for HF femtoscopy
```
victor-gonzalez, line 342: Please, incorporate `HF` in the table declaration

victor-gonzalez (changes requested):

PWGCF/DataModel/FemtoDerived.h
zhangbiao-phy: Ok, Fixed

PWGHF/HFC/Tasks/taskCharmHadronsFemtoDream.cxx
```diff
@@ -195,8 +189,9 @@ struct HfTaskCharmHadronsFemtoDream {
 
   SliceCache cache;
   Preslice<aod::FDParticles> perCol = aod::femtodreamparticle::fdCollisionId;
-  Produces<o2::aod::FDResultsHF> fillFemtoResult;
-  Produces<o2::aod::FDResultsHFTrkInfo> fillFemtoResultTrkInfo;
+  Produces<o2::aod::FDHfCharm> fillFemtoResultCharm;
+  Produces<o2::aod::FDHfTrk> fillFemtoResultTrk;
+  Produces<o2::aod::FDHfColl> fillFemtoResultColl;
```
vkucera, line 194: Don't call these `fill...`. It suggests that it's a function which is not the case. These are cursors to table rows and in HF they are usually called `row...`.

victor-gonzalez (approved):

alibuild (approved):

PR #11198, opened 2025-05-14
PWGJE: Adding 3D inclusive jet and track histograms

dajones2:
For analysing the inclusive jet and track distributions as a function of phi and eta in pT differential bins - useful for MC QA


dajones2:
Can this be merged please

nzardosh (approved):

PR #11201, opened 2025-05-14
[PWGLF] Add converters in strangeness data model for missing info/tables

romainschotter:
- add RCT info in `straevselsconverter5` to allow the use of RCT conditions for previous derived data (with StraEvSels_004 or below)
- add dummy ZDC neutron table if required in task subscription


alibuild (approved):

PR #11216, opened 2025-05-15
[PWGLF] Function to process generated V0s in sigma0 tasks + minor fixes

gianniliveraro:
This PR includes the following updates:

- New functions for processing generated photons and Lambdas, taking event selection into account. These are intended for use in efficiency calculations and follow the implementation used in derivedlambdakzeroanalysis. A similar approach for Sigma0s is planned for a future update;

- Removal of IR information from the Sigma0 data model;.

- Safeguard added to the analyzeV0CollAssoc function to prevent double-counting of V0s. A check to ensure that only primary particles are considered has also been introduced.


romainschotter (dismissed):

vkucera:
@gianniliveraro @romainschotter How about fixing some of those 375+ errors? :-)

skundu692 (approved):

PR #11218, opened 2025-05-15
[PWGLF] Add processGen function to have MC gen level info

abmodak:


alibuild (approved):
Auto-approving on behalf of @abmodak.

PR #11219, opened 2025-05-16
[Common,PWGEM,PWGHF,PWGLF,DPG,Trigger] Use std::array instead of deprecated gpustd::array

davidrohr:


davidrohr:
I already changed it in O2, gpu::gpustd::array is now just an alias to std::array.
So when we merge this, I would wait few days and then remove the alias in O2.

mpuccio (approved):

ktf:
Linter issues unrelated. Merging this to unblock O2 developments.

PR #11230, opened 2025-05-16
[PWGHF] taskFlow.cxx: change in histograms, change in MFT track selection, new process functions for MC

AlexianL:
What is new in this PR : 
- added new filters on MFT tracks (also, moved the MFT tracks eta cut from a function to a filter -> I tried it before I thought it didn't work but it should work, this will slightly improve memory consumption)
- changed the selectionFlag for HF
- added new ConfigurableAxis to make the difference between plots which requires eta/phi distribution or delta eta/delta phi distribution
- added process functions and all configurables + declarations needed to make it work for MC generated
- added an attempt to create process function for MFT track efficiency (to be extended to D0 and Lc efficiencies, also Tpc tracks ?)
- added differential invariant mass plots vs pT vs Multiplicity

### DIFF
--- a/PWGHF/HFC/Tasks/taskFlow.cxx
+++ b/PWGHF/HFC/Tasks/taskFlow.cxx
@@ -35,10 +35,14 @@
 #include "Framework/StepTHn.h"
 #include "ReconstructionDataFormats/GlobalTrackID.h"
 
+#include "Common/DataModel/CollisionAssociationTables.h"
 #include "Common/DataModel/EventSelection.h"
 #include "Common/DataModel/Multiplicity.h"
 #include "Common/DataModel/TrackSelectionTables.h"
 #include "CommonConstants/MathConstants.h"
+
+#include "PWGMM/Mult/DataModel/Index.h"
+#include "PWGMM/Mult/DataModel/bestCollisionTable.h"
 
 #include "PWGCF/Core/CorrelationContainer.h"
 #include "PWGCF/Core/PairCuts.h"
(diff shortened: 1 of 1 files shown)


PWGHF/HFC/Tasks/taskFlow.cxx
```diff
@@ -1330,33 +1467,113 @@ struct HfTaskFlow {
...
+      case -211:
+        return 0;
+      case 321: // Kaon
+      case -321:
+        return 1;
+      case 2212: // proton
+      case -2212:
+        return 2;
+      default: // NOTE. The efficiency histogram is hardcoded to contain 4 species. Anything special will have the last slot.
+        return 3;
```
vkucera, line 1527: No magic numbers

vkucera (changes requested):

PWGHF/HFC/Tasks/taskFlow.cxx
vkucera: Thanks.
* You don't need the positive and negative cases if you replace `switch (pdgCode)` with `switch (strd::abs(pdgCode))`.
* Your return values are still arbitrary. Map them to enumerated values, either defined here or you can use the `utilsPid` header.

AlexianL: Thanks for the suggestions, I added them in the new commit

PWGHF/HFC/Tasks/taskFlow.cxx
```diff
@@ -1330,33 +1469,110 @@ struct HfTaskFlow {
...
+      case PDG_t::kPiPlus: // positive or negative pion
+        return Pion;
+      case PDG_t::kKPlus: // positive or negative kaon
+        return Kaon;
+      case PDG_t::kProton: // proton or proton bar
+        return Proton;
+      default: // NOTE. The efficiency histogram is hardcoded to contain 4 species. Anything special will have the last slot.
+        return NHfProngSpecies;
+    }
+  }
```
vkucera, line 1528: Don't define helper methods between process functions.
```suggestion
  HfProngSpecies getSpecies(int pdgCode)
  {
    switch (std::abs(pdgCode)) {
      case PDG_t::kPiPlus: // positive or negative pion
        return HfProngSpecies::Pion;
      case PDG_t::kKPlus: // positive or negative kaon
        return HfProngSpecies::Kaon;
      case PDG_t::kProton: // proton or proton bar
        return HfProngSpecies::Proton;
      default: // NOTE. The efficiency histogram is hardcoded to contain 4 species. Anything special will have the last slot.
        return HfProngSpecies::NHfProngSpecies;
    }
  }
```

vkucera (changes requested):

PWGHF/HFC/Tasks/taskFlow.cxx
AlexianL: I added your suggestions and moved the function before the process function with other helper functions

PR #11234, opened 2025-05-17
[PWGDQ] Adding a process function in table-maker which does not use TOF as a …

iarsene:
…dependency; Removing the production of ZDC tables in dq-flow


alibuild (approved):

PR #11236, opened 2025-05-18
#8853 PR version workflow

blacwovie:


maciacco:
@blacwovie before we merge, could you briefly explain why you need to revert to the previous version? In principle you should be able to run with an older tag (unless there are issues I am overlooking)

blacwovie:
> @blacwovie before we merge, could you briefly explain why you need to revert to the previous version? In principle you should be able to run with an older tag (unless there are issues I am overlooking)

I want to make comparison between the old tag and old workflow with current tag.Because I'm not sure what caused changes in my result.(maybe the he3hadron workflow or maybe some changes in O2 frame?)

maciacco:
@blacwovie thanks for your reply. I would suggest that you first try and test this locally to assess where is the problem. Also, looking at your changes, you are mostly changing the naming of variables, which should not affect the outcome of the task. Let me and Francesco know if you manage to run this test.

PR #11241, opened 2025-05-18
[PWGHF] add a flag for vertex build and fix the issue of CPU time consumption 

zhangbiao-phy:
- add a flag for vertex build 
- fix the issue of CPU time consumption by moving the vertex rebuild into the innermost loop


alibuild (approved):

PR #11242, opened 2025-05-19
[PWGHF] HF invariant mass fitter: code cleaning and minor enhancement

lubynets:
**Code cleaning:**
* obsolete headers, variables and pieces of code removed;
* use `TH1` instead of `TH1F` to be more general with possible input histogram type;
* more safe type castings employed (e.g. avoid `reinterpret_cast` and use `Get<TH1>()` instead of `Get()` before `static_cast`);
* instances from the standard library (`string`, `vector`, `cout`, etc.) prepended with `std::`;
* removed **using** directive from header;
* the code is made less D0-hardcoded;
* fixed typos in comments and variable names.

**Enhancement:**
* added signal evaluation via bin counting (in addition to integral);
* added plotting features:
  * semi-opaque pure signal plot
  * background pre-fit function
  * highlight peak region (which is usually 3 sigma) with vertical lines.
  
The json config file is updated accordingly to mentioned changes, a link to CernBox with an example of input file is provided.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -38,7 +38,7 @@
 /PWGEM @alibuild @feisenhu @dsekihat @ivorobye
 /PWGEM/Dilepton @alibuild @mikesas @rbailhac @dsekihat @ivorobye @feisenhu
 /PWGEM/PhotonMeson @alibuild @mikesas @rbailhac @m-c-danisch @novitzky @mhemmer-cern @dsekihat
-/PWGHF @alibuild @vkucera @fcolamar @fgrosa @fcatalan92 @mfaggin @mmazzilli @deepathoms @NicoleBastid @hahassan7 @jpxrk @apalasciano @zhangbiao-phy
+/PWGHF @alibuild @vkucera @fcolamar @fgrosa @fcatalan92 @mfaggin @mmazzilli @deepathoms @NicoleBastid @hahassan7 @jpxrk @apalasciano @zhangbiao-phy @gluparel
 # PWG-LF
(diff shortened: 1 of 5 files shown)


lubynets:
> Thanks a lot @lubynets for the improvements. Can you also please fix the errors?

Hi @vkucera,
I addressed linter issues except of two:
- naming of the "workflow" file is not applicable since `HFInvMassFitter` is not an `O2Physics` workflow;
- printing into terminal via `iostream` instead of `O2` logging allows for compiling the `HFInvMassFitter` standalone, without need to compile the entire `O2`.

vkucera:
> > Thanks a lot @lubynets for the improvements. Can you also please fix the errors?
> 
> Hi @vkucera, I addressed linter issues except of two:
> 
>     * naming of the "workflow" file is not applicable since `HFInvMassFitter` is not an `O2Physics` workflow;
> 
>     * printing into terminal via `iostream` instead of `O2` logging allows for compiling the `HFInvMassFitter` standalone, without need to compile the entire `O2`.

Thanks @lubynets , agreed on the first point. I will fix that in the linter. The printing can be done with `printf` though.

vkucera:
Thanks a lot @lubynets for the nice work.
Last thing. Is there a reason to not take the masses from `o2::constants::physics`?

lubynets:
> Thanks a lot @lubynets for the nice work. Last thing. Is there a reason to not take the masses from `o2::constants::physics`?

Hi @vkucera, my idea was to keep fitter class independent on the full O2 framework.

vkucera (approved):

PR #11243, opened 2025-05-19
[PWGUD] Removed 2 process functions and fixed o2-linter issues

ananthapadmanabhan18:
Here are few changes that i made:
- Removed process functions (data and MC reco) which used only the filtered tables, instead i used filtering on the main functions
- Changed the `TLorentzVector` to recommended template.
- Added few more configurables for vertex cuts and number of PV Contributors. 
- Renamed few variables
- Only single truegapside=2 events selected

### DIFF
--- a/PWGUD/Tasks/exclusiveRhoTo4Pi.cxx
+++ b/PWGUD/Tasks/exclusiveRhoTo4Pi.cxx
@@ -493,8 +493,9 @@ DECLARE_SOA_TABLE(SignalMCreco, "AOD", "SignalMCreco",
 struct ExclusiveRhoTo4Pi {
   SGSelector sgSelector;
   int rhoPrime = 30113;
-  int four = 4;
-  int two = 2;
+  int numFourPionTracks = 4;
+  int numPiPlus = 2;
+  int numPiMinus = 2;
   float zeroPointEight = 0.8;
   Produces<aod::SignalData> sigFromData;
   Produces<aod::BkgroundData> bkgFromData;
@@ -553,7 +554,6 @@ struct ExclusiveRhoTo4Pi {
     histosData.add("tpcNClsFindable", "TPC N Cls Findable; N Cls Findable; Counts", kTH1F, {{200, 0, 200}});
 
     // TPC nSigma
-    // histosData.add("tpcNSigmaPi_WOTS", "TPC nSigma Pion without track selection; Events", kTH2F, {{1000, -15, 15}, {nBinsPt, 0, 10}});
     histosData.add("tpcNSigmaPi_WTS", "TPC nSigma Pion with track selection; Events", kTH2F, {{1000, -15, 15}, {nBinsPt, 0, 10}});
     histosData.add("tpcNSigmaPi_WTS_PID_Pi", "TPC nSigma Pion with track selection and PID Selection of Pi; Entries", kTH2F, {{1000, -15, 15}, {nBinsPt, 0, 10}});
 
@@ -565,7 +565,6 @@ struct ExclusiveRhoTo4Pi {
 
     // TOF nSigma
     histosData.add("tofNSigmaPi_WTS", "TOF nSigma Pion with track selection; Events", kTH2F, {{1000, -15, 15}, {nBinsPt, 0, 10}});
-    // histosData.add("tofNSigmaPi_WOTS", "TOF nSigma Pion without track selection; Events", kTH2F, {{1000, -15, 15}, {nBinsPt, 0, 10}});
     histosData.add("tofNSigmaPi_WTS_PID_Pi", "TOF nSigma Pion with track selection and PID Selection of Pi; Entries", kTH2F, {{1000, -15, 15}, {nBinsPt, 0, 10}});
 
     // TOF nSigma of other particles with selected pion tracks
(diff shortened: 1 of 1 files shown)


PWGUD/Tasks/exclusiveRhoTo4Pi.cxx
```diff
@@ -40,6 +39,8 @@ using namespace o2;
 using namespace o2::aod;
 using namespace o2::framework;
 using namespace o2::framework::expressions;
+typedef ROOT::Math::LorentzVector<ROOT::Math::PxPyPzM4D<double>> LorentzVecM;
```
rolavick, line 42: This does not seem to be placed correctly. Can you prove this will never affect others' code when someone else would accidentally use 'LorentzVecM' for the name of a local variable, too? Can't you just move this inside struct? The same applies to the next line, the definition of 'LorentzVecE'.

rolavick (changes requested):

PWGUD/Tasks/exclusiveRhoTo4Pi.cxx
vkucera: It's also completely useless. There is already `PxPyPzMVector` available directly from ROOT.

PWGUD/Tasks/exclusiveRhoTo4Pi.cxx
```diff
@@ -847,140 +841,102 @@ struct ExclusiveRhoTo4Pi {
...
+      histosData.fill(HIST("tofNSigmaKa_WTS_PID_Pi"), selectedPionTracks[i].tofNSigmaKa(), selectedPionTrackVector.Pt());
+      histosData.fill(HIST("tofNSigmaPr_WTS_PID_Pi"), selectedPionTracks[i].tofNSigmaPr(), selectedPionTrackVector.Pt());
+      histosData.fill(HIST("tofNSigmaEl_WTS_PID_Pi"), selectedPionTracks[i].tofNSigmaEl(), selectedPionTrackVector.Pt());
+      histosData.fill(HIST("tofNSigmaMu_WTS_PID_Pi"), selectedPionTracks[i].tofNSigmaMu(), selectedPionTrackVector.Pt());
+      histosData.fill(HIST("pT_track_WTS_PID_Pi"), selectedPionTrackVector.Pt());
+      histosData.fill(HIST("rapidity_track_WTS_PID_Pi"), selectedPionTrackVector.Rapidity());
     } // End of loop over tracks with selection and PID selection of Pions
 
-    if (numTracksWTSandPIDpi != 4) {
+    if (numSelectedPionTracks != four) {
```
vkucera, line 927: Seriously?

PWGUD/Tasks/exclusiveRhoTo4Pi.cxx
```diff
@@ -847,140 +841,102 @@ struct ExclusiveRhoTo4Pi {
...
       return;
     }
 
     // Selecting Events with net charge = 0
-    if (numPionMinusTRacks == 2 && numPiPlusTracks == 2) {
+    if (numPionMinusTracks == two && numPiPlusTracks == two) {
```
vkucera, line 932: Seriously?

PWGUD/Tasks/exclusiveRhoTo4Pi.cxx
```diff
@@ -494,7 +495,9 @@ DECLARE_SOA_TABLE(SignalMCreco, "AOD", "SignalMCreco",
 struct ExclusiveRhoTo4Pi {
   SGSelector sgSelector;
   int rhoPrime = 30113;
-  uint16_t numPVContrib = 4;
+  int four = 4;
+  int two = 2;
```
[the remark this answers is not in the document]
rolavick: I would not say it is meaningless. It has the meaning of what numbers usually mean, some value. Nevertheless, in the context of L927 and L932, the variable naming clearly gives the meaning to the number it is compared to: the number of tracks is/is not 4/2. Hence, I see the MagicNumbers errors as false positives. @vkucera  What solution would you suggest? Should we suppress the error with the extra comment line?

PWGUD/Tasks/exclusiveRhoTo4Pi.cxx
ananthapadmanabhan18: Sure @vkucera I didn't know that. I will add it. I added it because it was recommended in the o2-linter log.

PWGUD/Tasks/exclusiveRhoTo4Pi.cxx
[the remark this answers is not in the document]
ananthapadmanabhan18: I find it difficult to review the code if there are only variables (it would get difficult to cross check it)  : ) 
that's y i gave it as two and four. I can change it to something like numTracks, numPiPlusTracks etc if required.
Out of curiosity, may I ask you why this check exist? Does it improve performance or something :) idk...I am not an expert in c++.

PWGUD/Tasks/exclusiveRhoTo4Pi.cxx
vkucera: The linter did not recommend you to define `typedef ROOT::Math::LorentzVector<ROOT::Math::PxPyPzM4D<double>>`. You are supposed to read the documentation of the class and use it correctly.

PWGUD/Tasks/exclusiveRhoTo4Pi.cxx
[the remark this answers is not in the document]
vkucera: You are comparing a variable to another value. To make the comparison understandable, both sides have to be understandable. If you compare a variable to 4, it does not tell anything about the meaning of the value. Why is it 4 and not 5?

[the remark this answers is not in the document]
vkucera: @ananthapadmanabhan18 References for the rationale of each test are in the log of the linter.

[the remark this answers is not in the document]
ananthapadmanabhan18: Ah I see, I will change it. No issues. Thanks @vkucera

rolavick (approved):

PR #11249, opened 2025-05-19
[PWGCF] adding RCT evsel flag + ITS param. for PID in MC

glromane:
@njacazio @ercolessi


alibuild (approved):

PR #11252, opened 2025-05-19
[PWGCF] DptDpt - First step in implementing linter recommendations

victor-gonzalez:
It will take a few steps because we wanted to keep track of the changes 
This first step is just source file renaming when needed


alibuild (approved):

PR #11266, opened 2025-05-20
[PWGCF] fixed incorrect index and O2 linter warnings

Eloviyo:


PWGCF/TableProducer/filter2Prong.cxx
```diff
@@ -106,10 +106,10 @@ struct Filter2Prong {
                            prongCFId[0], prongCFId[1], c.pt(), c.eta(), c.phi(), hfHelper.invMassD0ToPiK(c), aod::cf2prongtrack::D0ToPiK);
         if constexpr (std::experimental::is_detected<HasMLProb, typename HFCandidatesType::iterator>::value) {
           mlvecd.clear();
-          for (float val : c.mlProbD0())
+          for (const float val : c.mlProbD0())
```
victor-gonzalez, line 109: The `linter` is saying that the `const float` is being used but not a constant reference to `float`, aka `const float&`, which is what is being expected

victor-gonzalez (approved):
Please, have a look at my comment in case it helps

PR #11268, opened 2025-05-20
[PWGLF] Add TPC PID Vs mass correlation cuts + other fix in the strangenessbuilder

romainschotter:
-  Add TPC PID Vs Mass correlation cuts in the strangeness builder 
-  Capture spurious cascade with TPC-only-V0 (thanks @gianniliveraro for catching this)

@ddobrigk for your information


alibuild (approved):

PR #11286, opened 2025-05-21
Fix vector error in candidateSelectorOmegac0ToOmegaPi.cxx

FDUEnrich:


FDUEnrich:
> Can you please explain the error that is fixed by this PR?

Hi Vit. The reason is that if we don't apply ML, std::vector<float> outputMlOmegac would be an empty vector. If I fill outputMlOmegac[0] with other variables(such as Invmass). This will result in out-of-bounds access on a null vector.

vkucera:
> > Can you please explain the error that is fixed by this PR?
> 
> Hi Vit. The reason is that if we don't apply ML, std::vector outputMlOmegac would be an empty vector. If I fill outputMlOmegac[0] with other variables(such as Invmass). This will result in out-of-bounds access on a null vector.

1. It is not clear to me why you need to store `outputMlOmegac[0]` in the `HfSelToOmegaPi` table when the full `outputMlOmegac` vector is already stored in the `HfMlSelOmegacToOmegaPi` table.
2. If you really need to store it, you can simply replace `outputMlOmegac[0]` with `outputMlOmegac.size() > 0 ? outputMlOmegac[0] : -1.f`.

fgrosa:
Hi @FDUEnrich do the last commits include everything you wanted to implement, or will you still add the ML table?

FDUEnrich:
> Hi @FDUEnrich do the last commits include everything you wanted to implement, or will you still add the ML table?
Hi Fabrizio, yes. I'm still testing.  I think I can fix this bug first. And add centrality and ML-joined table next time. When the test done I will open Ready for review. Tks

PR #11299, opened 2025-05-21
[PWGCF] DptDpt - Implementing `linter` recommendations

victor-gonzalez:


victor-gonzalez:
@vkucera Interesting that the `[cppcheck]` has not found errors in the `PROCESS_SWITCH` macros that are above in the same file. Perhaps the fact of being associated to methods that are actually defined outside of the `struct`?

vkucera:
> @vkucera Interesting that the `[cppcheck]` has not found errors in the `PROCESS_SWITCH` macros that are above in the same file. Perhaps the fact of being associated to methods that are actually defined outside of the `struct`?

Hi @victor-gonzalez , I admit it is not clear to me yet how cppcheck detects undefined macros. Another area where it seems to give false positives are declarations of O2 filters. I still have to look into the details of the reported problems.

alibuild (approved):

PR #11301, opened 2025-05-22
[PWGEM/Dilepton] update treeCreatorElectronMLDDA.cxx

dsekihat:
PWGEM/Dilepton: update treeCreatorElectronMLDDA.cxx


vkucera:
Hi @dsekihat , I'm not sure if that is what cppcheck complains about but if the casting to `uint8_t` in the filter is really needed, it should not be done with the C-style casting `(uint8_t)` but with `static_cast<uint8_t>`.

PR #11311, opened 2025-05-22
[DPG] Fix bug found by cppcheck

vkucera:
- `Shifting 32-bit value by 63 bits is undefined behaviour [shiftTooManyBits]`


njacazio (approved):
Thanks Vit!

PR #11325, opened 2025-05-23
[PWGJE] additional EMCal time calibration (fix)

fjonasALICE:
For some specific runs, a problem in the time calibration objects was observed, which leads to a shift of -2*8.8ns for low gain cells. The affected run ranges were identified by @jokonig , and we now apply a temporary shift according to the run ranges in the correction framework. Once new reconstruction passes are produced for the affected runs, the additional shift will be removed!


fjonasALICE:
@jokonig please double check the run ranges. A second set of eyes can't hurt I think.

jokonig:
@fjonasALICE looks good to me!

mhemmer-cern (approved):

PR #11327, opened 2025-05-23
Isolated lumi calculation in a dedicated task

ekryshen:
Main changes:
* lumi calculation moved from bc-selection-task to a dedicated lumi task
* introduced pileup correction using rates from CTP scalers
* introduced 2D lumi tables for various RCT selection
* Changed default ccdb path for RCT to RCT/Flags/RunFlags
* setting bit 31 in rct flags to 1 if rct map is not found in ccdb

### DIFF
--- a/Common/TableProducer/eventSelection.cxx
+++ b/Common/TableProducer/eventSelection.cxx
@@ -37,6 +37,8 @@
 #include "DataFormatsITSMFT/NoiseMap.h" // missing include in TimeDeadMap.h
 #include "DataFormatsITSMFT/TimeDeadMap.h"
 #include "ITSMFTReconstruction/ChipMappingITS.h"
+#include "DataFormatsCTP/Configuration.h"
+#include "DataFormatsCTP/Scalers.h"
 
 #include "TH1D.h"
 
@@ -80,7 +82,6 @@ struct BcSelectionTask {
   int mITSROFrameEndBorderMargin = 20;   // default value
   int mTimeFrameStartBorderMargin = 300; // default value
   int mTimeFrameEndBorderMargin = 4000;  // default value
-  bool isPP = 1;                         // default value
   TriggerAliases* aliases = nullptr;
   EventSelectionParams* par = nullptr;
   std::map<uint64_t, uint32_t>* mapRCT = nullptr;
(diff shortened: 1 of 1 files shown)


Common/TableProducer/eventSelection.cxx
```diff
@@ -1229,12 +1160,305 @@ struct EventSelectionTask {
...
+      histos.get<TH2>(HIST("hLumiZEMafterBCcutsRCT"))->GetYaxis()->SetBinLabel(4 * i + 3, (rctListName + "_zdc").Data());
+      histos.get<TH2>(HIST("hLumiZNCafterBCcutsRCT"))->GetYaxis()->SetBinLabel(4 * i + 3, (rctListName + "_zdc").Data());
+      histos.get<TH2>(HIST("hLumiTVXafterBCcutsRCT"))->GetYaxis()->SetBinLabel(4 * i + 4, (rctListName + "_zdc" + "_fullacc").Data());
+      histos.get<TH2>(HIST("hLumiTCEafterBCcutsRCT"))->GetYaxis()->SetBinLabel(4 * i + 4, (rctListName + "_zdc" + "_fullacc").Data());
+      histos.get<TH2>(HIST("hLumiZEMafterBCcutsRCT"))->GetYaxis()->SetBinLabel(4 * i + 4, (rctListName + "_zdc" + "_fullacc").Data());
+      histos.get<TH2>(HIST("hLumiZNCafterBCcutsRCT"))->GetYaxis()->SetBinLabel(4 * i + 4, (rctListName + "_zdc" + "_fullacc").Data());
+    }
+  }
+
+  void process(BCsWithBcSelsRun3 const& bcs, aod::FT0s const&)
```
victor-gonzalez, line 1235: @ekryshen I believe this is going to fail when running on converted Run1/2 data. What do you think?

ekryshen: @ddobrigk Should be fixed now (added processRun3)

victor-gonzalez:
@ekryshen Thanks!
But shouldn't it have a default when running on converted data? 
My guess is that the workflow will hang in such a case but perhaps that is an old behavior. Sorry for the noise if that is the case

ekryshen:
@victor-gonzalez Indeed, the workflow hangs without dummy process function for Run2. I've just committed another fix.

victor-gonzalez:
> @victor-gonzalez Indeed, the workflow hangs without dummy process function for Run2. I've just committed another fix.

Thanks @ekryshn!!!!
Fine from my side!!!

ddobrigk:
Thanks @ekryshen, I will approve and merge now! Since you added an extra struct, this PR may mean an increase in memory. I am not sure how relevant this is, but let's just go ahead so that we get the added features immediately and we can then evaluate the memory changes in a tag including this PR.

ddobrigk (approved):

lubynets:
Dear @ekryshen,
After you added LumiTask, by default arguments `lumi-task` `processRun2` and `processRun3` are switched to **false** (this is how they appear in the newly produced dpl-config.json file). In this case the job is stuck without any error message.
Is it worth adding an error Log and termination of the task if both arguments are **false**?

ekryshen:
> Dear @ekryshen, After you added LumiTask, by default arguments `lumi-task` `processRun2` and `processRun3` are switched to **false** (this is how they appear in the newly produced dpl-config.json file). In this case the job is stuck without any error message. Is it worth adding an error Log and termination of the task if both arguments are **false**?

@lubynets Thanks for the suggestion, I will implement it in the next iteration. I wonder if this issue (no process functions) can be catched by the central framework. @aalkin, what do you think?

aalkin:
@lubynets The task inputs are determined by enabled processes, when both are disabled in this case, the input list is cleared. If you run with the DebugGUI you'll notice the distinct lack of channels leading to the task as it has no inputs. Due to this, task's processing callback is not triggered and it is stuck. There are examples of tasks that have no process functions at all and are functioning, because they have `Spawns` or `Builds` declarations in them, that also add to input list.

PR #11372, opened 2025-05-26
[PWGEM] phosElId dispersion calculation fix

yhambard:
globalized the dispersion checker
code retains the ability to calculate in the original/wrong way, just in case
removed upper limit for phos energies while reconstructing pi0


mhemmer-cern:
Will merged since it got approved.

PR #11374, opened 2025-05-27
Add non-prompt info for MC gen loop

fmazzasc:


alibuild (approved):

PR #11417, opened 2025-05-30
[PWGHF] taskXic: Remove not needed PID from track tables

cterrevo:
Updated removing the TracksWPid, which joined TrackWDca, with not needed PID tables.  I also deleted a non used partition.

### DIFF
--- a/PWGHF/D2H/Tasks/taskXic.cxx
+++ b/PWGHF/D2H/Tasks/taskXic.cxx
@@ -67,7 +67,6 @@ struct HfTaskXic {
   float etaMaxAcceptance = 0.8;
   float ptMinAcceptance = 0.1;
 
-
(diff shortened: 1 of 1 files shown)


vkucera (changes requested):
Hi @cterrevo , please fix those few errors.

vkucera:
Dear @cterrevo , please see my replies below.

> Are these "errors" really necessary to fix?

It would be better fix them now. Otherwise they will appear again in the next PR that might be opened by someone else.

> I do not understand the types of error with "appears too early" (with respect to what?).

The message "B appears too early (before end of A)." means that B should appear after A.

> Do you have a link to a documentation I could follow?

Yes, the first error message says: "See the PWGHF coding guidelines." and the link to the guidelines is in the output of the linter check.

PR #11382, opened 2025-05-27
[PWGHF,PWGJE] Unify and extend MC flags in candidate tables (3-prong, D*+)

vkucera:
Add support for flagging all needed decay channels.

Background:

- MC-based corrections require knowledge about other decay channels than the ones expected in the candidate reconstruction and selection.
- Decay reconstruction flags used in the skim creator and decay matching flags used in the candidate creators are conceptually different. The former flags tag all decay hypotheses that satisfy selection conditions, the latter flags tag the actual decay channels.

Concept considerations:
- Separate decay reconstruction flags and decay matching flags.
- Enumerate all decay channels that need to be flagged for a given candidate table.
- All MC matching is be done in the candidate creators.
- Flags of main channels are stored in the existing columns: `FlagMcMatchRec`, `FlagMcMatchGen`.
- Flags of resonant channels are stored in the existing columns: `FlagMcDecayChanRec`, `FlagMcDecayChanGen`. (To be renamed.)
- Flags are stored as signed enumerator values (type `int8_t`), not as bitmaps.
  - `0` indicates no match.
  - Sign indicates particle/antiparticle.
 
Caveat:
- Invalidation of flags in existing derived data.

TODO: Extend to other species.

PR #11418, opened 2025-05-30
[PWGHF] Implement gen and reco MC matching for correlated bkg decays

Marcellocosti:
In this PR, the matching for the study of the correlated background decays implemented in the [linked](https://github.com/AliceO2Group/O2DPG/blob/2794ba5b82b80663b58471c7125f322d680c5218/MC/config/PWGHF/pythia8/generator/pythia8_charmhadronic_with_decays_Mode2_CorrBkg.cfg#L4) configuration script is implemented for 2-prong and 3-prong candidates both at reco and at gen level.

By now I keep the PR in draft mode as there are several debug prints for the checks on the matching I am performing and because I will need to digest the rearrangement of the MC flags proposed in the PR #11382 when merged. Nevertheless, a feedback on the general structure is more than welcome. Tagging @stefanopolitano for info.

The modifications to the D0 and D+ trees are for testing purposes.


PWGHF/TableProducer/candidateCreator3Prong.cxx
```diff
@@ -941,146 +945,269 @@ struct HfCandidateCreator3ProngExpressions {
...
+                  // LOG(info) << "CIAO 3.0.4";
+                  std::array<int, 5> finalStateParts3ProngAll = std::array{finalState[0], finalState[1], finalState[2], finalState[3], finalState[4]};
+                  if (sign < 0) {
+                    for (auto& part : finalStateParts3ProngAll) {
+                      if (part == kPi0) {
+                        part = -part; // The Pi0 pdg code does not change between particle and antiparticle
+                      }
+                    }
+                    // LOG(info) << "CIAO 3.0.5";
+                  }
```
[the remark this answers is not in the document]
mfaggin: Here I do not understand, sorry: why are you swapping the sign of the pi0 PDG code? It should be always 111 for pi0 and anti-pi0, right?

PWGHF/TableProducer/candidateCreator3Prong.cxx
```diff
@@ -941,146 +945,269 @@ struct HfCandidateCreator3ProngExpressions {
...
+                  std::array<int, 4> finalStateParts3ProngAll = std::array{finalState[0], finalState[1], finalState[2], finalState[3]};
+                  // std::array<int, 4> finalStateParts3ProngAll = std::array{finalState[0], finalState[1], finalState[2], sign*finalState[3]};
+                  if (sign < 0) {
+                    for (auto& part : finalStateParts3ProngAll) {
+                      if (part == kPi0) {
+                        part = -part; // The Pi0 pdg code does not change between particle and antiparticle
+                      }
+                    }
+                  }
+                  if (!RecoDecay::isMatchedMCGen(mcParticles, motherParticle, pdg, finalStateParts3ProngAll, false, &sign, depth)) {
```
mfaggin, line 992:
```suggestion
                  if (!RecoDecay::isMatchedMCGen(mcParticles, motherParticle, pdg, finalStateParts3ProngAll, true, &sign, depth)) {
```
What is the reason to put `acceptAntiParticle` parameter to `false`? I think that we can put it to `true` as we usually do for the MC matching of signal channels, right?

PWGHF/TableProducer/candidateCreator3Prong.cxx
```diff
@@ -941,146 +945,269 @@ struct HfCandidateCreator3ProngExpressions {
...
+                if (finalState.size() == 4) { // Check if the final state has 4 particles
+                  std::array<int, 4> finalStateParts3ProngAll = std::array{finalState[0], finalState[1], finalState[2], finalState[3]};
+                  // std::array<int, 4> finalStateParts3ProngAll = std::array{finalState[0], finalState[1], finalState[2], sign*finalState[3]};
+                  if (sign < 0) {
+                    for (auto& part : finalStateParts3ProngAll) {
+                      if (part == kPi0) {
+                        part = -part; // The Pi0 pdg code does not change between particle and antiparticle
+                      }
+                    }
+                  }
```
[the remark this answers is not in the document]
Marcellocosti: Since we have a case with 2 pi0s, the `break` statement would avoid a correct matching for as only one pion pdg code would be changed

PWGHF/TableProducer/candidateCreator3Prong.cxx
Marcellocosti: Thanks for noticing!

Marcellocosti:
Thanks @mfaggin for your comments, I have implemented them and I marked the PR as ready for review. Please let me know if you have further suggestions!

PWGHF/TableProducer/candidateCreator3Prong.cxx
```diff
@@ -941,146 +945,269 @@ struct HfCandidateCreator3ProngExpressions {
...
+      // LOG(info) << "CIAO REC MATCHING";
+      if (matchCorrBkgs) {
+        // LOG(info) << "CIAO 1";
+        // LOG(info) << "--------------------------------------------";
+        // LOG(info) << "Matching correlated bkgs";
+        std::array<int, 5> mothersPdgCodes = {Pdg::kDPlus, Pdg::kDS, Pdg::kDStar, Pdg::kLambdaCPlus, Pdg::kXiCPlus};
+        indexRec = -1; // Index of the matched reconstructed candidate
+        
+        // LOG(info) << "CIAO 2";
+        for (const auto& pdg : mothersPdgCodes) {
```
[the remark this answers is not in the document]
Marcellocosti: Good point. After talking with @fgrosa, I decided to implement the matching for all channels by default. The presence of `break` statements should reduce the CPU time, but indeed your proposal would save more resources so I can also implement this solution if you have a preference for it.

vkucera:
Converted to draft since this will take a few iterations and there is no point to build every single one of them.

PWGHF/Core/DecayChannels.h
```diff
@@ -53,65 +59,193 @@ enum DecayChannelResonant : int8_t {
...
+    {DecayChannelMain::D0ToPiK, std::vector<int>{+kKMinus, +kPiPlus}},
+    {DecayChannelMain::D0ToPiKPi0, std::vector<int>{+kKMinus, +kPiPlus, +kPi0}},
+    {DecayChannelMain::D0ToPiPi, std::vector<int>{+kPiMinus, +kPiPlus}},
+    {DecayChannelMain::D0ToPiPiPi0, std::vector<int>{+kPiMinus, +kPiPlus, +kPi0}},
+    {DecayChannelMain::D0ToKK, std::vector<int>{+kKMinus, +kKPlus}},
+};
+
+std::unordered_map<DecayChannelResonant, std::array<int, 2>> resoStatesD0 =
+  {
+    {DecayChannelResonant::D0ToRhoplusPi, std::array<int, 2>{213, +kPiMinus}},
```
vkucera, line 74: Avoid hard-coded PDG codes.

PWGHF/Core/DecayChannels.h
```diff
@@ -53,65 +59,193 @@ enum DecayChannelResonant : int8_t {
   //
   LastChannelResonant
 };
+
+std::unordered_map<DecayChannelMain, std::vector<int>> finalStates2Prongs =
```
vkucera, line 63:
```suggestion
std::unordered_map<DecayChannelMain, const std::vector<int>> finalStates2Prongs =
```

PWGHF/Core/DecayChannels.h
```diff
@@ -53,65 +59,193 @@ enum DecayChannelResonant : int8_t {
...
 enum DecayChannelMain : int8_t {
   // D+
-  DplusToPiKPi = 1, // π+ K− π+
-  DplusToPiKPiPi0,  // π+ K− π+ π0
-  DplusToPiPiPi,    // π+ π− π+
-  DplusToPiKK,      // π+ K− K+
+  DplusToPiKPi = 1, // 1, π+ K− π+
+  DplusToPiKPiPi0,  // 2, π+ K− π+ π0
+  DplusToPiPiPi,    // 3, π+ π− π+
+  DplusToPiKK,      // 4, π+ K− K+
```
vkucera, line 90: Remove the numbers. They are irrelevant.

PWGHF/Core/DecayChannels.h
```diff
@@ -177,4 +311,108 @@ enum DecayChannelResonant : int8_t {
...
+/// \tparam N size of the array of daughter PDG codes
+/// \param arrPdgResoChn array of daughter indices
+/// \param arrPdgDaugs array of PDG codes for the resonant decay
+/// \return true if the resonant channel is matched, false otherwise
+template <std::size_t N>
+bool checkResonantDecay(std::array<int, N> arrPdgResoChn, std::array<int, N> arrPdgDaugs)
+{
+  for (size_t i = 0; i < N; i++) {
+    bool findDaug = false;
+    for (size_t j = 0; j < N; j++) {
```
vkucera, line 372:
```suggestion
  for (std::size_t i = 0; i < N; i++) {
    bool findDaug = false;
    for (std::size_t j = 0; j < N; j++) {
```

PWGHF/Core/DecayChannels.h
```diff
@@ -177,4 +311,108 @@ enum DecayChannelResonant : int8_t {
...
+  }
+}
+
+/// Perform the matching for a single resonant channel
+/// \tparam N size of the array of daughter PDG codes
+/// \param arrPdgResoChn array of daughter indices
+/// \param arrPdgDaugs array of PDG codes for the resonant decay
+/// \return true if the resonant channel is matched, false otherwise
+template <std::size_t N>
+bool checkResonantDecay(std::array<int, N> arrPdgResoChn, std::array<int, N> arrPdgDaugs)
```
vkucera, line 368:
```suggestion
bool checkResonantDecay(std::array<int, N> const& arrPdgResoChn, std::array<int, N> const& arrPdgDaugs)
```

PWGHF/Core/DecayChannels.h
```diff
@@ -177,4 +311,108 @@ enum DecayChannelResonant : int8_t {
...
+namespace o2::hf_corrbkg
+{
+using namespace o2::hf_decay;
+
+/// Returns a map of the possible final states for a specific 3-prong particle specie
+/// \param pdgMother PDG code of the mother particle
+/// \return a map of final states with their corresponding PDG codes
+std::unordered_map<hf_cand_3prong::DecayChannelMain, std::vector<int>> getDecayChannel3Prong(int pdgMother)
+{
+  switch (pdgMother) {
```
vkucera, line 323: These utility functions should go into a separate utility header, e.g. `Utils/utilsMcMatching.h` together with the definitions of the maps which add constructors in every include. Plus, the matching utilities are only needed in the candidate creators and the McGen utils. They are not needed in any other workflows which only need to compare the flag columns with the enums.

PWGHF/TableProducer/candidateCreator3Prong.cxx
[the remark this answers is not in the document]
mfaggin: Personally I'd prefer indeed the solution of putting it as a `Configurable`. Despite the `break`, as it is now, the code calls many times the matching functions and this might be not optimal for the CPU time. You can leave the 5 cases that you put as default, then it's left to the analysers the possibility to reduce the amount of channels to be treated for the correlated background.

PWGHF/TableProducer/candidateCreator3Prong.cxx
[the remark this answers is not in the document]
Marcellocosti: I think the issues arises [here](https://github.com/AliceO2Group/O2Physics/blob/64b985131da0d1eb99bc6c2a42fdaff5e86f2a0d/Common/Core/RecoDecay.h#L828) for the reco matching and [here](https://github.com/AliceO2Group/O2Physics/blob/64b985131da0d1eb99bc6c2a42fdaff5e86f2a0d/Common/Core/RecoDecay.h#L961) for the generated matching. In both cases you can see that one side is multiplied by `sign` but the pdg code of the pi0 does not change between particles and antiparticles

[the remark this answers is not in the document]
mfaggin: ok, indeed that should be the reason. Many thanks @Marcellocosti

PWGHF/Core/DecayChannels.h
Marcellocosti: Dear Vit, I was able to change the hard coded values for the rho meson, but some pdg codes that are needed are not present neither in ROOT `TPDGCode.h` nor in O2, i.e.: Lambda(1520), eta meson, K(1430), f2(1270) and Delta++.

PWGHF/Core/DecayChannels.h
Marcellocosti: Here I marked as `const&` only the first argument, as I need to update the entries of the second one to `-1` when dealing with channels having more than one daughter for a given particle specie.

PWGHF/TableProducer/candidateCreator3Prong.cxx
```diff
@@ -850,6 +852,8 @@ struct HfCandidateCreator3ProngExpressions {
   Configurable<bool> rejectBackground{"rejectBackground", true, "Reject particles from background events"};
   Configurable<bool> matchKinkedDecayTopology{"matchKinkedDecayTopology", false, "Match also candidates with tracks that decay with kinked topology"};
   Configurable<bool> matchInteractionsWithMaterial{"matchInteractionsWithMaterial", false, "Match also candidates with tracks that interact with material"};
+  Configurable<bool> matchCorrBkgs{"matchCorrBkgs", false, "Match correlated background candidates"};
+  Configurable<std::vector<int>> mothersCorrBkgsPdgs{"mothersCorrBkgsPdgs", {411, 413, 431, 4122, 4232}, "PDG codes of the mother particles of correlated background candidates"};
```
mfaggin, line 856:
```suggestion
  Configurable<std::vector<int>> mothersCorrBkgsPdgs{"mothersCorrBkgsPdgs", {Pdg::kDPlus, Pdg::kDS, Pdg::kDStar, Pdg::kLambdaCPlus, Pdg::kXiCPlus}, "PDG codes of the mother particles of correlated background candidates"};
```

mfaggin:
> Dear @vkucera, @mfaggin, thanks for the inputs. I took care of your comments, implementing them or replying to them if I couldn't do so. Please let me know if you have further suggestions!

many thanks @Marcellocosti , aparta from a very minor last comment, I'm fine with your implementations and you can go ahead

PWGHF/Utils/utilsMcMatching.h
```diff
@@ -0,0 +1,253 @@
...
+#define PWGHF_UTILS_UTILSMCMATCHING_H_
+
+#include <array>
+#include <cstdint>
+#include <unordered_map>
+#include <vector>
+
+#include "PWGHF/Core/DecayChannels.h"
+
+namespace o2::hf_corrbkg
```
vkucera, line 26: This is misleading. The channels are not only for correlated background.

PWGHF/Utils/utilsMcMatching.h
```diff
@@ -0,0 +1,253 @@
...
+{
+using namespace o2::hf_decay;
+using namespace o2::constants::physics;
+
+namespace hf_chns_2prong
+{
+
+using namespace o2::hf_decay::hf_cand_2prong;
```
vkucera, line 34: Why don't you just extend `o2::hf_decay::hf_cand_2prong`?

PWGHF/TableProducer/candidateCreator2Prong.cxx
```diff
@@ -762,47 +770,111 @@ struct HfCandidateCreator2ProngExpressions {
...
+          indexRec = RecoDecay::getMatchedMCRec<false, false, false, true, false>(mcParticles, arrayDaughters, Pdg::kD0, std::array{+kPiPlus, -kKPlus}, true, &sign, 1, &nKinkedTracks);
+        } else if (!matchKinkedDecayTopology && matchInteractionsWithMaterial) {
+          indexRec = RecoDecay::getMatchedMCRec<false, false, false, false, true>(mcParticles, arrayDaughters, Pdg::kD0, std::array{+kPiPlus, -kKPlus}, true, &sign, 1, nullptr, &nInteractionsWithMaterial);
         } else {
-          indexRec = RecoDecay::getMatchedMCRec(mcParticles, arrayDaughters, Pdg::kJPsi, std::array{+kElectron, -kElectron}, true);
+          indexRec = RecoDecay::getMatchedMCRec(mcParticles, arrayDaughters, Pdg::kD0, std::array{+kPiPlus, -kKPlus}, true, &sign);
         }
         if (indexRec > -1) {
-          flag = 1 << DecayType::JpsiToEE;
+          flag = sign * (1 << DecayType::D0ToPiK);
```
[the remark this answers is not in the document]
vkucera: I don't understand. The 3-prongs are already using the new scheme everywhere. How is using the old scheme for 2-prongs consistent with that?

PWGHF/Utils/utilsMcMatching.h
Marcellocosti: Thanks, it is a better solution indeed

PWGHF/Utils/utilsMcMatching.h
```diff
@@ -0,0 +1,250 @@
...
+
+#include "PWGHF/Core/DecayChannels.h"
+
+namespace o2::hf_decay
+{
+
+namespace hf_cand_2prong
+{
+
+std::unordered_map<DecayChannelMain, const std::vector<int>> finalStates2Prongs =
```
vkucera, line 32:
```suggestion
std::unordered_map<DecayChannelMain, const std::vector<int>> DaughtersD0Main =
```

PWGHF/Utils/utilsMcMatching.h
```diff
@@ -0,0 +1,250 @@
...
+  {
+    {DecayChannelMain::D0ToPiK, {+kKMinus, +kPiPlus}},
+    {DecayChannelMain::D0ToPiKPi0, {+kKMinus, +kPiPlus, +kPi0}},
+    {DecayChannelMain::D0ToPiPi, {+kPiMinus, +kPiPlus}},
+    {DecayChannelMain::D0ToPiPiPi0, {+kPiMinus, +kPiPlus, +kPi0}},
+    {DecayChannelMain::D0ToKK, {+kKMinus, +kKPlus}},
+};
+
+std::unordered_map<DecayChannelResonant, const std::array<int, 2>> resoStatesD0 =
```
vkucera, line 41:
```suggestion
std::unordered_map<DecayChannelResonant, const std::array<int, 2>> DaughtersD0Resonant =
```

vkucera (changes requested):

PWGHF/Utils/utilsMcMatching.h
```diff
@@ -0,0 +1,244 @@
...
+    case o2::constants::physics::Pdg::kDS:
+      return DaughtersDsMain;
+    case o2::constants::physics::Pdg::kDStar:
+      return DaughtersDstarMain;
+    case o2::constants::physics::Pdg::kLambdaCPlus:
+      return DaughtersLcMain;
+    case o2::constants::physics::Pdg::kXiCPlus:
+      return DaughtersXiCMain;
+    default:
+      LOG(error) << "Unknown PDG code for 3-prong final states: " << pdgMother;
```
vkucera, line 159: I don't see the point of continuing if this happens. I would throw a fatal here.

PWGHF/Utils/utilsMcMatching.h
```diff
@@ -0,0 +1,244 @@
...
+};
+
+static const std::unordered_map<DecayChannelMain, const std::vector<int>> DaughtersLcMain{
+  {DecayChannelMain::LcToPKPi, {+kProton, +kKMinus, +kPiPlus}},
+  {DecayChannelMain::LcToPKPiPi0, {+kProton, +kKMinus, +kPiPlus, +kPi0}},
+  {DecayChannelMain::LcToPPiPi, {+kProton, +kPiMinus, +kPiPlus}},
+  {DecayChannelMain::LcToPKK, {+kProton, +kKMinus, +kKPlus}}};
+
+// Xic → p K∓ π±
+static const std::unordered_map<DecayChannelResonant, const std::array<int, 2>> DaughtersXiCResonant{
```
vkucera, line 131:
```suggestion
static const std::unordered_map<DecayChannelResonant, const std::array<int, 2>> DaughtersXicResonant{
```

PWGHF/Utils/utilsMcMatching.h
```diff
@@ -0,0 +1,244 @@
...
+  {DecayChannelResonant::D0ToKstar0Pi0, {-o2::constants::physics::kK0Star892, +kPi0}},
+  {DecayChannelResonant::D0ToKstarPi, {-o2::constants::physics::kKPlusStar892, +kPiPlus}},
+};
+
+} // namespace hf_cand_2prong
+
+namespace hf_cand_3prong
+{
+
+static const std::unordered_map<DecayChannelResonant, const std::array<int, 2>> DaughtersDPlusResonant{
```
vkucera, line 58:
```suggestion
static const std::unordered_map<DecayChannelResonant, const std::array<int, 2>> DaughtersDplusResonant{
```

PWGHF/Utils/utilsMcMatching.h
```diff
@@ -0,0 +1,244 @@
...
+  {DecayChannelResonant::DplusToPhiPi, {+o2::constants::physics::kPhi, +kPiPlus}},
+  {DecayChannelResonant::DplusToKstar0K, {-o2::constants::physics::kK0Star892, +kKPlus}},
+  {DecayChannelResonant::DplusToKstar1430_0K, {+10311, +kKPlus}},
+  {DecayChannelResonant::DplusToRho0Pi, {+kRho770_0, +kPiPlus}},
+  {DecayChannelResonant::DplusToF2_1270Pi, {+225, +kPiPlus}},
+};
+
+static const std::unordered_map<DecayChannelMain, const std::vector<int>> DaughtersDPlusMain{
```
vkucera, line 66: Keep `Main` before `Resonant`.

PWGHF/Utils/utilsMcMatching.h
```diff
@@ -0,0 +1,244 @@
...
+static const std::unordered_map<DecayChannelMain, const std::vector<int>> DaughtersXiCMain{
+  {DecayChannelMain::XicToPKPi, {+kProton, +kKMinus, +kPiPlus}},
+  {DecayChannelMain::XicToPKK, {+kProton, +kKMinus, +kKPlus}},
+  {DecayChannelMain::XicToSPiPi, {+kSigmaPlus, +kPiMinus, +kPiPlus}},
+};
+
+/// Returns a map of the possible final states for a specific 3-prong particle specie
+/// \param pdgMother PDG code of the mother particle
+/// \return a map of final states with their corresponding PDG codes
+inline std::unordered_map<DecayChannelMain, const std::vector<int>> getDecayChannel3Prong(int pdgMother)
```
vkucera, line 145: The suffix `3Prong` is redundant since the function is withing the `hf_cand_3prong` namespace.

vkucera: `Main` is missing.

PWGHF/Utils/utilsMcMatching.h
```diff
@@ -0,0 +1,244 @@
...
+    default:
+      LOG(error) << "Unknown PDG code for 3-prong final states: " << pdgMother;
+      return {};
+  }
+}
+
+/// Returns a map of the resonant decay channels for a specific 3-prong particle specie
+/// \param pdgMother PDG code of the mother particle
+/// \return a map of resonant decay channels with their corresponding PDG codes
+inline std::unordered_map<DecayChannelResonant, const std::array<int, 2>> getResoChannels3Prong(int pdgMother)
```
vkucera, line 167: Keep the names consistent between `Main` and `Resonant`.

PWGHF/Utils/utilsMcMatching.h
```diff
@@ -0,0 +1,244 @@
...
+
+} // namespace hf_cand_3prong
+
+/// Perform the matching for a single resonant channel
+/// \tparam N size of the array of daughter PDG codes
+/// \param arrPdgResoChn array of daughter indices
+/// \param arrPdgDaugs array of PDG codes for the resonant decay
+/// \return true if the resonant channel is matched, false otherwise
+template <std::size_t N>
+inline bool checkResonantDecay(std::array<int, N> const& arrPdgResoChn, std::array<int, N> arrPdgDaugs)
```
vkucera, line 194: There is nothing specific to resonant channels in this function which makes the name misleading.

vkucera (changes requested):

PWGHF/TableProducer/candidateCreator2Prong.cxx
```diff
@@ -762,47 +770,111 @@ struct HfCandidateCreator2ProngExpressions {
...
-      }
+      if (matchCorrBkgs) {
+        indexRec = -1; // Index of the matched reconstructed candidate
+        int depth = 2;
+        for (const auto& [chn, finalState] : hf_cand_2prong::DaughtersD0Main) {
+          std::array<int, 2> finalStateParts2Prong = std::array{finalState[0], finalState[1]};
+          if (finalState.size() == 3) { // Partly Reco 2-prong decays
+            if (matchKinkedDecayTopology && matchInteractionsWithMaterial) {
+              indexRec = RecoDecay::getMatchedMCRec<false, false, true, true, true>(mcParticles, arrayDaughters, Pdg::kD0, finalStateParts2Prong, true, &sign, depth, &nKinkedTracks, &nInteractionsWithMaterial);
+            } else if (matchKinkedDecayTopology && !matchInteractionsWithMaterial) {
```
vkucera, line 787: Add back the comment for D0.

PWGHF/TableProducer/candidateCreator2Prong.cxx
```diff
@@ -762,47 +770,111 @@ struct HfCandidateCreator2ProngExpressions {
...
+                    part = -part; // The Pi0 pdg code does not change between particle and antiparticle
+                  }
+                }
+              }
+              if (!RecoDecay::isMatchedMCGen(mcParticles, motherParticle, Pdg::kD0, finalStateParts2ProngAll, true, &sign, depth)) {
+                indexRec = -1; // Reset indexRec if the generated decay does not match the reconstructed one does not match the reconstructed one
+              }
+            }
+          } else if (finalState.size() == 2) { // Fully Reco 2-prong decays
+            if (matchKinkedDecayTopology && matchInteractionsWithMaterial) {
```
vkucera, line 810: What is the difference between `size == 3` and `size == 2`?

PWGHF/TableProducer/candidateCreator2Prong.cxx
```diff
@@ -762,47 +770,111 @@ struct HfCandidateCreator2ProngExpressions {
...
+              indexRec = RecoDecay::getMatchedMCRec<false, false, false, true, true>(mcParticles, arrayDaughters, Pdg::kD0, finalStateParts2Prong, true, &sign, depth, &nKinkedTracks, &nInteractionsWithMaterial);
+            } else if (matchKinkedDecayTopology && !matchInteractionsWithMaterial) {
+              indexRec = RecoDecay::getMatchedMCRec<false, false, false, true, false>(mcParticles, arrayDaughters, Pdg::kD0, finalStateParts2Prong, true, &sign, depth, &nKinkedTracks);
+            } else if (!matchKinkedDecayTopology && matchInteractionsWithMaterial) {
+              indexRec = RecoDecay::getMatchedMCRec<false, false, false, false, true>(mcParticles, arrayDaughters, Pdg::kD0, finalStateParts2Prong, true, &sign, depth, nullptr, &nInteractionsWithMaterial);
+            } else {
+              indexRec = RecoDecay::getMatchedMCRec<false, false, false, false, false>(mcParticles, arrayDaughters, Pdg::kD0, finalStateParts2Prong, true, &sign, depth);
+            }
+          } else {
+            LOG(info) << "Final state size not supported: " << finalStateParts2Prong.size();
```
vkucera, line 820: Should this not be a fatal?

PWGHF/TableProducer/candidateCreator3Prong.cxx
```diff
@@ -941,129 +947,220 @@ struct HfCandidateCreator3ProngExpressions {
...
+                auto motherParticle = mcParticles.rawIteratorAt(indexRec);
+                if (finalState.size() == 4) { // Check if the final state has 4 particles
+                  std::array<int, 4> finalStateParts3ProngAll = std::array{finalState[0], finalState[1], finalState[2], finalState[3]};
+                  if (sign < 0) {
+                    for (auto& part : finalStateParts3ProngAll) {
+                      if (part == kPi0) {
+                        part = -part; // The Pi0 pdg code does not change between particle and antiparticle
+                      }
+                    }
+                  }
```
vkucera, line 982: This could be a utility function that takes `sign`, `finalStateParts3ProngAll`, `pdg` as arguments.

PWGHF/Utils/utilsMcMatching.h
```diff
@@ -0,0 +1,244 @@
...
+/// Function documentation:
+/// \tparam is3Prong bool to specify if the mother decays with a 3-prong decay
+/// \tparam N size of the array of daughter PDG codes
+/// \param motherPdg PDG code of the mother particle
+/// \param channel decay channel flag to be set
+/// \param arrDaughPdgs array of daughter PDG codes
+template <bool is3Prong = false, std::size_t N>
+inline void flagResonantDecay(int motherPdg, int8_t* channel, std::array<int, N> const& arrDaughPdgs)
+{
+  if constexpr (is3Prong) {
```
vkucera, line 222: Why is `is3Prong` a template parameter?
Why is `is3Prong` needed at all?

PWGHF/Utils/utilsMcMatching.h
```diff
@@ -0,0 +1,244 @@
...
+    std::unordered_map<o2::hf_decay::hf_cand_3prong::DecayChannelResonant, const std::array<int, 2>> resoStates = o2::hf_decay::hf_cand_3prong::getResoChannels3Prong(motherPdg);
+    for (const auto& [flag, pdgCodes] : resoStates) {
+      if (o2::hf_decay::checkResonantDecay(arrDaughPdgs, pdgCodes)) {
+        *channel = flag;
+        break;
```
vkucera, line 227: It would be more transparent if the function simply returned `channel` since it is its only purpose.

PWGHF/Utils/utilsMcMatching.h
```diff
@@ -0,0 +1,244 @@
...
+
+/// Flag the resonant decays
+/// Function documentation:
+/// \tparam is3Prong bool to specify if the mother decays with a 3-prong decay
+/// \tparam N size of the array of daughter PDG codes
+/// \param motherPdg PDG code of the mother particle
+/// \param channel decay channel flag to be set
+/// \param arrDaughPdgs array of daughter PDG codes
+template <bool is3Prong = false, std::size_t N>
+inline void flagResonantDecay(int motherPdg, int8_t* channel, std::array<int, N> const& arrDaughPdgs)
```
vkucera, line 220: `channel` should be set to `0` by default since the function is supposed to always give a correct value.

vkucera (changes requested):

PWGHF/TableProducer/candidateCreator2Prong.cxx
Marcellocosti: Here, `size == 3` covers the cases where the final state has 3 final particles, thus the `acceptIncompleteReco` flag of the `getMatchedMCRec` must be turned on. The same occurs in the `candidateCreator3Prong` for the cases with `size == 4` and `size == 5`. Unfortunately, due to the need of passing `std::array` types to the functions for the generated matching, I couldn't find a better solution than explicitly separating the cases. If you have ideas on improvements please tell me!

PWGHF/Utils/utilsMcMatching.h
Marcellocosti: A possible solution would be to move the function `getDecayChannel3Prong` from the `hf_cand_3prong` namespace to the `hf_decay` one, and add the D0 in the switch cases. This would allow an easy extension to future particles. The drawback I see is that, since the maps are declared in different namespaces, one would need to use an `std::variant` data type to cover the cases. Do you want me to keep the implementation like this, go ahead with the solution I am proposing here or you have a further one?

PWGHF/TableProducer/candidateCreator2Prong.cxx
Marcellocosti: I am sorry but I don't get which comment you are referring to, can you please point me to what you meant? Thanks!

vkucera: I meant the `// D0 -> ....` comment. Now it is moved into the `else` block.

vkucera: Sure but the `if` block is also matching D0 so it would be good to have the comment there too.

PWGHF/Utils/utilsMcGen.h
```diff
@@ -17,56 +17,95 @@
...
-#include "Common/Core/RecoDecay.h"
-
-#include "PWGHF/Core/DecayChannels.h"
-#include "PWGHF/DataModel/CandidateReconstructionTables.h"
-
 namespace hf_mc_gen
 {
 
-template <typename T, typename U, typename V>
+template <bool matchCorrBkgs = false, typename T, typename U, typename V>
```
vkucera, line 39: Why does `matchCorrBkgs` need to be a template parameter?

PWGHF/Utils/utilsMcGen.h
```diff
@@ -17,56 +17,95 @@
...
 void fillMcMatchGen2Prong(T const& mcParticles, U const& mcParticlesPerMcColl, V& rowMcMatchGen, bool rejectBackground)
 {
   using namespace o2::constants::physics;
+  constexpr std::size_t NDaughtersResonant{2u};
 
   // Match generated particles.
   for (const auto& particle : mcParticlesPerMcColl) {
     int8_t flag = 0;
     int8_t origin = 0;
+    int8_t channel = 0; // Not used in 2-prong decays
```
vkucera, line 49: I see it used for D0 below.

PWGHF/Utils/utilsMcMatching.h
```diff
@@ -0,0 +1,269 @@
...
+    }
+    if (!findDaug) {
+      return false;
+    }
+  }
+  return true;
+}
+
+/// Flag the resonant decays
+/// Function documentation:
```
vkucera, line 191:
```suggestion
```

PWGHF/Utils/utilsMcMatching.h
```diff
@@ -0,0 +1,269 @@
...
+/// \tparam is3Prong bool to specify if the mother decays with a 3-prong decay
+/// \tparam N size of the array of daughter PDG codes
+/// \param motherPdg PDG code of the mother particle
+/// \param channel decay channel flag to be set
+/// \param arrDaughPdgs array of daughter PDG codes
+template <std::size_t N>
```
vkucera, line 197: Add documentation for `\return`.

PWGHF/Utils/utilsMcMatching.h
```diff
@@ -0,0 +1,269 @@
...
+/// of PDG codes representing either a final or a resonant state
+/// \tparam N size of the array of PDG codes
+/// \param motherPdgCode PDG code of the mother particle
+/// \param partPdgCode PDG code of the target particle
+/// \param arrFinalStatePdgs array of PDG codes to be modified
+template <std::size_t N>
+inline void changeFinalStatePdgSign(const int motherPdgCode, const int partPdgCode, std::array<int, N>& arrFinalStatePdgs)
+{
+  if (motherPdgCode >= 0)
+    return;
```
vkucera, line 260:
```suggestion
  if (motherPdgCode >= 0) {
    return;
  }
```

PWGHF/Utils/utilsMcMatching.h
```diff
@@ -0,0 +1,269 @@
...
+  return true;
+}
+
+/// Flag the resonant decays
+/// Function documentation:
+/// \tparam is3Prong bool to specify if the mother decays with a 3-prong decay
+/// \tparam N size of the array of daughter PDG codes
+/// \param motherPdg PDG code of the mother particle
+/// \param channel decay channel flag to be set
+/// \param arrDaughPdgs array of daughter PDG codes
```
vkucera, line 196: Fix the documentation.

PWGHF/Utils/utilsMcMatching.h
```diff
@@ -0,0 +1,269 @@
...
+    default:
+      LOG(fatal) << "Unknown PDG code for 3-prong final states: " << pdgMother;
+      return {};
+  }
+}
+} // namespace hf_cand_3prong
+
+/// Perform the matching for a single resonant channel
+/// \tparam N size of the array of daughter PDG codes
+/// \param arrPdgResoChn array of daughter indices
```
vkucera, line 168: arrPdgResoChn is not an array of daughter indices

PWGHF/Utils/utilsMcMatching.h
```diff
@@ -0,0 +1,269 @@
...
+      return daughtersXicMain;
+    default:
+      LOG(fatal) << "Unknown PDG code for 3-prong final states: " << pdgMother;
+      return {};
+  }
+}
+} // namespace hf_cand_3prong
+
+/// Perform the matching for a single resonant channel
+/// \tparam N size of the array of daughter PDG codes
```
vkucera, line 167: The function has nothing to do with resonant channels per se, as I pointed out earlier.

PWGHF/Utils/utilsMcMatching.h
```diff
@@ -0,0 +1,269 @@
...
+/// \param arrPdgResoChn array of daughter indices
+/// \param arrPdgDaugs array of PDG codes for the resonant decay
+/// \return true if the resonant channel is matched, false otherwise
+template <std::size_t N>
+inline bool checkDecayChannel(std::array<int, N> const& arrPdgResoChn, std::array<int, N> arrPdgDaugs)
```
vkucera, line 172: Make the names obvious to distinguish what is the expected array and what is the tested array.

vkucera (changes requested):

PWGHF/Utils/utilsMcGen.h
Marcellocosti: I implemented it as a template parameter since the matching of the generated candidates is performed in an header file, thus I cannot add configurables there. The only alternative I see is to have separate functions, which option do you prefer?

PWGHF/Core/DecayChannels.h
```diff
@@ -102,6 +108,16 @@ enum DecayChannelResonant : int8_t {
...
+  // D*+
+  DstarToD0ToRhoplusPi,       // ρ+ π−
+  DstarToD0ToRhoplusK,        // ρ+ K−
+  DstarToD0ToKstar0Pi0,       // anti-K*0 π0
+  DstarToD0ToKstarPi,         // K*− π+
+  DstarToDplusToPhiPi,        // φ π+
+  DstarToDplusToKstar0K,      // anti-K*0 K+
+  DstarToDplusToKstar1430_0K, // anti-K*0(1430) K+
+  DstarToDplusToRho0Pi,       // ρ0 π+
+  DstarToDplusToF2_1270Pi,    // f2(1270) π+
```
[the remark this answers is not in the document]
Marcellocosti: I think it is more precise and the interpretation is less error-prone, but I can remove the names if you prefer.

PWGHF/TableProducer/candidateCreator3Prong.cxx
vkucera: Seems addressed.

vkucera (approved):
Thanks a lot @Marcellocosti for the great work!

PWGHF/TableProducer/candidateCreator3Prong.cxx
```diff
@@ -941,129 +946,218 @@ struct HfCandidateCreator3ProngExpressions {
...
           }
-          RecoDecay::getDaughters(mcParticles.rawIteratorAt(indexRec), &arrDaughIndex, std::array{0}, 1);
-          if (arrDaughIndex.size() == NDaughtersResonant) {
-            for (auto iProng = 0u; iProng < arrDaughIndex.size(); ++iProng) {
-              auto daughI = mcParticles.rawIteratorAt(arrDaughIndex[iProng]);
-              arrPDGDaugh[iProng] = std::abs(daughI.pdgCode());
+          if (indexRec > -1) {
+            flag = sign * (isDplus ? DecayChannelMain::DplusToPiKK : DecayChannelMain::DsToPiKK);
+            if (arrayDaughters[0].has_mcParticle()) {
+              swapping = int8_t(std::abs(arrayDaughters[0].mcParticle().pdgCode()) == kPiPlus);
```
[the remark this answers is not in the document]
vkucera: Why?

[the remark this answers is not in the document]
vkucera: But the expected order is defined as KKPi here in:
```c++
auto arrPdgDaughtersDToPiKK{std::array{+kKPlus, -kKPlus, +kPiPlus}};
```

PR #11430, opened 2025-06-01
Additional histograms to calculate ZP over ZN v3 

omvazque:


abmodak (changes requested):
Hello @omvazque,
Could you please change the names of the following variables (in a more standard way)?
KcollEnergy => kCollEnergy
KzEro => kZero

Thanks,
Abhi

abmodak (approved):

PR #11434, opened 2025-06-02
Common - add trackdca and trackdcacov table producers for Run2 conver…

aimeric-landou:
…ter data


aimeric-landou:
All but one O2linter errors are from files I did not modify in this.
The one about the new `trackDcaCovFillerRun2` file is due to `Configurable<std::string> ccdburl{"ccdb-url",[...]` mismatch, but I decided to leave it as that is what is currently used in the `trackPropagation.cxx` file, which I used as base for this new workflow.

aimeric-landou:
I am not sure why the macOS build fails at `gpu::gpustd::array`: it works fine on my macOS local machine. Has this error been seen before?

PR #11439, opened 2025-06-02
tuned histogram bins etc

sashingo:


PWGHF/HFL/Tasks/taskElectronWeakBoson.cxx
```diff
@@ -147,27 +147,29 @@ struct HfTaskElectronWeakBoson {
       zorroSummary.setObject(zorro.getZorroSummary());
     }
 
+    // add configurable for CCDB path
+    zorro.setBaseCCDBPath(cfgCCDBPath.value);
+
```
fgrosa, line 152: Why do you need `zorro` if you don't use it to select the collisions? See e.g.
https://github.com/AliceO2Group/O2Physics/blob/2520da892049238a17af1cd415a568e650695978/PWGHF/Utils/utilsEvSelHf.h#L331-L333
You would also need the `initCCDB` to use `zorro` properly:
https://github.com/AliceO2Group/O2Physics/blob/2520da892049238a17af1cd415a568e650695978/PWGHF/Utils/utilsEvSelHf.h#L325

PWGHF/HFL/Tasks/taskElectronWeakBoson.cxx
```diff
@@ -147,27 +147,29 @@ struct HfTaskElectronWeakBoson {
...
     // define axes you want to use
-    const AxisSpec axisZvtx{400, -20, 20, "Zvtx"};
+    const AxisSpec axisZvtx{40, -20, 20, "Zvtx"};
```
fgrosa, line 154: Consider to make it configurable to avoid changing the code when you want to modify the number of bins and the range (same for all the axes below). See e.g.
https://github.com/AliceO2Group/O2Physics/blob/2520da892049238a17af1cd415a568e650695978/PWGLF/Tasks/QC/vertexQA.cxx#L80

fgrosa (commented):
Hi @sashingo, no major comments, but please take a look at my suggestions below and consider to implement them. Thanks!

PWGHF/HFL/Tasks/taskElectronWeakBoson.cxx
sashingo: Actually we use Zorro to select the software trigger (EMCal trigger), and we already confirmed that it works as we reported HFCL meeting.
We also use initCCDB
      // initialize Zorro
      zorro.initCCDB(ccdb.service, runNumber, currentTimestamp, cfgTriggerName);

fgrosa: Hi @sashingo sorry, I missed the `initCCDB`!
Indeed this does correctly the accounting of the TVX, but you still miss the call to `isSelected()`. Without it you are not sure that you are only selecting triggered collisions

fgrosa: It looks good to me (you can skip `&& cfgSkimmedProcessing` in the `if` statement before the return, since `cfgSkimmedProcessing` is always true if those lines are executed). When you push this change I will approve and merge, thanks!

fgrosa (approved):

PR #11444, opened 2025-06-02
[PWGLF] feat: Add nuclei pair table to data model

mpuccio:
This commit adds a new table called `NucleiPairTable` to the data model. The new table contains information about pairs of light nuclei, including their kinematic properties, detector signals, and quality flags.

The changes include:

- Defining a new namespace `NucleiPairTableNS` with columns for the first and second particles in the pair.
- Declaring a new table `NucleiPairTable` in the `o2::aod` namespace, which includes the columns from `NucleiPairTableNS`.
- Modifying the `nucleiSpectra.cxx` file to produce the new `NucleiPairTable` in addition to the existing `NucleiTable`.
- Adding a new configurable `cfgFillPairTree` to control whether the pair table should be filled.

These changes will enable the analysis of light nuclei pairs.


mpuccio:
@ivorobye

maciacco (approved):
@skundu692 could you have a look?

skundu692 (approved):

PR #11463, opened 2025-06-04
[PWGCF] Add histograms for monte carlo

nidamalikk:


PWGCF/EbyEFluctuations/Tasks/netchargeFluctuations.cxx
```diff
@@ -9,94 +9,732 @@
...
+
+DECLARE_SOA_TABLE(NetChargeGen, "AOD", "NETfluctGen",
+                  net_charge_gen::PosCharge,
+                  net_charge_gen::NegCharge,
+                  net_charge_gen::PosSqCharge,
+                  net_charge_gen::NegSqCharge,
+                  net_charge_gen::TermPCharge,
+                  net_charge_gen::TermNCharge,
+                  net_charge_gen::PosNegCharge,
+                  net_charge_gen::Centrality);
```
victor-gonzalez, line 119: Tables are declared with a plural name while their iterators have the same name but in singular
Why are two tables needed for reco data and reco MC? They are the same, right?

nidamalikk: Hello Victor, I tried to modify my code according to your comments.

PR #11471, opened 2025-06-04
[PWGCF] Made a change to the RD method cuts

JStaa:


PWGCF/MultiparticleCorrelations/Tasks/threeParticleCorrelations.cxx
```diff
@@ -881,36 +890,38 @@ struct ThreeParticleCorrelations {
...
   bool acceptEvent(const CollCand& collision, bool FillHist) // Event filter
   {
+
     if (FillHist) {
       rQARegistry.fill(HIST("hNEvents"), 0.5);
     }
 
     if (!collision.selection_bit(aod::evsel::kNoSameBunchPileup)) { // kNoSameBunchPileup
-      return false;
+      return kFALSE;
```
victor-gonzalez, line 899: Any reason for returning back to `ROOT` naming?

PWGCF/MultiparticleCorrelations/Tasks/threeParticleCorrelations.cxx
```diff
@@ -1075,6 +1081,7 @@ struct ThreeParticleCorrelations {
   bool radialDistanceFilter(const V0Cand& v0, const TrackCand& track, double B, bool Mix)
   {
 
+    bool pass = true;
```
victor-gonzalez, line 1084: I like this way
I would suggest consistency

PR #11472, opened 2025-06-04
[PWGDQ] Fill vertexing chi2PCA for triplets

sigurdnese:


PWGDQ/Core/VarManager.h
```diff
@@ -4006,6 +4007,10 @@ void VarManager::FillTripletVertexing(C const& collision, T const& t1, T const&
       o2::dataformats::VertexBase primaryVertex = {std::move(vtxXYZ), std::move(vtxCov)};
       auto covMatrixPV = primaryVertex.getCov();
 
+      auto chi2PCA = fgFitterThreeProngBarrel.getChi2AtPCACandidate();
+      if (fgUsedVars[kVertexingChi2PCA])
+        values[VarManager::kVertexingChi2PCA] = chi2PCA;
```
iarsene, line 4012: In order to save the actual computing in the case this variable is not used, you should include the chi2PCA also under the if statement. Also, not sure why the test did not catch this, but the lines guarded by the if statement should be placed inside braces {}, even if its just one line.

iarsene (approved):

PR #11475, opened 2025-06-05
[PWGCF] Addition of Monte Carlo study with changes of PID cuts

sweta29singh:


victor-gonzalez (changes requested):
Before reviewing it, could you please rebase your changes to the latest version of O2Physics? Your base is 4500 commits behind the master!

PR #11495, opened 2025-06-06
[Common] Unified, simple, fully self-configuring mult/cent provider

ddobrigk:
* adds a task to tentatively replace mult/cent services with a single, self-configuring, lightweight device following the rationale of the `strangenessbuilder`. Track loops / slices replaced by single operations for faster operation in all processing paths. 
* now with run 2 and run 3 modes tested thanks to @romainschotter. Still some tuning to be done with `defineDataProcessing` autodetect, but will follow in separate PR.

### DIFF
--- a/ALICE3/Core/DelphesO2LutWriter.cxx
+++ b/ALICE3/Core/DelphesO2LutWriter.cxx
@@ -23,9 +23,10 @@
 
 #include "ALICE3/Core/DelphesO2TrackSmearer.h"
 #include "ALICE3/Core/DelphesO2LutWriter.h"
-#include "iostream"
+#include <iostream>
 #include "TMatrixD.h"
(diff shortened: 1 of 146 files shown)


ddobrigk:
@fjonasALICE @nstrangm this PR will provide a unified mult/cent that will include BC-level centrality values for FT0 signals and, while at it, it will simplify the task, using less memory and less CPU. I will finish some testing soon and then move the PR to review, but even then, since this is intended to be a replacement for multiplicityTable and centralityTable, we'll have to commission it a bit. Hope this helps!

ddobrigk:
We're still missing the autodetect at `defineDataProcessing` but let's move this to review and take care of that extra feature later. Might still do a little cleanup etc before going ahead

PR #11505, opened 2025-06-07
[PWGEM/Dilepton] reduce memory consumption in createResolutionMap.cxx

dsekihat:
PWGEM/Dilepton: reduce memory consumption in createResolutionMap.cxx
change scheme of filling resolution maps in different event generators.


alibuild (dismissed):

PR #11515, opened 2025-06-09
[PWGCF] Added more histograms for Monte Carlo to calculate mean pt fluctuations for identified particles

sweta29singh:


sweta29singh:
Hii 

I'm encountering an error in CMakeLists.txt every time I try to commit, which seems to be related to previously having workflows configured. I've added my workflow to the file, but the issue persists. Could someone please take a look?

Thanks in Advance!

Regards
Sweta

PWGCF/EbyEFluctuations/Tasks/CMakeLists.txt
```diff
@@ -34,8 +34,8 @@ o2physics_add_dpl_workflow(antiproton-cumulants-mc
                     PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::AnalysisCore O2Physics::PWGCFCore
                     COMPONENT_NAME Analysis)
 		
-o2physics_add_dpl_workflow(identified-meanpt-fluctuations
-                    SOURCES IdentifiedMeanPtFluctuations.cxx
+o2physics_add_dpl_workflow(event-mean-pt-id
+                    SOURCES eventMeanPtId.cxx
```
victor-gonzalez, line 38: Here you have removed one file `IdentifiedMeanPtFluctuations.cxx` and one executable `identified-meanpt-fluctuations` from the build of `O2Physics`

victor-gonzalez:
Locally:
- Preserve your changed files somewhere else outside the `O2Physics` tree
- Roll back your `O2Physics` local repo to the point before your changes

If you want to change the name of a file which is under `git` version control, never do it manually or by using `mv`, you will lose the full file history on `git`. You should use `git mv old_filename new_filename`

If you need to delete a file which is under `git` version control never use `rm` or manually remove it. You should use `git rm filename`

Locally:
- rename the file you wanted to rename by using `git mv old_filename new_filename`
- recover the changes that you preserved somewhere else
- compile your changes and test that everything works as expected (you have not done this previously because the same error that you are getting in `gitHub` would have happen in the local compilation. Please don't create PR which have not been tested locally)
- commit your changes in your local repository

Force push your changes to your fork in `gitHub`
The PR will automatically be updated

sweta29singh:
> Locally:
> 
>     * Preserve your changed files somewhere else outside the `O2Physics` tree
> 
>     * Roll back your `O2Physics` local repo to the point before your changes
> 
> 
> If you want to change the name of a file which is under `git` version control, never do it manually or by using `mv`, you will lose the full file history on `git`. You should use `git mv old_filename new_filename`
> 
> If you need to delete a file which is under `git` version control never use `rm` or manually remove it. You should use `git rm filename`
> 
> Locally:
> 
>     * rename the file you wanted to rename by using `git mv old_filename new_filename`
> 
>     * recover the changes that you preserved somewhere else
> 
>     * compile your changes and test that everything works as expected (you have not done this previously because the same error that you are getting in `gitHub` would have happen in the local compilation. Please don't create PR which have not been tested locally)
> 
>     * commit your changes in your local repository
> 
> 
> Force push your changes to your fork in `gitHub` The PR will automatically be updated




Dear Victor,

I attempted to revert the o2physics tree to its previous state before I made changes while doing git commit. Unfortunately, my earlier file, IdentifiedMeanpTFluctuation.cxx, could not be restored. Could you please suggest what I should do now?

victor-gonzalez:
Your local repo is 22 commits ahead the central repo
I see this in your fork on `gitHub` but probably locally you have a few more (I cannot see that)
This means you have locally committed changes that you have to roll back until you get the local repo restored to the version before your changes. At that point the missing file will be recovered

My first point on preserve the local changes that are meaningful, in a place outside the local repo is related to this situation. Once the local changes are rolled back they will be gone unless they are preserved somewhere else

Once you recover the previous state, before proceeding, I would suggest updating the local repo. Your fork in `gitHub` suggests it is behind the central in about 100 commits

Once you are on pair with the central repository start using `git mv` and `git rm` as suggested in previous comments
Once you are done with that recover the changes you wanted to keep
Check that `O2Physics` compiles correctly and that your changes behave in the way you wanted

Commit your local changes
Force push your committed changes to the fork in `gitHub` and the PR will be automatically updated

victor-gonzalez:
The number of local commits keeps increasing which is the opposite I mentioned it should be done
All local changes have to be rolled back until there is not local commits at all

This is a drastic suggestion which should not be tried often unless assuming risk of losing local changes
- check how many commits the local repo is ahead of the `upstream/master` with `git status`
- `gitHub` shows that the fork is 28 commits ahead of the master. This number will be used next but if the local repo shows a higher number use the highest one
- rollback all the local changes with `git reset --hard HEAD~28`

If the number of local commits ahead was correctly used the local repo should be now ready
Update the local repo to the latest version of upstream/master

Use `git mv` (never delete a file with `rm` or manually change its name whit any other method than `git mv`)
Introduce the changes in the file which has been renamed and in the affected files

Compile the code locally
Test the code locally 

Push the changes to the fork in `gitHub`. The PR will be automatically updated

sweta29singh:
> The number of local commits keeps increasing which is the opposite I mentioned it should be done All local changes have to be rolled back until there is not local commits at all
> 
> This is a drastic suggestion which should not be tried often unless assuming risk of losing local changes
> 
>     * check how many commits the local repo is ahead of the `upstream/master` with `git status`
> 
>     * `gitHub` shows that the fork is 28 commits ahead of the master. This number will be used next but if the local repo shows a higher number use the highest one
> 
>     * rollback all the local changes with `git reset --hard HEAD~28`
> 
> 
> If the number of local commits ahead was correctly used the local repo should be now ready Update the local repo to the latest version of upstream/master
> 
> Use `git mv` (never delete a file with `rm` or manually change its name whit any other method than `git mv`) Introduce the changes in the file which has been renamed and in the affected files
> 
> Compile the code locally Test the code locally
> 
> Push the changes to the fork in `gitHub`. The PR will be automatically updated



Victor,
Sorry to bother you again.

I tried to roll back my local changes using git reset --hard HEAD~28, and it successfully restored my previous version of the code, i.e., IdentifiedMeanpTFluct.cxx.

As per your suggestion, I then updated my local repository to the latest version of upstream/master using:
git fetch upstream  
git rebase upstream/master

However, after doing this, the changes I had made locally are now gone. In my local directory, the macro IdentifiedMeanpTFluct.cxx is no longer visible, and instead, I only see the new macro eventMeanPtId.cxx.

I'm not sure what's going wrong and feeling a bit stuck at this point.

victor-gonzalez (approved):

PR #11518, opened 2025-06-09
[PWGHF] Add upc selection into HF framework

zhangbiao-phy:


PWGHF/Utils/utilsEvSelHf.h
```diff
@@ -145,6 +157,7 @@ struct HfEventSelection : o2::framework::ConfigurableGroup {
   o2::framework::Configurable<float> centralityMin{"centralityMin", 0., "Minimum centrality"};
   o2::framework::Configurable<float> centralityMax{"centralityMax", 100., "Maximum centrality"};
   o2::framework::Configurable<bool> useSel8Trigger{"useSel8Trigger", true, "Apply the sel8 event selection"};
+  o2::framework::Configurable<bool> useUpcTrigger{"useUpcTrigger", true, "Apply the upc event selection"};
```
mfaggin, line 160:
```suggestion
  o2::framework::Configurable<bool> useUpcTrigger{"useUpcTrigger", false, "Apply the upc event selection"};
```
I'd not enable it by default

PWGHF/Tasks/taskPidStudies.cxx
```diff
@@ -197,6 +197,7 @@ struct HfTaskPidStudies {
   using CollisionsMc = soa::Join<aod::Collisions, aod::McCollisionLabels, aod::EvSels, aod::CentFT0Cs, aod::CentFT0Ms>;
   using V0sMcRec = soa::Join<aod::V0Datas, aod::V0CoreMCLabels>;
   using CascsMcRec = soa::Join<aod::CascDatas, aod::CascCoreMCLabels>;
+  using BCs = soa::Join<aod::BCsWithTimestamps, aod::BcSels, aod::Run3MatchedToBCSparse>;
```
mfaggin, line 200: You define in each file this new data type, which looks like to be the same for each file, isn't it? If this is the case, I'd personally put it inside `PWGHF/Utils/utilsEvSelHf.h`

PWGHF/Tasks/taskMcValidation.cxx
```diff
@@ -1167,7 +1168,7 @@ struct HfTaskMcValidationRec {
                           aod::TracksWMc const& mcTracks,
                           aod::McParticles const& mcParticles,
                           aod::McCollisions const& mcCollisions,
-                          aod::BCsWithTimestamps const& bcs,
+                          BCsInfo const& bcs,
```
mfaggin, line 1171: I'd call the `BCsInfo` data type if and only if UPC selections are enabled. This means that in the process functions that do not apply UPC selections I'd not use at all the data type `BCsInfo`, but I'd keep the data type `aod::BCsWithTimestamps`

PWGHF/TableProducer/candidateCreator2Prong.cxx
```diff
@@ -95,6 +95,7 @@ struct HfCandidateCreator2Prong {
   Service<o2::ccdb::BasicCCDBManager> ccdb;
 
   using TracksWCovExtraPidPiKa = soa::Join<aod::TracksWCovExtra, aod::TracksPidPi, aod::PidTpcTofFullPi, aod::TracksPidKa, aod::PidTpcTofFullKa>;
+  using BCs = soa::Join<aod::BCsWithTimestamps, aod::BcSels, aod::Run3MatchedToBCSparse>;
```
vkucera, line 98: How will this run on Run 2 data?

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -155,46 +156,55 @@ struct HfTrackIndexSkimCreatorTagSelCollisions {
...
   {
     selectCollision<true, CentralityEstimator::None>(collision, bcs);
   }
   PROCESS_SWITCH(HfTrackIndexSkimCreatorTagSelCollisions, processTrigSel, "Use trigger selection", false);
 
   /// Event selection without trigger selection
-  void processNoTrigSel(aod::Collision const& collision, aod::BCsWithTimestamps const& bcs)
+  void processNoTrigSel(aod::Collision const& collision, BCs const& bcs)
   {
+    LOGF(info, " in processNoTrigSel xxxxxxxxxxxxxxxx");
```
vkucera, line 196: Why? This will just spam the log with plenty of useless messages.

vkucera (changes requested):

PWGHF/D2H/TableProducer/dataCreatorCharmHadPiReduced.cxx
```diff
@@ -195,6 +195,7 @@ struct HfDataCreatorCharmHadPiReduced {
   using CollisionsWCent = soa::Join<aod::Collisions, aod::EvSels, aod::CentFT0Ms, aod::CentFT0Cs>;
   using CollisionsWCentAndMcLabels = soa::Join<CollisionsWCent, aod::McCollisionLabels>;
   using CollisionsWCentAndQvectors = soa::Join<CollisionsWCent, aod::QvectorFT0Cs, aod::QvectorFT0As, aod::QvectorFT0Ms, aod::QvectorTPCposs, aod::QvectorTPCnegs, aod::QvectorTPCalls>;
+  using BCs = soa::Join<BCsWithTimestamps, aod::BcSels, aod::Run3MatchedToBCSparse>;
```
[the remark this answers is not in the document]
zhangbiao-phy: yes, I am try to change a name for that

PWGHF/Utils/utilsEvSelHf.h
```diff
@@ -181,8 +198,21 @@ struct HfEventSelection : o2::framework::ConfigurableGroup {
...
+  static constexpr int MinNdtcoll = 1;       // Default number of sigma (NDtcoll)
+  static constexpr int MinNBCs = 2;          // Minimum number of bunch crossings (NBCs)
+  static constexpr int MinNTracks = 2;       // Minimum number of PV contributors
+  static constexpr int MaxNTracks = 1000;    // Maximum number of PV contributors
+  static constexpr float MaxFITtime = 34.f;  // Maximum FIT time in ns
+  static constexpr float FITAmpFV0 = -1.f;   // FV0 amplitude
+  static constexpr float FITAmpFT0A = 150.f; // Max FT0A amplitude
+  static constexpr float FITAmpFT0C = 50.f;  // Max FT0C amplitude
+  static constexpr float FITAmpFDDA = -1.f;  // FDDA amplitude
+  static constexpr float FITAmpFDDC = -1.f;  // FDDC amplitude
```
vkucera, line 213: Use the recommended naming scheme.
Initialise with braces.
What is NDtcoll?

PWGHF/Utils/utilsEvSelHf.h
```diff
@@ -193,6 +223,21 @@ struct HfEventSelection : o2::framework::ConfigurableGroup {
   o2::framework::OutputObj<ZorroSummary> zorroSummary{"zorroSummary"};
   int currentRun{-1};
 
+  /// Set standard preselection gap trigger (values taken from UD group)
+  void setSGCutsFromUD(SGCutParHolder& sgCuts)
```
vkucera, line 227: "from UD" is not an attribute that should appear in the name.

PWGHF/TableProducer/trackIndexSkimCreator.cxx
zhangbiao-phy: sorry, just forgot remove it from debug mode

PWGHF/TableProducer/candidateCreator2Prong.cxx
zhangbiao-phy: I don't know, but from https://github.com/AliceO2Group/AliceO2/blob/8e0693280e134e41f0fe3d253bdda26b49831a80/Framework/Core/include/Framework/AnalysisDataModel.h#L2182-L2183, it miss FV0CId in run3

PWGHF/Utils/utilsEvSelHf.h
zhangbiao-phy: Thanks for spoting this

PWGHF/Utils/utilsEvSelHf.h
```diff
@@ -238,10 +283,9 @@ struct HfEventSelection : o2::framework::ConfigurableGroup {
   /// \param registry reference to the histogram registry needed for zorro
   /// \return bitmask with the event selection criteria not satisfied by the collision
   template <bool useEvSel, o2::hf_centrality::CentralityEstimator centEstimator, typename BCs, typename Coll>
-  uint16_t getHfCollisionRejectionMask(const Coll& collision, float& centrality, o2::framework::Service<o2::ccdb::BasicCCDBManager> const& ccdb, o2::framework::HistogramRegistry& registry)
+  uint32_t getHfCollisionRejectionMask(const Coll& collision, float& centrality, o2::framework::Service<o2::ccdb::BasicCCDBManager> const& ccdb, o2::framework::HistogramRegistry& registry, const BCs& bcs)
```
vkucera, line 286: The change of size should be propagated to other relevant places in O2Physics.

vkucera (changes requested):

PWGHF/Utils/utilsEvSelHf.h
zhangbiao-phy: Yes, and in all cases, the value is retrieved using `const auto rejectionMask = hfEvSel.getHfCollisionRejectionMask(...);`

PWGHF/Utils/utilsEvSelHf.h
zhangbiao-phy: ok, NDtcoll is minimum number of bunch crossings around the collision of interest, defined by collision time resolution (σ(tcoll)) over LHC bunch spacing time (∆t_{bc})

PWGHF/Tasks/taskPidStudies.cxx
mfaggin: I see that in many files (almost everywhere if I am not wrong) you have `using BCs = soa::Join<aod::BCsWithTimestamps, aod::BcSels, aod::Run3MatchedToBCSparse>;`. Since you have always the `#include "PWGHF/Utils/utilsEvSelHf.h"` maybe you can put the `using` there

PWGHF/TableProducer/candidateCreator2Prong.cxx
vkucera: There should be an option to process Run 2 data.

PWGHF/Utils/utilsEvSelHf.h
vkucera, line 286: The `bc` type should be templated to work with any `Join`.

PWGHF/Utils/utilsEvSelHf.h
```diff
@@ -108,10 +109,20 @@ enum EventRejection {
...
   NEventRejection
 };
 
+// upc event type
+enum UPCEventType {
+  kSingleGapA,
+  kSingleGapC,
+  kDoubleGap,
+  kNEventTypes
+};
```
vkucera, line 122: Remove the `k` prefixes.
Please call the enumerator `EventTypeUpc`.

PWGHF/Utils/utilsEvSelHf.h
```diff
@@ -181,8 +198,21 @@ struct HfEventSelection : o2::framework::ConfigurableGroup {
...
+  static constexpr int kMinNDtColl{1};        // Miminimum number of bunch crossings around the collision of interest
+  static constexpr int kMinNBCs{2};           // Minimum number of bunch crossings
+  static constexpr int kMinNTracks{2};        // Minimum number of PV contributors
+  static constexpr int kMaxNTracks{1000};     // Maximum number of PV contributors
+  static constexpr float kMaxFITTimeNs{34.f}; // Maximum FIT time in ns
+  static constexpr float kFITAmpFV0{-1.f};    // FV0 amplitude threshold
+  static constexpr float kFITAmpFT0A{150.f};  // Maximum FT0A amplitude
+  static constexpr float kFITAmpFT0C{50.f};   // Maximum FT0C amplitude
+  static constexpr float kFITAmpFDDA{-1.f};   // FDDA amplitude threshold
+  static constexpr float kFITAmpFDDC{-1.f};   // FDDC amplitude threshold
```
vkucera, line 213: Remove the `k` prefixes.
Follow the naming conventions and the naming scheme.

PWGHF/Utils/utilsEvSelHf.h
zhangbiao-phy: fixed

zhangbiao-phy:
> Hi @zhangbiao-phy many thanks for this development! Before a detailed review, I have two main comments:
> 
> 1. in the `hCollisions` histogram the `Zvtx` cut should always be at the end, in order to be able to evaluate the `Zvtx` cut efficiency
> 2. Since we are adding a dependency on a library, and on a table, it is very likely that this development will imply a memory increase (and we are always at the limit of the memory for our analyses). I would ask you if you could evaluate the increase of memory, and possibly put this event selection as optional to allow not to have it where not needed (for instance in the D2H reduced data creators). Thanks!
Hi @fgrosa, thanks for the comments!
> 1. in the `hCollisions` histogram the `Zvtx` cut should always be at the end, in order to be able to evaluate the `Zvtx` cut efficiency
Ok, thanks for reminding me of this! so I adjust the sequence now (after the ITS ROF border)
image

> 2. Since we are adding a dependency on a library, and on a table, it is very likely that this development will imply a memory increase (and we are always at the limit of the memory for our analyses). I would ask you if you could evaluate the increase of memory, and possibly put this event selection as optional to allow not to have it where not needed (for instance in the D2H reduced data creators). Thanks!

No, I didn’t evaluate the memory increase, but I can do that later. The only changes I made are adding a flag for event rejection and including a few additional columns in the BC-joined table, and adding the library from UD for the task that used the `getHfCollisionRejectionMask`.

To make it flexible, I introduced a configurable `useUpcTrigger` flag—so if a task doesn’t require UPC-related selections, it can simply set this flag to false.

However, many tasks that are not interested in UPC might still use `getHfCollisionRejectionMask`. To avoid unnecessary dependencies in such cases, we could consider adding a templated function with the same name but without the BC-related parameter—essentially reverting to the previous behavior. That way, tasks that don’t need UPC can avoid compiling with the SGSelector library and the joined BC table.

Of course, if you have a smarter or cleaner solution in mind, I’m happy to hear it. Thanks a lot!

zhangbiao-phy:
> > > Hi @zhangbiao-phy many thanks for this development! Before a detailed review, I have two main comments:
> > > 
> > > 1. in the `hCollisions` histogram the `Zvtx` cut should always be at the end, in order to be able to evaluate the `Zvtx` cut efficiency
> > > 2. Since we are adding a dependency on a library, and on a table, it is very likely that this development will imply a memory increase (and we are always at the limit of the memory for our analyses). I would ask you if you could evaluate the increase of memory, and possibly put this event selection as optional to allow not to have it where not needed (for instance in the D2H reduced data creators). Thanks!
> > >    Hi @fgrosa, thanks for the comments!
> > > 3. in the `hCollisions` histogram the `Zvtx` cut should always be at the end, in order to be able to evaluate the `Zvtx` cut efficiency
> 
> > Ok, thanks for reminding me of this! so I adjust the sequence now (after the ITS ROF border) image
> 
> Hi @zhangbiao-phy, thanks! Actually I would move also the timerange cuts before (sorry that's not related to your PR, but I did not notice before in the PR that introduced them).
> 
> > > 2. Since we are adding a dependency on a library, and on a table, it is very likely that this development will imply a memory increase (and we are always at the limit of the memory for our analyses). I would ask you if you could evaluate the increase of memory, and possibly put this event selection as optional to allow not to have it where not needed (for instance in the D2H reduced data creators). Thanks!
> > 
> > 
> > No, I didn’t evaluate the memory increase, but I can do that later. The only changes I made are adding a flag for event rejection and including a few additional columns in the BC-joined table, and adding the library from UD for the task that used the `getHfCollisionRejectionMask`.
> > To make it flexible, I introduced a configurable `useUpcTrigger` flag—so if a task doesn’t require UPC-related selections, it can simply set this flag to false.
> > However, many tasks that are not interested in UPC might still use `getHfCollisionRejectionMask`. To avoid unnecessary dependencies in such cases, we could consider adding a templated function with the same name but without the BC-related parameter—essentially reverting to the previous behavior. That way, tasks that don’t need UPC can avoid compiling with the SGSelector library and the joined BC table.
> > Of course, if you have a smarter or cleaner solution in mind, I’m happy to hear it. Thanks a lot!
> 
> That would be ideal, thanks! In any case if you could do a quick memory test also for the `trackIndexSkimCreator` before and after your implementation it would be great to know how much it changes.

Hi @fgrosa, Ok, actually I also guess you mean move the vz and chi2 to the end before :). this is current version: 
image

For the evaluate the memory increase, here is the table for different case, as you can see there is no much difference before and after the implementation for the same processing `processTrigSel`.

image

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -132,11 +132,19 @@ struct HfTrackIndexSkimCreatorTagSelCollisions {
...
-  template <bool applyTrigSel, o2::hf_centrality::CentralityEstimator centEstimator, typename Col, typename BCs>
-  void selectCollision(const Col& collision, const BCs&)
+  template <bool applyTrigSel, bool applyUPCSel, o2::hf_centrality::CentralityEstimator centEstimator, typename Col, typename BCs>
+  void selectCollision(const Col& collision, const BCs& bcs)
   {
     float centrality = -1.;
-    const auto rejectionMask = hfEvSel.getHfCollisionRejectionMask<applyTrigSel, centEstimator, BCs>(collision, centrality, ccdb, registry);
+    uint32_t rejectionMask;
+
+    if (applyUPCSel) {
```
fgrosa, line 141:
```suggestion
    if constexpr (applyUPCSel) {
```

PWGHF/Utils/utilsEvSelHf.h
```diff
@@ -238,10 +283,121 @@ struct HfEventSelection : o2::framework::ConfigurableGroup {
   /// \param registry reference to the histogram registry needed for zorro
   /// \return bitmask with the event selection criteria not satisfied by the collision
   template <bool useEvSel, o2::hf_centrality::CentralityEstimator centEstimator, typename BCs, typename Coll>
-  uint16_t getHfCollisionRejectionMask(const Coll& collision, float& centrality, o2::framework::Service<o2::ccdb::BasicCCDBManager> const& ccdb, o2::framework::HistogramRegistry& registry)
+  uint32_t getHfCollisionRejectionMask(const Coll& collision, float& centrality, o2::framework::Service<o2::ccdb::BasicCCDBManager> const& ccdb, o2::framework::HistogramRegistry& registry)
```
fgrosa, line 286: Isn't it possible to templetize the function to have or not the UPC selection instead of having the code duplicated in two functions? Another possibility is to call this function in the one applying also the UPC selection to avoid code duplication.

fgrosa (changes requested):
Hi @zhangbiao-phy many thanks for considering my main comments! It looks good for me, I just have a couple of remaining detailed comments

PWGHF/TableProducer/trackIndexSkimCreator.cxx
zhangbiao-phy: Fixed

PWGHF/Utils/utilsEvSelHf.h
```diff
@@ -193,6 +223,21 @@ struct HfEventSelection : o2::framework::ConfigurableGroup {
   o2::framework::OutputObj<ZorroSummary> zorroSummary{"zorroSummary"};
   int currentRun{-1};
 
+  /// Set standard preselection gap trigger (values taken from UD group)
+  void setSGPreselection(SGCutParHolder& sgCuts)
```
vkucera, line 227: Why do you pass `sgCuts`?

vkucera: Actually, this might be a better place to call `sgCuts = SGCutParHolder()` since it might not always be needed.

PWGHF/Utils/utilsEvSelHf.h
```diff
@@ -341,11 +386,49 @@ struct HfEventSelection : o2::framework::ConfigurableGroup {
...
+  /// \brief Applies event selection with upc flag and bc para
+  /// \tparam useEvSel use information from the EvSel table
+  /// \tparam centEstimator centrality estimator
+  /// \param collision collision to test against the selection criteria
+  /// \param centrality collision centrality variable to be set in this function
+  /// \param ccdb ccdb service needed to retrieve the needed info for zorro
+  /// \param registry reference to the histogram registry needed for zorro
+  /// \return bitmask with the event selection criteria not satisfied by the collision
+  template <bool useEvSel, o2::hf_centrality::CentralityEstimator centEstimator, typename BCs, typename Coll>
+  uint32_t getHfCollisionRejectionMask(const Coll& collision, float& centrality, o2::framework::Service<o2::ccdb::BasicCCDBManager> const& ccdb, o2::framework::HistogramRegistry& registry, const BCs& bcs)
```
vkucera, line 398: What's the benefit of having two functions?

zhangbiao-phy: > Templetisation should be enough (see my comment above)

No, there is a parameter for bcs, we have to load it in the UPC selection.

zhangbiao-phy: > If you use a pointer which is `nullptr` by default, you only need one function.

Thanks @vkucera, which should be the best solution!

PWGHF/Utils/utilsEvSelHf.h
```diff
@@ -229,18 +274,20 @@ struct HfEventSelection : o2::framework::ConfigurableGroup {
...
   /// \param collision collision to test against the selection criteria
   /// \param centrality collision centrality variable to be set in this function
   /// \param ccdb ccdb service needed to retrieve the needed info for zorro
   /// \param registry reference to the histogram registry needed for zorro
   /// \return bitmask with the event selection criteria not satisfied by the collision
-  template <bool useEvSel, o2::hf_centrality::CentralityEstimator centEstimator, typename BCs, typename Coll>
-  uint16_t getHfCollisionRejectionMask(const Coll& collision, float& centrality, o2::framework::Service<o2::ccdb::BasicCCDBManager> const& ccdb, o2::framework::HistogramRegistry& registry)
+  template <bool useEvSel, bool useUpcTrigger, o2::hf_centrality::CentralityEstimator centEstimator, typename BCs, typename Coll>
+  uint32_t getHfCollisionRejectionMask(const Coll& collision, float& centrality, o2::framework::Service<o2::ccdb::BasicCCDBManager> const& ccdb, o2::framework::HistogramRegistry& registry,
+                                       const BCs* bcs = nullptr)
```
vkucera, line 288:
```suggestion
  template <bool useEvSel, o2::hf_centrality::CentralityEstimator centEstimator, typename BCs, typename Coll>
  uint32_t getHfCollisionRejectionMask(const Coll& collision, float& centrality, o2::framework::Service<o2::ccdb::BasicCCDBManager> const& ccdb, o2::framework::HistogramRegistry& registry,
                                       const BCs* bcs = nullptr)
```
This is wrong. `useUpcTrigger` is a configurable. You should not declare a template parameter which shadows the configurable.

PWGHF/Utils/utilsEvSelHf.h
```diff
@@ -229,18 +274,20 @@ struct HfEventSelection : o2::framework::ConfigurableGroup {
     addHistograms(registry);
   }
 
-  /// \brief Applies event selection.
+  /// \brief Applies event selection
   /// \tparam useEvSel use information from the EvSel table
+  /// \tparam useUpcTrigger add upc event selection
```
vkucera, line 279:
```suggestion
```

vkucera (changes requested):

PWGHF/Utils/utilsEvSelHf.h
zhangbiao-phy: fixed

PWGHF/D2H/TableProducer/dataCreatorCharmHadPiReduced.cxx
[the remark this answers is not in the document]
vkucera: It still appears at several places.

PWGHF/Utils/utilsEvSelHf.h
```diff
@@ -297,6 +344,28 @@ struct HfEventSelection : o2::framework::ConfigurableGroup {
...
+
+      if (useUpcTrigger) {
+        if (!bcs) {
+          throw std::runtime_error("UPC selection requested but no BC table provided!");
+        }
+        SGCutParHolder sgCuts = setSGPreselection();
+        auto bc = collision.template foundBC_as<BCs>();
+        auto bcRange = udhelpers::compatibleBCs(collision, sgCuts.NDtcoll(), *bcs, sgCuts.minNBCs());
+        auto isSGEvent = sgSelector.IsSelected(sgCuts, collision, bcRange, bc);
+        int issgevent = isSGEvent.value;
```
vkucera, line 356: Please do not define variables whose names just differ in capitalisation.

PWGHF/Utils/utilsEvSelHf.h
```diff
@@ -297,6 +344,28 @@ struct HfEventSelection : o2::framework::ConfigurableGroup {
...
+        if (issgevent > EventTypeUpc::DoubleGap) {
+          SETBIT(rejectionMask, EventRejection::UpcEventCut);
+        } else {
+          if (issgevent == EventTypeUpc::SingleGapA) {
+            hUPCollisions->Fill(EventTypeUpc::SingleGapA);
+          } else if (issgevent == EventTypeUpc::SingleGapC) {
+            hUPCollisions->Fill(EventTypeUpc::SingleGapC);
+          } else if (issgevent == EventTypeUpc::DoubleGap) {
+            hUPCollisions->Fill(EventTypeUpc::DoubleGap);
+          }
```
vkucera, line 366: Why don't you just fill the `issgevent` value?

PWGHF/Utils/utilsEvSelHf.h
vkucera: Not addressed.

PWGHF/Utils/utilsEvSelHf.h
zhangbiao-phy: good point

PWGHF/Utils/utilsEvSelHf.h
```diff
@@ -103,15 +105,25 @@ enum EventRejection {
...
+  PositionZ,
   NEventRejection
 };
 
+// upc event type
+enum EventTypeUpc {
+  SingleGapA,
+  SingleGapC,
+  DoubleGap,
+  NEventTypes
```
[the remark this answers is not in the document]
zhangbiao-phy: fixed

PWGHF/Utils/utilsEvSelHf.h
zhangbiao-phy: fixed

PWGHF/TableProducer/candidateCreator2Prong.cxx
zhangbiao-phy: > There should be an option to process Run 2 data.

I would not do that, since it will never be used from our side, and I didn't see it be used even in UD group.

PWGHF/Utils/utilsEvSelHf.h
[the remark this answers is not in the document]
zhangbiao-phy: I assign explicit value in the last commit, then it's the explain that why i only cover from 0 to 2: 

0: Single gap on A side

1: Single gap on C side

2: Double gap (both A and C sides are clean)

3: Non-double-gap (both A and C are noisy)

4: Event not within the required multiplicity range

**For our UPC selection, we are only interested in cases 0, 1, and 2.
That’s why I reject the event when` issgevent > EventTypeUpc::DoubleGap`.**

PWGHF/Utils/utilsEvSelHf.h
```diff
@@ -297,6 +344,22 @@ struct HfEventSelection : o2::framework::ConfigurableGroup {
...
+      if (useUpcTrigger) {
+        if (!bcs) {
+          throw std::runtime_error("UPC selection requested but no BC table provided!");
+        }
+        SGCutParHolder sgCuts = setSGPreselection();
+        auto bc = collision.template foundBC_as<BCs>();
+        auto bcRange = udhelpers::compatibleBCs(collision, sgCuts.NDtcoll(), *bcs, sgCuts.minNBCs());
+        auto sgSelectionResult = sgSelector.IsSelected(sgCuts, collision, bcRange, bc);
+        int issgevent = sgSelectionResult.value;
+        if (issgevent > EventTypeUpc::DoubleGap) {
```
vkucera, line 357: - Please use easily readable names that follow the naming conventions. The correct name would be `isSgEvent`.
- The name implies a `bool` value which is not the case.
- The variable is used only once which makes it useless and can be replaced with expression itself.

PWGHF/Utils/utilsEvSelHf.h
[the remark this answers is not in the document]
zhangbiao-phy: Although I think it's already fine since the `IsSelected` function from `SGSelector. h' always returns 0 for SingleGapA, 1 for SingleGapC, and 2 for DoubleGap, meaning this enum matches exactly. I added explicit values to make it clearer for others.

PWGHF/Utils/utilsEvSelHf.h
```diff
@@ -103,15 +105,25 @@ enum EventRejection {
...
   NoCollInTimeRangeNarrow,
   NoCollInTimeRangeStandard,
   NoCollInRofStandard,
+  UpcEventCut,
+  Chi2,
+  PositionZ,
   NEventRejection
 };
 
+// upc event type, refer from https://github.com/AliceO2Group/O2Physics/blob/master/PWGUD/Core/SGSelector.h
```
vkucera, line 117:
```suggestion
// upc event type, refer from PWGUD/Core/SGSelector.h
```

fgrosa (dismissed):
Hi @zhangbiao-phy, thanks for considering the comments and the tests. Good for me!

PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
```diff
@@ -442,9 +443,10 @@ struct HfTaskFlowCharmHadrons {
   /// Compute the scalar product
   /// \param collision is the collision with the Q vector information and event plane
   /// \param candidates are the selected candidates
-  template <DecayChannel channel, typename T1>
+  template <DecayChannel channel, typename T1, typename BCs>
```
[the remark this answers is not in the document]
zhangbiao-phy: I don't know why you say that? since the changes for that be recovered long time ago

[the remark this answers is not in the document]
zhangbiao-phy: you see this is what I saw 
image

zhangbiao-phy:
Hi @fgrosa @vkucera, To ensure that the changes in this PR do not prevent potential Run 2 analyses, I revised the approach by introducing a rejection function `getHfCollisionRejectionMaskWithUPC`, which retrieves the rejection mask from `getHfCollisionRejectionMask`. I also added UPC-specific processing functions to the 2-prong and 3-prong data creators, along with a UPC flag in their main functions (`runCreator2ProngWithDCAFitterN` and `runCreator2ProngWithKFParticle`). Additionally, a fatal message is triggered if `nProcessesCollisionsUPC > 1 && isRun2`.  Let  me know if you any comments

vkucera:
> Hi @fgrosa @vkucera, To ensure that the changes in this PR do not prevent potential Run 2 analyses, I revised the approach by introducing a rejection function `getHfCollisionRejectionMaskWithUPC`, which retrieves the rejection mask from `getHfCollisionRejectionMask`. I also added UPC-specific processing functions to the 2-prong and 3-prong data creators, along with a UPC flag in their main functions (`runCreator2ProngWithDCAFitterN` and `runCreator2ProngWithKFParticle`). Additionally, a fatal message is triggered if `nProcessesCollisionsUPC > 1 && isRun2`. Let me know if you any comments

Hi @zhangbiao-phy , thanks for adding the Run 2 support. One things I would like to suggest as a potentially better way to do it is to template the task struct on the `isRun2` switch and push the correct task instance into the topology based on metadata.

vkucera:
OK @zhangbiao-phy , thanks for the clarification.

PWGHF/TableProducer/candidateCreator2Prong.cxx
```diff
@@ -185,19 +190,24 @@ struct HfCandidateCreator2Prong {
     setLabelHistoCands(hCandidates);
   }
 
-  template <bool doPvRefit, o2::hf_centrality::CentralityEstimator centEstimator, typename Coll, typename CandType, typename TTracks>
+  template <bool doPvRefit, bool useUpcSel = false, o2::hf_centrality::CentralityEstimator centEstimator, typename Coll, typename CandType, typename TTracks, typename BCsType>
```
vkucera, line 193:
```suggestion
  template <bool doPvRefit, bool useUpcSel, o2::hf_centrality::CentralityEstimator centEstimator, typename Coll, typename CandType, typename TTracks, typename BCsType>
```
Don't use default template arguments, please. The behaviour should be obvious from the call.

fgrosa:
> @fgrosa There might be a misunderstanding. I never advocated for an implementation of Run 2 support for UPC+HF analyses. I advocated for preserving the Run 2 support for HF analyses. One way to do it is to duplicated the process functions, as @zhangbiao-phy did; another way is to template the struct, as I proposed.

@vkucera I see! Sorry I misunderstood from the conversation, all good then!

PWGHF/TableProducer/candidateCreator2Prong.cxx
```diff
@@ -138,6 +140,9 @@ struct HfCandidateCreator2Prong {
         LOGP(fatal, "Process function for collision monitoring not correctly enabled. Did you enable \"processCollisionsCentFT0M\"?");
       }
     }
+    if (nProcessesCollisionsUPC > 1 && isRun2) {
+      LOGP(fatal, "Process function for UPC is only available in Run 3!");
+    }
```
vkucera, line 145: Why ` > 1` and not ` > 0`?

PWGHF/TableProducer/candidateCreator2Prong.cxx
```diff
@@ -619,10 +634,72 @@ struct HfCandidateCreator2Prong {
...
   PROCESS_SWITCH(HfCandidateCreator2Prong, processNoPvRefitWithKFParticleCentFT0M, "Run candidate creator using KFParticle package w/o PV refit and w/ centrality selection on FT0M", false);
 
+  /////////////////////////////////////////////
+  ///                                       ///
+  ///   with centrality selection on UPC   ///
+  ///                                       ///
+  /////////////////////////////////////////////
+
+  /// @brief process function using DCA fitter w/ PV refit and w/ centrality selection on UPC
+  void processPvRefitWithDCAFitterNUPC(soa::Join<aod::Collisions, aod::EvSels> const& collisions,
```
vkucera, line 648:
```suggestion
  void processPvRefitWithDCAFitterNUpc(soa::Join<aod::Collisions, aod::EvSels> const& collisions,
```
and elsewhere.

PWGHF/TableProducer/candidateCreator3Prong.cxx
```diff
@@ -236,19 +241,21 @@ struct HfCandidateCreator3Prong {
     }
   }
 
-  template <bool doPvRefit = false, o2::hf_centrality::CentralityEstimator centEstimator, typename Coll, typename Cand>
-  void runCreator3ProngWithDCAFitterN(Coll const&,
-                                      Cand const& rowsTrackIndexProng3,
-                                      TracksWCovExtraPidPiKaPr const&,
-                                      aod::BCsWithTimestamps const& /*bcWithTimeStamps*/)
+  template <bool doPvRefit = false, bool useUpcSel = false, o2::hf_centrality::CentralityEstimator centEstimator, typename Coll, typename Cand, typename BCsType>
+  void runCreator3ProngWithDCAFitterN(Coll const&, Cand const& rowsTrackIndexProng3, TracksWCovExtraPidPiKaPr const&, BCsType const& bcs)
```
vkucera, line 245: Keep one argument per line.

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -132,11 +132,19 @@ struct HfTrackIndexSkimCreatorTagSelCollisions {
 
   /// Collision selection
   /// \param collision  collision table with
-  template <bool applyTrigSel, o2::hf_centrality::CentralityEstimator centEstimator, typename Col, typename BCs>
-  void selectCollision(const Col& collision, const BCs&)
+  template <bool applyTrigSel, bool applyUPCSel, o2::hf_centrality::CentralityEstimator centEstimator, typename Col, typename BCsType>
```
vkucera, line 135:
```suggestion
  template <bool applyTrigSel, bool applyUpcSel, o2::hf_centrality::CentralityEstimator centEstimator, typename Col, typename BCsType>
```

PWGHF/Utils/utilsEvSelHf.h
```diff
@@ -181,8 +197,9 @@ struct HfEventSelection : o2::framework::ConfigurableGroup {
   static constexpr char NameHistPosYAfterEvSel[] = "hPosYAfterEvSel";
   static constexpr char NameHistNumPvContributorsAfterSel[] = "hNumPvContributorsAfterSel";
   static constexpr char NameHistCollisionsCentOcc[] = "hCollisionsCentOcc";
+  static constexpr char NameHistUPC[] = "hUPCollisions";
```
vkucera, line 200:
```suggestion
  static constexpr char NameHistUpCollisions[] = "hUpCollisions";
```
Keep the correspondence.

PWGHF/Utils/utilsEvSelHf.h
```diff
@@ -193,6 +210,21 @@ struct HfEventSelection : o2::framework::ConfigurableGroup {
   o2::framework::OutputObj<ZorroSummary> zorroSummary{"zorroSummary"};
   int currentRun{-1};
 
+  /// Set standard preselection gap trigger (values taken from UD group)
+  SGCutParHolder setSGPreselection()
```
vkucera, line 214:
```suggestion
  SGCutParHolder setSgPreselection()
```

PWGHF/Utils/utilsEvSelHf.h
```diff
@@ -341,11 +373,32 @@ struct HfEventSelection : o2::framework::ConfigurableGroup {
     return rejectionMask;
   }
 
+  template <bool useEvSel, o2::hf_centrality::CentralityEstimator centEstimator, typename BCsType, typename Coll>
+  uint32_t getHfCollisionRejectionMaskWithUPC(const Coll& collision, float& centrality, o2::framework::Service<o2::ccdb::BasicCCDBManager> const& ccdb, o2::framework::HistogramRegistry& registry, const BCsType& bcs)
+  {
+    auto rejectionMaskWithUPC = getHfCollisionRejectionMask<true, centEstimator, BCsType>(collision, centrality, ccdb, registry);
```
vkucera, line 379:
```suggestion
    auto rejectionMaskWithUpc = getHfCollisionRejectionMask<true, centEstimator, BCsType>(collision, centrality, ccdb, registry);
```

vkucera (changes requested):
Please make sure that there is `Upc` instead of `UPC` used everywhere and that the switch to use the event selection is named the same everywhere.

zhangbiao-phy:
hi @vkucera, thanks for the comments! I fixed all of them. Could you take a look? Thanks!

PWGHF/TableProducer/candidateCreator3Prong.cxx
```diff
@@ -236,19 +241,24 @@ struct HfCandidateCreator3Prong {
     }
   }
 
-  template <bool doPvRefit = false, o2::hf_centrality::CentralityEstimator centEstimator, typename Coll, typename Cand>
+  template <bool doPvRefit = false, bool useUpcSel, o2::hf_centrality::CentralityEstimator centEstimator, typename Coll, typename Cand, typename BCsType>
```
vkucera, line 244:
```suggestion
  template <bool doPvRefit, bool useUpcSel, o2::hf_centrality::CentralityEstimator centEstimator, typename Coll, typename Cand, typename BCsType>
```

vkucera:
> hi @vkucera, thanks for the comments! I fixed all of them. Could you take a look? Thanks!

You still use `useUpcSel` and `applyUpcSel`.

zhangbiao-phy:
> > hi @vkucera, thanks for the comments! I fixed all of them. Could you take a look? Thanks!
> 
> You still use `useUpcSel` and `applyUpcSel`.

I didn't get your point before, I fixed it now

PWGHF/DataModel/CandidateReconstructionTables.h
```diff
@@ -39,6 +39,7 @@
 namespace o2::aod
 {
 // Table aliases
+using BCFullInfos = soa::Join<aod::BCsWithTimestamps, aod::BcSels, aod::Run3MatchedToBCSparse>;
```
vkucera, line 42:
```suggestion
using BcFullInfos = soa::Join<aod::BCsWithTimestamps, aod::BcSels, aod::Run3MatchedToBCSparse>;
```

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -155,46 +163,54 @@ struct HfTrackIndexSkimCreatorTagSelCollisions {
   }
 
   /// Event selection with trigger and FT0A centrality selection
-  void processTrigAndCentFT0ASel(soa::Join<aod::Collisions, aod::EvSels, aod::CentFT0As>::iterator const& collision, aod::BCsWithTimestamps const& bcs)
+  void
+    processTrigAndCentFT0ASel(soa::Join<aod::Collisions, aod::EvSels, aod::CentFT0As>::iterator const& collision, aod::BCFullInfos const& bcs)
```
vkucera, line 167: Don't break the line.

vkucera: Keep one parameter per line.

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -155,46 +163,54 @@ struct HfTrackIndexSkimCreatorTagSelCollisions {
...
+    selectCollision<false, false, CentralityEstimator::None>(collision, bcs);
   }
   PROCESS_SWITCH(HfTrackIndexSkimCreatorTagSelCollisions, processNoTrigSel, "Do not use trigger selection", true);
+
+  /// Event selection with UPC
+  void processUpcSel(soa::Join<aod::Collisions, aod::EvSels>::iterator const& collision, aod::BCFullInfos const& bcs, aod::FT0s const& /*ft0s*/, aod::FV0As const& /*fv0as*/, aod::FDDs const& /*fdds*/, aod::Zdcs const& /*zdcs*/)
+  {
+    selectCollision<true, true, CentralityEstimator::None>(collision, bcs);
+  }
+  PROCESS_SWITCH(HfTrackIndexSkimCreatorTagSelCollisions, processUpcSel, "Use upc selection", false);
```
vkucera, line 213:
```suggestion
  PROCESS_SWITCH(HfTrackIndexSkimCreatorTagSelCollisions, processUpcSel, "Use UPC event selection", false);
```

PWGHF/Utils/utilsEvSelHf.h
```diff
@@ -205,7 +229,7 @@ struct HfEventSelection : o2::framework::ConfigurableGroup {
     hPosYAfterEvSel = registry.add<TH1>(NameHistPosYAfterEvSel, "selected events;#it{y}_{prim. vtx.} (cm);entries", {o2::framework::HistType::kTH1D, {{200, -0.5, 0.5}}});
     hNumPvContributorsAfterSel = registry.add<TH1>(NameHistNumPvContributorsAfterSel, "selected events;#it{y}_{prim. vtx.} (cm);entries", {o2::framework::HistType::kTH1D, {{500, -0.5, 499.5}}});
     setEventRejectionLabels(hCollisions, softwareTrigger);
-
+    hUpCollisions = registry.add<TH1>(NameHistUpCollisions, "HF upc counter;;# of upc events", {o2::framework::HistType::kTH1D, {axisUpcEvents}});
```
vkucera, line 232:
```suggestion
    hUpCollisions = registry.add<TH1>(NameHistUpCollisions, "HF UPC counter;;# of UPC events", {o2::framework::HistType::kTH1D, {axisUpcEvents}});
```

zhangbiao-phy:
hi @vkucera, thanks for the comments! could you have a look now?

vkucera (approved):

fgrosa:
> @mfaggin @fgrosa Can you also check if it looks good to you?

Good for me! I would suggest to move on and merge

PR #11533, opened 2025-06-10
[PWGHF] Add Xic0 flow task and fix previous O2 linter issues

Rrantu:
This PR introduces the flow analysis task for the decay channel Xic⁰ → Ξ⁻ π⁺ in taskFlowCharmHadrons.cxx, and fixes previously existing O2 linter issues as follows:
1. Replaced  2. * TMath::Pi() with the constant o2::constants::math::TwoPI.
2. Used RecoDecay::constrainAngle to normalize angles within the proper range.
3. Reorganized struct members in HfTaskFlowCharmHadrons to follow pwghf/struct-member-order: 
    Moved HfHelper and SliceCache before using declarations and 
    Placed ConfigurableAxis after Partition.
5. Renamed variables (axes_reso → axesReso) to follow lowerCamelCase convention.

### DIFF
--- a/PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
+++ b/PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
@@ -663,7 +663,7 @@ struct HfTaskFlowCharmHadrons {
   {
     runFlowAnalysis<DecayChannel::DplusToPiKPi>(collision, candidatesDplus);
   }
-  PROCESS_SWITCH(HfTaskFlowCharmHadrons, processDplusMl, "Process Dplus candidates with ML", false);
+  PROCESS_SWITCH(HfTaskFlowCharmHadrons, processDplusMl, "Process Dplus candidates with ML", true);
 
   // Dplus with rectangular cuts
   void processDplus(CollsWithQvecs::iterator const& collision,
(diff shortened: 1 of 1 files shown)


PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx
```diff
@@ -285,19 +291,52 @@ struct HfTaskFlowCharmHadrons {
...
+  /// Compute the Q vector for the candidate's tracks
+  /// \param cand is the candidate
+  /// \param tracksQx is the X component of the Q vector for the tracks
+  /// \param tracksQy is the Y component of the Q vector for the tracks
+  /// \param channel is the decay channel
+  template <typename T1>
+  void getQvecXic0Tracks(const T1& cand,
+                         std::vector<float>& tracksQx,
+                         std::vector<float>& tracksQy,
+                         float& ampl)
```
vkucera, line 303: Why is `ampl` passed by non-const reference?

vkucera: Small types are cheap to copy so if you don't need to modify the variable, you should pass it by value.

zhangbiao-phy:
nice work! @Rrantu  Ok for me!

zhangbiao-phy (approved):

vkucera (approved):

PR #11540, opened 2025-06-10
[PWGLF] do not include ITS internal code

f3sch:
CI will check if this compiles.
I think it is not advisable to rely on some internal tracking data structures, we might internally do something different in the future which is not visible from the outside, rather rely on the core library.

### DIFF
--- a/PWGLF/TableProducer/Strangeness/Converters/CMakeLists.txt
+++ b/PWGLF/TableProducer/Strangeness/Converters/CMakeLists.txt
@@ -41,22 +41,22 @@ o2physics_add_dpl_workflow(straevselsconverter
 
 o2physics_add_dpl_workflow(straevselsconverter2
                     SOURCES straevselsconverter2.cxx
-                    PUBLIC_LINK_LIBRARIES O2Physics::AnalysisCore O2::ITStracking
+                    PUBLIC_LINK_LIBRARIES O2Physics::AnalysisCore O2::ReconstructionDataFormats
                     COMPONENT_NAME Analysis)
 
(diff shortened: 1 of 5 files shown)


romainschotter (changes requested):
Hi @f3sch ! Thanks a lot for the changes! I also took the opportunity to remove the filling with dummy values, when it is not necessary.
Could you please also repeat those changes for the converter task `straevselsconverter5`? You might need to update your branch first.
Thanks a lot!

romainschotter:
@f3sch I completely understand and respect that you would prefer to keep the PR limited to your own changes and not take on responsibility for code you are not directly involved with. That said, please keep in mind that ultimately it's Francesca's and my responsibility to review and approve the PR, so sometimes we may propose adjustments that align with broader maintenance or consistency goals.

In any case, I have removed my commits as requested.

One point I would still like to emphasize: it would be really helpful if you could propagate the changes you have made to `straevselsconverter2`, `straevselsconverter3`, and `straevselsconverter4` also to the `straevselsconverter5` task.
Thanks a lot for your work!

f3sch:
I don't disagree with you if it would have been within the scope of the changes proposed.

Thanks for pointing to straevselsconverter5.
Note I changed now the template type to some bogus type to unneeded avoid linkage.

romainschotter:
@f3sch I understand, no worries :-)
The PR looks very good! One last comment: can you please change `O2::ITStracking` to `O2::ReconstructionDataFormats` for `straevselsconverter5` in the CMakeLists.txt too? 
Thanks a lot for all the work!

romainschotter (approved):

PR #11541, opened 2025-06-10
[PWGLF] Add option to disable SV radius cut for kink

wang-yuanzhe:


fmazzasc:
@skundu692 could you approve this? thank you

PR #11544, opened 2025-06-11
[Trigger] Add macro to inspect trigger menu per run

mpuccio:


alibuild (approved):

PR #11556, opened 2025-06-11
Remove indexRexCharmBaryon=-1 case for MC rec matching

chengtt0406:
Main updates:
1) Remove the indexRexCharmBaryon=-1 case for MC rec matching
2) Code structure for alternate decay channels is modified, which makes easier to extend for further development


zhangbiao-phy:
hi @chengtt0406, Thanks! Looks good to me!

PR #11579, opened 2025-06-12
[PWGLF] Update inelGt.h

njacazio:


PWGLF/Utils/inelGt.h
```diff
@@ -19,6 +19,10 @@
 #ifndef PWGLF_UTILS_INELGT_H_
 #define PWGLF_UTILS_INELGT_H_
 
+#include "Framework/O2DatabasePDGPlugin.h"
```
vkucera, line 22: O2DatabasePDGPlugin is not used.

vkucera (commented):
The header still does not compile.

PR #11593, opened 2025-06-12
[PWGLF] added configurable enablePureDCAHistogram for enabling pure DCA histogram based on nsigmaTPCKa cut

BanajitBarman:


PWGLF/Tasks/Nuspex/spectraTOF.cxx
```diff
@@ -18,6 +18,7 @@
 ///
 
 // O2 includes
+// o2-linter: disable=all
```
vkucera, line 21: First, there is no such directive, second, you are not supposed to silence error reports instead of fixing the reported errors.

BanajitBarman: Thank I will keep in mind next time.

PR #11600, opened 2025-06-13
[PWGLF] removed unnecessary disable directives for o2linter

alcaliva:
@vkucera, I fixed the issue you mentioned in the previous PR here

### DIFF
--- a/PWGJE/Core/JetBkgSubUtils.cxx
+++ b/PWGJE/Core/JetBkgSubUtils.cxx
@@ -147,7 +147,7 @@ std::tuple<double, double> JetBkgSubUtils::estimateRhoPerpCone(const std::vector
   return std::make_tuple(perpPtDensity, perpMdDensity);
 }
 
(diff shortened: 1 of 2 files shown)


vkucera:
Thanks @alcaliva

romainschotter (approved):

alcaliva:
Hi @vkucera,

I agree, because that function cannot (and does not) modify the jet, so it should not take a non-const ref.
In the meanwhile, I would close this PR as the existing code works. 
Alberto

vkucera:
> Hi @vkucera,
> 
> I agree, because that function cannot (and does not) modify the jet, so it should not take a non-const ref. In the meanwhile, I would close this PR as the existing code works. Alberto

It could have stayed open as a draft until https://github.com/AliceO2Group/O2Physics/pull/11618 is merged.

alcaliva:
I decided to close it because the new PR was "hiding" behind this one and I wanted it to be more visible.
I will fix it once the other PR (https://github.com/AliceO2Group/O2Physics/pull/11618) is merged.

PR #11611, opened 2025-06-13
[PWGHF] runMassFitter: add main() function and missing header, provide Readme

lubynets:
1. The `main()` function is added as an entry point to enable compilation as CMake project.
2. Added missing header.

_[Update]_
3. Added `CMakeLists_HFInvMassFitter.txt` and `HFInvMassFitterLinkDef.h ` for fitter compilation as a standalone project.
4. Added Readme with instructions how to compile and run the fitter.

### DIFF
--- a/PWGHF/D2H/Macros/README.md
+++ b/PWGHF/D2H/Macros/README.md
@@ -1,5 +1,5 @@
-# Invariant mass fitter
-Invariant mass fitter is implemented in the class `HFInvMassFitter` (`HFInvMassFitter.cxx/h` files), and its run is executed via `runMassFitter.C` macro.
+# Invariant-mass fitter
+Invariant-mass fitter is implemented in the class `HFInvMassFitter` (`HFInvMassFitter.cxx/h` files), and its run is executed via `runMassFitter.C` macro.
 Fitter is configured in `config_massfitter.json`.\
 The fitter is **not** a part of O2Physics source code.
 


vkucera:
Hi @lubynets , what exactly are you trying to do?
ROOT macros have the `.C` extension, their entry point is a function with the same name as the file and they are supposed to be executed in the ROOT interpreter.
You just added a `main` function but the rest remains unchanged.

lubynets:
> @lubynets OK. I think the use of the fitter should be documented somewhere. Do we not have an existing document for this?

Hi @vkucera,
As far as I know there is no such a document. I'll create one and add as a README file in the same folder (it can be further extended with instructions for other macros), ok?
I could also share the CMakeLists.txt for the fitter, but is it acceptable from the point of O2Physics structure? (since the fitter is not a part of common compilation).

PWGHF/D2H/Macros/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+## How to run
+### As a ROOT macro
+The `runMassFitter.C` can be compiled as ROOT macro.
+```
+cd path-to-o2physics-src/PWGHF/D2H/Macros
+source path-to-root-install/bin/thisroot.sh
+export ROOT_INCLUDE_PATH=$ROOT_INCLUDE_PATH:path-to-json-include
+root -l -x -b -q "HFInvMassFitter.cxx" "runMassFitter.C(\"config_massfitter.json\")"
+```
```
vkucera, line 20: Please add the language specification in the fenced code blocks as reported by markdownlint.

vkucera (changes requested):

PWGHF/D2H/Macros/README.md
lubynets: > Please add the language specification in the fenced code blocks as reported by markdownlint.

I am not 100% sure that I interpreted your comment correctly, but I think that I addressed it.

PWGHF/D2H/Macros/README.md
```diff
@@ -0,0 +1,59 @@
+# Invariant mass fitter
```
vkucera, line 1:
```suggestion
# Invariant-mass fitter
```

PWGHF/D2H/Macros/README.md
```diff
@@ -0,0 +1,59 @@
...
+Invariant mass fitter is implemented in the class `HFInvMassFitter` (`HFInvMassFitter.cxx/h` files), and its run is executed via `runMassFitter.C` macro.
```
vkucera, line 2:
```suggestion
Invariant-mass fitter is implemented in the class `HFInvMassFitter` (`HFInvMassFitter.cxx/h` files), and its run is executed via `runMassFitter.C` macro.
```

vkucera (commented):
Thanks @lubynets , looks good to me.
Just please fix the minor typo.

PR #11615, opened 2025-06-13
Fix HfMlResponse features naming for PID Exp

lubynets:
In ML features names for N sigmas PID in case of certain prong's type expectation (i.e. assumption of specific order of prongs) the index of the prong is not needed, moreover it is misleading. E.g. in `Lc->PKPi `decay the proton can be either 0-th or 2-nd, and therefore the name `tpcNSigmaPrExpPr0` is not correct.

### DIFF
--- a/PWGHF/Core/HfMlResponseDsToKKPi.h
+++ b/PWGHF/Core/HfMlResponseDsToKKPi.h
@@ -16,10 +16,10 @@
 #ifndef PWGHF_CORE_HFMLRESPONSEDSTOKKPI_H_
 #define PWGHF_CORE_HFMLRESPONSEDSTOKKPI_H_
 
-#include <vector>
-
 #include "PWGHF/Core/HfHelper.h"
 #include "PWGHF/Core/HfMlResponse.h"
+
+#include <vector>
 
 // Fill the map of available input features
 // the key is the feature's name (std::string)
@@ -132,12 +132,12 @@ enum class InputFeaturesDsToKKPi : uint8_t {
   nSigTpcTofKa0,
   nSigTpcTofKa1,
   nSigTpcTofKa2,
-  nSigTpcKaExpKa0,
-  nSigTpcPiExpPi2,
-  nSigTofKaExpKa0,
-  nSigTofPiExpPi2,
-  nSigTpcTofKaExpKa0,
-  nSigTpcTofPiExpPi2,
+  nSigTpcKaExpKa,
+  nSigTpcPiExpPi,
+  nSigTofKaExpKa,
+  nSigTofPiExpPi,
+  nSigTpcTofKaExpKa,
+  nSigTpcTofPiExpPi,
   absCos3PiK,
   deltaMassPhi
 };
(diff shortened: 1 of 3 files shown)


mfaggin:
> Hi @mfaggin, You write "`ExpPr0` means that we decide to expect the proton PID for the prong with index 0", but isn't it just `tpcNSigmaPr0`? 

It is not the same thing. `tpcNSigmaPr0` simply gives you the N sigma for the proton hypothesis for the prong 0, but this does not grant anything about whether the preselections selected this daughter as proton or not. This is the detail that makes the difference with `tpcNSigmaPrExpPr0`. Trying to be more explicit:

- if you are filling the candidate as a `pKpi`, then `tpcNSigmaPr0` and `tpcNSigmaPrExpPr0` are the same thing, because at position 0 you are already considering your proton daughter
- if instead you are filling the candidate as a `piKp`, then `tpcNSigmaPrExpPr0` contains the value of `tpcNSigmaPr2`, but the suffix `0` for `tpcNSigmaPrExpPr0` reminds us that we are storing the info "expecting to have the proton daughter at index 0". That's why I think this is still useful to keep for code readability

> What you write in the second sentence is exactly what I mean (and it is somehow contradictory to what you wrote in the first sentence): `tpcNSigmaPrExpPr` can stand either for `tpcNSigmaPr0` or `tpcNSigmaPr2` depending on the prongs' order hypothesis: `CHECK_AND_FILL_VEC_LCTOPKPI_SIGNED(candidate, tpcNSigmaPrExpPr0, nSigTpcPr0, nSigTpcPr2);`

Still I do not understand which contradiction you are talking about.

lubynets:
> > Hi @mfaggin, You write "`ExpPr0` means that we decide to expect the proton PID for the prong with index 0", but isn't it just `tpcNSigmaPr0`?
> 
> It is not the same thing. `tpcNSigmaPr0` simply gives you the N sigma for the proton hypothesis for the prong 0, but this does not grant anything about whether the preselections selected this daughter as proton or not. This is the detail that makes the difference with `tpcNSigmaPrExpPr0`. Trying to be more explicit:
> 
>     * if you are filling the candidate as a `pKpi`, then `tpcNSigmaPr0` and `tpcNSigmaPrExpPr0` are the same thing, because at position 0 you are already considering your proton daughter
Fully agree.
> 
>     * if instead you are filling the candidate as a `piKp`, then `tpcNSigmaPrExpPr0` contains the value of `tpcNSigmaPr2`, 
Agree as well.
> but the suffix `0` for `tpcNSigmaPrExpPr0` reminds us that we are storing the info "expecting to have the proton daughter at index 0". That's why I think this is still useful to keep for code readability

I do not understand it at all. We just applied the `piKp` order and filled the variable with `tpcNSigmaPr2` value, why do we expect proton to be 0-th when it is 2-nd?
> 
> 
> > What you write in the second sentence is exactly what I mean (and it is somehow contradictory to what you wrote in the first sentence): `tpcNSigmaPrExpPr` can stand either for `tpcNSigmaPr0` or `tpcNSigmaPr2` depending on the prongs' order hypothesis: `CHECK_AND_FILL_VEC_LCTOPKPI_SIGNED(candidate, tpcNSigmaPrExpPr0, nSigTpcPr0, nSigTpcPr2);`
> 
> Still I do not understand which contradiction you are talking about.

mfaggin:
hi @lubynets no problem at all!
Indeed there is no use-case for `tpcNSigmaPrExpPr2`, as you properly understand. And I perfectly understand you point, I agree with everything you say (and we are really saying the same things). My point of the "better code readability" is indeed a detail if you want, not strictly necessary, but in my opinion it simply helps to remember what is the convention that we have chosen.

Being explicit: reading `tpcNSigmaPrExpPr0` one understands that:
 - for a `pKpi` candidate the proton nsigma of prong `0` is used
 - for a `piKp` candidate the proton nsigma of prong `2` is used
The suffix `0` in `tpcNSigmaPrExpPr0` simply tells us that by convention we decide to write our candidates all as `pKpi` (namely if it is a `piKp` the info is swapped) and one can always know a posteriori this knowing this convention (given by the suffix itself) and by knowing the selection flag `pKpi` or `piKp`. Nothing else.
No other utility indeed, but this still remains more understandable, at least to me.

lubynets:
Hi @mfaggin,
Ok, now I understand what you mean - that number `0` is not a "degree of freedom" in variable naming, but an inseparable attribute of `Pr`, and it serves for reminding that we name `Lc` decay in this order: `PrKaPi` (although the prongs' order in a certain candidate can be opposite).
In my personal opinion it does not serve for readability (e.g. myself I was mislead with `0` index in the end and had to double-check what variable means), but since it is not a matter of code logic (in which we agree), but a matter of naming, I would not insist on these changes.
Shall I close the PR?

mfaggin:
Hi @lubynets good that now we understood each other. Actually we both agree on what discussed, but indeed since it's not a matter of code logic I think the PR can be closed and we can leave the code as it is.

PR #11618, opened 2025-06-14
[PWGJE] Fix: make doRhoAreaSub take jet by const reference

alcaliva:
@vkucera


vkucera:
Thanks @alcaliva !
@hahassan7 FYI

nzardosh:
Just for my education, since these are already passed as reference with "&" why would they be copied? I thought "const" only made sure they were not modified

nzardosh (approved):

PR #11622, opened 2025-06-15
[Common] Change index column identifier

ddobrigk:
to avoid HY unbound indices complaint (actual index value was fine)


alibuild (approved):

alibuild (approved):

PR #11632, opened 2025-06-16
[PWGDQ] Add cent FT0A and FT0M

ypwangg:


PWGDQ/Core/VarManager.cxx
```diff
@@ -201,6 +201,10 @@ void VarManager::SetDefaultVarNames()
   fgVariableUnits[kCentVZERO] = "%";
   fgVariableNames[kCentFT0C] = "Centrality FT0C";
   fgVariableUnits[kCentFT0C] = "%";
+  fgVariableNames[kCentFT0A] = "Centrality FT0A";
+  fgVariableUnits[kCentFT0A] = "%";
+  fgVariableNames[kCentFT0M] = "Centrality FT0M";
+  fgVariableUnits[kCentFT0M] = "%";
```
iarsene, line 207: you need to add these variables also in the fgVarNamesMap, so that they can be used in JSON expressions

PR #11639, opened 2025-06-17
[Common] Prevent CCDB crash -> always captured

ddobrigk:
This adds the `ccdb->setFatalWhenNull(false);` setting to the ccdb in the `multCentTable`. All exceptions are properly caught in subsequent code and dealt with so no need to crash.


alibuild (approved):

alibuild (approved):

PR #11642, opened 2025-06-17
Fix preselection bitmap check

romainschotter:


alibuild (approved):

PR #11654, opened 2025-06-17
[Common,PWGHF,PWGJE,Trigger,Tutorial] Drop support for old ONNXRuntime API

ktf:


ktf:
@singiamtel this one is also very slow to be picked up.

PR #11655, opened 2025-06-17
[PWGHF] Add exceptions for HF MC TOF recalibration to avoid human mistakes

fgrosa:


mfaggin (approved):
Many thanks @fgrosa !

PR #11657, opened 2025-06-17
[Common] Enclose Metadata and CollisionType Helper in o2 namespace

njacazio:


vkucera:
@njacazio @ddobrigk Missing include reported by MegaLinter was not fixed.
Please make sure that common headers follow IWYU to avoid nasty header dependencies.

PR #11669, opened 2025-06-18
Add missing metadata keys

fgrosa:


Common/Core/MetadataHelper.cxx
```diff
@@ -26,7 +26,10 @@ MetadataHelper::MetadataHelper()
                                               "RecoPassName",
                                               "Run",
                                               "AnchorPassName",
-                                              "AnchorProduction"};
+                                              "AnchorProduction",
+                                              "ROOTVersion",
+                                              "LPMProductionTag",
+                                              "O2Version"};
```
sawenzel, line 32: Does this mean that we need to hardcode all possible meta-data keys of an AO2D here? So far, it was not possible to read the LPMProductionTag from the meta section?

fgrosa: @sawenzel I think actually that is a good idea, should I try to implement it in this PR? Thanks!

ddobrigk (approved):

PR #11682, opened 2025-06-19
[PWGHF] Hard-code values of decay channels

vkucera:
This allows to preserve backward compatibility when extending the lists with more channels while keeping them sorted by species.


mfaggin:
apologize @vkucera I have a question: why did you decide to put `NChannels` always equal to the last one? I conceived it always as a counter, but in this way it is not such anymore

PR #11690, opened 2025-06-19
[PWGCF] add bool for pT ordering cut

Luzhiyongg:


Luzhiyongg:
@victor-gonzalez Hi Victor, I'd like to ask for a quick approval for this PR beacuse I want to catch up with the auto-submission of LHC23zzh_pass5 tomorrow.

Luzhiyongg:
Also, may I ask a question for the two particle correlation.
For the same event, I have no doubt that the pT ordering (trigger pT <= associated pT) should be applied to avoid double counting.
But for mixed events, I was wondering this is still the case since we are using "strictly upper pair" event mixing.

victor-gonzalez (approved):

victor-gonzalez:
> Also, may I ask a question for the two particle correlation. For the same event, I have no doubt that the pT ordering (trigger pT <= associated pT) should be applied to avoid double counting. But for mixed events, I was wondering this is still the case since we are using "strictly upper pair" event mixing.

Not quite sure about your question
I only see you are applying a pT ordering request which to me it is not the same as trigger pT < associated pT
If you apply the pT ordering to a mixed selecting just the upper pairs it looks to me that this is a kind of incompatible concurrent conditions: you select half of the combinations of which you will, in average, select an additional half. It sounds not good

PR #11700, opened 2025-06-20
[PWGDQ] Fix for invalid cursor call with 8 arguments instead of 9

aalkin:
The O2 PR https://github.com/AliceO2Group/AliceO2/pull/14421 adds sanity check for the table filling cursor and it fails in `PWGDQ/Tasks/dqEfficiency_withAssoc.cxx`:2080 - the filling call has 8 arguments while the table has 9 columns. From the context I assume that the centrality value was missing. Could you please verify that this is a correct fix? Thank you.


aalkin:
@mcoquet642 you also need to approve it before enabling the auto-merge. Thanks!

mcoquet642 (approved):

PR #11709, opened 2025-06-20
[PWGHF] Fix bug in D+ selector, due to wrong logic for correlated bkgs

fgrosa:
Thanks @Marcellocosti for spotting it!


alibuild (approved):

PR #11711, opened 2025-06-20
[PWGCF] removed obsolete variables and added new variables to process

hernasab:


PWGCF/Flow/Tasks/flowZdcTask.cxx
```diff
@@ -676,12 +662,88 @@ struct FlowZdcTask {
...
+      } else {
+        skipEvent = true;
+      }
+    }
+    // Skip event based on number of Nch sigmas
+    if (!skipEvent) {
+      return;
+    }
+    std::vector<float> pTs;
+    std::vector<float> vecFD;
```
victor-gonzalez, line 712: Any reason for variable definition without any usage?
@vkucera Should not this be captured by the compiler? or not any longer?

victor-gonzalez (approved):
I would suggest removing not used variables

PWGCF/Flow/Tasks/flowZdcTask.cxx
vkucera: Yes, it should. No idea why it is not the case here.

vkucera: Seems like the compiler only complains about trivial types. See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=55203

PR #11719, opened 2025-06-21
[PWGLF/Strangeness] add possibility to accept ITSsa tracks in strangenessbuilder.cxx

dsekihat:
PWGLF/Strangeness: add possibility to accept ITSsa tracks in strangenessbuilder.cxx

### DIFF
--- a/PWGLF/TableProducer/Strangeness/strangenessbuilder.cxx
+++ b/PWGLF/TableProducer/Strangeness/strangenessbuilder.cxx
@@ -318,6 +318,7 @@ struct StrangenessBuilder {
     Configurable<float> dcav0dau{"dcav0dau", 1.0, "DCA V0 Daughters"};
     Configurable<float> v0radius{"v0radius", 0.9, "v0radius"};
     Configurable<float> maxDaughterEta{"maxDaughterEta", 5.0, "Maximum daughter eta (in abs value)"};
+    Configurable<bool> acceptITSonly{"acceptITSonly", false, "flag to accept ITSonly tracks"};
 
     // MC builder options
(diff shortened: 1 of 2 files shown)


ddobrigk:
Hi @dsekihat, thanks, I agree that ITS-only tracks are extremely interesting to have here, and I have in the past also played with this concept (see slides 26+ of [this talk](https://www.dropbox.com/scl/fi/31b24wq50up9jrk99up1b/DDChinellato-2023-StrangenessReconstructionInO2-02.pdf?rlkey=qq85kaqbuifz7wwdvh2un5eto&st=5ce6nl4o&dl=0)). However, unless I am mistaken, your PR adds a boolean that is entirely equivalent to setting the existing variable `minCrossedRows` to `-1`, and if so I would stick to just using that alternative instead. I've also provided a core service that has `minCrossedRows=-1` for you to use: [check it out here](https://alimonitor.cern.ch/hyperloop/view-wagon/31133/wagon-settings). 

In actuality, if we see that accepting all ITS-only V0s consumes too much CPU or disk (for derived data), we could also investigate the option of having the TPC requirement removed not for all tracks but only for meson daughters of hyperons. This existed in the [previous generation of builders](https://github.com/AliceO2Group/O2Physics/blob/b1c4a26551040f0adad1989647d5b7e0bc90c7c9/PWGLF/TableProducer/Strangeness/lambdakzerobuilder.cxx#L1377), but hasn't been ported to the new `strangenessbuilder` yet; note, in any case, that the 'fully open' selection of `minCrossedRows=-1` includes this as a _subset_, of course.

(Tagging also coordinators for completeness @romainschotter @ercolessi)

dsekihat:
Hi David,indeed you are right. I close my PR.

PR #11726, opened 2025-06-22
[PWGHF] Cut variation - enhance plotting features, minor code cleaning and chore

lubynets:
Critical issues (preventing from script running) fixed:
- module `utils.style_formatter` replaced with `style_formatter` as it appears in the script folder
- `itemset` replaced with `[...] = ...` assignment in order to be compatible with `NumPy 2.x`

Not critical issues
- added check for not being a **nullptr** for histograms read from file
- reduced code duplication when printing histograms into pdf
- get rid of warnings about having histograms and canvases with the same string name

Plot enhancement
 - added text line with pT bin number and ranges to each pdf with plots

### DIFF
--- a/PWGHF/D2H/Macros/compute_fraction_cutvar.py
+++ b/PWGHF/D2H/Macros/compute_fraction_cutvar.py
@@ -155,25 +155,34 @@ def main(config):
             hist_frac_raw_nonprompt.SetBinContent(ipt + 1, raw_frac_nonprompt[0])
             hist_frac_raw_nonprompt.SetBinError(ipt + 1, raw_frac_nonprompt[1])
 
-        canv_rawy, histos_rawy, leg_r = minimiser.plot_result(f"_pt{pt_min:.0f}_{pt_max:.0f}")
+        hist_bin_title = ""
+        hist_bin_name = cfg.get("rawyields", {}).get("binwisehistoname")
+        if hist_bin_name is not None:
+            infile_rawy = ROOT.TFile.Open(os.path.join(cfg["rawyields"]["inputdir"], filename_rawy))
+            hist_bin_name = cfg["rawyields"]["binwisehistoname"] + str(ipt)
+            hist_bin = infile_rawy.Get(hist_bin_name)
+            hist_bin_title = hist_bin.GetTitle()
+            infile_rawy.Close()
+
+        canv_rawy, histos_rawy, leg_r = minimiser.plot_result(f"_pt{pt_min:.0f}_{pt_max:.0f}", hist_bin_title)
         output.cd()
         canv_rawy.Write()
         for _, hist in histos_rawy.items():
             hist.Write()
 
-        canv_eff, histos_eff, leg_e = minimiser.plot_efficiencies(f"_pt{pt_min:.0f}_{pt_max:.0f}")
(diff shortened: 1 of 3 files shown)


PWGHF/D2H/Macros/compute_fraction_cutvar.py
```diff
@@ -27,29 +27,45 @@ def main(config):
...
-        hist_effp.append(infile_eff.Get(cfg["efficiencies"]["histonames"]["prompt"]))
-        hist_effnp.append(infile_eff.Get(cfg["efficiencies"]["histonames"]["nonprompt"]))
+        hist_effp_name = cfg["efficiencies"]["histonames"]["prompt"]
+        hist_effnp_name = cfg["efficiencies"]["histonames"]["nonprompt"]
+        hist_effp.append(infile_eff.Get(hist_effp_name))
+        hist_effnp.append(infile_eff.Get(hist_effnp_name))
+        if(hist_effp[-1] is None):
+            sys.exit(f"Fatal error: Histogram with efficiency for prompt \"{hist_effp_name}\" is absent. Exit.")
+        if(hist_effnp[-1] is None):
+            sys.exit(f"Fatal error: Histogram with efficiency for nonprompt \"{hist_effnp}\" is absent. Exit.")
```
fgrosa, line 53: Maybe you can check also that the length of the raw yield and efficiency lists is the same?

PWGHF/D2H/Macros/compute_fraction_cutvar.py
```diff
@@ -155,30 +172,32 @@ def main(config):
             hist_frac_raw_nonprompt.SetBinContent(ipt + 1, raw_frac_nonprompt[0])
             hist_frac_raw_nonprompt.SetBinError(ipt + 1, raw_frac_nonprompt[1])
 
-        canv_rawy, histos_rawy, leg_r = minimiser.plot_result(f"_pt{pt_min:.0f}_{pt_max:.0f}")
+        hist_bin_title = "bin # " + str(ipt+1) + f"; {pt_axis_title}#in (" + str(pt_min) + "; " + str(pt_max) + ")"
```
fgrosa, line 175: I would suggest to use a single f-string (easier to read :-))
```suggestion
        hist_bin_title = f"bin # {ipt+1}; {pt_axis_title}#in ({pt_min:.0f}; {pt_max:.0f})"
```

fgrosa (changes requested):
Hi @lubynets thanks! I have just two minor comments before merging, see below

PWGHF/D2H/Macros/compute_fraction_cutvar.py
lubynets: Done

PWGHF/D2H/Macros/compute_fraction_cutvar.py
lubynets: As far as I understand it happens here https://github.com/AliceO2Group/O2Physics/blob/e501845932879c7295e35ac2114b90df03626b40/PWGHF/D2H/Macros/cut_variation.py#L76

lubynets: I now see that you meant another check. However it does not work, the point is that `hist_rawy`, `hist_effp` and `hist_effnp` lists by construction have the same size, and if in the config file the user will place different number of rows for yields and efficiencies, the smaller number of them will be processed.

PR #11729, opened 2025-06-23
PWGLF/Tasks/Nuspex/spectraTOF.cxx : Bug Fixes

RD0407:
Fix for Invalid MC collision index


maciacco:
hi @RD0407 you are missing a `std::` prefix in front of `pow` at line 1209 as pointed out by o2-linter. Please also check the many hard coded PDG codes, for readability it would be better to have them fetched from `PDG_t::` (you can check this later).

PR #11731, opened 2025-06-23
[Common,PWGUD,PWGLF,PWGHF] DPL Analysis: fix for reworked column getters

aalkin:
Due to changes in how the column getters work made in https://github.com/AliceO2Group/AliceO2/pull/14421 the single-valued columns return an rvalue, while fixed- and variable-size array columns return a span. This requires some small changed throughout O2Physics where these kinds of columns are accessed.

### DIFF
--- a/Common/TableProducer/eventSelection.cxx
+++ b/Common/TableProducer/eventSelection.cxx
@@ -1363,7 +1363,7 @@ struct LumiTask {
     const char* srun = Form("%d", run);
 
     for (const auto& bc : bcs) {
-      auto& selection = bc.selection_raw();
+      auto selection = bc.selection_raw();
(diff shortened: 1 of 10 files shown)


vkucera (approved):

rolavick (dismissed):

vkucera:
> @vkucera @ddobrigk thank you for your approval, but I'm still collecting the changes. There are some unexpected uses of the old API that are not always obvious how to fix. There will be more commits.

Thanks @aalkin for the clarification. I would keep the PR as a draft then until all commits are pushed.

aalkin:
The amount of changes required is too large for a one-off PR. It will be done in a smaller, manageable steps.

PR #11738, opened 2025-06-23
[PWGCF] Update meanPtFlucId.cxx

TGahlaut1:
Modified the PID selection task to have better purity.


PWGCF/EbyEFluctuations/Tasks/meanPtFlucId.cxx
```diff
@@ -455,50 +452,63 @@ struct MeanPtFlucId {
     return false;
   }
 
-  template <typename T>
-  PIDType identifyParticle(const T& track)
+  template <PIDType s1, PIDType s2, PIDType s3, typename T>
```
victor-gonzalez, line 455: Why is it needed to template the function with `s1`, `s2`, and `s3`? 
Everything is known at code writing time
Cannot them be passed as usual function arguments?

PR #11764, opened 2025-06-25
[PWGHF] Add omegac0 xic0 to omega ka framework reconstructed by kf

FDUEnrich:


PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx
```diff
@@ -118,10 +123,13 @@ struct HfCandidateCreatorXic0Omegac0 {
   Configurable<float> lambdaMassWindow{"lambdaMassWindow", 0.0075, "Distance from Lambda mass"};
   // cascade cuts
   Configurable<double> massToleranceCascade{"massToleranceCascade", 0.01, "Invariant mass tolerance for cascade"};
+  Configurable<double> massToleranceCascadeRej{"massToleranceCascadeRej", 0.01, "Invariant mass tolerance for rejected Xi"};
   // for KF particle operation
   Configurable<int> kfConstructMethod{"kfConstructMethod", 2, "KF Construct Method"};
   Configurable<bool> kfUseV0MassConstraint{"kfUseV0MassConstraint", false, "KF: use Lambda mass constraint"};
   Configurable<bool> kfUseCascadeMassConstraint{"kfUseCascadeMassConstraint", false, "KF: use Cascade mass constraint"};
+  Configurable<bool> constrainToPv{"constrainToPv", false, "Constrain particles to PV"};
+  Configurable<bool> constrainToMother{"constrainToMother", false, "Constrain daughter particles to mother particles"};
```
fgrosa, line 132: specify that these are configurables for the KF, as the previous ones (both in the name of the variable, and in the comment)
You might even consider to add `configurableGroups` to make it more organised, see e.g.  
https://github.com/AliceO2Group/O2Physics/blob/0882ed52c194504c30aea83b48ca377ac8fdba65/PWGHF/D2H/TableProducer/dataCreatorCharmResoReduced.cxx#L130-L136

PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx
```diff
@@ -1655,6 +1663,334 @@ struct HfCandidateCreatorXic0Omegac0 {
...
+      }
+      auto casc = cascAodElement.kfCascData_as<MyKfCascTable>();
+      hCascadesCounter->Fill(1);
+
+      // convert KaonFromCharm&KaFromOmega&V0DauPos&V0DauNeg tracks into KFParticle object
+      auto trackKaFromCharm = cand.prong0_as<MyKfTracks>();
+      auto trackKaFromOmega = casc.bachelor_as<MyKfTracks>(); // Ka <- Omega track
+      auto trackV0DauPos = casc.posTrack_as<MyKfTracks>();    // V0 positive daughter track
+      auto trackV0DauNeg = casc.negTrack_as<MyKfTracks>();    // V0 negative daughter track
+      auto KaFromOmegaCharge = trackKaFromOmega.signed1Pt() > 0 ? +1 : -1;
```
fgrosa, line 1715: Why can't you use the `trackKaFromOmega.sign()` directly?

PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx
```diff
@@ -1655,6 +1663,334 @@ struct HfCandidateCreatorXic0Omegac0 {
...
+      int signOmega = casc.sign() > 0 ? +1 : -1;
```
fgrosa, line 1716: Why can't you use `casc.sign()` directly?

fgrosa (changes requested):
Hi @FDUEnrich thanks for the implementation! As a general comment, I think it would have been beneficial to have a single process function for Omega+Pi and Omega+Ka (and also for Xi and Omega actually), where only the PID selection and the mass hypothesis is different for the bachelor, to avoid a large amount of code duplication, which is much more difficult to maintain. However since the other cases are already split we can keep this strategy (but it would be good to improve this sooner or later).
In the following you can see my detailed comments to your additionss

PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx
```diff
@@ -1655,6 +1663,334 @@ struct HfCandidateCreatorXic0Omegac0 {
...
+        continue;
+      }
+      auto casc = cascAodElement.kfCascData_as<MyKfCascTable>();
+      hCascadesCounter->Fill(1);
+
+      // convert KaonFromCharm&KaFromOmega&V0DauPos&V0DauNeg tracks into KFParticle object
+      auto trackKaFromCharm = cand.prong0_as<MyKfTracks>();
+      auto trackKaFromOmega = casc.bachelor_as<MyKfTracks>(); // Ka <- Omega track
+      auto trackV0DauPos = casc.posTrack_as<MyKfTracks>();    // V0 positive daughter track
+      auto trackV0DauNeg = casc.negTrack_as<MyKfTracks>();    // V0 negative daughter track
```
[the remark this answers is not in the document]
FDUEnrich: OK, let me check. Could you please give me a example? Because both OmegacToOmegaPi and XicToXiPiPi used this kind of table.

PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx
fgrosa: Thanks! You can anyway consider to use the configurable groups (they are also very useful to reduce the number of data members and avoid hitting the limit ;-)
It can also be done later in a second PR

PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx
fgrosa: @FDUEnrich did you check if you can use `trackKaFromOmega.sign()` directly instead of defining a new variable?

PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx
fgrosa: same as above

PWGHF/TableProducer/candidateSelectorOmegac0Xic0ToOmegaKa.cxx
```diff
@@ -0,0 +1,766 @@
...
+  Configurable<int> nClustersItsMin{"nClustersItsMin", 3, "Minimum number of ITS clusters requirement for pi <- charm baryon"};
+  Configurable<int> nClustersItsInnBarrMin{"nClustersItsInnBarrMin", 1, "Minimum number of ITS clusters in inner barrel requirement for pi <- charm baryon"};
+  Configurable<float> itsChi2PerClusterMax{"itsChi2PerClusterMax", 36, "Maximum value of chi2 fit over ITS clusters for pi <- charm baryon"};
+
+  ConfigurableAxis thnConfigAxisMass{"thnConfigAxisMass", {120, 2.4, 3.1}, "Cand. inv-mass bins"};
+  ConfigurableAxis thnConfigAxisPt{"thnConfigAxisPt", {100, 0, 20}, "Cand. pT bins"};
+  ConfigurableAxis thnConfigAxisCent{"thnConfigAxisCent", {100, 0, 100}, "Centrality bins"};
+  ConfigurableAxis thnConfigAxisPtKaon{"thnConfigAxisPtKaon", {100, 0, 10}, "PtPion from Omegac0 bins"};
+
+  struct : ConfigurableGroup {
```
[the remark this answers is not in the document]
fgrosa: Thanks for addressing this! However you added only the `pT Ka from Omegac`, while all the cuts in the configurable group should be added

PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx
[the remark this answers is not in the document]
fgrosa: This is done correctly in the process function with DCAFitter, see:
https://github.com/AliceO2Group/O2Physics/blob/4e484f3d6dab0517e5a3a0e1074d2a5d8e7a0a9b/PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx#L382-L383
In particular `lfTracks` are
https://github.com/AliceO2Group/O2Physics/blob/4e484f3d6dab0517e5a3a0e1074d2a5d8e7a0a9b/PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx#L147
All the process functions using the KF instead are wrong and must be fixed (tagging also the other analysers @Rrantu @Tao-Fang @xpeng)

PWGHF/TableProducer/candidateSelectorOmegac0Xic0ToOmegaKa.cxx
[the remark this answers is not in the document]
fgrosa: Ok for this PR! However, please add all the other cuts in a subsequent PR in the matrix, that allows for more neat configuration (also on hyperloop)
Maybe it can be done in @Rrantu PR https://github.com/AliceO2Group/O2Physics/pull/11860

PR #11765, opened 2025-06-25
Add dedicated light-ion GFW task

EmilGorm:


PWGCF/GenericFramework/Tasks/flowGFWLightIons.cxx
```diff
@@ -0,0 +1,1196 @@
...
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+/// \file flowGFWLightIons.cxx
```
victor-gonzalez, line 12: Is there anything tied to light ions in here that enforces to have that name?
I mean, cannot be the same code be run on PbPb or either pp? Perhaps with a reduced set of configurables for fine tuning?

EmilGorm: In principle yes, it can run on the pbpb code. But in case special considerations are needed, I feel not overloading the code with too many functionalities will make it easier to adapt in case any special requirements arise for the light ion data.

PR #11767, opened 2025-06-25
Dstar spin alignment in PbPb

Mingze129:
Adding new process for  Dstar spin alignment analysis in PbPb to CharmPolarisation task.

### DIFF
--- a/PWGHF/D2H/Tasks/taskCharmPolarisation.cxx
+++ b/PWGHF/D2H/Tasks/taskCharmPolarisation.cxx
@@ -15,6 +15,7 @@
 /// \author F. Grosa (CERN) fabrizio.grosa@cern.ch
 /// \author S. Kundu (CERN) sourav.kundu@cern.ch
 /// \author M. Faggin (CERN) mattia.faggin@cern.ch
+/// \author M. Li (CCNU) mingze.li@cern.ch
 
 #include <vector>
 
@@ -27,9 +28,10 @@
 #include "Framework/HistogramRegistry.h"
 #include "Framework/runDataProcessing.h"
 
-// #include "Common/Core/EventPlaneHelper.h"
-// #include "Common/DataModel/Qvectors.h"
+#include "Common/Core/EventPlaneHelper.h"
+#include "Common/DataModel/Qvectors.h"
 
+#include "PWGHF/Core/CentralityEstimation.h"
 #include "PWGHF/Core/HfHelper.h"
 #include "PWGHF/DataModel/CandidateSelectionTables.h"
 #include "PWGHF/DataModel/CandidateReconstructionTables.h"
@@ -48,6 +50,7 @@ enum CosThetaStarType : uint8_t {
   Production,
   Beam,
   Random,
+  EP,
   NTypes
 };
 enum DecayChannel : uint8_t {
@@ -60,6 +63,15 @@ enum MassHyposLcToPKPi : uint8_t {
   PKPi = 0,
   PiKP,
   NMassHypoLcToPKPi
+};
+enum QvecEstimator : uint8_t {
+  FV0A = 0,
+  FT0M,
+  FT0A,
+  FT0C,
+  TPCPos,
+  TPCNeg,
+  TPCTot
 };
 
 /// columns for table to study the Lc->PKPi background
@@ -120,6 +132,12 @@ struct TaskPolarisationCharmHadrons {
   Configurable<bool> selectionFlagDstarToD0Pi{"selectionFlagDstarToD0Pi", true, "Selection Flag for D* decay to D0 Pi"};
   Configurable<int> selectionFlagLcToPKPi{"selectionFlagLcToPKPi", 1, "Selection Flag for Lc decay to P K Pi"};
 
+  Configurable<int> harmonic{"harmonic", 2, "harmonic number"};
+  Configurable<int> qvecDetector{"qvecDetector", 3, "Detector for Q vector estimation (FV0A: 0, FT0M: 1, FT0A: 2, FT0C: 3, TPC Pos: 4, TPC Neg: 5, TPC Tot: 6)"};
+  Configurable<int> centEstimator{"centEstimator", 2, "Centrality estimator ((None: 0, FT0A: 1, FT0C: 2, FT0M: 3, FV0A: 4))"};
+  Configurable<int> minCent{"minCent", 0, "Minimum centrality (0-100) to be considered in the analysis"};
+  Configurable<int> maxCent{"maxCent", 100, "Maximum centrality (0-100) to be considered in the analysis"};
+
   ConfigurableAxis configThnAxisInvMass{"configThnAxisInvMass", {200, 0.139f, 0.179f}, "#it{M} (GeV/#it{c}^{2})"};
   ConfigurableAxis configThnAxisPt{"configThnAxisPt", {100, 0.f, 100.f}, "#it{p}_{T} (GeV/#it{c})"};
   ConfigurableAxis configThnAxisY{"configThnAxisY", {20, -1.f, 1.f}, "#it{y}"};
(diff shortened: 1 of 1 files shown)


PWGHF/D2H/Tasks/taskCharmPolarisation.cxx
```diff
@@ -520,10 +548,66 @@ struct TaskPolarisationCharmHadrons {
           registry.add("hRecoNonPromptRandom", "THn for polarisation studies with cosThStar w.r.t. random axis -- reco non-prompt signal", HistType::kTHnSparseF, {thnAxisInvMass, thnAxisPt, thnAxisNumPvContributors, thnAxisY, thnAxisInvMassKPiLc, thnAxisCosThetaStarRandom, thnAxisResoChannelLc, thnAxisAbsEtaTrackMin, thnAxisNumItsClsMin, thnAxisNumTpcClsMin, thnAxisIsRotatedCandidate, thnAxisCharge});
         }
       }
+    } else if (doprocessDstarInPbPb || doprocessDstarMcInPbPb) {
```
fgrosa, line 551: It would be good to preserve all the pp options also in Pb-Pb, but I agree that we should remove the axis for the rotational background to reduce the size. You could consider to define a `vector` of `AxisSpec` and add the axis for the rotational background only if it is not Pb-Pb, similarly to this:
https://github.com/AliceO2Group/O2Physics/blob/d7e0aab14e2fea3ced2893851fee3577b557c612/PWGHF/D2H/Tasks/taskFlowCharmHadrons.cxx#L175-L190
Moreover if EP is activated, you should throw a `FATAL` error

PWGHF/D2H/Tasks/taskCharmPolarisation.cxx
```diff
@@ -520,10 +548,66 @@ struct TaskPolarisationCharmHadrons {
...
+      if (doprocessDstarInPbPb) {
+        /// analysis for D*+ meson in Pb-Pb collisions, w/o rot. background axis
+        if (activateTHnSparseCosThStarHelicity) {
+          registry.add("hHelicity", "THn for polarisation studies with cosThStar w.r.t. helicity axis in Pb-Pb", HistType::kTHnSparseF, {thnAxisInvMass, thnAxisPt, thnAxisNumPvContributors, thnAxisY, thnAxisInvMassD0, thnAxisCosThetaStarHelicity, thnAxisAbsEtaTrackMin, thnAxisNumItsClsMin, thnAxisNumTpcClsMin, thnAxisIsRotatedCandidate});
```
fgrosa, line 555: From the comment I understood that you wanted to remove the rotational background axis (on which I would agree)
```suggestion
          registry.add("hHelicity", "THn for polarisation studies with cosThStar w.r.t. helicity axis in Pb-Pb", HistType::kTHnSparseF, {thnAxisInvMass, thnAxisPt, thnAxisNumPvContributors, thnAxisY, thnAxisInvMassD0, thnAxisCosThetaStarHelicity, thnAxisAbsEtaTrackMin, thnAxisNumItsClsMin, thnAxisNumTpcClsMin});
```

PWGHF/D2H/Tasks/taskCharmPolarisation.cxx
```diff
@@ -899,6 +991,38 @@ struct TaskPolarisationCharmHadrons {
           }
         }
       }
+    } else if constexpr (cosThetaStarType == charm_polarisation::CosThetaStarType::EP) { // EP
+      if constexpr (!doMc) {                                                             // data
+        if constexpr (withMl) {                                                          // with ML
+          if constexpr (channel == charm_polarisation::DecayChannel::DstarToDzeroPi) {   // D*+
+            registry.fill(HIST("hEP"), invMassCharmHad, ptCharmHad, numPvContributors, std::abs(rapCharmHad), invMassD0, cosThetaStar, outputMl[0], /*outputMl[1],*/ outputMl[2], absEtaMin, numItsClsMin, numTpcClsMin, isRotatedCandidate);
```
fgrosa, line 998: Shouldn't this be removed?
```suggestion
            registry.fill(HIST("hEP"), invMassCharmHad, ptCharmHad, numPvContributors, std::abs(rapCharmHad), invMassD0, cosThetaStar, outputMl[0], /*outputMl[1],*/ outputMl[2], absEtaMin, numItsClsMin, numTpcClsMin);
```

PWGHF/D2H/Tasks/taskCharmPolarisation.cxx
```diff
@@ -1068,15 +1206,67 @@ struct TaskPolarisationCharmHadrons {
     }
   }
 
+  /// Get the Q vector
```
fgrosa, line 1209: I would limit the options (see above)

PWGHF/D2H/Tasks/taskCharmPolarisation.cxx
```diff
@@ -1696,6 +1906,15 @@ struct TaskPolarisationCharmHadrons {
     ROOT::Math::PxPyPzMVector fourVecDauCM = boost(fourVecDau);
     ROOT::Math::XYZVector threeVecDauCM = fourVecDauCM.Vect();
 
+    if constexpr (withEP) {
+      ROOT::Math::XYZVector threeVecEP = ROOT::Math::XYZVector(-std::sin(*epPhi), std::cos(*epPhi), 0.f);
+      ROOT::Math::PxPyPzMVector fourVecEP = ROOT::Math::PxPyPzMVector(threeVecEP.X(), threeVecEP.Y(), threeVecEP.Z(), 0.f);
+      ROOT::Math::PxPyPzMVector fourVecEPCM = boost(fourVecEP);
+      ROOT::Math::XYZVector epVec = fourVecEPCM.Vect();
```
fgrosa, line 1913: I don't think this is the correct calculation of the cost* with respect to the EP (actually we need the normal direction)
I would suggest you to copy the Run 2 code:
1) First you compute the vector normal to the Q-vector:
https://github.com/alisw/AliPhysics/blob/4814a9a9c64f7d4b8d80fd8986d0ec3a47645430/PWGHF/vertexingHF/vHFML/AliAnalysisTaskSEDstarPolarization.cxx#L684
2) Then you compute the `cosThetaStar` as the scalar product between the normal vector and the daughter momentum boosted in the mother rest frame (this is what we need to boost)
https://github.com/alisw/AliPhysics/blob/4814a9a9c64f7d4b8d80fd8986d0ec3a47645430/PWGHF/vertexingHF/vHFML/AliAnalysisTaskSEDstarPolarization.cxx#L694C146-L694C155
As you can see in case of MC, the Q-vector direction was simply random (we have no flow in the MC, so it is equivalent)

PWGHF/D2H/Tasks/taskCharmPolarisation.cxx
```diff
@@ -1840,6 +2059,152 @@ struct TaskPolarisationCharmHadrons {
...
+    auto thisCollId = collision.globalIndex();
+    int numPvContributors = collision.numContrib();
+    auto groupedDstarCandidates = dstarCandidates.sliceBy(dstarPerCollision, thisCollId);
+    int nCands{0}, nCandsInSignalRegion{0};
+
+    std::vector<float> qVecs = getQvec(collision);
+    float xQVec = qVecs[0];
+    float yQVec = qVecs[1];
+    // float amplQVec = qVecs[2];
+    float evtPl = epHelper.GetEventPlane(xQVec, yQVec, harmonic);
```
fgrosa, line 2080: I would directly pass the Q-vector components to the run function (no need to compute the EP angle)
Same also below

fgrosa (changes requested):
Hi @Mingze129 thanks a lot for the development! The general implementation is good, but there are a couple of things to fix, see my comments below

Mingze129:
Hi, @fgrosa thanks for you detailed comments.
I have revised the task based on our previous discussion. Could you please check if this aligns with what you intended?

vkucera:
@Mingze129 The errors have not been fixed yet.

Mingze129:
Hi @vkucera , I am working on it. Once it's done I will restart the review.

Mingze129:
Hello @vkucera  I think I have fixed the most error except three point:
1. Line 253, we are try to define the axis range, it's not a particle mass.
2. Line 1732, 1698, 3 is the length of ML output, it doesn't have physics meaning.
3. Line 1552, 2090, We indeed involve some hard-coded PDG code, the most of them didn't not include in o2::constants::physics. 
https://aliceo2group.github.io/docs/d7/df9/namespaceo2_1_1constants_1_1physics.html

PWGHF/D2H/Tasks/taskCharmPolarisation.cxx
```diff
@@ -139,25 +150,11 @@ struct TaskPolarisationCharmHadrons {
...
-  ConfigurableAxis configThnAxisPtB{"configThnAxisPtB", {3000, 0.f, 300.f}, "#it{p}_{T}(B mother) (GeV/#it{c})"};
-  ConfigurableAxis configThnAxisAbsEtaTrackMin{"configThnAxisEtaTrackMin", {3, 0.f, 0.3f}, "min |#it{#eta_{track}}|"};
-  ConfigurableAxis configThnAxisNumItsClsMin{"configThnAxisNumItsClsMin", {4, 3.5f, 7.5f}, "min #it{N}_{cls ITS}"};
-  ConfigurableAxis configThnAxisNumTpcClsMin{"configThnAxisNumTpcClsMin", {3, 79.5f, 140.5f}, "min #it{N}_{cls TPC}"};
-  ConfigurableAxis configThnAxisCharge{"configThnAxisCharge", {2, -2.f, 2.f}, "electric charge"};
+  // Configurable<int> harmonic{"harmonic", 2, "harmonic number"};
+  Configurable<int> qvecDetector{"qvecDetector", 2, "Detector for Q vector estimation (FV0A: 0, FT0M: 1, FT0C: 2)"};
+  Configurable<int> centEstimator{"centEstimator", 2, "Centrality estimator ((None: 0, FT0C: 2, FT0M: 3))"};
+  Configurable<int> minCent{"minCent", 30, "Minimum centrality (0-100) to be considered in the analysis"};
+  Configurable<int> maxCent{"maxCent", 50, "Maximum centrality (0-100) to be considered in the analysis"};
```
vkucera, line 157:
```suggestion
  Configurable<int> qVecDetector{"qVecDetector", 2, "Detector for Q vector estimation (FV0A: 0, FT0M: 1, FT0C: 2)"};
  Configurable<int> centEstimator{"centEstimator", 2, "Centrality estimator ((None: 0, FT0C: 2, FT0M: 3))"};
  Configurable<int> centralityMin{"centralityMin", 30, "Minimum centrality (0-100) to be considered in the analysis"};
  Configurable<int> centralityMax{"centralityMax", 50, "Maximum centrality (0-100) to be considered in the analysis"};
```

PWGHF/D2H/Tasks/taskCharmPolarisation.cxx
```diff
@@ -1011,12 +1411,18 @@ struct TaskPolarisationCharmHadrons {
   template <charm_polarisation::DecayChannel channel>
   bool isInSignalRegion(float invMass)
   {
+    float minInvMass = 0.f;
+    float maxInvMass = 100.f;
```
vkucera, line 1415: Please stick to the naming conventions.
```suggestion
    float invMassMin = 0.f;
    float invMassMax = 100.f;
```

PWGHF/D2H/Tasks/taskCharmPolarisation.cxx
```diff
@@ -1087,15 +1493,48 @@ struct TaskPolarisationCharmHadrons {
     }
   }
 
+  /// Get the Q vector
+  /// \param collision is the collision with the Q vector information
+  template <typename CollsWithQvecs>
+  std::vector<float> getQvec(CollsWithQvecs const& collision)
```
vkucera, line 1499:
```suggestion
  template <typename CollsWithQVecs>
  std::vector<float> getQVec(CollsWithQVecs const& collision)
```

vkucera (changes requested):

PWGHF/D2H/Tasks/taskCharmPolarisation.cxx
```diff
@@ -1294,7 +1733,8 @@ struct TaskPolarisationCharmHadrons {
             invMassCharmHadForSparse = hfHelper.invMassLcToPiKP(candidate);
           }
           if constexpr (withMl) {
-            if (candidate.mlProbLcToPiKP().size() == 3) {
+            const int scoresNum = 3;
```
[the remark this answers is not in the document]
Mingze129: This constant now is define at beginning of the funciton "runPolarisationAnalysis", Line 1538

PR #11773, opened 2025-06-26
[PWGCF] Request to pull the needed code for hyperloop

CyF1rmiana:
Adding files for q2 calculation in PbPb_pass5.


victor-gonzalez:
And, please, never push code that has not been compiled and tested locally before
Central resources are scarce and shared. Please, help others

PR #11778, opened 2025-06-26
[Common] Fix autodetection string typos

ddobrigk:
@romainschotter thank you very much for spotting!


alibuild (dismissed):

PR #11785, opened 2025-06-26
add ntracks to the minicoll table + make trkl eta cuts configurable

maciacco:


alibuild (approved):

PR #11786, opened 2025-06-26
[PWGLF] nuclei-proton angular correlation: add DCA plots for secondary estimation

ercolessi:


maciacco:
As suggested by o2-linter, if possible use const refs in for loops to improve performance (can be checked later)

PR #11790, opened 2025-06-26
[PWGCF] Addition of extra event & track selections and configurable correlators

Preet-Bhanjan:


PWGCF/Flow/Tasks/resonancesGfwFlow.cxx
```diff
@@ -65,70 +64,136 @@ using namespace std;
...
+auto readMatrix(Array2D<T> const& mat, P& array)
+{
+  for (auto i = 0; i < static_cast<int>(mat.rows); ++i) {
+    for (auto j = 0; j < static_cast<int>(mat.cols); ++j) {
+      array[i][j] = mat(i, j);
+    }
+  }
+
+  return;
+}
```
victor-gonzalez, line 106: Why it is needed to template on the kind of array? Isn't it completely determined and unique?

PR #11810, opened 2025-06-27
[Infrastructure] Update LF Codeowners

maciacco:
@skundu692 @sustripathy this is the change that we proposed for the ownership of the `PWGLF/Utils` folder


ddobrigk:
Sorry for missing this and only approving before - please don't hesitate to ping me in such cases :-)

PR #11822, opened 2025-06-27
Fix event selection for generated signals

fgrosa:
@Luca610 @fchinu


alibuild (approved):

PR #11839, opened 2025-06-30
[Common] Batched evaluation for ONNX models to reduce memory consumption

ChSonnabend:


ddobrigk:
> @ddobrigk Can we merge this? Local tests show no difference in the output, so should leave the default behavior unchanged

Hi @ChSonnabend , sure thing - merged !

PR #11840, opened 2025-06-30
Adding protobuf model surgery

ChSonnabend:


ChSonnabend:
@ktf I added the model surgery stuff here. Hope thats ok, maybe in the future (depending on the usage) one could make a separate class for it, but its not much code yet.

PR #11841, opened 2025-06-30
[PWGLF] add truncated mean histograms and configurable to run on skimmed data

alcaliva:
- Introduced histograms that are filled using the cluster size truncated mean
- Added option to run on skimmed datasets via config flag


alcaliva:
Hi @skundu692 can you have a look?

PR #11843, opened 2025-06-30
[PWGEM] Implement histogram for momentum loss vs DCA, intention to select track w/wo Bremsstrahlung

feisenhu:
Implement pt_gen - pt_rec vs DCA histogram
Study the DCA shape of prompt and non-prompt sources which experience Bremsstrahlung.


alibuild (approved):

PR #11851, opened 2025-07-01
[PWGHF] Fix BC type in bc_as<> in B data creators

fgrosa:


alibuild (approved):

PR #11866, opened 2025-07-01
[PWGDQ] Creating Task Tag-And-Probe matching efficiency studies

rebeccacerri:


rebeccacerri:
Dear Maurice, thank you for your comments. This is the first draft of the code for the Tag and Probe matching efficiency (SW task). Together with Luca, we thought it would be useful to have a separate task from the table reader, allowing for an independent analysis without relying on the table reader, which makes use of many processes and could lead to a memory consumption very easily. The idea would be to potentially extend this task to the central barrel as well.
It works locally, but we are currently limited by the statistics we can analyse. Therefore, we would like to open a pull request so we can begin testing this task on Hyperloop.

mcoquet642:
@rebeccacerri thanks for the explanation. In that case, on my side I would be in favor of having this first version as light as possible and removing the unused stuff, that could be added later on when a given implementation requires it (in particular the type definitions and the flags)

rebeccacerri:
Dear Maurice, thank you for your comment. I have uploaded a minimal version with the unused parts removed. Let me know what you think about it. Thank you once again

mcoquet642 (approved):

PR #11877, opened 2025-07-02
[PWGHF,Trigger] Get rid of hard-coded mass cuts in HFFilter

fgrosa:
This PR makes all mass selections in the HFFilters configurable to work in case of bad mass resolutions @zhangbiao-phy @apalasciano


alibuild (approved):

EventFiltering/PWGHF/HFFilter.cxx
```diff
@@ -37,6 +37,7 @@
 #include "Common/DataModel/PIDResponseTPC.h"
 #include "Common/DataModel/TrackSelectionTables.h"
 
+#include "Framework/RunningWorkflowInfo.h"
```
vkucera, line 40: Use `<>` for all external headers.

PR #11881, opened 2025-07-02
OmegaC task, enable running one data and one MC process function at t…

fcatalan92:
…he same time

This could be useful when running on MC samples. @mtorresc15


alibuild (approved):

PR #11889, opened 2025-07-02
[PWGDQ] Temporary test for issues in running in hyperloop

lucamicheletti93:
Temporary removal of a part of the code that could be problematic in the current issue to run over derivd data on hyperloop. The removed part will be restored after the test


alibuild (approved):

PR #11902, opened 2025-07-03
[PWGLF] Add missing table in process function for Zorro

ercolessi:
@alcaliva @spucillo


alibuild (approved):

PR #11914, opened 2025-07-04
PWGHF: add down sample for data in Lc derived table creator

Mingyu3360715:


PWGHF/TableProducer/derivedDataCreatorLcToPKPi.cxx
```diff
@@ -87,7 +87,7 @@ struct HfDerivedDataCreatorLcToPKPi {
 
   HfHelper hfHelper;
   SliceCache cache;
-  static constexpr double mass{o2::constants::physics::MassLambdaCPlus};
+  static constexpr double MassLc{o2::constants::physics::MassLambdaCPlus};
```
vkucera, line 90: Please keep the name generic. It helps to keep the structure of all derived-data creators same.
```suggestion
  static constexpr double Mass{o2::constants::physics::MassLambdaCPlus};
```

vkucera (changes requested):

PWGHF/TableProducer/derivedDataCreatorLcToPKPi.cxx
Mingyu3360715: Thank you, I've applied it

PR #11916, opened 2025-07-04
[PWGJE,EMCAL-567] Add temperature calib. to EMC corr. task

jokonig:
- EMCal temperature calibration procedure is implemented in O2 and now used in the emcal correction task
- The temperature is assumed to be stable over the run and the temperature parameters are evaluated once at the initialization.


mhemmer-cern (approved):
Looks good to me. Thanks for the fix in processStandalone!

PR #11918, opened 2025-07-04
[PWGEM/PhotonMeson,PWGJE/EMCal] Add MC smearing functionality and add found FT0 gen histogram

nstrangm:


mhemmer-cern (dismissed):
Looks good to me.

PR #11919, opened 2025-07-04
[PWGEM] Efficiency with only MC varibles

rbailhac:
Efficiency with only MC varibles


dsekihat:
Hi Raphaelle, Thank you for your PR.
You don't need to add isSmeared many places. :)
mctrack.pt(), eta(), phi() returns generated value, when you run processQCMC. Therefore, adding another THnSparse is enough. If you have to do other things, shall I take care of it?

PR #11930, opened 2025-07-05
[PWGEM/Dilepton] add possibility to store only true electron in MC

dsekihat:
PWGEM/Dilepton: add possibility to store only true electron in MC


alibuild (approved):

PR #11932, opened 2025-07-06
[PWGHF] Add TTree to charm flow task for pt bin centering

Marcellocosti:
With this PR, a `TTree` is added to the flow task containing minimal information to extract the average `pt` for each `pt` interval. Tagging @stefanopolitano and @wuctlby for information.


zhangbiao-phy:
> We are afraid that the tree creators tables contain too many columns, leading to very large outputs when running on the full sample, so we implemented this minimal table. Furthermore, if we see that the output size is reasonable, the follow-up would be to add the centrality and scalar product columns and see if it is possible to perform the measurement directly f

PR #11945, opened 2025-07-07
[PWGJE,EMCAL-568] Fix mistake in gain calib shift

jokonig:
- paramsOld are the desired parameters, hence, the data has to be corrected using paramsOld/paramsNew
- Fixes https://github.com/AliceO2Group/O2Physics/pull/11935


mhemmer-cern (approved):
Looks good

PR #11946, opened 2025-07-07
[PWGHF] taskCharmPolarisation: Stop using the iterator at collisions for PbPb collision process

Mingze129:
Incorrect use of iterator, lead to error when slice on candidates.


vkucera:
Please make the PR title more informative by adding the scope of the changes.

Mingze129:
> > Incorrect use of iterator, lead to error when slice on candidates.
> 
> Why does the grouping not work directly with the iterator?

The grouping is working well with the iterator.
The issues is I slice the candidate again while the candidate already sliced (line 2351, 2321). 

And we don't want to slice the tracks by the collision, so we have to abandon the iterator.

vkucera (approved):

PR #11959, opened 2025-07-07
[PWGEM/Dilepton] fill gen. value for rec. tracks

dsekihat:
PWGEM/Dilepton: fill gen. value for rec. tracks


rbailhac:
Hi @dsekihat , it looks ok to me (as long as the output is not too big but it should be ok). Please proceed. Thanks

rbailhac (approved):

PR #11960, opened 2025-07-07
[PWGHF,PWGJE] Propagate new MC flags to beauty candidates

vkucera:
Continuation of #11382

- Propagate new MC flags for B0, B+, Bs, Lb.
- MC tables extended with columns for resonant channels and unified.
- Reduced workflows are not affected except for renaming of columns in the main data model.


nzardosh (approved):

vkucera:
> Hi @vkucera, it seems to me that all if correct, thanks! Currently the "non reduced" workflows do not include all the matchings for the correlated backgrounds, but I would say that they can be added in a second PR.

Thanks @fgrosa

PR #11979, opened 2025-07-09
Fix bug introduced in #11881

fcatalan92:


alibuild (approved):

PR #11987, opened 2025-07-09
[PWGCF] FemtoUniverse -- corrected mc truth filling for the secondary contamination estimation

zchochul:
Before mc truth was filled only for the truth particles associated with the mc reco particles. Now this is filled, which should enable correct secondary contamination estimation.


alibuild (approved):

PR #11997, opened 2025-07-10
[PWGUD] New 2D histos and vars (tracks, rct) in tau tree

amatyja:
New 2D histos (m3pi vs aco),
new variables (track DCA, TPC chi2, ITS chi2, rct) in tau tree,
MC histograms properly defined in different processes

### DIFF
--- a/PWGUD/Tasks/upcTauTau13topo.cxx
+++ b/PWGUD/Tasks/upcTauTau13topo.cxx
@@ -1641,7 +1641,7 @@ struct TauTau13topo {
   {
     return std::sqrt(px * px + py * py + pz * pz + mass * mass);
   }
-  
+
   //  float calculateDeltaPhi(TLorentzVector p, TLorentzVector p1)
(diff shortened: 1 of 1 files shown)


PWGUD/Tasks/upcTauTau13topo.cxx
```diff
@@ -1635,7 +1636,14 @@ struct TauTau13topo {
     return std::sqrt(E * E - px * px - py * py - pz * pz);
   }
 
-  float calculateDeltaPhi(TLorentzVector p, TLorentzVector p1)
+  float energy(float px, float py, float pz, float mass)
+  // Just a simple function to return energy
+  {
+    return std::sqrt(px * px + py * py + pz * pz + mass * mass);
+  }
```
vkucera, line 1643: You can just use `RecoDecay::e`

vkucera: Similar for the other kinematic calculations.

vkucera:
The code has plenty of useless functions which can be replace with direct calls of `RecoDecay` methods. Please avoid reinventing the wheel and use existing features instead.

rolavick (approved):

PR #12005, opened 2025-07-10
Add parametrisation of D-meson mass shape for 2025_apass1

fgrosa:


alibuild (approved):

PR #12015, opened 2025-07-11
[PWGHF] Fix converter for B-meson reduced data model (PID variables)

fgrosa:


alibuild (approved):

PR #12019, opened 2025-07-11
[Trigger] Zorro: Fix various C++ issues

vkucera:
- Member variable ... is not initialized in the constructor. (cppcheck)
- Function parameter ... should be passed by const reference. (cppcheck)
- Searching before insertion is not necessary. Use `try_emplace` instead. (cppcheck)
- Use `static_cast` instead of deprecated casting style. (cpplint)
- IWYU (clang-tidy)
- Include external headers with `<>`. (manual)
- Sort includes. (clang-format)
- Replace `size_t` with `std::size_t`.
- Use `const&` to iterate `std::vector<ZorroHelper>`. (O2 linter)

### DIFF
--- a/EventFiltering/Zorro.cxx
+++ b/EventFiltering/Zorro.cxx
@@ -12,13 +12,30 @@
 
 #include "Zorro.h"
 
+#include "EventFiltering/ZorroHelper.h"
+
+#include <CCDB/BasicCCDBManager.h>
+#include <CommonConstants/LHCConstants.h>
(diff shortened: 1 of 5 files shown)


mpuccio:
Have you checked that there is no functional change? If so, we can merge

vkucera:
> Have you checked that there is no functional change? If so, we can merge

Yes

mpuccio (approved):

PR #12027, opened 2025-07-11
[PWGLF] Add QC task for grouping tests + update sigma0builder

gianniliveraro:
This PR introduces a new task (sigma0QC) to evaluate the performance of different V0 grouping strategies in O2Physics using strangeness derived data.

It enables a direct comparison between four grouping approaches:

- Grouping via collision iterator;

- Grouping using Preslice;

- Grouping using PresliceUnsorted;

- Custom grouping using a simple std::vector<std::vector<>>.

Each approach is implemented in a separate struct. The base template, defined in "sigma0BuilderHelper.h", follows the logic of the sigma0builder task: V0s are grouped per collision, filtered using photon- and lambda-like criteria, and combined to build Sigma0 candidates.

Additionally, this PR also updates the sigma0builder task to adopt the custom grouping strategy for analysis.

Tagging @ddobrigk and @romainschotter for completeness.


romainschotter (dismissed):

ddobrigk:
O2 compile error is spurious (from before the fix), newest O2 CI should work

PR #12071, opened 2025-07-15
PWGDQ: Update the MuonForEMu cut in CutsLibrary.cxx

MaolinZH:
Add the MCHMID cut:

This is to reduce the rejection factor of e-mu trigger in CEFP


MaolinZH:
+label trigger

iarsene (approved):

PR #12085, opened 2025-07-16
PWGHF: Fix sliceBy for Bs data creator

fcatalan92:
Not sure if this bug had an impact on the analysis, to be checked after the PR is merged.

@Marcellocosti Thanks for spotting it!


alibuild (approved):

PR #12111, opened 2025-07-17
[PWGHF] Add D*Pi channel for B0 meson 

Marcellocosti:
With this PR, the decay channel B0 -> D*Pi is added to the `dataCreatorCharmHadPiReduced` workflow. Then, the default values of all process functions are set to false except for the first one and a grouping of the datamembers is implemented.

### DIFF
--- a/PWGHF/D2H/TableProducer/dataCreatorCharmHadPiReduced.cxx
+++ b/PWGHF/D2H/TableProducer/dataCreatorCharmHadPiReduced.cxx
@@ -247,8 +247,9 @@ struct HfDataCreatorCharmHadPiReduced {
   Preslice<CandsLcFiltered> candsLcPerCollision = aod::track_association::collisionId;
   Preslice<CandsLcFilteredWithMl> candsLcPerCollisionWithMl = aod::track_association::collisionId;
   Preslice<aod::TrackAssoc> trackIndicesPerCollision = aod::track_association::collisionId;
-  PresliceUnsorted<CollisionsWCentAndMcLabels> colPerMcCollision = aod::mccollisionlabel::mcCollisionId;
   Preslice<aod::McParticles> mcParticlesPerMcCollision = aod::mcparticle::mcCollisionId;
+
+  PresliceUnsorted<CollisionsWCentAndMcLabels> colPerMcCollision = aod::mccollisionlabel::mcCollisionId;
 
   std::shared_ptr<TH1> hCandidatesD0, hCandidatesDPlus, hCandidatesDs, hCandidatesLc;
   HistogramRegistry registry{"registry"};
@@ -1681,7 +1682,7 @@ struct HfDataCreatorCharmHadPiReduced {
       o2::hf_evsel::checkEvSel<true, o2::hf_centrality::CentralityEstimator::None, aod::BCsWithTimestamps>(collision, hfEvSel, zvtxColl, sel8Coll, zvtxAndSel8Coll, zvtxAndSel8CollAndSoftTrig, allSelColl, ccdb, registry);
 
       auto thisCollId = collision.globalIndex();
-      auto candsCThisColl = candsC.sliceBy(candsDplusPerCollision, thisCollId);
+      auto candsCThisColl = candsC.sliceBy(candsDsPerCollision, thisCollId);
       auto trackIdsThisCollision = trackIndices.sliceBy(trackIndicesPerCollision, thisCollId);
       runDataCreation<false, false, DecayChannel::BsToDsminusPi, false>(collision, candsCThisColl, trackIdsThisCollision, tracks, tracks, -1, bcs);
     }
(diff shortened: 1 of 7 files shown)


PWGHF/D2H/TableProducer/dataCreatorCharmHadPiReduced.cxx
```diff
@@ -109,47 +110,50 @@ enum WrongCollisionType : uint8_t {
...
+    Produces<aod::HfRedTracksCov> hfTrackCovPion;
+    Produces<aod::HfRedTracksPid> hfTrackPidPion;
+    // charm hadron related tables
+    Produces<aod::HfRed2Prongs> hfCand2Prong;
+    Produces<aod::HfRed2ProngsCov> hfCand2ProngCov;
+    Produces<aod::HfRed2ProngsMl> hfCand2ProngMl;
+    Produces<aod::HfRed3Prongs> hfCand3Prong;
+    Produces<aod::HfRed3ProngsCov> hfCand3ProngCov;
+    Produces<aod::HfRed3ProngsMl> hfCand3ProngMl;
+    Produces<aod::HfRedB0SoftPi> hfCandDstarSoftPi;
```
fgrosa, line 130: For the soft pion, we need the corresponding track tables as for the bachelor:
```
    Produces<aod::HfRedTrackBases> hfTrackPion;
    Produces<aod::HfRedTracksCov> hfTrackCovPion;
```
Without the track parameters otherwise we cannot re-reconstruct the B vertex from the derived data

PWGHF/D2H/TableProducer/dataCreatorCharmHadPiReduced.cxx
```diff
@@ -936,9 +963,44 @@ struct HfDataCreatorCharmHadPiReduced {
...
         }
-        rowHfLcPiMcCheckReduced(pdgCodeBeautyMother, pdgCodeCharmMother, pdgCodeProng0, pdgCodeProng1, pdgCodeProng2, pdgCodeProng3);
+        tables.rowHfLcPiMcCheckReduced(pdgCodeBeautyMother, pdgCodeCharmMother, pdgCodeProng0, pdgCodeProng1, pdgCodeProng2, pdgCodeProng3);
+      }
+      tables.rowHfLcPiMcRecReduced(indexHfCandCharm, selectedTracksPion[vecDaughtersB.back().globalIndex()], flag, flagWrongCollision, debug, motherPt);
+    } else if constexpr (decChannel == DecayChannel::B0ToDstarPi) {
+      // B0 → D*+ π- → (D0 π+) π- → (K- π+ π+) π-
+      auto indexRec = RecoDecay::getMatchedMCRec<true, false, false, true, true>(particlesMc, std::array{vecDaughtersB[0], vecDaughtersB[1], vecDaughtersB[2], vecDaughtersB[3]}, Pdg::kB0, std::array{-kKPlus, +kPiPlus, +kPiPlus, -kPiPlus}, true, &sign, 4);
+      if (indexRec > -1) {
+        // D*+ → (D0 π+) → K- π+ π+
```
fgrosa, line 973: In case of the `D*` it would also be good to match the D0 daughter (it should never happen that, if the `D*` is matched the `D0` is not, but just to be sure)

PWGHF/D2H/TableProducer/dataCreatorCharmHadPiReduced.cxx
```diff
@@ -1136,6 +1212,36 @@ struct HfDataCreatorCharmHadPiReduced {
...
+        hCandidatesDstar->Fill(SVFitting::FitOk);
+
+        auto secondaryVertexCharm = df2.getPCACandidate();
+        trackParCov0.propagateTo(secondaryVertexCharm[0], bz);
+        trackParCov1.propagateTo(secondaryVertexCharm[0], bz);
+        std::array<float, 3> pVecD0{};
+        df2.getTrack(0).getPxPyPzGlo(pVecD0); // D0
+        pVecCharm = RecoDecay::pVec(pVecD0, pVec2);
+        trackParCovCharmHad = df2.createParentTrackParCov();
+        trackParCovCharmHad.setAbsCharge(charmHadDauTracks[2].sign()); // sign of soft pion
```
fgrosa, line 1244: In the `D*` case, we should not reconstruct the `D*` track, because the `D*` does not fly. The `trackParCovCharmHad` should instead contain the `D0` track and then we should add an additional table with the soft pion track (see comment above). In this way later we can reconstruct B decay vertices by fitting the D0 + 2 pion tracks (since from the B vertex this is what you get)

fgrosa (commented):
Hi @Marcellocosti thanks a lot for the development!
In general it looks good to me, but there is something in the logic to be modified, especially for what concerns the charm track and the soft pion track (see below).

Marcellocosti:
Hi @fgrosa, I replaced the 3prong table with the 2prong one. Please let me know if I missed anything!

fgrosa (approved):

PR #12114, opened 2025-07-17
[PWGLF] updated datamodel of spin spin correlation with MC information

prottayCMT:


PWGLF/TableProducer/Strangeness/lambdaspincorrelation.cxx
```diff
@@ -340,7 +342,105 @@ struct lambdaspincorrelation {
...
+    std::vector<int> positiveIndex = {};
+    std::vector<int> negativeIndex = {};
+    std::vector<float> dcaBetweenDaughter = {};
+    int numbV0 = 0;
+    // LOGF(info, "event collisions: (%d)", collision.index());
+    auto centrality = collision.centFT0C();
+    auto vz = collision.posZ();
+    int occupancy = collision.trackOccupancyInTimeRange();
+    histos.fill(HIST("hEvtSelInfo"), 0.5);
+    if ((rctCut.requireRCTFlagChecker && rctChecker(collision)) && collision.selection_bit(aod::evsel::kNoSameBunchPileup) && collision.selection_bit(aod::evsel::kIsGoodZvtxFT0vsPV) && collision.selection_bit(o2::aod::evsel::kNoCollInTimeRangeStandard) && collision.sel8() && collision.selection_bit(o2::aod::evsel::kIsGoodITSLayersAll) && occupancy < cfgCutOccupancy) {
```
romainschotter, line 365: Hi @prottayCMT ! In the current implementation, if you don't want to check the RCT flags, the whole if-condition will be false.
I think you should write `(rctCut.requireRCTFlagChecker || rctChecker(collision))` instead of `(rctCut.requireRCTFlagChecker && rctChecker(collision))`

romainschotter (changes requested):

PWGLF/TableProducer/Strangeness/lambdaspincorrelation.cxx
prottayalice: its ok...i will fill the table only with RCT condition..if it is false it shouldn't fill anything

romainschotter: All good! Just wanted to let you know, I was afraid it was not deliberate

prottayalice: Thanks!

PR #12127, opened 2025-07-17
[Common] Temporary first-order fix to CCDB caching mistake

ddobrigk:
Addresses a situation in which orbit reset queries are responded with incorrect information by the CCDB manager if these were previously cached. Underlying root cause still to be investigated further - this solution addresses only the symptom of an as-of-yet unresolved issue as far as I can tell. I will keep digging. Tagging @fcatalan92 @fgrosa @mpuccio @lietava


alibuild (approved):

PR #12135, opened 2025-07-18
[PWGLF] increase counter only for (physical) primary particles

maciacco:


alibuild (approved):

PR #12144, opened 2025-07-18
[PWGCF] add pt container and remove some old code due to the limit num of o2 objects

Luzhiyongg:
add pt container for pt correlations and vn-pt correlations


victor-gonzalez (approved):
If the limit is in the number of `Configurable`s you can always use the `ConfigurableGroup`s
If the limit is in the number of histograms you can always use more than one `HistogramRegistry`. In one of my `QA` task I remember using `30` so it is perfectly safe

PR #12153, opened 2025-07-19
[PWGLF] make number of its clusters configurable

maciacco:


alibuild (approved):

PR #12157, opened 2025-07-19
ZorroSummary fix

lietava:


lietava:
When I follow failed check I see No error founds ?

lietava:
@mpuccio : please, have a look on this.

PR #12159, opened 2025-07-19
[Common] Fix timestamp module

ddobrigk:
Replaces `getForTimestamp` with `getSpecific` in CCDB calls of the timestamp module to ensure metadata is cleared in the internal state of the CCDB manager. Tagging @dsekihat


alibuild (approved):

PR #12176, opened 2025-07-21
[PWGHF] ML selections for D0 in D* selector

Marcellocosti:
This PR introduces the possibility of applying ML selection on the D0 daughter of a D* candidate, in order to allow a correct efficiency evaluation in triggered data.

### DIFF
--- a/PWGHF/TableProducer/candidateSelectorDstarToD0Pi.cxx
+++ b/PWGHF/TableProducer/candidateSelectorDstarToD0Pi.cxx
@@ -17,6 +17,7 @@
 
 #include "PWGHF/Core/HfHelper.h"
 #include "PWGHF/Core/HfMlResponseDstarToD0Pi.h"
+#include "PWGHF/Core/HfMlResponseD0ToKPi.h"
 #include "PWGHF/Core/SelectorCuts.h"
 #include "PWGHF/DataModel/CandidateReconstructionTables.h"
 #include "PWGHF/DataModel/CandidateSelectionTables.h"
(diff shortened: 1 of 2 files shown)


PWGHF/TableProducer/candidateSelectorDstarToD0Pi.cxx
```diff
@@ -110,7 +120,9 @@ struct HfCandidateSelectorDstarToD0Pi {
 
   HfHelper hfHelper;
   o2::analysis::HfMlResponseDstarToD0Pi<float> hfMlResponse;
+  o2::analysis::HfMlResponseDstarToD0Pi<float> hfMlResponseD0Daughter;
```
vkucera, line 123: Shouldn't this be an `HfMlResponseD0ToKPi` object?

Marcellocosti: Hi Vit, good point, I was unsure about this. In the end, I chose it to be a `HfMlResponseDstarToD0Pi` object because this implementation is necessary only when one wants to study the D*. Indeed, if one is interested in the D0, the BDT can be directly applied in the corresponding selector, but there was no possibility to apply a BDT selection on the D0s used to reconstruct D* candidates.

mfaggin: Hi @Marcellocosti , thanks a lot. Maybe this is the cleanest solution, because the usage of the `HfMlResponseD0ToKPi` class would imply either the existence of a `candidate-2-prong` object (that you do not have since the D* candidates come directly from the track-index-skim-creator) or a further customization of the `HfMlResponseD0ToKPi` to retrieve the D0 variables from the D* candidates. But I am missing one thing: why do not you use directly the `hfMlResponse` object which exists already? You duplicate all the code with `hfMlResponseD0Daughter` but ptobably it's not necessary, since what you need can be done by simply choosing only the D0 features as training variables. Am I right? Am I missing anything?

mfaggin: Hi Marcello, now I see. You are right, I was forgetting the double BDT when analysing triggered data. Fine for me then.

PR #12184, opened 2025-07-22
[PWGEM/PhotonMeson] Add omega EMC task and extend gg histograms

nstrangm:
- Add a dedicated task for omega mesons with the EMC for performance reasons in OO and PbPb
- Extend range of pT, time and mu for histograms in pi0 task


mhemmer-cern (approved):
looks good

PR #12208, opened 2025-07-23
[PWGHF] Fix filling of MC collision info in OmegaC creator

fcatalan92:
@mtorresc15


alibuild (approved):

PR #12209, opened 2025-07-23
[Common] Update centrality task with run dependant histograms

jesgum:
This PR includes the following changes 
-  Addition of `MultsRun3` which no longer joins the Run 3 multiplicity tables with the Run 2 `TrackletMults` and changing subscription in the process functions accordingly.
- Created `initRun` which creates the essential histograms dynamically for each Run. Each Run adds ~80 MB of memory needed by the histogram registry which initially needs ~50MB.
- Flag for ir fetcher which sometimes causes crashes in light ion runs.
- Add subscription and histograms for FV0AOuter in process functions calling `genericProcessCollision()`.
- Added 3d histogram FV0A vs FV0AOuter vs time which needs to be enabled with flag `doTimeStudyFV0AOuterVsFT0A3d` (not created per Run).


ddobrigk (approved):
thanks a lot @jesgum, looks very good!

PR #12212, opened 2025-07-23
[PWGLF] Add task to produce binned information about V0s/cascades

romainschotter:
Add task running over strangeness derived data and producing binned information about V0s/cascades 

@miranov25 for your information

### DIFF
--- a/PWGLF/TableProducer/Common/epvector.cxx
+++ b/PWGLF/TableProducer/Common/epvector.cxx
@@ -19,42 +19,46 @@
 ///
 
 // C++/ROOT includes.
+#include <TComplex.h>
 #include <TH1F.h>
+#include <TMath.h>
+
 #include <chrono>
+#include <cstdio>
 #include <string>
 #include <vector>
-#include <TComplex.h>
-#include <TMath.h>
-#include <cstdio>
 
 // o2Physics includes.
-#include "Framework/AnalysisDataModel.h"
-#include "Framework/AnalysisTask.h"
-#include "Framework/runDataProcessing.h"
-#include "Framework/HistogramRegistry.h"
-#include "Framework/StepTHn.h"
-#include "ReconstructionDataFormats/Track.h"
-#include "Common/DataModel/PIDResponse.h"
-#include "Common/DataModel/Multiplicity.h"
-#include "Common/DataModel/Centrality.h"
-#include "Common/DataModel/TrackSelectionTables.h"
-#include "Common/DataModel/EventSelection.h"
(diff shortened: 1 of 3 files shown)


alibuild (approved):

miranov25:
Thank you for your code, @romainschotter 

As discussed on Mattermost and in person, I have some comments on configuration and implementation and would like to share the reasoning behind the strategy I’ve successfully applied for multidimensional group-by statistics extraction since Run 1 (and which contributed to the development of **THn** in ROOT about 15 years ago).


* I’m aware the **memory limit is imposed by ROOT - imposed by buffer size in streames**; this is why I previously suggested using **THnS**. In practice, we will likely use a **set of THnS objects**, as I do in my performance parameterization code.

**Configurable Binning**
The variables and binning should remain **fully configurable**, as the current defaults look too coarse in some cases:
* **Invariant mass**: The bin width is roughly
$$\frac{0.547 - 0.447}{25} \approx 0.004$$
This is too coarse for precise fits. A normalized bin width in terms of \$p\_T\$ resolution would be more appropriate.

* **Radius**: Binning up to 250 cm should be reduced to match the **actual fiducial volume** accessible.
We will prototype improved functional binning with Rahul and Marian in our standalone code for the derived AO2D data
. For example:
* For **φ**, we often use **coarse global φ bins** or **local sector φ bins** to capture edge-related effectssqrtSaling for pt.
Once ready we can chack how to post it to the O2 code

---

**Why Functional Transformations Matter**
Rather than manually specifying non-uniform bins, I recommend using **smooth variable transformations**.
In my previous multidimensional analyses, the steps were:

1. Apply a **functional transformation** (e.g., sqrt, asinh) to flatten the distribution or ensure minimal bin width at high \$p\_T\$.
2. Perform **group-by statistics** (e.g., Gaussian fits) on the transformed grid.

---

**Dealing with Sparse Bins**
When working in high dimensions, statistics per bin decrease rapidly - 1/nBins. To mitigate this, I used a **moving average window** across neighboring bins:

* Compute **bin centers (COG)** dynamically.
* For smoothly varying properties, combine data from ±1 bin per dimension.
  This increases effective statistics by roughly $3^{\text{ndims}}$.

This methodology allowed me to generate **performance maps for Run 1, Run 2, and early Run 3**. The implementation is currently in **AliRoot**, but I’m considering porting it to **O2**, as mixing O2 and AliRoot in the same workflow caused some issues recently - I assume then name clashes for some code - if they used the same variable names in AliRoot and O2.

PR #12214, opened 2025-07-24
[PWGLF] To add Charged K* production task without ResoInitializer, with ME and rotational backgrounds, and MC processing

navneetkumar231295:
These changes add the charged K* production task without the ResoInitializer, along with mixed event and rotational backgrounds and MC processing.


dmallick2:
@navneetkumar231295 
thanks. could you please fix o2linter error 
For  Error: Avoid hard-coded PDG codes. 
Use named values from PDG_t or o2::constants::physics::Pdg instead.

navneetkumar231295:
ok, Thanks

PR #12215, opened 2025-07-24
[PWGHF] Fix misleading histogram in event selection utils

fcatalan92:
Small fix for the histogram of the MC collisions passing the event selections, the name and axis title was misleading


vkucera (approved):
Thanks @fcatalan92

PR #12277, opened 2025-07-27
[Common] single-device TPC PID service for testing

ddobrigk:
* adds a `pidTPCModule` class similar to other classes that deal with core services 
* this class is used in a `pidTPCService` test task that provides Nsigma values to users in the exact same manner (with identical output at least in the default mode). This task is able to switch between `Tracks` and `TracksIU` for testing differences due to use of `Tracks` vs `TracksIU` with existing calibrations in an easier way. 
* While previously the regular TPC PID required 3 DPL devices (one multiplicity counter, one corrected dE/dx device and one actual PID calculation task), this new refactored version has only one DPL device. 

The main practical consequence of this refactoring is that the memory footprint ('dirty private') decreased from 213 MB to 139 MB (-35%) and the PSS in a typical Pb-Pb test scenario went from approximately 500MB to 320MB (-35%). Further optimizations are still pending.


alibuild (approved):

PR #12293, opened 2025-07-28
[PWGLF] Added signal/event loss corrections

sawankumawat:


dmallick2:
@sawankumawat please fix o2linter error

sawankumawat:
> @sawankumawat please fix o2linter error

The only o2linter error is "Assign the value to a clearly named variable or constant". But then I have to create many variables in the code at every place, which will not look good. @vkucera Please suggest

dmallick2 (dismissed):

sawankumawat:
> > > @sawankumawat please fix o2linter error
> > 
> > 
> > The only o2linter error is "Assign the value to a clearly named variable or constant". But then I have to create many variables in the code at every place, which will not look good. @vkucera Please suggest
> 
> Hi @sawankumawat @dmallick2 , you are not supposed to "create many variables in the code at every place", on the contrary, you are supposed to define a few meaningful variables to replace the hard-coded numbers. In your case, it seems you only need a constant for the rapidity cut and for the number of daughters.
> 
> Btw, you have 4 other categories of errors which should also be fixed. You are hard-coding masses and PDG codes and reinventing the phi range constraints. Please use the existing functionalities as the linter suggests.

Hi @vkucera Thank you for the suggestion. I have solved all the linter errors now.

PWGLF/Tasks/Resonances/higherMassResonances.cxx
```diff
@@ -170,6 +170,11 @@ struct HigherMassResonances {
     // ConfigurableAxis axisdEdx{"axisdEdx", {20000, 0.0f, 200.0f}, "dE/dx (a.u.)"};
     // ConfigurableAxis axisPtfordEbydx{"axisPtfordEbydx", {2000, 0, 20}, "pT (GeV/c)"};
     // ConfigurableAxis axisMultdist{"axisMultdist", {3500, 0, 70000}, "Multiplicity distribution"};
+
+    // fixed variables
+    float rapidityMotherData = 0.5;
+    std::array<int, 6> numbers = {0, 1, 2, 3, 4, 5};
+    double beamMomentum = std::sqrt(13600 * 13600 / 4 - o2::constants::physics::MassProton * o2::constants::physics::MassProton); // GeV
   } config;
```
vkucera, line 178: The value `13600` is a magic number which appears multiple times in the code. Define a constant `beamEnergy` for it.

PWGLF/Tasks/Resonances/kstarqa.cxx
```diff
@@ -102,6 +102,12 @@ struct Kstarqa {
     Configurable<float> cfgRCRFC{"cfgRCRFC", 0.8f, "Crossed Rows to Findable Clusters"};
     Configurable<float> cfgITSChi2NCl{"cfgITSChi2NCl", 36.0, "ITS Chi2/NCl"};
     Configurable<float> cfgTPCChi2NCl{"cfgTPCChi2NCl", 4.0, "TPC Chi2/NCl"};
+
+    // Other fixed variables
+    float lowpTcutinpTdepPID = 0.5;
```
vkucera, line 107: Use proper camelCase. If you break `lowpTcutinpTdepPID` down into words, you get `lowp-tcutinp-tdep-p-i-d` which is not what you mean.

PR #12294, opened 2025-07-28
[PWGLF] added event selection cut in MC

sarjeetagami:


dmallick2:
@sarjeetagami please fix o2linter errors

PR #12304, opened 2025-07-29
[PWGHF] Add the configurable occupancy estimator in candidate creator

wuctlby:
Hi @zhangbiao-phy, this PR is for adding a configurable occupancy estimator in the candidate creator. Since the estimator was hard-coded as `ITS` from the beginning, this could cause some potential problems. 

Here, I would also like to @maciacco , even though he is on vacation, in case he wants to further check something when he returns.

I already tested it locally, and everything went smoothly.

PS: I couldn't find a method to read the estimator option directly from the 'hfEvSel', so I added a new variable instead. 
NOTE: If the FT0c is used, the occupancy range must also be changed accordingly.

Please let me know if you have any suggestions. Thanks a lot!

### DIFF
--- a/PWGHF/TableProducer/candidateCreator2Prong.cxx
+++ b/PWGHF/TableProducer/candidateCreator2Prong.cxx
@@ -105,6 +105,7 @@ struct HfCandidateCreator2Prong {
   Configurable<double> minParamChange{"minParamChange", 1.e-3, "stop iterations if largest change of any X is smaller than this"};
   Configurable<double> minRelChi2Change{"minRelChi2Change", 0.9, "stop iterations is chi2/chi2old > this"};
   Configurable<bool> fillHistograms{"fillHistograms", true, "do validation plots"};
+  Configurable<int> occEstimator{"occEstimator", 1, "Occupancy estimation (1: ITS, 2: FT0C)"};
   // magnetic field setting from CCDB
   Configurable<bool> isRun2{"isRun2", false, "enable Run 2 or Run 3 GRP objects for magnetic field"};
   Configurable<std::string> ccdbUrl{"ccdbUrl", "http://alice-ccdb.cern.ch", "url of the ccdb repository"};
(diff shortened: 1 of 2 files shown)


vkucera:
> > **O2 linter results:** ❌ 3 errors, ⚠️ 0 warnings, 🔕 12 disabled
> 
> Regarding the errors, as I am not the developer of either code, I am not sure if it's okay that I directly disable the check for the corresponding parts.

They are already disabled.

vkucera:
`occEstimator` is already included in `HfEventSelection` and you literally copy-pasted its definition from there.
What's the point of this PR?

PWGHF/TableProducer/candidateCreator2Prong.cxx
```diff
@@ -750,7 +751,7 @@ struct HfCandidateCreator2Prong {
 
       /// bitmask with event. selection info
       float centrality{-1.f};
-      float occupancy = getOccupancyColl(collision, OccupancyEstimator::Its);
```
wuctlby, line 753: Here? If I understand correctly, all highlighted parts are about the hard-coded `OccupancyEstimator::Its`.

vkucera:
I see your point now.
Please get the estimator using `hfEvSel.occEstimator`.

vkucera:
> > I see your point now. Please get the estimator using `hfEvSel.occEstimator`.
> 
> Yes, I also tried it. You mean something like `occEstimator = hfEvSel.occEstimator`?

That should work too but I don't see why you need the new variable.

wuctlby:
> > > I see your point now. Please get the estimator using `hfEvSel.occEstimator`.
> > 
> > 
> > Yes, I also tried it. You mean something like `occEstimator = hfEvSel.occEstimator`?
> 
> That should work too but I don't see why you need the new variable.

It's due to that I tried the  `occEstimator = hfEvSel.occEstimator`, but when compiling it, it fails. So ..., I will add the variable type and try it again in the docker of ubuntu 22.04

zhangbiao-phy:
hi @wuctlby, many thanks for your commits! Hi @vkucera, if you don't have further comments, I would like to approve it now. we are doing the study of D0 production vs occupancy.

zhangbiao-phy (approved):

vkucera (approved):

PR #12325, opened 2025-07-30
[PWGEM] Add DCA1 vs DCA2 hist to DielectronMC task

feisenhu:
Adding TH2 to DielectronMC task for promptPi0 and PromptJPsi, to check DCA1 vas DCA2 of tracks


alibuild (approved):

PR #12331, opened 2025-07-30
[PWGHF] Save ML scores in LcToK0sP tree creator

DelloStritto:


vkucera:
Why are the ML scores needed in the tree tables?

fgrosa (approved):

PR #12333, opened 2025-07-30
[DPG] Allowing to avoid collision index in track table. This is needed/usef…

arossi81:
…ul for running on hyperloop. Collision info (position, ncontrib) can be stored track-by-track in a dedicated table that is mergeable with the track one. Practically, with typical track-sampling criteria used, this allows to save space.


arossi81:
@alcaliva , could you merge this pull request (if that's ok for you of course) so that can enter in next tag? I would need it to run on hyper loop and get some output before next week.

PR #12363, opened 2025-07-31
[PWGEM/PhotonMeson,Trigger] update pcm data and add converters

dsekihat:
PWGEM/PhotonMeson, Trigger: update pcm data and add converters


vkucera:
> @dsekihat @mpuccio Linter reports several `abs` bugs in the trigger part. Please do not ignore the linter errors.

Header includes are missing.

PR #12385, opened 2025-08-01
Trigger+ Multiplicity+ LeadingConstituentCuts+ Modified MCCollision Matching + QA Split vertices + Zorro

Archita-Dash:
This big commit includes the following big changes:

1. added trigger based zorro utilities. Modified the trigger process function to be able to check for trigger overlaps
2. modified the multiplicity estimation tables that now use the latest ones implemented in the Jet framework
3. leading track and cluster cuts are in place now for the MCD level
4. leading track cut implemented for the MCP level
5. removed the previous mc collision matching from the jet loop to the event loop
6. the above bullet 5. now enables to QA the "split vertices"
7. changed the binning of the uncorrected cluster timing distribution in the Emcal-Hadronic-Correction task for QA purposes.


nzardosh (approved):

vkucera:
@Archita-Dash Use the standard logging features to print messages.
Your own message formats, like
```c++
std::cout << "WARNING: 
```
make the logs harder to read and process.
On top of that `cout` is defined in the `iostream` header which is forbidden.
Both issues are reported by the O2 linter.

Archita-Dash:
> @Archita-Dash Use the standard logging features to print messages. Your own message formats, like
> 
> ```c++
> std::cout << "WARNING: 
> ```
> 
> make the logs harder to read and process. On top of that `cout` is defined in the `iostream` header which is forbidden. Both issues are reported by the O2 linter.

@vkucera , that task was written quite sometime ago and some of its parameters need to be updated following the changes made in another core task; for this I am currently in the process of creating a new PR specific to the emcalClusterHadronicCorrectionTask.cxx. I didn't intend to make any changes to it in this commit. So yes, I will consider the above linter errors in the new PR that I am working on.

PR #12406, opened 2025-08-03
[PWGCF] cfgTrackBitMask type changed to uint16_t

nmallick19:
Previously, cfgTrackBitMask was decalred as uint8_t
This resulted in a code crash for any value of cfgTrackBitMask>=10
It is understood that any value with two digits or more is treated as character, hence the code crashes.

Solution is to use uint16_t instead.


jgrosseo:
I think this doesn't make sense because the value is then compared in line 126 to a 8 bit value in the stored derived data.

What is the problem you are trying to solve?

jgrosseo:
> @jgrosseo this is an unfortunate limitation of C++ itself, it expects a character there for the uint8_t that is internally equivalent to unsigned char, when the configuration is passed to the deployed task process. It is indeed a general recommendation to avoid using uint8_t for configurables.

But this happens only on the command line or also with json? With json we should have full control, don't we?

jgrosseo:
I see. We will test if this has any side effect on our side and then merge.

PR #12418, opened 2025-08-04
[Common] Event selection flags added

MaximVirta:
Event selection flags to distinguish selections between PbPb and smaller systems (OO, Ne-Ne)


Common/Tasks/qVectorsCorrection.cxx
```diff
@@ -541,16 +542,31 @@ struct qVectorsCorrection {
...
-    }
-    if (cfgAddEvtSelPileup && !qVec.selection_bit(o2::aod::evsel::kNoCollInTimeRangeStandard)) {
-      return;
+    if (cfgAddEvtSel) {
+      switch (cfgEvtSel) {
+        case 0: // Sel8
+          if (!qVec.sel8())
+            return;
+          break;
+        case 1: // PbPb standard
```
vkucera, line 551: Do not use magic numbers. Define an enumerator for the supported cases.

ddobrigk (approved):

PR #12423, opened 2025-08-04
[PWGCF] Fix bug in FV0 eta calculation

abmodak:
1) Fix the initialisation of `fv0det` geometry pointer  
2) Add `addHistos` function instead of using `addClone`

### DIFF
--- a/PWGCF/TwoParticleCorrelations/Tasks/longrangeCorrelation.cxx
+++ b/PWGCF/TwoParticleCorrelations/Tasks/longrangeCorrelation.cxx
@@ -75,6 +75,9 @@ static constexpr TrackSelectionFlags::flagtype TrackSelectionDca =
 static constexpr TrackSelectionFlags::flagtype TrackSelectionDcaxyOnly =
   TrackSelectionFlags::kDCAxy;
 
+static constexpr std::string_view kCorrType[] = {"Ft0aGlobal/", "Ft0cGlobal/", "Fv0Global/", "MftGlobal/", "Fv0Mft/"};
+static constexpr std::string_view kEvntType[] = {"SE/", "ME/"};
+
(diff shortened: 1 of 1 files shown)


PWGCF/TwoParticleCorrelations/Tasks/longrangeCorrelation.cxx
```diff
@@ -75,6 +75,9 @@ static constexpr TrackSelectionFlags::flagtype TrackSelectionDca =
 static constexpr TrackSelectionFlags::flagtype TrackSelectionDcaxyOnly =
   TrackSelectionFlags::kDCAxy;
 
+static constexpr std::string_view kCorrType[] = {"Ft0aGlobal/", "Ft0cGlobal/", "Fv0Global/", "MftGlobal/", "Fv0Mft/"};
+static constexpr std::string_view kEvntType[] = {"SE/", "ME/"};
+
```
victor-gonzalez, line 80: Thanks for considering my  suggestions!

I have an additional one for security and error avoidance in case you want to also consider it

I would suggest not trusting on remembering the meaning of `0` or `3` or of the position in the template parameters
Two additional `enum`s can be defined
```
enum KindOfMixing {
  kSE,
  kME
};

enum KindOfMultiplicity {
  kFT0AGLOBAL,
  kFT0CGLOBAL,
  kFV0GLOBAL,
  kMFTGLOBAL,
  kFV0MFT
};
```
and the functions can be templated with this types, for instance
```
  template <KindOfMultiplicity corrType, KindOfMixing evntType>
  void addHistos()
```
Now if always the named constants are used the compiler will advice if there is a misuse at any point
```
addHistos<kFT0GLOBAL, kSE>(...)
```
will compile correctly but
```
addHistos<kSE, kFT0GLOBAL>(...)
```
will give an error

The chosen names are, of course, representative

PR #12425, opened 2025-08-04
[PWGCF] Add CFMultiplicitySet to support multiple auxilary multiplicities/centralities

jaelpark:


PWGCF/TableProducer/filterCorrelations.cxx
```diff
@@ -548,9 +548,43 @@ struct MultiplicitySelector {
...
+      if (cfgEstimatorBitMask & kMultFV0A)
+        multiplicities.push_back(collision.multFV0A());
+      if (cfgEstimatorBitMask & kMultNTracksPV)
+        multiplicities.push_back(collision.multNTracksPV());
+      if (cfgEstimatorBitMask & kMultNTracksGlobal)
+        multiplicities.push_back(collision.multNTracksGlobal());
+      output(multiplicities);
+    }
+  }
+  PROCESS_SWITCH(MultiplicitySetBuilder, processEstimators, "Process auxiliary multiplicity/centrality estimators", false);
```
victor-gonzalez, line 581: The workflow will hang if there are not default `process` function
Is it intended to always run this task?

jaelpark: If it hangs, it remains to be solved. Most of the time it won't be needed - only in some specific cases we need the extra estimators (e.g. outliers, systematics). By default it should not run.

victor-gonzalez: As the task is incorporated to the workflow and there is not `process` function assigned to it most of the times, the workflow will hang. At least it used to be that way
Have you tested locally?

PWGCF/DataModel/CorrelationsDerived.h
```diff
@@ -86,6 +86,15 @@ using CFTrackWithLabel = CFTracksWithLabel::iterator;
 //------transient CF-filter to CF-2prong-filter
 DECLARE_SOA_TABLE(CFCollRefs, "AOD", "CFCOLLREF", o2::soa::Index<>, track::CollisionId); //! Transient cf collision index table
 
+//------multiplicity set
+namespace cfmultiplicityset
+{
+DECLARE_SOA_COLUMN(Multiplicities, multiplicities, std::vector<float>); //! List of auxiliary multiplicities
+}
+DECLARE_SOA_TABLE(CFMultiplicitySets, "AOD", "CFMULTSET", cfmultiplicityset::Multiplicities); //! Auxilary multiplicity set table
```
jgrosseo, line 94: The name doesn't work like this for merging. The first and third names must be the same + s

PWGCF/TableProducer/filterCorrelations.cxx
jgrosseo: It could not be part of FilterCF simply?

PWGCF/TableProducer/filterCorrelations.cxx
```diff
@@ -255,6 +269,19 @@ struct FilterCF {
...
+        multiplicities.push_back(collision.centFT0C());
+      if (cfgEstimatorBitMask & kMultFV0A)
+        multiplicities.push_back(collision.multFV0A());
+      if (cfgEstimatorBitMask & kMultNTracksPV)
+        multiplicities.push_back(collision.multNTracksPV());
+      if (cfgEstimatorBitMask & kMultNTracksGlobal)
+        multiplicities.push_back(collision.multNTracksGlobal());
+      outputMultSets(multiplicities);
+    }
+
```
jgrosseo, line 284: It does not matter much but in principle this code could be in `processDataMults`. Nothing should speak against filling one table there and the then one would not need the `HasMultTables`, or am missing something?

jgrosseo (approved):
I am ready to merge with the two doubts below. Feel free to still fix them in this PR or in the next one.

PWGCF/DataModel/CorrelationsDerived.h
```diff
@@ -86,6 +86,15 @@ using CFTrackWithLabel = CFTracksWithLabel::iterator;
 //------transient CF-filter to CF-2prong-filter
 DECLARE_SOA_TABLE(CFCollRefs, "AOD", "CFCOLLREF", o2::soa::Index<>, track::CollisionId); //! Transient cf collision index table
 
+//------multiplicity set
+namespace cfmultiplicityset
+{
+DECLARE_SOA_COLUMN(Multiplicities, multiplicities, std::vector<float>); //! List of auxiliary multiplicities
+}
+DECLARE_SOA_TABLE(CFMultiplicitySets, "AOD", "CFMULTIPLICITYSET", cfmultiplicityset::Multiplicities); //! Auxilary multiplicity set table
```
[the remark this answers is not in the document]
jaelpark: You're right that the name is clipped to `CFMULTIPLICITYS` (or `o2cfmultiplicitys`). It's not ideal so I'll change it in the next PR while making other fixes too.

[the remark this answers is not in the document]
jgrosseo: @ktf @aalkin in the past this (too long string in the third argument) caused a compilation error? Why is this gone?

PR #12509, opened 2025-08-09
Revert "[PWGCF] FemtoUniverse: Fixing the bug in phi calculation at R"

prchakra:
Reverts AliceO2Group/O2Physics#12315

The use of (-)ve sign takes care of proper bending of the charged particles due to the magnetic field.


alibuild (approved):

PR #12515, opened 2025-08-10
[PWGLF] Added invariant mass histogram of K0s

SuJeong-Ji:
Invariant mass histogram of K0s is added.


dmallick2:
@SuJeong-Ji could you please fix a few o2linter error related to "Do not use the TLorentzVector legacy class. Use std::array with RecoDecay methods or the ROOT::Math::LorentzVector template instead" and Avoid magic numbers in expressions. Assign the value to a clearly named variable or constant.

PR #12518, opened 2025-08-10
[PWGCF] FemtoUniverse: Adding segment to fill MCTruth tracks

prchakra:


prchakra:
@victor-gonzalez Hi, I need your suggestion here. The `clang-format` test of this PR is failing and suggesting to reorder the header files. The clang-formatting is suggesting to add `PWGCF/FemtoUniverse/Core/FemtoUniverseSHContainer.h` after `PWGCF/FemtoUniverse/Core/FemtoUniversePairSHCentMultKt.h`. Now, If I do that, I am getting a build error and the reason (I think) is `PWGCF/FemtoUniverse/Core/FemtoUniversePairSHCentMultKt.h` uses a namespace `femto_universe_sh_container` which is declared in `"PWGCF/FemtoUniverse/Core/FemtoUniversePairSHCentMultKt.h"`. Now, what should I do in this case?

PR #12521, opened 2025-08-10
[PWGCF] DptDpt - Using metadata for auto autoconfiguration

victor-gonzalez:


alibuild (approved):

PR #12550, opened 2025-08-12
[PWGUD] SGSelector: Fix const correctness

vkucera:
Make all non-muted input `const`.


vkucera:
@pbuehler @rolavick Can you please review the PR?

PR #12562, opened 2025-08-13
[DPG] new configs for nITScls, isGoodStaves etc.

altsybee:


alibuild (approved):

PR #12604, opened 2025-08-16
[PWGCF] DptDpt - Multiplicity observables correlations

victor-gonzalez:
Also
- configurables rationalization
- correlation between event selection triggers
- no active collision flags when not selected after trigger selection


alibuild (approved):

PR #12615, opened 2025-08-18
[PWGJE] Simulate hadron decay

pestratm:
Now, two histograms are filled with photons reconstructed using the cocktail method, where Pythia8 is used to simulate decay of neutral hadrons into photons (in one case taking neutral hadrons directly, in one case using charged pions as proxies).
Additionally, the pTHat of the collisions are saved and can be cut upon (if specified in the config files)

### DIFF
--- a/PWGJE/Tasks/CMakeLists.txt
+++ b/PWGJE/Tasks/CMakeLists.txt
@@ -38,10 +38,6 @@ o2physics_add_dpl_workflow(emc-tmmonitor
                     SOURCES emcTmMonitor.cxx
                     PUBLIC_LINK_LIBRARIES O2::Framework O2::EMCALBase O2::EMCALCalib O2Physics::AnalysisCore
                     COMPONENT_NAME Analysis)
-o2physics_add_dpl_workflow(hadron-photon-correlation
-                    SOURCES hadronPhotonCorrelation.cxx
-                    PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::AnalysisCore
-                    COMPONENT_NAME Analysis)
 o2physics_add_dpl_workflow(mc-generator-studies
                     SOURCES mcGeneratorStudies.cxx
                     PUBLIC_LINK_LIBRARIES O2::Framework O2::EMCALBase O2::EMCALCalib O2Physics::AnalysisCore
@@ -302,4 +298,17 @@ if(FastJet_FOUND)
                         SOURCES jetShape.cxx
                         PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::PWGJECore O2Physics::AnalysisCore
                         COMPONENT_NAME Analysis)
+endif()
+
+if(pythia_FOUND)
+  if(TARGET ROOT::EGPythia8)
+    set_target_properties(ROOT::EGPythia8 PROPERTIES
+        INTERFACE_LINK_LIBRARIES "ROOT::Core;ROOT::EG;ROOT::Graf;ROOT::Physics;${pythia_LIBRARY_SHARED}"
+    )
+
+    o2physics_add_dpl_workflow(hadron-photon-correlation
+                        SOURCES hadronPhotonCorrelation.cxx
+                        PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::AnalysisCore ROOT::EGPythia8
+                        COMPONENT_NAME Analysis)
+  endif()
 endif()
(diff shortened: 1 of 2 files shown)


pestratm:
Since the POWHEG directphoton process does not include a gg->gg process, I added a configurable switch to remove those events also in PYTHIA standalone if desired

pestratm:
Called Sumw2 on all histograms for correct uncertainties

PWGJE/Tasks/hadronPhotonCorrelation.cxx
```diff
@@ -39,6 +39,10 @@
 #include <Framework/InitContext.h>
 #include <Framework/runDataProcessing.h>
 
+#include "TClonesArray.h"
+#include "TLorentzVector.h"
+#include "TParticle.h"
+#include "TPythia8Decayer.h"
```
mhemmer-cern, line 45: Could you please change the include of ROOT and other outside includes to use <> instead of ""
In general for local includes (anything from O2Physics) in "" and everything else is considered outside and should use <>

PWGJE/Tasks/hadronPhotonCorrelation.cxx
```diff
@@ -93,18 +112,30 @@ struct HadronPhotonCorrelation {
   float ptMinAssoc;
   float ptMaxAssoc;
 
+  TPythia8Decayer* decayer = new TPythia8Decayer;
+  TLorentzVector* motherLV = new TLorentzVector(); // TLorentzVector required by TPythia8Decayer
```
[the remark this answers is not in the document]
pestratm: Thanks for the tip, that's good to know. I have added that now

PWGJE/Tasks/CMakeLists.txt
```diff
@@ -303,3 +299,16 @@ if(FastJet_FOUND)
                         PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::PWGJECore O2Physics::AnalysisCore
                         COMPONENT_NAME Analysis)
 endif()
+
```
fjonasALICE, line 302: no clue what this is, but did you test/confirm that this works? Will it work on hyperlool?

PWGJE/Tasks/hadronPhotonCorrelation.cxx
```diff
@@ -128,55 +159,179 @@ struct HadronPhotonCorrelation {
...
     registry.add("reconstructed/charged/hPionCorrelReco", "Reconstructed Trigger-Pion Correlation", kTHnSparseF, {axisPtTrig, axisPtAssoc, axisDeltaEta, axisDeltaPhi});
+    registry.get<THnSparse>(HIST("reconstructed/charged/hPionCorrelReco"))->Sumw2();
     registry.add("reconstructed/charged/hPionMultReco", "Reconstructed Pion Multiplicity", kTH1F, {axisMult});
+    registry.get<TH1>(HIST("reconstructed/charged/hPionMultReco"))->Sumw2();
     registry.add("reconstructed/charged/hPionSpectrumReco", "Reconstructed Pion Spectrum", kTHnSparseF, {axisPtAssoc, axisEtaAssoc, axisPhi});
+    registry.get<THnSparse>(HIST("reconstructed/charged/hPionSpectrumReco"))->Sumw2();
+  }
+
+  // Get pTHat of the event
+  template <typename P>
```
fjonasALICE, line 254: do we have a central JE function for this?

PWGJE/Tasks/hadronPhotonCorrelation.cxx
```diff
@@ -128,55 +159,179 @@ struct HadronPhotonCorrelation {
...
+          isDiffractive = true;
+        }
+      }
+    }
+
+    return pTHat;
+  }
+
+  template <typename P>
+  bool rejectOutliers(const P& particles, double pTHat)
```
fjonasALICE, line 283: do we have central functions for this in JE?

fjonasALICE (commented):
Made a few comments, apart from that fine by me to approve @nzardosh

PWGJE/Tasks/hadronPhotonCorrelation.cxx
pestratm: I checked other JE Tasks (e.g. [fullJetSpectra](https://github.com/AliceO2Group/O2Physics/blob/44edb11daa411d2a53d4dba7bfc1ed97226477c9/PWGJE/Tasks/fullJetSpectra.cxx#L1133)), and there, the pTHat is determined using the weight of the event. This works, however, only for the jj production, which is weighted, and not for example for the unweighted MB production or POWHEG+PYTHIA productions, which are weighted differently.

PWGJE/Tasks/hadronPhotonCorrelation.cxx
pestratm: SImilar to the pTHat, in the [fullJetSpectra](https://github.com/AliceO2Group/O2Physics/blob/44edb11daa411d2a53d4dba7bfc1ed97226477c9/PWGJE/Tasks/fullJetSpectra.cxx#L1165) task, outlier rejection is done explicitly

PR #12621, opened 2025-08-18
[DPG] Fix bug with physical primary

scattaru:


scattaru:
Dear @alcaliva , @chiarazampolli , @catalinristea ,
I would ask if it would be possible to go on with the merging.

Thanks a lot,
Cheers,
Samuele

PR #12624, opened 2025-08-18
[PWGHF] Adding Preslice to tracks and McParticles

samrangy:
Preslices are added to fix errors occurring in event mixing.


PWGHF/HFC/TableProducer/correlatorD0Hadrons.cxx
```diff
@@ -101,6 +101,8 @@ struct HfCorrelatorD0HadronsSelection {
   using SelCollisions = soa::Join<aod::Collisions, aod::EvSels>;
 
   Preslice<aod::HfCand2Prong> perCol = aod::hf_cand::collisionId;
+  Preslice<aod::Tracks> perCollisionID = aod::track::collisionId;
+  Preslice<aod::McParticles> perTrueCollision = o2::aod::mcparticle::mcCollisionId;
```
fgrosa, line 105: They are not used anywhere, why do you slice if then the table slice is not used?

fgrosa (changes requested):

PR #12638, opened 2025-08-19
[PWGJE] Addition of bins for vertex finding efficiency and vertex reconstruction corrections

Archita-Dash:
Hi @nzardosh ,

this commit includes additional bins before and after the Vtx Z cuts in order to calculate 

- vertex finding efficiency correction factor for the vertex Z cut of +- 10 cm in the analysis. This can now be obtained directly by taking the ratio between, for instance in my case, **Coll+TVX+Sel8Full+VtxZ+IsGoodZVtxFT0vsPV** and **Coll+TVX+Sel8Full+IsGoodZVtxFT0vsPV**
- vertex reconstruction bias correction factor which is basically the correction for the zero vertex. This can now be calculated by taking ratio between **Coll+TVX+Sel8Full+IsGoodZVtxFT0vsPV** and **BC+TVX+NoTFB+NoIsROB**.

Both can be obtained directly with data (reconstructed level). I tested locally with a small file and this seems to work as desired. I've attached here the histogram from my local test. As I mentioned the other day at the JP PAG meeting, I personally use Sel8Full+IsGoodVtxZ event selection for my analysis; however I am aware that other charged jet analyses are using sel8. Therefore I have included the corresponding bins for sel8 for the sake of completeness. 

Cheers,
Archita

Screenshot from 2025-08-19 18-29-38


Archita-Dash:
Update: Closing this PR because after discussing with Nima, we figured out that the lumi producer task itself needs refactoring in the jet framework and that for my correction factors estimation for my analysis, it is easiest and simplest to include a BC loop in my analysis which has now been implemented in my full jet task and the merged PR can be found here: https://github.com/AliceO2Group/O2Physics/pull/12721

Archita.

PR #12650, opened 2025-08-20
[Common] Protect against multiple downloads from TrackTuner

mfaggin:
Possible solution to avoid any rewrite of the same files if the `--pipeline` option is used in Hyperloop for the device instantiating the `TrackTuner`

@ddobrigk @fcatalan92 @arossi81 @phymanshu

### DIFF
--- a/Common/Tools/TrackTuner.h
+++ b/Common/Tools/TrackTuner.h
@@ -467,6 +467,7 @@ struct TrackTuner : o2::framework::ConfigurableGroup {
       // get the DCA correction file from CCDB if not yet downloaded before
       struct stat sbDca; // search utility
       if (stat(fullNameInputFile.c_str(), &sbDca) == 0) {
+        // file found
         LOG(info) << " [TrackTuner] File " << fullNameInputFile << " already downloaded. Not downloading it anymore";
       } else {
         LOG(info) << "[TrackTuner] downloading input file " << nameInputFile << " from CCDB...";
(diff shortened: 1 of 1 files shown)


Common/TableProducer/trackPropagation.cxx
```diff
@@ -133,7 +133,20 @@ struct TrackPropagation {
...
 
-      trackTunerObj.getDcaGraphs();
+      /// see if other track-propagation devices are present
+      int counter = 0;
+      LOG(info) << "  ---> looking for track-propagation devices";
+      const auto& workflows = initContext.services().get<RunningWorkflowInfo const>();
+      for (const DeviceSpec& device : workflows.devices) {
+        if (device.name.compare("track-propagation") == 0) {
+          // init HF event selection helper
+          LOG(info) << "    FOUND!";
```
ddobrigk, line 143: Thanks @mfaggin ! I would remove this log with "FOUND" in favor of something after the loop reporting the count of devices, i.e. `LOGF(info, "Number of track propagation tasks detected: %i", counter);` or so. Could you please adjust?

mfaggin:
Hi, I have just opened another draft PR here https://github.com/AliceO2Group/O2Physics/pull/12748 where I propose the usage of `getfromTimeStamp` instead of `retrieveBlob`, which should do what we want. This would imply the reload of all calibration files, but they are not too many (less than 10)

PR #12652, opened 2025-08-20
[PWGHF] Split PID tables for 2prong, 3prong, Dstar (DO NOT MERGE)

fcatalan92:


fcatalan92:
Hi @vkucera, I wanted to compile the code with the CI since I cannot do it locally at the moment.
It seems there is a maximum length for the table names, let's see if the fix works.

vkucera:
> Hi @vkucera, I wanted to compile the code with the CI since I cannot do it locally at the moment. It seems there is a maximum length for the table names, let's see if the fix works.

Yes, the maximum is 16 characters if I remember correctly.

fcatalan92:
@vkucera, all, the PR is ready to be reviewed and merged

PR #12670, opened 2025-08-21
[PWGJE] Add MC mixing and adjusting task for use on Hyperloop

jkinner2:
- added MC true mixing in the same way as reconstructed data
- split producer task to own file (to allow subwagons with different configurations on Hyperloop)
  - added internal multiplicity estimate based on the number of JetTracks
- extended invariant mass/correlation to eta meson
- adjusted an issue in the efficiency process regarding bias towards selected tracks

### DIFF
--- a/PWGJE/Tasks/photonChargedTriggerProducer.cxx
+++ b/PWGJE/Tasks/photonChargedTriggerProducer.cxx
@@ -1,0 +1,347 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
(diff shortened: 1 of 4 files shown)


fjonasALICE (commented):
It is too much to look through everything, but @nzardosh since this does not touch any other task, we can approve without fear of breaking?

PWGJE/Tasks/photonChargedTriggerProducer.cxx
```diff
@@ -0,0 +1,347 @@
...
+      if (!checkPipmTPC(trackPID))
+        continue;
+
+      // pipm
+      pipmTable(track.collisionId(), track.globalIndex(), track.pt(), track.phi(), track.eta());
+    }
+  }
+  PROCESS_SWITCH(PhotonChargedTriggerProducer, processRecoPipmTPC, "process pipm (TPC) table (reconstructed)", false);
+
+  void processRecoPhotonPCM(soa::Join<aod::JetCollisions, aod::JCollisionPIs>::iterator const& collision, aod::Collisions const&,
```
nzardosh, line 268: why do you need aod::Collisions const& here?

nzardosh: You still need aod::JCollisionPIs because thats where the Id colum is stored. However the aod::Collisions table is the one you probably dont need to subscribe to

PR #12677, opened 2025-08-21
[PWGCF] FemtoDream: added event sphericity calculation

RafRaf11:
New feature: event sphericity calculation in FemtoDream framework
- added Event sphericity histogram
- added sphericity calculation (with matrix elements taken from EPJC72:2124)
- added sphericity event cut code in producer tasks 
- deleted femtoDreamProducerTaskWithCascades.cxx due to incompatibility with sphericity calculation code (in consultation with @gmantzar: cascade building will be reintegrated in the normal producer task, making the ProducerWithCascades obsolete. PR is under preparation) 
- added event cut on minimal sphericity in femtoDreamPairTaskTrackTrack.cxx

### DIFF
--- a/PWGCF/FemtoDream/Core/femtoDreamCollisionSelection.h
+++ b/PWGCF/FemtoDream/Core/femtoDreamCollisionSelection.h
@@ -202,7 +202,7 @@ class FemtoDreamCollisionSelection
       double eta = track.eta();
       double px = track.px();
       double py = track.py();
-      if (TMath::Abs(pt) < /*lowerPtbound*/ 0.5 || TMath::Abs(eta) > 0.8) {
+      if (TMath::Abs(pt) < mSphericityPtmin || TMath::Abs(eta) > 0.8) {
         continue;
       }
 


PWGCF/FemtoDream/Core/femtoDreamCollisionSelection.h
```diff
@@ -180,9 +185,65 @@ class FemtoDreamCollisionSelection
...
+    double s00 = 0.; // elements of the sphericity matrix taken form EPJC72:2124
+    double s01 = 0.;
+    double s10 = 0.;
+    double s11 = 0.;
+
+    int numOfTracks = col.numContrib();
+    if (numOfTracks < 3)
+      return -9999.;
+
+    for (auto& track : tracks) {
```
lauraser, line 200: auto const& track

PWGCF/FemtoDream/Core/femtoDreamCollisionSelection.h
```diff
@@ -180,9 +185,65 @@ class FemtoDreamCollisionSelection
...
+      double py = track.py();
+      if (TMath::Abs(pt) < /*lowerPtbound*/ 0.5 || TMath::Abs(eta) > 0.8) {
+        continue;
+      }
+
+      ptTot += pt;
+
+      s00 += px * px / pt;
+      s01 += px * py / pt;
+      s10 = s01;
```
lauraser, line 213: you never use this variable later and also don't properly normalize to total pT, it is better to write it as a comment in s01 line that one of the other matrix elements you need is s10 but it is equal to s01 so you use them as quivalent.

lauraser (commented):
General comments on the code, I will approve the PR once I test it locally

PWGCF/FemtoDream/Core/femtoDreamCollisionSelection.h
RafRaf11: I changed it according to the comment.

PR #12679, opened 2025-08-21
[PWGDQ] Implemented a postcalibration scheme for electrons in Pb-Pb collisions, and some cleanup

iarsene:
The postcalibration implemented here relies on 4D maps used to correct the position of the TPC electron, pion and proton bands and the width of the band. The dimensions in which the correction is done are eta, number of PV, number of pileup tracks in the TPC and the median time of the pileup tracks in the TPC. One can still select between the already existing postcalibration (typically for pp) and the new one, via the configurable options added and by selecting the correct CCDB path. More instructions are available in the O2 DQ presentations.


alibuild (approved):

lucamicheletti93 (approved):

mcoquet642 (approved):

PR #12741, opened 2025-08-26
[PWGHF] updated to QA for Z->ee for Z0 trigger study modify to set non-linear centrality bins in histograms

sashingo:
- updated to QA for Z->ee for Z0 trigger study 
- modify to set non-linear centrality bins in histograms]


sashingo:
Hi @jpxrk and @apalasciano ,
could you review the commit when you have time ?
thanks in advance,

PR #12748, opened 2025-08-26
[Common] move from retrieveBlob to getFromtimeStamp for TrackTuner inputs

mfaggin:
This development should prevent conflicts in calibration file download/reading in case of multi-thread execution (`pipeline` option)

### DIFF
--- a/Common/Tools/TrackTuner.h
+++ b/Common/Tools/TrackTuner.h
@@ -447,11 +447,6 @@ struct TrackTuner : o2::framework::ConfigurableGroup {
     return outputString;
   }
 
-  std::string getFullNameInputPath()
-  {
-    return pathInputFile + std::string("/") + nameInputFile;
-  }
-
   void getDcaGraphs()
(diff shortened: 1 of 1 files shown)


Common/Tools/TrackTuner.h
```diff
@@ -447,53 +447,59 @@ struct TrackTuner : o2::framework::ConfigurableGroup {
     return outputString;
   }
 
+  std::string getFullNameInputPath()
+  {
+    return pathInputFile + std::string("/") + nameInputFile;
+  }
+
```
fcatalan92, line 454: This is not used later on in the code

fcatalan92 (commented):
@mfaggin Looks good to me! I have just a minor comment

Common/Tools/TrackTuner.h
mfaggin: thanks Fabio, I've removed it

PR #12753, opened 2025-08-26
[PWGHF] OmegaC task, rework output structure + add info for vs mult. analysis

fcatalan92:
@Yunfan-Liu @mtorresc15 Please have a look, since this likely impact your offline macros.
I re-worked the output of the task, now when running on data only a ThnSparse called `hReco` is produced, while running on MC you will get also one with the generated information `hMcGen` (and `hMcGenWithRecoColl` in case one runs with the centrality information on). The axes of the sparses depend on the process function selected (more axes are added when running with centrality).


alibuild (approved):

PR #12777, opened 2025-08-28
[PWGLF] Make sure that B field is updated when necessary

mpuccio:


maciacco (dismissed):

alibuild (dismissed):

maciacco (approved):

PR #12802, opened 2025-08-31
[Common] Cleanup: remove test tasks

ddobrigk:
Removes test tasks that were put in place in preparation of core service improvement campaign.


alibuild (approved):

PR #12812, opened 2025-09-01
Add possibility to avoid D0 daus swapping in D* ML application

fgrosa:
@Luca610 @Marcellocosti this PR should allow us to reproduce the BDT application as in the software triggers also for the D*, by turning on the swapping of the positive and negative tracks


alibuild (approved):

PR #12818, opened 2025-09-02
[PWGHF] Minor fix in taskDplus

apalasciano:
Missing arguments in THnSparse fill


stefanopolitano (approved):
Hi @apalasciano ! Thanks for the fix. No comment form my side.

PR #12820, opened 2025-09-02
[PWGCF] - FemtoDream: Cascade Update

gmantzar:
- Adding the Cascades back to the Producer Task
- Fixed bug in the Cascade Selection: Even thought, the selection was switching correctly between Omegas and Xis, the Producer Task wrote always only the Xi mass into the FemtoDerived table. Similar for the QA plots. This is fixed now
- Adding the CascadeCascade Pair Task:
--> Tested and working but missing the CPR (and the pair cleaning). Needs to be done!
- Implementing the "before selection" and "after selection" QA histograms for all particle types in the Producer Task


victor-gonzalez (approved):
Thanks!!

PR #12842, opened 2025-09-03
[PWGLF] Update V0 analysis task

romainschotter:
For derivedlambdakzeroanalysis:
- Add V0 Vs Phi vs Eta histograms
- Add cut to remove tracks close TPC sector edges
- Add QA histogram on V0 selection
- Add Event QA histograms

For cascademlselection:
- Switch from O2 grouping to manual grouping, as it is more CPU efficient and it allows to bypass index ordering checks


alibuild (dismissed):

ercolessi (approved):

PR #12856, opened 2025-09-03
[PWGLF] Added new method for track reassociation

gbencedi:
Major: 
- Added new method for track reassociation based on propagateToDCAhelix
Minor:
- Added new event selection based on RCT flags
- Added basic correlation histos Nch/PV tracks vs FT0C


alibuild (approved):

PR #12859, opened 2025-09-04
[PWGHF] tree creator for Sigmac correlated bkg. sources.

mfaggin:
@Mingyu3360715


mfaggin:
To address the two remaining O2linter errors I've just opened this PR in O2: https://github.com/AliceO2Group/AliceO2/pull/14649

I'd prefer to go ahead with both the PRs (this one with the O2linter errors and the O2 one) separately, and fix the O2linter error easily after the O2 PR will be merged.

PR #12864, opened 2025-09-04
[PWGHF,PWGCF] Store charm hadron + track invariant mass in derived tree

zhangbiao-phy:


PWGCF/DataModel/FemtoDerived.h
```diff
@@ -327,6 +328,7 @@ DECLARE_SOA_TABLE(FDHfPairs, "AOD", "FDHFPAIRS", //! table to store results for
                   fdhf::MultPercentile,
                   fdhf::Charge,
                   fdhf::PairSign,
+                  fdhf::CharmTrkM,
```
victor-gonzalez, line 331: Are you aware that this will make the task not backwards compatible unless a converter is in place?

zhangbiao-phy: > Are you aware that this will make the task not backwards compatible unless a converter is in place?

Hi @victor-gonzalez, Thanks for pointing this out. This table is currently used only in my analysis, so backward compatibility isn’t required. If it’s adopted more broadly later, I’ll provide a converter.

PR #12870, opened 2025-09-04
[PWGCF] updating PID sel for phi v2

prottayCMT:


otonvd (approved):
Please communicate with Victor about those check failures in column names

victor-gonzalez (commented):

PWGCF/DataModel/CorrelationsDerived.h
```diff
@@ -140,7 +140,9 @@ enum ParticleDecay {
   JPsiToEE,
   JPsiToMuMu,
   Generic2Prong,
-  PhiToKK,
+  PhiToKKPID1,
+  PhiToKKPID2,
+  PhiToKKPID3,
```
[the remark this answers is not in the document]
jaelpark: Yes, it does . But maybe not the non-phi related datasets.

PWGCF/TableProducer/filter2Prong.cxx
```diff
@@ -432,6 +433,79 @@ struct Filter2Prong {
     return false;
   }
 
+  template <typename T>
+  bool selectionPID2(const T& candidate)
```
jaelpark, line 437: I think this code could have been much, much shorter with some loops and arrays.

prottayCMT: I agree the selectionPID part added could be much shorter...May be I will do in any next commit.

jaelpark:
@prottayCMT (@otonvd) thanks for the reply. There are indeed other users of the phi codes already (other than myself also), but it's okay since I'm in touch with them. I also wanted to suggest that perhaps we could think of dedicating a small (upper) portion of the `decay` column for a bitmask to include various PID cuts simultaneously, then we don't need multiple datasets or filters to perform systematics. We could then just keep one `PhiToKK` value and add a dynamic column to extract the sub-cuts. Let me know if you're interested.

jaelpark:
Alternatively, please consider for the next commit implementing a `vector` configurable that includes both the pT limit and the required cut. You can loop over this configurable and reduce the size and amount of hard coding. This might be even better, and this way you can also have just one `PhiToKK` option.

PR #12905, opened 2025-09-08
[PWGHF] Improve the mass window selection and v0 type matched for lambda

zhangbiao-phy:
Hi @Luca610 and @fgrosa, I found some issues during doing the analysis of D0lambda channel, I try to fix it in this pr. Let me know if you have any comments on that.


alibuild (approved):

PR #12910, opened 2025-09-08
[Common] Refactor ft0CorrectedTable by removing unused processes

njacazio:
Removed the processStandard and processWithBypassFT0timeInMC functions, along with related logging and histogram additions. Updated the function signature to process.


njacazio:
On hold after discussing with @rbailhac

rbailhac:
Hi all, thanks for keeping us informed. "in any case the old version would be available via the previous O2Physics versions" is not really helping for us, because the format and information contained in our derived data is also changing. We need therefore latest O2Physics to produced derived data for MCs and at the same time this old bug fix for the old MCs. If you absolutely want to remove it, we will open our cuts artificially on old MCs to match with the expectations.

PR #12919, opened 2025-09-08
[PWGCF] Add plots with FT0C aplitude for UCC

ilikmeta:


victor-gonzalez:
I put a message in the enabling the merging but it seems it is not visible
The issue with `macOS-arm` is still not fixed. No PR have been merged today
Once it is fixed, the CI tests will slowly start over the open PRs
Usually this test is much quicker than the standard `OS` but in any case it will take a while
I enabled the auto-merging for have it ready for the CI tests succeed

PR #12926, opened 2025-09-09
[PWGCF] add cascades' run by run NUA correction

fuchuncui:


PWGCF/Flow/Tasks/flowGfwOmegaXi.cxx
```diff
@@ -79,6 +81,19 @@ std::shared_ptr<TProfile3D> omegac22[10];
...
+// std::shared_ptr<TH3D>histosPhiK0s[10];
+// std::shared_ptr<TH3D>histosPhiLambda[10];
+// std::shared_ptr<TH3D>histosPhiXi[10];
+// std::shared_ptr<TH3D>histosPhiOmega[10];
+// std::shared_ptr<TH3D>histosPhiK0scorr[10];
+// std::shared_ptr<TH3D>histosPhiLambdacorr[10];
+// std::shared_ptr<TH3D>histosPhiXicorr[10];
+// std::shared_ptr<TH3D>histosPhiOmegacorr[10];
+
+// std::shared_ptr<TH3D>nuaTH3[50];
```
victor-gonzalez, line 96: Pleas, don't leave code commented

victor-gonzalez (commented):
Using numbers (`magic numbers`) to address histograms is really dangerous, error prone, difficult of understand, hard to maintain and hard to evolve

Didn't work by having the run number in the `HistogramRegistry` histogram path?

Usually, unless the task is run in a concrete cluster or different test files from different run numbers are enforced, the task only addresses one single run number. If the histograms are defined having the actual run number in their paths, the merging will care of having the proper grouping by run number in the results file

fuchuncui:
> Using numbers (`magic numbers`) to address histograms is really dangerous, error prone, difficult of understand, hard to maintain and hard to evolve
> 
> Didn't work by having the run number in the `HistogramRegistry` histogram path?
> 
> Usually, unless the task is run in a concrete cluster or different test files from different run numbers are enforced, the task only addresses one single run number. If the histograms are defined having the actual run number in their paths, the merging will care of having the proper grouping by run number in the results file

Thanks for your comment.
In the function ".fill()", HIST name must be a const char and can not use HIST(Form()) (error: use of local variable with automatic storage from containing function). So I do need a ptr to fill these histos.
I have solved the magic numbers question by another method.

victor-gonzalez:
> > Using numbers (`magic numbers`) to address histograms is really dangerous, error prone, difficult of understand, hard to maintain and hard to evolve
> > Didn't work by having the run number in the `HistogramRegistry` histogram path?
> > Usually, unless the task is run in a concrete cluster or different test files from different run numbers are enforced, the task only addresses one single run number. If the histograms are defined having the actual run number in their paths, the merging will care of having the proper grouping by run number in the results file
> 
> Thanks for your comment. In the function ".fill()", HIST name must be a const char and can not use HIST(Form()) (error: use of local variable with automatic storage from containing function). So I do need a ptr to fill these histos. I have solved the magic numbers question by another method.

OK
But I guess that with your original structure, the one you commented and that is now gone, it should have worked because, presumably, you already stored `ptr`s

Anyway the named constants is by far better than the magic numbers
But have in mind that you have to allays double check the the naming matches with its purpose, no one else controls that, which is not the case with the previous, potentially, structure where you have the compiler help

victor-gonzalez (approved):
Not sure when it will get merged
The compiler errors are not from your code

PR #12929, opened 2025-09-09
[DPG] Added unit test for validating reconstruction software

alcaliva:


alcaliva:
Hi @f3sch,
thanks! We might also use that. I think that this code is sufficient to test randomness in the reconstruction software. If the pt distributions of the same events are exactly equal, then we can use this as a validation tool.

alcaliva:
Still, I think that quantifying the deviations due to this effect would be good.

alcaliva:
Hi, can this PR be approved? the failure of o2/macOS is not caused by this

ktf:
@singiamtel , I think you need to cleanup also this builder. It's probably the stale branch.

njacazio (approved):

PR #12943, opened 2025-09-10
Update femtoDreamCollisionSelection.h

RafRaf11:
Added a second sphericity histogram to check the sphericity distribution after the cut.


PWGCF/FemtoDream/Core/femtoDreamCollisionSelection.h
```diff
@@ -241,7 +242,9 @@ class FemtoDreamCollisionSelection
       spt = 2. * lambda2 / (lambda1 + lambda2);
     }
 
-    mHistogramRegistry->fill(HIST("Event/Sphericity"), spt);
+    mHistogramRegistry->fill(HIST("Event/Sphericity_before"), spt);
+    if (spt > mMinSphericity)
+      mHistogramRegistry->fill(HIST("Event/Sphericity_after"), spt);
```
lauraser, line 247: Hi @RafRaf11 ,
it seems to me a wrong place to do before and after histogras here as the cut is not applied here. In this part of the code, you just calculate the sphericity, which you later fill in collision tables. This is also done at the producer level and does not represent the cut that would be applied later in your analysis. From what I see in the code, this mMinSphericity is really not used for anything else besides filling this histogram. Why is it done here?

lauraser (changes requested):

PR #12947, opened 2025-09-10
[PWGDQ,PWGLF,DPG] TPC PID: fill clean kaon sample from Omega decay

lubynets:
Enabled filling a clean kaon sample for TPC PID NN training (in addition to existing samples of electrons, pions and protons).
* [PWGLF/DataModel/LFStrangenessTables.h](https://github.com/AliceO2Group/O2Physics/blob/b178c96d03ede873ee769ef8a4d7c1e21bf78332/PWGLF/DataModel/LFStrangenessTables.h): cascade table is extended with dynamic columns like in V0 table, i.e. Armenteros-Podolanski variables, Psi pair angle etc.
* [PWGDQ/Tasks/v0selector.cxx](https://github.com/AliceO2Group/O2Physics/blob/b178c96d03ede873ee769ef8a4d7c1e21bf78332/PWGDQ/Tasks/v0selector.cxx): implemented selection of $\Omega\rightarrow\Lambda K^-$ decays reconstructed by [PWGLF/TableProducer/Strangeness/strangenessbuilder.cxx](https://github.com/AliceO2Group/O2Physics/blob/b178c96d03ede873ee769ef8a4d7c1e21bf78332/PWGLF/TableProducer/Strangeness/strangenessbuilder.cxx). The cascade selection algorithm generally reproduces that of V0s;
* [DPG/Tasks/TPC/tpcSkimsTableCreator.cxx](https://github.com/AliceO2Group/O2Physics/blob/b178c96d03ede873ee769ef8a4d7c1e21bf78332/DPG/Tasks/TPC/tpcSkimsTableCreator.cxx): table with clean particle samples is extended with kaons - bachelor tracks of cascades selected at the previous step.

### DIFF
--- a/DPG/Tasks/TPC/tpcSkimsTableCreator.cxx
+++ b/DPG/Tasks/TPC/tpcSkimsTableCreator.cxx
@@ -107,8 +107,8 @@ struct TreeWriterTpcV0 {
   ctpRateFetcher mRateFetcher;
 
   /// Funktion to fill skimmed tables
-  template <bool doUseCorreceddEdx = false, typename T, typename C, typename V0>
-  void fillSkimmedV0Table(V0 const& v0, T const& track, C const& collision, const float nSigmaTPC, const float nSigmaTOF, const float dEdxExp, const o2::track::PID::ID id, int runnumber, double dwnSmplFactor, float hadronicRate)
+  template <bool doUseCorreceddEdx = false, typename T, typename C, typename V0Casc>
+  void fillSkimmedV0Table(V0Casc const& v0casc, T const& track, C const& collision, const float nSigmaTPC, const float nSigmaTOF, const float dEdxExp, const o2::track::PID::ID id, int runnumber, double dwnSmplFactor, float hadronicRate)
   {
 
     const double ncl = track.tpcNClsFound();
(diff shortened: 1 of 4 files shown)


romainschotter:
Hi @amaringarcia ! I completely agree with you. The changes in the strangeness data model should not interfere with anyone's work. 
From what I understand, the v0selector task is mainly used to obtain clean samples of various species: electrons from photon conversions, pions from K0s, protons from Lambda and kaons from Omegas. If that is correct, I am curious: wouldn’t it make more sense for the task to be located in LF or EM rather than in DQ?

amaringarcia:
Good morning @romainschotter, I think the reasons are historical, the J/psi analysers started very early on (pp @sqrt(s)= 7TeV) to use clean samples of electrons for electron PID efficiency. From my point of view, we could do a two stage move, the commit is approve as it is as we would like to add the kaons as soon as possible for the NN training, and in the mean time we check with the different players (DQ, LF,...) if a common v0selector can be centralized. Indeed many people use clean samples of identified particles and a common development stored in a central directory makes  sense.

romainschotter:
Hi @amaringarcia ! 
I’m not sure it’s necessary to have a centralized task for studying V0. My point is more about where the task should be located: as far as I understand, DQ focuses on quarkonia such as J/psi, which do not follow a V0 decay topology (since they don’t have a very displaced decay vertex).
In any case, I will not further delay the PR and I will approve the changes in the data model. 
Still, I would suggest considering moving the task to a directory more directly related to V0/cascade studies, such as LF or EM.

romainschotter (approved):

alcaliva (approved):

romainschotter:
> Hi @romainschotter, I think that unless we find a concrete reason for which to move this task to a different PWG directory, we should keep it here. In DQ, we are the main "consumer" of this task, so it makes sense that we maintain it and change it according to our needs.

Hi @iarsene!
Sorry for the confusion! I didn't mean to suggest moving the task. What I had in mind was rather making a copy of the task in another PWG. Anyway.

PR #12962, opened 2025-09-11
[PWGLF] Omegahm

lietava:
mult histos per run added
bool noSameBunchPileup added


romainschotter (dismissed):

lietava:
@romainschotter : thanks. (I thought there is some button on your side instaed of dummy :)

PR #12967, opened 2025-09-12
[Common] EventPlane: adding shift correction

jikim1290:
adding shift correction
based on the recent report in DPG meeting:
https://indico.cern.ch/event/1568257/contributions/6686897/attachments/3133733/5559683/DPG_AOT_EP_OO_and_Ne_Ne_110925.pdf

### DIFF
--- a/Common/TableProducer/qVectorsTable.cxx
+++ b/Common/TableProducer/qVectorsTable.cxx
@@ -18,34 +18,41 @@
 ///         (with or without corrections) and save the results in a dedicated table.
 ///
 
-// C++/ROOT includes.
-#include <chrono>
-#include <string>
-#include <vector>
-#include <unordered_map>
-#include <TComplex.h>
-#include <TH3F.h>
-#include <TProfile3D.h>
-
-// o2Physics includes.
-#include "Framework/AnalysisDataModel.h"
-#include "Framework/AnalysisTask.h"
-#include "Framework/runDataProcessing.h"
-#include "Framework/RunningWorkflowInfo.h"
-
 #include "Common/Core/EventPlaneHelper.h"
+#include "Common/DataModel/Centrality.h"
 #include "Common/DataModel/EventSelection.h"
 #include "Common/DataModel/FT0Corrected.h"
 #include "Common/DataModel/Multiplicity.h"
-#include "Common/DataModel/Centrality.h"
-
 #include "Common/DataModel/Qvectors.h"
-
-#include "Common/Core/TrackSelection.h"
 #include "Common/DataModel/TrackSelectionTables.h"
-// o2 includes.
-#include "CCDB/BasicCCDBManager.h"
-#include "DetectorsCommonDataFormats/AlignParam.h"
+
+#include <CCDB/BasicCCDBManager.h>
+#include <CommonConstants/MathConstants.h>
+#include <DetectorsCommonDataFormats/AlignParam.h>
+#include <FT0Base/Geometry.h>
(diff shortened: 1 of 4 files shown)


vkucera (changes requested):
Dear @jikim1290 , your changes are reverting my fixes made in https://github.com/AliceO2Group/O2Physics/pull/12833. Please remove your changes and be more careful when you rebase your branch.
Also please try to avoid invalid PR tags and typos in your PR titles.
It would also be more useful if you added more details in the description about what your PR does.

strogolo:
Dear @jikim1290 
thanks a lot for your PR and you don't need to add extra description since the content of this PR was deeply discussed at the DPG AOT meeting. So people are informed about this technical development and we all agree with this proposal.
The only doubt I have is related to the change at lines 100-101 in qVectorsTable.cxx, precisely where you are adding
`  Configurable<bool> cfgShiftCorr{"cfgShiftCorr", false, "configurable flag for shift correction"};
  Configurable<std::string> cfgShiftPath{"cfgShiftPath", "", "CCDB path for shift correction"};`
This means that we would have an additional object to upload to an additional path and to configure it properly, so increasing the possibility of errors, missing objects and misconfigurations.
1. Is it the only possible solution to duplicate adding a new path?
2. Have you checked that it is possible to run the frame even when there is not this object? Because there datasets for which we don't and will not have this correction.
3. I'd like to hear the opinion also from @ddobrigk 

Cheers,
Stefano

Common/TableProducer/qVectorsTable.cxx
```diff
@@ -176,7 +188,7 @@ struct qVectorsTable {
   void init(InitContext& initContext)
   {
     // Check the sub-detector used
-    const auto& workflows = initContext.services().get<RunningWorkflowInfo const>();
+    auto& workflows = initContext.services().get<RunningWorkflowInfo const>();
```
vkucera, line 191: Why did you remove the `const` qualifier?

Common/TableProducer/qVectorsTable.cxx
```diff
@@ -265,9 +277,22 @@ struct qVectorsTable {
...
+        fullPath += std::to_string(ind);
+        auto objshift = getForTsOrRun<TProfile3D>(fullPath, timestamp, runnumber);
+        shiftprofile.push_back(objshift);
+      }
+    }
+
     fullPath = cfgGainEqPath;
     fullPath += "/FT0";
-    const auto objft0Gain = getForTsOrRun<std::vector<float>>(fullPath, timestamp, runnumber);
+    auto objft0Gain = getForTsOrRun<std::vector<float>>(fullPath, timestamp, runnumber);
```
vkucera, line 295: Why did you remove the `const` qualifier?

Common/TableProducer/qVectorsTable.cxx
```diff
@@ -278,7 +303,7 @@ struct qVectorsTable {
 
     fullPath = cfgGainEqPath;
     fullPath += "/FV0";
-    const auto objfv0Gain = getForTsOrRun<std::vector<float>>(fullPath, timestamp, runnumber);
+    auto objfv0Gain = getForTsOrRun<std::vector<float>>(fullPath, timestamp, runnumber);
```
vkucera, line 306: Why did you remove the `const` qualifier?

jikim1290:
@strogolo this correction is implemented after applying previous full corrections. The input for shift correction is different from others and can not be done at the same stage. Therefore I think it is better to have independent configuration.
Please provide comments or merge this

PR #12978, opened 2025-09-14
[Common] Adjustments for ev sel service in Run 2 (prior to autodetect)

ddobrigk:
* adds an event selection service executable tailored to Run 2 that will not rely on the currently-to-be-fixed autodetect (@maciacco, this should help you not only to run, which you could anyway if you used the old event selection, but to not have to rely on the triple-DPL-device approach of the previous event selection, thus saving a lot of RAM).  
* adjusts the PID TPC Service such that a run 2 executable is invoked for Run 2 if needed and the default service does not include subscriptions to `Tracks`. It should be noted, however, that following the decision to anyhow save TPC Nsigmas to converted AO2D, this should not be used much anymore. As a side product, the removal of `Tracks` from the PIDTPCService subscriptions makes it possible to transparently enable the commented out parts of the propagation service (@romainschotter @gianniliveraro).

Will proceed to move `propagationService` to `Common` shortly: it is odd to invoke an LF task as a main common service at the moment.


alibuild (approved):

PR #12980, opened 2025-09-14
[Common] Propagation service moved to Common, tuned further

ddobrigk:
Moves the propagation service to common. Temporarily, `o2-analysis-lf-propagation-service` will coexist with `o2-analysis-propagation-service` but the former will be removed from both compilation and its current location (`PWGLF/TableProducers/Strangeness`) tomorrow, in favor of keeping only the task not identified as `lf`. 

This PR further enables the PID pre-selections for strangeness building and forks off an executable `propagationservice-Run2` following the rationale of `eventSelectionService` and `PIDTPCService`. Further cleanup to follow. 

Tagging @romainschotter @gianniliveraro


alibuild (approved):

PR #12989, opened 2025-09-15
[PWGHF] Refactor data creator for HF correlations

Marcellocosti:
Tests on hy with the current code produce very large outputs due to the large track multiplicity in OO collisions, especially for the D0 and D+ mesons. Therefore, in this PR, the data model is reshaped: separate process functions are implemented for the creation of Same Event and Mixed Event derived datasets, introducing pt- and centrality dependent down sample factors to handle the event mixing output size and the case of hadron-hadron correlations. This implementation also allows to get rid of the index columns in the output, as daughter-candidate pairs are rejected online.

Tagging @stefanopolitano, @scattaru, @wuctlby.


vkucera:
Converted to draft because of compilation errors.

PR #12991, opened 2025-09-16
[PWGUD] New Task : Incoherent JPsi Polarization with Run 3 Dataset

nivram-phy:
Basic task to create histograms and TTree on mass, pT, eta and phi


nivram-phy:
Hello @rolavick,
I fixed most of the errors. But it looks like it expects a UpperCamelCase for the name. I have modified the name from upcPolarisationJPsiIncorr.cxx to upcPolarisationJpsiIncorr.cxx. But kept "upc" as is, similar to other codes in the tasks folder. Would you like that also changed to "Upc"?

nivram-phy:
Hello @rolavick,
1) Sorry if I am missing soemthing. The name and the struct names does start with a lower case "upc...."
2) Ah! Incorr is just me spelling Incoherent badly. I can fix that overall, once I understnad the above issue.

rolavick:
Hi @nivram-phy ,
the rule is the struct should start with Upper case, while the file name is lower case.
Roman

nivram-phy:
@rolavick I think I have fixed all of the errors that you mentioned. Thanks for the explanation.

rolavick (approved):

PWGUD/Tasks/CMakeLists.txt
```diff
@@ -230,7 +235,7 @@ o2physics_add_dpl_workflow(upc-quarkonia-central-barrel
                            COMPONENT_NAME Analysis)
 
 o2physics_add_dpl_workflow(test-mc-std-tabs-rl
-                           SOURCES testMcStdTabsRl.cxx
+                           SOURCES testMCstdTabsRL.cxx
```
rolavick, line 238: I just noticed, why have you changed this? It will break the code.

PWGUD/Tasks/CMakeLists.txt
```diff
@@ -253,8 +258,3 @@ o2physics_add_dpl_workflow(analysis-mc-dpm-jet-sg-v3
                            SOURCES analysisMCDPMJetSGv3.cxx
                            PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::AnalysisCore
                            COMPONENT_NAME Analysis)
-
-o2physics_add_dpl_workflow(upc-test-rct-tables
```
rolavick, line 257: Also here, why have you deleted this whole block? It will break some hyperloop wagons at minimum...

rolavick (changes requested):

PWGUD/Tasks/CMakeLists.txt
nivram-phy: Strange, I don't remeber touching this at all. I am trying to trace this, but can I just change it back?

rolavick: Could be when you rebased you rebased to some older tag? Anyway, please synchronise it with the current tag and apply your changes on top of that.

rolavick (approved):

PR #13025, opened 2025-09-18
Add run info option and reorganize includes

njacazio:
Added option to include run info in histograms and updated includes.


njacazio:
@mario-krueger

PR #13038, opened 2025-09-19
fix absolute value bug for dca and pt cut

blacwovie:


vkucera:
You still have `abs` bugs in the code. See the O2 linter report.

maciacco (approved):

PR #13042, opened 2025-09-19
Update a3 geo in fasttracker

jesgum:
- Update x0 values for inert layers
- Add iTOF & coldplate
- Apply scale to VD x0
- I'm not sure about the xrho values in this setup so they are left at 0 and I'll run with e-loss disabled for now as it has previously shown to have a minimal, if any, effect. Will update and add the values as soon as I have them


njacazio:
Ciao @jesgum very good, many thanks! We should see that this gets taken from an external file fetched via the CCDB. As we approach the geometry configuration we want to change between geometries without touching the code.

PR #13086, opened 2025-09-22
[ALICE3] Add possibility to use generic geometry

njacazio:
@jesgum


jesgum:
Super! Thanks a lot @njacazio

PR #12488, opened 2025-08-08
[DPG] more checks in lightIonsEvSelQa

altsybee:

PR #13097, opened 2025-09-23
[Infrastructure] Enable more categories in Cppcheck

vkucera:
In the default configuration, cppcheck (which runs as one of the MegaLinter PR checks) only reports the most critical bugs, such as out of bound access, memory leaks, null pointer dereferences and uninitialised data members. The current codebase has around 90 such reported bugs.
With all checks enabled, cppcheck finds significantly more bugs, potential bugs (such as useless, duplicated, nonsensical, or unreachable code), inefficient code, and many other issues that would otherwise be considered compilation warnings if found by the compiler.
With this configuration, cppcheck has found over 4000 issues in the current codebase among which hundreds are genuine bugs, other hundreds are suspicious constructs (which are likely bugs too), and hundreds more are inefficient constructs.
Despite the high number of true positives the number of false positives stays very low.

EDIT: Instructions for local execution here: https://github.com/AliceO2Group/analysis-framework/pull/313

EDIT [2026-04-12]: Reopen the PR with updates.

- Extend suppression list with other harmless and very frequent errors (more can always be added later if needed). Reduces the total number of issues to around 2000.
- Add the `--inline-suppr` option to allow inline suppressions. (format: `// cppcheck-suppress warningId`)

Here are some recent examples of detected bugs:

- [AliceO2Group/O2Physics#15605](https://github.com/AliceO2Group/O2Physics/pull/15605#discussion_r3068323311)
- [AliceO2Group/O2Physics#14279](https://github.com/AliceO2Group/O2Physics/pull/14279#discussion_r3068340197)
- [AliceO2Group/O2Physics#13419](https://github.com/AliceO2Group/O2Physics/pull/13419#discussion_r3068344370)
- [AliceO2Group/O2Physics#12488](https://github.com/AliceO2Group/O2Physics/pull/12488#discussion_r3068358626)


ktf:
I think until the current reports are all fixed, we should not introduce further checks.

vkucera:
> I think until the current reports are all fixed, we should not introduce further checks.

They have been fixed for a few months.

dsekihat (approved):

PR #13113, opened 2025-09-24
[ALICE3] Implementation of tracker PID with ToT

hfribert:
This PR adds the use of tracker PID using ToT (Time-over-Threshold) to the OTF Fast simulation using LUTs, including light nuclei such as deuteron, triton, helium-3, and helium-4/alphas.

### DIFF
--- a/ALICE3/TableProducer/OTF/onTheFlyTracker.cxx
+++ b/ALICE3/TableProducer/OTF/onTheFlyTracker.cxx
@@ -113,7 +113,7 @@ struct OnTheFlyTracker {
   Configurable<std::string> lutDe{"lutDe", "lutCovm.de.dat", "LUT for deuterons"};
   Configurable<std::string> lutTr{"lutTr", "lutCovm.tr.dat", "LUT for tritons"};
   Configurable<std::string> lutHe3{"lutHe3", "lutCovm.he3.dat", "LUT for Helium-3"};
-  Configurable<std::string> lutAl{"lutAl", "lutCovm.he3.dat", "LUT for Alphas"}; // To be created, for now propagating as He-3
+  Configurable<std::string> lutAl{"lutAl", "lutCovm.he3.dat", "LUT for Alphas"};
 
   struct : ConfigurableGroup {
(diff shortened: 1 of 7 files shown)


ALICE3/TableProducer/OTF/onTheFlyTracker.cxx
```diff
@@ -113,6 +113,7 @@ struct OnTheFlyTracker {
   Configurable<std::string> lutDe{"lutDe", "lutCovm.de.dat", "LUT for deuterons"};
   Configurable<std::string> lutTr{"lutTr", "lutCovm.tr.dat", "LUT for tritons"};
   Configurable<std::string> lutHe3{"lutHe3", "lutCovm.he3.dat", "LUT for Helium-3"};
+  Configurable<std::string> lutAl{"lutAl", "lutCovm.he3.dat", "LUT for Alphas"};
```
njacazio, line 116: Ciao @hfribert is this used then? How are He4 propagated? The PDG of the he3 is not passed in line 883.

hfribert: Hi Nicolo, thanks for spotting that it's missing the proper propagation. I added back the loadLUT for alphas, and changed it now in the DelphesO2TrackSmearer.h accordingly.

Since the PDG code comes from the LUT header, I also made a change in the DelphesO2TrackSmearer.cxx (quite ugly if-else statement) to avoid fatal errors. But I am quite unsure if there is not a much cleaner way of dealing with this in the end. So any recommendations are welcome

hfribert:
@njacazio, I suppose I also need to then add ccdb support to the trackerPID task to load the ToT LUTs from alice-ccdb? This I will check then tomorrow

njacazio:
Ciao @hfribert I think it looks good, ready to merge?

hfribert:
> Ciao @hfribert I think it looks good, ready to merge?

Hi, tested it locally and seems fine to me, ccdb also seems to work

njacazio (approved):

PR #13120, opened 2025-09-24
[DPG] added condition on ambi tracks loop to avoid crashes

altsybee:
added condition on ambi tracks loop to avoid crashes (in particular, for some Pb-Pb MC)


alibuild (approved):

PR #13149, opened 2025-09-26
Added MC truth process

ananthapadmanabhan18:
- Added MC truth process


PWGUD/Tasks/exclusiveRhoTo4Pi.cxx
```diff
@@ -31,6 +31,8 @@
 #include "Math/Vector3D.h"
 #include "Math/Vector4D.h"
 #include "TPDGCode.h"
+#include <TCanvas.h>
+#include <TH1F.h>
```
rolavick, line 35: What do you need these two new includes for?

ananthapadmanabhan18: Oh, I was actually trying something, but then I forgot to remove it. I will remove it and update the PR

PR #13183, opened 2025-09-30
[PWGCF] adding shift correction for event plane

jikim1290:
adding shift correction for event plane
adding scalar product method


victor-gonzalez:
Please, in the new introduced code, follow `linter` recommendations

jikim1290:
Let me close this first and the other will open a new one

PR #13187, opened 2025-09-30
[ALICE3] Create a3geo.ini with ALICE3 geometry settings

njacazio:
Added initial ALICE3 geometry configuration in TEnv format.


alibuild (approved):

PR #13205, opened 2025-10-01

ariedel-cern:
Features added:
- derived to derived data prodcuer for tracks, lambda and k0shorts
- making the data model for collision similar
- adding more features to the collision selections via bitmasks
- overall small fixes and improvements


alibuild (approved):

PR #13232, opened 2025-10-04
[PWGHF] add a flag for the tables filling into charm femto task

zhangbiao-phy:


alibuild (approved):

PR #13239, opened 2025-10-05
[Common] Add FT0C variant 2 (Run 2 like truncation)

ddobrigk:
@abmodak - I will still do some testing but here is the tool for your cross-check


ddobrigk:
@abmodak this works as intended, now proceeding with the merge

alibuild (approved):

PR #13240, opened 2025-10-05
[PWGHF,PWGCF] fix the issue of int configurable caste to int8_t and add QA plot of charm mass vs pt in MC rec

zhangbiao-phy:
The issue is It will always use the default value when the int configurable caste to int8_t in Filter or Partition

### DIFF
--- a/PWGHF/HFC/Tasks/taskCharmHadronsFemtoDream.cxx
+++ b/PWGHF/HFC/Tasks/taskCharmHadronsFemtoDream.cxx
@@ -583,7 +583,7 @@ struct HfTaskCharmHadronsFemtoDream {
         part.tpcNSigmaPr(),
         part.tofNSigmaPr());
     }
-    if (sliceCharmHad.size() || sliceTrk1.size()) {
+    if (sliceCharmHad.size() > 0 || sliceTrk1.size() > 0) {
       rowFemtoResultColl(
         col.globalIndex(),
         timeStamp,
@@ -632,7 +632,7 @@ struct HfTaskCharmHadronsFemtoDream {
     auto sliceMcTrk1 = partitionMcTrk1->sliceByCached(aod::femtodreamparticle::fdCollisionId, col.globalIndex(), cache);
     auto sliceMcCharmHad = partitionMcCharmHadron->sliceByCached(aod::femtodreamparticle::fdCollisionId, col.globalIndex(), cache);
 
-    if (sliceMcCharmHad.size()) {
+    if (sliceMcCharmHad.size() > 0) {
       for (auto const& part : sliceMcCharmHad) {
         registryCharmHadronQa.fill(HIST("CharmHadronQA/hPtVsMass"), part.pt(), getCharmHadronMass(part));
       }


PWGHF/HFC/Tasks/taskCharmHadronsFemtoDream.cxx
```diff
@@ -583,7 +583,7 @@ struct HfTaskCharmHadronsFemtoDream {
         part.tpcNSigmaPr(),
         part.tofNSigmaPr());
     }
-    if (!sliceCharmHad.size() || !sliceTrk1.size()) {
+    if (sliceCharmHad.size() || sliceTrk1.size()) {
```
vkucera, line 586: `size` does not return `bool`. Use explicit comparison with `0`.

PWGCF/DataModel/FemtoDerived.h
```diff
@@ -245,14 +245,14 @@ DECLARE_SOA_COLUMN(Prong2Eta, prong2Eta, float);                    //! Track et
 DECLARE_SOA_COLUMN(Prong0Phi, prong0Phi, float);                    //! Track phi of charm hadron prong0
 DECLARE_SOA_COLUMN(Prong1Phi, prong1Phi, float);                    //! Track phi of charm hadron prong1
 DECLARE_SOA_COLUMN(Prong2Phi, prong2Phi, float);                    //! Track phi of charm hadron prong2
-DECLARE_SOA_COLUMN(CandidateSelFlag, candidateSelFlag, int8_t);     //! Selection of mass hypothesis for charm hadron (1 for Lc -> pkpi, 2 for Lc -> pikp, 4 for D+ -> pikpi)
+DECLARE_SOA_COLUMN(CandidateSelFlag, candidateSelFlag, int);        //! Selection of mass hypothesis for charm hadron (1 for Lc -> pkpi, 2 for Lc -> pikp, 4 for D+ -> pikpi)
```
vkucera, line 248: I believe that `int8_t` is the only integral type affected by this issue. Why not use `int16_t`?

PWGHF/HFC/Tasks/taskCharmHadronsFemtoDream.cxx
zhangbiao-phy: Thanks! fixed it!

PWGCF/DataModel/FemtoDerived.h
vkucera: Sorry, but I couldn't find a line where additional casting would be needed. Can you refer them?

shouqiye (approved):

PWGCF/DataModel/FemtoDerived.h
vkucera: I don't see why `charmHadMcSel` has to be `int`.

zhangbiao-phy: flagMc was `int_8`, so Configurable in charm femto task is:
`Configurable<int> charmHadMcSel{"charmHadMcSel", DecayChannelMain::LcToPKPi, "charm hadron selection for mc, DplusToPiKPi = 1, LcToPKPi = 17"};`
here caste Configurable `int` to `int8_t`
Filter hfMcSelFilter = nabs(aod::fdhf::flagMc) == static_cast<int8_t>(charmHadMcSel.value);
It looks good, but it always takes the default value casted to int8_t, So that Configurable is useless. 
Then the directly way is: align `flagMc` also to `int`. I think this issue also reported from someone in the O2 Analysis channel before

vkucera: Sorry but that does not answer my question.

zhangbiao-phy: sorry, I don't know why it should be `int16_t`? can you explain more?

zhangbiao-phy: > Your solution is to promote all concerned variables to `int`. I am proposing to make them the smallest working integral type, which AFAIK is `int16_t` .

 I agree with your point and I also want to improve it with this point. However as I mentioned above, I already tried `int16_t`. It doesn't work (seems not support). So could you please approve it with current version?

zhangbiao-phy: I already tried both case. it is not the case.

zhangbiao-phy:
> @zhangbiao-phy @shouqiye @ariedel-cern @lauraser Are we sure this does not break backward compatibility with already produced derived data?

Hi @victor-gonzalez, sorry for confused you again! I think the changes only related to the HF part, we don't produce derived data for common used, since there are only two analysis (pLc and pDplus) used such tables.

PR #13281, opened 2025-10-08
[PWGCF] Clean up event selection in femto framework

ariedel-cern:
Promote trigger selection to be part of the bitmask for collision selection.
With this change, multiple triggers can be selected by the producer task, and then specific triggers (via bitmask) can be used in analysis.


alibuild (approved):

PR #13287, opened 2025-10-08
Cleanup and bug fixes using cppcheck and clang-tidy

vkucera:


PWGHF/D2H/Tasks/taskCharmPolarisation.cxx
```diff
@@ -1576,7 +1576,7 @@ struct HfTaskCharmPolarisation {
       // variable definition
       float pxDau{-1000.f}, pyDau{-1000.f}, pzDau{-1000.f};
       float pxCharmHad{-1000.f}, pyCharmHad{-1000.f}, pzCharmHad{-1000.f};
-      float massDau{0.f}, invMassCharmHad{0.f}, invMassCharmHadForSparse{0.f}, invMassD0{0.f}, invMassKPiLc{0.f}, invMassPKLc{0.f}, invMassPPiLc{0.f};
+      double massDau{0.}, invMassCharmHad{0.}, invMassCharmHadForSparse{0.}, invMassD0{0.}, invMassKPiLc{0.}, invMassPKLc{0.}, invMassPPiLc{0.};
```
fgrosa, line 1579: Regarding this, it is not clear to me what is the benefit of having the mass as `double` when we compute it starting from `float` values (momenta and pdg masses).
Wouldn't it be better to have the function in `RecoDecay` returning the same type as the inputs instead of always `double`?
https://github.com/AliceO2Group/O2Physics/blob/689968149d4069e09f982edb23834b4e45d9a67b/Common/Core/RecoDecay.h#L449

PWGHF/Macros/computeFonllPlusPythiaPredictions.C
```diff
@@ -109,7 +109,7 @@ std::vector<std::string> splitString(const std::string& str, char delimiter)
 }
 
 //__________________________________________________________________________________________________
-std::array<TH1D*, 3> readFonll(std::string inFile, std::string histName)
+std::array<TH1D*, 3> readFonll(const std::string& inFile, const std::string& histName)
```
fgrosa, line 112: while I understand that it's a good practice, this is a macro, and this function is superfast anyway since it's executed 3 times before the simulation itself, that is the slow part :-)

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -958,10 +958,7 @@ struct HfTrackIndexSkimCreatorTagSelTracks {
         }
         if (config.debugPvRefit) {
           LOG(info) << "### vecPvContributorGlobId.size()=" << vecPvContributorGlobId.size() << ", vecPvContributorTrackParCov.size()=" << vecPvContributorTrackParCov.size() << ", N. original contributors=" << collision.numContrib();
-        }
-
-        /// Perform the PV refit only for tracks with an assigned collision
-        if (config.debugPvRefit) {
+          /// Perform the PV refit only for tracks with an assigned collision
```
fgrosa, line 961: I think this is more for @mfaggin than me (but it looks good to me)

PWGHF/Tasks/taskMcValidation.cxx
```diff
@@ -976,8 +976,7 @@ struct HfTaskMcValidationRec {
         }
         float deltaZ = -999.f;
         if (index) {
-          auto collision = track.collision_as<CollisionsWithMCLabels>();
-          auto mcCollision = particle.mcCollision_as<aod::McCollisions>();
```
fgrosa, line 980: right, thanks!

vkucera: Thanks for checking.

PR #13336, opened 2025-10-10
[Common] Adding overflow short-circuit

ChSonnabend:
This avoids index overflows encountered in my AO2Ds

### DIFF
--- a/Common/TableProducer/occupancyTableProducer.cxx
+++ b/Common/TableProducer/occupancyTableProducer.cxx
@@ -2708,15 +2708,17 @@ struct CreatePointerTables {
     // create pointer table
     int currentIDXforCheck = 0;
     int listSize = trackGIForTrackQAIndexList.size();
+    bool breakOnOverflow = false;
 
     for (const auto& track : tracks) {
-      while (track.globalIndex() > trackGIForTrackQAIndexList[currentIDXforCheck][0]) {
+      while (!breakOnOverflow && track.globalIndex() > trackGIForTrackQAIndexList[currentIDXforCheck][0]) {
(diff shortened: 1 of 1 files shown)


rahulverma012:
Hi @ChSonnabend, @fgrosa,
I've reviewed your changes, and they are correct — the task now works as intended.
@ddobrigk, could you please approve the PR?

Thank you

ChSonnabend:
> > > Hi @rahulverma012 I think that this fixes the longstanding issue with trackQA indices, can you check if it's fine?
> > 
> > 
> > At least its a workaround for now avoiding segfaults. Ultimately its better to find the root cause why the indices can go out of bounds, but I'm no expert here (and as mentioned before, I didn't investigate any further...). Once Rahul agrees maybe we could have it merged @ddobrigk ? (CI error seems unrelated)
> 
> To explain: Let's say the Tracks Table has size N and the TrackQA Table has size M. Initially, the task stores the global indices from the TrackQA table into an indexList vector. Then, it iterates over the Tracks table, checking whether each track exists in indexList. If it does, it stores the corresponding value in a new table.If not, it assigns -1. A while loop was used to keep the lookup bounded within N + M iterations. However, when all M entries from indexList had been processed, and there were still tracks remaining without assignment, the loop would continue and eventually access beyond the bounds of the vector—causing a segfault. Thanks to your out-of-bounds check, this issue is now avoided, and the behavior is now as intended—no crashes.
> 
> Thank you

Excellent, thank you! This was my suspicion and the reason why I implemented it in this way. I was just not sure if there was a smarter away to avoid this workaround. But from what I understand the size of the two tables will realistically always mismatch. In this case this is probably the best possible solution. @ddobrigk ready for the merge from my side then too 👍

PR #13395, opened 2025-10-15
[Common] Centrality: further protection against MC quadratic mapping overshooting 

ddobrigk:
Further adds a level of protection for events with unassigned FIT signals, which interfere destructively in very rare events in OO simulation and lead to a situation in which the quadratic multiplicity mapping produces unintended results. @romainschotter thanks for reporting!


alibuild (approved):

vkucera:
The numerous magic numbers are bug-prone and make the code quite difficult to maintain.
Also the VLA should be fixed.

PR #13400, opened 2025-10-15
[ALICE3] Update LUT configurations to default empty values for nuclei

njacazio:


alibuild (dismissed):

alibuild (approved):

jesgum:
Many thanks @njacazio! Some of these changes i've included in https://github.com/AliceO2Group/O2Physics/pull/13325

Specifically for the nuclei, where instead the loading of the luts is behind the existing treat nuclei flag

That pr also introduces the option to use the fassttracker I stead which is a feature that I think we want. Should I reopen the pr with only the fasttracker changes or how do you want to proceed?

alibuild (approved):

PR #13401, opened 2025-10-15
[Common,PWGLF] Task to study ZDC response and calibration in light ion data taking

coppedis:
We've added a task and a table to better understand and check the ZDC signals in the pO, OO and Ne-Ne runs.

### DIFF
--- a/PWGLF/TableProducer/CMakeLists.txt
+++ b/PWGLF/TableProducer/CMakeLists.txt
@@ -9,6 +9,7 @@
 # granted to it by virtue of its status as an Intergovernmental Organization
 # or submit itself to any jurisdiction.
 
+add_subdirectory(QC)
 add_subdirectory(Common)
 
 add_subdirectory(Nuspex)


Common/TableProducer/zdcTaskLightIons.cxx
```diff
@@ -0,0 +1,344 @@
...
+#include "Common/DataModel/Multiplicity.h"
+#include "Common/DataModel/ZDCLightIons.h"
+
+#include "Framework/AnalysisDataModel.h"
+#include "Framework/AnalysisTask.h"
+#include "Framework/HistogramRegistry.h"
+#include "Framework/runDataProcessing.h"
+
+#include "TH1F.h"
+#include "TH2F.h"
```
vkucera: Please include external headers with `<>`.

coppedis:
If I am not totally wrong, all the remaining linter issues are related to CMakeLists.txt of Common/TableProducer.
If I am not overlooking something, may I ask @ddobrigk to let me know whether we could merge this PR?

Thanks in advance for your help and understanding.
ch.

coppedis:
> Hi @coppedis , I think the change to the `CMakeLists.txt` of the PWGMM directory is accidental, right? Could you please remove that change from the PR? I will then approve. Thank you!

Dear @ddobrigk this was a fix (newline at the end of the file) that I erroneously put in  and Vit kindly asked me to remove. What do you want me to do? Leave it as approved by Vit or revert to some even previous version? Please, let me know.
Thanks a lot for your help

ddobrigk:
Hi @coppedis, I think the proper course of action is to leave that CMakeLists file totally unchanged, since you added a task only to `Common` - it looks confusing that this PR tags PWGLF when in fact you (from what I understand?) only fix one linter issue or so there and it is totally unrelated to the actual work in this PR. To proceed, can you please just take the CMakeLists as it is in the O2Physics master and use that in this PR? If really needed, we can adjust newlines there with a separate PR. Thank you!

PR #13417, opened 2025-10-16
[Common] Rename ZDCInterCalibTask and update it

udmitrie:
Rename workflow for clarity: zdc-task-inter-calib -> zdc-extra-table-producer;
Fix o2-linter connected problems and typos;
Add centroid calculations, centroid histos and new columns to the produced table.


udmitrie:
Dear @ddobrigk , please, take a look at this PR, which is required for further spectator plane measurements with ZDC.

PR #13418, opened 2025-10-16
[PWGJE] Updated GNN b-jet analysis histograms, applied proper event selections

choich08365:


nzardosh:
approving this, but tagging @hahassan7 to take a look too

PR #13422, opened 2025-10-16
[PWGCF] Addition of resonance correlations

Preet-Bhanjan:


PWGCF/TwoParticleCorrelations/Tasks/pidDiHadron.cxx
```diff
@@ -311,7 +496,86 @@ struct PidDiHadron {
...
+
+    bool isPion, isKaon, isProton;
+    bool isDetectedPion = nSigmaToUse[0] < detectorNsigmaCut[0] && nSigmaToUse[0] > detectorNsigmaCut[0 + 3];
+    bool isDetectedKaon = nSigmaToUse[1] < detectorNsigmaCut[1] && nSigmaToUse[1] > detectorNsigmaCut[1 + 3];
+    bool isDetectedProton = nSigmaToUse[2] < detectorNsigmaCut[2] && nSigmaToUse[2] > detectorNsigmaCut[2 + 3];
+
+    bool isTofPion = nSigmaTOF[0] < tofNsigmaCut[0] && nSigmaTOF[0] > tofNsigmaCut[0 + 3];
+    bool isTofKaon = nSigmaTOF[1] < tofNsigmaCut[1] && nSigmaTOF[1] > tofNsigmaCut[1 + 3];
+    bool isTofProton = nSigmaTOF[2] < tofNsigmaCut[2] && nSigmaTOF[2] > tofNsigmaCut[2 + 3];
+
```
victor-gonzalez: Please, use named constants instead of 'magic' number as indexes

PWGCF/TwoParticleCorrelations/Tasks/pidDiHadron.cxx
```diff
@@ -55,17 +57,108 @@ using namespace o2::framework::expressions;
...
+auto readMatrix(Array2D<T> const& mat, P& array)
+{
+  for (auto i = 0; i < static_cast<int>(mat.rows); ++i) {
+    for (auto j = 0; j < static_cast<int>(mat.cols); ++j) {
+      array[i][j] = mat(i, j);
+    }
+  }
+
+  return;
+}
```
victor-gonzalez: This is really dangerous
The vectors structures does not have checks on their range which means that if they read or written out of their bounds most of the times it will not be noticed but other data is being read or overwritten which is really hard to fix and produces random results
Will it not work passing references to empty arrays of the right type and then filling the array with the corresponding elements?
Also named constants can be used to directly access the elements of the matrix without the need of serializing its content

PWGCF/TwoParticleCorrelations/Tasks/pidDiHadron.cxx
Preet-Bhanjan: Thanks for pointing it out. I will be using named constants from now on

PR #13433, opened 2025-10-17
[PWGCF] Addition of named constants in PID function

Preet-Bhanjan:


PWGCF/TwoParticleCorrelations/Tasks/pidDiHadron.cxx
```diff
@@ -55,17 +57,109 @@ using namespace o2::framework::expressions;
...
+    kMassBins,
+    kDCABetDaug,
+    kUseProperLifetime,
+    kUseV0Radius,
+    kNParticleSwitches
+  };
+  enum Particles {
+    PIONS = 0,
+    KAONS,
+    PROTONS
```
victor-gonzalez: I still believe this is inconsistent a dangerous
Depending on the part of the code you look at this assignment is true or is false
I understand you try to use a single configurable to address identified or not particles and which of the identified particle is of interest but this clearly creates an inconsistency between parts of your code
I would suggest trying to address that situation in a different way

PR #13436, opened 2025-10-17
[PWGHF] Fix potential bug in Dstar candidate selection in debug mode 

zhangbiao-phy:
Fix potential bug in Dstar candidate selection in debug mode 

- whatever enables debug or not, we should return 0 when bin pt is not in the required pt range


alibuild (approved):

PR #13466, opened 2025-10-20
[PWGCF] FemtoUniverse: Add flag and condition to select only primaries

prchakra:


alibuild (approved):

PR #13489, opened 2025-10-22
[PWGHF,Trigger] Fix vertexer configuration in HF filters

fgrosa:
Thanks @vkucera for spotting it!
@apalasciano @zhangbiao-phy for info


alibuild (approved):

PR #13493, opened 2025-10-22
[Common] TrackTuner: dca calib. "autodetection" based on run number.

mfaggin:
In this PR I've implemented the possibility to enable an "auto-detection" of dca calibration files for the TrackTuner from CCDB based on the run number.
This is based on a list of predefined cases (look at the function `TrackTuner::getPathInputFileAutomaticFromCCDB()`), for which now we are forced to define different versions of the `propagationService` wagon (e.g. in the HF service wagons list in Hyperloop).
The development has been tested locally for a MC production anchored to pp 2023 data t 13.6 TeV, and the output before and after this development looks the same.

I open this PR in draft mode for the moment, waiting for your possible comments, especially on the following points:

1. I have implemented the usage of auto-detection only for the `propagationService` case, i.e. I have not done anything similar for the `Common/TableProducer/trackPropagation.cxx` workflow. Let me know if I need to extend it there as well (tagging @ddobrigk for it)

2. to use the auto detection, one has to slightly change the way the function `TrackTuner::getDcaGraphs()` is called. I implemented it only for the `propagationService`, but there are other codes where the `TrackTuner` is instantiated. Below the list of cases (hoping not to forget anything). Since they are not central workflows, but analysis-related ones, I'd leave it to the analysers, in case they need it (I tag here people that developed the codes, or possibly are responsible for them @iouribelikov @f3sch @lbariogl @njacazio @romainschotter @creetz16 @GiorgioAlbertoLucia @fmazzasc ):

- `DPG/Tasks/AOTTrack/V0Cascades/perfK0sResolution.cxx`
- `DPG/Tasks/AOTTrack/V0Cascades/qaLamMomResolution.cxx`
- `PWGLF/TableProducer/QC/nucleiQC.cxx`
- `PWGLF/TableProducer/Nuspex/nucleiFlowTree.cxx`

Tagging for their info also @arossi81 @phymanshu @fgrosa @gluparel

### DIFF
--- a/Common/Tools/TrackPropagationModule.h
+++ b/Common/Tools/TrackPropagationModule.h
@@ -213,9 +213,9 @@ class TrackPropagationModule
   void fillTrackTables(TConfigurableGroup const& cGroup, TrackTuner& trackTunerObj, TCCDBLoader const& ccdbLoader, TCollisions const& collisions, TTracks const& tracks, TOutputGroup& cursors, THistoRegistry& registry)
   {
 
-    /// retrieve the graphs *if not done yet*
+    /// retrieve the TrackTuner calibration graphs *if not done yet*
     /// i.e. if autodetect is required
-    if (autoDetectDcaCalib && !trackTunerObj.areGraphsConfigured) {
+    if (cGroup.useTrackTuner.value && autoDetectDcaCalib && !trackTunerObj.areGraphsConfigured) {
 
       /// get the run number from the ccdb loader, already initialized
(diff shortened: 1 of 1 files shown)


ddobrigk:
Hi @mfaggin - thanks, looks good to me! Then for the future I wonder if one could not make something a little bit more automated in terms of run ranges (directly using the CCDB instead of a list of "ifs" for the run list ranges), but that is optional: just a thought :-) . Approving!

PR #13502, opened 2025-10-22
[PWGCF] fix possible deref of nullptr error message

EmilGorm:


PWGCF/GenericFramework/Tasks/flowGfwLightIons.cxx
vkucera: Looks like you're introducing a couple of new memory leaks...

PR #13503, opened 2025-10-22
[PWGHF,PWGJE,Trigger] Divide candidate data model into skimming, reconstruction, aliases

vkucera:
Move parts of `PWGHF/DataModel/CandidateReconstructionTables.h` into:

- `PWGHF/DataModel/CandidateSkimmingTables.h`
  - Tables and flags used in the track index skimming
- `PWGHF/DataModel/AliasTables.h`
  - Definitions of joined table types
- `PWGHF/Core/DecayChannelsLegacy.h`
  - Legacy MC flags that should be replaced. 
- `PWGHF/ALICE3/Core/DecayChannelsLegacy.h`
  - Legacy MC flags used only for ALICE3. 

The goal is to factorise header dependencies and to reduce recompilations when modifying parts of the data model.

### DIFF
--- a/EventFiltering/PWGHF/HFFilter.cxx
+++ b/EventFiltering/PWGHF/HFFilter.cxx
@@ -23,7 +23,7 @@
 #include "EventFiltering/filterTables.h"
 //
 #include "PWGHF/Core/SelectorCuts.h"
-#include "PWGHF/DataModel/CandidateSkimmingTables.h"
+#include "PWGHF/DataModel/TrackIndexSkimmingTables.h"
 //
 #include "PWGEM/PhotonMeson/DataModel/gammaTables.h"
 #include "PWGLF/DataModel/LFStrangenessTables.h"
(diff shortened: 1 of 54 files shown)


PWGHF/DataModel/CandidateSkimmingTables.h
```diff
@@ -0,0 +1,323 @@
...
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+/// \file CandidateSkimmingTables.h
+/// \brief Definitions of tables produced by the candidate skimming
+///
+/// \author Gian Michele Innocenti <gian.michele.innocenti@cern.ch>, CERN
+/// \author Vít Kučera <vit.kucera@cern.ch>, CERN
+
+#ifndef PWGHF_DATAMODEL_CANDIDATESKIMMINGTABLES_H_
```
fgrosa: I would refrain from using the term "skimming" to avoid confusion, since in ALICE skimming is referring to the software triggers (I know that we called "skimCreator" our HF vertex finder, but this would also not be a proper name, since we are not skimming anything, but actually adding information to the AO2Ds). I propose to use the `TrackIndexSkimCreatorTables.h` or something similar, to make it a bit more clear that it refers to the `trackIndexSkimCreator` task.

fgrosa: I think it's not "skimming" because if it was it, it would imply a reduction of the information while in our case we add the information of candidates starting from tracks (so it's vertex finding, similar to the SVertexer in the reconstruction). In any case I think that `TrackIndexSkimmingTables` makes it less ambiguous, thanks!

PR #13515, opened 2025-10-23
[PWGHF,Trigger] Remove unused ONNX dependency

vkucera:


fgrosa:
Hi @vkucera, did you test this? In the past, not having it in the `CMakeList.cxx` was causing a crash of the filters runtime (that's why it's there)

vkucera:
> > > Hi @vkucera, did you test this? In the past, not having it in the `CMakeList.cxx` was causing a crash of the filters runtime (that's why it's there)
> 
> Sorry for the noise, I now realised that it was before we moved the ML inference in the `trackIndexSkimCreator`, so indeed now it should not be an issue anymore.

Thanks for checking!

PR #13522, opened 2025-10-24
[PWGLF] Refine jet definition for generated particles and code optimization

alcaliva:
- Jet for generated particles is now constructed using final-state charged particles that are physical primary or from HF decays
- Calculation of two perpendicular axes for UE studies is now done in a single function call to reduce CPU usage
- Defined per-event particle containers to reduce memory usage

### DIFF
--- a/PWGLF/Tasks/Strangeness/strangenessInJets.cxx
+++ b/PWGLF/Tasks/Strangeness/strangenessInJets.cxx
@@ -355,6 +355,7 @@ struct StrangenessInJets {
     u.SetXYZ(ux, uy, uz);
     return;
   }
+  */
 
   // Delta phi calculation
   double getDeltaPhi(double a1, double a2)
@@ -371,7 +372,6 @@ struct StrangenessInJets {
 
     return deltaPhi;
   }
-  */
 
   // Check if particle is a physical primary or a decay product of a heavy-flavor hadron
   bool isPhysicalPrimaryOrFromHF(aod::McParticle const& particle, aod::McParticles const& mcParticles)


PWGLF/Tasks/Strangeness/strangenessInJets.cxx
```diff
@@ -370,6 +371,96 @@ struct StrangenessInJets {
...
+  {
+    // Keep only pi, K, p, e, mu
+    int pdg = std::abs(particle.pdgCode());
+    if (!(pdg == PDG_t::kPiPlus || pdg == PDG_t::kKPlus || pdg == PDG_t::kProton || pdg == PDG_t::kElectron || pdg == PDG_t::kMuonMinus))
+      return false;
+
+    // Constants for identifying heavy-flavor (charm and bottom) content from PDG codes
+    static constexpr int kCharmQuark = 4;
+    static constexpr int kBottomQuark = 5;
+    static constexpr int hundreds = 100;
```
romainschotter: Hi @alcaliva ! The PR looks great! Would it be possible to change the name `hundreds` and `thousands` to something more meaningful like `FromMeson` and `FromBaryon`, maybe?

romainschotter (changes requested):

PWGLF/Tasks/Strangeness/strangenessInJets.cxx
romainschotter: Hi @alcaliva ! I know :-D
My point is that `hundreds = 100` looks weird and it would perhaps be better to give a more meaningfull name to `hundreds`.
I am fine either way, I will let you decide :-)
Thank you very much!

PR #13524, opened 2025-10-24
[PWGCF]   Resolved memory consumption issue while testing on hyperloop by modifying histogram binning and added configurables

sweta29singh:
Resolved memory consumption issue while testing on hyperloop


victor-gonzalez:
And, please, set a meaningful message for your PR

PR #13525, opened 2025-10-24
[PWGHF,Tutorial] HfHelper: Make methods static

vkucera:
Makes the instance unnecessary.


alibuild (approved):

PR #13548, opened 2025-10-25
[PWGCF] Add systematic variations on topological selections of V0s in flow analyses

stefanopolitano:
This PR adds the possibility to set loose/tight V0 selections in flow analyses for systematic studies


victor-gonzalez:
To me, it is OK. But all filter two prong wagons would need to  be reconfigured
What about @jgrosseo @jaelpark?

jaelpark:
I think this is ok, as no other things should be affected.

PR #13597, opened 2025-10-30
[PWGHF] taskCharmPolarisation: Enhance the resolution of the ML scores outputs

Mingze129:


Mingze129:
> a question from my side: is it really necessary? `configThnAxisMlBkg` and `configThnAxisMlNonPrompt` are `Configurables`, you should be able to set the values you want at runtime

As we checked this week for PbPb analysis, there are still too much background at low pt range (<5 GeV/c), and almost no signal peak at larger non-promtp cut. So we want a tighter cut on background.
I think it should be better to set `configThnAxisMlBkg` and `configThnAxisMlNonPrompt` as `Configurables` but now they are not. Maybe I can improved this.

Mingze129:
> > a question from my side: is it really necessary? `configThnAxisMlBkg` and `configThnAxisMlNonPrompt` are `Configurables`, you should be able to set the values you want at runtime
> 
> Absolutely agree. Btw, what is the commented out `configThnAxisMlPrompt` doing there?

The `configThnAxisMlPrompt` used to store the prompt scores, we move this now to reduce the output size. 
But base on Xinye's suggestion, Sometimes the analysis works better on prompt cut instead of nonprompt cut. So we want to keep this for possible update in the future.

mfaggin:
> > a question from my side: is it really necessary? `configThnAxisMlBkg` and `configThnAxisMlNonPrompt` are `Configurables`, you should be able to set the values you want at runtime
> 
> As we checked this week for PbPb analysis, there are still too much background at low pt range (<5 GeV/c), and almost no signal peak at larger non-promtp cut. So we want a tighter cut on background. I think it should be better to set `configThnAxisMlBkg` and `configThnAxisMlNonPrompt` as `Configurables` but now they are not. Maybe I can improved this.


"I think it should be better to set configThnAxisMlBkg and configThnAxisMlNonPrompt as Configurables but now they are not. Maybe I can improved this."
What does it mean? These two are `ConfigurableAxes` which can be set as you want at runtime:
Schermata del 2025-10-30 10-55-07

Are you saying that changing the values in Hyperloop does not give you the desired binning in the output?

Mingze129:
> > > a question from my side: is it really necessary? `configThnAxisMlBkg` and `configThnAxisMlNonPrompt` are `Configurables`, you should be able to set the values you want at runtime
> > 
> > 
> > As we checked this week for PbPb analysis, there are still too much background at low pt range (<5 GeV/c), and almost no signal peak at larger non-promtp cut. So we want a tighter cut on background. I think it should be better to set `configThnAxisMlBkg` and `configThnAxisMlNonPrompt` as `Configurables` but now they are not. Maybe I can improved this.
> 
> "I think it should be better to set configThnAxisMlBkg and configThnAxisMlNonPrompt as Configurables but now they are not. Maybe I can improved this." What does it mean? These two are `ConfigurableAxes` which can be set as you want at runtime: Schermata del 2025-10-30 10-55-07
> 
> Are you saying that changing the values in Hyperloop does not give you the desired binning in the output?

Ah, sorry. I didn't notice this. It's my fault.
Then this PR shouldn't be here.  I'm sorry for wasting your time.

PR #13604, opened 2025-10-30
[Common] TPC interface: skipTPCOnly auto mode, fix Run 2 bug

ddobrigk:
This PR: 

- Adds a new mode "-1" to the skipTPCOnly configurable of the TPC PID functionality, such that it will calculate the dEdx for TPC-only tracks only in case specific tasks that require it are detected. The rationale is to allow for a single hyperloop core service wagon to be used while still saving the CPU required to evaluate the Nsigma for TPC-only if not needed. 

- Fixes a Run 2 crash by correctly capturing the attempt to access a CCDB object for Run 2 converted data.


alibuild (dismissed):

PR #13609, opened 2025-10-30
[PWGHF] Bugfixes of correlated background processing in 3-prong decays

lubynets:
1. In evaluation of the `swapping` flag the case of $\Xi^+_c\rightarrow pK^-\pi^+$ decay was missing. This PR fixes it.
2. Removed `Only` from names of `keep` variables for signal, background and correlated background, since it is wrong (`keep`s are connected with logical OR).
3. Reflected $\Lambda^+_c\rightarrow pK^-\pi^+$ added into a definition of correlated background.


lubynets:
Hi @vkucera, I think after adding correlated BG by Mattia it is fully possible situation (from my experience) when one needs to save 2 out of 3 species - e.g. signal for efficiency determination and corrBG for its impact study. When there were only two "keeps", signal and BG, of course it did not make sense to enable both, but now it makes.

PR #13624, opened 2025-10-31
[PWGEM/Dilepton] restructure EMTrack and EMFwdTrack

dsekihat:
PWGEM/Dilepton: restructure EMTrack and EMFwdTrack

### DIFF
--- a/PWGEM/PhotonMeson/Core/Pi0EtaToGammaGamma.h
+++ b/PWGEM/PhotonMeson/Core/Pi0EtaToGammaGamma.h
@@ -77,7 +77,7 @@ struct Pi0EtaToGammaGamma {
   o2::framework::Configurable<std::string> grpmagPath{"grpmagPath", "GLO/Config/GRPMagField", "CCDB path of the GRPMagField object"};
   o2::framework::Configurable<bool> skipGRPOquery{"skipGRPOquery", true, "skip grpo query"};
   o2::framework::Configurable<float> d_bz_input{"d_bz_input", -999, "bz field in kG, -999 is automatic"};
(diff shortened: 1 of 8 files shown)


PWGEM/PhotonMeson/Core/Pi0EtaToGammaGamma.h
```diff
@@ -77,7 +77,7 @@ struct Pi0EtaToGammaGamma {
   o2::framework::Configurable<std::string> grpmagPath{"grpmagPath", "GLO/Config/GRPMagField", "CCDB path of the GRPMagField object"};
   o2::framework::Configurable<bool> skipGRPOquery{"skipGRPOquery", true, "skip grpo query"};
   o2::framework::Configurable<float> d_bz_input{"d_bz_input", -999, "bz field in kG, -999 is automatic"};
-  o2::framework::Configurable<uint64_t> ndiff_bc_mix{"ndiff_bc_mix", 198, "difference in global BC required in mixed events"};
+  o2::framework::Configurable<uint64_t> ndiff_bc_mix{"ndiff_bc_mix", 594, "difference in global BC required in mixed events"};
```
mhemmer-cern: Hey Daiki, is this changed default value good for all systems, or do we need different values for pp, OO and PbPb?
I would assume they might be different due to higher or lower IR, but maybe that's not correct?

dsekihat: Hi @mhemmer-cern , thank you for asking. It comes from ITS ROF length. 594 is for PbPb/OO, and 198 is for pp. It is not from IR. But, of course, 594 is better choice also for pp.

PR #13668, opened 2025-11-04
[PWGHF] Add a mass axis to final Thnsparse Correlation Data

zz951:
In correlatorLcHadrons.cxx, add a histogram to be filled with mass, pt, which will be filled only once in same delta phi bin even.
In taskCorrelationLcHadrons.cxx, add a histogram to be filled in a global region also add anther mass axis.


gluparel:
Hi @zz951,
it is not really clear to me why you want to add the TH3 histogram and in particular the additional ThNSparse in the task + the additional mass axis in the other 3 ThNSparse because it seems to me that you are only duplicating the information. 

In fact, if you add the ThNSparse hCorrel2DVsPtGlobalRegion with the mass axis you can then select on the invariant mass at analysis level and you do not need any more the other 3 ThNsparse (signal, sideband left and sideband right).

Do I understand correctly?

Grazia

apalasciano (dismissed):
Hi Zhen,
Thanks for implementing the suggestion.
I'm approving it

apalasciano (approved):

PR #13693, opened 2025-11-05
[PWGCF] adding sp method

jikim1290:
adding sp method
adding vertex selection


victor-gonzalez:
Please keep it as draft until everything is clear

PR #13687, opened 2025-11-04
[Common] ZDC LI task: subscription fix

ddobrigk:
@coppedis

PR #13698, opened 2025-11-05
[Common] Adding 2 entries for ZDC-light-ion task

coppedis:
ZPC PMC ADC were foreseen in the table but not filled in the task.


ddobrigk (approved):
Looks good, thanks, approving! (note: #13687 fixes the header guard already but isn't merged yet, CI's a little slow these days)

PR #13706, opened 2025-11-05
[PWGCF] adding sp method

jikim1290:
adding sp method


jikim1290:
Hello @victor-gonzalez , I made this with correct path. Please review it and approve/merge it if applicable, thanks!

PR #13750, opened 2025-11-09
[PWGLF] Fix typo

romainschotter:
Fix braket typo (calibration objects for cascades could not be loaded without loading the objects for V0s)


alibuild (approved):

PR #13761, opened 2025-11-10
[Common,PWGCF,PWGDQ,PWGEM,PWGHF,PWGJE,PWGLF,PWGUD,Trigger] Move Zorro into Common/Core

mpuccio:


vkucera:
@mpuccio I disabled the auto-merge because it only works when enabled after all other approvals are given.

mcoquet642 (approved):

zhangbiao-phy (approved):

PR #13775, opened 2025-11-11
[PWGCF] Adding a task for two particle correlation in pp

JosueMtzGar:
I am adding a new task for the calculus of two particle correlation in pp in order to calculate the baseline in a template fit

### DIFF
--- a/PWGCF/TwoParticleCorrelations/Tasks/twoParticleCorrelationPp.cxx
+++ b/PWGCF/TwoParticleCorrelations/Tasks/twoParticleCorrelationPp.cxx
@@ -378,7 +378,7 @@ struct TwoParticleCorrelationPp {
       if (fillCollision(mixed, multiplicity) == false) {
         return;
       }
-      //histos.fill(HIST("eventcount"), bindingOnVtx.getBin({collision1.posZ()}));
+      // histos.fill(HIST("eventcount"), bindingOnVtx.getBin({collision1.posZ()}));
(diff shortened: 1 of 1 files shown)


PWGCF/TwoParticleCorrelations/Tasks/twoParticleCorrelationPp.cxx
```diff
@@ -0,0 +1,574 @@
...
+          }
+          if (range3Min <= multiplicity && multiplicity <= range3Max) {
+            histos.fill(HIST("sameEvent_21_30"), deltaEta, deltaPhi);
+          }
+          if (range4Min <= multiplicity && multiplicity <= range4Max) {
+            histos.fill(HIST("sameEvent_31_40"), deltaEta, deltaPhi);
+          }
+          if (range5Min <= multiplicity && multiplicity <= range5Max) {
+            histos.fill(HIST("sameEvent_41_50"), deltaEta, deltaPhi);
+          }
```
victor-gonzalez: This is performed in a double inner loop of tracks, all the conditionals will be explored although only one, as much, will be `true`
This increase the execution CPU without any purpose
Is there a better way of implement it?

JosueMtzGar: Now, I am using a `else if`  conditional and a final `continue` to reduce the steps in the explore.

PR #13787, opened 2025-11-11
[PWGCF] add two subevent method for pt-pt correlations

Luzhiyongg:


EmilGorm:
Looks fine to me. Can be approved.

PR #13792, opened 2025-11-11
[PWGHF/D2H] Correcting delta mass definition in McRec proccess function

deependra170598:
Hi @fgrosa !
Could you please review the PR?
deltaMass defintion in McRec proccess function is corrected before filling it into container.


fgrosa (commented):
Just a minor comment and then it's good for me, thanks!

PR #13793, opened 2025-11-12
[PWGCF] Fix indexing between track and collision tables + dimension of THn histogram

abmodak:


PWGCF/TwoParticleCorrelations/Tasks/longrangecorrDerived.cxx
```diff
@@ -103,19 +103,33 @@ struct LongrangecorrDerived {
   OutputObj<CorrelationContainer> same{"sameEvent"};
   OutputObj<CorrelationContainer> mixed{"mixedEvent"};
 
+  using CollTable = aod::CollLRTables;
+  using TrkTable = aod::TrkLRTables;
+  using MftTrkTable = aod::MftTrkLRTables;
+  using Ft0aTrkTable = aod::Ft0aLRTables;
+  using Ft0cTrkTable = aod::Ft0cLRTables;
+  using MftbestTrkTable = aod::MftBestTrkLRTables;
+  using V0TrkTable = aod::V0TrkLRTables;
```
victor-gonzalez: I would keep the plurals because they are not iterators
But not in the same way as in `LR`
Is `Colls` what should be plural and not `Tables`
The same for the others

PWGCF/TwoParticleCorrelations/Tasks/longrangecorrDerived.cxx
```diff
@@ -239,92 +252,92 @@ struct LongrangecorrDerived {
...
+  void processMftbestft0aME(CollTable const& col, MftbestTrkTable const& mfts, Ft0aTrkTable const& ft0as)
   {
     processMixed(col, mfts, ft0as);
   }
 
-  void processV0mftbestME(aod::CollLRTables const& col, aod::V0TrkLRTables const& tracks, aod::MftBestTrkLRTables const& mfts)
+  void processV0mftbestME(CollTable const& col, V0TrkTable const& tracks, MftbestTrkTable const& mfts)
   {
     processMixed(col, tracks, mfts);
   }
```
victor-gonzalez: Also in here `col` should be `cols`

PR #13719, opened 2025-11-06
[PWGHF] Add centrality dependence to D0–hadron correlator and correlation task

ayankrmanna:
- Introduced centrality-dependent analysis in correlator and correlation task
 - Relaxed ML selection: changed '||' to '&&' in D0/D0bar BDT condition to improve efficiency

PR #13813, opened 2025-11-13
[PWGHF] Add centrality dependence to D⁰–hadron correlation analysis

ayankrmanna:
Centrality dependence has been added to enable centrality-dependent D0–hadron azimuthal correlation analysis.
Additionally, the ML filtering condition was updated by changing the || to && between the D0 and D0bar selection blocks.
This makes the selection less restrictive: a candidate is now rejected only when both D0 and D0bar fail their prompt/background ML thresholds


vkucera:
@ayankrmanna How is this different from https://github.com/AliceO2Group/O2Physics/pull/13719 ?

PR #13819, opened 2025-11-13
[Common] TrackTuner: add autodetect for OO.

mfaggin:
thanks @phymanshu for the parametrizations


alibuild (approved):

PR #13821, opened 2025-11-13
[PWGHF] Fix ML selection for D0 and D0bar on data and MCRec

samrangy:
Fixed this selection because some the true candidates (non-reflected) were getting rejected due to the default value -1.


apalasciano:
Dear @samrangy,
Please rebase your branch and/or check if the https://github.com/AliceO2Group/O2Physics/pull/13719 already includes all the modifications you are implementing in this PR.
If the other PR has already addressed all the points, please close this.

PR #13835, opened 2025-11-14
[PWGHF] Added PID histograms for V0 daughter and fixed bug for selection of V0 mass window

singhra1994:


singhra1994:
> Hi @singhra1994, sorry to jump in, but I wanted to warn you that using primary particle TOF (the standard TOF) for strange decay daughters may incur in a significant signal loss (up to 30-50% in proton-proton collisions). Further, this signal loss is not well reproduced by MC, which could lead to problems in case MC corrections are needed. The key problem is that primary TOF assumes the track-to-collision association stored in AO2Ds to be correct, while secondary decay daughters are often assigned to a collision that isn't the same as the V0 or Cascade they make up. To avoid this kind of issue, one should either not apply TOF at all or take these effects into account using the strangeness TOF: see [this talk in the strangeness PAG](https://indico.cern.ch/event/1605078/contributions/6769124/attachments/3162839/5620137/DDChinellato-StrangenessTOF-Update2-01.pdf), with the strangeness TOF also being mentioned in the tutorial [here](https://indico.cern.ch/event/1574136/contributions/6755145/attachments/3171040/5637479/O2AT5_PWGLF_Strangeness_Intro.pdf). A central core service wagon that provides strangeness TOF PID is also [provided in hyperloop](https://alimonitor.cern.ch/next-hyperloop/view-wagon/36011/wagon-settings). Hope this helps!



Thanks a lot @ddobrigk for mentioning this, indeed we were having this issue. I wrote this code to calculate data driven efficiency of proton for PID cuts  using V0_lambda. I will certainly look to the material you have provided and will come back to you in case of doubt.

ddobrigk:
> > Hi @singhra1994, sorry to jump in, but I wanted to warn you that using primary particle TOF (the standard TOF) for strange decay daughters may incur in a significant signal loss (up to 30-50% in proton-proton collisions). Further, this signal loss is not well reproduced by MC, which could lead to problems in case MC corrections are needed. The key problem is that primary TOF assumes the track-to-collision association stored in AO2Ds to be correct, while secondary decay daughters are often assigned to a collision that isn't the same as the V0 or Cascade they make up. To avoid this kind of issue, one should either not apply TOF at all or take these effects into account using the strangeness TOF: see [this talk in the strangeness PAG](https://indico.cern.ch/event/1605078/contributions/6769124/attachments/3162839/5620137/DDChinellato-StrangenessTOF-Update2-01.pdf), with the strangeness TOF also being mentioned in the tutorial [here](https://indico.cern.ch/event/1574136/contributions/6755145/attachments/3171040/5637479/O2AT5_PWGLF_Strangeness_Intro.pdf). A central core service wagon that provides strangeness TOF PID is also [provided in hyperloop](https://alimonitor.cern.ch/next-hyperloop/view-wagon/36011/wagon-settings). Hope this helps!
> 
> Thanks a lot @ddobrigk for mentioning this, indeed we were having this issue. I wrote this code to calculate data driven efficiency of proton for PID cuts using V0_lambda. I will certainly look to the material you have provided and will come back to you in case of doubt.

Just in case: From what we've seen in the strangeness TOF PID testing, in primary TOF QA plots, you might see two effects: 

* there will be a very slight offset from N-sigma = 0 (due to deep secondary tracks effectively traveling part of the distance to the TOF as a heavier particle). This should not be enough to explain a large signal loss if you apply a loose Nsigma cut. 
* there will be a significant population that won't be visible in an Nsigma plot, because they're displaced in time by multiples of 25000 picoseconds (BC start time), equivalent to many hundreds of sigma. You will only see that population if you check explicitly the under- and overflow bins of the QA plots.

In summary: the actual histogram of Nsigma might look like a gaussian, but the problem is in wildly large Nsigma values (still well-defined) and those will end up in under/overflow...

singhra1994:
> @zhangbiao-phy I disabled the auto-merge just to leave room for potential modifications resulting from the discussion. If the points have been resolved, feel free to proceed with the merging.

@vkucera @zhangbiao-phy, current version of strangeness TOF pid is not useful for our purpose. There is no harm to merge this PR, later if need we can add strangeness PID

PR #13837, opened 2025-11-14
[PWGJE] Change to loop over mccollisions for response filling

dajones2:


dajones2:
Hi Nima, unfortunately due to a long string of technical issues (that I no longer have the time or energy to solve!), I have not had a reliable and/or functional local installation of O2 for months. So I'm afraid this is as good as it gets for now!

nzardosh (approved):

PR #13845, opened 2025-11-15
[ALICE3] Implement task and tree creator for 3-prong candidates

Marcellocosti:
In this PR, the selector, task, and tree creator for simulation performance of HF 3-prong candidates is implemented. Moreover, an header to apply bdt selection is also added. The current code covers only the Lc baryon, but the aim is to cover all 3-prong species with proper templetization.

### DIFF
--- a/ALICE3/TableProducer/CMakeLists.txt
+++ b/ALICE3/TableProducer/CMakeLists.txt
@@ -53,8 +53,10 @@ o2physics_add_dpl_workflow(alice3-hf-selector-3prong
 
 o2physics_add_dpl_workflow(alice3-hf-tree-creator-3prong
                     SOURCES alice3HfTreeCreator3Prong.cxx
+                    PUBLIC_LINK_LIBRARIES O2Physics::AnalysisCore O2::DCAFitter
+                    COMPONENT_NAME Analysis)
 
 o2physics_add_dpl_workflow(alice3-tracking-translator
                     SOURCES alice3TrackingTranslator.cxx
-                    PUBLIC_LINK_LIBRARIES O2Physics::AnalysisCore O2::DCAFitter
+                    PUBLIC_LINK_LIBRARIES O2Physics::AnalysisCore
                     COMPONENT_NAME Analysis)


Marcellocosti:
> I don't like the mixing of Run 3 and Run 5 features. They should be as decoupled as possible. That's why there is the PWGHF/ALICE3 directory that contains (now obsolete) code used for the LoI. Now you are adding HF workflows in the main ALICE3 directory without any reference to HF in the names and you add the required features in the main PWGHF headers which adds more fragmentation.

I see your points. I was also thinking of what is the best solution to add my implementation. Should I use the obsolete `PWGHF/ALICE3` folder? Or it is better to move everything inside the main `ALICE3` folder? Regarding the naming conventions, I saw that HF code and ALICE3 code have quite different naming styles, so I tried to find a way in between. Do you have any suggestion?

vkucera:
> Regarding the naming conventions, I saw that HF code and ALICE3 code have quite different naming styles, so I tried to find a way in between. Do you have any suggestion?

ALICE3 code violates the O2 naming conventions, HF code doesn't. Stick to the official conventions, enforced by the O2 linter.

PWGHF/Core/HfHelper.h
```diff
@@ -1336,6 +1336,65 @@ class HfHelper
...
+  {
+    switch (CharmHad) {
+      case o2::analysis::CharmHadAlice3::Lc:
+        return cand.isSwapped() ? invMassLcToPiKP(cand)
+                                : invMassLcToPKPi(cand);
+      default:
+        LOG(fatal) << "Unsupported charm hadron type";
+        return -1.;
+    }
+  }
```
vkucera: These cannot be here. The best place would be to make dynamic columns `invMassLcTo...` in the ALICE3 HF data model or to make a separate header for ALICE3 HF utils and call `HfHelper::invMassLcTo...` from there.

vkucera:
> Hi all, I moved the implementation under the ALICE3 folder, please let me know if I overlooked anything!

Hi @Marcellocosti , thanks for the update, but now it seems that you are duplicating a lot of Run3 HF stuff. I thought you would just put in ALICE3 the stuff that is different from the Run3 stuff. Why do you need a new full candidate table a new selector array and a new ML response?

PR #13848, opened 2025-11-15

ariedel-cern:
This PR contains the following changes:
- Underlying logic/objects for computation of the bitmasks for the different particles has been improved
- the selction object now generate histograms, couting how many particles survive specific cuts and stores the configuration about each cut in the bin lable
- a cutculator macro has been added which can compute bitmask with user input from histograms stored by the selection objects
- enforce existence of TOF information above a PID specific threshold more strictly to not store uneccesary particles


alibuild (approved):

PR #14145, opened 2025-12-09
[Infrastructure] Use -gsplit-dwarf to improve linking

ktf:


ktf:
Indeed... It's only supported by Clang... :-/ We can have the version without the option, however it will mean that we have to do some extra juggling to get the debug symbols working.

PR #14355, opened 2025-12-27
[PWGHF] Include D0p channel in charm resonance task

apalasciano:
D0p channel is included in the charm resonance task.
New process functions are introduced only for data since no MC is available for these channels at the moment.


alibuild (approved):

PR #14531, opened 2026-01-17
[PWGCF] Update flowEventPlane.cxx

yashpatley:
1. Added track table to fix previous error of table size mismatch
2. Added v0 flow

### DIFF
--- a/PWGCF/Flow/Tasks/flowEventPlane.cxx
+++ b/PWGCF/Flow/Tasks/flowEventPlane.cxx
@@ -739,7 +739,7 @@ struct IdHadronFlow {
   bool identifyTrack(T const& track)
   {
     // Electron rejection
-    if (track.tpcNSigmaEl() > cTpcElRejCutMin && track.tpcNSigmaEl() < cTpcElRejCutMax) {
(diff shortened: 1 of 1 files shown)


victor-gonzalez (commented):
I guess there are no plans of using stored derived data
Why then recreate and store information that is already available in the `Tracks` tabel?

victor-gonzalez:
Well, the memory usage will increase quite a lot if the `Tracks` table is replicated in memory which is basically what this PR is doing. The right way should be a table with just the additional information needed but `joinable` to the `Tracks` table. For doing that all the `tracks` should have proper information in this additional table. By proper information I mean consistent information for the considered analysis, i.e., if the `track` should not be included in the analysis, its entry in the additional table should have a value which identifies such situation (perhaps `false` for pion, proton, and kaon flags?)

yashpatley:
Dear Victor, we need derived dataset in future for local runs and that's why I am adding tables also. I discussed this with Debojit and we need this derived data to get the results within approval timeline. Please consider this PR.

victor-gonzalez (approved):
Have a look at may comment in case you want to consider it (it will decrease the CPU consumption)

PR #14655, opened 2026-01-27
[PWGJE] Add PCH for JET tasks

singiamtel:
Heavily speeds up compilation

CC @ktf

### DIFF
--- a/PWGJE/Tasks/CMakeLists.txt
+++ b/PWGJE/Tasks/CMakeLists.txt
@@ -148,6 +148,15 @@ if(FastJet_FOUND)
                         SOURCES jetSubstructureDielectronOutput.cxx
                         PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::PWGJECore O2Physics::AnalysisCore
                         COMPONENT_NAME Analysis)
+
+    # PCH for jet substructure HF tasks (makes for much faster compilation)
+    target_precompile_headers(O2Physicsexe-analysis-je-jet-substructure-d0
+        PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}/jetSubstructureHFPCH.h")
+    foreach(T IN ITEMS dplus ds dstar lc b0 bplus xictoxipipi dielectron)
+        target_precompile_headers(O2Physicsexe-analysis-je-jet-substructure-${T}
+            REUSE_FROM O2Physicsexe-analysis-je-jet-substructure-d0)
+    endforeach()
+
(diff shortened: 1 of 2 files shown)


ktf:
@vkucera on a separate note, could we please make sure the O2 linter does not tell people to change QA to Qa ?

ktf:
I am not sure I understand. Can't you simply add the exceptions? There is hopefully a finite number of them. If the tools are wrong, we should fix the tools, or at least remove the check.

ktf:
> Because it's common sense to name files after their content. The struct-device-workflow-file name correspondence makes it trivial to deduce one name from another. 

So does removing `-` and doing a case insensitive search.


> E.g. if I want to know where a device is implemented, I get the file name by converting it to camelCase. Multiple tasks are not a problem. 

I do git grep or fuzzy finding in my editor, even without needing the name of the file. Please do not call "common sense" the limitations of your tools or their configuration. You can get to it in many ways. I also question the actual usefulness of the correspondence, given we do support custom TaskNames and multiple tasks per file.


> This was discussed one year ago and approved as error severity.

Yes, and in the same meeting I said that even one "false positive" to me it's enough to consider the check wrong and dangerous. Someone could have fixed the linter report by renaming the task to -q-a (which is less of a burden compared to changing the filename) and Hyperloop would have broken as well.


> > That said, my statement still holds. Tools should not enforce something which is against common sense because something else is broken. We should either fix the broken part or leave the common sense alone and disable the check.

> The statement holds only if you ignore the role of capitalisation as a word divider in camelCase.

Yes, and as this PR shows, I am not the only one. Moreover, I question the word divider idea in the first place, because that would mean that `hello-world` and `helloworld` would have sit in two files differing only by capitalisation (which is a mess, and not only for macOS).


> > In that regard, a better behaviour for `type_to_task_name` is to be found in [AliceO2Group/AliceO2#15006](https://github.com/AliceO2Group/AliceO2/pull/15006).

> As I said, this introduces ambiguity and gives green light to unreadable names.

There is no ambiguity if you do the substitutions in reverse size order of the exceptions, as per comment in the code. Concerning unreadable names, `TPCPIDQA.cxx` is as unreadable as `TpcPidQa.cxx`.


> > Can we have the linter stop complaining about QA and QC, please?

> Of course, if you change the device name generation algorithm, I will have to adjust O2 linter accordingly. But this should have been discussed in a meeting with the analysis coordinator.

Now that I know the implications, it will be in the meeting tomorrow, of course.

PR #14754, opened 2026-02-02
[Common] Unnecessary multiplication & division removed

MaximVirta:
Unnecessary complexity removed. The variable "nmode" was multiplied and divided within the same function call.


vkucera:
@MaximVirta @dsekihat Why is this pending?

Marcellocosti:
Hi all, probably this is implemented in #16244, so maybe the PR is not necessary anymore

PR #14949, opened 2026-02-13
[PWGDQ] added alignment corrections for MFT and MCH in global alignment task

aferrero2707:
MFT: only global alignment corrections are implemented, in the form of configurable rotations and translations of the top and bottom halves

MCH: the re-alignment and re-fitting of the MCH clusters and tracks is fully implemented, inspired from the muon QA task

Both can be enabled/disabled individually via configration options, and are disabled by default.


dsekihat:
Dear @aferrero2707 , Thank you very much for the PR. Is there any way to use this correction in common place so that everyone can apply?

mcoquet642 (approved):

PR #15205, opened 2026-02-28
[PWGHF] Add D0+gamma in charm reso data creator

fgrosa:


alibuild (approved):

PWGHF/D2H/Core/DataCreationCharmReso.h
```diff
@@ -438,6 +500,195 @@ bool buildAndSelectV0(const Coll& collision, const std::array<int, 3>& dDaughter
...
+  }
+  if (v0.radius < std::fabs(gammaDecayVtx.GetZ()) * std::tan(2 * std::atan(std::exp(-cfgGammaCuts.etaMax.value))) - marginTpc) {
+    return false; // RZ line cut
+  }
+  if (v0.radius < cfgGammaCuts.radiusMin.value || cfgGammaCuts.radiusMax.value < v0.radius) {
+    return false;
+  }
+
+  const float minRadTpcOnly{16.f};
+  if ((!trackNeg.hasITS() && !trackNeg.hasITS()) && v0.radius < minRadTpcOnly) { // TPConly tracks can detect conversion points larger than minRadTpcOnly.
```
vkucera, line 612: Looks like a bug.

fgrosa: It seems correct to me: if both tracks don't have ITS (-> TPC only), and the radius is smaller than the minimum for TPC only, it returns `false`. Am I overlooking something?

vkucera: You only check `trackNeg` twice.

PR #15377, opened 2026-03-13
[PWGCF] Merge multiparticle-correlations-ab files

vkucera:
Merge pieces of the `multiparticle-correlations-ab` task into a single task file.
FYI @abilandz

The task was for some reason scattered across six files by ripping task sections out into files which pretend to be header files but are not because:

- They neither implement reusable features nor declare a class.
- They are not self-contained.
- They are included inside the body of a struct definition.
- They are placed in the `Core` directory without actually providing any core functionality.

As a result of this fragmentation, the code cannot be analysed by compiler-based analysis tools, like Clang-Tidy.

For these reasons, I suggest to merge the code into a single task file, as done for all other analysis tasks in the repository.


abilandz:
Dear @vkucera, please can you drop your commit? 

I am not ready to move this way yet. I agreed with @victor-gonzalez some time ago to keep the code structured this way, as it is much easier for me to develop and maintain the code. Once I am done with the major development, I can consider to move the way you suggested. Honestly, I am not really keen to work on a monolithic file with 20K+ lines as a result of your commit, where everything is piled up on top of each other.

Instead, perhaps you can suggest an alternative design which will enable me to keep enum definitions in a separate file, global constants in another file, etc,, and which will be compatible with Clang-Tidy ?

On a side note, I would appreciate it if you first contact me for a discussion before making any direct intervention in my code out of the blue in the central repository. Thanks for your understanding!

abilandz:
> Dear @abilandz , thanks for your feedback.
> 
> > I am not really keen to work on a monolithic file with 20K+ lines as a result of your commit, where everything is piled up on top of each other.
> 
> Please consider the following facts about your current code:
> 
> * The `MuPa-MemberFunctions.h` file itself is "a monolithic file with 20K+ lines", so you are already in that situation regardless of my commit.

Dear @vkucera,

That's not a realistic size for that file, as several functions are marked as obsolete because I eventually came up with a better design or a more efficient implementation. Once I validate the new implementation until the end, the obsolete code will be purged, and this file will shrink. 

> * The `MuPa-MemberFunctions.h` file contains over 92 % of the total lines of the code related to the task, so merging the code into a single file makes it more piled up by less than 8 %.

After I remove the obsolete code mentioned in the previous reply, that percentage will shoot up.  

> * 19 % of the lines are comments. If you reduce those 4000 lines of comments, you can easily get below 20k lines in total.

I do not comment on the weather here. These comments are an integral part of my analysis task, describing, for instance, in detail the algorithms used in the code implementations, etc. 

They also mark the weak points in the code that still need improvement or that need to be implemented from scratch.

But, I have avoided at all costs leaving the commented source code in my task. 

> * You can still keep the enumerators and global constants as separate headers if you prefer.

Yes, thanks! 

> * You can also promote your existing struct members inheriting from `ConfigurableGroup` to
>   named structs with other data members and methods. See e.g. `HfEventSelection`. https://github.com/AliceO2Group/O2Physics/blob/d8bc5a4a6b00267731b491db5ba60e96ca0dc4d4/PWGHF/Utils/utilsEvSelHf.h#L154

Thanks! i will have a look! 

> * All the data members defined with named structs in `MuPa-DataMembers.h` can be declared as instances of those structs inside the task, while keeping the definitions in the header.

Thanks!

> 
> This can all be done while guaranteeing that each file can be compiled on its own.
> 
> Btw, I tagged you in the PR description precisely for the purpose of notifying you of this proposal.

Thanks for your feedback! 

Ante

PR #15497, opened 2026-03-23
[PWGJE] skipping collisions without mc collision

nzardosh:


alibuild (approved):

PR #15517, opened 2026-03-25
[PWGDQ] Add QA matching derived tables AO2D outputs

nivram-phy:
This PR extends [qaMatching.cxx](app://-/index.html#) to produce QA-focused AO2D tables and improves true-match indexing logic to use track-type–based muon selection. Add AOD schema and producers for QAMEVT, QAMCHTRK, and QAMCAND.

QAMEVT
Event-level summary: reduced event ID, MFT multiplicity, and collision vertex position.
QAMCHTRK
Per-MCH track info: track ID/type, kinematics, vertex-propagated params, and MFT matching attempts.
QAMCAND
Per-candidate info: matching label, parent track ID/kinematics, match type, score, and rank.


aferrero2707:
@nivram-phy I am testing the code locally, will give some feedback here after the weekend.

nivram-phy:
@mcoquet642 Two separate issues showed up: PWGDQ/Tasks/qaMatching.cxx is now being fully linted and fails on many pre-existing whole-file linter violations, OS violations from outside of this code
Should I do something?

@aferrero2707 Latest code with my modifications has been pushed.

PWGDQ/Tasks/qaMatching.cxx
vkucera, line 1: Don't revert the include fixes.

vkucera (changes requested):

PWGDQ/Tasks/qaMatching.cxx
nivram-phy: I wont and I have removed the iostream include and cout also

vkucera:
```suggestion
      double pt = std::hypot(px, py);
```

PWGDQ/Tasks/qaMatching.cxx
```diff
@@ -79,12 +78,81 @@
 #include <utility>
 #include <vector>
 
-#include <math.h>
```
vkucera, line 81: Why are you removing this?

nivram-phy: resolved. I think I copied block of includes prior to this.

vkucera:
You can just run the linter and formatting locally and fix all issues at once before you push an update.

nivram-phy:
@vkucera I couldn't do it locally or understand where it was failing, so I am doing it this way. I am on very unfamiliar territory here.

vkucera:
> @vkucera I couldn't do it locally or understand where it was failing, so I am doing it this way. I am on very unfamiliar territory here.

- https://aliceo2group.github.io/analysis-framework/docs/tools/o2linter.html#how-to-run-o2-linter
- https://aliceo2group.github.io/analysis-framework/docs/tools/#pre-commit-hooks

vkucera (dismissed):

vkucera:
You didn't test your changes before pushing.

nivram-phy:
@vkucera I noticed that macOS-arm and o2 check seems to happen only if the PR is not in draft, is that correct or maybe I haev misunderstood it?

vkucera:
> @vkucera I noticed that macOS-arm and o2 check seems to happen only if the PR is not in draft, is that correct or maybe I haev misunderstood it?

It's correct. We don't compile draft PRs.

nivram-phy:
@aferrero2707 Back to a stable point, needed to fix a lot of things as some baseline linter requirements have changed.

vkucera:
> @aferrero2707 Back to a stable point, needed to fix a lot of things as some baseline linter requirements have changed.

@nivram-phy What do you mean? No requirements have changed since you opened the PR.

nivram-phy:
> > @aferrero2707 Back to a stable point, needed to fix a lot of things as some baseline linter requirements have changed.
> 
> @nivram-phy What do you mean? No requirements have changed since you opened the PR.

@vkucera We were working with a different version of the code from a month ago locally for building a derived table table. From that point, the include files were modified in the master that I had not noticed. This is what I mentioned. 
Maybe my wording could have been better

PR #15569, opened 2026-03-30
[ALICE3] Strangeness finding with otf decayer output

jesgum:
@lhusova fyi: with these changes we are able to run with the decayer in otf simulations. I've also tried to make sure not to break the ACTS workflow and that both workflows can be enabled and run separately. I also moved the dca fitter settings into a configurable group which might update some parameters on HY in case the default one is not used


njacazio (approved):

vkucera:
@jesgum When are you going to fix the bugs that I reported three months ago? https://github.com/AliceO2Group/O2Physics/pull/14441#issuecomment-3767219673

PR #15653, opened 2026-04-05
[PWGHF] Added safety checks before using ccdb efficiency objects

deependra170598:
Hi @singhra1994, @apalasciano and @gluparel,

Could you please review this PR?

In this PR, I tried to add safety checks before using ccdb efficiency object so that NAN dose not propagate to the THnSparse.

Let me know you any suggestion or improvement needed, Thanks.

### DIFF
--- a/PWGHF/HFC/Tasks/taskCorrelationDstarHadrons.cxx
+++ b/PWGHF/HFC/Tasks/taskCorrelationDstarHadrons.cxx
@@ -118,8 +118,8 @@ struct HfTaskCorrelationDstarHadrons {
   HistogramRegistry registry{"registry", {}, OutputObjHandlingPolicy::AnalysisObject, true, true};
 
   o2::ccdb::CcdbApi ccdbApi;
-  std::vector<TH1F*> vecHistEfficiencyDstar;
-  std::vector<TH1D*> vecHistEfficiencyTracks;
+  std::vector<TH1*> vecHistEfficiencyDstar;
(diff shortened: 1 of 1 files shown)


apalasciano (approved):

PWGHF/HFC/Tasks/taskCorrelationDstarHadrons.cxx
```diff
@@ -212,19 +212,31 @@ struct HfTaskCorrelationDstarHadrons {
...
+          if (!efficiencyWeightDstar) {
+            LOGF(fatal, "Dstar efficiency weight can't be zero.");
+          }
+        }
+        float const ptEffLowEdgeTrack = vecHistEfficiencyTracks[0]->GetBinLowEdge(1);
+        if (ptTrack <= ptEffLowEdgeTrack) { // pT of current track is lower than the lower edge of the pT axis
+          efficiencyWeightTracks = vecHistEfficiencyTracks[0]->GetBinContent(1);
+        } else {
+          efficiencyWeightTracks = vecHistEfficiencyTracks[0]->GetBinContent(vecHistEfficiencyTracks[0]->GetXaxis()->FindBin(ptTrack));
+          if (!efficiencyWeightTracks) {
```
vkucera, line 236: This is wrong. You convert `float` to `bool` and compare it to `false`. For that to happen, the `float` would have to be exactly `0` which almost never happens if the value is a result of a calculation. Furthermore, the syntax signals the intention of using a `bool` value which is not the case.

PR #15664, opened 2026-04-06
[Common] Fetch FIT ccdb also if only BC cent requested

ddobrigk:
@abmodak, thanks for spotting!


alibuild (approved):

PR #15681, opened 2026-04-08
[PWGDQ] Added a process function in dqFlow to compute FT0C cumulant for flow analysis

Ankyyadav:
For flow analysis using 2 particle correlations at midrapidity, one needs the RefFlow table filled with sum of squares of amplitude from FT0C


Ankyyadav:
tableMaker_withAssoc changed by Luca to include Qvector tables there. These changes in dqFlow would be obsolete with those changes.

PR #15711, opened 2026-04-10
[ALICE3] Add improved LUT format

aalkin:
Hello @njacazio,

This is the replacement for TrackSmearer and LUTWriter, using a flattened data structure FlatLutData. It should be functionally identical (needs validation), but in addition it allows to construct the TrackSmearer as a view over shared memory instead of copying memory or loading the LUT from file. From the API standpoint, you should be able to use the new TrackSmearer in exactly the same way as the old one, and after some additional developments in O2 you can switch to the view approach.

### DIFF
--- a/ALICE3/Core/FlatTrackSmearer.cxx
+++ b/ALICE3/Core/FlatTrackSmearer.cxx
@@ -11,7 +11,8 @@
 
 #include "FlatTrackSmearer.h"
 
-namespace o2::delphes {
+namespace o2::delphes
+{
 int TrackSmearer::getIndexPDG(int pdg) const
 {
   switch (std::abs(pdg)) {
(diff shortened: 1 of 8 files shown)


ALICE3/Core/FlatTrackSmearer.cxx
```diff
@@ -0,0 +1,414 @@
...
+    case 1000010020:
+      return 5; // Deuteron
+    case 1000010030:
+      return 6; // Triton
+    case 1000020030:
+      return 7; // Helium3
+    case 1000020040:
+      return 8; // Alphas
+    default:
+      return 2; // Default: pion
```
vkucera, line 44: Use named PDG codes.

aalkin:
Superseded by #15826

PR #15714, opened 2026-04-10
[PWGEM] update of emcalPi0Qc task to allow running over (weighted) MC productions

fjonasALICE:


fjonasALICE:
i screwed up and pushed other commits into this PR. Will fix

PR #15727, opened 2026-04-10
[PWGJE] Rename task to jetDsSpecSubs and add angularity

monamelop:
- Rename task from jetDsSpectrumAndSubstructure to jetDsSpecSubs
- Introduce new THnSparse histograms


vkucera:
@monamelop Why is this opened?

PR #15741, opened 2026-04-13
[PWGDQ] Fill q-vector tables for all PbPb process functions in tableMaker_withAssoc

Ankyyadav:
Q-vector tables are to be filled in tableMaker_withAssoc now rather than using dqFlow to save memory. For all PbPb barrel processes, q-vector tables are filled now except for processPbPbWithFilterBarrelOnly. Added a cofigurable to check if q-vector calibrations exist.
For flow analysis using 2 particle correlations at midrapidity, one needs the RefFlow table filled with sum of squares of amplitude from FT0C. These are now filled in dqFlow. Switched to using a Produces group for all q-vector related tables

### DIFF
--- a/PWGDQ/TableProducer/tableMaker_withAssoc.cxx
+++ b/PWGDQ/TableProducer/tableMaker_withAssoc.cxx
@@ -295,10 +295,10 @@ struct TableMaker {
     Configurable<bool> fConfigSaveElectronSample{"cfgSaveElectronSample", false, "If true, only save electron sample"};
   } fConfigPostCalibTPC;
 
-  // 
+  //
   struct : ConfigurableGroup {
     Configurable<bool> fConfigFT0CCumulant{"cfgFT0CCumulant", false,
-    "If true, compute RefFlow cumulants from FT0C amplitudes (requires FT0s subscription)"};
+                                           "If true, compute RefFlow cumulants from FT0C amplitudes (requires FT0s subscription)"};
   } fConfigQvector;
 
   struct : ConfigurableGroup {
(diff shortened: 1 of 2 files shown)


PWGDQ/TableProducer/tableMaker_withAssoc.cxx
```diff
@@ -208,6 +206,14 @@ struct TableMaker {
   Produces<ReducedMFTsExtra> mftTrackExtra;
   Produces<ReducedMFTAssoc> mftAssoc;
 
+  // Q-vector related tables, to be filled only if the user selects the corresponding option; since they are not needed for the skimming, we keep them in a separate group to avoid filling them when not needed
```
lucamicheletti93, line 210: I would remove for the momento the q-vecotrs calculated by the CF framework. We need more tests to validate it properly

lucamicheletti93:
Dear @Ankyyadav , thanks for the PR! I think that for the moment you can remove the part for the cumulants. We need more test on the memory and on the consistency of the results. My suggestion is to just add the process function to include the filling of the q-vecotor tables when you use central barre j/psi.

PWGDQ/TableProducer/tableMaker_withAssoc.cxx
Ankyyadav: Okay, I would remove the cumulant related tables and the part where I compute it. Though I do not use the GFW to fill these tables. I just need the FT0C amplitudes that I get from collisions tables.
Would it work if I add the FT0C cumulant part in dqFlow and a new process along with the existing processAllQvector and processQvecterFromCenter?

lucamicheletti93: @Ankyyadav maybe it is better to keep all the cumulant part in the dqFlow, and fully integrate it in the table-maker in a second moment

Ankyyadav: Okay. I have made the changes and put the cumulant part in dqFlow

lucamicheletti93 (approved):

PR #15746, opened 2026-04-13
[PWGDQ] Integration of dqFlow workflow into the tableMaker

lucamicheletti93:
- removal for the dq-flow of the q-vector tables produced in the table-maker
- some cleaning of the histograms for flow
- new minimal muon selection with 10sigma pDCA
- additional process function to do the mixing with muons with q-vectors in the table-reader-with-assoc


alibuild (approved):

PR #15755, opened 2026-04-14
[PWGHF] Species-dependent minimum prong pt cut in skimming

Marcellocosti:
This PR implements the possibility to have different minimum prong pt selections for the HF 3 prong candidates in the `trackIndexSkimCreator` workflow.

### DIFF
--- a/PWGHF/Core/SelectorCuts.h
+++ b/PWGHF/Core/SelectorCuts.h
@@ -251,6 +251,17 @@ static const std::vector<std::string> labelsPt{};
 
 // column labels
 static const std::vector<std::string> labelsCutVar = {"massMin", "massMax", "cosp", "decL"};
+
+// Single prong min pt selection
+static constexpr int NSpecies3Prong = 8; // number of species for 3-prong candidates
+static const float cutsProngMinPt3Prong[NSpecies3Prong] = {0.4, 0.4, 0.4, 0.4, 0.4, 0.4, 0.4, 0.4};
+
+// 3-prong species labels
+static const std::vector<std::string> labelsSpecies3Prong = {"Dplus", "Lc", "Ds", "Xic", "Cd", "Ct", "Ch", "Ca"};
+
+// column label
+static const std::vector<std::string> labelsMinPt = {"Minimum track transverse momentum (GeV/c)"};
+
 } // namespace hf_cuts_presel_3prong
 
 namespace hf_cuts_presel_ds
--- a/PWGHF/Core/SelectorCuts.h
+++ b/PWGHF/Core/SelectorCuts.h
@@ -254,10 +254,10 @@ static const std::vector<std::string> labelsCutVar = {"massMin", "massMax", "cos
 
 // Single prong min pt selection
 static constexpr int NSpecies3Prong = 8; // number of species for 3-prong candidates
-static const float cutsProngMinPt3Prong[NSpecies3Prong] = {0.4, 0.4, 0.4, 0.4, 0.4, 0.4, 0.4, 0.4};
+static const float ptProngMin[NSpecies3Prong] = {0.4, 0.4, 0.4, 0.4, 0.4, 0.4, 0.4, 0.4};
 
 // 3-prong species labels
-static const std::vector<std::string> labelsSpecies3Prong = {"Dplus", "Lc", "Ds", "Xic", "Cd", "Ct", "Ch", "Ca"};
+static const std::vector<std::string> labelsPtProngMin = {"Dplus", "Lc", "Ds", "Xic", "Cd", "Ct", "Ch", "Ca"};
 
 // column label
 static const std::vector<std::string> labelsMinPt = {"Minimum track transverse momentum (GeV/c)"};
--- a/PWGHF/Core/SelectorCuts.h
+++ b/PWGHF/Core/SelectorCuts.h
@@ -232,7 +232,7 @@ static const std::vector<std::string> labelsCutVar = {"massMin", "massMax", "cos
 namespace hf_cuts_presel_3prong
 {
 static constexpr int NBinsPt = 2;
-static constexpr int NCutVars = 4;
+static constexpr int NCutVars = 5;
 // default values for the pT bin edges (can be used to configure histogram axis)
 // common for any 3-prong candidate
 // offset by 1 from the bin numbers in cuts array
@@ -243,31 +243,21 @@ constexpr double BinsPt[NBinsPt + 1] = {
 const auto vecBinsPt = std::vector<double>{BinsPt, BinsPt + NBinsPt + 1};
 
 // default values for the cuts
-constexpr double Cuts[NBinsPt][NCutVars] = {{1.75, 2.05, 0.7, 0.02},  /* 1 < pt < 5 */
-                                            {1.75, 2.05, 0.5, 0.02}}; /* 5 < pt < 1000 */
+constexpr double Cuts[NBinsPt][NCutVars] = {{1.75, 2.05, 0.7, 0.02, 0.3},  /* 1 < pt < 5 */
+                                            {1.75, 2.05, 0.5, 0.02, 0.3}}; /* 5 < pt < 1000 */
 
 // row labels
 static const std::vector<std::string> labelsPt{};
 
 // column labels
-static const std::vector<std::string> labelsCutVar = {"massMin", "massMax", "cosp", "decL"};
-
-// Single prong min pt selection
-static constexpr int NSpecies = 8; // number of species for 3-prong candidates
-constexpr float PtProngMin[NSpecies][1] = {{0.4f}, {0.4f}, {0.4f}, {0.4f}, {0.4f}, {0.4f}, {0.4f}, {0.4f}};
-
-// 3-prong species labels
-static const std::vector<std::string> labelsSpecies = {"Dplus", "Lc", "Ds", "Xic", "Cd", "Ct", "Ch", "Ca"};
-
-// column label
-static const std::vector<std::string> labelsPtProngMin = {"Minimum pT (GeV/c)"};
+static const std::vector<std::string> labelsCutVar = {"massMin", "massMax", "cosp", "decL", "ptProngMin"};
 
 } // namespace hf_cuts_presel_3prong
 
 namespace hf_cuts_presel_ds
 {
 static constexpr int NBinsPt = 2;
-static constexpr int NCutVars = 5;
+static constexpr int NCutVars = 6;
 // default values for the pT bin edges (can be used to configure histogram axis)
 // offset by 1 from the bin numbers in cuts array
 constexpr double BinsPt[NBinsPt + 1] = {
@@ -277,14 +267,14 @@ constexpr double BinsPt[NBinsPt + 1] = {
 const auto vecBinsPt = std::vector<double>{BinsPt, BinsPt + NBinsPt + 1};
 
 // default values for the cuts
-constexpr double Cuts[NBinsPt][NCutVars] = {{1.70, 2.15, 0.7, 0.02, 0.02},  /* 1 < pt < 5 */
-                                            {1.70, 2.15, 0.5, 0.02, 0.02}}; /* 5 < pt < 1000 */
+constexpr double Cuts[NBinsPt][NCutVars] = {{1.70, 2.15, 0.7, 0.02, 0.3, 0.02},  /* 1 < pt < 5 */
+                                            {1.70, 2.15, 0.5, 0.02, 0.3, 0.02}}; /* 5 < pt < 1000 */
 
 // row labels
 static const std::vector<std::string> labelsPt{};
 
 // column labels
-static const std::vector<std::string> labelsCutVar = {"massMin", "massMax", "cosp", "decL", "deltaMassKK"};
+static const std::vector<std::string> labelsCutVar = {"massMin", "massMax", "cosp", "decL", "ptProngMin", "deltaMassKK"};
 } // namespace hf_cuts_presel_ds
 
 namespace hf_cuts_presel_dstar
--- a/PWGHF/Core/SelectorCuts.h
+++ b/PWGHF/Core/SelectorCuts.h
@@ -254,7 +254,7 @@ static const std::vector<std::string> labelsCutVar = {"massMin", "massMax", "cos
 
 // Single prong min pt selection
 static constexpr int NSpecies = 8; // number of species for 3-prong candidates
-static const float ptProngMin[NSpecies] = {0.4, 0.4, 0.4, 0.4, 0.4, 0.4, 0.4, 0.4};
+static const float ptProngMin[NSpecies][1] = {{0.4}, {0.4}, {0.4}, {0.4}, {0.4}, {0.4}, {0.4}, {0.4}};
 
 // 3-prong species labels
 static const std::vector<std::string> labelsPtProngMin = {"Dplus", "Lc", "Ds", "Xic", "Cd", "Ct", "Ch", "Ca"};
--- a/PWGHF/Core/SelectorCuts.h
+++ b/PWGHF/Core/SelectorCuts.h
@@ -253,8 +253,8 @@ static const std::vector<std::string> labelsPt{};
 static const std::vector<std::string> labelsCutVar = {"massMin", "massMax", "cosp", "decL"};
 
 // Single prong min pt selection
-static constexpr int NSpecies3Prong = 8; // number of species for 3-prong candidates
-static const float ptProngMin[NSpecies3Prong] = {0.4, 0.4, 0.4, 0.4, 0.4, 0.4, 0.4, 0.4};
+static constexpr int NSpecies = 8; // number of species for 3-prong candidates
+static const float ptProngMin[NSpecies] = {0.4, 0.4, 0.4, 0.4, 0.4, 0.4, 0.4, 0.4};
 
 // 3-prong species labels
 static const std::vector<std::string> labelsPtProngMin = {"Dplus", "Lc", "Ds", "Xic", "Cd", "Ct", "Ch", "Ca"};
--- a/PWGHF/Core/SelectorCuts.h
+++ b/PWGHF/Core/SelectorCuts.h
@@ -254,13 +254,13 @@ static const std::vector<std::string> labelsCutVar = {"massMin", "massMax", "cos
 
 // Single prong min pt selection
 static constexpr int NSpecies = 8; // number of species for 3-prong candidates
-static const float ptProngMin[NSpecies][1] = {{0.4}, {0.4}, {0.4}, {0.4}, {0.4}, {0.4}, {0.4}, {0.4}};
+constexpr float ptProngMin[NSpecies][1] = {{0.4f}, {0.4f}, {0.4f}, {0.4f}, {0.4f}, {0.4f}, {0.4f}, {0.4f}};
 
 // 3-prong species labels
-static const std::vector<std::string> labelsPtProngMin = {"Dplus", "Lc", "Ds", "Xic", "Cd", "Ct", "Ch", "Ca"};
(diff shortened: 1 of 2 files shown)


fgrosa:
Hi @Marcellocosti, thanks! Please before merging this PR, it is important to profile the CPU consumption (for Pb-Pb especially).

Marcellocosti:
Hi @fgrosa, thanks for the suggestion! 

I just have a doubt: do you want me to profile the CPU consumption related to these code changes or to lower min-pt selections with respect to the ones currently used? Because with this implementation we will be anyway able to run with the min-pt selections used until now with no increase in CPU usage, given that just one check is added. And maybe hyperloop is best suited to monitor the impact of lowering the pt thresholds.

fgrosa:
> Hi @fgrosa, thanks for the suggestion!
> 
> I just have a doubt: do you want me to profile the CPU consumption related to these code changes or to lower min-pt selections with respect to the ones currently used? Because with this implementation we will be anyway able to run with the min-pt selections used until now with no increase in CPU usage, given that just one check is added. And maybe hyperloop is best suited to monitor the impact of lowering the pt thresholds.

I would say that at least you should check that without changing the cut (i.e. setting them all as we have now) it does not introduce an overhead. Then, I would suggest also to check what is the impact of lowering even only one since keep in mind that this usecase will be only useful for Pb-Pb, since in pp anyway there is no point in reducing it further.
For local tests, you can then use this tool to visualise the performance metrics: https://alimonitor.cern.ch/hyperloop/performance-graphs

Marcellocosti:
Okay, I will post here the checks you suggested before marking the PR as ready for review. For the use-cases, I was thinking more of lowering it in OO collisions (from the current 0.4 GeV/c), but of course it would be nice to see if also the Pb-Pb case can be lowered.

Marcellocosti:
Hi @fgrosa,

I performed local tests reconstructing 55 Pb-Pb collisions in 0-20% centrality. I performed 3 tests and I report the `cpuUsedPercentage` for each of them:


- Test 1: run the old version of the code with minimum track pt for 3-prongs > 0.6 GeV/c
image


- Test 2: run the new version of the code with minimum track pt for 3-prongs > 0.6 GeV/c
image


- Test 3: run the new version of the code with minimum track pt for 3-prongs > 0.4 GeV/c
image


Please let me know if this is what you had in mind or if you think additional checks are required. Thanks a lot!

fgrosa:
> Hi @fgrosa,
> 
> I performed local tests reconstructing 55 Pb-Pb collisions in 0-20% centrality. I performed 3 tests and I report the `cpuUsedPercentage` for each of them:
> 
>     * Test 1: run the old version of the code with minimum track pt for 3-prongs > 0.6 GeV/c
> 
> 
> image
> 
>     * Test 2: run the new version of the code with minimum track pt for 3-prongs > 0.6 GeV/c
> 
> 
> image
> 
>     * Test 3: run the new version of the code with minimum track pt for 3-prongs > 0.4 GeV/c
> 
> 
> image
> 
> Please let me know if this is what you had in mind or if you think additional checks are required. Thanks a lot!

Thanks @Marcellocosti! This shows that the implementation for the current cuts does not introduce overhead, so it's safe to be merged, but releasing the cut for central Pb-Pb collisions increases x3 the time, so it is prohibitive unless we implement other optimisations. This might instead work for OO however, given the lower combinatorics.

gluparel (dismissed):

PWGHF/Core/SelectorCuts.h
```diff
@@ -251,6 +251,17 @@ static const std::vector<std::string> labelsPt{};
...
+
+// Single prong min pt selection
+static constexpr int NSpecies3Prong = 8; // number of species for 3-prong candidates
+static const float cutsProngMinPt3Prong[NSpecies3Prong] = {0.4, 0.4, 0.4, 0.4, 0.4, 0.4, 0.4, 0.4};
+
+// 3-prong species labels
+static const std::vector<std::string> labelsSpecies3Prong = {"Dplus", "Lc", "Ds", "Xic", "Cd", "Ct", "Ch", "Ca"};
+
+// column label
+static const std::vector<std::string> labelsMinPt = {"Minimum track transverse momentum (GeV/c)"};
```
vkucera:
```suggestion
static const std::vector<std::string> labelsPtProngMin = {"Minimum prong transverse momentum (GeV/c)"};
```

PWGHF/Core/SelectorCuts.h
```diff
@@ -251,6 +251,17 @@ static const std::vector<std::string> labelsPt{};
 
 // column labels
 static const std::vector<std::string> labelsCutVar = {"massMin", "massMax", "cosp", "decL"};
+
+// Single prong min pt selection
+static constexpr int NSpecies3Prong = 8; // number of species for 3-prong candidates
```
vkucera:
```suggestion
static constexpr int NSpecies = 8; // number of species for 3-prong candidates
```

PWGHF/Core/SelectorCuts.h
```diff
@@ -251,6 +251,17 @@ static const std::vector<std::string> labelsPt{};
...
+static const float cutsProngMinPt3Prong[NSpecies3Prong] = {0.4, 0.4, 0.4, 0.4, 0.4, 0.4, 0.4, 0.4};
```
vkucera:
```suggestion
static const float ptProngMin[NSpecies3Prong] = {0.4, 0.4, 0.4, 0.4, 0.4, 0.4, 0.4, 0.4};
```

PWGHF/Core/SelectorCuts.h
```diff
@@ -251,6 +251,17 @@ static const std::vector<std::string> labelsPt{};
...
+
+// 3-prong species labels
+static const std::vector<std::string> labelsSpecies3Prong = {"Dplus", "Lc", "Ds", "Xic", "Cd", "Ct", "Ch", "Ca"};
```
vkucera:
```suggestion
static const std::vector<std::string> labelsPtProngMin = {"Dplus", "Lc", "Ds", "Xic", "Cd", "Ct", "Ch", "Ca"};
```

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -1456,6 +1456,9 @@ struct HfTrackIndexSkimCreator {
     Configurable<std::vector<double>> binsPtDstarToD0Pi{"binsPtDstarToD0Pi", std::vector<double>{hf_cuts_presel_dstar::vecBinsPt}, "pT bin limits for D*+->D0pi pT-dependent cuts"};
     Configurable<LabeledArray<double>> cutsDstarToD0Pi{"cutsDstarToD0Pi", {hf_cuts_presel_dstar::Cuts[0], hf_cuts_presel_dstar::NBinsPt, hf_cuts_presel_dstar::NCutVars, hf_cuts_presel_dstar::labelsPt, hf_cuts_presel_dstar::labelsCutVar}, "D*+->D0pi selections per pT bin"};
 
+    // Species-differential track min pT selection for 3-prong candidates
+    Configurable<LabeledArray<float>> cutsProngPtMin{"cutsProngPtMin", {hf_cuts_presel_3prong::cutsProngMinPt3Prong, hf_cuts_presel_3prong::NSpecies3Prong, 1, hf_cuts_presel_3prong::labelsSpecies3Prong, hf_cuts_presel_3prong::labelsMinPt}, "Min pT selection for prongs of 3-prong candidates in GeV/c"};
```
vkucera:
```suggestion
    Configurable<LabeledArray<float>> ptProngMin3Prong{"ptProngMin3Prong", {hf_cuts_presel_3prong::cutsProngMinPt3Prong, hf_cuts_presel_3prong::NSpecies3Prong, 1, hf_cuts_presel_3prong::labelsSpecies3Prong, hf_cuts_presel_3prong::labelsMinPt}, "Min pT selection for prongs of 3-prong candidates in GeV/c"};
```

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -1999,8 +2002,8 @@ struct HfTrackIndexSkimCreator {
   /// \param primVtx is the primary vertex
   /// \param cutStatus is a 2D array with outcome of each selection (filled only in debug mode)
   /// \param isSelected ia s bitmap with selection outcome
-  template <typename T1, typename T2, typename T3, typename T4>
-  void applySelection3Prong(const T1& pVecCand, const T2& secVtx, const T3& primVtx, T4& cutStatus, auto& isSelected)
+  template <typename T1, typename T2, typename T3, typename T4, typename T5>
+  void applySelection3Prong(const T1& pVecCand, const T2& prongDaugsPt, const T3& secVtx, const T4& primVtx, T5& cutStatus, auto& isSelected)
```
vkucera: Why does the type of `prongDaugsPt` need to be templated?

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -2643,6 +2657,7 @@ struct HfTrackIndexSkimCreator {
 
               auto trackParVarPos2 = getTrackParCov(trackPos2);
               std::array dcaInfoPos2{trackPos2.dcaXY(), trackPos2.dcaZ()};
+              std::array prongPtInfo{trackPos1.pt(), trackNeg1.pt(), trackPos2.pt()};
```
vkucera:
```suggestion
              std::array ptProngs{trackPos1.pt(), trackNeg1.pt(), trackPos2.pt()};
```

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -2917,6 +2932,7 @@ struct HfTrackIndexSkimCreator {
               auto trackNeg2 = trackIndexNeg2.template track_as<TTracks>();
               auto trackParVarNeg2 = getTrackParCov(trackNeg2);
               std::array dcaInfoNeg2{trackNeg2.dcaXY(), trackNeg2.dcaZ()};
+              std::array prongPtInfo{trackPos1.pt(), trackNeg1.pt(), trackNeg2.pt()};
```
vkucera:
```suggestion
              std::array ptProngs{trackPos1.pt(), trackNeg1.pt(), trackNeg2.pt()};
```

PWGHF/TableProducer/trackIndexSkimCreator.cxx
vkucera:
```suggestion
  template <typename T1, typename T2, typename T3, typename T4, typename T5>
  void applySelection3Prong(const T1& pVecCand, const T2& ptProngs, const T3& secVtx, const T4& primVtx, T5& cutStatus, auto& isSelected)
```

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -2038,6 +2041,17 @@ struct HfTrackIndexSkimCreator {
             }
           }
         }
+
+        // prong daughter pT
+        if ((config.debug || TESTBIT(isSelected, iDecay3P))) {
+          const auto minPtDaug = config.cutsProngPtMin->get(iDecay3P);
```
vkucera:
```suggestion
          const auto ptProngMin = config.cutsProngPtMin->get(iDecay3P);
```

vkucera (changes requested):

Marcellocosti:
Hi @vkucera, thanks for the feedback, I have implemented your comments!

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -1456,6 +1456,9 @@ struct HfTrackIndexSkimCreator {
     Configurable<std::vector<double>> binsPtDstarToD0Pi{"binsPtDstarToD0Pi", std::vector<double>{hf_cuts_presel_dstar::vecBinsPt}, "pT bin limits for D*+->D0pi pT-dependent cuts"};
     Configurable<LabeledArray<double>> cutsDstarToD0Pi{"cutsDstarToD0Pi", {hf_cuts_presel_dstar::Cuts[0], hf_cuts_presel_dstar::NBinsPt, hf_cuts_presel_dstar::NCutVars, hf_cuts_presel_dstar::labelsPt, hf_cuts_presel_dstar::labelsCutVar}, "D*+->D0pi selections per pT bin"};
 
+    // Species-differential track min pT selection for 3-prong candidates
+    Configurable<LabeledArray<float>> ptProngMin3Prong{"ptProngMin3Prong", {hf_cuts_presel_3prong::ptProngMin, hf_cuts_presel_3prong::NSpecies, 1, hf_cuts_presel_3prong::labelsPtProngMin, hf_cuts_presel_3prong::labelsMinPt}, "Min pT selection for prongs of 3-prong candidates in GeV/c"};
```
vkucera: I'm a bit confused. This looks like a declaration of a 2D array, but yours is 1D.

Marcellocosti: Hi Vit, this was done to have the entry in hyperloop looking like the current `onnxFileNames` of the workflow `hf-track-index-skim-creator`, which seems to me the best layout for this config

vkucera: But `onnxFileNames` is a 2D array.

Marcellocosti: I saw that `onnxFileNames` results in a table with rows containing a fixed particle name + a configurable, wouldn't the current implementation achieve the same result for `ptProngMin3Prong`? The `dpl-config.json` files from my local test show that, with this implementation, they have the same structure. Or maybe do you have in mind a cleaner alternative?

vkucera: Yes, they result in the same structure but your declaration of the default value does not have the same structure. Look for example at the declaration of `selectionsPid` which has the same data type as `ptProngMin3Prong`.

Marcellocosti: Thanks for the clarifications, I should have fixed the configurable now!

PWGHF/Core/SelectorCuts.h
```diff
@@ -251,6 +251,17 @@ static const std::vector<std::string> labelsPt{};
 
 // column labels
 static const std::vector<std::string> labelsCutVar = {"massMin", "massMax", "cosp", "decL"};
+
+// Single prong min pt selection
+static constexpr int NSpecies = 8; // number of species for 3-prong candidates
+static const float ptProngMin[NSpecies][1] = {{0.4}, {0.4}, {0.4}, {0.4}, {0.4}, {0.4}, {0.4}, {0.4}};
```
vkucera:
```suggestion
constexpr float ptProngMin[NSpecies][1] = {{0.4f}, {0.4f}, {0.4f}, {0.4f}, {0.4f}, {0.4f}, {0.4f}, {0.4f}};
```

PWGHF/Core/SelectorCuts.h
```diff
@@ -251,6 +251,17 @@ static const std::vector<std::string> labelsPt{};
...
+
+// 3-prong species labels
+static const std::vector<std::string> labelsPtProngMin = {"Dplus", "Lc", "Ds", "Xic", "Cd", "Ct", "Ch", "Ca"};
```
vkucera:
```suggestion
static const std::vector<std::string> labelsSpecies = {"Dplus", "Lc", "Ds", "Xic", "Cd", "Ct", "Ch", "Ca"};
```

PWGHF/Core/SelectorCuts.h
```diff
@@ -251,6 +251,17 @@ static const std::vector<std::string> labelsPt{};
...
+
+// column label
+static const std::vector<std::string> labelsMinPt = {"Minimum prong transverse momentum (GeV/c)"};
```
vkucera: It would be a super long axis label in the Hyperloop menu otherwise.
```suggestion
static const std::vector<std::string> labelsPtProngMin = {"minimum pT (GeV/c)"};
```

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -2038,6 +2041,17 @@ struct HfTrackIndexSkimCreator {
             }
           }
         }
+
+        // prong daughter pT
+        if ((config.debug || TESTBIT(isSelected, iDecay3P))) {
```
vkucera:
```suggestion
        if (config.debug || TESTBIT(isSelected, iDecay3P)) {
```

vkucera (commented):
Thanks for the updates. After the clarification, I also understand better what labels represent what. Please see my updated suggestions.

Marcellocosti:
Thanks @vkucera, I have implemented your comments!

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -2643,6 +2657,7 @@ struct HfTrackIndexSkimCreator {
 
               auto trackParVarPos2 = getTrackParCov(trackPos2);
               std::array dcaInfoPos2{trackPos2.dcaXY(), trackPos2.dcaZ()};
+              std::array ptProngs{trackPos1.pt(), trackNeg1.pt(), trackPos2.pt()};
```
vkucera: You create this variable 142 lines before it's needed while the program might not even reach there. Create variables as local as possible.

Marcellocosti: Thanks for spotting it, I have moved the declarations just before the function call.

PWGHF/Core/SelectorCuts.h
```diff
@@ -251,6 +251,17 @@ static const std::vector<std::string> labelsPt{};
 
 // column labels
 static const std::vector<std::string> labelsCutVar = {"massMin", "massMax", "cosp", "decL"};
+
+// Single prong min pt selection
+static constexpr int NSpecies = 8; // number of species for 3-prong candidates
+constexpr float ptProngMin[NSpecies][1] = {{0.4f}, {0.4f}, {0.4f}, {0.4f}, {0.4f}, {0.4f}, {0.4f}, {0.4f}};
```
vkucera:
```suggestion
constexpr float PtProngMin[NSpecies][1] = {{0.4f}, {0.4f}, {0.4f}, {0.4f}, {0.4f}, {0.4f}, {0.4f}, {0.4f}};
```

Marcellocosti: Thanks, fixed!

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -2038,6 +2041,17 @@ struct HfTrackIndexSkimCreator {
...
           }
         }
+
+        // prong daughter pT
+        if (config.debug || TESTBIT(isSelected, iDecay3P)) {
+          const auto ptProngMin = config.ptProngMin3Prong->get(iDecay3P);
+          if (ptProngs[0] < ptProngMin || ptProngs[1] < ptProngMin || ptProngs[2] < ptProngMin) {
+            CLRBIT(isSelected, iDecay3P);
+            if (config.debug) {
+              cutStatus[iDecay3P][4] = false;
```
vkucera, line 2048: This seems like you are using an existing reserved slot. Should you not extend the numbers of cuts instead?

Marcellocosti:
Hi @vkucera, thanks for spotting this! Now I have extended the cuts by one for the 3-prongs and for the Ds, which has a separate namespace because it has an additional cut on the phi mass. Please let me know if you have further suggestions!

PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -2038,6 +2038,17 @@ struct HfTrackIndexSkimCreator {
             }
           }
         }
+
+        // prong daughter pT
```
vkucera:
```suggestion
        // prong pT
```

vkucera (approved):
Thanks @Marcellocosti , it looks good to me now.

vkucera:
@fgrosa @gluparel Back to you.

stefanopolitano (approved):

PR #15780, opened 2026-04-15
[PWGLF] Change of PID selections for 'k1analysis.cxx'

SuJeong-Ji:
Previous PID selection only relied on the TPC nsigma cut, and the change is made to apply a circular cut.


dmallick2:
As checked with analyzer regarding o2-linter issue related to  PDG, as these particle is not available , it takes some time do for PR . For data, this is not required. for time being it is approved ignoring o2-linter issue. it will take care later.

dmallick2 (approved):

PR #15783, opened 2026-04-15
[PWGDQ] Improve memory consumption of DQ filter task

fgrosa:
Memory improvement for DQ filter, as discussed in the trigger meeting today: https://indico.cern.ch/event/1674294/?note=378937#27-general-news-2026-menu 
@lupengzhong @lucamicheletti93

### DIFF
--- a/PWGDQ/Core/HistogramsLibrary.cxx
+++ b/PWGDQ/Core/HistogramsLibrary.cxx
@@ -2339,6 +2339,40 @@ void o2::aod::dqhistograms::DefineHistograms(HistogramManager* hm, const char* h
     hm->AddHistogram(histClass, "Mass_Pt", "", false, 500, 0.0, 5.0, VarManager::kMassDau, 200, 0.0, 20.0, VarManager::kPt);
     hm->AddHistogram(histClass, "Rapidity", "", false, 400, -4.0, 4.0, VarManager::kRap);
   }
+
+  // specific group/subgroups for trigger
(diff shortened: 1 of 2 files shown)


lupengzhong:
Hi @fgrosa,

I have checked the new and old AnalysisResults.root files that you produced with this PR or not. As you mentioned in today’s trigger meeting, I can confirm that the two outputs are identical, and all the necessary information for the trigger QA is included in the new root file. I am able to generate the same QA plots with both root files.

lucamicheletti93:
Thanks @fgrosa ! The modifications looks fine to me. Just one extra comment, if you want to add this modification to the new tag for the trigger probably you need to include also this PR: https://github.com/AliceO2Group/O2Physics/pull/15784 . It avoids the increase of memory consumption due to the load of the full geometry.

fgrosa:
> Thanks @fgrosa ! The modifications looks fine to me. Just one extra comment, if you want to add this modification to the new tag for the trigger probably you need to include also this PR: #15784 . It avoids the increase of memory consumption due to the load of the full geometry.

Hi @lucamicheletti93, thanks but I don't need to include also the other one, since I will only cherry-pick this on top of the current CEFP tag. Can you please approve this one if it's good for you? Thanks!

lucamicheletti93 (approved):

lucamicheletti93:
@fgrosa Thanks for the clarification!! I have approved it 😉

PR #15786, opened 2026-04-15
[PWGHF] Separate UPC Lc analysis into a dedicated task file

Rrantu:
In this PR, I move the original UPC Lc part from taskLc.cxx into a separate file, taskUpcLc.cxx.

### DIFF
--- a/PWGHF/D2H/Tasks/taskLc.cxx
+++ b/PWGHF/D2H/Tasks/taskLc.cxx
@@ -17,7 +17,6 @@
 /// \author Vít Kučera <vit.kucera@cern.ch>, CERN
 /// \author Annalena Kalteyer <annalena.sophie.kalteyer@cern.ch>, GSI Darmstadt
 /// \author Biao Zhang <biao.zhang@cern.ch>, Heidelberg University
-/// \author Ran Tu <ran.tu@cern.ch>, Fudan University
 /// \author Oleksii Lubynets <oleksii.lubynets@cern.ch>, Heidelberg University, GSI Darmstadt
 
 #include "PWGHF/Core/CentralityEstimation.h"
@@ -29,9 +28,6 @@
 #include "PWGHF/DataModel/CandidateSelectionTables.h"
 #include "PWGHF/DataModel/TrackIndexSkimmingTables.h"
 #include "PWGHF/Utils/utilsEvSelHf.h"
-#include "PWGHF/Utils/utilsUpcHf.h"
-#include "PWGUD/Core/SGSelector.h"
-#include "PWGUD/Core/UPCHelpers.h"
 
 #include "Common/Core/RecoDecay.h"
 #include "Common/DataModel/Centrality.h"
@@ -69,7 +65,6 @@ using namespace o2::framework::expressions;
 using namespace o2::hf_centrality;
 using namespace o2::hf_occupancy;
 using namespace o2::hf_evsel;
-using namespace o2::analysis::hf_upc;
 
 /// Λc± → p± K∓ π± analysis task
 struct HfTaskLc {
@@ -79,7 +74,6 @@ struct HfTaskLc {
   Configurable<std::vector<double>> binsPt{"binsPt", std::vector<double>{hf_cuts_lc_to_p_k_pi::vecBinsPt}, "pT bin limits"};
   // ThnSparse for ML outputScores and Vars
   Configurable<bool> fillTHn{"fillTHn", false, "fill THn"};
-  Configurable<bool> fillUPCTHnLite{"fillUPCTHnLite", false, "fill THn"};
   Configurable<bool> storeOccupancy{"storeOccupancy", true, "Flag to store occupancy information"};
   Configurable<int> occEstimator{"occEstimator", 2, "Occupancy estimation (None: 0, ITS: 1, FT0C: 2)"};
   Configurable<bool> storeProperLifetime{"storeProperLifetime", false, "Flag to store proper lifetime"};
@@ -88,8 +82,7 @@ struct HfTaskLc {
   Configurable<std::string> ccdbPathGrp{"ccdbPathGrp", "GLO/GRP/GRP", "Path of the grp file (Run 2)"};
   Configurable<std::string> ccdbPathGrpMag{"ccdbPathGrpMag", "GLO/Config/GRPMagField", "CCDB path of the GRPMagField object (Run 3)"};
 
-  HfEventSelection hfEvSel;         // event selection and monitoring
-  HfUpcGapThresholds upcThresholds; // UPC gap determination thresholds
+  HfEventSelection hfEvSel; // event selection and monitoring
   SliceCache cache;
   Service<o2::ccdb::BasicCCDBManager> ccdb{};
 
@@ -127,13 +120,7 @@ struct HfTaskLc {
   ConfigurableAxis thnConfigAxisNumPvContr{"thnConfigAxisNumPvContr", {200, -0.5, 199.5}, "Number of PV contributors"};
   ConfigurableAxis thnConfigAxisOccupancy{"thnConfigAxisOccupancy", {14, 0, 14000}, "axis for centrality"};
   ConfigurableAxis thnConfigAxisProperLifetime{"thnConfigAxisProperLifetime", {200, 0, 2}, "Proper lifetime, ps"};
-  ConfigurableAxis thnConfigAxisGapType{"thnConfigAxisGapType", {7, -1.5, 5.5}, "axis for UPC gap type (see TrueGap enum in o2::aod::sgselector)"};
-  ConfigurableAxis thnConfigAxisFV0A{"thnConfigAxisFV0A", {1001, -1.5, 999.5}, "axis for FV0-A amplitude (a.u.)"};
-  ConfigurableAxis thnConfigAxisFT0{"thnConfigAxisFT0", {1001, -1.5, 999.5}, "axis for FT0 amplitude (a.u.)"};
-  ConfigurableAxis thnConfigAxisZN{"thnConfigAxisZN", {510, -1.5, 49.5}, "axis for ZN energy (a.u.)"};
-  ConfigurableAxis thnConfigAxisZNTime{"thnConfigAxisZNTime", {200, -10, 10}, "axis for ZN energy (a.u.)"};
   HistogramRegistry registry{"registry", {}};
-  HistogramRegistry qaRegistry{"QAHistos", {}, OutputObjHandlingPolicy::AnalysisObject};
 
   // Factors for conversion between units
   constexpr static float CtToProperLifetimePs = 1.f / o2::constants::physics::LightSpeedCm2PS;
@@ -157,13 +144,12 @@ struct HfTaskLc {
 
   void init(InitContext&)
   {
-    const std::array<bool, 14> doprocess{doprocessDataStd, doprocessDataStdWithFT0C, doprocessDataStdWithFT0M, doprocessDataWithMl, doprocessDataWithMlWithFT0C, doprocessDataWithMlWithFT0M, doprocessDataWithMlWithUpc, doprocessMcStd, doprocessMcStdWithFT0C, doprocessMcStdWithFT0M, doprocessMcWithMl, doprocessMcWithMlWithFT0C, doprocessMcWithMlWithFT0M, doprocessDataStdWithUpc};
+    const std::array<bool, 12> doprocess{doprocessDataStd, doprocessDataStdWithFT0C, doprocessDataStdWithFT0M, doprocessDataWithMl, doprocessDataWithMlWithFT0C, doprocessDataWithMlWithFT0M, doprocessMcStd, doprocessMcStdWithFT0C, doprocessMcStdWithFT0M, doprocessMcWithMl, doprocessMcWithMlWithFT0C, doprocessMcWithMlWithFT0M};
     if ((std::accumulate(doprocess.begin(), doprocess.end(), 0)) != 1) {
       LOGP(fatal, "no or more than one process function enabled! Please check your configuration!");
     }
 
-    const bool isData = doprocessDataStd || doprocessDataStdWithFT0C || doprocessDataStdWithFT0M || doprocessDataWithMl || doprocessDataWithMlWithFT0C || doprocessDataWithMlWithFT0M || doprocessDataWithMlWithUpc;
-    const bool isUpc = doprocessDataWithMlWithUpc || doprocessDataStdWithUpc;
+    const bool isData = doprocessDataStd || doprocessDataStdWithFT0C || doprocessDataStdWithFT0M || doprocessDataWithMl || doprocessDataWithMlWithFT0C || doprocessDataWithMlWithFT0M;
 
     auto addHistogramsRec = [&](const std::string& histoName, const std::string& xAxisTitle, const std::string& yAxisTitle, const HistogramConfigSpec& configSpec) {
       if (isData) {
@@ -266,14 +252,6 @@ struct HfTaskLc {
     /// decay length error
     addHistogramsRec("hDecLenErrVsPt", "decay length error (cm)", "#it{p}_{T} (GeV/#it{c})", {HistType::kTH2F, {{100, 0., 1.}, {vbins}}});
 
-    if (isUpc) {
-      qaRegistry.add("Data/fitInfo/ampFT0A_vs_ampFT0C", "FT0-A vs FT0-C amplitude;FT0-A amplitude (a.u.);FT0-C amplitude (a.u.)", {HistType::kTH2F, {{500, 0., 500}, {500, 0., 500}}});
-      qaRegistry.add("Data/zdc/energyZNA_vs_energyZNC", "ZNA vs ZNC common energy;E_{ZNA}^{common} (a.u.);E_{ZNC}^{common} (a.u.)", {HistType::kTH2F, {{100, 0., 10}, {100, 0., 10}}});
-      qaRegistry.add("Data/zdc/timeZNA_vs_timeZNC", "ZNA vs ZNC time;ZNA Time;ZNC time", {HistType::kTH2F, {{200, -10., 10}, {200, -10., 10}}});
-      qaRegistry.add("Data/hUpcGapAfterSelection", "UPC gap type after selection;Gap side;Counts", {HistType::kTH1F, {{7, -1.5, 5.5}}});
-      qaRegistry.add("Data/hUpcMulti", "Multiplicity of UPC events;Multiplicity;Counts", {HistType::kTH1F, {{200, -0.5, 199.5}}});
-      qaRegistry.add("Data/hUpcVtz", "Vertex Z position of UPC events;Vz (cm);Counts", {HistType::kTH1F, {{200, -10., 10.}}});
-    }
     if (fillTHn) {
       const AxisSpec thnAxisMass{thnConfigAxisMass, "inv. mass (p K #pi) (GeV/#it{c}^{2})"};
       const AxisSpec thnAxisPt{thnConfigAxisPt, "#it{p}_{T}(#Lambda_{c}^{+}) (GeV/#it{c})"};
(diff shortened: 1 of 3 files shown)


vkucera:
@Rrantu Columns and tables are parts of the data model. They don't belong in the utility headers.

Rrantu:
Hi @vkucera, thanks for the comment, where would you recommend moving the columns and tables to? Would it be acceptable to define them directly in the task?

vkucera:
> Hi @vkucera, thanks for the comment, where would you recommend moving the columns and tables to? Would it be acceptable to define them directly in the task?

Yes, if the tables are not used anywhere else, please put them in the task.

PWGHF/D2H/Tasks/taskLc.cxx
```diff
@@ -71,15 +71,77 @@ using namespace o2::hf_occupancy;
...
+DECLARE_SOA_COLUMN(DecayLength, decayLength, float);
+DECLARE_SOA_COLUMN(CPA, cPA, float);
+DECLARE_SOA_COLUMN(PvContributors, pvContributors, float);
+DECLARE_SOA_COLUMN(AmpFV0A, ampFV0A, float);
+DECLARE_SOA_COLUMN(AmpFT0A, ampFT0A, float);
+DECLARE_SOA_COLUMN(AmpFT0C, ampFT0C, float);
+DECLARE_SOA_COLUMN(ZdcEnergyZNA, zdcEnergyZNA, float);
+DECLARE_SOA_COLUMN(ZdcEnergyZNC, zdcEnergyZNC, float);
+DECLARE_SOA_COLUMN(ZdcTimeZNA, zdcTimeZNA, float);
+DECLARE_SOA_COLUMN(ZdcTimeZNC, zdcTimeZNC, float);
```
vkucera: Follow the naming conventions.

```suggestion
DECLARE_SOA_COLUMN(FdScore, fdScore, float);
DECLARE_SOA_COLUMN(PtProng0, ptProng0, float);
DECLARE_SOA_COLUMN(PtProng1, ptProng1, float);
DECLARE_SOA_COLUMN(PtProng2, ptProng2, float);
DECLARE_SOA_COLUMN(Chi2PCA, chi2PCA, float);
DECLARE_SOA_COLUMN(DecayLength, decayLength, float);
DECLARE_SOA_COLUMN(Cpa, cpa, float);
DECLARE_SOA_COLUMN(PvContributors, pvContributors, float);
DECLARE_SOA_COLUMN(AmpFV0A, ampFV0A, float);
DECLARE_SOA_COLUMN(AmpFT0A, ampFT0A, float);
DECLARE_SOA_COLUMN(AmpFT0C, ampFT0C, float);
DECLARE_SOA_COLUMN(ZdcEnergyZNA, zdcEnergyZNA, float);
DECLARE_SOA_COLUMN(ZdcEnergyZNC, zdcEnergyZNC, float);
DECLARE_SOA_COLUMN(ZdcTimeZNA, zdcTimeZNA, float);
DECLARE_SOA_COLUMN(ZdcTimeZNC, zdcTimeZNC, float);
```

Rrantu:
> > Hi @vkucera, thanks for the comment, where would you recommend moving the columns and tables to? Would it be acceptable to define them directly in the task?
> 
> Yes, if the tables are not used anywhere else, please put them in the task.

Thanks! I‘ve moved them to the task file

PWGHF/D2H/Tasks/taskLc.cxx
```diff
@@ -71,15 +71,77 @@ using namespace o2::hf_occupancy;
...
+                  full::AmpFT0C,
+                  full::ZdcEnergyZNA,
+                  full::ZdcEnergyZNC,
+                  full::ZdcTimeZNA,
+                  full::ZdcTimeZNC);
+
+DECLARE_SOA_TABLE(HfUpcLcInfos, "AOD", "HFUPCLCINFOS",
+                  full::M,
+                  full::Pt,
+                  full::PtProng0,
```
vkucera: Why do you duplicate columns in both tables?

PWGHF/D2H/Tasks/taskLc.cxx
```diff
@@ -71,15 +71,77 @@ using namespace o2::hf_occupancy;
...
+DECLARE_SOA_COLUMN(AmpFV0A, ampFV0A, float);
+DECLARE_SOA_COLUMN(AmpFT0A, ampFT0A, float);
+DECLARE_SOA_COLUMN(AmpFT0C, ampFT0C, float);
+DECLARE_SOA_COLUMN(ZdcEnergyZNA, zdcEnergyZNA, float);
+DECLARE_SOA_COLUMN(ZdcEnergyZNC, zdcEnergyZNC, float);
+DECLARE_SOA_COLUMN(ZdcTimeZNA, zdcTimeZNA, float);
+DECLARE_SOA_COLUMN(ZdcTimeZNC, zdcTimeZNC, float);
+} // namespace full
+
+DECLARE_SOA_TABLE(HfUpcLcBdtInfos, "AOD", "HFUPCLCBDTINFOS",
```
vkucera: Why do you call this table BDT?

PWGHF/D2H/Tasks/taskLc.cxx
```diff
@@ -71,15 +71,77 @@ using namespace o2::hf_occupancy;
...
+  Produces<o2::aod::HfUpcLcInfos> rowCandUpc;
+
   Configurable<int> selectionFlagLc{"selectionFlagLc", 1, "Selection Flag for Lc"};
   Configurable<double> yCandGenMax{"yCandGenMax", 0.5, "max. gen particle rapidity"};
   Configurable<double> yCandRecoMax{"yCandRecoMax", 0.8, "max. cand. rapidity"};
   Configurable<std::vector<double>> binsPt{"binsPt", std::vector<double>{hf_cuts_lc_to_p_k_pi::vecBinsPt}, "pT bin limits"};
   // ThnSparse for ML outputScores and Vars
   Configurable<bool> fillTHn{"fillTHn", false, "fill THn"};
-  Configurable<bool> fillUPCTHnLite{"fillUPCTHnLite", false, "fill THn"};
+  Configurable<bool> fillUPCTreeLite{"fillUPCTreeLite", false, "fill THn"};
```
vkucera: You only have a single option to fill the trees. It seems that the attributes `UPC` and `Lite` are redundant.

vkucera (changes requested):

PWGHF/D2H/Tasks/taskLc.cxx
vkucera: And the comment is wrong.

vkucera:
@Rrantu A conceptual question: Why can you not use the THnSparse?

PWGHF/D2H/Tasks/taskLc.cxx
Rrantu: Thanks for the comment! The Lite one fills fewer candidates, including only those passing the single-gap selection.

PWGHF/D2H/Tasks/taskLc.cxx
Rrantu: Because the tree includes BDT scores for the candidates.

PWGHF/D2H/Tasks/taskLc.cxx
Rrantu: One table corresponds to the BDT-applied version, while the other does not. However, both require the mass and pt information.

Rrantu:
> @Rrantu A conceptual question: Why can you not use the THnSparse?

Thanks for the question! For large statistics, merging THnSparse objects becomes very heavy, making it difficult to obtain merged AnalysisResults.root. I also need to repeatedly adjust selections on multiple axes during the offline analysis. Projections from high-dimensional THnSparse are not very flexible.

PWGHF/D2H/Tasks/taskLc.cxx
vkucera: You never fill both tables. It would be better to have one table with the common columns that you fill in any case and a second table with the extra columns for the BDT case.

PWGHF/D2H/Tasks/taskLc.cxx
vkucera: But it also includes other BDT-unrelated columns. See my comment above.

lubynets:
Dear @Rrantu, @vkucera, all codeowners, sorry for interfering,
Are `Tasks`, and not `TableProducer`s, supposed to produce trees? https://aliceo2group.github.io/analysis-framework/docs/gettingstarted/theo2physicsrepo.html#folder-structure

PWGHF/D2H/Tasks/taskLc.cxx
vkucera: Then it should be called something like `fillTreeOnlySingleGap` and explained in the comment.

vkucera:
> Dear @Rrantu, @vkucera, all codeowners, sorry for interfering, Are `Tasks`, and not `TableProducer`s, supposed to produce trees? https://aliceo2group.github.io/analysis-framework/docs/gettingstarted/theo2physicsrepo.html#folder-structure

No, they are not.

PWGHF/D2H/Tasks/taskLc.cxx
Rrantu: Thanks for the suggestion. Here the detector variables and the BDT output are both required together for offline selections. Given this use case, I would prefer to keep the current structure.

Rrantu:
Hi @lubynets, Thanks for the question. In this case, the tables are only used within this task, and the trees are produced specifically for this analysis. Also, there are other tasks that produce trees in a similar way.

PWGHF/D2H/Tasks/taskLc.cxx
Rrantu: Thanks for the suggestion, I have renamed it.

Rrantu:
Hi @vkucera, could you please take a look at my updated code?

PWGHF/D2H/Tasks/taskLc.cxx
```diff
@@ -71,15 +71,82 @@ using namespace o2::hf_occupancy;
...
+
   Configurable<int> selectionFlagLc{"selectionFlagLc", 1, "Selection Flag for Lc"};
   Configurable<double> yCandGenMax{"yCandGenMax", 0.5, "max. gen particle rapidity"};
   Configurable<double> yCandRecoMax{"yCandRecoMax", 0.8, "max. cand. rapidity"};
   Configurable<std::vector<double>> binsPt{"binsPt", std::vector<double>{hf_cuts_lc_to_p_k_pi::vecBinsPt}, "pT bin limits"};
   // ThnSparse for ML outputScores and Vars
   Configurable<bool> fillTHn{"fillTHn", false, "fill THn"};
-  Configurable<bool> fillUPCTHnLite{"fillUPCTHnLite", false, "fill THn"};
+  Configurable<bool> fillTreeOnlySingleGap{"fillTreeOnlySingleGap", false, "fill THn"};
+  Configurable<bool> fillTreeUPCQa{"fillTreeUPCQa", false, "fill THn"};
```
vkucera:
```suggestion
  Configurable<bool> fillTreeOnlySingleGap{"fillTreeOnlySingleGap", false, "fill THn"};
  Configurable<bool> fillTreeUpcQa{"fillTreeUpcQa", false, "fill THn"};
```
Also use some relevant documentation comments.

vkucera: You still have not put any meaningful documentation comments.

Rrantu: Sorry, I missed that earlier. I've added documentation comments now.

Rrantu:
Hi @vkucera, when you have time, could you please take a look at my updated? Thanks!

PWGHF/D2H/Tasks/taskLc.cxx
```diff
@@ -71,15 +71,84 @@ using namespace o2::hf_occupancy;
...
   Configurable<double> yCandGenMax{"yCandGenMax", 0.5, "max. gen particle rapidity"};
   Configurable<double> yCandRecoMax{"yCandRecoMax", 0.8, "max. cand. rapidity"};
   Configurable<std::vector<double>> binsPt{"binsPt", std::vector<double>{hf_cuts_lc_to_p_k_pi::vecBinsPt}, "pT bin limits"};
   // ThnSparse for ML outputScores and Vars
   Configurable<bool> fillTHn{"fillTHn", false, "fill THn"};
-  Configurable<bool> fillUPCTHnLite{"fillUPCTHnLite", false, "fill THn"};
+  // Option to fill TTree for candidates passing only single gap condition (for UPC analysis)
+  Configurable<bool> fillTreeOnlySingleGap{"fillTreeOnlySingleGap", false, "fill THn"};
+  // Option to fill TTree for UPC QA (for UPC analysis)
+  Configurable<bool> fillTreeUpcQa{"fillTreeUpcQa", false, "fill THn"};
```
vkucera: These are not documentation comments. You are supposed to document the function of the configurable in its definition, not outside.

Rrantu: Thanks for the correction. I've updated it accordingly.

vkucera: Thanks but you left the redundant comment lines there. Please remove them.

Rrantu: Thanks, I've removed them.

PWGHF/D2H/Tasks/taskLc.cxx
```diff
@@ -71,15 +71,82 @@ using namespace o2::hf_occupancy;
...
+  Produces<o2::aod::HfUpcLcBdtInfos> rowCandUpcBdt;
+  Produces<o2::aod::HfUpcLcInfos> rowCandUpc;
+  Produces<o2::aod::HfUpcQa> rowCandUpcQa;
+
   Configurable<int> selectionFlagLc{"selectionFlagLc", 1, "Selection Flag for Lc"};
   Configurable<double> yCandGenMax{"yCandGenMax", 0.5, "max. gen particle rapidity"};
   Configurable<double> yCandRecoMax{"yCandRecoMax", 0.8, "max. cand. rapidity"};
   Configurable<std::vector<double>> binsPt{"binsPt", std::vector<double>{hf_cuts_lc_to_p_k_pi::vecBinsPt}, "pT bin limits"};
   // ThnSparse for ML outputScores and Vars
   Configurable<bool> fillTHn{"fillTHn", false, "fill THn"};
```
vkucera, line 81: This comment should also go in the documentation string.

PWGHF/D2H/Tasks/taskLc.cxx
```diff
@@ -819,27 +863,27 @@ struct HfTaskLc {
                 outputFD = mlProb[MlClassNonPrompt];   /// non-prompt score
               }
               /// Fill the ML outputScores and variables of candidate
-              if (fillUPCTHnLite) {
+              if (fillTreeOnlySingleGap) {
                 if (gap == o2::aod::sgselector::TrueGap::SingleGapA || gap == o2::aod::sgselector::TrueGap::SingleGapC) {
-                  std::vector<double> valuesToFill{massLc, pt, outputBkg, outputPrompt, outputFD, static_cast<double>(numPvContributors), static_cast<double>(fitInfo.ampFV0A), static_cast<double>(fitInfo.ampFT0A), static_cast<double>(fitInfo.ampFT0C), static_cast<double>(zdcEnergyZNA), static_cast<double>(zdcEnergyZNC), static_cast<double>(zdcTimeZNA), static_cast<double>(zdcTimeZNC)};
-                  registry.get<THnSparse>(HIST("hnLcUpcVarsWithBdt"))->Fill(valuesToFill.data());
+                  rowCandUpcBdt(massLc, pt, outputBkg, outputPrompt, outputFD, static_cast<float>(fitInfo.ampFV0A), static_cast<float>(fitInfo.ampFT0A), static_cast<float>(fitInfo.ampFT0C), static_cast<float>(zdcTimeZNA), static_cast<float>(zdcTimeZNC));
```
vkucera: What types are the values before casting?

Rrantu: https://github.com/AliceO2Group/O2Physics/blob/a839c00a43566a199475fd87663c4a26c5644a98/PWGUD/Core/UDHelpers.h#L300
Before casting the values are of type float.

gluparel:
> > Dear @Rrantu, @vkucera, all codeowners, sorry for interfering, Are `Tasks`, and not `TableProducer`s, supposed to produce trees? https://aliceo2group.github.io/analysis-framework/docs/gettingstarted/theo2physicsrepo.html#folder-structure
> 
> No, they are not.

Hi @vkucera, actually I am in favour of what suggested by @lubynets and I would avoid to add trees here and I would suggest to use the Lc TreeCreator: https://github.com/AliceO2Group/O2Physics/blob/master/PWGHF/TableProducer/treeCreatorLcToPKPi.cxx

Grazia

PWGHF/D2H/Tasks/taskLc.cxx
vkucera: So why do you cast them to the same type?

vkucera:
> > > Dear @Rrantu, @vkucera, all codeowners, sorry for interfering, Are `Tasks`, and not `TableProducer`s, supposed to produce trees? https://aliceo2group.github.io/analysis-framework/docs/gettingstarted/theo2physicsrepo.html#folder-structure
> > 
> > 
> > No, they are not.
> 
> Hi @vkucera, actually I am in favour of what suggested by @lubynets and I would avoid to add trees here and I would suggest to use the Lc TreeCreator: https://github.com/AliceO2Group/O2Physics/blob/master/PWGHF/TableProducer/treeCreatorLcToPKPi.cxx
> 
> Grazia

Hi @gluparel , what are your reasons in favour of that?
My main reason against is that the logic of selecting candidates and filling them in the trees in the task is quite complicated and very different from the way the tree creator output is filled. Adding the same logic in the tree creator for one specific analysis does not seem justified to me and it would cause code duplication which would bring further maintenance work.

zhangbiao-phy:
> > > > Dear @Rrantu, @vkucera, all codeowners, sorry for interfering, Are `Tasks`, and not `TableProducer`s, supposed to produce trees? https://aliceo2group.github.io/analysis-framework/docs/gettingstarted/theo2physicsrepo.html#folder-structure
> > > 
> > > 
> > > No, they are not.
> > 
> > 
> > Hi @vkucera, actually I am in favour of what suggested by @lubynets and I would avoid to add trees here and I would suggest to use the Lc TreeCreator: https://github.com/AliceO2Group/O2Physics/blob/master/PWGHF/TableProducer/treeCreatorLcToPKPi.cxx
> > Grazia
> 
> Hi @gluparel , what are your reasons in favour of that? My main reason against is that the logic of selecting candidates and filling them in the trees in the task is quite complicated and very different from the way the tree creator output is filled. Adding the same logic in the tree creator for one specific analysis does not seem justified to me and it would cause code duplication which would bring further maintenance work.

Hi @vkucera @gluparel @lubynets, I agree with Vít’s comments here. Regarding adding a tree to the Lc task, I also understand the concerns raised by Grazia and Oleksii. To avoid making things more complicated, I would suggest that @Rrantu create a new dedicated task in the D2H folder (e.g. taskLcUPC.cxx, or more generally taskCharmUPC.cxx). We could also consider using such a task for all HF UPC analyses (different process function for different particle). This would be more appropriate since the event selection and logic are quite different from the standard Lc or other HF task. what do you think?

gluparel:
> > > > > Dear @Rrantu, @vkucera, all codeowners, sorry for interfering, Are `Tasks`, and not `TableProducer`s, supposed to produce trees? https://aliceo2group.github.io/analysis-framework/docs/gettingstarted/theo2physicsrepo.html#folder-structure
> > > > 
> > > > 
> > > > No, they are not.
> > > 
> > > 
> > > Hi @vkucera, actually I am in favour of what suggested by @lubynets and I would avoid to add trees here and I would suggest to use the Lc TreeCreator: https://github.com/AliceO2Group/O2Physics/blob/master/PWGHF/TableProducer/treeCreatorLcToPKPi.cxx
> > > Grazia
> > 
> > 
> > Hi @gluparel , what are your reasons in favour of that? My main reason against is that the logic of selecting candidates and filling them in the trees in the task is quite complicated and very different from the way the tree creator output is filled. Adding the same logic in the tree creator for one specific analysis does not seem justified to me and it would cause code duplication which would bring further maintenance work.
> 
> Hi @vkucera @gluparel @lubynets, I agree with Vít’s comments here. Regarding adding a tree to the Lc task, I also understand the concerns raised by Grazia and Oleksii. To avoid making things more complicated, I would suggest that @Rrantu create a new dedicated task in the D2H folder (e.g. taskLcUPC.cxx, or more generally taskCharmUPC.cxx). We could also consider using such a task for all HF UPC analyses (different process function for different particle). This would be more appropriate since the event selection and logic are quite different from the standard Lc or other HF task. what do you think?

Hello @zhangbiao-phy , @vkucera , my point is that this task is used by various analyses and we would start to add various trees for the various analyses which need different information. Is this the direction we would like to take? Another option is to have a dedicated task for HF UPC analysis as indeed suggested by Biao and this could be indeed a good option for me, since the various UPC analysis use the same event selections but just a different particle creator.

vkucera:
Hi @gluparel @zhangbiao-phy , thanks for your feedback. I like the idea of a common workflow for UPC analyses if a majority of the code can be shared by multiple species. And if the primary output is expected to have many dimensions, it would indeed be worth making it as a tree creator, as @lubynets pointed out.

Rrantu:
> Hi @gluparel @zhangbiao-phy , thanks for your feedback. I like the idea of a common workflow for UPC analyses if a majority of the code can be shared by multiple species. And if the primary output is expected to have many dimensions, it would indeed be worth making it as a tree creator, as @lubynets pointed out.

Thanks for the suggestion. I will try to separate the UPC part into a new task file

vkucera:
> > Hi @gluparel @zhangbiao-phy , thanks for your feedback. I like the idea of a common workflow for UPC analyses if a majority of the code can be shared by multiple species. And if the primary output is expected to have many dimensions, it would indeed be worth making it as a tree creator, as @lubynets pointed out.
> 
> Thanks for the suggestion. I will try to separate the UPC part into a new task file

Thanks @Rrantu , feel free to mark the PR as ready for review when you push the updates. Please do not create a new PR. Just updated this one.

Rrantu:
> > > Hi @gluparel @zhangbiao-phy , thanks for your feedback. I like the idea of a common workflow for UPC analyses if a majority of the code can be shared by multiple species. And if the primary output is expected to have many dimensions, it would indeed be worth making it as a tree creator, as @lubynets pointed out.
> > 
> > 
> > Thanks for the suggestion. I will try to separate the UPC part into a new task file
> 
> Thanks @Rrantu , feel free to mark the PR as ready for review when you push the updates. Please do not create a new PR. Just updated this one.

Sorry, when I noticed the comment, I had already updated my branch, so this PR was automatically closed by GitHub. I have created a new PR here: https://github.com/AliceO2Group/O2Physics/pull/16132

vkucera:
> > > > Hi @gluparel @zhangbiao-phy , thanks for your feedback. I like the idea of a common workflow for UPC analyses if a majority of the code can be shared by multiple species. And if the primary output is expected to have many dimensions, it would indeed be worth making it as a tree creator, as @lubynets pointed out.
> > > 
> > > 
> > > Thanks for the suggestion. I will try to separate the UPC part into a new task file
> > 
> > 
> > Thanks @Rrantu , feel free to mark the PR as ready for review when you push the updates. Please do not create a new PR. Just updated this one.
> 
> Sorry, when I noticed the comment, I had already updated my branch, so this PR was automatically closed by GitHub. I have created a new PR here: #16132

As I said before, don't open a new PR.

vkucera:
@Rrantu Please fix the warning and update the PR title.

Rrantu:
> @Rrantu Please fix the warning and update the PR title.

Thanks for the commet, I have updated the PR title. Regarding the warning, it seems that it was not introduced by my changes.

vkucera:
> > @Rrantu Please fix the warning and update the PR title.
> 
> Thanks for the commet, I have updated the PR title. Regarding the warning, it seems that it was not introduced by my changes.

What makes you think so?

PWGHF/D2H/Tasks/taskLc.cxx
vkucera: Still there

vkucera (dismissed):
Good from my side.

zhangbiao-phy (approved):

PR #15797, opened 2026-04-15
[PWGEM] remove unnecessary variables in EMTrack

dsekihat:
Remove unnecessary variables in EMTrack for dilepton PAG.
Add EMPhoton for photon PAG.
This PR should reduce memory in event mixing for both dilepton and photon PAG.


alibuild (approved):

PR #15818, opened 2026-04-16
[PWGCF] Define + fill correlation tables for both MC gen and reco levels

abmodak:


jaelpark (approved):

PWGCF/TwoParticleCorrelations/TableProducer/longrangeMaker.cxx
```diff
@@ -502,6 +495,173 @@ struct LongrangeMaker {
     } // SG events
   }
 
+  using ColMCTrueTable = soa::Join<aod::McCollisions, aod::MultMCExtras, aod::McCollsExtra>;
```
vkucera, line 498: Include what you use!

PR #15820, opened 2026-04-16
[PWGEM] Track matching checks for phosElId.cxx

yhambard:
Added several cases, to see how the track matching changes -  for TPC identified electrons
for True electrons (from mc)
for good dispersed clusters
for E/p ratio being around 1.05
and for all the combinations of the listed above

### DIFF
--- a/PWGEM/Tasks/phosElId.cxx
+++ b/PWGEM/Tasks/phosElId.cxx
@@ -16,8 +16,8 @@
 /// \author Yeghishe Hambardzumyan, MIPT
 /// \since Apr, 2024
 
-#include "Common/CCDB/TriggerAliases.h"
-#include "Common/Core/RecoDecay.h"
+#include "Common/Core/TrackSelection.h"
+#include "Common/Core/TrackSelectionDefaults.h"
 #include "Common/Core/trackUtilities.h"
 #include "Common/DataModel/CaloClusters.h"
 #include "Common/DataModel/Centrality.h"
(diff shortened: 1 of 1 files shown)


yhambard:
@alibuild @feisenhu @dsekihat @jokonig @hscheid, 
Could I, please, get a review to merge?
The macOS build fails on the protobuf package, it doesn't get to check my code. It isn't a required check.
Thanks in advance!

PWGEM/Tasks/phosElId.cxx
```diff
@@ -28,33 +28,27 @@
...
+#include "Framework/AnalysisDataModel.h"
+#include "Framework/AnalysisTask.h"
+#include "Framework/ConfigParamSpec.h"
+#include "Framework/HistogramRegistry.h"
+#include "Framework/runDataProcessing.h"
+#include "PHOSBase/Geometry.h"
+#include "ReconstructionDataFormats/TrackParametrization.h"
+
+#include "TF1.h"
+#include "TPDGCode.h"
```
mhemmer-cern: Could you please not revert the includes that were fixed recently.
Any include not from inside O2Physics should use `<>` instead of `""`

yhambard: Okay, I'll get on doing that within an hour! 👍

yhambard: Thanks for noticing! I had my fork behind. I resolved it by just reverting the includes back to the recently fixed ones

dsekihat (approved):

PR #15826, opened 2026-04-17
[ALICE3] Introduce flat LUT container for the TrackSmearer and rebase OTF code to use it

aalkin:
@njacazio here is the final working version (clean - without test tasks) that can be integrated. It still needs [#15275](https://github.com/AliceO2Group/AliceO2/pull/15275) in O2 to work - we should merge that on Monday.

@vkucera please disregard the linter errors here, as it rearranges a very old code. It can be brought to compliance with subsequent changes instead of trying to fix everything immediately.

### DIFF
--- a/ALICE3/Core/FlatTrackSmearer.cxx
+++ b/ALICE3/Core/FlatTrackSmearer.cxx
@@ -304,7 +304,7 @@ bool TrackSmearer::smearTrack(O2Track& o2track, const lutEntry_t* lutEntry, floa
     if (mInterpolateEfficiency) {
       eff = interpolatedEff;
     }
-    if (gRandom->Uniform() > eff) {//FIXME: use a fixed RNG instead of whatever ROOT has as a default
+    if (gRandom->Uniform() > eff) { // FIXME: use a fixed RNG instead of whatever ROOT has as a default
       isReconstructed = false;
     }
   }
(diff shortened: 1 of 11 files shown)


aalkin:
The required CI tests are done, legacy macOS issue is unrelated to this code and the linter issues, while being legitimate for the most part, should be addressed separately.

ALICE3/Core/FlatTrackSmearer.cxx
```diff
@@ -0,0 +1,415 @@
...
+    LOGF(info, "LUT table for PDG %d already loaded (index %d)", pdg, ipdg);
+    return false;
+  }
+
+  LOGF(info, "Loading %s LUT file: '%s'", getParticleName(pdg), filename);
+  const std::string localFilename = o2::fastsim::GeometryEntry::accessFile(filename, "./.ALICE3/LUTs/", mCcdbManager, 10);
+
+  std::ifstream lutFile(localFilename, std::ifstream::binary);
+  if (!lutFile.is_open()) {
+    throw framework::runtime_error_f("Cannot open LUT file: %s", localFilename);
```
aalkin:
```suggestion
    throw framework::runtime_error_f("Cannot open LUT file: %s", localFilename.c_str());
```

njacazio (approved):

PR #15835, opened 2026-04-17
[Common] Minor fix to the FwdTrackExtensionTask

lucamicheletti93:
Minor fix to reduce the memory consumption of the task


Common/TableProducer/fwdtrackextension.cxx
```diff
@@ -82,34 +82,34 @@ struct FwdTrackExtension {
...
+          if (refitGlobalMuon && (trackType == o2::aod::fwdtrack::ForwardTrackTypeEnum::GlobalMuonTrack || trackType == o2::aod::fwdtrack::ForwardTrackTypeEnum::GlobalForwardTrack)) {
+            auto muontrack = track.template matchMCHTrack_as<MuonsWithCov>();
+            auto mfttrack = track.template matchMFTTrack_as<aod::MFTTracks>();
+            o2::dataformats::GlobalFwdTrack propmuon = o2::aod::fwdtrackutils::propagateMuon(muontrack, muontrack, collision, o2::aod::fwdtrackutils::propagationPoint::kToVertex, 0.f, zField);
+            SMatrix5 tpars(mfttrack.x(), mfttrack.y(), mfttrack.phi(), mfttrack.tgl(), mfttrack.signed1Pt());
+            SMatrix55 tcovs{};
+            o2::track::TrackParCovFwd mft{mfttrack.z(), tpars, tcovs, mfttrack.chi2()};
+            fwdtrack = o2::aod::fwdtrackutils::refitGlobalMuonCov(propmuon, mft);
+          }
+          if (trackType == o2::aod::fwdtrack::ForwardTrackTypeEnum::MuonStandaloneTrack || trackType == o2::aod::fwdtrack::ForwardTrackTypeEnum::MCHStandaloneTrack) {
```
aferrero2707: I have a doubt regarding the logic here... how do you compute the DCA in the case of `trackType == GlobalMuonTrack || trackType == GlobalForwardTrack`?

Thanks!

lucamicheletti93: @aferrero2707  , I move back to a minimal fix configuration. Indeed with this change we reduce by half the memory consumption for this task (~450MB --> ~200MB), but I think we can go down to ~40MB with other modifications. I will work on that at the beginning of the week, for the moment I fix the issue with the geometry to avoid further issues.

dsekihat (approved):

PR #15850, opened 2026-04-17

ariedel-cern:
Refactor resonance handling in the femto framework.
Resonances are now constructed by combining tracks from the femto tables rather than directly from raw data. This allows them to be produced later and accessed via linked derived data.
Additionally, an explicit mask for the resonance is no longer required. Instead, the track mask is inherited from the daughter tracks, with an additional boolean flag to control whether TOF information is used.


alibuild (approved):

PR #15860, opened 2026-04-18
[PWGHF] HFInvMassFitter enhancement

lubynets:
1. **Added a possibility of slight change of initial fit parameters (total integral of signal and background and background's function parameters) in order to avoid fit's convergence to a local minimum instead of the global one.**
_Detailed explanation_: it was observed that the raw yield as a function of BDT NP score lower-value cut demonstrates non-smooth behavior (see some outlier points in left panel as well as points with error bars larger than the neighbor points have). It was assumed to happen due to possible fit's convergence to a local minimum instead of the global one exactly at this NP score cut. In order to avoid this, the initial fit parameters are now randomized around their original values, using the `TRandom3` class seeded with `RandomSeed` configurable. It turns out that for 100 trials most of raw yield values clusterize at a certain value (which is belived to be a "true" value), wile some number of outliers is present (middle panel). The median out of 100 trials value of each variable of interest (raw yield, mean, sigma, or statistical error on them) can be extracted, then the raw yield vs NP score cut distribution does not have outliers (right panel). This is critically important for stability of the cut variation method implemented in [cut_variation.py](https://github.com/AliceO2Group/O2Physics/blob/c70b32253e703fd9deef6e2ab27800c2eef88a95/PWGHF/D2H/Macros/cut_variation.py).
Screenshot from 2026-04-18 23-22-00

2. **Added double-sided Crystal Ball function to possible signal functions.**
3. **Configuration via json made more user- and reading-friendly:**
vector of values (of the size of number of slice intervals) can be replaced with a single number when all vector components are equal;
check for correct parsing the config added (to replace the json package error message with the custom one);
histograms which are optional in some mode (e.g. MC) are not requested to be present in the config.

4. **Bugs and code imperfections cleaned**


lubynets:
All checks except for O2Physics build are passed. Since this PR does not affect files compiled within O2Physics, the pending test do not make sense for it, therefore the PR is converted to draft to prevent the server from unnecessary job.

vkucera:
> All checks except for O2Physics build are passed. Since this PR does not affect files compiled within O2Physics, the pending test do not make sense for it, therefore the PR is converted to draft to prevent the server from unnecessary job.

Hi @lubynets , do you mean that you request a review + approval before marking the PR as ready for review because you want to avoid starting the build tests unnecessarily?

lubynets:
> > All checks except for O2Physics build are passed. Since this PR does not affect files compiled within O2Physics, the pending test do not make sense for it, therefore the PR is converted to draft to prevent the server from unnecessary job.
> 
> Hi @lubynets , do you mean that you request a review + approval before marking the PR as ready for review because you want to avoid starting the build tests unnecessarily?

Yes, effectively my PR is ready for review. I marked it as `Draft` in order not to occupy the resources, which can be used more efficiently for those PRs which modify the compiled part of the O2Physics.

vkucera:
> > > All checks except for O2Physics build are passed. Since this PR does not affect files compiled within O2Physics, the pending test do not make sense for it, therefore the PR is converted to draft to prevent the server from unnecessary job.
> > 
> > 
> > Hi @lubynets , do you mean that you request a review + approval before marking the PR as ready for review because you want to avoid starting the build tests unnecessarily?
> 
> Yes, effectively my PR is ready for review. I marked it as `Draft` in order not to occupy the resources, which can be used more efficiently for those PRs which modify the compiled part of the O2Physics.

OK, thanks for clarification.

@stefanopolitano @zhangbiao-phy Please have a look.

stefanopolitano (approved):
Hi @lubynets ! Thanks a lot for you development. Good for me

zhangbiao-phy:
Good to me as well! Thanks!

PR #15861, opened 2026-04-18
[PWGLF] Fix file sink crash

romainschotter:
- fix crash at the file sink level
- group event, V0, cascade histograms into different folders

@ChiaraDeMartin95 for your information


alibuild (approved):

PR #15867, opened 2026-04-20
[PWGLF] added primary convolution

jtanwar2212:


vkucera:
@jtanwar2212 
- Don't add the prefix if you are not sure about the correct form. It will be added automatically.
- What is convulation?

jtanwar2212:
Hi @vkucera, thank you for seeing the request, we added primary particle histogram which help us to estimate the primary fraction of Helium in our data.

vkucera:
> Hi @vkucera, thank you for seeing the request, we added primary particle histogram which help us to estimate the primary fraction of Helium in our data.

I don't think it is clear from the PR title. Did you mean "convolution"?

maciacco (approved):

PR #15881, opened 2026-04-20
[PWGCF] adding an additional Mult histo (counting events with >1 particle candidate)

glromane:


alibuild (approved):

PR #15886, opened 2026-04-20
[PWGLF] Update heavyionMultiplicity.cxx

murioni:
adding Z vtx sel in processBC, configurable to use multiplicity instead of centrality in MC

### DIFF
--- a/PWGLF/Tasks/GlobalEventProperties/heavyionMultiplicity.cxx
+++ b/PWGLF/Tasks/GlobalEventProperties/heavyionMultiplicity.cxx
@@ -589,11 +589,11 @@ struct HeavyionMultiplicity {
         continue;
       }
       auto centrality = 0;
-      if(isApplyGenMult)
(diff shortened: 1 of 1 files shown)


PWGLF/Tasks/GlobalEventProperties/heavyionMultiplicity.cxx
```diff
@@ -586,17 +588,23 @@ struct HeavyionMultiplicity {
       if (RecCol.globalIndex() != mcCollision.bestCollisionIndex()) {
         continue;
       }
+      auto centrality = 0;
```
abmodak, line 591: Hello @murioni, can you please initialize the `centrality` value to some large negative value e.g. `-999`, just to avoid the unnecessary filling of zero entries (in case you do not have centrality value) in the histogram?

Thanks,
Abhi

abmodak (dismissed):

abmodak (approved):

PR #15910, opened 2026-04-22

ariedel-cern:
This PR includes 
- fixing slowdowns in the event mixing due to execcisve Qa
- adding pass through mode in the producer for collisions, tracks and v0s (more are following)
- add requirement on filling event histograms in the Qa tasks to only fill when a particle of interest is present in the collision


alibuild (approved):

PR #15914, opened 2026-04-22
[PWGJE] Updating jetD0AngSubstructure.cxx

pdhankhe:
- Fixed the formatting issues of columns
- Added task to match detector level and particle level Monte Carlo particles.
- Added an additional task to calculate the Monte Carlo efficiency.
- Using templated analyse functions to easily define process functions for other particles than D0 if desired.


PWGJE/Tasks/jetD0AngSubstructure.cxx
```diff
@@ -8,102 +8,320 @@
...
 #include <Framework/Configurable.h>
+#include <Framework/DataProcessorSpec.h>
 #include <Framework/HistogramRegistry.h>
 #include <Framework/HistogramSpec.h>
 #include <Framework/InitContext.h>
 #include <Framework/runDataProcessing.h>
-
+//
+#include <PWGHF/Core/DecayChannels.h>
+//
```
nzardosh, line 30: please rebase to the latest master to avoid changing includes if not needed

PWGJE/Tasks/jetD0AngSubstructure.cxx
```diff
@@ -8,102 +8,320 @@
...
+mcd: Monte Carlo data (detector level)
+mcp: Monte Carlo data (particle level)
+*/
+namespace histnames
+{
+#define HNAME(Name) constexpr const char* Name = #Name;
+/*
+// Experimental Data (analyseDataChargedSubstructure)
+*/
+HNAME(ex_col);           // Collision Counter
```
nzardosh, line 218: why do you define these here? What is the operational benefit?

PWGJE/Tasks/jetD0AngSubstructure.cxx
```diff
@@ -170,32 +495,388 @@ struct JetD0AngSubstructure {
...
+                                   MCPJetsPerMCCollissionPreslice const& jetmcppreslice,
+                                   aod::JetMcCollisions const& mccollisions,
+                                   aod::JetCollisionsMCD const& collisions,
+                                   JetTableMCD const& mcdjets,
+                                   JetTableMCP const& mcpjets,
+                                   CandidatesMCD const& /*mcdCandidates*/,
+                                   CandidatesMCP const& /*mcpCandidates*/,
+                                   aod::JetTracks const& tracks,
+                                   aod::JetParticles const& particles,
+                                   HistogramRegistry& registry)
```
nzardosh, line 551: the registry should be a member of your struct and accesible everywhere

vkucera (changes requested):
Don't break the includes.

PWGJE/Tasks/jetD0AngSubstructure.cxx
Lolle2001: There are two main reasons I prefer to define the histogram name this way:
- using HNAME allows me to only insert the name of a histogram once. If I were to make a change it will be consistent and change both the variable name and string definition.
- the code is more readable to me and it is easy to add a new definition.

PWGJE/Tasks/jetD0AngSubstructure.cxx
Lolle2001: The registry _is_ a member of the struct. I have `process`  methods at the bottom where I pass it as an argument to the templated `analyse` methods. 
- Currently I use only one registry, but I might want to add one later in my planned analysis. 
- I prefer to make the methods independent of which particle I analyse (e.g. the tables and registry).

I am not aware of all requirements so if this is something that is not recommended or impractical according to your experience. I am happy to change it. Otherwise I rather leave it this way.

Lolle2001:
> Don't break the includes.

I do not see how the includes are broken. Could you perhaps elaborate on this or send a link to the relevant part in the documentation?

vkucera:
> > Don't break the includes.
> 
> I do not see how the includes are broken. Could you perhaps elaborate on this or send a link to the relevant part in the documentation?

https://aliceo2group.github.io/analysis-framework/docs/tools/#cleaning-include-statements-and-using-statements

PWGJE/Tasks/jetD0AngSubstructure.cxx
vkucera: Your macro just defines strings with the same name as the content. If you want to rename a histogram, you have to change all occurrences of that string anyway. It just creates another level of indirection, making the references actually longer than their content. How does that make anything easier or more readable?

PWGJE/Tasks/jetD0AngSubstructure.cxx
vkucera: What would you try to achieve by having multiple registries?

PWGJE/Tasks/jetD0AngSubstructure.cxx
Lolle2001: It is a preference, it prevents me making typos in the histogram names. I added this, because before I had string and found myself making mistakes when writing the names as a string. The intellisense I use recognises variable names, but not bare strings. This makes it a lot easier for me to change names and prevent typos. 

So indeed, if I would change the variable name, I have to do that elsewhere in the code (ctrl+f makes this easy in combination with the namespace), but that is not the point I intended to make. You do not seem to tell me that it is not allowed, so I rather keep it this way since it fits my style of working.

PWGJE/Tasks/jetD0AngSubstructure.cxx
Lolle2001: When I intend to add a `process` method for other particle species. However, I realise that this would require a different pipeline of tasks anyway and therefore seperate outputs. Therefore I can just as well use one registry since it will never happen that in the same pipelen the task is used for both particle species.

PWGJE/Tasks/jetD0AngSubstructure.cxx
vkucera: Well, I just gave you several reasons why I think it makes the code worse in both aspects that you seem to care about.
Can you please explain how this setup allows you to rename a histogram without changing all the string occurrences?
The point of preventing typos seems irrelevant to me because you are supposed to test your changes and an attempt to fill a non-existent histogram should throw a run time error.

PWGJE/Tasks/jetD0AngSubstructure.cxx
vkucera: Even if you did have processing of multiple species in different `process` functions of the same task, I don't see how multiple registries would help you do anything that cannot be done with one.

PR #15929, opened 2026-04-23
[Common] Extend option to reassociate PV contributors to different collisions

fgrosa:
This PR extends the possibility to reassociate tracks that are PV contributors if they belong to low mult vertices, without changing the default behaviour. @ddobrigk can you please approve it? Thanks!

### DIFF
--- a/Common/Core/CollisionAssociation.h
+++ b/Common/Core/CollisionAssociation.h
@@ -46,6 +46,12 @@ enum TrackSelection {
   QualityTracksITS = 2
 };
 
+enum PVContrReassocOpt {
+  Disabled = 0,
+  OnlySameBc = 1,
+  SameBcAndLowMult = 2
+};
+
 } // namespace track_association
 } // namespace o2::aod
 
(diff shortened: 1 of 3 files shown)


ddobrigk (approved):

dsekihat:
Ciao @fgrosa , Thank you for this PR. Just for my curiosity, does this PR somehow solve split vertices in high-multiplicity pp collisions?

fgrosa:
> Ciao @fgrosa , Thank you for this PR. Just for my curiosity, does this PR somehow solve split vertices in high-multiplicity pp collisions?

Hi @dsekihat we cannot solve split vertices at the analysis level, since the primary vertex reconstruction is done at the reconstruction time. 
This PR allows to have more relaxed re-association criteria for tracks that are PV contributors of low multiplicity vertices. In fact, low multiplicity PVs produced by HF decays might have poor time resolution, and drift to BCs that are different than the "real" PV. In the current version we might not be able to reassociate the tracks originated from the HF decay to the right vertex because of the PVcontributor constraint, while this PR allows to relax that constrain for low multiplicity PVs.

PR #15931, opened 2026-04-23
[Trigger,PWGHF] Small fix to QA output

fgrosa:
@zhangbiao-phy @apalasciano


alibuild (approved):

PR #15937, opened 2026-04-23
[PWGDQ] Change some code for the energy correlator study

YazhenLin:
By default, the TH2::Interpolate(x, y) method uses bilinear interpolation. This algorithm calculates the weighted average based on the four nearest bin centers.
When the input coordinates (x,y) fall outside the range of the first or last bin centers, the function attempts to interpolate between the edge bin and the Underflow/Overflow bins. Since these bins are often empty (value = 0), the resulting weight drops sharply toward zero, which is physically incorrect for efficiency/acceptance corrections.
To prevent these, I keep the input coordinates to the safe range defined by the centers of the first and last bins. It does not alter the interpolated values for points within the internal range, it simply ensures that points beyond the boundary bin centers returns  to the edge values.


mcoquet642:
Please add some description to explain what you are changing in your PRs

YazhenLin:
> Please add some description to explain what you are changing in your PRs

Hi @mcoquet642, I add some description now. Thank you!

mcoquet642 (approved):

PR #15943, opened 2026-04-23
[PWGLF] deuteron proton correlaton: fix gen mixed event

ercolessi:


maciacco (approved):
@ercolessi check the missing const ref at l. 685 in future iterations

PR #15944, opened 2026-04-23
[PWGDQ] Adding process function for muon only streaming + q-vectors

lucamicheletti93:
New process function to run the muon streaming in Pb-Pb collisions. This process do not use the FwdTrackExtension in order to minimize the memory consumption.


alibuild (dismissed):

alibuild (approved):

PR #15948, opened 2026-04-23
[PWGDQ] add tag and probe method for dalitz selection

glegras:


PWGDQ/Core/CutsLibrary.cxx
```diff
@@ -5064,6 +5095,42 @@ AnalysisCut* o2::aod::dqcuts::GetAnalysisCut(const char* cutName)
     return cut;
   }
 
+  TF1* fDCAxyresLow = new TF1("fDCAxyresLow", "[0] + [1] * pow(x, -[2])", 0.1, 1000.);
+  TF1* fDCAzresLow = new TF1("fDCAzresLow", "[0] + [1] * pow(x, -[2])", 0.1, 1000.);
+  TF1* fDCAxyresUp = new TF1("fDCAxyresUp", "[0] + [1] * pow(x, -[2])", 0.1, 1000.);
+  TF1* fDCAzresUp = new TF1("fDCAzresUp", "[0] + [1] * pow(x, -[2])", 0.1, 1000.);
```
vkucera: How do these objects get deleted?

glegras: Since they are used in the AnalysisCut, they cannot be deleted until the end. But now I modified so that these TF1 are defined only when necessary

vkucera: You didn't answer my question.

glegras: I am not sure how they are deleted, maybe @iarsene you could check the PR and comment on that? Also I am using the standard coding of the PWGDQ framework, so if something is wrong about object deletions it should be modified in the entire framework.

vkucera: I'm asking because I don't see a single `delete` statement in the `AnalysisCut` class which would indicate that all the dynamically allocated objects with raw pointers leak, unless they are owned and handled by other classes.

iarsene: Hi Vit, in principle you are right. However, this function is allocated at init time, and then we need it until the end of the job. So this will not "leak" memory during the time the job is running. We can probably make the AnalysisCut to own this TF1 and then have an explicit delete when the cut gets destroyed at the end of the job.

vkucera: Hi Ionut, I don't know what you mean by "job" here, but I would expect the leak to live from the first call of `~AnalysisCut` till the end of the program. Either the destructor should handle the deallocation or the raw pointers should be replaced with smart pointers.

glegras: I changed from TF1* to shared_ptr<TF1> now

iarsene (approved):

PWGDQ/Core/AnalysisCut.h
```diff
@@ -49,29 +50,29 @@ class AnalysisCut : public TNamed
...
-    short fDepVar2;    // second (optional) variable on which the cut depends
+    int16_t fDepVar2;  // second (optional) variable on which the cut depends
     float fDep2Low;    // lower limit for the second dependent var
     float fDep2High;   // upper limit for the second dependent var
     bool fDep2Exclude; // if true, then use the dependent variable range as exclusion
 
-    TF1* fFuncLow;  // function for the lower limit cut
-    TF1* fFuncHigh; // function for the upper limit cut
+    std::shared_ptr<TF1> fFuncLow;  // function for the lower limit cut
+    std::shared_ptr<TF1> fFuncHigh; // function for the upper limit cut
```
vkucera, line 69: These are all uninitialised.

PR #15957, opened 2026-04-24
[PWGLF] Inclusion of conditions to evaluate systematic uncertainties

lucasjsilva:
The inclusion concern pT extrapolation to zero, phi cut variation and the evaluation of particles coming from weak decays


PWGLF/Tasks/GlobalEventProperties/studyPnch.cxx
vkucera, line 1: @lucasjsilva You are supposed to test your changes with the latest O2Physics and know how to rebase. The fact that your PR does not compile and reverts fixes of includes indicates that was not the case.

vkucera (changes requested):

PR #15991, opened 2026-04-28
[PWGJE] Added functionality to calculate efficiency for non-hooked MC

smaff92:


PWGJE/Tasks/statPromptPhoton.cxx
```diff
@@ -1513,6 +1520,15 @@ struct statPromptPhoton {
           continue;
         if (std::fabs(mcParticle.getGenStatusCode()) >= 81 || !mcParticle.isPhysicalPrimary())
           continue;
+        if (cfgReqRecPS_GEN) {
+          if (std::fabs(mcParticle.eta()) > 0.62)
+            continue;
+          bool insideCalPhi = false;
+          if ((mcParticle.phi() > 1.42 && mcParticle.phi() < 3.26) || (mcParticle.phi() > 4.56 && mcParticle.phi() < 5.70))
```
nzardosh: would be good if you could make these hard coded values configurable

smaff92: Hi Nima,

Done! I also changed the previous cfg to int, so the build should clear now.

Cheers!

nzardosh (approved):

PR #15993, opened 2026-04-28
[PWGCF] FemtoUniverse: Add option to randomize pair for DeltaEta-DeltaPhi* histograms in SH task

alicja-pp:
`PWGCF/FemtoUniverse/Tasks/femtoUniversePairTaskTrackTrackSpherHarMultKtExtended.cxx`:
 - Add option to randomize pair for DeltaEta-DeltaPhi* histograms in SH task also for plus-minus pairs.
- This was default for same sign pairs. For plus-minus pairs it was implemented only for mixed events.
- Now it is still applied for same sign pairs, and for plus minus it can be turned on with _confRandomizeCPR_ configurable in both same and mixed event functions.


PWGCF/FemtoUniverse/Tasks/femtoUniversePairTaskTrackTrackSpherHarMultKtExtended.cxx
```diff
@@ -622,18 +623,31 @@ struct FemtoUniversePairTaskTrackTrackSpherHarMultKtExtended {
...
         }
 
+        auto part1 = p1;
+        auto part2 = p2;
+
+        if (twotracksconfigs.confRandomizeCPR) {
+          double rand;
+          rand = randgen->Rndm();
+
+          if (rand > 0.5) {
```
prchakra, line 633: May be use <twotracksconfigs.confRandValue> instead of 0.5? It'll resolve the o2-linter issue.

prchakra (changes requested):

PWGCF/FemtoUniverse/Tasks/femtoUniversePairTaskTrackTrackSpherHarMultKtExtended.cxx
alicja-pp: That's right, I already commited the change.

prchakra (approved):

PR #15998, opened 2026-04-28
[PWGLF] Inclusion of conditions to evaluate systematic uncertainties

lucasjsilva:
The inclusion considers pT extrapolation to zero, phi cut variation, and the evaluation of particles coming from weak decays


vkucera:
@lucasjsilva Why are you duplicating https://github.com/AliceO2Group/O2Physics/pull/15957 ?

abmodak (approved):

PR #16001, opened 2026-04-28
[PWGUD] fix efficiency and cuts

miedema-11:
sorry, there is something wrong when I make PR, I rebased the local repo and make a new one to solve conflicts


amatyja (dismissed):

vkucera:
@miedema-11 Can you please explain your workflow? I don't understand how you can create a  PR which doesn't compile on any system.

miedema-11:
I did test it locally, but something went wrong when I pushed it. But I have found the problem, and it is ready to merge now

amatyja (approved):

PR #16029, opened 2026-04-29
[PWGCF] updated code for mixing hadrons with mixed phi from diff events

prottayCMT:


prottayCMT:
@jaelpark could you please approve

PWGCF/Tasks/correlations.cxx
```diff
@@ -654,6 +661,29 @@ struct CorrelationTask {
             continue;
           }
         }
+
+        if (rejectMixedPhiProngEventsNow && cfgRejectMixedPhiProngEvents) {
```
jaelpark: Why is `rejectMixedPhiProngEventsNow` needed? Why not just check if `target` is mixed and `doprocessMixed2ProngDerivedMixedPhi` true?

jgrosseo: I agree that if possible we should get rid of that variable, or if really needed make it a template variable and not a member

PR #16031, opened 2026-04-29
[PWGJE] Increase Ds sparse hist binning

monamelop:


monamelop:
Hi, the only failing check is build/O2Physics/o2/macOS. From the log, it looks like a protobuf/abseil linking issue on macOS, before my O2Physics task is built.

This PR only changes the number of bins of one histogram axis, so I believe the failure is unrelated to my change. Could someone please confirm whether this can be ignored or re-run?

nzardosh (approved):

PR #16044, opened 2026-04-30
[PWGJE] Modification 3-prong to n-prong for b-jet

hanseopark:
DataModel?JetTagging.h
- To add N-Prong table

TableProducer/secondaryVertexReconstrcution.cxx
- To add NProng producer

TableProducer/jetTaggerHF.cxx

- To change 3Prong to N-Prong
 
Tasks/bjetCentMult.cxx
- To change 3Prong to N-Prong and modify the histogram name

Tasks/jetTaggerHFQA.cxx
- To remove the 2-prong task and change 3prong to n-prong

### DIFF
--- a/PWGJE/DataModel/JetTagging.h
+++ b/PWGJE/DataModel/JetTagging.h
@@ -49,9 +49,9 @@ using JTrackTag = JTracksTag::iterator;
 
 namespace secondary_vertex_params
 {
-DECLARE_SOA_COLUMN(XPrimaryVertex, xPVertex, float); // o2-linter: disable=name/o2-column
-DECLARE_SOA_COLUMN(YPrimaryVertex, yPVertex, float); // o2-linter: disable=name/o2-column
-DECLARE_SOA_COLUMN(ZPrimaryVertex, zPVertex, float); // o2-linter: disable=name/o2-column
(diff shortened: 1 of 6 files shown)


hanseopark:
@hahassan7 Hello Hadi,
Please could you check the code?

hahassan7:
Hi @hanseopark ,
Thanks for making these changes, the code looks good to me. Although there are some limitations with your method, it only allows for either 2 or 3 prong SV reconstruction. If you want both, you can't use this method, instead you will have to activate the 2prong and 3prong functions. But for tagging the old limitation is still there, you can only tag either 2 or 3 prong SV you can't use both. But this is not a big problem. I think the code can approved the way it is.

hanseopark:
> Hi @hanseopark , Thanks for making these changes, the code looks good to me. Although there are some limitations with your method, it only allows for either 2 or 3 prong SV reconstruction. If you want both, you can't use this method, instead you will have to activate the 2prong and 3prong functions. But for tagging the old limitation is still there, you can only tag either 2 or 3 prong SV you can't use both. But this is not a big problem. I think the code can approved the way it is.

This is the issue I was worried about.
In my opinion, performing a side-by-side comparison using tags will ultimately be difficult.

nzardosh (approved):

PR #16046, opened 2026-04-30
[PWGLF] Added code for model prediction for Kstar, Phi resonance

sawankumawat:


sawankumawat:
- Derived from mcParticlePrediction.cxx
- Fixes limitations of filling only the final-state particles and long-lived V0S
- Applies physical primary selection on daughters (not the mother) for short-lived resonance decays
- Added PDG selection on individual daughters.

@ercolessi @romainschotter could you please review and approve?

Thank you

PR #16072, opened 2026-05-03
[PWGCF] added RCT event selection with Qa plots

Thorkj:


PWGCF/TwoParticleCorrelations/Tasks/corrFit.cxx
```diff
@@ -408,6 +378,20 @@ struct CorrFit {
 
     LOGF(info, "End of init");
   }
+  template <typename TCollision>
+  bool eventRct(TCollision collision, const bool fillCounter)
```
vkucera: Don't copy collisions.

PR #16083, opened 2026-05-04
[PWGUD] add true gap service and TPC-FT0 correlations

Luzhiyongg:


rolavick (commented):
Why do you need the UpcService? What is the usage of it?

Luzhiyongg:
> Why do you need the UpcService? What is the usage of it?

Hi roman,
For event-mixing, it's required to put the major event selection in the filter, otherwise the event-mixing pairs will include many unused pairs and significantly reduce the statistics.
We found this issue in pp and it's expected that the case will be more significant in UPC.

As the major event selection here are gapside and truegapside, I create a service task to produce truegapside so that I can use truegapside in the filter.
UpcService.h is the header defining truegapside output, and it's included by the upcTrueGapService.cxx and flowLongRangeCorrUpc.cxx

PWGUD/DataModel/UDTruegapsideTables.h
rolavick, line 1: This is not really a service, it is just a table definition. Put this header to PWGUD/DataModel and change the name appropriatelly, that is xxxxTables.h. Also, be specific in \brief

PWGUD/TableProducer/trueGapProducer.cxx
rolavick, line 1: Also, this only produces one desired table. Put the file to PWGUD/TableProducer. Change the appropriately, that is xxxxProducer.cxx

rolavick (changes requested):

PWGUD/TableProducer/trueGapProducer.cxx
Luzhiyongg: got it. I will do that

rolavick (approved):

PWGUD/DataModel/UDTruegapsideTables.h
```diff
@@ -0,0 +1,41 @@
...
+#include <Framework/AnalysisDataModel.h>
+#include <Framework/AnalysisHelpers.h>
+#include <Framework/AnalysisTask.h>
+#include <Framework/BinningPolicy.h>
+#include <Framework/Configurable.h>
+#include <Framework/GroupedCombinations.h>
+#include <Framework/HistogramRegistry.h>
+#include <Framework/HistogramSpec.h>
+#include <Framework/InitContext.h>
+#include <Framework/runDataProcessing.h>
```
vkucera, line 30: Don't include unused headers.

PWGUD/TableProducer/trueGapProducer.cxx
```diff
@@ -0,0 +1,73 @@
...
+#include "PWGUD/Core/SGSelector.h"
+#include "PWGUD/DataModel/UDTables.h"
+#include "PWGUD/DataModel/UDTruegapsideTables.h"
+
+#include <Framework/ASoA.h>
+#include <Framework/AnalysisDataModel.h>
+#include <Framework/AnalysisHelpers.h>
+#include <Framework/AnalysisTask.h>
+#include <Framework/BinningPolicy.h>
+#include <Framework/Configurable.h>
```
vkucera, line 26: Don't include unused headers.

vkucera:
@Luzhiyongg Why have you not fixed the includes yet?

PR #16097, opened 2026-05-04
[PWGCF] Fixes in femto framework

ariedel-cern:
Apply these two fixes:
- change index policy to striclyUpper in pair efficiency task
- make kstar/q3 compuation more robust


alibuild (approved):

PR #16100, opened 2026-05-04
[PWGEM] Remove unused file

vkucera:
@emeninno Is there a plan to use this?


emeninno:
@vkucera yes, I still would like to work on this, as soon as I have time. Thanks

vkucera:
> @vkucera yes, I still would like to work on this, as soon as I have time. Thanks

Hi @emeninno , thanks for your reply.

PR #16110, opened 2026-05-05
[PWGHF] Change default cutInvMassCascCharmNuclei threshold to 0 in the track skimming task

zhangbiao-phy:
Change default cutInvMassCascCharmNuclei threshold to 0, make more convenient for derived data creation, given it's no need in most of datasets


alibuild (approved):

PR #16114, opened 2026-05-05
[ALICE3,PWGHF] Cleanup legacy ALICE3

njacazio:


PWGHF/DataModel/CandidateReconstructionTables.h
```diff
@@ -23,7 +23,6 @@
 //
 #include "PWGLF/DataModel/LFStrangenessTables.h"
 
-#include "ALICE3/DataModel/ECAL.h"
```
vkucera, line 26: This currently provides `ECALs` for the `hf_cand_chic::Prong1` index column.

vkucera (changes requested):

PWGHF/DataModel/CandidateReconstructionTables.h
vkucera: However, the related table seems to be used only in the legacy ALICE3 code. I can deprecated it to unblock this PR.

njacazio: I removed the usage

vkucera: Won't help. You kept the table in.

njacazio: True :)

vkucera (approved):

alibuild (approved):

PR #16129, opened 2026-05-06
[PWGDQ] make qaMatching AO2D index merger-safe

nivram-phy:
## Summary
- make the `qaMatching` event-reference column use a merger-safe custom index label
- preserve canonical AO2D tree naming for the qaMatching output tables

## Problem
The previous `qaMatching` AO2D output produced an event-reference branch that was not resolved correctly by `o2-aod-merger` and caused downstream Hyperloop issues.


nivram-phy:
@aferrero2707 This is the new PR with the unbounded index issue resolved (hopefully). I chcked it locally with the o2-aod-merger

mcoquet642 (approved):

PR #16149, opened 2026-05-07
[PWGCF] Added RCTevent selection, and track counter configurable

Thorkj:


Thorkj:
The O2 build fails, but locally I have no issue. Is there anything to be done from my side?

wenyaCern:
> The O2 build fails, but locally I have no issue. Is there anything to be done from my side?

Hi @Thorkj , seems the error also happens in other PR in PWG-CF. I tried to check the problem, and it looks like a CI infrastructure issue: the build fails already at the CMake compiler check, before compiling O2Physics code. `recc` falls back to `localhost:8085` because `RECC_SERVER` is not set, then fails with “Connection refused”.

wenyaCern (approved):

PR #16164, opened 2026-05-08
[PWGLF] Added some configurables and changed cuts

SwatiSaha-1997:


PWGLF/Tasks/Resonances/lambda1520Analysispo.cxx
```diff
@@ -94,8 +94,12 @@ struct Lambda1520Analysispo {
   // ── Track quality cuts ───────────────────────────────────────────────────
   Configurable<float> minTrackPt{
     "minTrackPt", 0.15f, "Minimum transverse momentum pT of a track [GeV/c]"};
-  Configurable<float> minTrackMomentum{
-    "minTrackMomentum", 0.f, "Minimum total momentum p of a track [GeV/c]"};
+  Configurable<float> minProtonMomentum{
+    "minProtonMomentum", 0.f,
+    "Minimum total momentum p for proton PID [GeV/c]"};
```
vkucera, line 99: Don't break the lines like this. The linter cannot parse it and it increases the number of lines for no reason.

vkucera (changes requested):

PR #16166, opened 2026-05-08
[PWGLF] fix O2 linter errors

bhattdurgesh921-sys:


vkucera:
@bhattdurgesh921-sys What is this PR about?

PR #16168, opened 2026-05-08
[PWGLF] fix O2-linter errors and rename filename

bhattdurgesh921-sys:


bhattdurgesh921-sys:
The O2-linter errors are not related to my code, since to fix one last error regarding filename, I have updated the CMakeList, and that brought the other errors now.

Thanks,
Durgesh

dmallick2 (approved):

PR #16200, opened 2026-05-11
[PWGCF] new deuteron PID cut

blacwovie:


vkucera:
new deuteron PID cut *where?*

shouqiye (approved):

vkucera:
@blacwovie @shouqiye The PR title is too vague. I suppose there is not just a single place in CF which can have a deuteron PID cut.

PR #16203, opened 2026-05-11
[PWGDQ] add efficiency weight method

zjxiongOvO:


PWGDQ/Tasks/tableReader_withAssoc.cxx
```diff
@@ -1853,6 +1861,10 @@ struct AnalysisSameEventPairing {
             VarManager::FillPairVn<TEventFillMap, TPairType>(t1, t2);
           }
 
+          if (fConfigOptions.useEfficiencyWeighting) {
+            VarManager::FillEfficiency();
+          }
```
iarsene, line 1866: I think that we don't need the special FillEfficiency function. If the efficiency object has been set, and if we have histograms which require the efficiency, then these should just be filled as any normal variable. In this case, these would be filled in the FillPair() function

iarsene (approved):

PR #16204, opened 2026-05-11
[ALICE3,PWGCF] replace -1 with 1 in getForTimeStamp

dsekihat:
[ALICE3,PWGCF] replace -1 with 1 in getForTimeStamp
Don't use -1 as time stamp to access CCDB.


jgrosseo:
Doesn't this fail all the tasks where the object is not valid on 01.01.1900 at 0:00:01?

PR #16206, opened 2026-05-11
[PWGLF] Add a configurable to turn on and off INELGt0 condition

SuJeong-Ji:
Add a configurable to be a switch of INELGt0 condition


PWGLF/Tasks/Resonances/chk892LI.cxx
vkucera, line 19: Remove

prottayCMT (approved):

PR #16208, opened 2026-05-11
[PWGLF] use reconstructed eventType in cascade analysis

nepeivodaRS:
The reconstructed event type is used for MC.
Several QA hists, and eventType labels are added.

### DIFF
--- a/PWGLF/TableProducer/Strangeness/cascqaanalysis.cxx
+++ b/PWGLF/TableProducer/Strangeness/cascqaanalysis.cxx
@@ -44,6 +44,8 @@
 #include <Framework/runDataProcessing.h>
 
 #include <TH1.h>
+#include <TH2.h>
+#include <TH3.h>
 #include <TPDGCode.h>
 #include <TRandom2.h>
 #include <TString.h>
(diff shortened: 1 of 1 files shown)


PWGLF/TableProducer/Strangeness/cascqaanalysis.cxx
```diff
@@ -194,6 +196,15 @@ struct Cascqaanalysis {
...
   }
 
+  template <typename TAxisType>
+  static void setEventTypeAxisLabels(TAxisType* axis)
+  {
+    const char* labels[] = {"INEL", "INEL>0", "INEL>1"};
+    for (int i = 0; i < static_cast<int>(sizeof(labels) / sizeof(labels[0])); ++i) {
+      axis->SetBinLabel(i + 1, labels[i]);
+    }
+  }
```
vkucera, line 219: - Why don't you use `std::array`?
- You have no guarantee that these labels match how you fill the histograms. Why don't you use `enum`?

nepeivodaRS: Hey,
There is already an `EvFlags` enum, but it is a bit mask:

```cpp
EvINEL    = 0x1
EvINELgt0 = 0x2
EvINELgt1 = 0x4
```

So I cannot use those values directly as histogram-bin indices.
Do you suggest adding a separate ordered enum for the histogram axis like this:

```cpp
enum EventType {
  kINEL = 0,
  kINELgt0,
  kINELgt1,
  kNEventTypes
};

static constexpr std::array<const char*, kNEventTypes> EventTypeLabels = {
  "INEL",
  "INEL>0",
  "INEL>1"
};
```
I haven't done this because of the possible future confusion between these two enums.

PWGLF/TableProducer/Strangeness/cascqaanalysis.cxx
```diff
@@ -74,6 +77,19 @@ struct Cascqaanalysis {
 
   HistogramRegistry registry{"registry"};
 
+  enum EventTypeBin {
+    kINEL = 0,
+    kINELgt0,
+    kINELgt1,
+    kNEventTypeBins
+  };
```
vkucera, line 85: Thanks, that looks way safer.
Please remove the `k` prefixes. They server no purpose, conflict with the naming conventions and appear only in ROOT code.

romainschotter (approved):

PR #16209, opened 2026-05-12
[ALICE3] Refactor OTF tracker

Marcellocosti:
First draft of a refactoring of the `onTheFlyTracker` task, moving the code dedicated to the V0 and cascades studies to separate functions with respect to the main process function, so that one can more clearly understand the various operations performed on the tracks.

Tagging other users of the task to inform and receive feedback: @njacazio @jesgum @lhusova


Marcellocosti:
Tested with @jesgum that the results are consistent after the refactoring, so marking as ready.

njacazio (approved):

PR #16216, opened 2026-05-12
[PWGHF] Added THn for QA of electrons from Z

sashingo:
- added THn for QA of electrons from Z
- reduced ConfigurableAxis


vkucera:
Why is this 72 commits long?

sashingo:
I don't know why.
I committed as usual. (add, commit, push)

vkucera:
> I don't know why. I committed as usual. (add, commit, push)

But you are not updating and developing the right way.

- You develop in the master branch. You shouldn't.
- You keep merging the upstream into your master branch instead of rebasing. Like this you just keep cumulating merge commits and every PR you make will have more and more commits.
- There is no reason why your PR should contain your commits since last July.

Btw, fix the errors and warnings before you make the PR ready for review.

PR #16218, opened 2026-05-12
[PWGLF] Bug Fix in hStrangeCorrelation.cxx

kcui1:


PWGLF/Tasks/Strangeness/hStrangeCorrelation.cxx
```diff
@@ -1218,6 +1218,9 @@ struct HStrangeCorrelation {
 
         static_for<0, 3>([&](auto i) {
           constexpr int Index = i.value;
+          if ((Index == IndexOmegaMinus || Index == IndexOmegaPlus) && assocCandidate.compatible(Index, trackSelection.dEdxCompatibility) && std::abs(assocCandidate.invMassNSigma(Index - 2)) < massWindowConfigurations.nSigmaNearXiMassCenter) {
```
lhusova, line 1221: shouldn't it be `assocCandidate.compatible(Index-2, trackSelection.dEdxCompatibility)` ?

kcui1: What I want to do is : if one particle is OmegaMinus or OmegaPlus and it passed the dE/dx selection of Omega but the invMass is so close to Xi, then continue. So `assocCandidate.compatible(Index, trackSelection.dEdxCompatibility)` means this particle passed dE/dx selection.

kcui1: But you're right. It would be more reasonable to change it to -2.

lhusova: Ok, I see, I will approve then.

lhusova (approved):

PR #16225, opened 2026-05-12
[PWGLF] Update on Event and Signal Loss

zhengqi-w:
- [General] Update with the Event Split calculation and the Event,Signal Loss histograms
- Change the Histogram definition and filling steps of Generated Events with at least one recon. event. (since only generated events which has been reconstructed can get the physical CentralityFT0C information)
- Change the filling scheme for reconstructed MC collisions (since one generated collision may be reconstructed more than once) calculating the Event Splitting part of Event Loss coefficient.


maciacco:
@zhengqi-w approving now, please check the o2 linter errors in the next iteration

maciacco (approved):

PR #16226, opened 2026-05-12
[Trigger,PWGHF] Add mass vs ct histogram also for Xic0

fgrosa:
@creetz16 @lubynets I added the mass vs ct histogram also for the Xic0, in case we will need it


alibuild (approved):

PR #16228, opened 2026-05-13
[PWGDQ] extend the input features in MLResponse class

aferrero2707:
The additional features are needed to match those currently used for training the ML models for the MFT-MCH matching.

### DIFF
--- a/PWGDQ/TableProducer/tableMakerMC_withAssoc.cxx
+++ b/PWGDQ/TableProducer/tableMakerMC_withAssoc.cxx
@@ -1041,12 +1041,6 @@ struct TableMakerMC {
       if (static_cast<int>(muon.trackType()) < 2) {
         auto muonID = muon.matchMCHTrackId();
         auto chi2 = muon.chi2MatchMCHMFT();
-        if (fConfigVariousOptions.fUseML.value) {
-          std::vector<float> output;
-          std::vector<float> inputML = matchingMlResponse.getInputFeaturesTest(muon);
-          matchingMlResponse.isSelectedMl(inputML, 0, output);
-          chi2 = output[0];
-        }
         if (mCandidates.find(muonID) == mCandidates.end()) {
           mCandidates[muonID] = {chi2, muon.globalIndex()};
         } else {
@@ -1078,7 +1072,7 @@ struct TableMakerMC {
           muonprop = VarManager::PropagateMuon(muontrack, collision, VarManager::kToMatching);
         }
         std::vector<float> output;
-        std::vector<float> inputML = matchingMlResponse.getInputFeaturesGlob(muon, muonprop, mftprop, collision);
+        std::vector<float> inputML = matchingMlResponse.getInputFeatures(muon, mfttrack, muontrack, mftprop, muonprop, collision);
         matchingMlResponse.isSelectedMl(inputML, 0, output);
         float score = output[0];
         if (mCandidates.find(muonID) == mCandidates.end()) {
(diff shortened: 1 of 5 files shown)


vkucera:
You are supposed to test your changes before you open a PR.

PWGDQ/TableProducer/tableMakerMC_withAssoc.cxx
```diff
@@ -1078,7 +1084,7 @@ struct TableMakerMC {
           muonprop = VarManager::PropagateMuon(muontrack, collision, VarManager::kToMatching);
         }
         std::vector<float> output;
-        std::vector<float> inputML = matchingMlResponse.getInputFeaturesGlob(muon, muonprop, mftprop, collision);
+        std::vector<float> inputML = matchingMlResponse.getInputFeatures(muon, mfttrack, muontrack, mftprop, muonprop, collision);
```
mcoquet642, line 1075: Please don't comment out this full part. Keep at least one InputFeature function, if its dummy, to leave in tableMaker/tableMakerMC

aferrero2707:
@mcoquet642 @lucamicheletti93 the additional features included in the MLResponse class require some dynamic tables that are currently not included in the tableMaker tasks. Since the ML matching is highly experimental and not yet used in any analysis, I propose to disable the corresponding code for the time being (see last commit in this PR) and work on a proper fix in a separate PR.

mcoquet642:
> @mcoquet642 @lucamicheletti93 the additional features included in the MLResponse class require some dynamic tables that are currently not included in the tableMaker tasks. Since the ML matching is highly experimental and not yet used in any analysis, I propose to disable the corresponding code for the time being (see last commit in this PR) and work on a proper fix in a separate PR.

As I said, I would prefer to keep a dummy input feature method in the response class, and use it in tableMaker, instead of disabling the code

aferrero2707:
> You are supposed to test your changes before you open a PR.

Sorry, I had to prepare this one without updating my local O2Physics, and I didn't realise the conflicts with the DQ table makers...

aferrero2707:
> As I said, I would prefer to keep a dummy input feature method in the response class, and use it in tableMaker, instead of disabling the code

@mcoquet642 I might have found a more elegant solution, see [this commit](https://github.com/AliceO2Group/O2Physics/pull/16228/commits/055f746c365d42805f538e2bc5d11179d448186a). It uses the `requires` concept to test the availability of the getter function at compile time, and if unavailable  sets the feature to zero.
This would allow to compile the tableMaker core without having to join more tables to the collisions and tracks.

What do you think?

mcoquet642:
> @mcoquet642 I might have found a more elegant solution, see [this commit](https://github.com/AliceO2Group/O2Physics/pull/16228/commits/055f746c365d42805f538e2bc5d11179d448186a). It uses the `requires` concept to test the availability of the getter function at compile time, and if unavailable sets the feature to zero. This would allow to compile the tableMaker core without having to join more tables to the collisions and tracks.
> 
> What do you think?

I am not familiar with this `requires` method, but if it works it would be excellent !

aferrero2707:
> I am not familiar with this `requires` method, but if it works it would be excellent !

I tested it locally and it definitely works. The `requires` expression simply checks the existence of the method at compile-time, and if the method is not available the compiler generates the code with the dummy assignment instead.

aferrero2707:
@mcoquet642 this PR is finally ready for review...

mcoquet642 (approved):

PR #16230, opened 2026-05-13
[PWGHF] update the default settings and correct the `getCandMassAndSign` in `taskPtFlucCharmHadrons.cxx`

wuctlby:
Hi @zhangbiao-phy and @stefanopolitano 
I made some updates on the `taskPtFlucCharmHadrons.cxx` in this PR, including

- Since the sigma of the mean pT of tracks is around 50 MeV, I optimized the default corresponding settings and modified the variable name of the pT selection on D
- Corrected the `getCandMassAndSign`, because the typename of `candidate` in the lopper is different.

Please let me know if you have any comments. Thanks a lot!
Also tagging @prottayCMT and @SwatiSaha-1997


vkucera:
Location missing in the PR title

wuctlby:
Hi @zhangbiao-phy and @stefanopolitano, it seems this macOS error always occurs and is not related to my implementation. So if there are no further comments, could you please approve this PR?

stefanopolitano (approved):

PR #16250, opened 2026-05-14
[PWGJE] Update analysis code based on ARC comments

ynishida-style:
Address ARC comments and improve Jet Shape task logic

- Merge MC processes into a single process.
- Move tracking efficiency histogram filling to after jet matching.
- Fix the incorrect position of the histogram registry.
- Unify nSigma condition for DCAxy with the one used for PID purity calculation.
- Update Perp Cone data extraction to use the same ring regions as the Jet In-cone extraction.


vkucera:
@ynishida-style Update *which* code?

nzardosh (approved):

PR #16255, opened 2026-05-14
[PWGUD] Add RCT flag checks in UPCCandidateProducer

ariffero:


rolavick:
Hi Andrea,
why do you need the RCT flags selection in this producer? You are about to use the flags with CBT prefix, that is a shortcut for central barrel tracking. So far, this producer was used for muon arm analyses only. Do you plan to use it for central barrel too? 
Roman

ariffero:
Hi Roman,
I implemented the RCT flag selection here because it was asked some time ago in the UPC PAG both to me and Niveditha. Note that the default option for the rct label is CBT_muon, that requires "FT0, ITS, MCH, MID with quality status Good or Limited Acceptance MC Reproducible, TPC  Good or Limited Acceptance MC Reproducible or BadPID" and is the standard for muon arm analyses without the MFT. Anyway, the rct lablel to use is a configurable so it can be used also for muons analyses with the MFT or in the central barrel.
Cheers,
Andrea

rolavick:
Hi Andrea,
but, is this candidate creator using ITS at all? Who actually asked you about RCT flags?
Roman

rolavick (approved):

PR #16257, opened 2026-05-15
[DPG] major_update_producer

rliotino99:
Major updates on the hmpid table producer task and associated change in CMakeLists file

### DIFF
--- a/DPG/Tasks/AOTTrack/PID/HMPID/hmpidTableProducer.cxx
+++ b/DPG/Tasks/AOTTrack/PID/HMPID/hmpidTableProducer.cxx
@@ -11,38 +11,24 @@
 
 #include "tableHMPID.h"
 
-#include "Common/Core/PID/PIDTOF.h"
-#include "Common/Core/trackUtilities.h"
 #include "Common/DataModel/Centrality.h"
 #include "Common/DataModel/EventSelection.h"
 #include "Common/DataModel/Multiplicity.h"
 #include "Common/DataModel/PIDResponseTOF.h"
 #include "Common/DataModel/PIDResponseTPC.h"
 #include "Common/DataModel/TrackSelectionTables.h"
-#include "Common/TableProducer/PID/pidTOFBase.h"
 
 #include "CCDB/BasicCCDBManager.h"
-#include "CCDB/CcdbApi.h"
-#include "CommonConstants/GeomConstants.h"
 #include "DataFormatsParameters/GRPMagField.h"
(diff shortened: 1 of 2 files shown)


DPG/Tasks/AOTTrack/PID/HMPID/hmpidTableProducer.cxx
```diff
@@ -11,25 +11,42 @@
...
-#include <CCDB/BasicCCDBManager.h>
+#include "CCDB/BasicCCDBManager.h"
+#include "CCDB/CcdbApi.h"
+#include "CommonConstants/GeomConstants.h"
+#include "DataFormatsParameters/GRPMagField.h"
+#include "DetectorsBase/GeometryManager.h"
+#include "DetectorsBase/MatLayerCylSet.h"
+#include "DetectorsBase/Propagator.h"
+#include "MathUtils/Utils.h"
+#include "ReconstructionDataFormats/TrackParametrizationWithError.h"
```
vkucera: Fix the format and make sure that you include what you use.

rliotino99: Ok Thank you. They were leftovers from an earlier development stage and are not referenced anywhere in the code — confirmed by a clean build without them.

vkucera: Thanks for the update. Please note the following:

- O2 headers are external to O2Physics and therefore should be included using `<>`.
- A successful compilation is not enough to verify that you include what you use. Please follow the [instructions](https://aliceo2group.github.io/analysis-framework/docs/tools/#cleaning-include-statements-and-using-statements).

rliotino99: Ok, I think I have implemented everything correctly, deleted unused includes, tested locally and fixed with the instructions given. 
This was my first time using clang-tidy, so please let me know if anything else needs to be addressed before merging. Thanks

vkucera: Thanks. Now you are including also O2Physics headers with `<>` which is wrong because they are local headers (i.e. from the same project).

rliotino99: Yes, I'm sorry it was a typo between versions

njacazio (approved):

PR #16275, opened 2026-05-16
[Common] Add run 3 hyperfine percentiling

ddobrigk:
* adds an adaptive percentiling setting in which 0.00001% is viable, which should be fine in any run with well over 10 million events. Based on inspecting the typical signals, a very high multiplicity selection is likely to be fine with midrapidity counters but caution should be exercised with FIT signals, as these have a tail that is likely beam-background or otherwise spurious: QA is always needed!
* adds sanity checks to capture problematic calibration situations and issue an invalid (but mathematically functional) calibration in that case. Checks include excessively fine binning requested for the input information as well as situations in which the detector signal is zero (e.g. no MFT in datataking). 
* N.B.: the calibrator that is touched in this PR is legacy code (it's very old...) and thus spawns several linter warnings: will be fixed in a future PR.


alibuild (approved):

PR #16308, opened 2026-05-19
[PWGJE] isolation fixes & MC prompt photon information fixes

fjonasALICE:
This PR fixes two critical bugs:
1. The identification of direct prompt photons in MC was not done correctly, and direct prompt was falsely identified as fragmentation prompt
2. The isolation enegry calculation using KDTrees did not correctly take into account that phi is a periodic variable. This is now fixed, in the same way it was fixed for jet matching in the framework (which also uses KDTrees)


nzardosh (approved):

PWGJE/Tasks/gammaJetTreeProducer.cxx
```diff
@@ -494,6 +558,46 @@ struct GammaJetTreeProducer {
...
 
+  /// \brief Prints the mother chain of a given MC particle.
+  /// \param particle The particle to print the mother chain for
+  template <typename T>
+  void printMotherChain(const T& particle) const
+  {
+    T current = particle;
+    int depth = 0;
+    bool continueTracing = true;
+    while (continueTracing) {
```
vkucera, line 569:
```suggestion
    while (true) {
```

PR #16323, opened 2026-05-20
[PWGLF] Add quadratic response option

lbariogl:


PWGLF/TableProducer/QC/flowQC.cxx
```diff
@@ -270,74 +263,77 @@ struct flowQC {
 
     float centrality = getCentrality(collision);
 
+    const bool quadraticResponse = cfgQuadraticResponse;
```
vkucera: Why do you duplicate the `cfgQuadraticResponse`?

lbariogl: Changed

maciacco (approved):

PR #16325, opened 2026-05-20
[PWGDQ] make matching QA table filling more flexible

aferrero2707:
Instead of using an hard-coded list of matching methods and corresponding indexes when filling the derived tables, the filling is done using a dynamic list of methods and corresponding indexes that is based on what is actually requested in the configuration.

The horizontal bin labels of a dedicated TH1 histogram provide the correspondence between method labels and their indexes, to be used when reading back the derived tables to generate final histograms.

An `isTagged` column is also added to the derived MCH table to flag the tracks that have good quality and cross the active area of the MFT, and therefore are expected to have a matchable MFT track. The tagging does not rely on MC truth, and can therefore be also applied to real data.


aferrero2707:
@mcoquet642 this is now ready for review, could you please have a look? Thanks!

mcoquet642 (approved):

PR #16333, opened 2026-05-21
Added a new axis for the number of primary vertex (PV) contributors  in data processing

Fan-Cao:
Added a new axis for the number of primary vertex (PV) contributors to all THnSparse histograms in data processing.
This axis is only created and filled when the `storeCentrality` configuration flag is enabled, maintaining full backward compatibility.


vkucera:
The PR is empty.

PR #16334, opened 2026-05-21
[PWGJE] Add PCH to more Jet tasks

ktf:


PWGJE/Tasks/CMakeLists.txt
```diff
@@ -207,233 +209,290 @@ if(FastJet_FOUND)
...
                         PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::PWGJECore O2Physics::AnalysisCore
                         COMPONENT_NAME Analysis)
     o2physics_add_dpl_workflow(jet-correlation-d0
                         SOURCES jetCorrelationD0.cxx
+                        REUSE_FROM JetSubstructureHFPCH
                         PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::PWGJECore O2Physics::AnalysisCore
                         COMPONENT_NAME Analysis)
     o2physics_add_dpl_workflow(hf-debug
                         SOURCES hfDebug.cxx
+                        REUSE_FROM JetSubstructureHFOutputPCH
```
fjonasALICE, line 470: is it intended that this is the only workflow where is uses ...OutputPCH instead of ...HFPCH.?

vkucera: I guess this is because it needs `PWGHF/DataModel/DerivedTables.h` while the other tasks don't.

PR #16335, opened 2026-05-21
[PWGLF] Fix Omega2012 input subscription

jjh740707-prog:
This patch adds ResoTracks as an unused input to Omega2012Analysis::processData.

The table is not used in the analysis logic, but subscribing to it aligns the process signature with the resonance daughter initializer output and avoids DPL input framing issues when running the task with ResoCascades and ResoV0s.

### DIFF
--- a/PWGLF/Tasks/Resonances/omega2012Analysis.cxx
+++ b/PWGLF/Tasks/Resonances/omega2012Analysis.cxx
@@ -698,6 +698,7 @@ struct Omega2012Analysis {
   PROCESS_SWITCH(Omega2012Analysis, processDummy, "Process Dummy", true);
 
   void processData(const aod::ResoCollision& collision,
+                   aod::ResoTracks const& /*resotracks*/,
                    aod::ResoCascades const& resocasc,
                    aod::ResoV0s const& resov0s)
   {


vkucera:
@jjh740707-prog Can you please elaborate what kind of "DPL input framing issues" is this fixing?

vkucera:
@jjh740707-prog Please set up your GitHub account such that colleagues can identify you.

jjh740707-prog:
@vkucera Thanks for asking. I should have described this more clearly.

The issue was not related to the Omega(2012) physics selection itself, but to the DPL input subscription configuration used in the resonance workflow.

In my local workflow, I run

o2-analysis-lf-resonance-module-initializer
o2-analysis-lf-omega2012analysis

with the resonance daughter tables enabled. The initializer produces the daughter tables used by the analysis task, namely:

ResoTracks
ResoCascades
ResoV0s

Before this patch, the task subscribed only to:

ResoCollision
ResoCascades
ResoV0s

through

void processData(const aod::ResoCollision& collision,
aod::ResoCascades const& resocasc,
aod::ResoV0s const& resov0s)

In my local tests, this configuration did not behave consistently when used together with the resonance initializer workflow. The symptoms included DPL/runtime messages such as

"Header is not a DataHeader?"

and the analysis output was not filled correctly in the failing runs.

After comparing with the subscription pattern used in other resonance tasks, I found that explicitly subscribing to ResoTracks resolves the issue by making the task subscribe to the complete set of resonance daughter tables produced by the initializer.

The patch therefore changes the signature to

void processData(const aod::ResoCollision& collision,
aod::ResoTracks const& /*resotracks*/,
aod::ResoCascades const& resocasc,
aod::ResoV0s const& resov0s)

The ResoTracks table is intentionally unused in the analysis logic. It is added only to ensure the expected DPL input subscription configuration.

The actual Omega(2012) reconstruction remains unchanged: the analysis still uses only ResoCascades and ResoV0s in fill<false>(), so the selection, invariant-mass calculation, and histogram filling logic are unaffected.

After this change, the same local workflow completed successfully, all devices exited with code 0, and the same-event and mixed-event Omega(2012) histograms were filled as expected.

mjkim525:
Hi @jjh740707-prog,

Is there any reason for using o2-analysis-lf-resonance-module-initializer instead of o2-analysis-lf-resonance-initializer?

Also, are these results based on the discussion with @BongHwi, the author?

jjh740707-prog:
Hi, thank you for your comment.

These results were not based on a prior discussion with @BongHwi. I used `o2-analysis-lf-resonance-module-initializer` based on my current understanding of the workflow, but I may be missing the recommended setup for this channel.

Could you please clarify why `o2-analysis-lf-resonance-initializer` would be preferred here, especially for producing the inputs needed for the `Xi + K0S` analysis in `omega2012Analysis`? I would like to understand the intended workflow before making further changes.

For the next iteration, I will first consult @BongHwi, as the author, and then update the initializer and configuration accordingly.

vkucera:
@jjh740707-prog Thanks for the details but from your description I get the impression that you don't actually know why it works now and before it didn't. In particular,

> The ResoTracks table is intentionally unused in the analysis logic. It is added only to ensure the expected DPL input subscription configuration.

This does not make sense. Either a table is used and the subscription is needed or it is not used and the subscription should be omitted.
Please clarify which code uses the added table, otherwise I don't think your changes are well justified.

jjh740707-prog:
Thank you for pointing this out.

Following your advice, I rechecked the workflow and resolved the issue by using o2-analysis-lf-resonance-initializer, which provides the resonance tables required for this analysis.

Thank you again for your careful review and helpful guidance.

PR #16340, opened 2026-05-21
[PWGDQ] enable processing of real data in matching tasks

aferrero2707:
The ML training data producer and the matching QA tasks are modified to allow processing real data as well as MC data. A dedicated process function is added to both tasks.

It requires https://github.com/AliceO2Group/O2Physics/pull/16325 to be merged first.


aferrero2707:
@mcoquet642 the commits can be squashed when merging, they are only there to hopefully simplify the review. The code is by the way ready for review from my side. Thanks!

mcoquet642 (approved):

PR #16355, opened 2026-05-22
[PWGEM/PhotonMeson] Add histograms for resolution studies

jokonig:
- Eta and Phi dependent momentum resolution of V0 photons Ideally, these histograms can be used to construct the meson eficiency from the single photon efficiency


alibuild (approved):

PR #16360, opened 2026-05-23
[PWGEM] Add b2cc2ll signal to generated output

feisenhu:


alibuild (approved):

PR #16372, opened 2026-05-25
[PWGHF] Add task h2fromlb

MRazza879:
New task for the analysis of non-prompt deuterons from beauty-hadron decays.
The task runs on MC and data, selecting anti-deuteron candidates.


vkucera:
- Why is this considered an HF workflow?
- Why do you call a deuteron "h2"?

ercolessi:
Hi @vkucera  
- this is an HF analysis, presented and discussed in PAG D2H
- What is the problem with referring to deuteron with H2 (Hydrogen-2) ? It is consistent with EventFiltering

vkucera:
>     * What is the problem with referring to deuteron with H2 (Hydrogen-2) ? It is consistent with EventFiltering

Well, the symbol for deuteron has always been d. I don't see why anyone would call it H2, which has always been the symbol for the hydrogen molecule.

ercolessi:
> >     * What is the problem with referring to deuteron with H2 (Hydrogen-2) ? It is consistent with EventFiltering
> 
> 
> 
> Well, the symbol for deuteron has always been d. I don't see why anyone would call it H2, which has always been the symbol for the hydrogen molecule.

I don't think anybody is going to assume we are reconstructing hydrogen molecules in this task, in my opinion it would be more useful to keep it consistent with the corresponding EventFiltering task. 

But if you feel strongly about the notation sure, h2 can be changed to d, just let us know

vkucera:
> > > ```
> > > * What is the problem with referring to deuteron with H2 (Hydrogen-2) ? It is consistent with EventFiltering
> > > ```
> > 
> > 
> > Well, the symbol for deuteron has always been d. I don't see why anyone would call it H2, which has always been the symbol for the hydrogen molecule.
> 
> I don't think anybody is going to assume we are reconstructing hydrogen molecules in this task, in my opinion it would be more useful to keep it consistent with the corresponding EventFiltering task.
> 
> But if you feel strongly about the notation sure, h2 can be changed to d, just let us know

I think the best would be to just call it what it is, i.e. "deuteron".

stefanopolitano (approved):

PR #16380, opened 2026-05-25
[PWGCF] Add Bertsch-Pratt LCMS decomposition for femto pairs

IgorPtak:


PWGCF/Femto/Core/pairHistManager.h
```diff
@@ -920,6 +954,42 @@ class PairHistManager
...
+    const double kside1 = (-p1.Px() * tPy + p1.Py() * tPx) / tPt;
+    const double klong1 = gammaL * (p1.Pz() - betaL * p1.E());
+
+    const double kout2 = (p2.Px() * tPx + p2.Py() * tPy) / tPt;
+    const double kside2 = (p2.Py() * tPx - p2.Px() * tPy) / tPt;
+    const double klong2 = gammaL * (p2.Pz() - betaL * p2.E());
+
+    mQout = static_cast<float>(kout1 - kout2);
+    mQside = static_cast<float>(kside1 - kside2);
+    mQlong = static_cast<float>(klong1 - klong2);
```
ariedel-cern: I think it would be good to return these values as a tuple, so one can call the same function for MC. Something like

```
std::tuple<float, float, float> computeBertschPrattLCMS(/* params */)
{
    // ...
    return { qOut, qSide, qLong };
}
```
then they could be unpaced like
```
[mQuut, mQside, mQlong] = computeBertschPrattLCMS(p1, p2);
```

and then at some later point you could simply add

```
[mQuotTrue, mQsideTrue, mQlongTrue] = computeBertschPrattLCMS(p1True, p2True);
```

ariedel-cern (dismissed):
Good for now, maybe implement the comment when you add MC processing.

ariedel-cern (approved):
Thanks for implementing the comment

PR #16381, opened 2026-05-26
[PWGHF] [xic0omegac0] Update codes to fix PID assignment in KFParticle

jinhyunni:
This PR aims to fix PID assignment when using `getTrackParCovFromKFP()` method in `candidateCreatorXic0Omegac0Qa.cxx` code.

@fcatalan92, @mtorresc15


fcatalan92:
@jinhyunni  Thanks for spotting and fixing this issue! The PR is fine for me, however I think you can solve the same issue in a cleaner way if you define a helper function that converts the PDG code to the correct o2::track::PID one, as Fernanda did in https://github.com/AliceO2Group/O2Physics/pull/16352 .

jinhyunni:
Hi @fcatalan92, thank you for reviewing this PR and for sharing your thoughts on this point.

I agree that introducing a helper function would make the implementation cleaner. My initial intention was to keep these assignments together with the other decay-channel-dependent variables in the `init` function, so that all hypotheses for each reconstruction channel are defined in one place.

That said, if you think the helper-function approach would be better for maintainability or consistency with the existing code style, I would be happy to update the implementation accordingly.

fcatalan92:
> That said, if you think the helper-function approach would be better for maintainability or consistency with the existing code style, I would be happy to update the implementation accordingly.

Hi @jinhyunni, yes, if you have time, it would be great if you could update the PR using the helper-function approach. Thanks!

jinhyunni:
Hi @fcatalan92, thank you for your guidance on how to better organize the code.
I have revised this PR according to your recommendation.
Could you please take another look, and if there are no further changes needed, would you be able to merge this PR?

jinhyunni:
> Error while checking build/O2Physics/staging for [4fc0b25](https://github.com/AliceO2Group/O2Physics/commit/4fc0b252041b44979bbad0ac524867af1430d1c1) at 2026-05-27 07:12:
> 
> ```
> ## sw/BUILD/O2Physics-latest/log
> /sw/SOURCES/O2Physics/slc9_x86-64-slc9_x86-64/0/PWGHF/TableProducer/candidateCreatorXic0Omegac0Qa.cxx:516:3: error: control reaches end of non-void function [-Werror=return-type]
> ninja: build stopped: subcommand failed.
> ```
> 
> Full log [here](https://ali-ci.cern.ch/alice-build-logs/AliceO2Group/O2Physics/16381/4fc0b252041b44979bbad0ac524867af1430d1c1/build_O2Physics_staging/pretty.html).

This issue was caused by missing return value at default case, and it was fixed with the latest commit(https://github.com/AliceO2Group/O2Physics/pull/16381/commits/ab891357f60c6ecf804e77a87b3afd071885a03c).

fcatalan92 (approved):

PR #16384, opened 2026-05-26
[PWGDQ] Remove string creation from inner loop

ktf:


ktf:
@iarsene could you cross check the names are correct?

mcoquet642 (approved):

PR #16393, opened 2026-05-26
[Common,PWGLF] Add TOF NSigma from TOF PID framework

romainschotter:
-  Add TOF NSigma from TOF PID framework in the strangeness TOF PID (thanks @dsekihat )
- Collision time and time error are calculated as the average of the tracks' time and time error
- change PIDTOFParamService to allow using the TOF Nsigma from TOF PID framework when running over strangeness derived data (otherwise ccdb configurables are always fetched from the tof-signal task)

@njacazio @dsekihat for your information


njacazio:
Ciao @romainschotter that's great thanks!
Since they are orthogonal, could you split the two changes in two PRs? i.e. one for the inheritance, one for the rest?
Apart from that looks good, many thanks!
Cheers,
N.

PWGLF/DataModel/LFStrangenessPIDTables.h
```diff
@@ -112,6 +112,7 @@ DECLARE_SOA_INDEX_COLUMN(DauTrackExtra, dauTrackExtra); //! point to daughter th
 DECLARE_SOA_INDEX_COLUMN(StraCollision, straCollision); //! point to collision associated with this track (not the V0/Casc)
 DECLARE_SOA_COLUMN(TOFSignal, tofSignal, float);        //! track TOF signal
 DECLARE_SOA_COLUMN(TOFEvTime, tofEvTime, float);        //! event time
+DECLARE_SOA_COLUMN(TOFEvTimeErr, tofEvTimeErr, float);  //! event time error for TOF
```
njacazio, line 115: This in principle could be left and in the table the value from https://github.com/AliceO2Group/O2Physics/blob/master/Common/DataModel/PIDResponseTOF.h#L304

njacazio (dismissed):

romainschotter:
Ciao @njacazio ! 
Of course, I can split the PR into two PRs but the `strangenesstofpid` task won't work anymore when running over derived data and using the TOF framework to calculate the TOF NSigma (at least, until the second PR would be merged).
That's because the PIDTOFParamService expects the workflow to depend on the `tof-signal` task, which carries the CCDB configuration for the PIDTOFParamService. In order to have a self-consistent PR, I have put together the changes in the strangeness and TOF frameworks.
Actually, this makes me realize that I forgot to add a commit to the PR (the one where we tell the PIDTOFParamService from which task the CCDB configuration should be taken from)

njacazio:
> Ciao @njacazio ! Of course, I can split the PR into two PRs but the `strangenesstofpid` task won't work anymore when running over derived data and using the TOF framework to calculate the TOF NSigma (at least, until the second PR would be merged). That's because the PIDTOFParamService expects the workflow to depend on the `tof-signal` task, which carries the CCDB configuration for the PIDTOFParamService. In order to have a self-consistent PR, I have put together the changes in the strangeness and TOF frameworks. Actually, this makes me realize that I forgot to add a commit to the PR (the one where we tell the PIDTOFParamService from which task the CCDB configuration should be taken from)

Ciao @romainschotter, thank you for your clear and well-formed answer (not always granted!), then it's fine with me

lhusova (dismissed):

lhusova (approved):

njacazio (approved):

dsekihat (approved):

Common/Core/PID/PIDTOFParamService.cxx
```diff
@@ -69,7 +69,8 @@ void o2::pid::tof::TOFResponseImpl::inheritFromBaseTask(o2::framework::InitConte
 }
 
 void o2::pid::tof::TOFResponseImpl::initSetup(o2::ccdb::BasicCCDBManager* ccdb,
-                                              o2::framework::InitContext& initContext)
+                                              o2::framework::InitContext& initContext,
+                                              const std::string task)
```
vkucera, line 73: Don't pass strings by value.

PR #16394, opened 2026-05-26
Complete Delta resonance Analysis

bhattdurgesh921-sys:


dmallick2:
@bhattdurgesh921-sys @arpit-hep  may I know what is the motivation for adding similar analysis task on delta-analysis. we have similar task for pp collisions on delta-analysis  (https://github.com/AliceO2Group/O2Physics/blob/master/PWGLF/Tasks/Resonances/deltaanalysis.cxx). I would request to you please use this one. If there is no strong reason, we should avoid adding similar task twice.

bhattdurgesh921-sys:
> @bhattdurgesh921-sys @arpit-hep may I know what is the motivation for adding similar analysis task on delta-analysis. we have similar task for pp collisions on delta-analysis (https://github.com/AliceO2Group/O2Physics/blob/master/PWGLF/Tasks/Resonances/deltaanalysis.cxx). I would request to you please use this one. If there is no strong reason, we should avoid adding similar task twice.



Task is similar but there are so much additional things which i use in this code so that's why i add a separate code which is also applicable for asymmetric systems also

bhattdurgesh921-sys:
This pull request creates DeltaAnalysis1232, transitioning the codebase to the modern O2 resonance framework by integrating ResoInitializer for optimized task setup and ResoTracks for streamlined data access. The update adopts high-performance DF (Derived Data) processing workflows, implementing memory-efficient event mixing via o2::soa::combinations to replace manual pool management with a scalable approach that supports event-plane dependent mixing. PID and track selection have been significantly hardened through species-specific, momentum-binned TPC+TOF circular nσ cuts and pT-dependent DCA constraints, effectively mitigating contamination in high-multiplicity environments. To ensure robust systematic studies, the analysis now incorporates rotational background estimation, occupancy-in-time-range pile-up mitigation, and a dedicated processMCGen workflow for precise efficiency bookkeeping. Finally, the output infrastructure has been standardized into 5D THnSparse histograms, complemented by comprehensive QA monitoring directories to streamline debugging and systematic validation.

PR #16396, opened 2026-05-26
Avoid std::map in EventSelectionModule

ktf:


ktf:
@ddobrigk can you have a look here? Looking up in a map is very expensive due to the pointer traversal and constructing it requires one allocation per node which is also expensive. A binary search in a vector of pairs should be much faster. Also, @aalkin was suggesting to use a index table for this?

ktf:
@ekryshen can you have a look at this?

ekryshen:
@ktf Thanks, looks reasonable. 

Actually I suspect that we don't need this map or vector of pairs any longer. It was introduced as a workaround for early passes of pp2022 where trigger info wasn't aligned with actual bcs. As far as I understand, the trigger functionality is used by EMCAL analysers only. I will check with them if they still need support of old passes. If not, we can simply remove this workaround.

ekryshen:
@ktf I've just got a confirmation that early passes are not used by ECAL analysers so this functionality is not needed any longer. How should we proceed? We can either merge or close this PR, then I will clean up this code.

ktf:
removing the mapping completely is of course fine with me. can you take care of it?

ekryshen:
Done here https://github.com/AliceO2Group/O2Physics/pull/16489. I guess we can close this PR

ktf:
Thanks

PR #16397, opened 2026-05-26
[Common] Improve performance of FulfillsITSHitRequirements

ktf:
Use a bitmask rather than an std::set.


ktf:
@ddobrigk can you please have a look also here? This also pops up in the profiles.

ktf:
@dsekihat any reason why this is not merged? Any reason for the TrackSelection class to have a dictionary, BTW?

dsekihat (approved):

dsekihat:
Dear @ktf , My fault. I overlooked the automatic notification. It is merged just now.

PR #16398, opened 2026-05-26
[PWGDQ] Add globalBc to DQ MC data model for MC truth BC checks

sigurdnese:
- Add globalBc to reducedMcEvents, in order to calculate MC truth TF and ITSROF border checks
- New column is filled in tableMakerMc(_withAssoc), passing the globalBc from central data model McCollisions
- Add converter to fill 'globalBc' column with dummy 0
- Implement calculation of MC IsTFBorder and IsITSROFBorder variables in the VarManager
- New helper function to set TF border calculation configurables and metadata
- The TF border calculation ('recomputed') is also implemented for the real data case, akin to the existing ITSROFBorderRecomputed variable. The initialization of metadata and configurables are added to tableReader_withAssoc
- All TFBorder and ITSROFBorder variables given names and units in the VarManager 
- The asymmetric-pairing task in MC (used for the UPC D0 analysis) also gets access to the MC TFBorder and ITSROFBorder calculations, by fetching the configurables from analysis-event-selection

https://indico.cern.ch/event/1688270/#11-update-of-the-reducedmceven

### DIFF
--- a/PWGDQ/Tasks/ModelConverterReducedMCEvents002.cxx
+++ b/PWGDQ/Tasks/ModelConverterReducedMCEvents002.cxx
@@ -20,7 +20,6 @@
 #include <Framework/AnalysisHelpers.h>
 #include <Framework/AnalysisTask.h>
 #include <Framework/runDataProcessing.h>
-#include <sys/types.h>
 
 using namespace o2;
 using namespace o2::framework;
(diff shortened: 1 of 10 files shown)


PWGDQ/Tasks/ModelConverterReducedMCEvents002.cxx
vkucera, line 1: This is not a task. This is a table producer.

vkucera:
This PR is stale.

PR #16400, opened 2026-05-26
[Common] CTP: avoid useless work to count non-zero bits

ktf:


ktf:
similar one in O2 already approved

PR #16408, opened 2026-05-27
[Common] Fix includes

vkucera:


vkucera:
@dsekihat @ddobrigk Please merge the PR.

Common/Tasks/qVectorsCorrection.cxx
```diff
@@ -39,6 +39,7 @@
 
 #include <sys/types.h>
 
+#include <concepts>
```
ktf, line 42: I do not think it's a good idea to tell people to include <concepts> if they do not use it. Maybe this should be fixed at the framework level? What is actually requiring it?

vkucera: `std::derived_from`

ktf: ah, ok, I thought it was some dependency percolating from the Framework.

PR #16429, opened 2026-05-28
[PWGDQ] fix reduced MCH track index in output table

aferrero2707:
The reduced index of the MCH tracks in the output table needs to be declared as an index column in order to allow to properly retrieve the MCH track from the index stored in the candidates table.


aferrero2707:
@mcoquet642 this is ready for review, thanks!

mcoquet642 (approved):

PR #16439, opened 2026-05-28
[PWGHF] Fill in empty PDG value for KFParticle of V0 and CharmBaryon

jinhyunni:
This PR aims to asign PDG values to KFParticle object of V0 and Charm Baryon.
Because of this missing PDG value assignment, usage of helper function `convertPDGIntoPID` was not able to return the right PID value.

@fcatalan92, @mtorresc15


fcatalan92 (approved):
Thanks @jinhyunni

PR #16448, opened 2026-05-29
[ALICE3] Switch from McPartWithDaus to McParticles with --aod-origin-replace in otf decayer

jesgum:
Switch to using `--aod-origin-replace` so we can reuse the McParticle datamodel instead of making a copy
Subsequent workflows have been adjusted accordingly -- but may need further adjustments (which will come in a future pr)


njacazio (approved):
Great work, thanks!

PR #16459, opened 2026-05-29
[PWGCF] Use ITS pid instead of TPC when configured

SwatiSaha-1997:
Use ITS pid instead of TPC: make the choice based on cofigurable


alibuild (approved):

PR #16468, opened 2026-05-30
[PWGEM/Dilepton] reduce data size for SCT

dsekihat:
[PWGEM/Dilepton] reduce data size for SCT


alibuild (approved):

PR #16470, opened 2026-05-30
[PWGHF] add more deuteron candidate preselection flags

zhangbiao-phy:
Added configurations for deuteron candidate selection and DCA preselection.


alibuild (approved):

PR #16488, opened 2026-06-01
[Infrastructure] Adding rights for the new PWG-CF convener (me) and new femtoscopy PAG…

lauraser:
… coordinator (Wiola)


jgrosseo:
Shoudn't we also remove the previous convener(s)?

lauraser:
Good point, apologies for forgetting that. Done in a new PR, thanks Jan!

PR #16528, opened 2026-06-04
[PWGDQ] Adding functionality for event mixing across TFs

iarsene:
This is implemented so far just for the dielectron channel, in the table-reader-with-assoc. The event mixing categories are created in the same way as for the usual mixing (within a single TF), but they are configured directly in the same-event-pairing task.
In this mixing, the candidate electrons are temporarily stored in event pools which are permanent (data members of the task) and included in the mixing using a rolling buffer of configurable depth. The important aspect is that in the case when mixing is done with very many categories or the signal is very rare, and the pools cannot be filled during a single TF, this method will allow using all the TFs in the job, so all pools can get enough events.


alibuild (approved):

vkucera:
Please fix the issues reported in the bug tracker.

PR #16547, opened 2026-06-05
[PWGCF] Add multNTracksPV multiplicity estimator in the task

abmodak:


lauraser (approved):
I approve as this is urgent. However, please update it later to include a check that if both estimators are on, you get an error. It can else introduce errors if by accident you turn on both.

PR #16552, opened 2026-06-07
[ALICE3] Implement conditional saving of hit maps

njacazio:
Added conditional saving of hit maps in destructor.


alibuild (approved):

PR #16559, opened 2026-06-07
[PWGLF] Fix PDG code in strangenessBuilderModule

romainschotter:
Fix PDG code check when running generated cascades

Thanks a lot @nepeivodaRS for reporting it!


alibuild (approved):

PR #16560, opened 2026-06-08
[PWGLF] Fix strangenesstofpid converter

romainschotter:
- fill `DauTrackTOFPIDs_002` table with available columns in `DauTrackTOFPIDs_001`
- add safety in the `strangenesstofpid` to prevent heap corruption and warn the user (even though it could intentional in case using specific converters)


alibuild (approved):

PR #16565, opened 2026-06-08
[Common] Reserve memory upfront when creating the track propagation table

ktf:


ktf:
Only changes the upfront reservation. Verified on hyperloop with https://alimonitor.cern.ch/hyperloop/ which produces bit by bit the same results for the 85 datasets which are in common. Merging.

PR #16569, opened 2026-06-08
[PWGHF] trackIndexSkimCreator: Avoid computing the pVector of bachelor1 in the inner loop

ktf:


PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -4131,6 +4131,12 @@ struct HfTrackIndexSkimCreatorLfCascades {
               continue;
             }
 
+            // bachelor1 momentum is invariant across the inner (bachelor2) loop;
+            // pVector() is a dynamic column with no caching (it recomputes
+            // sin/cos on every call), so evaluate it once here instead of on
+            // every bachelor2 iteration.
+            const std::array<float, 3> pVecCharmBachelor1 = trackCharmBachelor1.pVector();
```
vkucera: Thanks @ktf . I think the verbose comment is not needed.

vkucera:
```suggestion
            const auto pVecCharmBachelor1 = trackCharmBachelor1.pVector();
```

vkucera (changes requested):

vkucera (approved):

PR #16587, opened 2026-06-09
[ALICE3] Fix indices bug in otf decayer

jesgum:


jesgum:
Indices are now properly mapped between mothers and daughters within all collisions in a dataframe instead of just the first one. Applies for both for particles that already decayed in the generator and for particles decayed in the decayer

@njacazio fyi

alibuild (dismissed):

alibuild (approved):

PR #16592, opened 2026-06-09
[PWGDQ,PWGEM] Adding missing reduced table dependence for q-vectors

lucamicheletti93:
Adding missing reduced table dependence for q-vectors


vkucera (changes requested):
Fix the title prefix. (There is no need to add it by hand.)

rbailhac (approved):

vkucera (approved):

alibuild (approved):

PR #16593, opened 2026-06-09
[PWGCF] FemtoUniverse: Fix hardcoded pion mass, add filling only primary MC pairs

alicja-pp:
`PWGCF/FemtoUniverse/Core/FemtoUniverseFemtoContainer.h`
add option to fill only primary MC pairs in MC histograms

`PWGCF/FemtoUniverse/Core/FemtoUniversePairSHCentMultKt.h`
fix harcoded setPionPairMass() into setPDGCodes that sets masses according to configurable

`PWGCF/FemtoUniverse/Tasks/femtoUniversePairTaskTrackTrackMultKtExtended.cxx`
add configurable _confOnlyPrimaryMCPair_

[pdg/database] O2 linter error will be fixed in the future, as it is not needed right now, and there were issues with fix

### DIFF
--- a/PWGCF/FemtoUniverse/Core/FemtoUniversePairSHCentMultKt.h
+++ b/PWGCF/FemtoUniverse/Core/FemtoUniversePairSHCentMultKt.h
@@ -244,10 +244,10 @@ class PairSHCentMultKt
   /// Set the PDG codes of the two particles involved
   /// \param pdg1 PDG code of particle one
   /// \param pdg2 PDG code of particle two
-  void setPDGCodes(const int pdg1, const int pdg2)
+  void setPairMass(const float mass1, const float mass2)
   {
-    mMassOne = TDatabasePDG::Instance()->GetParticle(pdg1)->Mass();
-    mMassTwo = TDatabasePDG::Instance()->GetParticle(pdg2)->Mass();
+    mMassOne = mass1;
+    mMassTwo = mass2;
   }
 
   /// To compute the bin value for cavariance matrix
--- a/PWGCF/FemtoUniverse/Core/FemtoUniversePairSHCentMultKt.h
+++ b/PWGCF/FemtoUniverse/Core/FemtoUniversePairSHCentMultKt.h
@@ -84,7 +84,6 @@ class PairSHCentMultKt
     int em = 0;
     int il = 0;
     do {
-
       fels[il] = el;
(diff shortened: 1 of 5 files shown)


PWGCF/FemtoUniverse/Core/FemtoUniversePairSHCentMultKt.h
```diff
@@ -245,10 +244,10 @@ class PairSHCentMultKt
...
   /// \param pdg1 PDG code of particle one
   /// \param pdg2 PDG code of particle two
-  void setPionPairMass()
+  void setPDGCodes(const int pdg1, const int pdg2)
   {
-    mMassOne = o2::constants::physics::MassPiPlus; // FIXME: Get from the PDG service of the common header
-    mMassTwo = o2::constants::physics::MassPiPlus; // FIXME: Get from the PDG service of the common header
+    mMassOne = TDatabasePDG::Instance()->GetParticle(pdg1)->Mass();
+    mMassTwo = TDatabasePDG::Instance()->GetParticle(pdg2)->Mass();
   }
```
vkucera, line 251: I already commented in the other PR that this is not allowed. Why are you adding it yet to another place?

vkucera (changes requested):

PWGCF/FemtoUniverse/Core/FemtoUniversePairSHCentMultKt.h
alicja-pp: I tried to change this but the solution gave a segmentation fault in this case. I need this function fix quicker. I plan to debug this O2 linter error when possible.

vkucera: - You need to pass a pointer to the PDG service plugin from the task.
- When will you fix it?
- At least add a `FIXME` comment.

alicja-pp: I have managed to find other way around it. Is it ok?

vkucera: Yes, this is much better. Thanks.

vkucera (approved):

prchakra (approved):

PR #16606, opened 2026-06-10
[PWGCF] Add v0-resonance pair task

ariedel-cern:
Adding part of the v0-resonance pair task (so far only lambda-phi, other combinations are following).


alibuild (approved):

PR #16608, opened 2026-06-10
[Common] Change default MatLUT

ddobrigk:
thanks to @mvolkl


vkucera:
Marked as stale due to 30 days of inactivity.

PR #16622, opened 2026-06-11
[PWGLF] Fix two bugs

gengjiabin12:


PWGLF/TableProducer/Strangeness/hStrangeCorrelationFilter.cxx
```diff
@@ -764,7 +764,8 @@ struct HStrangeCorrelationFilter {
         continue;
       if (posdau.tpcNClsCrossedRows() < trackSelections.minTPCNCrossedRows)
         continue;
-
+      if (v0.dcaV0daughters() > v0Selection.dcaV0dau)
```
romainschotter, line 767: Hi @gengjiabin12 ! The PR looks fine, however there is one thing I am not sure to understand. Why adding an explicit cut on the DCA between V0 daughters while the process functions loop on prefiltered sample of V0s already passing the very same DCA between V0 daughters cut (l.203 for V0s and l.205 for cascades)? I might be missing something

romainschotter (changes requested):

PR #16625, opened 2026-06-11
[PWGCF] Update fluctuation table

fsii:
Update:
1. fluctuation tables
2. fluctuation calculator keys


SwatiSaha-1997 (approved):

vkucera:
```text
PWGCF/EbyEFluctuations/Tasks/partNumFluc.cxx:144:8: warning: the type of the loop variable 'iExponent' is different from the one returned by the iterator and generates an implicit conversion; you can either change the type to the matching one ('const int &' but 'const auto&' is always a valid option) or remove the reference to make it explicit that you are creating a new value [performance-implicit-conversion-in-loop]
PWGCF/EbyEFluctuations/Tasks/partNumFluc.cxx:145:10: warning: the type of the loop variable 'jExponent' is different from the one returned by the iterator and generates an implicit conversion; you can either change the type to the matching one ('const int &' but 'const auto&' is always a valid option) or remove the reference to make it explicit that you are creating a new value [performance-implicit-conversion-in-loop]
```

PR #16637, opened 2026-06-12
[PWGHF] Remove unsafe reserve in charm FemtoDream producers

zhangbiao-phy:
EDIT by @vkucera : Mitigates fatal caused by https://github.com/AliceO2Group/AliceO2/pull/15450

### DIFF
--- a/PWGHF/HFC/TableProducer/producerCharmHadronsTrackFemtoDream.cxx
+++ b/PWGHF/HFC/TableProducer/producerCharmHadronsTrackFemtoDream.cxx
@@ -253,8 +253,8 @@ struct HfProducerCharmHadronsTrackFemtoDream {
     trackRegistry.add("AnalysisQA/CutCounter", "; Bit; Counter", kTH1F, {{cutBits + 1, -0.5, cutBits + 0.5}});
 
     // event QA histograms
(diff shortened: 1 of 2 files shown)


alibuild (approved):

vkucera:
Hi @zhangbiao-phy , can you please briefly describe in the PR description the problem that this is solving?

zhangbiao-phy:
> Hi @zhangbiao-phy , can you please briefly describe in the PR description the problem that this is solving?

Hi @vkucera, This PR fixes a runtime fatal error triggered by the stricter `reserve()` checks introduced recently from this PR： https://github.com/AliceO2Group/O2Physics/commit/473e3981794d03629537ed763c03291633c10b54.

In these FemtoDream charm producers, the reconstructed charm candidate outputs were reserved with `candidates.size()` for a per-collision filtered candidate range. In some workflows this size can be zero while the range still yields a candidate during iteration, leading to errors such as:

`Table 'FDHfCand2Prong': filled 1 rows after reserve(0)`

Since the candidate range size is not a reliable upper bound for the number of rows filled in these tasks, this PR removes the unsafe `reserve()` calls for the reconstructed charm candidate output tables.

PR #16643, opened 2026-06-12
[PWGCF] flowSP: cleanup tasks add fitvalues multcut

cnkoster:


cnkoster:
Dear reviewers, 
Would it be possible to review and merge this PR?

wenyaCern (approved):

jaelpark:
Please add `delete` for the corresponding `new`. The correct place to do this is in the destructor. I think that'll be alright then. Even better idea might be to just upgrade to `unique_ptr`.

PR #16646, opened 2026-06-12
[PWGHF] Charm Baryon correlator: modified V0 selection and correlation with V0

singhra1994:
These are the changes in this commit
- Applied V0 selection in the correlator struct, initially it was at collision selection struct
- Changed the method of counting generated lambda0, now we can calculate efficiency of lambda0 with and without requirement of reconstructed vertex of generated collisions


alibuild (approved):
Auto-approving on behalf of @singhra1994.

PR #16663, opened 2026-06-15
[PWGCF/Femto] ~35% faster track-track mixing: hoist the per-window track slice

ktf:
In processMixedEvent the partner partitions are sliced per
  (collision1, collision2) pair, but collision1 is fixed across its entire
  mixing window. sliceByCached only caches the offset/count — it re-runs
  arrow::Table::Slice on every call (a per-column ArrayData allocation +
  shared_ptr refcount storm) — so at mixing depth D, collision1's track
  slice was rebuilt D times per window for no reason.

  Hoist sliceParticle1 into a std::optional declared before the
  combinations loop and (re)materialize it only when collision1.globalIndex()
  changes, i.e. once per window. collision2 genuinely varies per pair and is
  left untouched.

### DIFF
--- a/PWGCF/Femto/Core/pairProcessHelpers.h
+++ b/PWGCF/Femto/Core/pairProcessHelpers.h
@@ -23,6 +23,7 @@
 #include <Framework/Logger.h>
 
 #include <cstdint>
+#include <optional>
 
 namespace o2::analysis::femto
 {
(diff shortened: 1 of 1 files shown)


ktf:
@aalkin this is one of the hotspots for the slicing we were discussing today. This actually avoids half of the cost by moving the construction outside.

ktf:
https://alimonitor.cern.ch/hyperloop/view-wagon/57243/test-statistics

ktf:
~~Needs a bit more refinement... I think it's actually leaking right now.~~

ktf:
Nevermind, it's fine as it is. @dsekihat Can someone review and eventually merge this? It's potentially a massive speedup.

ariedel-cern:
Hi @ktf, many thanks for the contribution! I'll do a quick local check to verify the mixed-event output is consistent with what we had before, and will merge once confirmed. I also plan to carry over the equivalent change to our triplet-case implementation.

ariedel-cern (approved):

PR #16672, opened 2026-06-15
[ALICE3] Add basic cascade selections and mc association for a3 strangeness

jesgum:
@lhusova fyi


alibuild (approved):

PR #16688, opened 2026-06-16
[PWGCF] Fix strongly intensive eta-gap histograms

isputows:
This PR fixes a crash in the strongly intensive FB correlation task caused by missing eta-gap histogram handling after the histogram structure update.

The obsolete inclusive eta-gap histogram filling was removed and the task was cleaned up accordingly.


alibuild (approved):
Auto-approving on behalf of @isputows.

PR #16694, opened 2026-06-16
[PWGHF] Add MC related infomations into the c-deuteron task

zhangbiao-phy:
Tag @fgrosa for information


alibuild (approved):

PR #16698, opened 2026-06-16
[PWGDQ] Code development for coherent Jpsi A2 analysis

ypwangg:
1. Defined some observables in VarManager.h/cxx
2. Added some histograms and cuts
3. Added configurables and tasks in table-reader-with-assoc for A2 analysis
4. Enabled the coherent Jpsi A2 workflow in table-reader-with-assoc-direct

### DIFF
--- a/PWGDQ/Core/VarManager.h
+++ b/PWGDQ/Core/VarManager.h
@@ -823,8 +823,8 @@ class VarManager : public TObject
     kCos2DeltaPhiPP_FT0C,
     kNullA2,
     kInfA2,
-    kSel1,
-    kSel2,
+    kSel1, // if track1 is used in TPC Q vector calculation
+    kSel2, // if track2 is used in TPC Q vector calculation
     kDeltaPhiPair,
     kOpeningAngle,
     kQuadDCAabsXY,
--- a/PWGDQ/Core/VarManager.h
+++ b/PWGDQ/Core/VarManager.h
@@ -60,12 +60,12 @@
 #include <Math/Vector3Dfwd.h>
 #include <Math/Vector4D.h> // IWYU pragma: keep (do not replace with Math/Vector4Dfwd.h)
 #include <Math/Vector4Dfwd.h>
+#include <TH3.h>
 #include <TMath.h>
 #include <TMathBase.h>
 #include <TObject.h>
 #include <TRandom.h>
 #include <TString.h>
-#include <TH3.h>
 
 #include <KFPTrack.h>
 #include <KFPVertex.h>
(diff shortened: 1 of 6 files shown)


PWGDQ/Core/VarManager.h
```diff
@@ -6790,6 +7044,49 @@ float VarManager::LorentzTransformJpsihadroncosChi(TString Option, T1 const& v1,
...
+      trackSelected = false;
+      return trackSelected;
+    }
+    if (t.dcaXY() > 0.0105 + 0.0350 / std::pow(t.pt(), 1.1)) {
+      trackSelected = false;
+      return trackSelected;
+    }
+  }
+  return trackSelected;
+}
```
iarsene: These are all hardcoded selections, which will make it difficult for other analyses (that might use other cuts to compute event plane). I suggest to just have a configurable param in the same-event-pairing to specify an AnalysisCut for the tracks used in the Q vector calculation. You can apply that in the same-event-pairing task, and if it evaluates to true, you can transmit via parameter to the VarManager fill function whether that track should be subtracted or not. The track cut itself can be computed in the track-selection task in parallel with all the other track cuts.

PWGDQ/Tasks/tableReader_withAssoc.cxx
```diff
@@ -1353,6 +1360,10 @@ struct AnalysisSameEventPairing {
     Configurable<bool> useRemoteCollisionInfo{"cfgUseRemoteCollisionInfo", false, "Use remote collision information from CCDB"};
     Configurable<bool> useEfficiencyWeighting{"cfgUseEfficiencyWeighting", false, "Apply efficiency weighting to the pairs from CCDB"};
     Configurable<int> efficiencyType{"cfgEfficiencyType", 0, "Type of efficiency to apply from CCDB: 0 no efficiency, 1 pt-cent-costhetastar"};
+    Configurable<bool> useRemoteFlow{"cfgUseRemoteFlow", false, "Use remote flow information from CCDB"};
+    Configurable<bool> useLocalFlow{"cfgUseLocalFlow", false, "Use flow information from local cache"};
```
iarsene: I would discourage the use of local files for this. Everything in ALICE should be reproducible, and using correction files directly from your local machine is breaking that. Is there a problem in using CCDB ?

PWGDQ/Tasks/tableReader_withAssoc.cxx
```diff
@@ -1808,6 +1819,37 @@ struct AnalysisSameEventPairing {
...
+      TFile* fileFlow = TFile::Open(pathFlow.Data(), "READ");
+      if (fileFlow == nullptr || fileFlow->IsZombie()) {
+        LOGF(fatal, "Flow resolution file %s cannot be opened", pathFlow.Data());
+      }
+      fileFlow->GetObject("ScalarProduct", ResoFlowSP);
+      fileFlow->GetObject("EventPlane", ResoFlowEP);
+      if (ResoFlowSP == nullptr || ResoFlowEP == nullptr) {
+        LOGF(fatal, "Flow resolution histograms not available in file %s", pathFlow.Data());
+      }
+    }
```
iarsene: Same comment as above. The files you plugin here are local and not reproducible

iarsene (approved):

PR #16703, opened 2026-06-17
[PWGDQ] adds single muon pTcut 1p5 1p7 and 2 GeV/c with MCH-MID matching

iamkoushikb-bit:
This PR adds new predefined single-muon cuts in CutsLibrary.cxx.

Added:

muonLowPt1p5
muonLowPt1p7
muonLowPt1p5_MchMidmatching
muonLowPt1p7_MchMidmatching
muonLowPt6_MchMidmatching

These cuts provide additional pT thresholds combined with MCH-MID matching for PWGDQ analyses and Hyperloop workflows.

### DIFF
--- a/PWGDQ/Core/CutsLibrary.cxx
+++ b/PWGDQ/Core/CutsLibrary.cxx
@@ -3042,7 +3042,25 @@ AnalysisCompositeCut* o2::aod::dqcuts::GetCompositeCut(const char* cutName)
     cut->AddCut(GetAnalysisCut("MCHMID"));
     return cut;
   }
-
+ 
+  if (!nameStr.compare("muonLowPt1p5_MchMidmatching")) {
+    cut->AddCut(GetAnalysisCut("muonLowPt1p5"));
+    cut->AddCut(GetAnalysisCut("matchedMchMid"));
+    return cut;
+  }
+  
+  if (!nameStr.compare("muonLowPt1p7_MchMidmatching")) {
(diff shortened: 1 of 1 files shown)


vkucera:
Don't open a PR for review if does not pass the checks.

vkucera:
@iamkoushikb-bit Please add your full name on your profile.

iamkoushikb-bit:
is this fine now? can it be merged now?

lucamicheletti93 (commented):
Dear @iamkoushikb-bit is it really needed to add these cuts? Are they necessary for the Upsilona analysis? In this case you can directly use minitree that allows you to do more cuts

iamkoushikb-bit:
Dear @lucamicheletti93, Thank you so much for your comment. Actually I want to do mixed event background subtraction and then perform centrality and dimuon-pt differential yield extraction. and mixed events studies are not possible with minitree right?

Currently I am using **DQ_LHC23_PbPb_pass4_muon** dataset - this was skimmed using muonQualityCuts, and I am putting **matchedMchMid** as my muon_selection criterion in O2_dq_table_reader workflow. So as of now without these new cuts my current mixed event study has no single muon pT cuts applied, I have followed your suggestion of using minitree and found that with single muon pT cuts the S/B improves, I hope this can improve my mixed event study. 

I would be eager to know your comments and suggestions on this. also I would really appreciate if you can suggest another way around like more suitable cuts.

PR #16713, opened 2026-06-17
[PWGHF] Add generated ct and resonant channel info for c-deuteron analysis

fgrosa:
@zhangbiao-phy can you check if it's fine? Please note that I also changed all the `TH1F` into `TH1D`, to avoid truncation of entries. Thanks a lot!

### DIFF
--- a/PWGHF/D2H/Tasks/taskCd.cxx
+++ b/PWGHF/D2H/Tasks/taskCd.cxx
@@ -681,7 +681,7 @@ struct HfTaskCd {
   void fillHistosMcGen(CandCdMcGen const& mcParticles, Coll const& recoCollisions)
   {
     for (const auto& particle : mcParticles) {
-      if (std::abs(particle.flagMcMatchGen()) == hf_decay::hf_cand_3prong::DecayChannelMain::CDeuteronToDeKPi) {
+      if (std::abs(particle.flagMcMatchGen()) != hf_decay::hf_cand_3prong::DecayChannelMain::CDeuteronToDeKPi) {
         continue;
       }
       const auto yGen = RecoDecay::y(particle.pVector(), o2::constants::physics::MassCDeuteron);
--- a/PWGHF/D2H/Tasks/taskCd.cxx
+++ b/PWGHF/D2H/Tasks/taskCd.cxx
@@ -117,6 +117,7 @@ DECLARE_SOA_COLUMN(FlagMcDecayChanRec, flagMcDecayChanRec, int8_t); //! Resonant
 DECLARE_SOA_COLUMN(OriginMcGen, originMcGen, int8_t);               //! MC origin for generated particles
 DECLARE_SOA_COLUMN(FlagMcDecayChanGen, flagMcDecayChanGen, int8_t); //! Resonant MC decay channel for generated candidates
 DECLARE_SOA_COLUMN(CtGen, ctGen, float);                            //! Generated ct computed wrt to c-deuteron production vertex, which can be either PV (prompt) or B-hadron decay vertex (non-prompt)
+DECLARE_SOA_COLUMN(CtRec, ctRec, float);                            //! Reconstructed ct computed wrt to PV
 DECLARE_SOA_COLUMN(Cent, cent, float);                              //! Centrality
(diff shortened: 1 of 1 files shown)


zhangbiao-phy:
Hi @fgrosa! Thanks a lot! They looks good to me! Maybe we can also add ct in the lite tree for MC reco and data? we can calculated it by ourself only from full tree.

fgrosa:
> Hi @fgrosa! Thanks a lot! They looks good to me! Maybe we can also add ct in the lite tree for MC reco and data? we can calculated it by ourself only from full tree.

Currently I put the `ctGen`, which is computed considering the actual decay length of the c-deuteron (so in case of non-prompt starting from the decay vertex of the b hadron and not the PV) to be able to reweight it to test the impact on the efficiency. But if you prefer I can also add the reco one (which would instead start from the PV regardless the origin of the candidate, since we cannot have the information of the b-hadron decay vertex at reco level). Let me know, thanks a lot!

zhangbiao-phy:
> > Hi @fgrosa! Thanks a lot! They looks good to me! Maybe we can also add ct in the lite tree for MC reco and data? we can calculated it by ourself only from full tree.
> 
> Currently I put the `ctGen`, which is computed considering the actual decay length of the c-deuteron (so in case of non-prompt starting from the decay vertex of the b hadron and not the PV) to be able to reweight it to test the impact on the efficiency. But if you prefer I can also add the reco one (which would instead start from the PV regardless the origin of the candidate, since we cannot have the information of the b-hadron decay vertex at reco level). Let me know, thanks a lot!


Thanks! My consideration is that we may need it in the future. Having ctReco in the lite tree would allow us to directly compare data and MC without accessing the full tree. I agree that measuring ct will be difficult with our current statistics, but it would still be useful to have the variable available for checks and possible future studies. Of course we can also add it in the future! so as you prefer ;)

fgrosa:
> > > Hi @fgrosa! Thanks a lot! They looks good to me! Maybe we can also add ct in the lite tree for MC reco and data? we can calculated it by ourself only from full tree.
> > 
> > 
> > Currently I put the `ctGen`, which is computed considering the actual decay length of the c-deuteron (so in case of non-prompt starting from the decay vertex of the b hadron and not the PV) to be able to reweight it to test the impact on the efficiency. But if you prefer I can also add the reco one (which would instead start from the PV regardless the origin of the candidate, since we cannot have the information of the b-hadron decay vertex at reco level). Let me know, thanks a lot!
> 
> Thanks! My consideration is that we may need it in the future. Having ctReco in the lite tree would allow us to directly compare data and MC without accessing the full tree. I agree that measuring ct will be difficult with our current statistics, but it would still be useful to have the variable available for checks and possible future studies. Of course we can also add it in the future! so as you prefer ;)

Added in this commit [1b11593](https://github.com/AliceO2Group/O2Physics/pull/16713/commits/1b11593c406c89225ddbebe5483dfb1a311c4671)!

zhangbiao-phy:
> > > > Hi @fgrosa! Thanks a lot! They looks good to me! Maybe we can also add ct in the lite tree for MC reco and data? we can calculated it by ourself only from full tree.
> > > 
> > > 
> > > Currently I put the `ctGen`, which is computed considering the actual decay length of the c-deuteron (so in case of non-prompt starting from the decay vertex of the b hadron and not the PV) to be able to reweight it to test the impact on the efficiency. But if you prefer I can also add the reco one (which would instead start from the PV regardless the origin of the candidate, since we cannot have the information of the b-hadron decay vertex at reco level). Let me know, thanks a lot!
> > 
> > 
> > Thanks! My consideration is that we may need it in the future. Having ctReco in the lite tree would allow us to directly compare data and MC without accessing the full tree. I agree that measuring ct will be difficult with our current statistics, but it would still be useful to have the variable available for checks and possible future studies. Of course we can also add it in the future! so as you prefer ;)
> 
> Added in this commit [1b11593](https://github.com/AliceO2Group/O2Physics/pull/16713/commits/1b11593c406c89225ddbebe5483dfb1a311c4671)!

Thanks a lot! all good to me!

zhangbiao-phy (approved):

PR #16714, opened 2026-06-17
[PWGCF] Add configurable FB eta-window width and clean warnings

isputows:
This PR makes the FB pseudorapidity window width configurable via `cfgEtaWidth` instead of using fixed η intervals.

The FB windows are now defined by their center positions and a configurable width, allowing systematic studies of the dependence on the η-window size.

In addition, minor code cleanup was performed to address linter and include-cleaner warnings.


alibuild (approved):
Auto-approving on behalf of @isputows.

PR #16716, opened 2026-06-17
[PWGDQ] add computation of chi2-like values from ML score

aferrero2707:
The ML score, which takes values between 0 and 1, is converted to a chi2-like value ranging from zero to infinity with this simple formula:

    chi2 = 1 / score - 1

The chi2-like definition allows to apply certain chi2-based selection, and particulartly the leading/next-to-leading chi2 gap, also in the case of matching performed with ML models.

### DIFF
--- a/PWGDQ/Tasks/qaMatching.cxx
+++ b/PWGDQ/Tasks/qaMatching.cxx
@@ -1999,6 +1999,21 @@ struct QaMatching {
     return attempts;
   }
 
+  template <class TMUON>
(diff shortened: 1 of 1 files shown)


aferrero2707:
@mcoquet642 do you prefer that I fix the linter error (it's an explicit include that is missing, but it is not preventing the correct compilation), or I postpone the fix to the next PR to avoid re-triggering the CI?

mcoquet642 (approved):

vkucera:
Next time, please fix the errors immediately.

PR #16719, opened 2026-06-17
[PWGHF] Implement workaround to for check in MlResponse not compatible with fixed features

fgrosa:
This PR implements a workaround to the check introduced in https://github.com/AliceO2Group/O2Physics/pull/16424 which is not compatible in case of model application with a fixed array of features, not read by a configurable.


alibuild (approved):

PR #16732, opened 2026-06-18
[PWGHF] Add possibility to reapply trigger selections in Xic+ candidate creator

fgrosa:
This PR implements the possibility to reapply the selections applied in the software triggers, since the topological variables in the software trigger are computed using the DCAFitter and only the two pion tracks. This is in particular important for the MC, since software triggers are not simulated.
@fchinu for info


alibuild (approved):

vkucera:
FYI @JaeYoonCHO

PR #16740, opened 2026-06-19
[DPG] tpcSkimsTableCreator: read IR source and sqrtSNN from CCDB; fix nclPID definition

lubynets:
1. IR source is read from CCDB automatically in runtime (using timestamp of the first BC), so there is no need to provide `irSource` configurable anymore;
2. Fixed `nclPID` definition

Tagging @amaringarcia 

_UPD:_
3. sqrtSNN is read from CCDB automatically.
If the Genreal Run Parameters pointer is not available, the irSource is defaulted to empty string (and hadronic rate will be filled with zeros), and sqrtSNN to 5360 GeV.


lubynets:
Convert to draft to add reading the $\sqrt{s_{\mathrm{NN}}}$ from CCDB (if possible).
_UPD:_ done.

alcaliva (approved):

PR #16748, opened 2026-06-19
[PWGCF] Fix triplet event mixing

ariedel-cern:
Fix bug in same and mixed event mixing in triplet process helpers and add mixing qa.


alibuild (approved):

PR #16751, opened 2026-06-20
[PWGCF] Changing PID selection at low-pT

SwatiSaha-1997:
Improve PID selection at low-pT by rejecting ambiguous TPC candidates
- Count the number of particle hypotheses (π, K, p, e) satisfying the TPC nσ criterion.
- Reject tracks compatible with more than one particle species


alibuild (approved):

PR #16752, opened 2026-06-21
[PWGDQ] Adding minitree to table-reader-with-assoc

lucamicheletti93:
Adding minitree to table-reader-with-assoc


alibuild (approved):

PR #16768, opened 2026-06-23
[PWGLF,PWGMM] Change ev and trk selection flags

jmunozme:


vkucera:
@jmunozme Your PR does not compile. Please test it before you open a PR for review.

PWGMM/UE/Tasks/uecharged.cxx
```diff
@@ -13,14 +13,15 @@
 /// \file uecharged.cxx
 /// \brief Underlying event analysis task
 /// \since November 2021
-/// \last update: April 2026
+/// \last update: June 2026
```
vkucera: There is no `\last` Doxygen command.

aortizve (approved):

PR #16770, opened 2026-06-23
[PWGDQ] optimized the computation of match attempts in QA task

aferrero2707:
The computation of the number of match attempts for each MCH track is rather heavy and the code was calling the corresponding function multiple times.
The number of attempts is now computed only once and stored internally.

This gives a nice 3x speedup in the processing time for a typical 2024 Pb-Pb AO2D.

### DIFF
--- a/PWGDQ/Tasks/qaMatching.cxx
+++ b/PWGDQ/Tasks/qaMatching.cxx
@@ -74,6 +74,7 @@
 #include <cstddef>
 #include <cstdint>
 #include <functional>
+#include <limits>
 #include <map>
 #include <memory>
 #include <optional>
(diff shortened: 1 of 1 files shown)


PWGDQ/Tasks/qaMatching.cxx
```diff
@@ -1891,10 +1909,11 @@ struct QaMatching {
                     std::vector<GlobalMuonPair>& globalMuonPairs)
   {
     // outer loop over muon tracks
-    for (const auto& mchIndex1 : collisionInfo.mchTracks) {
-
+    for (const auto& [mchIndex1, mchTrackInfo1] : collisionInfo.mchTracks) {
+      (void)mchTrackInfo1;
```
vkucera: Why do you capture `mchTrackInfo1` if you don't use it?

aferrero2707: @vkucera If preferred, I can iterate over pairs and access the index via `pair.first`.

vkucera: That is much better indeed.

mcoquet642 (approved):

PR #16788, opened 2026-06-24
[DPG] Add flag to select events based on subGenId for HF MC.

mfaggin:
As done in https://github.com/AliceO2Group/O2Physics/blob/master/PWGHF/Tasks/taskMcValidation.cxx#L858C1-L861C6 , this PR adds the possibility to select events in MC analyses based on the subgenerator ID. This is useful to isolate MB gap events in HF-dedicated MC productions.


alibuild (approved):

PR #16801, opened 2026-06-24
[PWGHF] Fix reserve in derived-data creators

vkucera:
- Reserve candidate tables outside the collision loop with the size of the full candidate partition times the number of possible candidate selection hypotheses.
- Do not reserve the (MC)collision tables. (Most of them have no candidates/particles and are not stored.)


alibuild (approved):

PR #16808, opened 2026-06-24
[PWGDQ] event mixing for MFT-MCH candidates

aferrero2707:


mcoquet642:
will be addressed in future PR

PR #16819, opened 2026-06-25
[PWGEM/Dilepton] remove axes from outputs

dsekihat:
[PWGEM/Dilepton] remove axes from outputs


alibuild (approved):

PR #16823, opened 2026-06-25
[Common,Tutorial,PWGJE,PWGUD] DPL Analysis: avoid including ASoA.h in dictionary generation

aalkin:
Various PWGs create ROOT dictionaries for some of their classes, which, unfortunately, through their inclusion chains pull ASoA.h and expose it to rootcling. ROOT's internal llvm is outdated and thus each time we need to update ASoA.h with new code, we have to employ workarounds in oder to keep O2Physics functional. This is highly inconvenient, and thus there will be an update in O2 that will explicitly forbid exposing ASoA.h to rootcling - the compilation will fail with an error. This PR removes all the instances of ASoA.h inclusion in dictionaries, that have been exposed by the upcoming O2 change.

This is a suggested change, the final fix may be different. Preferably, the dictionaries that are not required should be simply removed from CMakeLists.txt. From the codeowners, we request reviewing these changes, suggesting improvements, if needed, and validating that the updated code does not affect the analysis.

requires https://github.com/AliceO2Group/AliceO2/pull/15557

### DIFF
--- a/Common/Core/fwdtrackUtilities.h
+++ b/Common/Core/fwdtrackUtilities.h
@@ -18,7 +18,6 @@
 #ifndef COMMON_CORE_FWDTRACKUTILITIES_H_
 #define COMMON_CORE_FWDTRACKUTILITIES_H_
 
-#include <Framework/AnalysisDataModel.h>
 #include <Framework/DataTypes.h>
 #include <GlobalTracking/MatchGlobalFwd.h>
 #include <MCHTracking/TrackExtrap.h>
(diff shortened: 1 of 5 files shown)


ktf:
@aalkin very nice! It looks like you forgot to add one file, though? Maybe we should split this by PWG so that the approval is faster?

aalkin:
@ktf most of those are dependent on code from one another, unfortunately.

iarsene (dismissed):

iarsene:
Hi @aalkin, @ktf,
After a short chat with Anton, we agreed that I will have a look regarding the DQ changes. I think that we do not really need the dictionaries and maybe most of these changes are not needed. I will test and comunicate with Anton on this.
Ionut

ktf:
Perfect, thanks.

mhemmer-cern:
Hello @aalkin, @ktf,
Thanks for the ongoing work on the framework.
For the PhotonMeson part of the EM I removed the dictionary generation after talking with Anton via private channels. The PR went just up and should be merged soon.
Cheers,
Marvin

rolavick (approved):

aalkin:
Closing if favor of separate PRs:
https://github.com/AliceO2Group/O2Physics/pull/16876
https://github.com/AliceO2Group/O2Physics/pull/16877
https://github.com/AliceO2Group/O2Physics/pull/16878
https://github.com/AliceO2Group/O2Physics/pull/16879

PR #16838, opened 2026-06-26
[PWGDQ] Removing ROOT dictionaries

mcoquet642:
@iarsene @aalkin @ktf as suggested in [PR#16823](https://github.com/AliceO2Group/O2Physics/pull/16823)


PWGDQ/Core/AnalysisCompositeCut.h
```diff
@@ -28,16 +28,16 @@
 class AnalysisCompositeCut : public AnalysisCut
 {
  public:
-  AnalysisCompositeCut(bool useAND = kTRUE);
+  explicit AnalysisCompositeCut(bool useAND = kTRUE);
```
aalkin: If you use this class as a data member in an analysis task, the explicit constructor will probably break the decomposition that is used to parse the task's content.

mcoquet642: Indeed ... I removed the explicit, thanks

PWGDQ/Core/AnalysisCut.cxx
aalkin, line 65: Not really related to this PR, but since these classes are probably not used outside of O2Physics, it makes sense to use framework log facilities instead of iostream.

iarsene (approved):

PWGDQ/Core/AnalysisCut.cxx
vkucera: As reported by the O2 linter...

PR #16842, opened 2026-06-27
[PWGJE] Refactor collision index handling in diffWakeTreeProducer

wirthni:
Ci is added, because the produced tree is used out of an O2 analysis framework (bare root) and a direct correspondence between collisions and tracks is needed


PWGJE/TableProducer/diffWakeTreeProducer.cxx
```diff
@@ -67,7 +65,7 @@ namespace o2::aod
 namespace testcol
 {
 // Event properties
-// DECLARE_SOA_COLUMN(Gi, gi, int64_t);
+DECLARE_SOA_COLUMN(Ci, ci, int64_t);         // Collision index - needed because is used as tree
```
fjonasALICE: I think you should declare here an SOA_INDEX_COLUMN instead (the same as for the tracks) , which will allow you to do the matching also offline. It will simply appear as a int64_t bracnh outside of o2

fjonasALICE (dismissed):

nzardosh (approved):

nzardosh (approved):

PR #16848, opened 2026-06-28
[PWGLF] Apply missing cuts before filling dE/dx vs momentum histograms

PaolaVT:
-multiplicityPt.cxx:
The dE/dx vs. momentum histograms were being filled before the required track selection cuts. The filling order has been corrected so that only tracks passing all cuts contribute to the histograms. As well as, the unused CentClasses member was deleted.

-dedxPidAnalysis.cxx:
 The fphiCutLow/fphiCutHigh arguments from passedPhiCutPri and passedPhiCutSecondaries were removed and struct members were used directly instead.


maciacco (approved):

PWGLF/Tasks/Nuspex/dedxPidAnalysis.cxx
```diff
@@ -43,13 +43,15 @@
...
 #include <TH3.h>
 #include <TString.h>
 
 #include <cmath>
 #include <cstdint>
 #include <memory>
-#include <string_view>
+#include <string>
+// #include <string_view>
 #include <vector>
```
vkucera, line 55: Don't put comments in the includes.

PR #16882, opened 2026-06-30
[PWGLF] added NUA fill and use

prottayCMT:


romainschotter (changes requested):
Hi @prottayCMT ! The PR looks fine but can you please fix the O2 linter error related to `Use constant references for non-modified iterators in range-based for loops.`? Thank you very much!

romainschotter (approved):

PR #16886, opened 2026-07-01
[ALICE3] Pass trackparcov as reference instead of copy in a3 track utilities

jesgum:


ALICE3/Core/TrackUtilities.cxx
```diff
@@ -48,14 +48,14 @@ void o2::upgrade::convertTLorentzVectorToO2Track(const int charge,
   new (&o2track)(o2::track::TrackParCov)(x, particle.Phi(), params, covm);
```
vkucera, line 48: What is the rationale behind this memory allocation?

jesgum: That line is not written by me so I'm not sure

alibuild (approved):

PR #16916, opened 2026-07-03
[PWGCF] add some particle cuts and event cuts

pengchon:


abilandz (approved):
Please address the remaining errors and warnings, which are not mandatory now, alongside next PR - approving this PR so you can proceed with more urgent checks in the analysis itself.

PR #16922, opened 2026-07-03
[PWGLF] Autoset the tables produced based on the task topology

njacazio:


vkucera (changes requested):
@njacazio Please fix the bugs.

njacazio:
I did not realise this was still open, thanks!

vkucera:
> I did not realise this was still open, thanks!

What about the bugs?

PR #16938, opened 2026-07-06
[PWGCF] Fixes in femto framework

ariedel-cern:
- Fix minor bugs in V0 and cascade selection
- Fix namespace across the full framework
- Implement first batch of warnings indicated by the new clang-tidy check

### DIFF
--- a/PWGCF/Femto/Core/partitions.h
+++ b/PWGCF/Femto/Core/partitions.h
@@ -17,157 +17,169 @@
 #define PWGCF_FEMTO_CORE_PARTITIONS_H_
 
 // collsion selection
-#define MAKE_COLLISION_FILTER(selection)                                                                            \
-  (o2::aod::femtocollisions::posZ >= selection.vtxZMin && o2::aod::femtocollisions::posZ <= selection.vtxZMax) &&   \
-    (o2::aod::femtocollisions::mult >= selection.multMin && o2::aod::femtocollisions::mult <= selection.multMax) && \
-    (o2::aod::femtocollisions::cent >= selection.centMin && o2::aod::femtocollisions::cent <= selection.centMax) && \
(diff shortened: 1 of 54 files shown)


alibuild (approved):

PWGCF/Femto/Core/partitions.h
```diff
@@ -17,157 +17,169 @@
...
 
 // collsion selection
-#define MAKE_COLLISION_FILTER(selection)                                                                            \
-  (o2::aod::femtocollisions::posZ >= selection.vtxZMin && o2::aod::femtocollisions::posZ <= selection.vtxZMax) &&   \
-    (o2::aod::femtocollisions::mult >= selection.multMin && o2::aod::femtocollisions::mult <= selection.multMax) && \
-    (o2::aod::femtocollisions::cent >= selection.centMin && o2::aod::femtocollisions::cent <= selection.centMax) && \
-    (o2::aod::femtocollisions::magField >= o2::framework::expressions::as<int8_t>(selection.magFieldMin) &&         \
-     o2::aod::femtocollisions::magField <= o2::framework::expressions::as<int8_t>(selection.magFieldMax)) &&        \
-    ncheckbit(o2::aod::femtocollisions::mask, selection.collisionMask)
+// NOLINTNEXTLINE(cppcoreguidelines-macro-usage)
```
vkucera, line 20: Why did you disable the check?

ariedel-cern: Hi Vit, I found that using macros is a convenient way to reuse the same partitions and filters across different files. Without suppressing the check, clang-tidy warns that the macro should be replaced with a constexpr template function. I am not sure that is actually possible in this case (I tried a few approaches, but I couldn't get the code to compile). If you know of a way to implement this without macros, please let me know, and I will replace them with functions.

vkucera: Hi Anton, indeed. In your case, it probably makes sense. Thanks for the clarification.

PR #16939, opened 2026-07-06
[Common] Add light ion centrality tables with anchoring at 90%

jesgum:
This PR adds two new centrality tables, specifically meant for cross checks for light ions.
The approach here is the same approach as in Pb--Pb, where we fit with Glauber+NBD, anchor at 90%, use data for 0-90% and Glauber+NBD for 90-100%. One table is added for a collision based calibration and one table for a bc based calibration. 

Tagging @ddobrigk


dsekihat (approved):

vkucera:
Marked as stale due to 30 days of no activity.

dsekihat (approved):

PR #16947, opened 2026-07-07
[ALICE3] Fix production time for daughters from cascade decays in a3 decayer

jesgum:


alibuild (approved):

vkucera:
@jesgum What is the strategy to fix the bugs and other errors?

PR #16948, opened 2026-07-07
[PWGHF] Fix OmegaC selector

fcatalan92:
Properly reject candidates when ML selection is not satisfied.

@mtorresc15


alibuild (approved):

PR #16871, opened 2026-06-30
[PWGHF] Adapt status code checks for c-deuteron resonant decays

fgrosa:

PR #16953, opened 2026-07-07
[PWGHF] Fix MC flagging for c-deuteron

fgrosa:
@zhangbiao-phy

### DIFF
--- a/PWGHF/Utils/utilsMcMatching.h
+++ b/PWGHF/Utils/utilsMcMatching.h
@@ -160,9 +160,9 @@ static const std::unordered_map<DecayChannelMain, const std::vector<int>> daught
   {DecayChannelMain::CDeuteronToDeKPi, {+o2::constants::physics::Pdg::kDeuteron, +PDG_t::kKMinus, +PDG_t::kPiPlus}}};
 
 /// resonances in c-deuteron decay are not stored in the particle stack for c-deuteron, but tagged with specific status codes
-static const std::array<int, 2> statusCodeCDeuteronToDeKstar0{-85, -95};
-static const std::array<int, 2> statusCodeCDeuteronToNeDeltaplusK{-86, -96};
-static const std::array<int, 2> statusCodeCDeuteronToNeL1520Pi{-87, -97};
+static constexpr std::array<int, 2> statusCodeCDeuteronToDeKstar0{-85, -95};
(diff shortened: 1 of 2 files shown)


vkucera:
@fgrosa Please fix the warnings introduced in https://github.com/AliceO2Group/O2Physics/pull/16871

PWGHF/Utils/utilsMcMatching.h
```diff
@@ -160,9 +160,9 @@ static const std::unordered_map<DecayChannelMain, const std::vector<int>> daught
   {DecayChannelMain::CDeuteronToDeKPi, {+o2::constants::physics::Pdg::kDeuteron, +PDG_t::kKMinus, +PDG_t::kPiPlus}}};
 
 /// resonances in c-deuteron decay are not stored in the particle stack for c-deuteron, but tagged with specific status codes
-static const std::array<int, 2> StatusCodeCDeuteronToDeKstar0{-85, -95};
-static const std::array<int, 2> StatusCodeCDeuteronToNeDeltaplusK{-86, -96};
-static const std::array<int, 2> StatusCodeCDeuteronToNeL1520Pi{-87, -97};
+static const std::array<int, 2> statusCodeCDeuteronToDeKstar0{-85, -95};
+static const std::array<int, 2> statusCodeCDeuteronToNeDeltaplusK{-86, -96};
+static const std::array<int, 2> statusCodeCDeuteronToNeL1520Pi{-87, -97};
```
vkucera: These used to be `constexpr` constants. Why are they not any more?

fgrosa: I thought it was not necessary to have them `constexpr` in first place, but if you want I can revert them. What about all the other variables that are `const` (the maps)? Is there any advantage of having them `constexpr`?

vkucera: > I thought it was not necessary to have them `constexpr` in first place, but if you want I can revert them.

Yes, please.

> What about all the other variables that are `const` (the maps)? Is there any advantage of having them `constexpr`?

I don't remember if they can be `constexpr` but if yes, it would be better for compile-time checks and optimisations. I will check later.

fgrosa: Apparently `unordered_map` cannot be `constexpr` in C++17 / C++20 / C++23

alibuild (approved):

PR #16958, opened 2026-07-08
[PWGCF] Update cascade processing

ariedel-cern:
In this PR:
- add a process function for tracks, v0s, and cascades for PbPb processing
- add flags so rejection of competing mass hypotheses in cascade builder (Xi, Omega) can be turned off.


alibuild (approved):

PR #16965, opened 2026-07-08
[PWGCF] Speed up triplet mixing in femto framework

ariedel-cern:
Avoid unnecessary particle slicing during mixed-event processing in triplet tasks.


alibuild (approved):

PR #16974, opened 2026-07-08
[PWGUD] Fix wrong offset

aalkin:
@rolavick @amatyja @siragoni 
It appears that I indeed made a mistake in #16790 - `rawSlice()` requires start and end, rather than start and size. Could you please verify this change in analysis and confirm that it is enough?


amatyja (approved):
Dear Anton, thank you for the correction. We will check it after merging. Thanks.

PR #16978, opened 2026-07-09
[PWGCF] Add spherical harmonics decomposition to unified Femto framework

IgorPtak:


PWGCF/Femto/Core/pairHistManager.h
```diff
@@ -941,6 +965,66 @@ class PairHistManager
...
+          dir += "/";
+          dir += ktFolder;
+          dir += "/";
+
+          int ihist = 0;
+          for (int l = 0; l <= mShLMax; ++l) {
+            for (int m = -l; m <= l; ++m) {
+              std::string lm = std::to_string(l);
+              lm += (m < 0) ? std::to_string(l - m) : std::to_string(m);
+              std::string nameRe = dir;
```
ariedel-cern, line 1003: I think it would be good to include folders for the histograms into the existing structure under {SE/ME}/{Analysis/MC}/

You should be able to just append std::string(prefix) to the title.

PWGCF/Femto/Core/pairHistManager.h
```diff
@@ -941,6 +965,66 @@ class PairHistManager
...
+              nameRe += "ReYlm";
+              nameRe += lm;
+              std::string nameIm = dir;
```
ariedel-cern, line 1006: Same here as above

ariedel-cern:
Hi @IgorPtak ,
Many thanks for the implementation. The code looks good to me (I understand why you want to allocate the histograms for the decomposition dynamically, this is fine with me). As mentioned above, could you adjust the naming of the histograms

PWGCF/Femto/Core/pairHistManager.h
IgorPtak: Do you mean I shouldn't add the extra SH/mult/kt subfolders, since SH is already placed under {SE/ME}/{Analysis/MC}/?

IgorPtak: Nvm, I think I see what you mean now.

ariedel-cern (approved):

PR #16980, opened 2026-07-09
[PWGCF] Fix process guard in femto producer

ariedel-cern:
Fix process guard in femto producer


alibuild (approved):

PR #16981, opened 2026-07-09
[PWGJE] Enhance GammaJetAnalysisTree with additional MC information

fjonasALICE:
- Extend GjGammaMCInfos with generator-level kinematics (eta, phi, energy, pt) and MC isolation for the particle associated with each reconstructed cluster.
- Update getClusterOrigin to return both the origin bitmap and the matched MC particle index (including special handling for merged π⁰/η and converted photons).
- Add a new GjMCJetSubstructures table and fill  substructure for particle-level jets when substructure calculation is enabled.
- Consolidate PROCESS_SWITCH definitions at the end of the task with explicit ordering comments


jaimenorman:
+1

jaimenorman (approved):

PR #16996, opened 2026-07-09
[PWGLF] Compile headers

vkucera:
Provides missing compile commands needed by Clang-Tidy.

### DIFF
--- a/PWGLF/Utils/CMakeLists.txt
+++ b/PWGLF/Utils/CMakeLists.txt
@@ -24,3 +24,15 @@ o2physics_add_library(collisionCutsGroup
 o2physics_target_root_dictionary(collisionCutsGroup
     HEADERS collisionCutsGroup.h
     LINKDEF collisionCutsGroupLinkDef.h)
+
+o2physics_add_library(LfDecay3bodyBuilderHelper
+    SOURCES decay3bodyBuilderHelper.cxx
+    PUBLIC_LINK_LIBRARIES O2Physics::AnalysisCore KFParticle::KFParticle)
+
+o2physics_add_library(LfStrangenessBuilderHelper
+    SOURCES strangenessBuilderHelper.cxx
+    PUBLIC_LINK_LIBRARIES O2Physics::AnalysisCore KFParticle::KFParticle)
+
+o2physics_add_library(LfStrangenessBuilderModule
+    SOURCES strangenessBuilderModule.cxx
+    PUBLIC_LINK_LIBRARIES O2Physics::AnalysisCore KFParticle::KFParticle)
--- a/PWGLF/Utils/decay3bodyBuilderHelper.cxx
+++ b/PWGLF/Utils/decay3bodyBuilderHelper.cxx
@@ -14,4 +14,5 @@
 ///
 /// \author Vít Kučera <vit.kucera@cern.ch>, Inha University
 
+// o2-linter: disable=name/workflow-file (not a workflow file)
 #include "decay3bodyBuilderHelper.h" // IWYU pragma: keep
--- a/PWGLF/Utils/decay3bodyBuilderHelper.cxx
+++ b/PWGLF/Utils/decay3bodyBuilderHelper.cxx
@@ -1,0 +1,17 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+/// \file decay3bodyBuilderHelper.cxx
+/// \brief Helper file to provide compilation command for headers.
+///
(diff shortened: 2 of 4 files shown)


romainschotter (changes requested):
Hi @vkucera !
I'm not sure I understand the rationale behind these changes. Could you please elaborate? 
In particular for the updates to the \author entries, my understanding is that the \author field is generally intended to reflect the primary authors/contributors/maintainers of the code, i.e. people who could be contacted for clarification of the code, maintenance, debugging, etc.
Thank you very much!

vkucera:
Hi @romainschotter 

> Hi @vkucera ! I'm not sure I understand the rationale behind these changes. Could you please elaborate?

It's written in the description. What would you like me to elaborate?

> In particular for the updates to the \author entries, my understanding is that the \author field is generally intended to reflect the primary authors/contributors/maintainers of the code, i.e. people who could be contacted for clarification of the code, maintenance, debugging, etc.

Yes. What is the problem?

romainschotter:
Hi @vkucera !
Thank you very much for the clarification.

> > In particular for the updates to the \author entries, my understanding is that the \author field is generally intended to reflect the primary authors/contributors/maintainers of the code, i.e. people who could be contacted for clarification of the code, maintenance, debugging, etc.
> 
> Yes. What is the problem?

The issue is precisely whether you are the primary author, contributor, and/or maintainer of the `3bodydecaybuilder` and `strangenessbuilder`. If not, then the \author field should at least include the original authors of those files, i.e. Carolina Anna Jauch and David Chinellato respectively, who are also the primary author, contributor and maintainer. Replacing their names with your own does not accurately reflect the authorship and maintenance of these files, and removes the attribution to the people who originally developed and have maintained them.  Alternatively, since these are utility tasks used by the central tasks, one could put "ALICE Collaboration" as \author field (as it is sometimes done)

Similarly, in my understanding, the \brief field should describe the purpose of the helper or module itself, which is not `Helper file to provide compilation command for headers.`

Could you please update these accordingly?

vkucera:
Hi @romainschotter , I think there is a misunderstanding. Please check the PR carefully. I am not modifying any C++ files or replacing author names. I am adding new files, therefore I am the contributor, and their purpose is exactly what the `\brief` field states.

romainschotter:
Hi @vkucera !
Apologies. Somehow, I read that you modified the header files while you actually added cxx files. My mistake
Nevertheless, I think you should still consider putting the primary authors in the \author entry of the cxx files. Otherwise it might create some confusion as people will naturally assume that the cxx author is the same as the header.
Thank you very much!

vkucera:
> I think you should still consider putting the primary authors in the \author entry of the cxx files. Otherwise it might create some confusion as people will naturally assume that the cxx author is the same as the header. Thank you very much!

Sorry, I suspect that what you suggest will have exactly the opposite effect of what you want. Are you suggesting that if the header authors are not mentioned in the cxx files, people will assume that they are authors of them? And to prevent people from assuming that they are authors of the cxx files, you suggest to mention them there as authors???

romainschotter:
Now it is me who is confused :-D

What I meant is that we should keep the \author entries consistent between the header and the corresponding cxx files to avoid confusion. 

Otherwise, someone looking at the implementation of `3bodydecaybuilder` or `strangenessbuilder` in the cxx files might reasonably conclude that you are the primary contact for those modules because your name appears as the sole \author. If they have questions, need clarifications, want to report a bug, or plan further developments, they would naturally contact you instead of the actual maintainers of those modules, namely the authors listed in the corresponding header files.

vkucera:
Hi @romainschotter , thanks for the clarification but I'm afraid it still does not make any sense to me.

- The author information should be true first. What you suggest makes it false.
- I don't want to be mentioned as an author of a file that I never touched. Why do you assume that the header authors do?
- And why do you think that adding wrong author information creates less confusion?
- The cxx files do not contain any implementation information. I will argue that any reasonable person who is interested in the code will end up opening the header and read the author information there to figure out the relevant people.
- Even if someone opens the cxx file, they will quickly understand from its emptiness and from the `\brief` that there is nothing interesting there and will move on to the header.
- And if they want to ask about the cxx file, they should ask me, not the header authors.

romainschotter:
Hi @vkucera,

How about the following compromise?

We add the primary authors and maintainers of the module in the \author field of the corresponding cxx files, and explicitly clarify that questions about the cxx files themselves should be directed to you, while questions about the module, its implementation, maintenance, or future developments should be directed to the primary authors of the module.

In that way, the distinction is explicit: people interested in the module know whom to contact, while your contribution is also properly acknowledged.

What do you think?

vkucera:
Hi @romainschotter , I genuinely don't understand what problem you are trying to solve here. 

If you want explicit distinction of contact people for respective files and you agree that the header authors are contacts for the headers and I am contact for the cxx files, that's exactly what the current PR follows and that's exactly what any literate person will see when they open any of the files. Where is the problem?

Instead, you keep suggesting adding wrong documentation information. Why? That is not a compromise. That is just misinformation.

vkucera:
@romainschotter Please either give a valid reason for blocking this PR or merge it.

romainschotter:
Hi @vkucera !

You can consider this approved from my side. 
I will clarify the author documentation separately at a later stage.

Thank you very much.

romainschotter (approved):

vkucera:
> Hi @vkucera !
> 
> You can consider this approved from my side. I will clarify the author documentation separately at a later stage.
> 
> Thank you very much.

Thanks

PR #17003, opened 2026-07-10
[PWGLF] Update KF cascade building in strangenessBuilderHelper

creetz16:
Update LF strangenessbuilderHelper KF cascade building: 
- store unconstrained V0 and cascade masses in StoredKFCascCores table in all cases

Hi @ddobrigk and @romainschotter :) 
I noticed that in the KF cascade building I had implemented the computation of the V0 and cascade masses in a way that the constrained masses (PDG masses) are stored in the StoredKFCascCores table in case the mass constraints are enabled. On the one hand this makes sense from the point of view of storing consistently the constrained particle parameters (mom, E, mass) and covariance matrix. But on the other hand it is essential to store the unconstrained invariant mass to for example apply selections on it. In our case of the Xic to Xipipi analysis of triggered data with the KF we indeed need the unconstrained V0 and cascade masses while using the constrained parameters.
I therefore updated the strangenessBuilderHelper accordingly in this PR. I would be happy to have your thoughts on this, maybe there is a better way to do it. :)
I am not aware of any other analysis using the KF cascade building. Do you know of anyone whom we should include in this discussion/update?

### DIFF
--- a/PWGLF/Utils/strangenessBuilderHelper.h
+++ b/PWGLF/Utils/strangenessBuilderHelper.h
@@ -1059,6 +1059,11 @@ class strangenessBuilderHelper
       return false;
     }
 
+    // Calculate V0 mass before mass constraint
+    float MLambda, SigmaLambda;
+    KFV0.GetMass(MLambda, SigmaLambda);
+    cascade.kfMLambda = MLambda;
+
     if (kfUseV0MassConstraint) {
       KFV0.SetNonlinearMassConstraint(o2::constants::physics::MassLambda);
     }
@@ -1117,9 +1122,20 @@ class strangenessBuilderHelper
       cascade = {};
       return false;
     }
+
+    // Calculate masses a priori before potentially applying mass constraint
+    // --> this is the invariant mass of the decay products, not the mass-constrained value
+    float MXi, SigmaXi, MOmega, SigmaOmega;
+    KFXi.GetMass(MXi, SigmaXi);
+    KFOmega.GetMass(MOmega, SigmaOmega);
+    cascade.massXi = MXi;
+    cascade.massOmega = MOmega;
+
     if (kfUseCascadeMassConstraint) {
       // set mass constraint if requested
       // WARNING: this is only adequate for decay chains, i.e. XiC -> Xi or OmegaC -> Omega
+      // WARNING: be aware of the fact that if enabled, the stored particle momentum and covariance matrix will be mass-constrained
+      // while the stored mass will be the invariant mass of the decay products (i.e. not mass-constrained)
       KFXi.SetNonlinearMassConstraint(o2::constants::physics::MassXiMinus);
       KFOmega.SetNonlinearMassConstraint(o2::constants::physics::MassOmegaMinus);
     }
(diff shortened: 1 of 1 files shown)


romainschotter:
Hi @creetz16 !
I had a look at which tasks are using the KFCascCores or the KFCascDatas, and here is what I found:
- `PWGLF/Tasks/QC/kfStrangenessStudy.cxx`,
- `PWGLF/Tasks/QC/kfPerformanceStudy.cxx`,
- `PWGHF/TableProducer/candidateCreatorXicToXiPiPi.cxx`
- `PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx`.

I think there are two possible strategies:
1) store the unconstrained masses in the `MXi, `MOmega`, `MLambda` columns and obtain the constrained masses either by calculating them manually or through dynamic columns using the constrained momentum information. The downside is that the tasks listed above would need to be updated accordingly.
2) use the `KFToCascRefs` table (which an interlink table `KFCascCores` --> `CascCores`) to de-reference the `KFCascCore` into a `CascCore` and access the unconstrained information directly. A similar approach could also be implemented using `KFCascDataLink` and `CascDataLink`, although it would only work on original data and would require a few extra steps than the simple de-referencing.

Both approaches work but I personally find the second one more practical, since it keeps the constrained and unconstrained information clearly separated while still making both easily accessible. However, it requires a tiny development (and some testing) on our side regarding this process function (https://github.com/AliceO2Group/O2Physics/blob/9b9fa30ec1aea5b3bc68486dbae381e9b241c03b/PWGLF/Utils/strangenessBuilderModule.h#L828) if one wants to use the `KFToCascRefs`.

Maybe @ddobrigk or you have a different opinion, so please let me know what you think ! :-)

In the meantime, I'll start implementing the second approach. It will most likely be next week, though, as I won't have enough time to implement and test it over the weekend :cry: )

Thank you very much!

creetz16:
Hi @romainschotter 

Many thanks for having a look!! This is not urgent, so please don't worry! :)
Getting the constrained mass is not difficult at all since we are constraining to the PDG mass and therefore it will simply be a delta function at the PDG mass value.
But I agree that keeping the constrained and unconstrained information separate is for sure the "safest" way. I chose here the easy way in a sense. ;) 
`PWGLF/Tasks/QC/kfStrangenessStudy.cxx` and `PWGLF/Tasks/QC/kfPerformanceStudy.cxx` are QA tasks that David and I wrote some time ago to compare cascades built with the DCA fitter and with the KF. `PWGHF/TableProducer/candidateCreatorXicToXiPiPi.cxx` is our HF analysis code for the Xic to Xipipi reconstruction where I am one of the analyser (at least for the KF part of the code). Indeed, I forgot about the Xic0 task `PWGHF/TableProducer/candidateCreatorXic0Omegac0.cxx`. I am not sure if anyone is using it for analysis at the moment but it is implemented in a way that it is supposed to be run without any mass constraints in the strangenessbuilder. The mass constraints are only set in the candidateCreatorXic0Omegac0 itself. 
I will also look into your second suggestion.

Thanks again for having a look. I would be happy to discuss next week at some point. :)

vkucera:
@creetz16 Please fix the missing initialisations.

romainschotter (approved):

vkucera:
@romainschotter What is the plan for fixing the bugs?

PR #17011, opened 2026-07-10
[PWGEM] update trigger analyses

dsekihat:
[PWGEM] update trigger analyses
This PR is necessary, because zorro makes many CCDB calls when zorro.isSelected is used on derived data. This leads interruption of hyperloop trains. After this PR merged, EMSWTAT and EMSWTATOI must be stored in EM's derived data for trigger analyses.
This PR removes remnant dependency of swt in the photon directory on the dilepton directory.


alibuild (approved):

PR #17034, opened 2026-07-13
[PWGHF] Implement checks for correlated backgrounds and for B->J/Psi(->ee)X decays in B->J/Psi workflow

fchinu:


PWGHF/D2H/TableProducer/dataCreatorJpsiHadReduced.cxx
```diff
@@ -206,9 +217,19 @@ struct HfDataCreatorJpsiHadReduced {
...
+    }
+    if (doprocessJpsiPhiData && doprocessJpsiPhiMc) {
+      LOGP(fatal, "Cannot enable both J/Psi-Phi data and MC processing at the same time, please fix your configuration!");
+    }
+
     std::array<bool, 6> doProcess = {doprocessJpsiKData, doprocessJpsiKMc, doprocessJpsiK0StarData, doprocessJpsiK0StarMc, doprocessJpsiPhiData, doprocessJpsiPhiMc};
-    if (std::accumulate(doProcess.begin(), doProcess.end(), 0) != 1) {
-      LOGP(fatal, "One and only one process function can be enabled at a time, please fix your configuration!");
+    if (std::accumulate(doProcess.begin(), doProcess.end(), 0) == 0) {
+      LOGP(fatal, "Enable at least one process function, please fix your configuration!");
```
fgrosa, line 235: Is more than one process function enabled allowed?

PWGHF/D2H/TableProducer/dataCreatorJpsiHadReduced.cxx
```diff
@@ -504,6 +525,65 @@ struct HfDataCreatorJpsiHadReduced {
...
+                  if (std::abs(particleProng0.pdgCode()) == kMuonMinus && std::abs(particleProng1.pdgCode()) == kMuonMinus) {
+                    flag = partlyRecoDecayMapMuMu[std::abs(pdgCodeBeautyMother)];
+                  } else if (std::abs(particleProng0.pdgCode()) == kElectron && std::abs(particleProng1.pdgCode()) == kElectron) {
+                    flag = partlyRecoDecayMapEE[std::abs(pdgCodeBeautyMother)];
+                  }
+                  break;
+                }
+              }
+            }
+          }
```
fgrosa, line 593: In principle this could be done more easily with the function https://github.com/AliceO2Group/O2Physics/blob/13b7152150042efb4eea55e103744e7a364195fc/Common/Core/RecoDecay.h#L733-L734 
enabling the `acceptIncompleteReco` flag, or am I missing something?
It can also be modified in a subsequent PR in case

PWGHF/D2H/TableProducer/dataCreatorJpsiHadReduced.cxx
```diff
@@ -504,6 +525,65 @@ struct HfDataCreatorJpsiHadReduced {
           checkWrongCollision(particleMother, collision, indexCollisionMaxNumContrib, flagWrongCollision);
         }
       }
+
+      // additional checks for correlated backgrounds
+      if (checkDecayTypeMc) {
+        // B+ → J/Psi K+ → (e+e-) K+
+        if (!flag) {
+          indexRec = RecoDecay::getMatchedMCRec<true, false, false, true, true>(particlesMc, std::array{vecDaughtersB[0], vecDaughtersB[1], vecDaughtersB[2]}, Pdg::kBPlus, std::array{-kElectron, +kElectron, +kKPlus}, true, &sign, 3);
```
fgrosa: This is the only important comment: we have to accept incomplete decays, because we often have a photon emitted which ends up in the decay chain (in particular for the EE channel, but also with MuMu if I am not mistaken)

fgrosa (changes requested):
Thanks @fchinu! A couple of comments below

PWGHF/D2H/TableProducer/dataCreatorJpsiHadReduced.cxx
fchinu: Yes, I have removed the restriction. Now you can enable more than one process function (though never both data and MC for the same particle)

PWGHF/D2H/TableProducer/dataCreatorJpsiHadReduced.cxx
fchinu: Indeed, thanks! One needs to require that the other particle stems from the same mother as of the J/Psi. I have implemented it

fchinu:
Hi @fgrosa, thanks a lot for your review! I have implemented your comments, please let me know if you spot any other issues in the new code :)

fgrosa:
Thanks @fchinu it looks good to me now!

fgrosa (approved):

PR #17036, opened 2026-07-13
[PWGLF] added event selection options

GiorgioAlbertoLucia:


maciacco (approved):

PWGLF/TableProducer/QC/nucleiQC.cxx
```diff
@@ -52,6 +52,7 @@
 
 #include <fmt/format.h>
 
+#include <algorithm>
```
vkucera, line 55: Why?

GiorgioAlbertoLucia: MegaLinter requires it for min, I tried to remove it in a new PR and it complained

PR #17038, opened 2026-07-13
[ALICE3] alice3 dilepton: machinery for study with LUTS based on ACTS

rbailhac:
alice3 dilepton: machinery for study with LUTS based on ACTS, making further smearing and PID efficiency based on histograms in ccdb


rbailhac:
Hi @njacazio , 
here is a pull request to add the possibility for the ALICE 3 dielectron studies to run with LUTs based on ACTS with in addition a after-burner to better smear the momentum according to ACTS (based on 2D maps in the ccdb) and the possibility to add from outside PID efficiencies (2D histos in the ccdb) based on full simulation. 

Could you have a look maybe ?  Am I supposed to fix all the output of the checked etc... ? I tried to reduce them a bit but there are still some.

Thanks,
Raphaelle

njacazio (approved):

vkucera:
@rbailhac @njacazio You reverted my fixes of includes despite being explicitly warned about it by the error messages from the code-check! Fix it asap!

PR #17040, opened 2026-07-13
[PWGEM] Rewise phi-star separation cuts

Steffimro:
Same-sign conversion legs merging in the TPC deplete the SE pair distribution at low q_inv. This PR adds a separation variable u (Delta eta and signed Delta phi* averaged over the nine standard TPC radii, min over both same-sign leg pairs), filled vs q_inv and k_T in SE and ME to derive a data-driven cut; the cut itself is off by default.


mhemmer-cern (approved):
LGTM
A commit that fixes the code-checker warnings and errors as a follow up please.

PR #17043, opened 2026-07-13
[PWGCF] massive fix for O2linter + minor fix for vs M calculus in Internal Va…

abilandz:
…lidation + added new eff.


vkucera:
Hi @abilandz , thanks a lot for the fixes!
Please note that you might see incorrect error messages in results of the code-check because the headers are not self-contained and cannot be compiled and analysed properly by Clang-Tidy.

abilandz:
> Hi @abilandz , thanks a lot for the fixes! Please note that you might see incorrect error messages in results of the code-check because the headers are not self-contained and cannot be compiled and analysed properly by Clang-Tidy.

Hi @vkucera , thanks for the feedback, noted! Per our previous conversation, I will inline the content of those temporary include headers when I am done with the major code development. 

On a side note, and related to O2linter, please can you comment on:

1. `Error: Do not use TDatabasePDG directly. Use o2::constants::physics::Mass... or Service<o2::framework::O2DatabasePDG> instead.` => I use TDatabasePDG, because otherwise there is an unnecessary memory blow-up when I have O2DatabasePDG in the code, but do not use it in a particular analysis. On the other hand, I can book TDatabaseDPG object per request. There was a discussion about this some time ago, but I didn't follow up to see whether there were any changes in O2DatabasePDG related to it (it was related back then to the missing lazy initialization of the service O2DatabasePDG)
2. `Error: Avoid hard-coded PDG codes. Use named values from PDG_t or o2::constants::physics::Pdg instead.` => Nothing is hard-coded here related to PDG code, these are merely array indices

I might have more questions later, if I do not manage to handle all the MegaLinter thingies. 

Btw, fixing O2Linter naming conventions triggered a lot of MegaLinter warnings related to variable name shadowing... Is it really such a problem to have an example code snippet:

`int var = Var(); // get the content of variable 'var', from the corresponding function Var()`

After O2Linter intervention and renaming the function Var() to var(), this runs into a shadowing problem...

Thanks a lot for all the feedback!

abilandz:
@vkucera : Please can you also comment on this MegaLinter failure, i think this shall be OK:

```
if(!ptr) { 
  LOGF(fatal, ...); 
}
ptr->DoSomething();
``` 

The `if` statement is not redundant => namely if ptr is NULL, I will bail out in that branch. Therefore, it shall be safe to use ptr after that branch.

vkucera:
> There was a discussion about this some time ago, but I didn't follow up to see whether there were any changes in O2DatabasePDG related to it (it was related back then to the missing lazy initialization of the service O2DatabasePDG)

I admit I also lost track of that issue. Can you please point me to the latest discussion and the overhead estimate? I will follow up with the developers.

> Nothing is hard-coded here related to PDG code, these are merely array indices

Indeed, those are false positives. Please silence them with the disable directive.

>```c++
>int var = Var(); // get the content of variable 'var', from the corresponding function Var()
>```
>
> After O2Linter intervention and renaming the function Var() to var(), this runs into a shadowing problem...

My suggestion would be to rename `Var()` to `getVar()`, following the general convention for getter functions. (According to the O2 naming conventions, `Var()` would be a call of the default constructor of a class `Var`.)

>```c++
>if(!ptr) { 
>  LOGF(fatal, ...); 
>}
>ptr->DoSomething();
>``` 
>
>The `if` statement is not redundant => namely if ptr is NULL, I will bail out in that branch. Therefore, it shall be safe to use ptr after that branch.

You are right. This could be considered a false positive from Cppcheck but there is a valid reasoning behind it. The reason why it complains is that Cppcheck does not have access to the implementation of `LOGF` and therefore cannot evaluate whether the code guarantees the block termination before the dereference of the null pointer. You can either add a `return` statement after `LOGF` to guarantee the termination explicitly or silence the Cppcheck error on that line.

abilandz:
> > There was a discussion about this some time ago, but I didn't follow up to see whether there were any changes in O2DatabasePDG related to it (it was related back then to the missing lazy initialization of the service O2DatabasePDG)
> 
> I admit I also lost track of that issue. Can you please point me to the latest discussion and the overhead estimate? I will follow up with the developers.

I just forwarded an email containing all relevant findings and links related to the memory blow-ups. There was another related post back then by @victor-gonzalez that mentioned the missing lazy initialization of the O2DatabasePDG service as the cause of the memory blow-up in that case, but I cannot dig that one up quickly now.

Thanks for all the suggestions below, I will address them, but most likely only in the next commit tomorrow, as today my candle already burned at both ends...

> 
> > Nothing is hard-coded here related to PDG code, these are merely array indices
> 
> Indeed, those are false positives. Please silence them with the disable directive.
> 
> > ```c++
> > int var = Var(); // get the content of variable 'var', from the corresponding function Var()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > After O2Linter intervention and renaming the function Var() to var(), this runs into a shadowing problem...
> 
> My suggestion would be to rename `Var()` to `getVar()`, following the general convention for getter functions. (According to the O2 naming conventions, `Var()` would be a call of the default constructor of a class `Var`.)
> 
> > ```c++
> > if(!ptr) { 
> >  LOGF(fatal, ...); 
> > }
> > ptr->DoSomething();
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The `if` statement is not redundant => namely if ptr is NULL, I will bail out in that branch. Therefore, it shall be safe to use ptr after that branch.
> 
> You are right. This could be considered a false positive from Cppcheck but there is a valid reasoning behind it. The reason why it complains is that Cppcheck does not have access to the implementation of `LOGF` and therefore cannot evaluate whether the code guarantees the block termination before the dereference of the null pointer. You can either add a `return` statement after `LOGF` to guarantee the termination explicitly or silence the Cppcheck error on that line.

vkucera:
> I just forwarded an email containing all relevant findings and links related to the memory blow-ups.

@abilandz Thanks! I will try to poke further on that one.

abilandz:
@victor-gonzalez : Ola Señor, please can you review and, if the compilation goes through, approve this big PR? The remaining error in O2linter is not related to my code. And the remaining warnings in O2linter about the file names are irrelevant, because in any case I will inline the content of those temporarily fragmented files into the main file?

Thanks!

victor-gonzalez (approved):

PR #17044, opened 2026-07-13
[PWGCF] Add pair-wise kstar histograms to the triplet hist manager

ariedel-cern:
Add pair-wise kstar histograms to the triplet hist manager


alibuild (approved):

PR #17058, opened 2026-07-14
[PWGUD] Adding bc-by-bc trigger info and getting info from DB

jgcn:
Major changes:
get ITS ROF info from database instead of having it hardwired
add info on bc-by.bc triggers
Minor changes:
better definition of TF axis

### DIFF
--- a/PWGUD/Tasks/upcVmRof.cxx
+++ b/PWGUD/Tasks/upcVmRof.cxx
@@ -25,6 +25,8 @@
 #include <CommonDataFormat/TimeStamp.h>
 #include <DataFormatsParameters/AggregatedRunInfo.h>
 #include <DataFormatsParameters/GRPLHCIFData.h>
+//#include <DataFormatsITSMFT/DPLAlpideParam.h>
+#include <ITSMFTBase/DPLAlpideParam.h>
 #include <Framework/ASoA.h>
 #include <Framework/AnalysisDataModel.h>
 #include <Framework/AnalysisHelpers.h>
@@ -180,10 +182,13 @@ struct UpcVmRof {
   std::bitset<o2::constants::lhc::LHCMaxBunches> bcPatternA;
   std::bitset<o2::constants::lhc::LHCMaxBunches> bcPatternB;
   std::bitset<o2::constants::lhc::LHCMaxBunches> bcPatternC;
+  std::vector<int> bcbIdx;
+  int nbcB = 0;
 
-  // constants related to ITS ROFs
-  static constexpr int RofPerOrbit = 6; // valid for pO, OO and PbPb in Run 3
-  static constexpr int RofShift = 64;   // bc shift of ITS. Valid for pO, OO and PbPb in Run 3
+  // variables to store ITS ROF info
+  int rofPerOrbit = -1; // number of rofs per orbit
+  int rofLength = -1; // number of bcs per ROF
+  int rofShift = -1;   // bc shift of ITS. 
 
   // variables to store run info
   int runNumberBc = 0;     // run number used to process BCs
@@ -192,7 +197,6 @@ struct UpcVmRof {
   int64_t orbitsPerTF = 0; // number of orbits per TF
   int64_t bcSOR = 0;       // first bc of the first orbit
   int64_t nBCsPerTF = 0;   // duration of TF in bcs
-  int64_t nBCsPerROF = 0;  //  number of bcs per ROF
   int64_t currentTF = -1;  // current time frame being looked at
   int64_t nTF = 0;         // number of time frames in run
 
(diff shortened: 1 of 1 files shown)


PWGUD/Tasks/upcVmRof.cxx
```diff
@@ -25,6 +25,8 @@
 #include <CommonDataFormat/TimeStamp.h>
 #include <DataFormatsParameters/AggregatedRunInfo.h>
 #include <DataFormatsParameters/GRPLHCIFData.h>
+#include <DataFormatsITSMFT/DPLAlpideParam.h>
+//#include <ITSMFTBase/DPLAlpideParam.h>
```
vkucera: Please do not leave comment lines in between includes. It is useless and it breaks the automatic sorting and grouping.

vkucera (changes requested):

PWGUD/Tasks/upcVmRof.cxx
```diff
@@ -671,7 +711,7 @@ struct UpcVmRof {
     } // ZDC info
 
     // fill output table
-    int recoFlag = (col.flags() & dataformats::Vertex<o2::dataformats::TimeStamp<int>>::Flags::UPCMode) ? 1 : 0;
+    int recoFlag = static_cast<int>((col.flags() & dataformats::Vertex<o2::dataformats::TimeStamp<int>>::Flags::UPCMode) ? true : false);
```
vkucera: This is wrong. You are converting `int` -> `bool` on the left side of `?` and converting `bool` -> `int` on the right side.
You can simply use `TESTBIT(col.flags(), dataformats::Vertex<o2::dataformats::TimeStamp<int>>::Flags::UPCMode) ? 1 : 0`.

vkucera: Also, what do `1` and `0` mean?

jgcn: Hi 
thanks for the tip, and for the email explanations. I will try to follow as much as I can, and push again. I know that C++ is my tool and as such part of my job, but my real job is something else, so the learning curve for me takes a long time to travel ... I am trying, hopefully my learning will converge somewhen ... I will try to push this and whatever I understand from the email tomorrow. Thanks a lot for your help.
guillermo

PWGUD/Tasks/upcVmRof.cxx
```diff
@@ -660,19 +729,23 @@ struct UpcVmRof {
...
+      }
+      if (!std::isfinite(eZNC)) {
         eZNC = -999;
+      }
     } // ZDC info
 
     // fill output table
-    int recoFlag = (col.flags() & dataformats::Vertex<o2::dataformats::TimeStamp<int>>::Flags::UPCMode) ? 1 : 0;
-    if (selTrks.size() == NTrksTwoBody) {
+    const int recoFlag = ((col.flags() & dataformats::Vertex<o2::dataformats::TimeStamp<int>>::Flags::UPCMode) != stdReco) ? upcReco : stdReco;
```
vkucera: Why are you comparing the result of `&` to `stdReco` instead of `0`?

jgcn: Because stdReco is defined as zero. From previous iterations I got the impression that you did not like just numbers, but to give a meaning to them. So I defined the 0 for this particular useage as stdReco and not zero (one) as upcReco (the new definitions were introduced in the last iteration and are towards the start of the code)

vkucera: But `stdReco` seems unrelated to this expression. If `stdReco` had a different value, this expression would still have to be compared to `0`. There is no logical connection between the two.

jgcn: It is related to the expression: the flag UPCMode tells you if it is stdReco or upcReco. But if you want, I can put a zero. I just need this to be approved! Please, let me know what to do exactly to get it approved, so that I can run over data, and see if this new code helps in solving the problems we are facing in PWDUD ...

jgcn: and just to clarify. This is a convention used in reco, vertexing and the full UPC groupd. So, one cannot just change it in one piece of code, ao "If stdReco had a different value," does not apply. It is as if I unilaterally decide that a piece of code uses inches and not cm and want to use other ALICE code in cm .... of course, it will not work. But it should not happen.

jgcn: so it would be great if you could just unblock and approve the PR, please. We can discuss this offline and I can try to explan, or do other test or revert to use zero in the next iteration

amatyja: Hi Vit, It is a convention in the UPC/UD group. We set 0 or 1 as Guillermo pointed out. For me it if fine and we need this code. I would approve it at soon as Guillermo get rid of "Formatting errors".

vkucera: @jgcn @amatyja I don't think you understood me.

The code has the syntax: `flag = condition ? flagYes : flagNo` where your condition is `(col.flags() & dataformats::Vertex<o2::dataformats::TimeStamp<int>>::Flags::UPCMode) != 0` which is independent of the values of `flagYes` and `flagNo`.
Your condition is testing whether `flags` has the `UPCMode` bit set. This has nothing to do with the values of `upcReco` and `stdReco`.

Or am I missing anything?

amatyja: For standard reconstruction col.flags() is set to 0, For UPC reconstruction col.flags() is 2 (so called UPCMode). Mostly flag check returns 0. Guillermo defines 0 as a condition and as a flagNo. I see no sense to multiply definitions of 0 via different flags. flagYes is defined as upcReco. In UD we need just this two cases: standardReco or upcReco.

vkucera: These are not multiple definitions of 0. These are two different entities which happen to have the same value. It's the same as if your condition was testing validity of a pointer by comparing it to a UPC constant which happens to be equal to 0. It's a category error which can turn into an execution bug if the flags change values. Even if they don't, it's simply a wrong expression of what the code is supposed to do.

jgcn: Hi
ok, I'll revert it to zero and try to clean the remaining tabs/spaces tomorrow early morning.
I am hoping that this is really the last iteration (I know that the iterations are due to my ignorance and my editor adding tabs without me noticing :) 
thanks for the comments and help
g

vkucera: Thanks @jgcn . Please note that it is completely unnecessary to fix the formatting by hand. It is literally a matter of two clicks to merge the formatting PR and zero clicks if you use the pre-commit hook.

jgcn: Hi Vit, thakns, but I am really ignorant. I do not know where to click twice ... I will ask later Roman/Kuba to teach me this trick. In the meanwhile I have removed the tabs (and restore the zero) and it seems that the required formatting checks have passed without error (hurra!). Now I hope that the other two checks are successful and that I did not introduced a new problem ...
Anyway: thanks for your help and  patience :)
guillermo

amatyja (approved):

vkucera (approved):

PR #17073, opened 2026-07-15
[PWGCF] Add cascades to derived to derived producer

ariedel-cern:
Implement derived-to-derived builder for cascades and them to the derived-to-derived producer task.


alibuild (approved):

PR #17085, opened 2026-07-16
[PWGHF] Fix selection of deltaMassKK for Ds and move min track pT cut before vertex

fgrosa:
This PR fixes the KK mass selection for the Ds, in fact currently it's not possible to release completely the cut since if `deltaMassKK>massPhi`, every candidate gets rejected.
It also moves the selection on the min track pt for 3 prongs before the decay vertex reconstruction, as it doesn't require the vertex reconstruction.
@fchinu @Marcellocosti

### DIFF
--- a/PWGHF/TableProducer/trackIndexSkimCreator.cxx
+++ b/PWGHF/TableProducer/trackIndexSkimCreator.cxx
@@ -1833,13 +1833,13 @@ struct HfTrackIndexSkimCreator {
     const double deltaMassMax = cut3Prong[hf_cand_3prong::DecayType::DsToKKPi].get(binPt, 5u);
     if (TESTBIT(whichHypo[hf_cand_3prong::DecayType::DsToKKPi], 0)) {
       const double mass2PhiKKPi = RecoDecay::m2(std::array{pVecTrack0, pVecTrack1}, std::array{arrMass3Prong[hf_cand_3prong::DecayType::DsToKKPi][0][0], arrMass3Prong[hf_cand_3prong::DecayType::DsToKKPi][0][1]});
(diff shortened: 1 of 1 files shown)


PWGHF/TableProducer/trackIndexSkimCreator.cxx
```diff
@@ -1833,13 +1833,13 @@ struct HfTrackIndexSkimCreator {
     const double deltaMassMax = cut3Prong[hf_cand_3prong::DecayType::DsToKKPi].get(binPt, 5u);
     if (TESTBIT(whichHypo[hf_cand_3prong::DecayType::DsToKKPi], 0)) {
       const double mass2PhiKKPi = RecoDecay::m2(std::array{pVecTrack0, pVecTrack1}, std::array{arrMass3Prong[hf_cand_3prong::DecayType::DsToKKPi][0][0], arrMass3Prong[hf_cand_3prong::DecayType::DsToKKPi][0][1]});
-      if (mass2PhiKKPi > (MassPhi + deltaMassMax) * (MassPhi + deltaMassMax) || mass2PhiKKPi < (MassPhi - deltaMassMax) * (MassPhi - deltaMassMax)) {
+      if (mass2PhiKKPi > (MassPhi + deltaMassMax) * (MassPhi + deltaMassMax) || (mass2PhiKKPi < (MassPhi - deltaMassMax) * (MassPhi - deltaMassMax) && deltaMassMax < MassPhi)) {
```
fchinu: Thanks! My only comment would be to consider putting `deltaMassMax < MassPhi` before `mass2PhiKKPi < (MassPhi - deltaMassMax) * (MassPhi - deltaMassMax` so that in case one puts very large values for `deltaMassMax`  we don't have to do all the the calculations for checking the delta mass condition:

```suggestion
      if (mass2PhiKKPi > (MassPhi + deltaMassMax) * (MassPhi + deltaMassMax) || deltaMassMax < MassPhi && (mass2PhiKKPi < (MassPhi - deltaMassMax) * (MassPhi - deltaMassMax))) {
```

alibuild (approved):

PR #17091, opened 2026-07-16
[Common] TrackTuner: fix mean shift.

mfaggin:
This PR fixes a conceptual mistake in the dca mean shift applied with the TrackTuner (with no impact for current analyses, see later).
From the PR https://github.com/AliceO2Group/O2Physics/pull/7009 the dca correction is the following:

Schermata del 2026-07-16 14-05-31

In particular, that PR introduced the subtraction of the dcaMC mean from the reco dca in MC, in order to center it at 0 when computing the residual to the generated y parameter, i.e. in the calculation of the dca resolution term. This operation already centers the reconstructed dca in MC to 0, i.e. the `-<dcaXYMC_reco>` term in the 3rd row is a leftover. In this PR we remove this term, ending up with the following formula:

Schermata del 2026-07-16 14-09-17

In particular, this formula now provides a robust correction, which in the ideal case of MC perfectly matching the data returns exactly the reconstructed dca in MC as it is.

The discussed issue is fixed with this PR. However, this should be totally transparent in analysis given that the dca mean in MC looks always at 0. Below some plots with some tests done on the pp reference HF MC `LHC25b4a6`, showing the performance of the mean and sigma w/o the usage of the TrackTuner (black), with the usage of the TrackTuner as in O2Physics up to now (violet) and with the usage of the TrackTuner as fixed in this PR (green). As expected, the TrackTuner changes the mean and resolution of the DCA, but the fix in the TrackTuner mean shift of this PR does not have a big impact given that `<dca_MC>=0` (in other words: the violet and the green points coincide, even if the green are conceptually more correct).

Mean:

can_meanUpgrDcaXY
can_meanUpgrDcaZ
can_withPvRefit_meanUpgrDcaXY
can_withPvRefit_meanUpgrDcaZ

Resolution:
can_resUpgrDcaXY
can_resDcUpgraZ
can_withPvRefit_resUpgrDcaXY
can_withPvRefit_resUpgrDcaZ

Thanks @fchinu for spotting it.
Tagging also @arossi81 and @phymanshu for info

### DIFF
--- a/Common/Tools/TrackTuner.h
+++ b/Common/Tools/TrackTuner.h
@@ -940,7 +940,7 @@ struct TrackTuner : o2::framework::ConfigurableGroup {
 
       // double dd0mrpn=std::abs(sd0mrpn)-std::abs(sd0mrpo);
       // double deltaDcaXYmean = std::abs(dcaXYMeanData) - std::abs(dcaXYMeanMC) ;
-      double deltaDcaXYmean = dcaXYMeanData - dcaXYMeanMC;
+      double deltaDcaXYmean = dcaXYMeanData; // - dcaXYMeanMC; // Remove the shift of the mean, already introduced in deltaDcaXYTuned
 
       // double d0rpn =d0rpmc+dd0rpn-dd0mrpn;
       double trackParDcaXYTuned = trackParDcaXYMC + deltaDcaXYTuned + deltaDcaXYmean;


mfaggin:
@dsekihat @jgrosseo @ddobrigk @ktf

dsekihat (approved):

PR #17115, opened 2026-07-17
[ALICE3] Fix to be able to run the dielectron chain properly

rbailhac:
We found some small mistake to be fixed in order to properly estimate the combinatorial background


njacazio:
Hi, good catch. I'll approve once the test passes

njacazio (approved):

vkucera:
@rbailhac @njacazio Why are the includes still not fixed?

PR #17150, opened 2026-07-21
[PWGCF] Add lambda mass to cascade extras

ariedel-cern:
Add the mass of the daughter lambda lambda to the cascade extras table, along with the corresponding QA histogram, so the histogram manager.


alibuild (approved):

PR #17161, opened 2026-07-22
[PWGJE] Add UPC jet QA task

JaehyeokRyu:
This PR adds a standalone PWGJE UPC jet QA task.
The task is intended to validate UPC event tagging and charged-jet observables before deciding how the required UPC information should be integrated into the PWGJE derived-data workflow.
It provides dedicated QA AOD tables for selected UPC events, charged tracks, charged jets, and jet constituents, with support for data, detector-level MC, and particle-level MC


nzardosh:
Hi @JaehyeokRyu  Thanks for your PR, however there is already a UPC interface in the jet framework (you can find this in the derived data producer). Unless there is something you spot there that is unworkable I would prefer if we use that implementation so that we have one central UPC workflow for the framework.

JaehyeokRyu:
Hi @nzardosh, thanks for pointing this out. I agree that it would be better to avoid introducing a separate UPC path in the jet framework.
The motivation for this PR was mainly a validation/QA study following my recent PWG-JE-JSUB status report. Based on the ongoing UPC event-tagging discussions in UD, I wanted to test whether a FIT-gap-only tagging approach is sufficient for the jet QA use case. For that reason I added the ZDC-side information as well, in particular the ZNA/ZNC common energy and time, the derived neutron class such as XnXn/Xn0n/0nXn/0n0n, and independent gap/ZDC tagging switches.
From what I understand, the current JE derived-data workflow already provides SGSelector-based gap tagging and stores the FIT/FDD amplitudes. Following the recent discussions and presentations in UD, I included the ZDC information to study its usefulness for UPC jet QA. I did not find equivalent ZDC tagging or QA information in the current JE workflow, although I may have overlooked it.
If this functionality would be more appropriate in the central JE derived-data UPC workflow, I would be happy to rework the PR accordingly rather than keep it as a standalone task. Would that be closer to what you had in mind?

nzardosh:
Dear @JaehyeokRyu thanks for the detailed reply. It is indeed useful to check the ZDC requirements and if we find they are needed then we should absoluately add this to the central JE framework. Maybe the part I am missing right now is whether the validation of this new event selection requires jets or if it something that should be common for all UPC analyses?

JaehyeokRyu:
Dear @nzardosh, Thanks for the reply.
Indeed, the validation of the event selection is a common requirement for all UPC analyses, which the UD group is nowadays actively investigating through their QA efforts.
We wanted to include this in the framework because, for our UPC jet analysis, we need to go a step further and perform jet-wise QA. I think it is essential for us to understand how these event taggings influence the kinematics of jets, their constituents, and tracks.
The reason for adding the ZDC is that recent UD reports indicated that FIT-only tagging may not be sufficiently robust in all cases. Therefore, we plan to test multiple taggings—either individually or combined—while monitoring the corresponding jet and track properties.

nzardosh:
Ok then in this case lets do the following. Please remove the table definitions from the derived data model and put them all in your QA task. Then run your QA task and validate what we need. Then when the validation is done we can change the core framework to include what is needed and then we remove the table definitions you made and change your QA task to work within the core framework. Does that sound ok to you?

JaehyeokRyu:
Perfect. I'll revert the core framework changes in this PR and keep the table definitions purely within my QA task for now.
I'll let you know once we test this on the hyperloop and the validation is complete.

nzardosh (approved):

vkucera:
@JaehyeokRyu The includes are misplaced and wrong. Also the using directives are misplaced. Please fix them.
@nzardosh Please do not merge PRs before the tests finish.

vkucera:
> @JaehyeokRyu The includes are misplaced and wrong. Also the using directives are misplaced. Please fix them. @nzardosh Please do not merge PRs before the tests finish.

See https://github.com/AliceO2Group/O2Physics/pull/17316

JaehyeokRyu:
Hi, @vkucera! Thanks for pointing this out, and sorry about that. The linter issue seemed to be related to existing workflows outside my task, but I agree that the include and using-directive placement in my file should follow the repository convention. I’ll fix those positions in a follow-up commit and submit it again.

PR #17172, opened 2026-07-22
[DPG] Update deuteron analysis

rliotino99:
Changes in table hmpid, table producer and deuteron analysis task


njacazio (approved):

vkucera:
@rliotino99 How did you come up with the include changes?
@njacazio Why did you approve the PR despite the include errors?

rliotino99:
Hello @vkucera, regarding the include changes, as far as I can remember, I used VS Code IntelliSense to trace the unresolved symbols to the headers where they are defined. Based on that, I added the corresponding #includes rather than relying on trial and error.

I also remember noticing that, across different commits, some of the code-check errors disappeared automatically. I observed the same behavior recently in another pull request, where those particular errors no longer appeared.

vkucera:
@rliotino99 But how did that make you remove the `Framework` headers?

PR #17188, opened 2026-07-23
[PWGCF] Add lite track and collision tables to femto framework

ariedel-cern:
In a typical derived data production, the track table accounts for roughly 80% of the total data size. This PR introduces new "lite" tables (FLiteTracks, FLiteCols) that store pt, eta, and phi as binned uint16_t values rather than full-precision floats, reducing the overall derived data size by roughly 20% with no significant loss of kinematic accuracy. The same binning approach is applied to select collision properties (vertex Z, multiplicity, centrality).


alibuild (approved):

PR #17199, opened 2026-07-24
[ALICE3] Add option to smear and add charged Xi and Sigma to the track table in otf tracker

jesgum:


alibuild (approved):

PR #17216, opened 2026-07-27
[PWGDQ] add match attempts in ML training data producer

aferrero2707:
The number of match attempts is currently the best estimator of the multiplicity of MFT tracks that are combined with each MCH track when the global muon matching is performed.


aferrero2707:
@mcoquet642 the PR is ready and tested locally. I have fixed most of the errors reported by code-check, for the remaining linter errors I propose to address them in a dedicated PR.

mcoquet642 (approved):

PR #17221, opened 2026-07-27
[PWGLF,PWGMM] Update for the latest data model

aalkin:
The particle primary definition was changed, which breaks the filter. Since this wagon is used in nightly simulation tests, it has to be updated.


alibuild (approved):

PR #17232, opened 2026-07-28
[PWGLF] Remove stale duplicate nucleitpcpbpb.cxx index entry

mpuccio:
The rename to nucleitpcPbPb.cxx in d5e065ddd (#17177) was done from a case-insensitive filesystem, so the old lowercase path never got removed from the tree. Both paths collide into one file on case-insensitive filesystems (macOS/Windows), making the tracked file unrestorable via reset/checkout since git pathspec matching is case-insensitive here.


alibuild (approved):

mpuccio:
@vkucera could you have a look at the macos CI? On my mac I do not get the issue with the configuration, I get, however, a nasty issue with the fact that we have two files with the same name but different case.

vkucera:
@jtanwar2212 Please fix the name of the workflow.

PR #17240, opened 2026-07-28
[PWGCF] Fix DCA selections of daughter tracks

ariedel-cern:
Previously, we used the full DCA (sqrt(DCAxy² + DCAz²)) of the daughter tracks for V0s and cascades when applying the minimum distance-to-primary-vertex cut. This has now been corrected: the minimum daughter-track vertex requirement now uses |DCAxy| instead of the full DCA.


alibuild (approved):

PR #17245, opened 2026-07-29
[PWGDQ] Apply phi-dependent correction to track rotation

ayatsujiringo:
add an option to read phi distribution from ccdb to correct track rotation angle
The track-rotation workflow was tested locally.


XiaozhiBai (approved):

vkucera:
@ayatsujiringo You are supposed to verify your includes.

PR #17255, opened 2026-07-29
[PWGCF] Fix V0V0 processing in femto framework

ariedel-cern:
This PR addresses two issues:
* Fixed the `processLambda` path in the derived-to-derived producer. Following a DF switch, the track table index was not reinitialized, leading to a mismatch between V0 entries and their daughter-track indices in the output.
* Fixed the `Lambda–AntiLambda` processing in the `v0v0` task. The fix required introducing and renaming several configurables to properly separate the processing paths.


prchakra (approved):

alibuild (approved):

PR #17269, opened 2026-07-30
[ALICE3] Fix O2Linter issues in ALICE 3 DQ Framework

axti98:


axti98:
@njacazio We might rename the tasks at some point so the linter errors there and in the CMakeList  are also resolved, just to keep in mind

njacazio (approved):

PR #17274, opened 2026-07-31
[PWGUD] fix ptdiff

miedema-11:
initialize the GFW objects for pT-diff vn correctly, otherwise it will crash


amatyja (approved):

amatyja:
Hi @miedema-11, Write some description, please!
Best regards
Adam

miedema-11:
> Hi @miedema-11, Write some description, please! Best regards Adam

Hi Adam, I have added. Thank you

amatyja:
Hi @miedema-11, after some conversation with Roman, it may happen that the HY train will be crashed due to large output. Could you introduce a switch which handle output messages, and output logs are stored only when flag is set to true? 
Best regards
Adam

PR #17296, opened 2026-08-03
[PWGCF] Femto: Allow multiple registries for SH histograms

IgorPtak:
## Problem

SH histograms scale as `nCent * nKt * (2 * nJM + 2)` and overflow the
512-histogram limit of a single `HistogramRegistry`. 

## Solution

`PairHistManager::init()` takes an optional pool of registries for SH
histograms and assigns each (cent, kT) cell to one with room. Builders
forward separate pools for same-event and mixed-event.


vkucera:
If you need more than 512 histograms, I think you are doing something wrong.

PR #17303, opened 2026-08-03
[PWGLF] Task for sigma+ reco via PCM

hfribert:
This task builds/reconstructs Sigma+ baryons via their decay to a proton + pi0 --> a proton + 2 photons. Only one photon is reconstructed via photon conversion (PCM) through V0's, and the Sigma+ invariant mass is reconstructed using kinematic constraints of the decay daughters (similar approach as performed for Sigma+ correlations in Run2).


lhusova (approved):

vkucera:
@lhusova This PR adds several errors which would have been reported here if you had not merged the PR prematurely. Please do not merge PRs before you see the results of the checks.

PR #17315, opened 2026-08-04
[PWGCF] Change nsigma TOF selections to optional cuts, force bit=1 when no TOF is found

allesgrau:


PWGCF/Femto/Core/cascadeBuilder.h
```diff
@@ -78,7 +79,8 @@ struct ConfCascadeFilters : o2::framework::ConfigurableGroup {
   o2::framework::Configurable<std::vector<float>> posDauTpc{"posDauTpc", {5.f}, "Maximum |nsimga_Pion/Proton| TPC for positive daughter tracks"};                      \
   o2::framework::Configurable<std::vector<float>> negDauTpc{"negDauTpc", {5.f}, "Maximum |nsimga_Pion/Proton| TPC for negative daughter tracks"};                      \
   o2::framework::Configurable<std::vector<float>> posDauTof{"posDauTof", {}, "Maximum |nsimga_Pion/Proton| TOF for positive daughter tracks"};                         \
-  o2::framework::Configurable<std::vector<float>> negDauTof{"negDauTof", {}, "Maximum |nsigma_Pion/Proton| TOF for negative daughter tracks"};
+  o2::framework::Configurable<std::vector<float>> negDauTof{"negDauTof", {}, "Maximum |nsigma_Pion/Proton| TOF for negative daughter tracks"};                         \
+  o2::framework::Configurable<bool> requireTof{"requireTof", false, "If true, only keep candidates whose daughters have a TOF signal"};
```
ariedel-cern: I would introduce 2 new configurables
- The first one you have, i.e. making TOF a minimal (true) or optional (false) selection
- The second configurable should be called like "keepTracksWithoutTof", such that daughters without TOF can be kept

PWGCF/Femto/Core/cascadeBuilder.h
```diff
@@ -242,7 +244,7 @@ class CascadeSelection : public baseselection::BaseSelection<float, o2::analysis
       mOmegaMassLowerLimit = filter.rejectMassOmegaMin.value;
       mOmegaMassUpperLimit = filter.rejectMassOmegaMax.value;
       this->addSelection(kBachelorTpcPion, cascadeSelectionNames.at(kBachelorTpcPion), config.bachelorTpcPion.value, limits::kAbsUpperLimit, true, true, false);
-      this->addSelection(kBachelorTofPion, cascadeSelectionNames.at(kBachelorTofPion), config.bachelorTofPion.value, limits::kAbsUpperLimit, true, true, false);
+      this->addSelection(kBachelorTofPion, cascadeSelectionNames.at(kBachelorTofPion), config.bachelorTofPion.value, limits::kAbsUpperLimit, true, false, false);
```
ariedel-cern: Here the flag should be set according to "requireTof" from the configuration

PWGCF/Femto/Core/cascadeBuilder.h
```diff
@@ -251,7 +253,7 @@ class CascadeSelection : public baseselection::BaseSelection<float, o2::analysis
       mXiMassLowerLimit = filter.rejectMassXiMin.value;
       mXiMassUpperLimit = filter.rejectMassXiMax.value;
       this->addSelection(kBachelorTpcKaon, cascadeSelectionNames.at(kBachelorTpcKaon), config.bachelorTpcKaon.value, limits::kAbsUpperLimit, true, true, false);
-      this->addSelection(kBachelorTofKaon, cascadeSelectionNames.at(kBachelorTofKaon), config.bachelorTofKaon.value, limits::kAbsUpperLimit, true, true, false);
+      this->addSelection(kBachelorTofKaon, cascadeSelectionNames.at(kBachelorTofKaon), config.bachelorTofKaon.value, limits::kAbsUpperLimit, true, false, false);
```
ariedel-cern: same as above

PWGCF/Femto/Core/cascadeBuilder.h
```diff
@@ -336,6 +339,9 @@ class CascadeSelection : public baseselection::BaseSelection<float, o2::analysis
     if (bachelor.hasTOF()) {
       this->evaluateObservable(kBachelorTofPion, bachelor.tofNSigmaPi());
       this->evaluateObservable(kBachelorTofKaon, bachelor.tofNSigmaKa());
+    } else if (!mRequireTof) {
```
ariedel-cern: here you should then check against the flag whether to keep tracks without Tof or not
Also, the limit to check against should be 0 (since it is a AbsUpperCut, so the only value which should alwasy succeed is 0. Using max value would always fail)

ariedel-cern:
Hi @allesgrau ,
many thanks for the PR. I think the PR still needs some work. I commented here only the cascade builder, but the same comments apply also for the V0 builder.

PWGCF/Femto/Core/cascadeBuilder.h
```diff
@@ -33,6 +33,7 @@
 #include <array>
 #include <cmath>
 #include <cstdint>
+#include <limits>
```
ariedel-cern: This include is no longer needed. Same in V0Builder

PWGCF/Femto/Core/cascadeBuilder.h
```diff
@@ -336,6 +341,9 @@ class CascadeSelection : public baseselection::BaseSelection<float, o2::analysis
     if (bachelor.hasTOF()) {
       this->evaluateObservable(kBachelorTofPion, bachelor.tofNSigmaPi());
       this->evaluateObservable(kBachelorTofKaon, bachelor.tofNSigmaKa());
+    } else if (mKeepTracksWithoutTof) {
+      this->setBitmask(kBachelorTofPion, 0);
```
ariedel-cern: This call would set the bitmask itself to 0, i.e. the selection would not be passed, which is what we want to avoid in this branch if the track has no TOF.
You have to call here
```
this->evaluateObservable(kBachelorTofPion, 0);
```
Like this we compare 0 against all the limits (which are positive definite), so all bits in the bitmask will be set to 1.
This applies to all other calls below (and in the V0Builder).

ariedel-cern (changes requested):

ariedel-cern (approved):

PR #17318, opened 2026-08-04
[PWGJE] Adding bkg indexing to jetCorrelationD0 task

MattOckleton:
Removing matchedFrom and selectedAs columns and replacing with category which will be filled with an int corresponding to the signal, reflection, or correlated background.


nzardosh (approved):

vkucera:
@MattOckleton Please fix the errors.

MattOckleton:
> @MattOckleton Please fix the errors.
Hi @vkucera ,
They should be fixed here: https://github.com/AliceO2Group/O2Physics/pull/17383
But this in itself is confusing as alibuild warns of some build errors which do not seem to appear in the log (the log suggests the build succeeded with no errors).

PR #17320, opened 2026-08-04
[PWGEM] Update SkimmerForDalitzEE

lgansbartl:
Adjust primary dilepton skimmer code in PM to be closer to Dilepton skimmer code.
- Add track to collision association
- Add hadron rejection
- Add TOF recovery
- Add option to fill tables only with tracks not with pairs


mhemmer-cern (approved):
MegaLinter and code-check errors and warnings should be fixed asap. However, since all Cl for build are broken as of today and the changes from this PR are needed, it can be merged now.

PR #17321, opened 2026-08-04
[PWGCF] Implement passThrough mode in femto producer for MC processing

ariedel-cern:
In preparation for an update to the pair efficiency task, add a pass-through mode for MC processing so that all generated collisions and primary charged particles are written to the derived tables, independently of whether a candidate was reconstructed.

Also improve the identification of tracks associated to the wrong collision. Since this is a property of the track/particle link rather than of the particle itself, it cannot be determined by the producer and is now evaluated in the histogram managers.


alibuild (dismissed):

alibuild (approved):

PR #17348, opened 2026-08-05
[PWGHF] HFInvMassFitter: stabilize fit quality, add fit-quality status to the output

lubynets:
**1. Fit quality stabilization**
With the existing version of the fitter some fits are exceptionally poor, and the shape of background and / or peak is completely wrong. It happens due to wrong estimation of initial values of background's and signal's integral. In this PR:
-- Estimate of background and signal integral was fixed;
-- Pre-fit of background sidebands is now done with `chi2FitTo()` always, since `fitTo()` function (likelihood) turned out to be agnostic to the `Ranges()` parameter, i.e. it also fitted the peak when sidebands were requested (that's why at some fit plots the gray line representing sidebands pre-fit was sometimes completely misfittig, while the total fit looked healthy).

**2. Add fit-quality status to the output**
-- `hFitResult` histogram is added to the output, which contains RooFit's `status`, covariance matrix quality (`cov qual`), estimated distance to minimum (`edm`), minimum value of the negative log-likelihood (`minNLL`) and global correlation coefficient of the signal's integral with other fit parameters (`N Sig GCC`);
-- `CovCorrMatrices/hCovCorrMatrix` histograms are added, which contain covariance and correlation across all fit parameters.

**3. Code cleaning**
-- Avoided crash if a single histogram fitting is failed. Instead the exception is handled and fitting of remaining histograms continues;
-- Fixed (most of) clang-tidy and cpp check errors.

### DIFF
--- a/PWGHF/D2H/Macros/runMassFitter.C
+++ b/PWGHF/D2H/Macros/runMassFitter.C
@@ -218,7 +218,7 @@ void runMassFitter(const std::string& configFileName)
 
   std::vector<double> sliceVarLimits(nHistograms + 1);
 
-  auto checkVectorSize = [&](const auto& vec, const std::string& name = "", const bool isEmptyOk = false) {
+  auto checkVectorSize = [&](const auto& vec, const std::string& name, const bool isEmptyOk = false) {
     if (vec.size() != static_cast<size_t>(nHistograms)) {
       if (isEmptyOk && vec.empty()) {
         return;
--- a/PWGHF/D2H/Macros/runMassFitter.C
+++ b/PWGHF/D2H/Macros/runMassFitter.C
@@ -542,7 +542,11 @@ void runMassFitter(const std::string& configFileName)
     setDscbParameter(dscbNRLower, &HFInvMassFitter::setDscbNRLowLimit);
     setDscbParameter(dscbNRUpper, &HFInvMassFitter::setDscbNRUpLimit);
 
-    massFitter->doFit();
+    try {
+      massFitter->doFit();
+    } catch (...) {
+      std::cout << "exception caught while doing fit\n";
+    }
 
     auto drawOnCanvas = [&](std::vector<TCanvas*>& canvas, std::function<void()> drawer) {
       if (nHistograms > 1) {
--- a/PWGHF/D2H/Macros/runMassFitter.C
+++ b/PWGHF/D2H/Macros/runMassFitter.C
@@ -481,6 +481,17 @@ void runMassFitter(const std::string& configFileName)
       hMassSgn[iSliceVar]->Rebin(nRebin[iSliceVar]);
     }
 
+    const auto hMassLo = hMass[iSliceVar]->GetXaxis()->GetXmin();
+    if (massMin[iSliceVar] < hMassLo) {
+      printf("Warning! massMin[%d] is less than hMass[%d] left edge (%f vs %f) and will be assigned the value of the latter\n", iSliceVar, iSliceVar, massMin[iSliceVar], hMassLo);
+      massMin[iSliceVar] = hMassLo;
+    }
+    const auto hMassUp = hMass[iSliceVar]->GetXaxis()->GetXmax();
+    if (massMax[iSliceVar] > hMassUp) {
+      printf("Warning! massMax[%d] is greater than hMass[%d] right edge (%f vs %f) and will be assigned the value of the latter\n", iSliceVar, iSliceVar, massMax[iSliceVar], hMassUp);
+      massMax[iSliceVar] = hMassUp;
+    }
+
     double reflOverSgn = 0;
 
     HFInvMassFitter* massFitter = new HFInvMassFitter(hMass[iSliceVar], massMin[iSliceVar], massMax[iSliceVar], bkgFunc[iSliceVar], sgnFunc[iSliceVar], randomSeed);
--- a/PWGHF/D2H/Macros/runMassFitter.C
+++ b/PWGHF/D2H/Macros/runMassFitter.C
@@ -490,7 +490,7 @@ void runMassFitter(const std::string& configFileName)
     hMass[iSliceVar]->SetTitle(Form("%s;%s;Counts per %0.1f MeV/#it{c}^{2}",
                                     ptTitle.Data(), massAxisTitle.c_str(),
                                     hMass[iSliceVar]->GetBinWidth(1) * 1000));
-    hMass[iSliceVar]->SetName(Form("MassForFit%d", iSliceVar));
+    hMass[iSliceVar]->SetName(Form("hMassForFit%d", iSliceVar + 1));
 
     if (enableRefl) {
       hMassRefl[iSliceVar]->Rebin(nRebin[iSliceVar]);
@@ -702,9 +702,16 @@ void runMassFitter(const std::string& configFileName)
   }
 
   for (int iSliceVar = 0; iSliceVar < nHistograms; iSliceVar++) {
+    if (iSliceVar == 0) {
+      outputFile.mkdir("MassHistograms");
+      outputFile.mkdir("SgnCorrHistograms");
+    }
+    outputFile.cd("MassHistograms");
     hMass[iSliceVar]->Write();
+    outputFile.cd("SgnCorrHistograms");
     hSgnCorr[iSliceVar]->Write();
   }
+  outputFile.cd();
   hRawYieldsSignal->Write();
   hRawYieldsSignalCounted->Write();
   hRawYieldsBkg->Write();
--- a/PWGHF/D2H/Macros/runMassFitter.C
+++ b/PWGHF/D2H/Macros/runMassFitter.C
@@ -574,8 +574,8 @@ void runMassFitter(const std::string& configFileName)
 
     try {
       massFitter->doFit();
-    } catch (...) {
-      std::cout << "exception caught while doing fit\n";
+    } catch (const std::exception& e) {
+      printf("Warinig! Exception \"%s\" caught while doing fit of the histogram no. %d. The fitting process will be continued without it.\n", e.what(), iSliceVar);
     }
 
     auto drawOnCanvas = [&](std::vector<TCanvas*>& canvas, std::function<void()> drawer) {
@@ -703,7 +703,9 @@ void runMassFitter(const std::string& configFileName)
     outputFile.cd("MassHistograms");
     hMass[iSliceVar]->Write();
     outputFile.cd("CovCorrMatrices");
-    hCovCorr[iSliceVar]->Write(Form("hCovCorrMatrix%d", iSliceVar + 1));
+    if (hCovCorr[iSliceVar] != nullptr) {
+      hCovCorr[iSliceVar]->Write(Form("hCovCorrMatrix%d", iSliceVar + 1));
+    }
   }
   outputFile.cd();
   hRawYieldsSignal->Write();
--- a/PWGHF/D2H/Macros/runMassFitter.C
+++ b/PWGHF/D2H/Macros/runMassFitter.C
@@ -321,7 +321,6 @@ void runMassFitter(const std::string& configFileName)
   std::vector<TH1*> hMassSgn(nHistograms);
   std::vector<TH1*> hMassRefl(nHistograms);
   std::vector<TH1*> hMass(nHistograms);
-  std::vector<TH1*> hSgnCorr(nHistograms);
   std::vector<TH2*> hCovCorr(nHistograms);
 
   for (int iSliceVar = 0; iSliceVar < nHistograms; iSliceVar++) {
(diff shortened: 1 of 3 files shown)


PWGHF/D2H/Macros/runMassFitter.C
```diff
@@ -495,27 +523,27 @@ void runMassFitter(const std::string& configFileName)
       massFitter->setUseChi2Fit();
     }
 
-    auto setFixedValue = [&iSliceVar](bool const& isFix, std::vector<double> const& fixManual, const TH1* histToFix, std::function<void(double)> setFunc, std::string const& var) -> void {
+    auto setFixedValue = [&iSliceVar, massFitter](bool const& isFix, std::vector<double> const& fixManual, const TH1* histToFix, void (HFInvMassFitter::*setter)(double), std::string const& var) -> void {
```
vkucera, line 529: What is the motivation for removing `std::function` here?

vkucera (commented):
Hi @lubynets , thanks for the fixes!

PWGHF/D2H/Macros/runMassFitter.C
lubynets: I wanted to make `setFixedValue()` uniform with [`setDscbParameter()`](https://github.com/AliceO2Group/O2Physics/blob/ede5910ec1e08ac14a6db845829558bfca637957/PWGHF/D2H/Macros/runMassFitter.C#L527)

vkucera: OK, thanks. What speaks against using `std::function` everywhere?

lubynets: No special reasoning against `std::function`, it was a random choice. Are there reasons not to use `void()()`?

vkucera: > No special reasoning against `std::function`, it was a random choice. Are there reasons not to use `void()()`?

Isn't it some kind of old unsafe C-style syntax that has been superseded by stuff in the standard library?

lubynets: Maybe the style is indeed C-like, but what unsafe things can happen? Naively I would say that explicit requirement of `HFInvMassFitter`'s member function should prevent from wrong usage of the `setFixedValue` (?)
A little bit context: the reason why I touched this part of code was to avoid using `std::bind()`, discouraged by clang-tidy. I switched from `std::function` to `void()()`, but now see that there is another [place](https://github.com/lubynets/O2Physics/blob/c3737d3086e9e8270cb794224135f5c1c6888d9e/PWGHF/D2H/Macros/runMassFitter.C#L580) where `std::function` remains, but instead of `std::bind()` the lambda is [used](https://github.com/lubynets/O2Physics/blob/c3737d3086e9e8270cb794224135f5c1c6888d9e/PWGHF/D2H/Macros/runMassFitter.C#L591).
I don't have strong preferences to any of the two approaches. Maybe it makes sense to use everywhere the same style, but I don't know which one is more academically correct.

vkucera:
@lubynets What is the status? Do you plan to fix the remaining errors in this PR or later?

lubynets:
> @lubynets What is the status? Do you plan to fix the remaining errors in this PR or later?

The fixes, proposed by code-check (designated initializer list and `std::map::contains()`) require c++20 standard. Since the HFInvMassFitter is not a part of compiled O2Physics code, and can be built standalone, I would like to preserve a compatibility with older standards, namely c++17.
What about the `TMatrixDSym`, I think the reason for the error is that the [TMatrixDSym.h](https://root.cern/doc/v620/TMatrixDSym_8h_source.html) contains other `include`s only, which in their turn define the TMatrixDSym.

vkucera:
> > @lubynets What is the status? Do you plan to fix the remaining errors in this PR or later?
> 
> The fixes, proposed by code-check (designated initializer list and `std::map::contains()`) require c++20 standard. Since the HFInvMassFitter is not a part of compiled O2Physics code, and can be built standalone, I would like to preserve a compatibility with older standards, namely c++17. What about the `TMatrixDSym`, I think the reason for the error is that the [TMatrixDSym.h](https://root.cern/doc/v620/TMatrixDSym_8h_source.html) contains other `include`s only, which in their turn define the TMatrixDSym.

For the suggestions unsupported by C++17, please silence the errors and add an explanatory comment.
For the `TMatrixDSym.h` include, you need to do this trick:

https://github.com/AliceO2Group/O2Physics/blob/2f2c4e000af37bbed2c73990900993d1c36a4740/PWGHF/HFC/Macros/DhCorrelationFitter.cxx#L25-L26

lubynets:
> > > @lubynets What is the status? Do you plan to fix the remaining errors in this PR or later?
> > 
> > 
> > The fixes, proposed by code-check (designated initializer list and `std::map::contains()`) require c++20 standard. Since the HFInvMassFitter is not a part of compiled O2Physics code, and can be built standalone, I would like to preserve a compatibility with older standards, namely c++17. What about the `TMatrixDSym`, I think the reason for the error is that the [TMatrixDSym.h](https://root.cern/doc/v620/TMatrixDSym_8h_source.html) contains other `include`s only, which in their turn define the TMatrixDSym.
> 
> For the suggestions unsupported by C++17, please silence the errors and add an explanatory comment. For the `TMatrixDSym.h` include, you need to do this trick:
> 
> https://github.com/AliceO2Group/O2Physics/blob/2f2c4e000af37bbed2c73990900993d1c36a4740/PWGHF/HFC/Macros/DhCorrelationFitter.cxx#L25-L26

@vkucera, thanks for tips. Is silencing of code-check errors done in the same way as that of o2-linter?
`// o2-linter: disable=name_of_check (explanation)`

vkucera:
> > > > @lubynets What is the status? Do you plan to fix the remaining errors in this PR or later?
> > > 
> > > 
> > > The fixes, proposed by code-check (designated initializer list and `std::map::contains()`) require c++20 standard. Since the HFInvMassFitter is not a part of compiled O2Physics code, and can be built standalone, I would like to preserve a compatibility with older standards, namely c++17. What about the `TMatrixDSym`, I think the reason for the error is that the [TMatrixDSym.h](https://root.cern/doc/v620/TMatrixDSym_8h_source.html) contains other `include`s only, which in their turn define the TMatrixDSym.
> > 
> > 
> > For the suggestions unsupported by C++17, please silence the errors and add an explanatory comment. For the `TMatrixDSym.h` include, you need to do this trick:
> > https://github.com/AliceO2Group/O2Physics/blob/2f2c4e000af37bbed2c73990900993d1c36a4740/PWGHF/HFC/Macros/DhCorrelationFitter.cxx#L25-L26
> 
> @vkucera, thanks for tips. Is silencing of code-check errors done in the same way as that of o2-linter? `// o2-linter: disable=name_of_check (explanation)`

See https://aliceo2group.github.io/analysis-framework/docs/tools/#clang-tidy

lubynets:
> > > @lubynets What is the status? Do you plan to fix the remaining errors in this PR or later?
> > 
> > 
> > The fixes, proposed by code-check (designated initializer list and `std::map::contains()`) require c++20 standard. Since the HFInvMassFitter is not a part of compiled O2Physics code, and can be built standalone, I would like to preserve a compatibility with older standards, namely c++17. What about the `TMatrixDSym`, I think the reason for the error is that the [TMatrixDSym.h](https://root.cern/doc/v620/TMatrixDSym_8h_source.html) contains other `include`s only, which in their turn define the TMatrixDSym.
> 
> For the suggestions unsupported by C++17, please silence the errors and add an explanatory comment. For the `TMatrixDSym.h` include, you need to do this trick:
> 
> https://github.com/AliceO2Group/O2Physics/blob/2f2c4e000af37bbed2c73990900993d1c36a4740/PWGHF/HFC/Macros/DhCorrelationFitter.cxx#L25-L26

Done.
But another error is caught by MegaLinter due to `// NOLINTEND(...)`
Will it be appropriate to add a filter `-readability/nolint` to the [CPPLINT.cfg](https://github.com/AliceO2Group/O2Physics/blob/bfeb5086dfc0582bba8a07161cf63a3786c1ba0b/CPPLINT.cfg)?

vkucera:
> > > > @lubynets What is the status? Do you plan to fix the remaining errors in this PR or later?
> > > 
> > > 
> > > The fixes, proposed by code-check (designated initializer list and `std::map::contains()`) require c++20 standard. Since the HFInvMassFitter is not a part of compiled O2Physics code, and can be built standalone, I would like to preserve a compatibility with older standards, namely c++17. What about the `TMatrixDSym`, I think the reason for the error is that the [TMatrixDSym.h](https://root.cern/doc/v620/TMatrixDSym_8h_source.html) contains other `include`s only, which in their turn define the TMatrixDSym.
> > 
> > 
> > For the suggestions unsupported by C++17, please silence the errors and add an explanatory comment. For the `TMatrixDSym.h` include, you need to do this trick:
> > https://github.com/AliceO2Group/O2Physics/blob/2f2c4e000af37bbed2c73990900993d1c36a4740/PWGHF/HFC/Macros/DhCorrelationFitter.cxx#L25-L26
> 
> Done. But another error is caught by MegaLinter due to `// NOLINTEND(...)` Will it be appropriate to add a filter `-readability/nolint` to the [CPPLINT.cfg](https://github.com/AliceO2Group/O2Physics/blob/bfeb5086dfc0582bba8a07161cf63a3786c1ba0b/CPPLINT.cfg)?

The error seems legit to me. The begin and end comments are not in the same scope.

lubynets:
> > > > > @lubynets What is the status? Do you plan to fix the remaining errors in this PR or later?
> > > > 
> > > > 
> > > > The fixes, proposed by code-check (designated initializer list and `std::map::contains()`) require c++20 standard. Since the HFInvMassFitter is not a part of compiled O2Physics code, and can be built standalone, I would like to preserve a compatibility with older standards, namely c++17. What about the `TMatrixDSym`, I think the reason for the error is that the [TMatrixDSym.h](https://root.cern/doc/v620/TMatrixDSym_8h_source.html) contains other `include`s only, which in their turn define the TMatrixDSym.
> > > 
> > > 
> > > For the suggestions unsupported by C++17, please silence the errors and add an explanatory comment. For the `TMatrixDSym.h` include, you need to do this trick:
> > > https://github.com/AliceO2Group/O2Physics/blob/2f2c4e000af37bbed2c73990900993d1c36a4740/PWGHF/HFC/Macros/DhCorrelationFitter.cxx#L25-L26
> > 
> > 
> > Done. But another error is caught by MegaLinter due to `// NOLINTEND(...)` Will it be appropriate to add a filter `-readability/nolint` to the [CPPLINT.cfg](https://github.com/AliceO2Group/O2Physics/blob/bfeb5086dfc0582bba8a07161cf63a3786c1ba0b/CPPLINT.cfg)?
> 
> The error seems legit to me. The begin and end comments are not in the same scope.

The `NOLINTBEGIN` and `NOLINTEND` are indeed inconsistent with respect to the scope, but placing them in the same scope does not solve the problem, the error is the same (tested by MegaLinter locally).

vkucera:
> > > > > > @lubynets What is the status? Do you plan to fix the remaining errors in this PR or later?
> > > > > 
> > > > > 
> > > > > The fixes, proposed by code-check (designated initializer list and `std::map::contains()`) require c++20 standard. Since the HFInvMassFitter is not a part of compiled O2Physics code, and can be built standalone, I would like to preserve a compatibility with older standards, namely c++17. What about the `TMatrixDSym`, I think the reason for the error is that the [TMatrixDSym.h](https://root.cern/doc/v620/TMatrixDSym_8h_source.html) contains other `include`s only, which in their turn define the TMatrixDSym.
> > > > 
> > > > 
> > > > For the suggestions unsupported by C++17, please silence the errors and add an explanatory comment. For the `TMatrixDSym.h` include, you need to do this trick:
> > > > https://github.com/AliceO2Group/O2Physics/blob/2f2c4e000af37bbed2c73990900993d1c36a4740/PWGHF/HFC/Macros/DhCorrelationFitter.cxx#L25-L26
> > > 
> > > 
> > > Done. But another error is caught by MegaLinter due to `// NOLINTEND(...)` Will it be appropriate to add a filter `-readability/nolint` to the [CPPLINT.cfg](https://github.com/AliceO2Group/O2Physics/blob/bfeb5086dfc0582bba8a07161cf63a3786c1ba0b/CPPLINT.cfg)?
> > 
> > 
> > The error seems legit to me. The begin and end comments are not in the same scope.
> 
> The `NOLINTBEGIN` and `NOLINTEND` are indeed inconsistent with respect to the scope, but placing them in the same scope does not solve the problem, the error is the same (tested by MegaLinter locally).

Indeed. It's a bug that was fixed in March. I will add an exception in the config as you suggested until we get a new release.

vkucera:
@lubynets https://github.com/AliceO2Group/O2Physics/pull/17499

vkucera (approved):

PR #17349, opened 2026-08-05
[Common] Reduce verbosity of ctpRateFetcher when trigger classes are not found

fchinu:
In part of 2023 the trigger name changed, so the `CMTVX-B-NOPF` trigger class for some periods  is not found. Both the "Trying different class" and the "class not found" message were logged at info/warn level, and since fetch() is called per collision in the TPC PID service task https://github.com/AliceO2Group/O2Physics/blob/ede5910ec1e08ac14a6db845829558bfca637957/Common/Tools/PID/pidTPCModule.h#L523
this produced very large log files, causing jobs to fail because of `Too much disk space used`: https://alimonitor.cern.ch/agent/jobs/details.jsp?pid=3677551603
This PR demotes both messages to `LOG(debug)`

### DIFF
--- a/Common/CCDB/ctpRateFetcher.cxx
+++ b/Common/CCDB/ctpRateFetcher.cxx
@@ -47,7 +47,7 @@ double ctpRateFetcher::fetch(o2::ccdb::BasicCCDBManager* ccdb, uint64_t timeStam
     } else {
       double_t ret = fetchCTPratesClasses(ccdb, timeStamp, runNumber, "CMTVX-B-NOPF");
       if (ret < 0.) {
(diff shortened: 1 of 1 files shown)


fchinu:
@dsekihat CI failures do not seem to be related to the changes implemented in this PR, do you think we can merge it? Thanks a lot!

ddobrigk (approved):

dsekihat:
Ciao @fchinu , I am sorry for my delay. It seems that I don't have approval right in this directory. Thanks to @ddobrigk for merging this PR.

PR #17367, opened 2026-08-06
[PWGLF] Add SigmaPlus PCM tables

hfribert:
Added more tables for the SigmaPlus reco via PCM, in order to have more variables available for cutting background and for a BDT (boosted decision tree) method, as was done in Run 2.


romainschotter (approved):

vkucera:
@romainschotter Do not merge PRs with errors.

PR #17370, opened 2026-08-07
[ALICE3] Adapt input file handling to new ACTS input format

lhusova:
update of the translator to be consistent with the new ACTS output structure

### DIFF
--- a/ALICE3/TableProducer/alice3TrackingTranslator.cxx
+++ b/ALICE3/TableProducer/alice3TrackingTranslator.cxx
@@ -297,7 +297,7 @@ struct Alice3TrackingTranslator {
   };
   void addMCParticle(int collIndex, ParticleStruct& fileParticles, int iParticle, uint8_t flags, int firstMother, int firstDaughter, int secondDaughter, int numberOfHits)
   {
-    int mothers[2] = {firstMother, -1};
+    int mothers[2] = {firstMother, firstMother};
     int daughters[2] = {firstDaughter, secondDaughter};
     tableStoredMcParticles(collIndex,                                                                      // mcCollisionId
                            fileParticles.m_particle_type->at(iParticle),                                   // pdgCode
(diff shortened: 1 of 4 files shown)


ALICE3/TableProducer/alice3TrackingTranslator.cxx
```diff
@@ -228,10 +295,10 @@ struct Alice3TrackingTranslator {
     }
     std::vector<unsigned int>* barcode = nullptr;
   };
-  void addMCParticle(int collIndex, ParticleStruct& fileParticles, int iParticle, uint8_t flags, int firstMother, int firstDaughter, int numberOfHits)
+  void addMCParticle(int collIndex, ParticleStruct& fileParticles, int iParticle, uint8_t flags, int firstMother, int firstDaughter, int secondDaughter, int numberOfHits)
   {
     int mothers[2] = {firstMother, -1};
```
jesgum, line 300: The array of mother indices stores the first and last index similar to the daughter one, so what I did for with the decayer for one mother, is that i stored the index both first and last. This is what lets you loop over the mother array like we do in the sameMotherCheck
```
for (const auto& mcParticleMother : mcParticle.template mothers_as<aod::McParticles>())
```
It might still be okay to keep it like this but I would change it just in case

```suggestion
    int mothers[2] = {firstMother, firstMother};
```

ALICE3/TableProducer/alice3strangenessFinder.cxx
jesgum, line 582: This is maybe beyond the scope of this PR but would it not be possible to wrap this block into 

```
if (posTrack.has_mcParticle() && negTrack.has_mcParticle()) {
 // de-reference mcparticle
}
```

From what I remember, all ACTS tracks would not necessarily have an associated MC particle, but if that's the case, the index in `aod::McTrackLabels` could be set to -1 (maybe it is already). Because what `has_mcParticle()` checks if the index is greater than zero, so we would never enter this block if there is no associated mcParticle

jesgum (commented):
Hi @lhusova I just have some minor suggestions :-)

njacazio (approved):

njacazio:
Based on the xp on the LUT maker, I think the best would be to move this to the ACTSO2, the problem would be the table creation, since actsO2 does not depend on O2, but we might find a workaround

PR #17378, opened 2026-08-07
[PWGLF] reduced derived table size and updated the rejection condition of tracks

prottayCMT:


alibuild (approved):

PR #17387, opened 2026-08-09
[PWGLF] reducing table size by storing only events with k0s

prottayCMT:


alibuild (approved):

PR #17407, opened 2026-08-11
[PWGEM] add new deduplication modes + MC information

Steffimro:
Two new deduplication modes added:

**Deduplaction Mode 2**: Instead of ranking V0 candidates by score and accepting them one by one, as the greedy mode does, the whole group of candidates connected through shared tracks is decided at once, keeping the maximum number of V0s that do not share a track. Among equally large solutions the one with the smallest sum of scores wins, so the score still decides, but only between solutions of the same size.

**Dedpulication Mode 3**: Keeps all V0s. No deduplication - more for cross-checks.

For better diagnostics of the deduplication procedure truth categories where added. processMC now classifies every candidate from the MC mothers of its two legs: true (both legs from the same photon), cross-leg fake (legs from two different photons), other fake (at least one leg is not a conversion leg) — and records which surviving candidate took a leg from a rejected one. A true photon counts as lost when all of its true candidates were rejected, and those blockers give the reason.

### DIFF
--- a/PWGEM/PhotonMeson/TableProducer/photonconversionbuilder.cxx
+++ b/PWGEM/PhotonMeson/TableProducer/photonconversionbuilder.cxx
@@ -74,7 +74,6 @@
 #include <string>
 #include <tuple>
 #include <unordered_map>
-#include <unordered_set>
 #include <utility>
 #include <vector>
 
@@ -106,8 +105,10 @@ enum TrackPropMode {
 };
 
 enum V0DeduplicationMode {
-  Pairwise = 0,      // old algorithm
-  GreedyMatching = 1 // new algorithm
+  Pairwise = 0,       // old algorithm, V0 with best PCA + cosPA at the same time in a group wins
+  GreedyMatching = 1, // new algorithm, best score wins (score combines PCA and cosPA -- see detailed implementation in PCMUtilities.cxx)
+  GroupMatching = 2,  // new algorithm, best combination of leg-disjoint V0s wins, not the single best V0
+  KeepAll = 3         // no deduplication, keep all V0s
 };
 
 // Struct needed for deduplication mode 1. Used to keep track  of v0 properties and a score based on pca and cosPA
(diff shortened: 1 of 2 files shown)


PWGEM/PhotonMeson/Utils/PCMUtilities.h
```diff
@@ -212,5 +214,53 @@ inline float getScoreV0(float cosPA, float pca, float weight)
...
+  kV0True = 0,         // both legs come from the SAME true photon
+  kV0CrossLegFake = 1, // legs come from TWO different true photons
+  kV0OtherFake = 2     // at least one leg is not a photon conversion leg
+};
+//_______________________________________________________________________
+/// \brief global index of the photon a track descends from
+/// \param track track with an MC label
+/// \param mcparticles the full McParticles table
+/// \return index of the mother if it is a photon, -1 otherwise
+template <typename TTrack, typename TMCParticles>
```
mhemmer-cern: replace typename with the correct concept for iterators and tables to speed up build time please

PWGEM/PhotonMeson/Utils/PCMUtilities.h
```diff
@@ -212,5 +214,53 @@ inline float getScoreV0(float cosPA, float pca, float weight)
...
+  }
+  return (mcparticles.iteratorAt(mothers[0]).pdgCode() == PDG_t::kGamma) ? static_cast<int64_t>(mothers[0]) : -1;
+}
+
+//_______________________________________________________________________
+/// \brief classify a V0 candidate by the photon mothers of its two legs
+/// \param motherPos photon mother of the positive leg, -1 if none
+/// \param motherEle photon mother of the negative leg, -1 if none
+/// \return kV0True, kV0CrossLegFake or kV0OtherFake
+template <typename TTrack, typename TMCParticles>
```
mhemmer-cern: replace typename with the correct concept for iterators and tables to speed up build time please

PWGEM/PhotonMeson/TableProducer/photonconversionbuilder.cxx
```diff
@@ -164,8 +169,10 @@ struct PhotonConversionBuilder {
   Configurable<double> d_bz_input{"d_bz", -999, "bz field, -999 is automatic"};
   Configurable<int> useMatCorrType{"useMatCorrType", 0, "0: none, 1: TGeo, 2: LUT"};
   Configurable<int> modeTrackPropagation{"modeTrackPropagation", 0, "0: use real track propagation, including material, 1: use fast approximation using only geometry, 2: Use real track propagation and make comparison to fast propagation (only for debugging and testing)"};
-  Configurable<int> deduplicationMode{"deduplicationMode", 0, "0: Pairwise deduplication, 1: Based on Greedy matching (best score wins)"};
+  Configurable<int> deduplicationMode{"deduplicationMode", 0, "0: Pairwise deduplication, 1: Based on Greedy matching (best score wins), 2: Based on Group matching (crossed pairs are both kept), 3: Keep all V0s"};
```
mhemmer-cern: Maybe add comment that method 1 is fall back method

mhemmer-cern (changes requested):

mhemmer-cern (approved):

PR #17410, opened 2026-08-11
[PWGCF] Add lite colshape table

ariedel-cern:
Adding the lite version of FColShape table and other smaller fixes.


alibuild (approved):

PR #17416, opened 2026-08-11
[PWGDQ] add dimuon analysis in global alignment task

aferrero2707:
Added comparison of dimuon properties (invariant mass, angles, DCA) for different alignment configurations.

### DIFF
--- a/PWGDQ/Tasks/muonGlobalAlignment.cxx
+++ b/PWGDQ/Tasks/muonGlobalAlignment.cxx
@@ -18,12 +18,14 @@
 #include "Common/CCDB/EventSelectionParams.h"
 #include "Common/CCDB/RCTSelectionFlags.h"
 #include "Common/Core/RecoDecay.h"
+#include "Common/Core/fwdtrackUtilities.h"
 #include "Common/DataModel/EventSelection.h"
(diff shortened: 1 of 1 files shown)


aferrero2707:
@iarsene @mcoquet642 @XiaozhiBai @mguilbau this PR is needed quite urgently to analyze the impact of the different alignments on the dimuon invariant mass. Could any of you please have a look?

There are few remaining code-check errors that I need to look in detail:
- `the parameter 'mftTrackPar' is copied for each invocation` is a genuine one, but it is harmless. I will fix it in the next PR.
- `do not use static_cast to downcast from a base to a derived class` refers to [this line](https://github.com/AliceO2Group/O2Physics/blob/5b4bf8c3f19d1061b04864bd4034e51dbd551e74/PWGDQ/Tasks/muonGlobalAlignment.cxx#L388), which is also present in other sources. I propose to fix all of them in one single dedicated PR.
- `included header Vector3D.h is not used directly` seems to be a false positive, since the header is needed to fix the other error `no header providing "ROOT::Math::XYZVector" is directly included`
- same for `Vector4D.h` and `ROOT::Math::PxPyPzMVector`

Thanks a lot!

mcoquet642 (approved):

PR #17419, opened 2026-08-11
[Tools] Add PID feature extractor and ONNX inference tasks

forynski:
Adds two new tasks under Tools/PIDFeatureExtractor/:

- pidFeatureExtractor.cxx: extracts a 34-feature / 7-detector-group PID
  feature set (TPC, TOF, TRD, ITS, EMCal, HMPID, centrality) from AO2D
  data for both MC (reconstructed + truth) and real/raw data, writing a
  ROOT TTree (and optionally CSV). Includes optional DPG-style track
  cuts (off by default) and an optional Bayesian PID posterior for
  comparison against the ML model.

- pidOnnxInference.cxx: runs a trained ONNX model over the extractor's
  output and writes per-track class probabilities, using
  o2::analysis::MlResponse (Tools/ML) for CCDB/local model loading.

Notes for reviewers:
- pidFeatureExtractor.cxx writes a plain TFile/TTree rather than an AOD
  table (DECLARE_SOA_TABLE/Produces<>). This is deliberate: two
  Produces<> tables sharing a column prefix in one struct triggered a
  reproducible compile failure against the current O2 build
  (ASoA.h/MetadataTrait constraint errors). Manual ROOT I/O avoids it.
- pidOnnxInference.cxx is a plain DataProcessorSpec, not an
  AnalysisTask - it has no AOD subscription (it reads the extractor's
  output file directly), and runs standalone after the extractor
  rather than chained in the same pipeline.
- The model's expected input feature order (in pidOnnxInference.cxx) is
  a reasonable default but not yet verified against the actual training
  code - flagged in the README.
- modelPathsCcdb currently holds a placeholder CCDB path; needs a real
  location once the model is deployed.

See README.md in the new folder for full usage and configuration
details.


ktf (dismissed):
Approving to be able to run tests.

ktf (approved):

ktf (approved):

dsekihat (approved):

PR #17424, opened 2026-08-12
[PWGCF] Femto: add Lc support to the femto framework

IgorPtak:
Adds Lc baryons as a new particle species to the unified femto framework, enabling
Lc-hadron femtoscopic correlations. 

**Data model** - four new tables:
- `FLcs` - kinematics, mass of the accepted hypothesis and the three daughter indices
- `FLcMasks` - selection bitmask
- `FLcExtras` - QA/debug: mass of the competing hypothesis, three-prong topological
  variables, BDT scores and the PWGHF selection flags
- `FLcLabels` - MC labels

**Core:** `charmHadronBuilder.h` and `charmHadronHistManager.h` extended to the
three-prong topology, so both species share the same builder and histogram manager.
Lc candidates are taken from PWGHF (`HfCand3Prong`), with the acceptance applied as
a rapidity cut via `HfHelper`. Unlike D0, the sign of the stored signed pT is the
physical baryon charge, taken from the charge of the first prong - the PWGHF flags
only tell which mass assignment passed. Prongs are stored remapped onto the accepted
hypothesis, so `Prong0Dau` always points at the proton.

The naming of everything shared by the family was generalised from D0 to charm
hadron (`ConfD0Filters` -> `ConfCharmHadronFilters`, `fillD0s` -> `fillCharmHadrons`,
and so on). Species-level names (`D0Bits`, `D0Selection1`) are unchanged.

**Tasks:** the two D0 workflows now cover the whole family. 
- `femto-charm-hadron-qa` (was `femto-d0-qa`) - QA for D0 and Lc candidates
  and their daughters
- `femto-pair-track-charm-hadron` (was `femto-pair-track-d0`) - D0-hadron and
  Lc-hadron correlations, same and mixed event

JSON prefixes changed accordingly: `D0Filters` -> `CharmHadronFilters`,
`D0Tables` -> `CharmHadronTables`.

**Pairs:** `TrackLcPairCleaner` and `ClosePairRejectionTrackLc` are new. 

**MC:** Lc candidates have no direct MC label, so the three prongs are
matched to a generated Lc -> p K pi decay with `RecoDecay::getMatchedMCRec`.

**Mass hypotheses:** an Lc candidate can pass both the pKpi and the piKp hypothesis.
One row is stored per accepted hypothesis, so such a candidate appears twice, each
row with its own mass and prong assignment. `CharmHadronFilters.rejectAmbiguousHypothesis` allows dropping them entirely.


PWGCF/Femto/Core/pairBuilder.h
```diff
@@ -653,7 +653,7 @@ class PairD0D0Builder
     mD0Cleaner1.init(confD0Cleaner1);
     mD0Cleaner2.init(confD0Cleaner2);
 
-    mD0HistManager1.template init<modeSe>(registry, D0HistSpec1, confD0Selection1, PosDauHistSpec1, NegDauHistSpec1);
+    mD0HistManager1.template init<modeSe>(registry, D0HistSpec1, confD0Selection1, PosDauHistSpec1, NegDauHistSpec1, NegDauHistSpec1);
```
ariedel-cern, line 656: I guess you add the negative daughter twice to match the signature of the new function. I would define two functions with different signatures for the 2 and 3 pronged candidates. This, I think, will be very confusing in the future.

PWGCF/Femto/TableProducer/femtoProducer.cxx
```diff
@@ -418,14 +440,28 @@ struct FemtoProducer {
   template <modes::System system, typename T1, typename T2, typename T3>
   void processD0s(T1 const& col, T2 const& tracks, T3 const& candidates)
```
ariedel-cern, line 441: Now looking at the code again, I think it is a bit wasteful to have a process function just for D0s and Lcs (I think we already have more than enough). Can you remove them and have at least processTracks{D0S,Lcs}?

ariedel-cern (changes requested):
I left some small comment. Overall, everything looks good. I think the code-check posted one warning with regards to an used header.

PWGCF/Femto/TableProducer/femtoProducer.cxx
IgorPtak: Do you mean mergin D0s and Lcs into processCharmHadrons? D0 and Lc come from different input tables, so  merging them would make the DPL signature require both tables in every AO2D. Is there any solution to prevent this?

ariedel-cern: No merge, you're right that combining them would force both input tables into every AO2D. I meant just dropping the standalone processD0s and processLcs and keeping processTracksD0s and processTracksLcs.
Actually, now I see there's a D0-D0 correlation task, so processD0s has a use case on its own. Just dropping processLcs makes more sense, unless you're planning an Lc-Lc task I'm not aware of?

ariedel-cern (approved):

PR #17430, opened 2026-08-12
[PWGHF] Fix unused headers in HF tree creators

fchinu:


vkucera:
@fchinu Thanks!

fchinu:
The error on `channelsResonant` is related to #11382 and/or #13206, and will be fixed in another PR, as agreed with @vkucera

xinyepeng (approved):

PR #17431, opened 2026-08-12
[PWGCF] Femto: Add optional pT-dependent CPA selection for cascades

IgorPtak:
Adds an optional pT-dependent minimum CPA selection for cascades in `PWGCF/Femto/Core/cascadeBuilder.h`.

Two new configurables in `CASCADE_DEFAULT_BITS`:
- `cascadeCpaPtDependent` (bool, default `false`) - switches between the constant and the pT-dependent threshold
- `cascadeCpaMinFormula` (vector<string>, default `{"0.95"}`) - minimum CPA as a TFormula of pT (`x` = pT), used only when the flag is enabled

### DIFF
--- a/PWGCF/Femto/Core/cascadeBuilder.h
+++ b/PWGCF/Femto/Core/cascadeBuilder.h
@@ -65,7 +65,8 @@ struct ConfCascadeFilters : o2::framework::ConfigurableGroup {
 // NOLINTNEXTLINE(cppcoreguidelines-macro-usage)
 #define CASCADE_DEFAULT_BITS                                                                                                                                           \
   o2::framework::Configurable<bool> passThrough{"passThrough", false, "If true, all Cascades are passed through. Bits for all selections are stored."};                \
-  o2::framework::Configurable<std::vector<float>> cascadeCpaMin{"cascadeCpaMin", {0.95f}, "Minimum cosine of pointing angle"};                                         \
+  o2::framework::Configurable<std::vector<std::string>> cascadeCpaMin{"cascadeCpaMin", {"0.95"}, "Minimum CPA as TFormula, x=pt"};                                     \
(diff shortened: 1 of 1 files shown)


ariedel-cern:
Hi @IgorPtak ,
many thanks for this PR. I think the easiest way is to just promote the cut on CPA to a functional cut, and if some analyzers prefer a constant cut, they can just set a constant value.
In principle, we can then make a change to the SelectionContainer class, where we can check at instantiation if a function cut is a constant (like we could check if the string has an "x," for example), flip a flag, and then the updateLimits function becomes a NoOp. Though this should probably be its own PR.
Just one follow-up: I think there was a suggestion to also be able to cut directly on the Pointing angle, not just its cosine. Did you also intend to have this functionality with this PR?

ariedel-cern (changes requested):
See my previous comment

IgorPtak:
> Hi @IgorPtak , many thanks for this PR. I think the easiest way is to just promote the cut on CPA to a functional cut, and if some analyzers prefer a constant cut, they can just set a constant value. In principle, we can then make a change to the SelectionContainer class, where we can check at instantiation if a function cut is a constant (like we could check if the string has an "x," for example), flip a flag, and then the updateLimits function becomes a NoOp. Though this should probably be its own PR. Just one follow-up: I think there was a suggestion to also be able to cut directly on the Pointing angle, not just its cosine. Did you also intend to have this functionality with this PR?

Sure, I will update the PR to promote the CPA cut to a functional cut. Regarding the `SelectionContainer` change, would you like me to open a follow-up PR for that, or would you prefer to handle it? You're right about the direct Pointing angle cut. I'll add that to this PR.

ariedel-cern:
I will handle the changes in the SelectionContainer.

ariedel-cern (approved):

PR #17443, opened 2026-08-13
[PWGCF] flowGenericFramework.cxx: fix missing semi-colon in ConfigurableGroup…

EmilGorm:
… inheritance
- Accidentally shadows later ConfigurableGroups, causing them to inherit from the Wrong type


alibuild (approved):
Auto-approving on behalf of @EmilGorm.

PR #17461, opened 2026-08-14
[PWGDQ] fix MFT momentum rescaling

aferrero2707:
The MCH tracks are extrapolated to the first measured MFT point using `TrackExtrap::extrapToVertex()` instead of `TrackExtrap::extrapToVertexWithoutBranson()`, which cures large inaccuracies in the MCH momentum estimation at the MFT front for some tracks.
Those inaccuracies were contributing to high-mass tails in the di-muon invariant mass  forward tracks with rescaled momentum.

Below is a comparison of the J/psi invariant mass before and after the fix, estimated from the same AO2Ds of LHC25i4 (OO MC simulation with injected J/psi and psi(2S)).

Before:
image

After:
image

### DIFF
--- a/PWGDQ/Tasks/muonGlobalAlignment.cxx
+++ b/PWGDQ/Tasks/muonGlobalAlignment.cxx
@@ -41,7 +41,6 @@
 #include <Framework/InitContext.h>
 #include <Framework/runDataProcessing.h>
 #include <GPU/GPUROOTCartesianFwd.h>
-#include <GlobalTracking/MatchGlobalFwd.h>
 #include <MCHBase/TrackerParam.h>
 #include <MCHGeometryTransformer/Transformations.h>
 #include <MCHTracking/Track.h>
@@ -348,8 +347,8 @@ struct muonGlobalAlignment { // o2-linter: disable=name/workflow-file,name/struc
       base::Propagator::initFieldFromGRP(grpmag);
       TrackExtrap::setField();
       TrackExtrap::useExtrapV2();
-      fieldB = static_cast<o2::field::MagneticField*>(TGeoGlobalMagField::Instance()->GetField()); // for MFT
-      std::array<double, 3> centerMFT{0, 0, -61.4};                                                // or use middle point between Vtx and MFT?
+      fieldB = dynamic_cast<o2::field::MagneticField*>(TGeoGlobalMagField::Instance()->GetField()); // for MFT
+      std::array<double, 3> centerMFT{0, 0, -61.4};                                                 // or use middle point between Vtx and MFT?
       mBzAtMftCenter = fieldB->getBz(centerMFT.data());
     } else {
       LOGF(fatal, "GRP object is not available in CCDB at timestamp=%llu", bc.timestamp());
@@ -1301,8 +1300,8 @@ struct muonGlobalAlignment { // o2-linter: disable=name/workflow-file,name/struc
                                 collision.posX(),
                                 collision.posY(),
                                 collision.posZ(),
-                                collision.covXX(),
-                                collision.covYY());
+                                std::sqrt(collision.covXX()),
+                                std::sqrt(collision.covYY()));
   }
 
   template <class TMFT>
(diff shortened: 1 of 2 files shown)


mcoquet642:
why not use the common tool for rescaling introduced in `fwdtrackUtilities.h` ?

aferrero2707:
> why not use the common tool for rescaling introduced in `fwdtrackUtilities.h` ?

Mostly to have the freedom of experimenting without affecting the common tools used in analysis.
For this specific case, if I understand correctly one could use `refitGlobalMuonCov()` for the momentum rescaling, right? However, the issue is in the determination of the MCH momentum to be used for the rescaling, so the common function would not help.

Once the fix is confirmed we can see how to properly propagate it to the common code. What do you think?

mcoquet642:
> > why not use the common tool for rescaling introduced in `fwdtrackUtilities.h` ?
> 
> Mostly to have the freedom of experimenting without affecting the common tools used in analysis. For this specific case, if I understand correctly one could use `refitGlobalMuonCov()` for the momentum rescaling, right? However, the issue is in the determination of the MCH momentum to be used for the rescaling, so the common function would not help.
> 
> Once the fix is confirmed we can see how to properly propagate it to the common code. What do you think?

Indeed `refitGlobalMuonCov` does not fix the choice of the rescaling momentum (which indeed should be computed with Branson correction as you point out). But I think it would be good to be consistent with how the parameters are computed from the provided inputs. The one implemented in `refitGlobalMuonCov`  has been extensively tested and is used by both EM and DQ (btw they also already use Branson correction to compute the rescaling of the momentum)

aferrero2707:
@mcoquet642 I have replaced my custom code by `refitGlobalMuonCov()` in all places where I extrapolate the global tracks to the vertex. I am still using my simpler `UpdateTrackMomentum()` in a couple of other places, I will look how to properly replace it at a later stage.

I have also fixed all the code-check errors (except the two remaining ones that really look like false positives). There are still several warnings, but I propose to fix them in a separate PR to avoid making lots of cosmetic changes in this one.

@iarsene @mcoquet642 @XiaozhiBai @mguilbau given that the changes here are relevant and needed for all the ongoing muon alignment studies, could you please have a look and merge if there are no more objections?

Thanks a lot!

mcoquet642 (approved):

PWGDQ/Tasks/muonGlobalAlignment.cxx
```diff
@@ -1449,24 +1458,20 @@ struct muonGlobalAlignment { // o2-linter: disable=name/workflow-file,name/struc
...
 
-    o2::mch::TrackExtrap::extrapToVertex(mftTrackProp,
-                                         collision.posX(),
-                                         collision.posY(),
-                                         collision.posZ(),
-                                         collision.covXX(),
-                                         collision.covYY());
+    auto geoMan = o2::base::GeometryManager::meanMaterialBudget(fwdTrackProp.getX(), fwdTrackProp.getY(), fwdTrackProp.getZ(), collision.posX(), collision.posY(), collision.posZ());
+    auto x2x0 = static_cast<float>(geoMan.meanX2X0);
+    fwdTrackProp.propagateToVtxhelixWithMCS(collision.posZ(), {collision.posX(), collision.posY()}, {collision.covXX(), collision.covYY()}, mBzAtMftCenter, x2x0);
```
mcoquet642, line 1472: @aferrero2707 Sorry, I missed this before merging. I would recommend to not use this method, unless you properly assessed its performance and are happy with it. It is not used for analysis because it was shown in the past to be unreliable and to artificially distort the eta distribution.

PWGDQ/Tasks/qaMatching.cxx
```diff
@@ -1678,24 +1678,19 @@ struct QaMatching {
...
-
-    o2::mch::TrackExtrap::extrapToVertex(mftTrackProp,
-                                         collision.posX(),
-                                         collision.posY(),
-                                         collision.posZ(),
-                                         collision.covXX(),
-                                         collision.covYY());
+    auto geoMan = o2::base::GeometryManager::meanMaterialBudget(fwdTrackProp.getX(), fwdTrackProp.getY(), fwdTrackProp.getZ(), collision.posX(), collision.posY(), collision.posZ());
+    auto x2x0 = static_cast<float>(geoMan.meanX2X0);
+    fwdTrackProp.propagateToVtxhelixWithMCS(collision.posZ(), {collision.posX(), collision.posY()}, {collision.covXX(), collision.covYY()}, mBzAtMftCenter, x2x0);
```
mcoquet642, line 1691: same

PWGDQ/Tasks/muonGlobalAlignment.cxx
aferrero2707: Thanks for pointing this out, I will replace it with the method that is used in the latest version of the forward track utilities.

PWGDQ/Tasks/qaMatching.cxx
aferrero2707: See answer above.

PR #17488, opened 2026-08-18
[PWGCF] Femto: add optional kT-dependent dphistar cut in CPR

IgorPtak:


ariedel-cern:
Hi @IgorPtak ,
again, many thanks for this PR. I think we can add here a bit more general functionality.
Some other analyzers might like to cut on kstar or Q3, rather than kT, and giving kT a special status seems not very general. I would propose to declare an enum, like
```
enum class CprCutKinematicVariable : int8_t {
  kKstar=0,
  kKt = 1,
  KMt = 2,
  kQ3 = 3
};

```
in modes.h such that, in a configurable, the analyzers can then set an integer indicating which kinematic variable they want to cut on. Then we also have to rework the fill function of the CPR manager. Here, we can, in principle, pass the PairHistManager by reference instead of a float and then call a specific getter; there is no need to recompute any pair property. And to keep the interface simple, we can generally set dphistar to a formula (maybe we should also do the same for deta to be symmetric?).
I think there was also a suggestion at some point to allow for a rectangular cut (instead of an ellipsoidal one as what we have now). Was this also your intention to allow for that with this change?

ariedel-cern (changes requested):
see previous comment

IgorPtak:
> Hi @IgorPtak , again, many thanks for this PR. I think we can add here a bit more general functionality. Some other analyzers might like to cut on kstar or Q3, rather than kT, and giving kT a special status seems not very general. I would propose to declare an enum, like
> 
> ```
> enum class CprCutKinematicVariable : int8_t {
>   kKstar=0,
>   kKt = 1,
>   KMt = 2,
>   kQ3 = 3
> };
> ```
> 
> in modes.h such that, in a configurable, the analyzers can then set an integer indicating which kinematic variable they want to cut on. Then we also have to rework the fill function of the CPR manager. Here, we can, in principle, pass the PairHistManager by reference instead of a float and then call a specific getter; there is no need to recompute any pair property. And to keep the interface simple, we can generally set dphistar to a formula (maybe we should also do the same for deta to be symmetric?). I think there was also a suggestion at some point to allow for a rectangular cut (instead of an ellipsoidal one as what we have now). Was this also your intention to allow for that with this change?

I'll add CprCutKinematicVariable to modes.h and expose it as a configurable, and make deta a formula too for symmetry.

On passing the PairHistManager: agreed. Two things I hit: only getKstar() exists, so getKt()/getMt() need adding and isClosePair() is called before PairHistManager.setPair(), so the values aren't computed yet.

The rectangular cut wasn't in my original scope but I'm happy to add it.

PR #17508, opened 2026-08-19
[PWGCF] Fix derived to derived processing of femto tracks with mass

ariedel-cern:
The FemtoTracksWithMass partitions in femtoProducerDerivedToDerived.cxx were declared incorrectly, so processTracksWithMass was broken.


alibuild (approved):

PR #17512, opened 2026-08-19
[PWGHF] Add pT-dependent invariant mass cut for Lc->pKpi

fgrosa:


alibuild (approved):

PR #17518, opened 2026-08-19
[PWGHF] Lc decay time: fix definition and provide additional info

lubynets:
**1. Fix definition of MC-gen decay time**
In the current version of code the decay time of MC generated $\Lambda_{\mathrm{c}}^+$ particle is defined as $t^{\mathrm{gen}} = t / \gamma$, where $t$ is the lab-frame production time of one of the daughters, and $\gamma$ is the Lorentz factor of $\Lambda_{\mathrm{c}}^+$.
This PR proposes to define the decay time of MC generated $\Lambda_{\mathrm{c}}^+$ particle as $t^{\mathrm{kin}} = \frac{Lm}{pc}$, where $L$ and $p$ are MC-defined decay length and momentum of the mother particle.
For prompt $\Lambda_{\mathrm{c}}^+$ there should be no difference between these two definitions. The existing difference (left panels of the attached picture) are probably related to numeric artifacts and/or neglecting the curvature of $\Lambda_{\mathrm{c}}^+$ trajectory (?)
For nonprompt $\Lambda_{\mathrm{c}}^+$ the difference is more pronounced (right panels of the picture). The motivation to use the $t^{\mathrm{kin}}$ definition is that we are interested not in real proper decay time of nonprompt particle, but in its proxy, the quantity which we would obtain in data-driven way, _if this particle was prompt_ (since for reconstructed decay candidates both in MC and data we use exactly this definition).
out
**2.** The THnSparse with reconstructed MC candidates is extended with yet another axis, containing decay time of matched generated particle (for obtaining the response matrix in order to enable unfolding / forward-folding).
**3.** The name `lifetime` in code variables is replaced with `decay time`, since the former is the property of particle species (PDG constant), while the latter is the property of particular particle instance.
**4. Code cleaning**
-- unneeded template typename replaced with explicit specification of parameter's type;
-- some variables are named less cryptic;
-- fixed clang-tidy errors.

### DIFF
--- a/PWGHF/D2H/Tasks/taskLc.cxx
+++ b/PWGHF/D2H/Tasks/taskLc.cxx
@@ -338,6 +338,10 @@ struct HfTaskLc {
     return o2::hf_centrality::getCentralityColl<Coll>(collision);
   }
 
+  /// Evaluate decay time of generated particle
+  /// \param mcParticleProng0 one of generated particle's daughters
+  /// \param motherParticle generated particle
+  /// \return decay time in picoseconds. For nonprompt particles it is evaluated as if it was prompt (as it is calculated for data)
   float evaluateMcGenDecayTime(const McParticles3ProngMatched::iterator& mcParticleProng0, const McParticles3ProngMatched::iterator& motherParticle)
   {
     const auto mcCollision = motherParticle.template mcCollision_as<aod::McCollisions>();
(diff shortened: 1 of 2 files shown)


lubynets:
Dear codeowners, could you kindly review the PR?

zhangbiao-phy:
> Dear codeowners, could you kindly review the PR?
Thanks @lubynets! Looks good to me!

zhangbiao-phy (approved):

PR #17521, opened 2026-08-20
[Infrastructure,Common] Fix precompiled header creation

ktf:
Needs to propagate defines to the precompiled header.

Why it now depends on rANS, yet to be understood.


ktf:
Seems to fix compilation on linux. Currently broken. Merging.

PR #17522, opened 2026-08-20
[ALICE3] Update ALICE3 centrality task

axti98:
- Make CCDB timestamp configurable
- Replace for loop and if statements with Filter + .size() call
- Remove LOG call 
- Fix O2Linter errors

### DIFF
--- a/ALICE3/TableProducer/alice3Centrality.cxx
+++ b/ALICE3/TableProducer/alice3Centrality.cxx
@@ -33,14 +33,13 @@
 #include <TString.h>
 
 #include <cmath>
+#include <cstdint>
 #include <cstdlib>
 #include <string>
 
 using namespace o2;
 using namespace o2::framework;
 using namespace o2::framework::expressions;
-
-constexpr float DefaultMult = 10000.0f;
 
 struct Alice3Centrality {
   Produces<aod::CentRun2V0Ms> cent;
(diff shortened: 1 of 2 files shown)


ALICE3/TableProducer/alice3Centrality.cxx
```diff
@@ -31,68 +34,66 @@
...
+
+  Filter trackFilter = (aod::track::eta >= minEta) && (aod::track::eta <= maxEta) && (nabs(aod::track::dcaXY) <= maxDCA) && (nabs(aod::track::dcaZ) <= maxDCA);
 
   bool centralityLoaded = false;
   TH1D* hCumMultALICE3 = nullptr;
 
   void init(InitContext&)
   {
-    const AxisSpec axisMult{MaxMult.value > 10000.f ? 10000 : (int)MaxMult, 0, MaxMult, "Reconstructed tracks"};
+    const AxisSpec axisMult{maxMult.value > DefaultMult ? static_cast<int>(DefaultMult) : static_cast<int>(maxMult), 0, maxMult, "Reconstructed tracks"};
```
jesgum: Might be more convenient with ConfigurableAxis

ALICE3/TableProducer/alice3Centrality.cxx
```diff
@@ -31,68 +34,66 @@
...
     ccdb->setLocalObjectValidityChecking();
   }
 
-  int nevs = 0;
-  void process(const o2::aod::Collision& collision, const soa::Join<aod::Tracks, aod::TracksDCA>& tracks)
+  void process(const o2::aod::Collision& collision, const soa::Filtered<soa::Join<aod::Tracks, aod::TracksDCA>>& tracks)
   {
     if (!centralityLoaded) {
-      hCumMultALICE3 = ccdb->getForTimeStamp<TH1D>("Analysis/ALICE3/Centrality", 1);
+      hCumMultALICE3 = ccdb->getForTimeStamp<TH1D>("Analysis/ALICE3/Centrality", ccdbNoLaterThan.value);
```
jesgum: I would suggest to make the ccdb path a configurable as well, since it might be a bit more convenient if new calibrations are produced (it might be annoying to always try and upload them to `Analysis/ALICE3/Centrality` as everyone might not have the permissions and instead one could just use their own directory)

jesgum (commented):
Hi @axti98 ! Thanks a lot for the PR! I just have some minor suggestions regarding the code itself.
I also have a general comment regarding the calibration object that is currently in `Analysis/ALICE3/Centrality`. This looks to be created with v2 in mind (looks like `|eta| < 4` is used), so it would probably have to be updated. Also, it is not clear to me, just from inspecting it, what generator was used (the calibration object would be different depending on the generator and the generator settings). Something to consider is if we should have some form of safeguard for this

ALICE3/TableProducer/alice3Centrality.cxx
axti98: Hi Jesper, yes I agree with your concerns. So the Object behind Analysis/ALICE3/Centrality is just the centrality percentile versus multiplicity, so I will make this path a configurable and upload my own calibration based on PYTHIA

jesgum (approved):

jesgum:
Approving from my side, leaving it for @njacazio to have a look and merge

PR #17523, opened 2026-08-20
[PWGCF,PWGHF] Add MC information to table for single charm had in femto tasks

fgrosa:
For info @zhangbiao-phy @FDUEnrich


zhangbiao-phy:
> For info @zhangbiao-phy @FDUEnrich

Thanks @fgrosa! actually we have it for derived table of charm pairs, so we just need do mix-event online for MC, but It's fine you added it, then we can do everything based on this framework without runing the D2H task now

fgrosa:
> > For info @zhangbiao-phy @FDUEnrich
> 
> Thanks @fgrosa! actually we have it for derived table of charm pairs, so we just need do mix-event online for MC, but It's fine you added it, then we can do everything based on this framework without runing the D2H task now

Yes indeed the idea was to use it to do everything locally, to have more flexibility for the ME :)

lauraser (approved):

alibuild (approved):

PR #17527, opened 2026-08-20
[PWGCF] Add pointing angle column+histograms to FemtoV0s and FemtoCascades

ariedel-cern:
Add a dynamic column to compute the pointing angle from the cosine of the pointing angle stored in the extra tables of v0s and cascades. Similarly, added QA histograms to V0 and cascade histManager.


alibuild (approved):

PR #17535, opened 2026-08-21
[PWGDQ] fix match type assignment in ML training data producer

aferrero2707:
The match type assigned to candidate MFT-MCH matches in the ML training data producer task was wrongly set to "fake" in cases where the MCH track is pairable but the corresponding MFT track does not have an associated MC particle. In such cases the match type is now correrctly assigned as " wrong".


aferrero2707:
@iarsene @mcoquet642 @XiaozhiBai @mguilbau this is a bug fix for the task that produces the ML training input data for the MFT-MCH matching. I have split the PR in two commits, the first with the actual bugfix and the second with the unrelated linter fixes.

Could you please have a look? Thanks!

mcoquet642 (approved):

PR #17536, opened 2026-08-21
[PWGLF] Move double-cascade inspector to double-omega search analysis

fmazzasc:


romainschotter (changes requested):
Hi @fmazzasc !
I am not entirely sure to understand why it is necessary to re-build the Lambda and Omega candidates. Why not using the V0 and cascade tables produced by the strangeness builder?
Could you please also add:
- #include <string> in doubleOmegaTreeCreator.cxx?
Thank you very much!

romainschotter:
As discussed privately with @fmazzasc , I understand now that the Lambda and Omega had to be rebuilt because, somehow, the double-Omega signal events were lost in the strangeness building of the propagationService.

However, we should definitely look into it because there are no preselections in the strangeness builder in the propagation service, that should suppress double Omegas. In order to dive deeper into this, it was agreed with @fmazzasc that a switch should be added later this week to use the V0s/cascades from the strangeness builder in the propagation service.

In the meantime, I merge the PR.

Thank you very much!

romainschotter (approved):

PR #17540, opened 2026-08-21
[ALICE3] Added Multiplicity estimator task for ALICE3

axti98:
Added basic multiplicity estimator task in preparation for PbPb studies and multiplicity dependent studies. Currently, two estimator types are implemented:
- Global Tracks: Any track with certain DCA, nSiliconHits and reconstruction flag. Properties are configurable
- PVContributors: Any track that has contributed to the fit of the primary vertex. For now, this is actually every track with isReconstructed = true, since the full track array is passed to the computeVertex function. Might be adjusted later.
- MCTruth estimation: All charged physical primaries from the generator and in the |eta| intervals 2.5, 1.25, and 0.9

Necessary changes in the TrackingTranslator because of data model change as well as in OTF Tracker


ALICE3/TableProducer/OTF/onTheFlyTracker.cxx
```diff
@@ -1776,7 +1776,7 @@ struct OnTheFlyTracker {
                               trackParCov.getSigmaTgl2(), trackParCov.getSigma1PtY(), trackParCov.getSigma1PtZ(), trackParCov.getSigma1PtSnp(), trackParCov.getSigma1PtTgl(),
                               trackParCov.getSigma1Pt2());
       tableMcTrackLabels(trackParCov.mcLabel, 0);
-      tableTracksExtraA3(trackParCov.nSiliconHits, trackParCov.nTPCHits, trackParCov.trackType);
+      tableTracksExtraA3(trackParCov.nSiliconHits, trackParCov.nTPCHits, trackParCov.trackType, true);
```
njacazio: shouldn't it be only if the track is a contributor?

ALICE3/TableProducer/alice3Multiplicity.cxx
```diff
@@ -0,0 +1,180 @@
...
+    int numMCParticlesEta09 = 0;
+
+    for (const auto& mcParticle : mcParticles) {
+      if (!mcParticle.isPhysicalPrimary()) {
+        continue;
+      }
+
+      auto charge = 0.;
+      auto* p = pdg->GetParticle(mcParticle.pdgCode());
+      if (p != nullptr) {
```
njacazio, line 140: for counting particles we can use https://github.com/AliceO2Group/O2Physics/blob/master/PWGLF/Utils/inelGt.h#L96

ALICE3/TableProducer/alice3Multiplicity.cxx
```diff
@@ -0,0 +1,180 @@
...
+      histos.fill(HIST("multiplicity/nTracksMC"), numMCParticles);
+      histos.fill(HIST("multiplicity/nTracksMCEta25"), numMCParticlesEta25);
+      histos.fill(HIST("multiplicity/nTracksMCEta125"), numMCParticlesEta125);
+      histos.fill(HIST("multiplicity/nTracksMCEta09"), numMCParticlesEta09);
+    }
+
+    multMC(numMCParticles, numMCParticlesEta25, numMCParticlesEta125, numMCParticlesEta09);
+  }
+
+  void processDummy(const aod::Collision&)
```
njacazio, line 166: why do you need the dummy?

axti98: Its sometimes more convenient on hyperloop to have a dummy process so you don't need to remove the task all the time

ALICE3/TableProducer/alice3Multiplicity.cxx
axti98: I count it in several eta intervals, I would not use this function since then we have to loop again for each eta interval over all tracks which is quite inefficient?

ALICE3/TableProducer/OTF/onTheFlyTracker.cxx
axti98: Yes, I think I misunderstood something earlier. I have implemented it now that every track that contributed has a flag now which is checked in the table filling

njacazio (approved):

PR #17505, opened 2026-08-19
[PWGJE] fix code-check issues in QG ML producer

mulewicz:
Follow-up to #17400 fixing the code-check issues reported after the merge.

Changes:
- add the direct `Framework/AnalysisHelpers.h` include required for `Produces`
- mark `hasITSLayerHit` as `[[nodiscard]]`

The O2 linter passes locally.

PR #17550, opened 2026-08-21
[PWGJE] Fix DECLARE SOA INDEX

mulewicz:


PWGJE/TableProducer/quarkGluonJetsProducer.cxx
nzardosh, line 246: this was already fixed here wasnt it?
https://github.com/AliceO2Group/O2Physics/pull/17505/changes

Not sure why its showing up agains as a change. @vkucera do you know?

nzardosh (approved):

PWGJE/TableProducer/quarkGluonJetsProducer.cxx
vkucera: Because the PR branch does not contain the other previously merged PR. I would suggest to rebase on top of the latest upstream master.

mulewicz: done, thank you

PWGJE/TableProducer/quarkGluonJetsProducer.cxx
```diff
@@ -94,7 +94,7 @@ using QGMLJet = QGMLJets::iterator;
 
 namespace qgmlconst
 {
-DECLARE_SOA_INDEX_COLUMN(QGMLJet, qGMLJet);
+DECLARE_SOA_INDEX_COLUMN(QGMLJET, qGMLJet);
```
vkucera, line 97: What are you trying to fix?

mulewicz: When I try to run on the Hyperloop I get the error: Index on O2qgmljet but no tree found

vkucera: This is not going to fix it. You need the name of the iterator here which was correct. Your change does not even compile which is the minimum to test before you even commit your changes locally. Your problem on Hyperloop is caused by something else. Maybe you don't save the referenced table in the output which makes the merging step fail.

nzardosh: Sorry this was my bad advice, indeed DECLARE_SOA_INDEX_COLUMN(QGMLJet, qGMLJet); is the correcr syntax

nzardosh:
Hi @mulewicz let me know when you have updated the PR and it is ready to go :)

PR #17566, opened 2026-08-23
[PWGCF] Add systematic variation export to cutculator gui

ariedel-cern:


ariedel-cern:
MegaLinter wants the executable bit set because of the shebang, but flipping the file mode to 755 makes the clang-format check fail. I'd rather keep the mode as is.

alibuild (approved):

PR #17572, opened 2026-08-24
[Common] Improve track selection

ktf:
* Avoid redundant IsRun2 check
* Avoid doing IsSelected() twice
* Use std::pow (float) rather than C pow (double)


ktf:
@dsekihat can you have a look here? It should be safe.

dsekihat (approved):

dsekihat:
Dear @ktf , Thank you for the PR. It looks good to me.

PR #17602, opened 2026-08-25
[PWGLF] Add README for strangeness table converters

romainschotter:
- add README for strangeness table converters
- add Python script producing the README file (to be executed every now and then)


alibuild (approved):

PR #17610, opened 2026-08-26
[PWGUD] Shrink 8 flag variables to one variable

amatyja:
Shrink 8 flag variables to one variable
Corrected code-check errors
Changed order of selection cuts


alibuild (approved):

PR #17613, opened 2026-08-26
[PWGCF] Modified the MC reconstructed task

SwatiSaha-1997:
[1] When multiple reconstructed collisions map to the same generated MC collision, keep only the one with the most contributors and discard the rest
[2] Changed the filename to fix o2-linter error in the CMakeLists.txt file

### DIFF
--- a/PWGCF/EbyEFluctuations/Tasks/meanptFluctuations.cxx
+++ b/PWGCF/EbyEFluctuations/Tasks/meanptFluctuations.cxx
@@ -33,7 +33,6 @@
 #include <Framework/InitContext.h>
 #include <Framework/OutputObjHeader.h>
 #include <Framework/runDataProcessing.h>
-#include <ReconstructionDataFormats/PID.h>
 
 #include <TF1.h>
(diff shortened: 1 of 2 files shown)


alibuild (approved):

PWGCF/EbyEFluctuations/Tasks/meanptFluctuations.cxx
```diff
@@ -350,7 +349,7 @@ struct MeanptFluctuationsAnalysis {
   } //! end init function
 
   template <typename TCollision>
-  bool eventSelected(TCollision collision, const int& multTrk, const float& centrality)
+  bool eventSelected(TCollision const& collision, const int& multTrk, const float& centrality)
```
vkucera, line 352: Do not pass small types (`int`, `float`) by `const&`.

SwatiSaha-1997: Thank @vkucera , i will fix this with the new PR.

PR #17625, opened 2026-08-27
[PWGLF] added isMIDDaughters function at processGenJets for making eff correction factor

JimunLee:
Dear experts, 

It's added "isMIDDaughters" function at processGenJets and QA plots at processJetQA also. 

Thanks a lot


JimunLee:
I found a few things I have to fix more, so that I'll cancel to merge this code.

PR #17628, opened 2026-08-27
[PWGLF] Change object name in ccdb and remove string formatting

romainschotter:
- Change object name to be pulled from ccdb to match the object name in the centralityQa task
- Remove string formatting in log messages


alibuild (approved):

PR #17629, opened 2026-08-27
[PWGDQ] Using info from dalitz selected tracks  in dqEfficiency_withAssoc_direct.cxx

glegras:


PWGDQ/Core/CutsLibrary.cxx
```diff
@@ -5471,7 +5458,7 @@ AnalysisCut* o2::aod::dqcuts::GetAnalysisCut(const char* cutName)
 
   for (int i = 1; i <= 8; i++) { // o2-linter: disable=magic-number (number of cuts)
     if (nameStr == Form("dalitzLeg%d", i)) {
-      cut->AddCut(VarManager::kIsDalitzLeg + i - 1, 0.5, 1.5);
+      cut->AddCut(VarManager::kIsDalitzLeg + i - 1, -0.5, 0.5, true);
```
iarsene, line 5461: This might be dangerous, since the order of the variables in the VarManager is not protected by any mechanism

PWGDQ/Core/MCSignalLibrary.cxx
```diff
@@ -522,6 +522,18 @@ MCSignal* o2::aod::dqmcsignals::GetMCSignal(const char* name)
...
+    prong.SetSourceBit(0, MCProng::kPhysicalPrimary);
+    signal = new MCSignal(name, "Electrons from primary pi0 decays", {prong}, {-1});
+    return signal;
+  }
+  if (nameStr == "eSecondaryFromPi0") {
+    MCProng prong(2, {11, 111}, {true, true}, {false, false}, {0, 0}, {0, 0}, {false, false});
+    prong.SetSourceBit(0, MCProng::kPhysicalPrimary, true);
+    signal = new MCSignal(name, "Electrons from secondary pi0 decays", {prong}, {-1});
+    return signal;
+  }
```
iarsene, line 536: I am not sure I fully understand these signals. I think that the decay products of primary pi0's are considered ALICE primary particles. Then, the secondary electron signal is for when you have secondary neutral pions (created from some long lived particle or from material)?

PWGDQ/Core/MCSignalLibrary.cxx
```diff
@@ -885,6 +909,12 @@ MCSignal* o2::aod::dqmcsignals::GetMCSignal(const char* name)
 
   //_________________________________________________________________________________________________________________________
   // LMEE pair signals for LF, same mother
+  if (nameStr == "eeDuplicated") { // check whether we have two tracks pointing to the same MC particle
+    MCProng prong(1, {11}, {true}, {false}, {0}, {0}, {false});
+    prong.SetSourceBit(0, MCProng::kPhysicalPrimary);
+    signal = new MCSignal(name, "duplicated electron", {prong, prong}, {0, 0}); // signal at pair level
+    return signal;
+  }
```
iarsene, line 917: Not sure why is this needed. Don't we have such a signal already ? (i did not check)

PWGDQ/Core/MCSignalLibrary.cxx
glegras: Hi, this is indeed to separate primary electrons from the e from pi0 from strange particle decays to check whether they have an influence in the ITS-TPC matching efficiency or not

PWGDQ/Core/MCSignalLibrary.cxx
glegras: I did not see such signal in the MCSignalLibrary. Maybe I have missed it?

PWGDQ/Core/CutsLibrary.cxx
glegras: I am not sure I understand why it is dangerous. The reson for this change was a bug introduced in the VarManager when Maurice solved some linter errors, for which this variable is not necessarily 1 but can be 2^i. But I can also correct it in the VarManager if necessary

iarsene: You are right, I think i misunderstood the change

iarsene (approved):

PR #17633, opened 2026-08-27
[PWGUD] Split MC table to generated and MCrec ones

amatyja:


alibuild (approved):

PR #17636, opened 2026-08-27
[PWGLF] Add missing converter in CMakeLists.txt

romainschotter:


alibuild (approved):

PR #17638, opened 2026-08-27
[PWGJE] adding optional phi exclusion zone to jet finders

fjonasALICE:
- added two new configurables to all jetfinders phiExclusionMin and phiExclusionMax which will be applied to track level, gen level and jet level. For jet level, phiExclusionMin - R and phiExclusionMax + R is automatically used to avoid edge effects and mistakes by user
- exclusion region will be used to investigate tracking effects at high pt in data vs MC

### DIFF
--- a/PWGJE/Core/JetFinder.cxx
+++ b/PWGJE/Core/JetFinder.cxx
@@ -48,6 +48,11 @@ void JetFinder::setParams()
   jetDef.set_jet_algorithm(algorithm);
   areaDef = fastjet::AreaDefinition(areaType, ghostAreaSpec);
   selJets = fastjet::SelectorPtRange(jetPtMin, jetPtMax) && fastjet::SelectorEtaRange(jetEtaMin, jetEtaMax) && fastjet::SelectorPhiRange(jetPhiMin, jetPhiMax);
+
+  if (phiExclusionMin < 999.0 && phiExclusionMax > -999.0) {
+    // automatically jets around exclusion region
+    selJets = selJets && (!fastjet::SelectorPhiRange(phiExclusionMin - jetR, phiExclusionMax + jetR));
+  }
 }
 
 /// Performs jet finding
--- a/PWGJE/Core/JetFinder.cxx
+++ b/PWGJE/Core/JetFinder.cxx
@@ -49,7 +49,7 @@ void JetFinder::setParams()
   areaDef = fastjet::AreaDefinition(areaType, ghostAreaSpec);
   selJets = fastjet::SelectorPtRange(jetPtMin, jetPtMax) && fastjet::SelectorEtaRange(jetEtaMin, jetEtaMax) && fastjet::SelectorPhiRange(jetPhiMin, jetPhiMax);
 
-  if (phiExclusionMin < 998.0) {
+  if (phiExclusionMin > -998.0) {
     // automatically jets around exclusion region
     selJets = selJets && (!fastjet::SelectorPhiRange(phiExclusionMin - jetR, phiExclusionMax + jetR));
   }
--- a/PWGJE/Core/JetFinder.cxx
+++ b/PWGJE/Core/JetFinder.cxx
@@ -49,7 +49,7 @@ void JetFinder::setParams()
   areaDef = fastjet::AreaDefinition(areaType, ghostAreaSpec);
   selJets = fastjet::SelectorPtRange(jetPtMin, jetPtMax) && fastjet::SelectorEtaRange(jetEtaMin, jetEtaMax) && fastjet::SelectorPhiRange(jetPhiMin, jetPhiMax);
 
-  if (phiExclusionMin < 999.0 && phiExclusionMax > -999.0) {
+  if (phiExclusionMin < 998.0) {
     // automatically jets around exclusion region
     selJets = selJets && (!fastjet::SelectorPhiRange(phiExclusionMin - jetR, phiExclusionMax + jetR));
   }
--- a/PWGJE/Core/JetFinder.h
+++ b/PWGJE/Core/JetFinder.h
@@ -50,6 +50,8 @@ class JetFinder
 
   float phiMin = -1. * M_PI;
   float phiMax = 2. * M_PI;
+  float phiExclusionMin = 999.;
+  float phiExclusionMax = -999.;
   float etaMin = -.9;
   float etaMax = .9;
 
--- a/PWGJE/Core/JetFinder.h
+++ b/PWGJE/Core/JetFinder.h
@@ -50,8 +50,8 @@ class JetFinder
 
   float phiMin = -1. * M_PI;
   float phiMax = 2. * M_PI;
-  float phiExclusionMin = 999.;
-  float phiExclusionMax = -999.;
+  float phiExclusionMin = -999.;
+  float phiExclusionMax = 999.;
   float etaMin = -.9;
   float etaMax = .9;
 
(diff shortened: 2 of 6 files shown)


PWGJE/Core/JetFinder.cxx
```diff
@@ -48,6 +48,11 @@ void JetFinder::setParams()
   jetDef.set_jet_algorithm(algorithm);
   areaDef = fastjet::AreaDefinition(areaType, ghostAreaSpec);
   selJets = fastjet::SelectorPtRange(jetPtMin, jetPtMax) && fastjet::SelectorEtaRange(jetEtaMin, jetEtaMax) && fastjet::SelectorPhiRange(jetPhiMin, jetPhiMax);
+
+  if (phiExclusionMin < 999.0 && phiExclusionMax > -999.0) {
```
nzardosh: make these 998 just to be safe

PWGJE/JetFinders/jetFinder.h
```diff
@@ -139,6 +142,15 @@ struct JetFinderTask {
       jetFinder.phiMin = -1.0 * M_PI;
       jetFinder.phiMax = 2.0 * M_PI;
     }
+    if (phiExclusionMin != 999.0 && phiExclusionMax != 999.0 && phiExclusionMin >= phiExclusionMax) {
```
nzardosh: !=999.0 might be dangerous for a float. I think use < 998.0. Also shouldnt one be negative?

PWGJE/JetFinders/jetFinder.h
```diff
@@ -71,6 +72,8 @@ struct JetFinderTask {
   o2::framework::Configurable<float> trackEtaMax{"trackEtaMax", 0.9, "maximum track eta"};
   o2::framework::Configurable<float> trackPhiMin{"trackPhiMin", -999, "minimum track phi"};
   o2::framework::Configurable<float> trackPhiMax{"trackPhiMax", 999, "maximum track phi"};
+  o2::framework::Configurable<float> phiExclusionMin{"phiExclusionMin", 999, "minimum of phi exclusion region which is applied for tracks and for jets (jetR is added to the exclusion region for jets)"};
+  o2::framework::Configurable<float> phiExclusionMax{"phiExclusionMax", -999, "maximum of phi exclusion region which is applied for tracks and for jets (jetR is added to the exclusion region for jets)"};
```
nzardosh: arent the positive and negative the opposite way around?

nzardosh: or if its because you want to use the min < max inequality?

PWGJE/JetFinders/jetFinder.h
```diff
@@ -184,8 +196,8 @@ struct JetFinderTask {
...
   o2::framework::expressions::Filter mcCollisionFilter = (nabs(o2::aod::jmccollision::posZ) < vertexZCut);
-  o2::framework::expressions::Filter trackCuts = (o2::aod::jtrack::pt >= trackPtMin && o2::aod::jtrack::pt < trackPtMax && o2::aod::jtrack::eta >= trackEtaMin && o2::aod::jtrack::eta <= trackEtaMax && o2::aod::jtrack::phi >= trackPhiMin && o2::aod::jtrack::phi <= trackPhiMax); // do we need eta cut both here and in globalselection?
-  o2::framework::expressions::Filter partCuts = (o2::aod::jmcparticle::pt >= trackPtMin && o2::aod::jmcparticle::pt < trackPtMax && o2::aod::jmcparticle::eta >= trackEtaMin && o2::aod::jmcparticle::eta <= trackEtaMax && o2::aod::jmcparticle::phi >= trackPhiMin && o2::aod::jmcparticle::phi <= trackPhiMax);
+  o2::framework::expressions::Filter trackCuts = (o2::aod::jtrack::pt >= trackPtMin && o2::aod::jtrack::pt < trackPtMax && o2::aod::jtrack::eta >= trackEtaMin && o2::aod::jtrack::eta <= trackEtaMax && o2::aod::jtrack::phi >= trackPhiMin && o2::aod::jtrack::phi <= trackPhiMax && (o2::aod::jtrack::phi <= phiExclusionMin || o2::aod::jtrack::phi >= phiExclusionMax)); // do we need eta cut both here and in globalselection?
```
nzardosh: dont these overwrite the other ones? lets say if the other one is set to the EMCAL phi acceptance does it mean the exclusion one has to be as well?

PWGJE/JetFinders/jetFinderHF.h
```diff
@@ -41,6 +41,7 @@
 #include <fastjet/JetDefinition.hh>
 #include <fastjet/PseudoJet.hh>
 
+#include <stdexcept>
```
nzardosh, line 44: why is this needed?

PWGJE/Core/JetFinder.cxx
fjonasALICE: changed to just check min and use 998

PWGJE/JetFinders/jetFinder.h
fjonasALICE: changes as suggested to only check < 998 for min

PWGJE/JetFinders/jetFinder.h
fjonasALICE: positive and negative have to be swapped like this, because we want that if default values are used that nothing gets rejected, where we in particular by default only select tracks smalles than phiMinExclusion and larger than phiMaxExclusion

PWGJE/JetFinders/jetFinder.h
fjonasALICE: for default configurables, the last part of the added selection always evaluates to true

PWGJE/JetFinders/jetFinderHF.h
fjonasALICE: it is needed to throw a runtime exception for cases where min>max

PWGJE/Core/JetFinder.cxx
```diff
@@ -48,6 +48,11 @@ void JetFinder::setParams()
   jetDef.set_jet_algorithm(algorithm);
   areaDef = fastjet::AreaDefinition(areaType, ghostAreaSpec);
   selJets = fastjet::SelectorPtRange(jetPtMin, jetPtMax) && fastjet::SelectorEtaRange(jetEtaMin, jetEtaMax) && fastjet::SelectorPhiRange(jetPhiMin, jetPhiMax);
+
+  if (phiExclusionMin < 998.0) {
```
nzardosh: if here  you put the min as -999 and max as 999 then you could check 
 if (phiExclusionMin > -998.0) {

This way the users would know that the correct thing is to make the minimum negative

PWGJE/JetFinders/jetFinder.h
nzardosh: or you can add an else after the  condition you added in the init which then swaps them. This way it might be easier for the user?

PWGJE/JetFinders/jetFinder.h
```diff
@@ -139,6 +142,19 @@ struct JetFinderTask {
...
+
+    if (phiExclusionMin > -998.0) {
+      jetFinder.phiExclusionMin = phiExclusionMin;
+      jetFinder.phiExclusionMax = phiExclusionMax;
+      if (phiExclusionMin >= phiExclusionMax) {
+        throw std::runtime_error("Invalid phi exclusion range: require phiExclusionMin < phiExclusionMax when both are set.");
+      }
+    } else {
+      // default case
+      phiExclusionMin = -1.0 * phiExclusionMin;
```
nzardosh: here you can just add the magic numbers of -999 and 999

PWGJE/Core/JetFinder.h
```diff
@@ -50,6 +50,8 @@ class JetFinder
 
   float phiMin = -1. * M_PI;
   float phiMax = 2. * M_PI;
+  float phiExclusionMin = 999.;
```
nzardosh: here also you could swap the min and max

nzardosh (approved):

PR #17639, opened 2026-08-27
[PWGLF] Fixed feed-down and added multiplicity-differential selection

sprasadphy:
Following changes have been made:
1. For Lambda feeddown: MC-matched candidates now also require `isPhysicalPrimary()`, so the efficiency numerator is consistent with the generated denominator. Non-primary candidates are filled into new `hMassFeedDown*pTFlat`, `hFeedDown*PtVsMotherPt` and `hFeedDown*MotherPdg` histograms.

2. Added `aod::CentFT0Ms` for the multiplicity-differential analysis.

3. The configurables are now organised into `ConfigurableGroups`

4. Fixed the code-check issues: removed the unused `PIDResponseTOF` include, initialised the PDG service and loop variables in `estimateFlattenicityFV0MC`, and replaced the C-style arrays with `std::array`. `clang-tidy` is now clean for this file.

### DIFF
--- a/PWGLF/Tasks/Strangeness/lambdak0sflattenicity.cxx
+++ b/PWGLF/Tasks/Strangeness/lambdak0sflattenicity.cxx
@@ -21,9 +21,9 @@
 #include "PWGLF/Utils/inelGt.h"
 
 #include "Common/CCDB/EventSelectionParams.h"
+#include "Common/DataModel/Centrality.h"
 #include "Common/DataModel/EventSelection.h"
 #include "Common/DataModel/Multiplicity.h"
-#include "Common/DataModel/PIDResponseTOF.h"
 #include "Common/DataModel/PIDResponseTPC.h"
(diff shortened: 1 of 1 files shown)


romainschotter (changes requested):
Hi @sprasadphy !
I had a look at the PR and it looks fine!
I just would like to draw your attention to fact that, because some configurables have been grouped into configurable groups, the configurables will be reset to their default values when resyncing the wagon in Hyperloop (Hyperloop sees it as you removed the configurable and created a new one). This means that you will have to change each configurable to its previous value by hand (which can be a long and painful operation)
I just want to be sure that you are aware of it before I merge these changes :-)
Thank you very much!

sprasadphy:
Hi @romainschotter Thank you for your concern and message. I previously did not plan on grouping them but I was reaching the limit (I think it is 100). I am well aware that I will have to change each configurable to its previous value by hand.

romainschotter (approved):

PR #17640, opened 2026-08-27
[PWGCF] Add process function for tracks with mass and lambdas in femto derived-to-derived producer

ariedel-cern:


alibuild (approved):

PR #17649, opened 2026-08-28
[Infrastructure] Keep into account cgroups / docker memory limits

ktf:
Using the system memory results in overallocating the pool.


ktf:
Currently over-allocating memory. Works as expected in private tests. Merging.

PR #17651, opened 2026-08-28
[PWGJE] Extend MC efficiency for Ds-tagged jets

monamelop:
- Extend the MC collision and jet counters for detector-level, particle-level, and matched jets.
- Add AOD output tables for:
  - detector-level Ds-tagged jets;
  - particle-level Ds-tagged jets;
  - matched detector-level and particle-level jet pairs.
- Store Ds MC origin and candidate MC matching information in the output tables.
- Store the jet matching status from the existing HF jet matching relations.
- Extend the MC-origin histogram axes to preserve the original MC origin categories.

### DIFF
--- a/PWGJE/Tasks/jetDsSpecSubs.cxx
+++ b/PWGJE/Tasks/jetDsSpecSubs.cxx
@@ -13,6 +13,7 @@
 /// \brief Ds-tagged jet analysis with substructure histogram outputs
 /// \author Monalisa Melo <monalisa.melo@cern.ch>, Universidade de São Paulo
 
+#include "PWGHF/Core/DecayChannels.h"
 #include "PWGJE/Core/JetDerivedDataUtilities.h"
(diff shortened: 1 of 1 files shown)


PWGJE/Tasks/jetDsSpecSubs.cxx
```diff
@@ -13,6 +13,7 @@
 /// \brief Ds-tagged jet analysis with substructure histogram outputs
 /// \author Monalisa Melo <monalisa.melo@cern.ch>, Universidade de São Paulo
 
+#include "PWGHF/Core/DecayChannels.h"
```
nzardosh: I think this header is not needed. Could you remove it please?

monamelop: Removed, thanks!

nzardosh (approved):

PR #17653, opened 2026-08-28
[PWGUD] Remove ZDC fields from MC output

amatyja:
Remove ZDC energy and time fields from MC output. These ZDC information does not exist in MC
Shifted information about reconstruction in a code


alibuild (approved):

PR #17659, opened 2026-08-28
[ALICE3] Propagate more information from otf decayer to otf tracker

jesgum:
Fasttracker needs the decay radius when smearing short lived particles since they might not traverse all layers


alibuild (approved):

PR #17662, opened 2026-08-29
[PWGLF] Fix typos in strangenesstofpid

romainschotter:
- fix typo in strangenesstofpid which, after calculating the Nsigma TOF, the values were set to their default values


alibuild (approved):

PR #17663, opened 2026-08-30
[PWGDQ] correct MFT propagation to vertex in QA tasks

aferrero2707:
The propagation of the MFT and global forward tracks to the vertex is now performed with the official `propagateTrackParCovFwd()` function from the forward track utilities.


aferrero2707:
@mcoquet642 this is the fix for the FWD track extrapolation to the vertex that you suggested in https://github.com/AliceO2Group/O2Physics/pull/17461.

Could you please have a look?
The `code-check` errors are actually false positives, for the warnings I will prepare a separate PR.
Thanks!

mcoquet642 (approved):

PR #17666, opened 2026-08-31
[PWGJE] Add Berkeley tree producer

mhwang285:
Add a task to produce BerkeleyTrees for MC events for further analysis in the Berkeley HPC systems. In our testing the resulting compression factor is around 80, so we'd like to merge this in and test on Hyperloop and see if we can produce standard derived data with this task. (O2 linter are unrelated to this task)


PWGJE/TableProducer/berkeleyTreeProducer.cxx
```diff
@@ -0,0 +1,204 @@
...
+                  berkeleytree::DetPhi,
+                  berkeleytree::DetTrackSel,
+                  berkeleytree::DetMcId,
+                  berkeleytree::GenPt,
+                  berkeleytree::GenEta,
+                  berkeleytree::GenPhi,
+                  berkeleytree::GenE,
+                  berkeleytree::GenCharge,
+                  berkeleytree::GenMcId,
+                  berkeleytree::PdgId);
```
nzardosh, line 80: probably this is fine but not sure if the Id keyword causes any issues later for the merger if it thinks its an index coloumn

nzardosh (approved):

PWGJE/TableProducer/berkeleyTreeProducer.cxx
mhwang285: Hmm, I see, do you think it triggers on the letters "Id" being anywhere in the column name? I can't really think of another name to use that wouldn't contain "Id" somewhere in it...

mhwang285:
thanks @nzardosh , the required checks say (for me at least) "Security: approval needed, not starting", do you know how we can get these to start? (might be because this is my first O2Physics PR)

nzardosh:
They should start running soon :) Btw can you update your username here to your CERN one please?

PR #17669, opened 2026-08-31
[PWGHF] Add D meson invariant mass selection to B0 reduced selector and fix process function handling in B0 task

fchinu:


fchinu:
MegaLinter fails because `isDmesInMassRange` always returns `true` for DstarPi, but the function is never called for DstarPi:
https://github.com/fchinu/O2Physics/blob/03db920491a1a9b5ae16cdc3e617256a1ed7d398/PWGHF/D2H/TableProducer/candidateSelectorB0ToDPiReduced.cxx#L125-L127

stefanopolitano (approved):

PR #17675, opened 2026-08-31
[PWGUD] Remove true tau info from MC tree

amatyja:
Remove true tau info from MC trees: gen and reconstructed
minor corrections in histogram names and labels


alibuild (approved):

PR #17679, opened 2026-08-31
[PWGJE] jetSpectraCharged: add rhoShift, jetBkgAnalysis: mcp option

aimeric-landou:
also fix shadowing centrality variable, fix default processes all false


aimeric-landou:
Linter magic number issue:
```
    const bool hasLead = jets.size() >= 1;
    const bool hasSub = jets.size() >= 2;
```
what would be a better way to write this?

nzardosh (approved):

PR #17685, opened 2026-08-31
[Common] Fill MC hists in a loop over MC collisions in CentralityQa

romainschotter:
- Add option to loop over MC collisions to fill MC histograms, while before all MC histograms were filled in a loop over reconstructed collisions


vkucera:
@romainschotter Please stop using the central repository for pushing your personal branches.

dsekihat (approved):

PR #17689, opened 2026-08-31
[PWGDQ] Allowing independent shifts of global tracks for MFT top or bottom halves

mcoquet642:
This introduces the possibility of applying shifts to global muons in the bottom or the top half of the MFT. This is activating by passing as a forward-shift CCDB object a vector of 10 entries (corresponding to the following shifts : x_top, y_top, z_top, slopeX_top, slopeY_top, x_bottom y_bottom, z_bottom, slopeX_bottom, slopeY_bottom). The purpose is to correct for misalignment between the two halves of the MFT at analysis level.


alibuild (approved):

PR #17698, opened 2026-09-01
[PWGEM] Add configurable for min. number of photons in EOI

jokonig:
- This is to reduce the datasize of derived data. Use min. number of photons >= 2 for pi0 etc. analysis


alibuild (approved):

PR #17709, opened 2026-09-01
[PWGHF] Use `const&` and `std::move` to avoid copies

vkucera:


alibuild (approved):

PR #17710, opened 2026-09-01
[PWGJE] Use `const&` and `std::move` to avoid copies

vkucera:


vkucera:
Ping

nzardosh (approved):

nzardosh:
there is an error in the o2 build

PR #17713, opened 2026-09-01
[PWGUD] Use `const&` and `std::move` to avoid copies

vkucera:


amatyja (approved):
Thanks a lot for corrections. UD is fine with them.

amatyja (approved):
Thnak you for corrections. UD part is fine.

PR #17720, opened 2026-09-01
[PWGLF] Remove cascade wrong hypothesis assignments in strangenesstofpid

romainschotter:
Removed assignments of -999. for wrong hypotheses in nSigma calculations
--> already assigned by default in the cascTofInfo struct


alibuild (approved):

PR #17721, opened 2026-09-01
[PWGLF] Add special treatment for decays into two positive PDG code particles

romainschotter:
-  Add special treatment for decays into two positive PDG code particles 
- Add new strategy to look at generated secondary V0s


alibuild (approved):

PR #17727, opened 2026-09-02
[PWGCF] Flowcontainer: memory fixes

EmilGorm:
* Remove redundant copies
* Fix memory leaks


alibuild (approved):
Auto-approving on behalf of @EmilGorm.

PR #17731, opened 2026-09-02
[PWGEM] Unify QC + MaterialBudgetTasks

Steffimro:
This task restructures the PCM QC code: the standalone pcmQCMC task and the material-budget part of MaterialBudget.cxx are merged into pcm-qc as process functions, the old tasks are moved to the legacy folder.

Process functions (all off by default except processQC):

processQC / processQCML: data QC as before, with classical or ML-based selection. Added default histograms: Rxy/eta/phi vs pT, an (alpha, qT, pT) sparse for pT-sliced Armenteros plots, psi_pair vs pT (ML path), TPC nsigma_el vs eta.

processPCMQCMC / processPCMQCMCML: former pcmQCMC. QC observables split by MC truth class (primary, fromWD, fromHS, fromPi0Dalitz, fromEtaDalitz, candidate as purity denominator) including pT/eta/phi and conversion-point resolution. MC histograms are only booked when an MC process is enabled.

processGen: generator-level reference (needs binned generated-pT derived data), plus wide-R conversion-point maps

processRecoQA: new, runs on AO2Ds before skimming. Single-photon efficiency analysis possible (converted -> both legs tracked -> both legs in V0Legs -> candidate in V0PhotonsKF) as {pT, eta, phi, Rconv} sparses, duplicate counters and missing-leg diagnostics, to separate tracking losses from SVertexer losses.

**New optional groups:** materialBudgetSettings_group (conversion-point maps, overall and per detector region), mcAnalysisModeSettings_group (detailed resolution, collision-association QA, true-association requirement), genSettings_group, recoQASettings_group, qaSettings_group (cut variables vs pT before/after the selection). All new output is off by default.


mhemmer-cern:
It looks like the old tasks are not in the Legacy folder with this commit.

jokonig (approved):

PR #17734, opened 2026-09-02
[PWGEM] Add table for detailed event normalization purposes

jokonig:
- Table is filled only once per DF to reduce output size. Essentially an event-normalization histogram that can then be loaded in the analysis task runnnig over derived data
- Allows to keep track of reject reasons for the events
- Should be extended for multiplicity dependent analyses


dsekihat:
Dear @jokonig , there is a table for normalization information. https://github.com/AliceO2Group/O2Physics/blob/86ccf3cb386368ed3004e8972ee817b8e0edcf5e/PWGEM/Dilepton/DataModel/dileptonTables.h#L450

PR #17738, opened 2026-09-02
[PWGLF] Add new propagation method in forwardlambdakzeroanalysis

romainschotter:
- add new propagation method, closer to what is currently used for propagating MFT tracks to PV
- fix axes used in DCA QA histograms

Thank you very much @MauriceCoke !


alibuild (approved):

PR #17749, opened 2026-09-03
[PWGLF] Update TOF calculation for wrong PID hypothesis in strangenesstofpid

romainschotter:
- remove unnecessary event time map
- update TOF calculation for wrong PID hypothesis
- adapt data model for storing all wrong PID hypothesis
- use centralized TOF calibration in derived data


alibuild (approved):

PR #17756, opened 2026-09-03
[ALICE3] Add short-lived particle table to otf tof pid

jesgum:
@njacazio please have a look :-)

### DIFF
--- a/ALICE3/DataModel/OTFTOF.h
+++ b/ALICE3/DataModel/OTFTOF.h
@@ -150,6 +150,26 @@ DECLARE_SOA_DYNAMIC_COLUMN(NSigmaOuterTOF, nSigmaOuterTOF, //! General function 
                              }
                            });
 
+DECLARE_SOA_COLUMN(NSigmaSplusInnerTOF, nSigmaSplusInnerTOF, float);   //! NSigma Sigma plus InnerTOF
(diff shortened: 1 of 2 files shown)


njacazio (dismissed):
Very good!

ALICE3/DataModel/OTFTOF.h
```diff
@@ -222,9 +242,28 @@ DECLARE_SOA_TABLE(UpgradeTofExpectedTimes, "AOD", "UPGRADETOFEXPT",
                   upgrade_tof::OuterTOFExpectedTimeHe3,
                   upgrade_tof::OuterTOFExpectedTimeAl);
 
+DECLARE_SOA_TABLE(UpgradeTofExtras, "AOD", "UPGRADETOFEXTRAS",
```
njacazio: I'm not a fan of the naming actually, could we write that it has to do with long lived particles?

jesgum: I renamed the table, in principle I think we could probably instead also split it to nsigma/times like the other ones but considering that the table is rather small right now we could keep it like this. I didn't start preparing the task that is supposed to use the table so if we want to change it later it shouldn't really be an issue

njacazio (approved):

njacazio:
Warnings and friends to be fixed in a coming pr not to pollute this one

PR #17772, opened 2026-09-04
[PWGJE] add hard scattering scale to gammejettreeproducer

fjonasALICE:
- addition is required to allow offline rejection of outliers in JJ and GJ MC


mhemmer-cern (approved):
LGTM

PR #17783, opened 2026-09-05
[PWGCF] Add efficiency-corrected multiplicity information to CF derived data

EmilGorm:
* CorrelationsDerived.h: Efficiency-corrected multiplicity now stored in dedicated CFCollisionsExtra table as sum of per track weight
* Keep track of track-based multiplicity in transient CFMultiplicities for efficiency correction purposes
* filterCorrelations.cxx: fill efficiency-corrected multiplicity when an efficiency file is supplied to the table producer
   * Only filled when the track-based multiplicity is used
* twoParticleCorrelationsMpi.cxx: utilise the efficiency-corrected multiplicity
    * dedicated process functions
    * compile-time lookup to switch on multiplicity
    * Added option for MC efficiency calculations to only use one reconstructed event per generated collision

### DIFF
--- a/PWGCF/TableProducer/filterCorrelations.cxx
+++ b/PWGCF/TableProducer/filterCorrelations.cxx
@@ -21,6 +21,7 @@
 #include "Common/DataModel/PIDResponseTPC.h"
 #include "Common/DataModel/TrackSelectionTables.h"
 
+#include <CCDB/BasicCCDBManager.h>
 #include <Framework/AnalysisDataModel.h>
 #include <Framework/AnalysisHelpers.h>
 #include <Framework/AnalysisTask.h>
@@ -35,14 +36,21 @@
 #include <MathUtils/detail/TypeTruncation.h>
 #include <ReconstructionDataFormats/PID.h>
 
+#include <TFile.h>
 #include <TH3.h>
+#include <THn.h>
 #include <TParticlePDG.h>
 
 #include <Rtypes.h>
 
 #include <algorithm>
+#include <array>
+#include <cmath>
 #include <cstdint>
 #include <experimental/type_traits> // required for is_detected
+#include <limits>
+#include <memory>
+#include <string>
 #include <type_traits>
 #include <vector>
 
@@ -58,6 +66,7 @@ using namespace o2::math_utils::detail;
 
 struct FilterCF {
   Service<o2::framework::O2DatabasePDG> pdg;
+  Service<o2::ccdb::BasicCCDBManager> ccdb;
 
   enum TrackSelectionCuts1 : uint8_t {
     kTrackSelected = BIT(0),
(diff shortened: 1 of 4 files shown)


wenyaCern (dismissed):

PWGCF/TableProducer/filterCorrelations.cxx
```diff
@@ -280,6 +341,69 @@ struct FilterCF {
...
+    }
+    return mEfficiency;
+  }
+
+  template <typename TCollision, typename TTracks>
+  float getCorrectedMultiplicity(const TCollision& collision, const TTracks& tracks, uint64_t timestamp)
+  {
+    if (!collision.isTrackMultiplicity()) {
+      LOGF(fatal, "Efficiency-corrected multiplicity requires MultiplicitySelector::processTracks");
+    }
```
jgrosseo, line 363: As this is anyway a transient table, I was wondering if instead of adding a new bool, one could use this feature to check the configuration of MultiplicitySelector:
https://github.com/AliceO2Group/O2Physics/blob/master/Common/Core/TableHelper.h#L71

PWGCF/TwoParticleCorrelations/Tasks/twoParticleCorrelationsMpi.cxx
```diff
@@ -284,14 +286,32 @@ struct TwoParticleCorrelationsMpi {
   using AodTracks = soa::Filtered<soa::Join<aod::Tracks, aod::TrackSelection>>;
 
   using DerivedCollisions = soa::Filtered<aod::CFCollisions>;
+  using DerivedCollisionsCorrected = soa::Filtered<aod::CFCollisionsWithExtra>;
+  using DerivedCollisionsMultSet = soa::Filtered<soa::Join<aod::CFCollisions, aod::CFMultSets>>;
+  using DerivedCollisionsMultSetCorrected = soa::Filtered<soa::Join<aod::CFCollisions, aod::CFCollisionsExtra, aod::CFMultSets>>;
   using DerivedTracks = soa::Filtered<aod::CFTracks>;
+  static constexpr int RequireRecoCollisionBestMode = 2;
```
jgrosseo, line 293: Why not an enum for all 3 modes then?

jgrosseo (commented):
See my two comments.

What I find not good is that we have now a processMCEfficiency in correlations.cxx and in this task. I think we should only duplicate things which are strictly MPI related, while the efficiency improvements could also be useful for other analysis and I don't see any issue for others with your changes. What do you think?

EmilGorm:
> See my two comments.
> 
> What I find not good is that we have now a processMCEfficiency in correlations.cxx and in this task. I think we should only duplicate things which are strictly MPI related, while the efficiency improvements could also be useful for other analysis and I don't see any issue for others with your changes. What do you think?

Agreed, I have removed the processMCEfficiency from twoParticleCorrelationsMpi.cxx and moved the updates to correlations.cxx so it can remain the default task for processing efficiencies.

PWGCF/TableProducer/filterCorrelations.cxx
EmilGorm: This is a very nice solution, I have put it in

PWGCF/TwoParticleCorrelations/Tasks/twoParticleCorrelationsMpi.cxx
EmilGorm: Done.

jgrosseo (approved):

wenyaCern (approved):

PR #17803, opened 2026-09-07
[PWGDQ] Add muon low pt1p5 mch midmatching cut

iamkoushikb-bit:
https://indico.cern.ch/event/1675395/#4-upsilon-signal-extraction-in - Here it has been shown that the region of single muon pT cut between 1.5 - 1.7 GeV/c is optimal for background reduction as well as retaining most of the signal statistics. Kindly refer to my talk (slide 7).


PWGDQ/Core/CutsLibrary.cxx
```diff
@@ -3062,662 +2995,681 @@ AnalysisCompositeCut* o2::aod::dqcuts::GetCompositeCut(const char* cutName)
...
+    return cut;
+  }
+
+  if (!nameStr.compare("muonLowPt1p7_MchMidmatching")) {
+    cut->AddCut(GetAnalysisCut("muonLowPt1p7"));
+    cut->AddCut(GetAnalysisCut("matchedMchMid"));
+    return cut;
+  }
+
+  if (!nameStr.compare("muonLowPt6_MchMidmatching")) {
```
mguilbau: What is this used for?

iamkoushikb-bit: Dear Maxime, 

As I pointed out in my talk, 1.5 - 2.0 GeV/c single muon pT cut is the optimal range. And I was also suggested not to use any pDCA cut so this cut which couples Single muon pT cut with MCHMID matching has been added here in the PR. That is the reason of adding the 2.0 GeV/c cut and what is your opinion about this?

cheers,
Koushik

mcoquet642 (approved):

PR #17743, opened 2026-09-03
[PWGDQ] various improvements to the global alignment task histograms

aferrero2707:
* added histograms for di-muon decay lengths
* added possibility to modify some of the histogram axis definitions via configurable options
* added option that reduces the size of MFT DCA histograms (enabled by default)

PR #17809, opened 2026-09-07
[PWGDQ] added derived tables to global muon alignment task

aferrero2707:
The derived tables provide the minimal information neederd to run the alignment analysis on new geometries, thus avoiding to pocess each time the original large-size AO2Ds.

A pre-processor switch allows to use the same task for either producing or processing the derived tables.

### DIFF
--- a/PWGDQ/Tasks/muonGlobalAlignment.cxx
+++ b/PWGDQ/Tasks/muonGlobalAlignment.cxx
@@ -119,18 +119,126 @@ using SMatrix5 = ROOT::Math::SVector<Double_t, 5>;
 using o2::dataformats::GlobalFwdTrack;
 using o2::track::TrackParCovFwd;
 
+namespace muonaligncoll
+{
+DECLARE_SOA_COLUMN(RunNumber, runNumber, int);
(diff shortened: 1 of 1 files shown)


aferrero2707:
@mcoquet642 @javierecc @ppoava here is another relatively urgent PR, that should allow to iterate over the alignment geometries much faster with small derived tables.

I still need to rebase once https://github.com/AliceO2Group/O2Physics/pull/17743 is merged, but you can already have a look at the changes if you have the time.

Thanks!

mguilbau:
@mcoquet642 I looked at the changes. This look fine to me but I believe it is better if you take a quick look before merging

aferrero2707:
@mcoquet642 @mguilbau note that the changes only affect the task we use for the in-depth analysis of the global muon alignment, and do not touch other parts of the DQ code.

Unless there are serious errors reported by the `code-check`, it should be good to go.

mguilbau:
No serious error reported indeed. I will approve it then

mguilbau (approved):

mcoquet642:
@mguilbau In future please wait for the code-check to finish before merging

PR #17192, opened 2026-07-24
[Tools/ML] Fix code-check issues in MLResponse

fcatalan92:

PR #17810, opened 2026-09-07
[PWGHF] Remove unnecessary caching of feature names in HF track index skim creator

fgrosa:
This PR removes the unnecessary caching of feature names, which is incompatible with https://github.com/AliceO2Group/O2Physics/pull/17192/changes


alibuild (approved):

vkucera:
@fgrosa The linked PR did not change any functionality. How can it be the reason for this PR?

fgrosa:
> @fgrosa The linked PR did not change any functionality. How can it be the reason for this PR?

If you try to execute the track index skim creator now with BDTs applied, you will get the following error:
```
[ERROR] SEVERE: Device hf-track-index-skim-creator (21008) had at least one message above severity 7: Input feature ptProng0 not available! Please check your configurables.
```
See e.g. https://alimonitor.cern.ch/train-workdir/tests/0075/00755074/stdout.log. I assumed that the issue was in the change of the check in `cacheInputFeaturesIndices`, since it was not happening before (and it makes sense because `mAvailableInputFeatures.contains(inputFeature)` cannot be true, as we do not pass the features via configurables given that they are fixed).

fcatalan92:
I think the behavior of `mAvailableInputFeatures.contains(inputFeature)` is very similar to that previously implemented with `mAvailableInputFeatures.count(inputFeature)`. Maybe the path now crashing in the index-skim-creator was never run before or using `count` the check was not effective.
In any case, the current behavior of `MlResponse` is what we want, so all looks sane to me.

fcatalan92 (approved):

fcatalan92:
BTW, after some investigation, the changes in this PR remove the workaround implemented in #16719 for the issue introduced by #16424 and fixed in #16722. Again, all makes sense.

PR #17820, opened 2026-09-07
[PWGHF] Add optional event-status filling to taskD0

zhangbiao-phy:
Add an optional `fillEventStatus` mode to `taskD0` to support centrality-dependent signal-loss studies.

When enabled, `hSparseAcc` uses calibrated MC centrality in its existing centrality axis and appends an event-status axis:
- 0: no associated reconstructed collision;
- 1: associated reconstructed collisions exist, but none pass the HF event selection;
- 2: at least one associated reconstructed collision passes the HF event selection.

Generated particles are filled once, including those without a reconstructed collision, provided they pass the upstream generated-particle selection.


alibuild (approved):

PR #17828, opened 2026-09-08
[PWGHF] Fix a typo in the selector

JaeYoonCHO:
This PR fixes a typo in `candidateSelectorXicToXiPiPi.cxx`, where the `isSelectedTrackItsQuality` check was originally applied only to the first pion. After this fix, `isSelectedTrackItsQuality` is applied to both pions.


fgrosa:
> This PR fixes a typo in `candidateSelectorXicToXiPiPi.cxx`, where the `isSelectedTrackItsQuality` check was originally applied only to the first pion. After this fix, `isSelectedTrackItsQuality` is applied to both pions.

Thanks for spotting this! It should not have any effect anyway, as globalTracksWoDCA are selected in the creation of the HF derived data, but good to have it correct in the code.

fgrosa (approved):

PR #17829, opened 2026-09-08
[PWGCF] Update event shape processing in track-track task

ariedel-cern:
- Fix code check error
- Add additional dimensions to the sparse histogram


alibuild (approved):

PR #17831, opened 2026-09-08
[Common] Avoid zero-initialising the NN TPC PID prediction buffer

ktf:
The per-DataFrame NN prediction buffer (masked tracks × output dims × 9 mass hypotheses, ~72 MB for a 1M-track PbPb DataFrame) is a value-initialised `std::vector<float>`: a full memset touching every page of the allocation, immediately overwritten in full by the per-hypothesis evaluation loop, so every page is paid for twice.

Carry it as an uninitialised `unique_ptr<float[]>` instead, so each page is touched exactly once, by the write that fills it. `makePidTables` takes `const float*` accordingly.

### DIFF
--- a/Common/Tools/PID/pidTPCModule.h
+++ b/Common/Tools/PID/pidTPCModule.h
@@ -443,11 +443,8 @@ class pidTPCModule
 
   //__________________________________________________
   template <typename TCCDB, typename M, typename T, typename B>
-  std::vector<float> createNetworkPrediction(TCCDB& ccdb, soa::Join<aod::Collisions, aod::EvSels> const& collisions, M const& mults, T const& tracks, B const& bcs, const size_t size)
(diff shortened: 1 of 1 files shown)


ktf:
@ChSonnabend can you have a look?
@vkucera can you please turn of the code-checker until it stops making PRs red for issues unrelated to the actual PRs?

ChSonnabend:
@ktf Sounds reasonable to me. Did you benchmark the performance and see if this makes any difference?

ktf:
Do you have a benchmark for it?

ChSonnabend:
> Do you have a benchmark for it?

The benchmark would be the PID task. No I don't as I don't actively maintain it anymore. I would need to ask the PID group. Thats mainly why I asked if you already did that, otherwise we should probably forward it to @amaringarcia

ktf:
To be clear, I run the PID task for the whole "do preparations in a standalone ONNXRuntime node" where I do see some improvements, which we would need to discuss / validate though. This is PR is a split off that development because it's self contained, removes an unneeded zeroing, and simplifies the final PR.

dsekihat (approved):

PR #17832, opened 2026-09-08
[PWGEM/Dilepton] remove unnecessary histogram from single muon QC MC

dsekihat:
[PWGEM/Dilepton] remove unnecessary histogram from single muon QC MC


alibuild (approved):

PR #17835, opened 2026-09-08
[PWGLF] SigmaPlus MCgen info & PCM reco fix

hfribert:
- The PCM reconstruction for the SigmaPlus does not use the V0 tables anymore, but searches and fits the electron/positron itself (many are missing in the V0 tables)
- Added SigmaPlus MC truth pT to the LFKinkDecayTables
- Added in both the sigmaplusbuilder and the sigmaHadCorr task the ability to save the trueMC-generated SigmaPlus info for efficiency studies

- Fixed linter issues of the sigmaHadCorr task

### DIFF
--- a/PWGLF/DataModel/LFKinkDecayTables.h
+++ b/PWGLF/DataModel/LFKinkDecayTables.h
@@ -25,6 +25,7 @@
 
 #include <array>
 #include <cmath>
+#include <cstdint>
 
 namespace o2::aod
 {
--- a/PWGLF/DataModel/LFKinkDecayTables.h
+++ b/PWGLF/DataModel/LFKinkDecayTables.h
@@ -216,6 +216,7 @@ DECLARE_SOA_COLUMN(GammaGMotherPdgCode, gammaGMotherPdgCode, int); //! PDG code 
 DECLARE_SOA_COLUMN(XDecVtxMC, xDecVtxMC, float);         //! MC-truth Sigma+ decay vertex (x direction)
 DECLARE_SOA_COLUMN(YDecVtxMC, yDecVtxMC, float);         //! MC-truth Sigma+ decay vertex (y direction)
 DECLARE_SOA_COLUMN(ZDecVtxMC, zDecVtxMC, float);         //! MC-truth Sigma+ decay vertex (z direction)
+DECLARE_SOA_COLUMN(DecayRadiusMC, decayRadiusMC, float); //! MC-truth Sigma+ decay radius
 DECLARE_SOA_COLUMN(PxSigmaPlusMC, pxSigmaPlusMC, float); //! MC-truth Sigma+ mother Px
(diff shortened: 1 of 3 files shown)


romainschotter (changes requested):
Hi @hfribert !
The PR looks fine!
However, I understand that you need to use the DCA fitter to reconstruct the kink topology of the Sigma+ --> proton gamma (why not proton pi0?) but I am not entirely sure to understand why it is necessary to rebuild the photon from scratch and not take them from the V0 table produced in the propagationService. 
You wrote that a lot of photons were lost in the propagationService. Could you please elaborate a bit more?
It would be useful to understand how and why they are lost in order to fix it centrally. 
Thank you very much!

hfribert:
Hi @romainschotter !

Actually we reconstruct the decay to proton + pi0, but due to the short lifetime of the pi0 we assume that the proton and 2 gammas come from one vertex. To boost efficiency we only reconstruct one of the photons via PCM and apply some kinematic constraints such as the pion mass and decay planes etc. to reconstruct the full decay in the end.

Regarding the V0 tables, they have some preselections applying CPA and DCA cuts which seem to be too tight for the dielectrons we try to reconstruct; the 'findable' histograms that we implemented with MC truth showed at least that many dielectrons are not saved in the tables due to that. (If their quality is good enough for reliable dca fits is to be seen)

Screenshot 2026-09-08 at 23 09 42

romainschotter:
Hi @hfribert ! 
Thank you very much for the prompt reply!
Regarding the V0 table: I am confused because the propagationService uses very loose selections (cosPA > 0.95, DCA pos/neg to PV > 0.05 cm, which are the loosest selections possible). The only selection that could be "tight" is DCA V0 daughters < 1 cm and we could consider to loosen it, either centrally or in your own analysis ticket in Hyperloop by making a clone of the propagationService. 
The other reason for sticking to the propagationService is that, because it has been used and is currently used in photon analyses, it includes useful features to address issues like the fact that TPC-only tracks (and thus PCM photons) are badly constrained in time and are compatible with several collisions. The propagationService includes, for instance, different strategies to address this "deduplication" problem.

hfribert:
Hi @romainschotter, thanks a lot for the insights on the propagationService!

I will follow the recommendation to use the propagationService instead, after checking out the available deduplication options, etc., and considering the possibility of just creating my own propagation-service wagon with looser selections if needed. Thanks again for the quick reply!

I am making some more checks on which cuts specifically have to be loosened, and will update the PR

romainschotter:
Hi @hfribert !
The choice of the deduplication algorithm can be set in the propagationService via the configurable `deduplicationAlgorithm`, which is an integer. There are different modes and, by default, the mode 1 is used for PCM photons as it is the one showing the best performances. You can find more information about the different deduplication modes in this presentation, for example: https://indico.cern.ch/event/1547630/#2-update-on-sigma0-baryon-prod
If needed, we can also discuss either privately or at the PAG-Strangeness meeting.
Thank you very much!

romainschotter (approved):

PR #17836, opened 2026-09-08
[Infrastructure,Tools,Common,Tutorial,PWGDQ,PWGHF,PWGLF] [WIP] [ML] Fix for memory issue in ML headers

ChSonnabend:
This PR fixes the dangling float* pointers and replaces it with a std::vector allocation.


ChSonnabend:
This still needs to be tested but should fix a memory issue spotted by @mhemmer-cern and discussed with @fmazzasc 
Please do not merge yet.

romainschotter (dismissed):

PR #17846, opened 2026-09-09
[Common] Fix TPC side determination in VDrift correction

matthias-kleiner:


ddobrigk (dismissed):

ddobrigk:
Hi @matthias-kleiner, thanks a lot for this fix! 

However, I was wondering what will happen with data for which the TPC side flags are not stored: this is relevant, since those were added only in February this year. Further, since they were added "adiabatically" by taking unused bits in the flags integer, older datasets will appear to have all tracks only in one side, and thus the fix will break in that situation. Is there perhaps some way in which the TPC side information could be inferred from other parameters that existed already before? For instance, a combination of tgl and the sign of Z (given we store tracks at the position of their innermost update?)... Sorry if I am overlooking something. Thanks again!

matthias-kleiner:
> Hi @matthias-kleiner, thanks a lot for this fix!
> 
> However, I was wondering what will happen with data for which the TPC side flags are not stored: this is relevant, since those were added only in February this year. Further, since they were added "adiabatically" by taking unused bits in the flags integer, older datasets will appear to have all tracks only in one side, and thus the fix will break in that situation. Is there perhaps some way in which the TPC side information could be inferred from other parameters that existed already before? For instance, a combination of tgl and the sign of Z (given we store tracks at the position of their innermost update?)... Sorry if I am overlooking something. Thanks again!

Hi @ddobrigk ,
I added a fallback solution for the older data as @shahor02  suggested. @shahor02 could you have  look if this is correct?

shahor02 (dismissed):

ddobrigk:
Hi @matthias-kleiner, thanks a lot! I have one question and a request, if I may: 

* Do you know what effect this change has? I would presume this was making us miss valid candidates, so it might recover some efficiency for photon conversions? 
* Could you please add a switch between old and new behaviour, with the default being old behaviour? This will allow for testing without silently changing the result of analysis code that hasn't been touched; then, at a later stage, we can switch the default to new behaviour and send out a notification. 

Thanks again - and please only worry about this whenever convenient and don't let me disturb your weekend!

matthias-kleiner:
> Hi @matthias-kleiner, thanks a lot! I have one question and a request, if I may:
> 
>     * Do you know what effect this change has? I would presume this was making us miss valid candidates, so it might recover some efficiency for photon conversions?
> 
>     * Could you please add a switch between old and new behaviour, with the default being old behaviour? This will allow for testing without silently changing the result of analysis code that hasn't been touched; then, at a later stage, we can switch the default to new behaviour and send out a notification.
> 
> 
> Thanks again - and please only worry about this whenever convenient and don't let me disturb your weekend!

Hi @ddobrigk ,
yes, I added a switch that is off by default and the new mode can be activated with `--configKeyValues "TPCVDriftManager.useSideBasedCorrection=true"`

I think it could in principle recover some efficiency loss: when the sign of the tgl disagrees with the actual TPC side, the old code shifts the tracks Z the wrong way instead of correcting it, which would hurt topological cuts. How much depends on how large the vDrift correction is and how often the vertex position/track topology puts tgl and Z sign at different sign. The best would be to make a direct A/B comparison with the switch.

ddobrigk (dismissed):

romainschotter:
Hi @matthias-kleiner ! 

Thank you very much for the PR! 

I have taken the liberty to try out your changes in order to test whether the configurable is available in the `propagationService` configuration and it turns out that it won't appear in the configuration. As far as I understand, it is because the configurable has been introduced here via `O2ParamDef`, which declares only a static member, that needs to be defined in a source file via `O2ParamImpl` (which does not exist in the PR). If the file would exist, the other issue is that the `TPCDriftManager` is declared as `o2physics_add_header_only_library`, which prevents to link the source file with `O2ParamImpl` and prevents the task calling the TPCDriftManager to compile (hence the compilation error).

The simplest alternative is maybe to add the variable `mUseSideBasedCorrection` as a private data member of the class, along side a setter function in order to set the value to `true` or `false`. In this way, we can later add a configurable in any task using the `TPCVDriftManager` (`propagationService`, `photonconversionbuilder`,...) in order to enable/disable the new side-based correction. If it helps, I have added comments with the changes to be made.

My 2 cents.

Common/Core/TPCVDriftManager.h
```diff
@@ -14,6 +14,8 @@
 
 #include <CCDB/BasicCCDBManager.h>
 #include <CommonConstants/LHCConstants.h>
+#include <CommonUtils/ConfigurableParam.h>
+#include <CommonUtils/ConfigurableParamHelper.h>
```
romainschotter:
```suggestion

```

Common/Core/TPCVDriftManager.h
```diff
@@ -26,6 +28,15 @@
...
 
+struct TPCVDriftManagerParam : public o2::conf::ConfigurableParamHelper<TPCVDriftManagerParam> {
+  // Use the TPC side flags (with legacy-data fallback) instead of the tgl-sign-based correction.
+  // Off by default so that existing analyses see no change in results until this is explicitly
+  // enabled for testing.
+  bool useSideBasedCorrection = false;
+
+  O2ParamDef(TPCVDriftManagerParam, "TPCVDriftManager");
+};
+
```
romainschotter:
```suggestion


```

Common/Core/TPCVDriftManager.h
romainschotter, line 38:
```suggestion
class TPCVDriftManager
{
 public:
  void init(o2::ccdb::BasicCCDBManager* ccdb) noexcept
  {
    mCCDB = ccdb;
  }

  void setUseSideBasedCorrection(bool value) noexcept
  {
    mUseSideBasedCorrection = value;
  }
```

Common/Core/TPCVDriftManager.h
```diff
@@ -131,7 +142,54 @@ class TPCVDriftManager
     }
 
     // impose new Z coordinate
-    track.setZ(track.getZ() + ((track.getTgl() < 0.) ? -dDrift : dDrift));
+    float zShift = 0.f;
+    if (!TPCVDriftManagerParam::Instance().useSideBasedCorrection) {
```
romainschotter:
```suggestion
    if (!mUseSideBasedCorrection) {
```

Common/Core/TPCVDriftManager.h
```diff
@@ -156,6 +214,8 @@ class TPCVDriftManager
   o2::ccdb::BasicCCDBManager* mCCDB{};  // reference to initialized ccdb manager
```
romainschotter, line 213:
```suggestion
bool mUseSideBasedCorrection{false};
```

matthias-kleiner:
> Hi @matthias-kleiner !
> 
> Thank you very much for the PR!
> 
> I have taken the liberty to try out your changes in order to test whether the configurable is available in the `propagationService` configuration and it turns out that it won't appear in the configuration. As far as I understand, it is because the configurable has been introduced here via `O2ParamDef`, which declares only a static member, that needs to be defined in a source file via `O2ParamImpl` (which does not exist in the PR). If the file would exist, the other issue is that the `TPCDriftManager` is declared as `o2physics_add_header_only_library`, which prevents to link the source file with `O2ParamImpl` and prevents the task calling the TPCDriftManager to compile (hence the compilation error).
> 
> The simplest alternative is maybe to add the variable `mUseSideBasedCorrection` as a private data member of the class, along side a setter function in order to set the value to `true` or `false`. In this way, we can later add a configurable in any task using the `TPCVDriftManager` (`propagationService`, `photonconversionbuilder`,...) in order to enable/disable the new side-based correction. If it helps, I have added comments with the changes to be made.
> 
> My 2 cents.

Hi @romainschotter ,
many thanks for checking it, you are indeed correct. I have changed it as you suggested to a simple setter that is `false` per default.

ddobrigk:
Many, many thanks @matthias-kleiner ! (and @romainschotter for taking a look too). Will approve

ddobrigk (dismissed):

ddobrigk (approved):

PR #17847, opened 2026-09-09
[PWGEM] PhotonMeson: Fix Preslice and indexing in Dalitz skimmer

jokonig:
- Preslices have to be defined outside process function
- Legs are now accessed from V0s and indexing from v0 table fixed


alibuild (dismissed):

mhemmer-cern (approved):

PR #17852, opened 2026-09-09
[PWGCF] Add MC processing to derived-to-derived femto producer

ariedel-cern:
Adding MC processing to derived-to-derived producer (for tracks only for now). Also update data model to have version and staged mc labels.


alibuild (approved):

PR #17865, opened 2026-09-10
[PWGLF] Add centrality and gen daughter momenta components to trackedHypertritonRecoTask table producer

creetz16:


maciacco (approved):

maciacco:
@creetz16 approving now as this should run asap in Hyperloop. Please fix linter errors in the next one (missing initializers, variable shadowing, unnecessary else if)

PR #17871, opened 2026-09-10
[PWGHF] fix a mistake from last commit for mc collision globalIndex in D0 task

zhangbiao-phy:


alibuild (approved):

vkucera:
@zhangbiao-phy Please always link the PR that introduced the bug.

PR #17881, opened 2026-09-11
[ALICE3] fix code check in otf particle

jesgum:


alibuild (approved):

PR #17887, opened 2026-09-11
[PWGLF] Split MC process + optimize histo filling

gbencedi:
- split MC process + remove unused functions and redundant histograms
- optimize event/track counter histogram filling
- remove pT cut for generated particles


alibuild (approved):

PR #17898, opened 2026-09-13
[PWGDQ] Removing FwdTracksDCA dependencies from tableMaker(MC)_withAssoc

mcoquet642:
In order to save resources, computation of standalone and global muon dca is done on the fly in tableMaker(MC). Removing dependency of tables produced by fwdtrackextension


alibuild (approved):

PR #17903, opened 2026-09-14
[PWGJE] Bugfixes in JE tasks

mhwang285:
- Kappa and alpha were not being used in some angularity calculations
- Resolved some O2 linter issues (most but not all)
- Fixed two mislabeled histogram axes in jetSpectraCharged.cxx


PWGJE/Tasks/jetFinderQA.cxx
```diff
@@ -471,16 +471,16 @@ struct JetFinderQATask {
   Filter eventCuts = (nabs(aod::jcollision::posZ) < vertexZCut &&
                       ((checkCentFT0M ? aod::jcollision::centFT0M : aod::jcollision::centFT0C) >= centralityMin) &&
                       ((checkCentFT0M ? aod::jcollision::centFT0M : aod::jcollision::centFT0C) < centralityMax));
-  PresliceUnsorted<soa::Filtered<aod::JetCollisionsMCD>> CollisionsPerMCPCollision = aod::jmccollisionlb::mcCollisionId;
```
nzardosh, line 474: For these could you put them back please? even though it goes against the convention in most of the framework I use capitals for preslices to differentiate them. We might decide to change this globally for later but for now would be nice to keep it consistent

PWGJE/Tasks/jetSpectraCharged.cxx
```diff
@@ -743,7 +743,7 @@ struct JetSpectraCharged {
           double dcorrpt = corrBasejetpt - corrTagjetpt;
           float angularityMcd = 0.;
           for (const auto& constituent : jetMCD.template tracks_as<aod::JetTracks>()) {
-            angularityMcd += std::pow(constituent.pt(), kappa) * std::pow(jetutilities::deltaR(jetMCD, constituent), alpha);
+            angularityMcd += std::pow(constituent.pt(), kappa) * std::pow(jetutilities::deltaR(jetMCD, constituent) / (jetMCD.r() / 100.f), alpha);
           }
           angularityMcd /= (corrBasejetpt * (jetMCD.r() / 100.f));
```
nzardosh: in line 746 there is a division by the radius. There shouldnt be another one here i think?

nzardosh (commented):
Dear Tucker, thanks for making the changes. I have left a couple of comments. 

For a future note it is best not to change the analysis tasks written by other analysers, even if the changes are cosmetic (unless discussed with them first, in which case please disregard this) as they might be making changes of their own in their local branches. @aimeric-landou @joonsukbae could you please check and make sure you are ok with the changes proposed.

PWGJE/Tasks/jetSpectraCharged.cxx
mhwang285: Ah thanks for catching that, I fixed 746 but not 748, both lines are fixed now. I also had to change `corrTagjetpt`, `corrBasejetpt`, and `dcorrpt` to floats so that we can use `std::pow` with a `kappa` (which is `float`)

PWGJE/Tasks/jetFinderQA.cxx
mhwang285: Yes, recapitalized now in [81405af](https://github.com/AliceO2Group/O2Physics/pull/17903/commits/81405afc4856c0920d542390ad9f2ed719c93a81)

PWGJE/Tasks/jetSpectraCharged.cxx
```diff
@@ -542,9 +542,9 @@ struct JetSpectraCharged {
     float angularity = 0.;
     for (const auto& constituent : jet.template tracks_as<aod::JetTracks>()) {
       registry.fill(HIST("h2_jet_pt_track_pt_rhoareasubtracted"), jetcorrpt, constituent.pt(), weight);
-      angularity += std::pow(constituent.pt(), kappa) * std::pow(jetutilities::deltaR(jet, constituent), alpha);
+      angularity += std::pow(constituent.pt(), kappa) * std::pow(jetutilities::deltaR(jet, constituent) / (jet.r() / 100.f), alpha);
     }
-    angularity /= (jet.pt() * (jet.r() / 100.f));
+    angularity /= std::pow(jet.pt(), kappa);
```
mhwang285: @nzardosh should this normalization be `jetcorrpt` instead of `jet.pt()`? In the original code it was `jet.pt()` so I kept it but now looking at L746 maybe it should be the corrected pT rather than the uncorrected?

joonsukbae:
> Dear Tucker, thanks for making the changes. I have left a couple of comments.
> 
> For a future note it is best not to change the analysis tasks written by other analysers, even if the changes are cosmetic (unless discussed with them first, in which case please disregard this) as they might be making changes of their own in their local branches. @aimeric-landou @joonsukbae could you please check and make sure you are ok with the changes proposed.

Dear @mhwang285 and @nzardosh, it looks good to me. The type/name changes and the angularity expansion also look good. Thank you also for catching the switched MCD vs. MCP axis labels.

nzardosh (approved):

PR #17906, opened 2026-09-14
[PWGHF] Cache prongs propagation in trackIndexSkimCreator

fchinu:
While developing #17878, I realised that the CPU time of the track-index skim creator can be reduced significantly by building each track once per collision, reducing the `track_as`, `getTrackParCov` calls and, if needed, the propagation to the collision's primary vertex, and reusing it in the pair and triplet loops.
In the tests I did locally on central Pb-Pb data, I see a reduction in CPU time by a factor ~2x:
[performanceMetricsCached.json](https://github.com/user-attachments/files/32192168/performanceMetricsCached.json)
[performanceMetricsDefault.json](https://github.com/user-attachments/files/32192169/performanceMetricsDefault.json)


vkucera:
Hi @fchinu , thanks for the development. I have a few doubts/questions.

- I don't see how the caching can help for two-prong candidates where there should be no duplicated calculation.
- For the three-prong candidates, it seems to me that you are calculating the track properties for all tracks upfront while the current implementation does the calculations only after applying the track selection.

fchinu:
Hi @vkucera, thanks for having a look! I would say that it helps also for 2 prongs, since the two loops are nested, and the negative prong information is extracted once per positive track. 

https://github.com/AliceO2Group/O2Physics/blob/abd50728fa350bd5b3eb7e28aa78af74b76cce1a/PWGHF/TableProducer/trackIndexSkimCreator.cxx#L2401-L2418

Maybe I am missing something? For the 3 prongs, it is true that we calculate everything upfront without applying preselections, but it seems like that calculating the properties only once per collision is faster than once per triplet being built.
One thing worth mentioning is that the tests were ran by considering only Ds and D+, but I have now ran tests with all candidates and only for 2p, and there is an improvement in all the cases:
All channels: default: 3428s, cached: 1938s, ratio: 0.57
2P only: 	default: 727s cached: 569s, ratio: 0.78
As one would expect, the effect is smaller for 2p, since there is less combinatorics there, and the amount of duplicated work is lower, but in the end we produce derived data also considering 3p, where the effect is larger

vkucera:
Hi @fchinu , you are right, thanks for the clarification. Indeed, there is also duplication for the negative tracks in the 2-prong loop.
From the graphs, there seems to be no obvious memory increase, even though, AFAIU, you duplicate the grouped track table and the track-collision association table for each collision. If you can validate the output as well, it would be good to go.
Very nice improvement! Thanks again!

vkucera (approved):

fchinu:
Thanks a lot @vkucera for approving the PR! For bookkeeping I write here what we discussed earlier:
The outputs are validated, and show no difference in the tables produced by the track index skim creator. Indeed, it is very nice to see that there is no visible increase in memory consumption, maybe that could be due to how tracks retrieval from the tables is performed in the framework, though I'm not fully sure

vkucera:
@fgrosa Please don't enable the auto-merge before the tests finish. Now the result of the code check cannot be reported.

fchinu:
Hi, today I ran a test on hyperloop using today's tag (with the implementation of caching) and with yesterday's tag (without caching) and indeed it seems that caching is doing its job, throughput increased by a factor ~2x: (left with caching, right without caching)
image

PR #17912, opened 2026-09-14
[Common] pidTPCModule::createNetworkPrediction(): refactor and add NN version 5

lubynets:
**1. Refactor `pidTPCModule::createNetworkPrediction()`:**
* follow naming conventions;
* replace magic numbers with named constants;
* get rid of code repetition;
* add `const` qualifier where appropriate
This part does not have intention to change the observable behavior.

**2. Introduce NN versions "dictionary"**
* add `fatal` error if NN name and number if input features do not match any entry in the dictionary.

**3. Introduce NN version 5**
which is similar to the version 4, but the azimuthal angle is recalculated to the value at TPC entrance.


lubynets:
Dear codeowners, could you please review the PR?

dsekihat (approved):

PR #17920, opened 2026-09-15
[PWGHF] Add MC species templates for eID in taskSingleElectron

sejeong8:
Add TPC nSigma(e) vs pT per MC truth species (e, pi, K, p, other) after track selection and after TOF selection, as templates for the eID fit.


vkucera:
Please fix the title. Also notice that the tag prefix is added automatically.

sejeong8:
> Please fix the title. Also notice that the tag prefix is added automatically.

Fixed, thank you.

apalasciano (approved):

PR #17931, opened 2026-09-16
[PWGCF] Fix q-vector and event plane calculation in femto frame work

ariedel-cern:
- Use the configured harmonic for |q_n| (previously always used n=2)
- Allow the event plane angle and the q-vector to come from different detectors
- Guard against out-of-range harmonic indices and non-positive FT0 amplitude sums
- Fix histogram axes for event shape variables


alibuild (dismissed):

alibuild (approved):

PR #17953, opened 2026-09-17
[PWGLF] NucleiTask - Change reco condition on MC

giovannimalfattore:


giovannimalfattore:
@maciacco @mpuccio can you have a look? it seems an issue with MegaLinter

mpuccio (approved):

PR #17955, opened 2026-09-17
[PWGHF] Add DCAFitter configs and PID/pt info for B->JPsi channels

fgrosa:


alibuild (approved):

PR #17962, opened 2026-09-18
[PWGEM] Add the possibility to check ITS-TPC matching

rbailhac:
Add the possibility to check ITS-TPC matching efficiency. Later are exactly the same lines of code if ITS standalone are not included: 
 if (!mIncludeITSsa && (!track.hasITS() || !track.hasTPC())) { // track has to be ITS-TPC matched track
      return false;
    }


PWGEM/Dilepton/Core/DielectronCut.h
```diff
@@ -177,10 +177,6 @@ class DielectronCut : public TNamed
   template <bool dont_require_pteta = false, typename TTrack>
   bool IsSelectedTrack(TTrack const& track) const
   {
-    if (!track.hasITS() || !track.hasTPC()) {
```
feisenhu, line 180: I agree with the PR to remove  these lines.
As there is a check, requiring a ITS and TPC hit if standalone ITS tracks not used [1], this should not have an effect for analysis in which the flag is not used.

For future PRs, there is still some need to introduce a TPC only case.
And currently there is no check for hasITS() when ITS standalone tracks are used.

[1] Line 238-240

feisenhu (approved):

PR #17963, opened 2026-09-18
[PWGCF] Add multiplicity estimator task with pair trigger to femto framework

ariedel-cern:
Adding a new task and the surrounding infrastructure for estimating event multiplicity under the condition that a specific particle pair is present.


alibuild (approved):

PR #17965, opened 2026-09-18
[ALICE3] Move configuration index to be a struct member in otf tracker

jesgum:
Move `icfg` to be a struct member that is set when a new configuration is processed, removing the need to passing it as an argument in most function calls


alibuild (approved):

PR #17970, opened 2026-09-19
[PWGCF] Fix bugs for RecSim mode and add multiparticle correlation calculation

meiALICE:
Fix bugs on filling histograms in RecSim mode.

Add two and recursion for multiparticle correlation. 

Renaming things and other improvements.


abilandz (approved):
Please address the remaining suggestions from code-check along the next PR. This one is approved now!

PR #17971, opened 2026-09-19
[PWGJE] TOF-veto selection fixed

arvindkhuntia:


PWGJE/Tasks/nucleiInJets.cxx
```diff
@@ -892,6 +941,9 @@ struct nucleiInJets {
     float jetPt = -999.;
 
     if (isWithLeadingJet) {
+      if (!isConeAxisAccepted(leadingJetPtEtaPhi[1])) {
```
nzardosh, line 945: can you check this logic please?
It seems that if you dont find a jet and isWithJetEvents is false, then isWithLeadingJet is still true and you go into the fillTrackInfo and pass it leadingJetPtEtaPhi[1] which has a default value of zero. This gives you a fake jet where there shouldnt be one?

arvindkhuntia: Hi @nzardosh,

Thank you for pointing this out.  Without a valid leading-jet guard, the default {0,0,0} vector could create a fake jet/perpendicular cone when isWithJetEvents=false and no jet is found. I fixed it by adding an explicit valid-leading-jet check. Many thanks.

nzardosh (approved):

nzardosh:
Btw isConeAxisAccepted is not applied in processMCRec but it is in the others

arvindkhuntia:
Currently, I am not using processMCRec process function but for consistency I can add isConeAxisAccepted in the next PR. Many thanks.

PR #17984, opened 2026-09-20
[PWGEM] add new pairQC Task + clean up photonhbt

Steffimro:
New task pairQcTask: analysis of pairs of V0 photon candidates (same event only). The following analyses are possible:

- pair observables before and after the pair cuts: conversion-point distances, opening angle, (dEta, dPhi), leg classes, pointing to the primary vertex
- duplicate-candidate QA: leg-level similarity of candidates of one collision, with the MC classification of duplicate pairs
- cross-pair QA: the two alternative e+e- pairings of the four legs, built with the DCAFitter and with the analytic helix builder
  (Utils/AnalyticV0.h), and the veto derived from them
- MC: pair truth type, fake subtype, ancestry of the parent photons, true conversion points and q_true vs q_reco

added new functions into MCUtilities and PairUtilities, so that multiple tasks can use those functions to analyse certain reconstruction artifacts.

Moved the PairQA out from photonhbt, so that only the correlation function analysis and the MC analysis stays there. Added a new Mixing with V0 Legs only, based on V0Leg information and V0LegXYZ information.


PWGEM/PhotonMeson/Tasks/CMakeLists.txt
```diff
@@ -196,7 +196,7 @@ o2physics_add_dpl_workflow(emcal-mc-task
                     PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::AnalysisCore O2Physics::PWGEMPhotonMesonCore
                     COMPONENT_NAME Analysis)
 
-o2physics_add_dpl_workflow(event-normalization-task
-                    SOURCES eventNormalizationTask.cxx
-                    PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::AnalysisCore
+o2physics_add_dpl_workflow(pair-qc-task
+                    SOURCES pairQcTask.cxx
+                    PUBLIC_LINK_LIBRARIES O2::Framework O2Physics::AnalysisCore O2Physics::PWGEMPhotonMesonCore
```
mhemmer-cern, line 206: it looks like you are removing a task from the CMakeList here. Please check again carefully and fix if needed.

mhemmer-cern (changes requested):
Please check the CMakeList changes again and while you are at it take a look at the code-check results to at least fix the introduced errors please.

PWGEM/PhotonMeson/Tasks/CMakeLists.txt
Steffimro: Thanks, I re-included it. Issue was because of older O2Physics version on my branch.

PWGEM/PhotonMeson/Utils/MCUtilities.h
```diff
@@ -18,17 +18,19 @@
...
 
 #include <TMCProcess.h>
 #include <TPDGCode.h>
 
 #include <algorithm>
+#include <array>
 #include <cmath>
 #include <cstdint>
 #include <cstdlib>
-#include <ranges>
```
mhemmer-cern, line 31: ranges is actually used in this header file for `std::ranges::input_range`. Could you add it back in please

mhemmer-cern (changes requested):

mhemmer-cern (approved):
Looks good to me now

mhemmer-cern:
Since code check errors found here were fixed I will enable auto-merge, so this can make it into tomorrows tag without needing to wait for the builds

